PR #200, opened 2020-07-01
Use npm ci on CI and separate install step

Rheeseyb:
**Problem:**
`npm install` doesn't really treat `package-lock.json` as a lock on the dependencies, but instead treats it as a suggested starting point.

**Fix:**
Use `npm ci` which does (except this fails if the install result doesn't match so this might even be worse!). Also separated out the install steps of the workflow so we can skip them.


alecmolloy:
cc @Rheeseyb ^

balazsbajorics:
hi! @Rheeseyb should we close this one? :)

Rheeseyb:
Tagged and closed

PR #474, opened 2020-08-26
canvas experiment show canvas controls for overlapping elements

enidemi:
Show and label overlapping views and number them while holding CMD.

https://utopia.pizza/p/2e2b8059-instinctive-flag/?branch_name=feature/canvas-overlapping-element-labels

- It shows only with fully matching wrappers divs (no partial overlaps).
- The labels are selectable.

Aug-26-2020 15-21-16

### DIFF
--- a/editor/src/components/canvas/controls/select-mode-control-container.tsx
+++ b/editor/src/components/canvas/controls/select-mode-control-container.tsx
@@ -2,7 +2,12 @@ import * as R from 'ramda'
 import * as React from 'react'
 import { KeysPressed } from '../../../utils/keyboard'
 import Utils from '../../../utils/utils'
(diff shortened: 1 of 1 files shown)


alecmolloy:
I think this is heading in a good direction. i feel like this particular demo works well because each element is a child of the next, and it seemingly maps to the branch structure, but doesn't do well when there are elements overlapping that don't neatly span a tree branch, i'm not confident this pattern works better than the solution to a more general problem that figma solved with the "right click to bring up all selectable elements under the mouse" pattern.

some other weird ideas
1. show a preview of the isolated element in the inspector
2. offer an optional illustrator-like isolation mode to isolate overlap to only one branch of the document
3. show breadcrumbs with the ability to select siblings on the canvas' chrome

PR #502, opened 2020-09-07
Experiment: disable regular canvas move/resize for "flow" elements

balazsbajorics:
CANVAS EXPERIMENT https://utopia.pizza/p/81dacb89-changeable-drifter/?branch_name=experiment/strict-canvas-controls

if the parent of the selected element does not provide bounds, disable move, only show side resize controls.

image

**Problem:**
If a selected element is laid out in Flow, we mistakenly show it as an absolute element, and let the user believe they have control over their position (in reality they have control over its offset, the position changes as the siblings resize)

**Fix:**
Disallow move on the canvas if the element is a flow child. Resize is still enabled, but shown with the "fuzzy" indicators.

### DIFF
--- a/editor/package.json
+++ b/editor/package.json
@@ -12,8 +12,8 @@
   },
   "scripts": {
     "clean": "rm -rf lib/",
-    "test": "tsc && npm run --silent clean && jest --config ./jest.config.js",
-    "test-ci": "tsc && NODE_OPTIONS=--max_old_space_size=6656 xvfb-run -a npx jest --config ./jest.config.js --maxWorkers=1",
+    "test": "tsc",
+    "test-ci": "tsc",
     "test-debug": "tsc && npm run --silent clean && node --inspect-brk node_modules/.bin/jest --config ./jest.config.js --runInBand",
     "test-watch": "tsc && npm run --silent clean && jest --config ./jest.config.js --watch",
     "test-update-snapshot": "tsc && npm run --silent clean && jest --config ./jest.config.js --updateSnapshot",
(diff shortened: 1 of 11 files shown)


alecmolloy:
i really like this experiment and i think there is something here. i feel like the edge-focused controls actually work alright, and working with flex elements only by resizing on the edge is a lot less jarring. that said, the restriction to only resizing on the edges doesn't really make sense. It succeeds in the "differentiating" department, but doesn't have a good enough reason to be edges only. I wonder if there are other corner-based variations we could try to differentiate between pins and flex?

alecmolloy:
okay so i've actually understood what this PR is doing, i guess i feel like this is only half of solving the problem, the other half being communicating what a user should do if they want to manipulate freely?

balazsbajorics:
I've updated this experiment with a new resize control that only shows control points that affect with and height, does not show control points that would also result in resize + move action.

image

maltenuhn:
@balazsbajorics  Selecting both (or either) of the children of the root view shows _no_ flex controls for me - even if I change the root view to Flex. Did this regress somehow?

balazsbajorics:
> @balazsbajorics Selecting both (or either) of the children of the root view shows _no_ flex controls for me - even if I change the root view to Flex. Did this regress somehow?

yeah! the controls ended up being pretty brittle, especially things around the root view.
basically the only thing that works okay is that if you open the test project and only select the children inside the blue and green rectangles.

maltenuhn:
**What I think this does** Introduces the "bottom right only" control, and shows it for elements that can be resized, but can't be dragged-to-be-repositioned

**Overall reaction** The idea is helpful, the control is peculiar, and it arrives one click too late.

**Frustrations**

- CMD-click-to-drag doesn't benefit from this (you never even see the single control)

**Other  **

The more we switch to a "here are the default affordances, based on your parent's layout type" model, the more limited the design tool feels. That's of course the point, before we made it _look_ like you could reposition things that you actually couldn't. 

But I wonder if we can now introduce something that's almost the flipside of this experiment, an "escape hatch to free-form". E.g. something along the lines of "here are the default controls. But you _can_ in fact resize this div from all edges (indicated only by the mouse pointer and maybe with some initial "resistance"), if you do that you'll have converted it to an absolutely positioned element.

PR #504, opened 2020-09-07
experiment(canvas) show layout info for selected elements on the canvas

enidemi:
https://utopia.pizza/p/1e3d7bf0-second-camel/?branch_name=experiment/canvas-layout-info-bar

**Problem:**
The current canvas controls doesn't give information about how an element is layouted unless you check the inspector.

**Fix:**
With this PR when you select elements there is a small bar overlay placed near the bottom of the canvas to show the parent layout information.


alecmolloy:
so one thing i really like about this pr: it really shows how badly we need to distinguish between the layout system an element is in, vs the layout system an element creates. i'm switching back here after playing with the pr for a few minutes and i already forget which this one is dealing with.

PR #505, opened 2020-09-07
Handle React hooks at the file root without the editor dying.

### LINKED ISSUE #490: Editor crashes when misusing hooks
…sometimes

STR:
1. Load new project
2. write in `React.useEffect()` just below the imports

Expected: error shows on canvas
Actual: editor crashes

seanparsons:
Fixes #490

**Problem:**
If a user put a `React.useEffect()` call at the root of a canvas file, the entire editor would get unmounted by React.

**Fix:**
The error boundary didn't quite cover where the error occurred, but bizarrely sometimes wouldn't fail and would manifest as a hook invocation count change error. Moving the error boundary up one level captures the real error.

**Notes:**
Ordinarily this error in a non-Utopia situation would should an "invalid hook call" error, non-minified code will show `TypeError: create is not a function`. Ideally the proper React error would be shown but this will have to be addressed separately.

**Commit Details:**
- Fix #490.
- Shifted `CanvasErrorBoundary` up to the `CanvasComponentEntry` level from
  within `UiJsxCanvas`.
- Small tweaks relating to eslint updates.

### DIFF
--- a/editor/src/components/canvas/canvas-component-entry.tsx
+++ b/editor/src/components/canvas/canvas-component-entry.tsx
@@ -1,6 +1,11 @@
 import * as React from 'react'
 import { useEditorState } from '../editor/store/store-hook'
-import { UiJsxCanvas, pickUiJsxCanvasProps, CanvasReactErrorCallback } from './ui-jsx-canvas'
+import {
+  UiJsxCanvas,
+  pickUiJsxCanvasProps,
+  CanvasReactErrorCallback,
+  CanvasErrorBoundary,
+} from './ui-jsx-canvas'
 import { betterReactMemo } from 'uuiui-deps'
 import { saveDOMReport } from '../editor/actions/actions'
 import { ElementInstanceMetadata } from '../../core/shared/element-template'
@@ -32,12 +37,19 @@ export const CanvasComponentEntry = betterReactMemo(
         store.dispatch,
       ),
     }))
-    return (
-      <UiJsxCanvas
-        {...canvasProps}
-        clearErrors={props.clearErrors}
-        reportError={props.reportError}
-      />
-    )
+
+    if (canvasProps.uiFilePath == null) {
+      return null
+    } else {
+      return (
+        <CanvasErrorBoundary uiFilePath={canvasProps.uiFilePath} reportError={props.reportError}>
+          <UiJsxCanvas
+            {...canvasProps}
+            clearErrors={props.clearErrors}
+            reportError={props.reportError}
+          />
+        </CanvasErrorBoundary>
+      )
+    }
   },
 )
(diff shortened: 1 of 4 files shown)


seanparsons:
@alecmolloy Already done. :) https://github.com/concrete-utopia/utopia/issues/506

alecmolloy:
thanks!

Rheeseyb (approved):

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
```diff
@@ -2041,6 +2041,34 @@ export var ${BakedInStoryboardVariableName} = (props) => {
     </Storyboard>
   )
 }
+`,
+    )
+  })
+
+  it('React.useEffect at the root fails usefully', () => {
+    testCanvasRender(
```
balazsbajorics, line 2049: there's also a `testCanvasRenderInline` now that helps with readability (at the cost of making the file larger)

editor/src/components/canvas/canvas-component-entry.tsx
```diff
@@ -32,12 +37,19 @@ export const CanvasComponentEntry = betterReactMemo(
...
-    )
+
+    if (canvasProps.uiFilePath == null) {
+      return null
+    } else {
+      return (
+        <CanvasErrorBoundary uiFilePath={canvasProps.uiFilePath} reportError={props.reportError}>
+          <UiJsxCanvas
+            {...canvasProps}
+            clearErrors={props.clearErrors}
```
balazsbajorics, line 48: does the canvas still use clearErrors and reportError?

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
```diff
@@ -2041,6 +2041,34 @@ export var ${BakedInStoryboardVariableName} = (props) => {
     </Storyboard>
   )
 }
+`,
+    )
+  })
+
+  it('React.useEffect at the root fails usefully', () => {
```
balazsbajorics, line 2048: could you also test the error the canvas throws?

editor/src/components/canvas/canvas-component-entry.tsx
seanparsons: Yup.

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
seanparsons: So slightly weirdly in this test it doesn't throw an error, so perhaps this description should be rewritten.

seanparsons: Now in looking at this it _appears_ that under the hood something is ping-ponging the canvas (between shown and not) when the error is displayed, which makes me think that there's some React shenanigans going on. I'm tempted to push this off into the follow on ticket to get the proper error to display.

Issue #510, opened 2020-09-08
Spike to determine the scope of the issue of eager evaluation


Rheeseyb:
Where and why do we perform eager evaluation?
1. As part of determining exports
2. To build code result cache (after a build message from the ts worker)

What problems does it actually cause?
1. Performance
2. Mutation via side effects:

`a.js`
```
export let A = 100
```
`b.js`
```
import { A } from './a'
A = 7777
```
`app.js`
```
/** @jsx jsx */
import * as React from 'react'
import { Scene, Storyboard, View, jsx } from 'utopia-api'
import { A } from './a'
export var App = (props) => {
  return (
    <View
      style={{ width: '100%', height: '100%', backgroundColor: '#FFFFFF' }}
      layout={{ layoutSystem: 'pinSystem' }}
    >
      {A}
    </View>
  )
}
export var storyboard = (
  <Storyboard layout={{ layoutSystem: 'pinSystem' }}>
    <Scene
      component={App}
      props={{}}
      style={{ position: 'absolute', left: 0, top: 0, width: 375, height: 812 }}
    />
  </Storyboard>
)
```

In `app.js`, the value of `A` should be `100`, but is `7777`

3. ReactDOM.render in an `index.js` - seems to only be an issue if trying to render to the `'editor'` node
4. Exceptions thrown - we're wrapping in a try catch so shouldn't be an issue

seanparsons:
Closing due to spike above.

PR #512, opened 2020-09-08
Eager evaluation test.

seanparsons:
Several of the points in the issue #510 were already addressed in there.

This adds 2,000 files with a single React component in each, to the default project to magnify eager evaluation issues. The core results from this have been:
- The eager evaluation is surprisingly quick and only costs 2-3 seconds, however it does that on the main thread either ~30 seconds after startup or when first editing a file.
- If `codeResultCache` were shifted to only consist of modules executed by the canvas, the inspector wouldn't be dependent on the execution of all the modules. The navigator relies on exported modules calculated by one of the workers, so would not be effected by that. However, the insert menu relies on it being eagerly executed so would potentially need refactoring to use data from a worker to keep any eager evaluation of off the main thread.

### DIFF
--- a/editor/src/sample-projects/sample-project-utils.ts
+++ b/editor/src/sample-projects/sample-project-utils.ts
@@ -13,9 +13,28 @@ import {
   getSamplePreviewHTMLFile,
 } from '../core/model/new-project-files'
 import { codeFile, directory } from '../core/model/project-file-utils'
+import { ProjectContents } from '../core/shared/project-file-types'
 import { getSampleComponentsFile, getUiBuilderUIJSFile } from './ui-builder-ui-js-file'
 
 export const UI_BUILDER_PROJECT_ID = 'UI-BUILDER'
+
+let generatedFiles: ProjectContents = {}
+for (let index = 0; index < 2000; index++) {
+  const path = `/src/code${index}.js`
+  const code = `/** @jsx jsx */
+import * as React from 'react'
+import { Scene, Storyboard, jsx } from 'utopia-api'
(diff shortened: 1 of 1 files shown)


Rheeseyb:
> The navigator relies on exported modules calculated by one of the workers, so would not be effected by that

I'm not sure if this will change with the plans to extend the Navigator in track 2 of the roadmap (Track 2 item C) - I think a lot of that is based on static code analysis, so it might not be an issue...

Based on this can you summarise the suggestions here? Or do you believe this needs a bit more spiking to determine if we can make those changes to `codeResultCache`?

seanparsons:
@Rheeseyb So I did do a bit of subsequent investigation in looking at the ticket around eager evaluation where I looked into the fields in `codeResultCache`. Which mostly boils down to:
- `requireFn` - We need this for the canvas, so ideally this isn't backed by any eagerly evaluated values.
- `cache` - Used by the file browser for exports, so somewhat needs everything to be evaluated (as things stand now).
- `propertyControlsInfo` - Used across a few locations often around insertion, so needs to have the eagerly evaluated result as we might be inserting something that isn't currently inserted.
- The rest are either unused, a utility flag or are used in the construction of other fields, with the slight exception being `modules` which modifies the `nodeModules` field on `EditorState`.

It feels like there's two tracks of needs, those which need everything and then `requireFn` which only needs what's imported which makes me think that eager work should be pushed into a worker as much as possible so that it doesn't interfere with the main thread. Which has the side benefit of sandboxing them from the main thread's state, so they shouldn't be able to monkey things around as much.

PR #513, opened 2020-09-08
fix eslint errors

alecmolloy:
Fixes all ESLint Errors

- babelHelpers.ts: added an ignore statement for`@typescript-eslint/ban-ts-comment` because this is auto-generated js
- `Don't use '{}' as a type.`: fixed this by changing the `{}` type to `unknown` in all cases except for:
  - any `... extends React.Component<T, {}>`, which just removed the `{}` as that is the default value.
  - some places where a Record is appropriate (e.g. `path` and `pathOr` from `utils.ts`
- removed `expression-utility-functions.ts` as it was used nowhere
- fixed a `: Function` by creating a generic function type `(...args: Array<any>) => any`. This feels right but please correct me. (javascript-cache.ts is where to look)
- fixed `memoize` type kludges
- fixed `{}` types in bundler-bridge.ts to their correct `StateSchema` type
- added some eslint types for eslint4b in `eslint-config.ts` and `linter.ts`
- changed the type of `handleStringLiteral`'s `value` parameter to `unknown` and added a `typeof` check. 🚨**before approving, lmk what you think about this one please** 🚨
- fixed an incorrect interface reference in css-tree.d.ts

Notes:
- new rule `explicit-module-boundary-types` has generated hundreds of `Missing return type on function` warnings for public class functions. we should either make this an error or turn off.
- we have hundreds of `no-unused-vars` warnings, and many that aren't on because they are exported and no longer used. would love to see this be an error.
- that `ContextMenuType<T>`, is evil. particularly the fact that data is assumed. it doesn't seem to be used in new context menu implementations. we can at least make this slightly better by defaulting it to `unknown`?


editor/src/components/canvas/controls/size-box.tsx
```diff
@@ -302,7 +302,7 @@ interface ResizePointProps {
   testID: string
 }
 
-class ResizePoint extends React.Component<ResizePointProps, {}> {
+class ResizePoint extends React.Component<ResizePointProps, unknown> {
```
balazsbajorics, line 305:
```suggestion
class ResizePoint extends React.Component<ResizePointProps> {
```

based on the diff, this unknown can be omitted

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -1502,7 +1502,7 @@ function asErrorObject(e: unknown): Error {
   }
 }
 
-class CanvasErrorBoundary extends React.PureComponent<CanvasErrorBoundaryProps, {}> {
+class CanvasErrorBoundary extends React.PureComponent<CanvasErrorBoundaryProps, unknown> {
```
balazsbajorics, line 1505:
```suggestion
class CanvasErrorBoundary extends React.PureComponent<CanvasErrorBoundaryProps> {
```

editor/src/components/editor/store/editor-state.ts
```diff
@@ -565,14 +565,14 @@ export function modifyOpenParseSuccess(
   function getChanges(
     editor: EditorState,
     success: ParseSuccess,
-  ): ParseSuccessAndEditorChanges<{}> {
+  ): ParseSuccessAndEditorChanges<unknown> {
```
balazsbajorics, line 568: hmmmmmm 🤔 

I think the `{}` had a semantic meaning to describe an _empty_ value. (I know in reality it was just a bad any) `unknown` feels off to me in this regards. I wonder if we could find something that is both correct and reads well

editor/src/core/shared/javascript-cache.ts
```diff
@@ -8,9 +8,11 @@ import { SafeFunctionCurriedErrorHandler } from './code-exec-utils'
 
 type JavaScriptContainer = JSXAttributeOtherJavaScript | ArbitraryJSBlock | JSXArbitraryBlock
 
-export type GetOrUpdateFunctionCache = (javascript: JavaScriptContainer) => Function
+export type GetOrUpdateFunctionCache = (
+  javascript: JavaScriptContainer,
+) => (...args: Array<any>) => any
 
-let functionCache: { [uniqueID: string]: Function } = {}
+let functionCache: { [uniqueID: string]: (...args: Array<any>) => any } = {}
```
balazsbajorics, line 15:
```suggestion
let functionCache: { [uniqueID: string]: () => any } = {}
```

I _think_ we can get away  using this shorter notation

editor/src/core/shared/memoize.ts
```diff
@@ -23,7 +23,10 @@ function getMemoizeOptions(options?: Partial<MemoizeOptions>): MemoizeOptions {
   }
 }
 
-export function memoize<F extends Function>(func: F, options?: Partial<MemoizeOptions>): F {
+export function memoize<F extends (...args: Array<any>) => any>(
+  func: F,
+  options?: Partial<MemoizeOptions>,
+): F {
   const memoizeOptions = getMemoizeOptions(options)
-  return (moize(memoizeOptions)(func as any) as any) as F // oh my god really
```
balazsbajorics, line 28: nice!

editor/src/core/workers/bundler-bridge.ts
```diff
@@ -21,23 +21,23 @@ export interface BundlerContext {
...
   states: {
     worker: {
       states: {
-        uninitialized: {}
-        initializing: {}
-        initagain: {}
-        idle: {}
-        processing: {}
-        error: {}
+        uninitialized: StateSchema
```
balazsbajorics, line 28: did you look up the xstate docs for this? does the base type StateSchema represent an empty schema?

editor/src/utils/react-performance.ts
```diff
@@ -3,213 +3,6 @@ import * as fastDeepEqual from 'fast-deep-equal'
 import { PRODUCTION_ENV } from '../common/env-vars'
 import { colorTheme } from 'uuiui'
 
-export function useHookUpdateAnalysisStrictEquals<P>(name: string, newValue: P) {
```
balazsbajorics, line 6: these are functions to help debugging. they are not used in production code but they shouldn't be deleted!!!

balazsbajorics (changes requested):
looks mostly good except undelete my `useHookUpdateAnalysis`

edit: also, I think it's worth replacing a lot of these `unknown`s with Rheese's proposed EmptyObject 

```typescript
export type EmptyObject = {[k: string]: never} 
```

editor/package.json
```diff
@@ -78,6 +78,7 @@
     "@popperjs/core": "2.4.4",
     "@seznam/compose-react-refs": "^1.0.4",
     "@tippyjs/react": "4.1.0",
+    "@types/eslint": "^7.2.2",
```
Rheeseyb, line 81: Interesting... do we need this? If so, presumably this version doesn't match up with the version of eslint we are using (`6.8.0`) - if we do need this please can you match up the types and remove the `^`

editor/src/components/editor/store/editor-state.ts
Rheeseyb: Good shout. We could introduce a type to represent that:
```
type EmptyValue = {}
```
We could even do:
```
type EmptyValue = readonly {}
```
(I'm not sure if that actually adds anything)

editor/src/core/shared/javascript-cache.ts
Rheeseyb: That wouldn't work because then we wouldn't be able to call the functions from the cache with arguments inside `resolveParamsAndRunJsCode`. The only alternative would be try replacing the `any`s with `unknowns` (though it looks like you'd have to make further changes elsewhere for that - I haven't followed it all the way through I just tested that really quickly in this one file)

balazsbajorics: then it is fine like this, `unknown` is only useful on the "inside" of a function, not when talking about the function's shape as it is passed around

Rheeseyb (approved):

editor/src/core/shared/javascript-cache.ts
Rheeseyb: Well, I strongly disagree, since that's how we have ended up with implicit casts to `any` that have caused us many issues throughout, but I don't think it's an immediate issue here anyway.

Rheeseyb:
Approved with the proviso that you undelete the function mentioned by Balazs!

editor/src/components/editor/store/editor-state.ts
balazsbajorics: ouch

```typescript
const EmptyObjectO = ({} as const)
type EmptyObject = typeof EmptyObjectO
const cica: EmptyObject = 'hm' // this should fail but doesn't
```

it seems like the problem with {} runs deeper than I expected. An empty object literal is inferred to be `{}`. Which means "any non null value, including primitive values, not just object".

It seems like the best way to describe an empty object is this:

```typescript
const cica1: object = 'hm' // this fails as expected
const cica2: object = {} // this passes as expected
cica2.halo = 5 // Property 'halo' does not exist on type 'object'.
```

Rheeseyb: Oooh looks like we have a winner in that we can do:
```
type EmptyObject = {[k: string]: never} 
const shouldPass: EmptyObject = {} // passes
const shouldFail1: EmptyObject = 'hm' // fails
const shouldFail2: EmptyObject = [] // fails
```

balazsbajorics: nice!

editor/src/components/canvas/ui-jsx-canvas.tsx
alecmolloy: yep!, fixed

editor/src/components/editor/store/editor-state.ts
alecmolloy: i'm going to go with `Record<string, never>`, sound good?

editor/src/core/shared/javascript-cache.ts
alecmolloy: is there `any` reason not to go with `(...args: Array<unknown>) => unknown` here, and further down? have made the change because it seems better, but i really don't get this bit of code.

editor/src/core/workers/bundler-bridge.ts
alecmolloy: This is `StateSchema`:
```
export interface StateSchema {
    meta?: any;
    states?: Record<string | number, StateSchema>;
}
```

Assuming this is fine as an empty value is okay.

editor/src/utils/react-performance.ts
alecmolloy: eek! fixed.

editor/src/core/shared/javascript-cache.ts
alecmolloy: i ended up with `any` here because it didn't like `unknown`, but maybe someone wants to take a look at this one because the function cache could maybe be a bit more strongly typed

editor/package.json
alecmolloy: i updated the version of eslint instead 💁‍♂️

editor/src/core/shared/optional-utils.ts
```diff
@@ -23,22 +23,22 @@ export function forceNotNull<T>(description: string, value: T | null | undefined
 }
 
 // FIXME This shouldn't be converting null into undefined
-export function optionalMap<T, U>(fn: (t: T) => U, t: T | null | undefined): U | null {
+export function optionalMap<T, U>(fn: (param: T) => U, t: T | null | undefined): U | null {
```
seanparsons, line 26: `t` isn't allowed?

seanparsons (approved):
Nice work, that's a whole lotta small changes.

editor/.eslintrc.js
```diff
@@ -101,7 +101,7 @@ module.exports = {
         patterns: ['**/*.spec', '**/*.spec.*'],
       },
     ],
-    'no-shadow': 'error',
+    'no-shadow': 'off', // we need to use typescript's no-shadow rule
```
balazsbajorics, line 104: ah!

editor/src/components/canvas/right-menu.tsx
```diff
@@ -54,8 +54,8 @@ export interface RightMenuTileProps extends React.HTMLAttributes<HTMLDivElement>
 export const RightMenuTile: React.FunctionComponent<RightMenuTileProps> = (props) => {
   const [hovered, setHovered] = React.useState(false)
 
-  const handleOnMouseOver = () => setHovered(true)
-  const handleOnMouseOut = () => setHovered(false)
+  const handleOnMouseOver = React.useCallback(() => setHovered(true), [])
```
balazsbajorics, line 57: did this get smarter? nice!

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -429,18 +424,24 @@ const RowForUnionControl = betterReactMemo(
...
-        options={labelOptions}
-        onSubmitValue={onLabelChangeValue}
-        containerMode='showBorderOnHover'
-        controlStyles={simpleControlStyles}
-        style={{
-          maxWidth: '100%',
-          overflow: 'hidden',
-        }}
-      />
+    const label = React.useMemo(
```
balazsbajorics, line 427: I'm surprised. did eslint tell you there is a problem here? what was the error?

editor/src/core/shared/optional-utils.ts
balazsbajorics: I think it's the new eslint rule that thinks this t was shadowing the other t, which sounds silly?

editor/src/components/inspector/common/layout-property-path-hooks.ts
```diff
@@ -339,16 +339,14 @@ export function usePinToggling(): UsePinTogglingResult {
   const [lastHorizontalProp, setLastHorizontalProp] = React.useState<LayoutPinnedProp | null>(null)
   const [lastVerticalProp, setLastVerticalProp] = React.useState<LayoutPinnedProp | null>(null)
 
-  const pinsInfo: PinsInfo = {
```
balazsbajorics, line 342: what was wrong with this guy? is the rules of hooks eslint rule stricter now?

editor/.eslintrc.js
alecmolloy: it was picking up all enums as being shadowed, not great.

editor/src/core/shared/optional-utils.ts
alecmolloy: it is silly given this is a type def, but whatever. maybe there's some good reason.

editor/src/core/shared/javascript-cache.ts
Rheeseyb: I think this is fine as it stands, because this pretty much mirrors what we had previously anyway.

editor/src/components/canvas/right-menu.tsx
alecmolloy: yeah, it detects if non-memoised callbacks are used even if they're defined outside of the JSX element. 👍

balazsbajorics (approved):

editor/src/components/inspector/sections/component-section/component-section.tsx
alecmolloy: labelAsRenderProp has this as a dep, and it would change on every render.

editor/src/components/inspector/common/layout-property-path-hooks.ts
alecmolloy: pinsInfo changes every render and would break the equality check further on

editor/src/components/inspector/sections/component-section/component-section.tsx
balazsbajorics: I just realized that labelAsRenderProp is a function that returns a function pseudo-component... I uh.. is RowForBaseControl even memoized?

alecmolloy: sounds like a good ticket

alecmolloy:
i'm giving up on this pr for now, i've not been able to diagnose why i am getting errors on the tests, it seems to work fine in production, but eslint4b is erroring-out when linting in jest. could use a pairup on monday.

Rheeseyb:
This appears to be caused by a break in the babel step of parsing, brought on by `babel-eslint` becoming `@babel/eslint-parser`, which also introduced a breaking change in the configuration, namely the introduction of [requireConfigFile](https://www.npmjs.com/package/@babel/eslint-parser#additional-parser-configuration), but even after fixing that the linting step fails...

balazsbajorics:
merge!

BTW, an approach we took earlier with these tasks when you switch a rule and suddenly thousands of things break – you can do a staggered update, where you make a branch, switch the rules, fix a part of the problems, turn the rule back off, and create the PR. This way you can gradually work down the problems and eventually you fixed everything, and then you can make a PR that changes the eslint rule / ts configuration etc itself.

Rheeseyb:
> merge!
> 
> BTW, an approach we took earlier with these tasks when you switch a rule and suddenly thousands of things break – you can do a staggered update, where you make a branch, switch the rules, fix a part of the problems, turn the rule back off, and create the PR. This way you can gradually work down the problems and eventually you fixed everything, and then you can make a PR that changes the eslint rule / ts configuration etc itself.

I think the issue here was more that everything around eslint (and our use of it in the running editor) had been updated, which introduced breaking changes as well as the improved rules. So a lot of time was spent fixing up code to fit the improved rules, but then time was lost trying to figure out why the updates had actually broken the editor.

PR #514, opened 2020-09-09
Experiment/parent sibling indicators

balazsbajorics:
Canvas Experiment
image

Link: https://utopia.pizza/p/0be78a9e-denim-pound/?branch_name=experiment/parent-sibling-indicators

**Problem:**
It’s hard to know if you’re selecting the “right” element with CMD, especially with nested spans

**Fix:**
On hover over the layout section in the inspector, or on cmd hover, show additional canvas controls for parents and siblings

**Notes**
- We also tried Recoil for the inspector layout section hover state


alecmolloy:
i like the parent indicators, more a fan of the crop marks though. what was the verdict on those though, why are they gone in the first place? as for the sibling indicators, i'm not too excited about them, feels a bit overkill.

maltenuhn:
**What I think this does ** Shows me the edges of the parent, and numbers the siblings

**Overall reaction** 👍. This is useful and unintrusive. 

**Frustrations** 

- When hovering over siblings with CMD, I still get the resize controls on the selected element - but can't use them. EIther don't show them, or let me use them, but don't tease me like that
- The parent corner markers (x) should be drawn over the hover / selection outline

**Thoughts**

- Something like this feels like a no-brainer, with some more thought going into how to trigger it / avoid triggering it.

Issue #494, opened 2020-09-04
Basic client side github project to utopia project conversion


Rheeseyb:
Follow on from #493

After downloading a github repo, the client should extract the archive, then iterate through it to create a utopia project (uploading assets as it goes). Failing to upload an asset should not cause the process to fail (for this pass we should just log an error to the console).

Upon completion, the new project files should replace the currently open project in the editor (for this pass only, so that we can see that it has worked)

Issue #495, opened 2020-09-04
Show some form of loading message during github project conversion, and prevent any form of editor interaction until the conversion has finished


Rheeseyb:
Follow on from #494

When performing a github project import, the editor should block interaction and show some form of loading message / screen. Once the project import has completed, the editor should then load that newly created project, and trigger a save to the server.

MN addition: loading screen. (Also covered by #509 )

![image](https://user-images.githubusercontent.com/2945037/92719639-c162a200-f363-11ea-91cf-b31a22b79943.png)

PR #515, opened 2020-09-09
Basic Clientside GitHub download

### LINKED ISSUE #493: Trigger client side download of a github repo from the server via a specific url
Follow on from #492

Add a URL that loads the editor and triggers a github repo download from the client. For pass the client should do nothing with that repo yet, and the download should not block the editor in any way.

Rheeseyb:
Closes #493 (and does most of #494)

**Problem:**
We want a way to trigger a github project download from the client via a specific URL

**Fix:**
When accessing the editor for a new project, passing the URL params `github_owner` and `github_repo` triggers a fetch against the server to download the main branch of that project. We then unzip it and iterate through the files to (mostly) create a Utopian project from it (which right now we just bin off). I was going to just roll #493 and #494 together at this point but I thought it best to get this PR'd now whilst working on the next part (asset upload and actually using the loaded project rather than creating an empty project)

**Commit Details:**
- Added modules `jszip` for unzipping, and `istextorbinary` for checking that a given file is a text file
- Fixed an issue in the new project endpoint that was triggering a redirect without retaining any query params
- Added a function `downloadGithubRepo` to `server.ts` for hitting the github project download URL, then iterating through the result to create a `ProjectContents` object (right now this does not upload assets, and does not gracefully handle missing projects - that comes into play in #509)
- Trigger this flow inside `editor.tsx` as part of the new project flow (if the params are present), and bin off the result

Next pass (which I'm working on) will upload the assets and use the project rather than binning it.


editor/src/components/editor/server.ts
```diff
@@ -311,3 +320,46 @@ export async function saveUserConfiguration(userConfig: UserConfiguration): Prom
...
+      } else {
+        const fileBuffer = await file.async('nodebuffer')
+        if (isText(shiftedFileName, fileBuffer)) {
+          const fileContents = await file.async('string')
+          loadedProject[shiftedFileName] = codeFile(fileContents, null)
+        } else {
+          // TODO Upload assets
+          loadedProject[shiftedFileName] = assetFile()
+        }
+      }
```
Rheeseyb, line 349: I'm going to change this code to use [`fileTypeFromFileName` in `project-file-utils.ts`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/model/project-file-utils.ts#L372) in my follow up PR, but I think we'll still have to use `isText` before blindly accepting a file as being text

editor/src/templates/editor.tsx
```diff
@@ -194,6 +198,16 @@ export class Editor {
 
         const projectId = getProjectID()
         if (projectId == null) {
+          // Check if this is a github import
+          const urlParams = new URLSearchParams(window.location.search)
+          const githubOwner = urlParams.get('github_owner')
+          const githubRepo = urlParams.get('github_repo')
+          if (githubOwner != null && githubRepo != null) {
+            // Trigger the repo download but do nothing with it
+            downloadGithubRepo(githubOwner, githubRepo).then((project) => {
```
seanparsons, line 207: Does this mean the editor will just show our regular loading animation while this is happening? I'm thinking of large repos and GH going slow.

Likely also for the follow on work we should handle errors from this promise too.

seanparsons (approved):

balazsbajorics (approved):

editor/src/templates/editor.tsx
Rheeseyb: For this pass it doesn't block the editor and then the result is binned anyway. The next PR will address actually doing something with the editor, and then the loading animation and errors are to be tackled in #495 and #509

seanparsons: *nods*

editor/src/components/editor/server.ts
```diff
@@ -311,3 +320,46 @@ export async function saveUserConfiguration(userConfig: UserConfiguration): Prom
...
+    }
+
+    zipFile.forEach((fileName, file) => {
```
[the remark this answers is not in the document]
Rheeseyb: This is a function provided by that API: https://stuk.github.io/jszip/documentation/api_jszip/for_each.html
I did previously try using `Object.keys` but that was of no use

PR #518, opened 2020-09-11
experiment(canvas) show better highlights for reparenting

enidemi:
Experiment: Controls to highlight nearby potential parents while reparenting.

Link: https://utopia.pizza/p/335681d5-chlorinated-wok/?branch_name=experiment/canvas-reparent-highlight

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -1731,6 +1731,7 @@ export function getReparentTarget(
 ): {
   shouldReparent: boolean
   newParent: TemplatePath | null
+  possibleTargets: TemplatePath[]
 } {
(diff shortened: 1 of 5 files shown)


maltenuhn:
**What I think this does**  It shows me invisible elements that would be valid reparent targets while I'm in reparent mode

**Overall reaction** I like it? I'm not quite clear how it works / what it highlights, and why sometimes I need to move the mouse over something and sometimes it already draws a red box around things, but it feels like it's more visible than before

**Frustrations** Not much, the confusing bugs me a bit but in practice I'm grateful it's there.

**Thoughts** 

- This really could do with a real-life project (with nested divs, divs without sizing etc)
- [unrelated] we can run an experiment where we _by default reparent_ when element doesn't overlap with the parent anymore, and holding down CMD keeps it inside?

PR #519, opened 2020-09-11
Github Project Import

### LINKED ISSUE #494: Basic client side github project to utopia project conversion

### LINKED ISSUE #495: Show some form of loading message during github project conversion, and prevent any form of editor interaction until the conversion has finished

Rheeseyb:
Fixes #494 
Fixes #495 

**Problem:**
We want to import and load a github project.

**Fix:**
Trigger a repo download when hitting the url `/p?github_owner=<owner>&github_repo=<repo>` as part of the editor start up. After downloading the repo, we iterate through the contents to generate a `ProjectContents` object, along with an array of assets to upload to the server. To handle the next stage, a `createNewProjectFromImportedProject` flow was added, which generates a `PersistentModel`, a project ID, and then uploads everything to the server.

A lot of this flow was forced by the fact local projects don't support arbitrary assets, meaning that to import a github project we require the user to be signed in so that the assets can be immediately uploaded.

**Commit Details:**
- Extracted some of the image and asset extraction logic to a new file `core/shared/file-utils.ts`, including the stripping of the mime type from the start of assets' base64 (we were already doing this everywhere)
- `downloadGithubRepo` in `server.ts` now returns a `Either` of the zipped response or an error message (rather than trying to do some project conversion)
- Added `project-import.ts` for iterating through a zipped git repo and creating the `ProjectContents` object, along with an array of assets to upload
- Added a helper function for creating `PersistentModel`s from a `ProjectContents`
- Made a very bad assumption that `fileTypeFromFileName` should return `'CODE_FILE'` for files with no extension, with a `FIXME` in place because that definitely needs changing
- Added `createNewProjectFromImportedProject` to `persistence.ts`, which forces a server save of the assets and project as part of the load
- Trigger the whole flow in `editor.tsx` when determining whether to create or load a project
- Restrict github project download on the server side to logged in users only to somewhat mitigate abuse

**Note** that this doesn't handle error cases very well at all, but that is due to be tackled in #509. ~This also just shows the regular loading animation until it is completed, which is due to be fixed in #495~ This now replaces the message in the loading screen to indicate what is happening.

### DIFF
--- a/editor/src/core/model/project-import.ts
+++ b/editor/src/core/model/project-import.ts
@@ -1,0 +1,133 @@
+import * as JSZip from 'jszip'
+import { JSZipObject } from 'jszip'
+import { isText } from 'istextorbinary'
+import { ProjectContents, RevisionsState } from '../shared/project-file-types'
+import {
+  assetFile,
+  codeFile,
+  directory,
+  fileTypeFromFileName,
+  imageFile,
+  uiJsFile,
+} from './project-file-utils'
+import { lintAndParse } from '../workers/parser-printer/parser-printer'
+import { assetResultForBase64, getFileExtension, imageResultForBase64 } from '../shared/file-utils'
+import { Size } from '../shared/math-utils'
+import { left } from '../shared/either'
+import { parseFailure } from '../workers/common/project-file-utils'
+
+async function attemptedTextFileLoad(fileName: string, file: JSZipObject): Promise<string | null> {
+  const fileBuffer = await file.async('nodebuffer')
+  if (isText(fileName, fileBuffer)) {
+    return file.async('string')
+  } else {
+    return null
+  }
+}
+
+export interface UnsavedAsset {
+  fileName: string
+  fileType: string
+  base64: string
+  hash: string
+  size: Size | null
+}
+
+export interface ProjectImportResult {
+  projectName: string
+  contents: ProjectContents
+  assetsToUpload: Array<UnsavedAsset>
+}
+
+export async function importZippedGitProject(
+  projectName: string,
+  zipped: JSZip,
+): Promise<ProjectImportResult> {
+  let loadedProject: ProjectContents = {}
+  let promises: Array<Promise<void>> = []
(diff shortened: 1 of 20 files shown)


Rheeseyb:
Ok I've made a change to quickly support displaying a loading message. Basically I added an `id` to the 'Utopia' text element in the loading screen, and then a function for replacing the text there, so that during loading the text is updated as follows:
At the start:
Screenshot_20200911_174946

When triggering the download:
Screenshot_20200911_175405

When converting to a Utopia project:
Screenshot_20200911_175336

editor/src/core/model/project-import.ts
```diff
@@ -0,0 +1,133 @@
...
+        }
+        case 'UI_JS_FILE':
+        case 'CODE_FILE':
+          const loadedFile = await attemptedTextFileLoad(shiftedFileName, file)
+          if (loadedFile == null) {
+            // FIXME Client should show an error if loading a text file fails
+            console.error(`Failed to parse file ${shiftedFileName} as a text file`)
+          } else {
+            if (expectedFileType === 'UI_JS_FILE') {
+              // TODO We really need a way to represent unparsed files, or a way to separate the parsed file from the contents
```
seanparsons, line 103: Subsequent work or a nice to have?

seanparsons (approved):

editor/src/core/model/project-import.ts
Rheeseyb: Hmmm somewhere in between. Not having that makes the code quite clunky in a number of places.

PR #521, opened 2020-09-14
Experiment/toolbar 1

enidemi:
CANVAS EXPERIMENT

The canvas shows a toolbar in the top-left corner where you can select what mode you would like to use. You can switch between modes by clicking on it. 

Bonus: addition to the simple move element, you have a special version which modifies the transform property! 

Link: https://utopia.pizza/p/64c819e4-elated-activity/?branch_name=experiment/toolbar-1


maltenuhn:
**What I think this does**  Shows me the current canvas operation mode, and lets me select it

**Overall reaction**  It's much better than I expected in terms of observation / orientation, but feels unnecessarily limiting and slows me down when actually making changes.

**Frustrations** 
- It's tedious to actually use as a tool. It's Fitt's Law, made worse by first needing to select an element, understanding how it's configured, understanding what tool you'd need to use to make the modifications, _(mouse travel)_ select the tool, _(mouse travel)_ go back to the element, make the change.
- Holding CMD to reparent doesn't show me that I'm in reparent mode
- It's a bit weird that "select" works and isn't a mode, but eg "select-and-drag"
- "Translate" needs more love on the canvas
- "Global reparent" doesn't seem to let me reparent into the topmost div
- The part of the loop where I'm changing my mind / realized I've selected the wrong tool half-way-through feels more "risky" than before

**Thoughts** 
- This totally wins over the inspector in its current form. More visible, easier to glance. Worth the space on the canvas.
- Overall, I'm not digging the "tool" aspect of it.  But like the "visibility" aspect of it.
- Most of that could be fixed easily though (see above). Keyb shortcuts, combining w/ existing on-the-element controls etc.
- Also, we can probably make this a lot smarter by remembering / showing / playing with defaults or presets. E.g. 
 (e.g. "I'm in flow" + "I already have 12px `margin-left`) defaults to 'Margin tool'.
- If you combine this with CSS classes, or layout components / components with presets, it isn't yet very useful (but could definitely be enhanced)

PR #523, opened 2020-09-15
Canvas Experiment context aware highlight controls approach 2

enidemi:
Canvas experiment

Shows labels on mouseover on resizecontrols about what properties are used for the size.

Try it here: https://utopia.pizza/p/e53a49c9-invincible-century/?branch_name=experiment/highlight-context


maltenuhn:
**What I think this does**  Show me the property I'd be affecting if I were to drag on a control

**Overall reaction** Works better than I expected. I like that it's small and unintrusive.

**Frustrations** (1) As soon as I actually drag, it "goes away" and is replaced by the other tooltip. (2) I found myself trying to click on it to _change_ that variable, and felt pretty stupid for trying, but still kept trying.

**Thoughts** (1)Could we unify this with the other tooltip? (2) Or morph it / grow it into that when you actually start dragging?
(3) What mechanics could we use to then _change_ the target variable?

enidemi:
@maltenuhn We started an experiment that's related to this, it would allow the user to change the target variable using the tab key and show different options. 
I agree it should be visible while dragging!

Issue #524, opened 2020-09-15
Update ESLint + deps


alecmolloy:
We are currently stuck on old versions of ESLint because of a series of dependency issues:

1. `eslint-plugin-react-hooks@4.0.0` introduced a change that causes `eslint4b` to break some tests. 
2. something in `parser-printer.spec.ts` that stalls tests when updating things. (not exactly sure what in this case)

Some banging about with a hammer is needed if we want to update ESLint and associated things.

PR #525, opened 2020-09-15
fix eslint typescript no-shadow rules

alecmolloy:
Fixes the hundreds of `no-shadow` ESLint errors

**Problem:**
The typescript eslint plugin has a new no-shadow rule that is separate from the stock ESLint rule. I changed to using this one in the last ESLint-related PR, but didn't update the typescript plugin, so this does that.

As a side-effect, the new TS plugin messes up the react hooks plugin, by making the rule think that types should be included as deps. This is wrong, and has been fixed in the plugin, but we cannot update to this for reasons explained here: https://github.com/concrete-utopia/utopia/issues/524. For now I have either removed explicit type references that were not needed, or added suppressions.


balazsbajorics:
I'm merging this to fix the broken master.

PR #526, opened 2020-09-15
Shift eager evaluation of property controls info to an iframe.

### LINKED ISSUE #249: Don't eagerly require every user code file for exports evaluation
related TODO in code 
https://github.com/concrete-utopia/utopia/blob/02485af9fb19f697e5721b7d161506cb7f8151c7/editor/src/components/custom-code/code-file.ts#L100-L111

Problem: we need to evaluate the code files to find `.propertyControls` values there. But we eagerly evaluate every code file made by the user, even if that code file is not used in the project. The evaluation has a potential issue: some modules can have side effects (like loading a css stylesheet, or calling ReactDOM.render()), and we don't want to trigger these side effects if the user doesn't `require()` these modules. Tiggering the side effects without the file being imported can lead to mysterious behavior, very hard for any user to understand/debug and it's a very typical way of how Utopia can ambush people.

Solution?
I don't have a clear solution in my mind. Maybe the require function could as a second job collect and save the results and tell it to the editor, so only the files required by the user would get the export info, but there is no way to edit a component and need its .propertyControls without the component being imported, so I think it's not an issue
Or maybe we need a new way to get to the propertyControls and we should just set this system on fire here.

seanparsons:
Fixes #249

**Problem:**
Currently the main thread is used to determine the property controls info taken from components. But this is problematic because it will block the UI if large dependencies are added or the project is very large. Also running the risk of the modules being "executed" to get the info might also interfere with the editor behaviour because they have access to the DOM for example.

**Fix:**
The property controls info process has been moved into an iframe, separating out that part from the type it used to live in.

**Notes:**
Something is causing the browser tests to hang endlessly, which I'm investigating.

**Commit Details:**
- Fixes #249.
- Removed `propertyControlsInfo` from `CodeResultCache`.
- Changed `CodeResult.exports` to only include the types not values from them.
- Added `propertyControlsInfo` to `EditorState`.
- Added the `UPDATE_PROPERTY_CONTROLS_INFO` action for updating the above property.
- Added the `PROPERTY_CONTROLS_IFRAME_READY` action to indicate that it's now valid to start sending messages to the new iframe.
- Added type guards for the types of the new actions so that they can be checked when coming back from the new iframe.
- Created a `PropertyControlsInfoComponent` which holds the iframe for property controls and is included within `EditorComponentInner`.
- Modified the insert menu to cater for property controls info coming in subsequently to the contents of `CodeResultCache`.
- Fixed some hook errors in `insertmenu.tsx` that were introduced recently.
- `editorModelFromPersistentModel` now doesn't create an empty `EditorState`, so that unnecessary asynchronous property controls info requests are avoided.
- `packageJsonFileFromProjectContents` is a tiny bit more tolerant of file paths.
- `sendPropertyControlsInfoRequest` is the entry point for the new iframe.
- Prefixed some unused function parameters with underscores so that eslint stops complaining about them.
- `property-controls-info.tsx` is the driving logic of the iframe.


editor/src/components/editor/editor-component.tsx
```diff
@@ -565,3 +569,24 @@ const ToastRenderer = betterReactMemo('ToastRenderer', () => {
     </FlexColumn>
   )
 })
+
+const PropertyControlsInfoComponent = betterReactMemo('PropertyControlsInfoComponent', () => {
+  const iframeSrc = urljoin(PROPERTY_CONTROLS_INFO_BASE_URL, 'editor', 'property-controls-info.html')
```
Rheeseyb, line 574: I don't think you need the `.html` on the end

editor/src/components/filebrowser/filebrowser.tsx
```diff
@@ -101,26 +101,20 @@ function collectFileBrowserItems(
...
+            if (exportedVar != 'default') {
+              fileBrowserItems.push({
+                path: `${fullPath}/${exportedVar}`,
+                type: 'export',
+                fileType: null,
+                typeInformation: codeResult.exports[exportedVar].type,
+                originatingPath: originatingPath,
+                hasErrorMessages: false,
+                modified: false,
+                exportedFunction: false, // FIXME: Check if type covers this. exportedVarIsFunction,
```
Rheeseyb, line 114: Was this `FIXME` something to be fixed as part of the PR or for later?

editor/src/core/property-controls/property-controls-utils.ts
```diff
@@ -283,3 +306,68 @@ export function getPropertyControlsForTarget(
...
+export function setPropertyControlsIFrameReady(value: boolean): void {
+  propertyControlsIFrameReady = value
+}
+
+export function sendPropertyControlsInfoRequest(
+  exportsInfo: ReadonlyArray<ExportsInfo>,
+  nodeModules: NodeModules,
+  projectContents: ProjectContents,
+): void {
+  sendToIFrame()
```
Rheeseyb, line 324: I really dislike that this is called before the function is defined. Please move it down :)

editor/src/core/workers/test-workers.ts
```diff
@@ -85,19 +87,19 @@ export class FakeLinterWorker implements LinterWorker {
 }
 
 export class FakeWatchdogWorker implements WatchdogWorker {
-  initWatchdogWorker(projectID: string): void {
+  initWatchdogWorker(_projectID: string): void {
```
Rheeseyb, line 90: Why the underscores?

editor/src/templates/property-controls-info.html
```diff
@@ -0,0 +1,25 @@
...
+      crossorigin="anonymous"
+    />
+    <!-- Begin Generated Utopia External Links -->
+    <!-- End Generated Utopia External Links -->
+  </head>
+  <body>
+    <!-- projectIDScript -->
+    <div id="preview"></div>
+    <div id="preview-ui"></div>
+  </body>
```
Rheeseyb, line 24: Seems pretty much all of this can go in the bin?

editor/src/core/workers/test-workers.ts
seanparsons: ESLint warnings over unused parameters.

editor/src/templates/property-controls-info.html
seanparsons: Very true.

Rheeseyb: Well, sorry, we'd need the `<head>` tags, but no contents. I'm pretty sure we don't need the `<body>` tags.

website/src/common/env-vars.ts
```diff
@@ -32,6 +32,11 @@ export const FLOATING_PREVIEW_BASE_URL: string = PRODUCTION_CONFIG
   : STAGING_CONFIG
   ? 'https://utopia.baby/'
   : BASE_URL
+export const PROPERTY_CONTROLS_INFO_BASE_URL: string = PRODUCTION_CONFIG
```
Rheeseyb, line 35: Should this just be `export const PROPERTY_CONTROLS_INFO_BASE_URL: string = FLOATING_PREVIEW_BASE_URL`?

seanparsons: I'm a little dubious of that just because someone changing it for the preview might accidentally bust the property controls.

editor/src/components/editor/store/editor-update.spec.ts
```diff
@@ -61,6 +61,8 @@ import {
 } from '../../../core/model/test-ui-js-file'
 import { npmDependency } from '../../../core/shared/npm-dependency-types'
 import { emptyUiJsxCanvasContextData } from '../../canvas/ui-jsx-canvas'
+import {setPropertyControlsIFrameAvailable} from '../../../core/property-controls/property-controls-utils'
+
```
Rheeseyb, line 65:
```suggestion
```

editor/src/components/filebrowser/filebrowser.tsx
seanparsons: Fixed in 3f385e6b8.

editor/src/core/property-controls/property-controls-utils.ts
seanparsons: Done in 3f385e6b8.

editor/src/templates/property-controls-info.html
seanparsons: Snipped in 3f385e6b8.

editor/src/components/editor/store/editor-update.spec.ts
seanparsons: Removed in 3f385e6b8.

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -935,9 +935,7 @@ const SceneRoot: React.FunctionComponent<SceneRootProps> = (props) => {
   let rootElement = null
   let validPaths: Array<InstancePath> = []
   if (content != null) {
-    const defaultProps = defaultPropertiesForComponent(content)
```
balazsbajorics, line 938: this is a pretty important change, I think the old behavior was explicitly broken since it didn't produce the same results as the Preview

seanparsons: This would create different behaviour to the preview because the components wont have default props streamlined in.

balazsbajorics: yeah, we are in agreement :) "old behavior was explicitly broken" what I meant was that it was a broken behavior, not that it was broken by this change :)

balazsbajorics (commented):
👍

PR #528, opened 2020-09-16
run the linter as part of running npm run test

balazsbajorics:
**Problem:**
We only ran `eslint` on pre-commit hooks. The most recent master was broken with a bad eslint config and we didn't notice on the PR, because we don't run eslint in the tests.

**Fix:**
Run eslint as part of `npm run test`

**Commit Details:**
- Added a new command `npm run lint` to run the linter for all source files
- Added a new command `npm run jest` to run just the jest tests without running tsc first


enidemi (approved):

editor/package.json
```diff
@@ -12,11 +12,13 @@
   },
   "scripts": {
     "clean": "rm -rf lib/",
-    "test": "tsc && npm run --silent clean && jest --config ./jest.config.js",
-    "test-ci": "tsc && NODE_OPTIONS=--max_old_space_size=6656 xvfb-run -a npx jest --config ./jest.config.js --maxWorkers=1",
-    "test-debug": "tsc && npm run --silent clean && node --inspect-brk node_modules/.bin/jest --config ./jest.config.js --runInBand",
-    "test-watch": "tsc && npm run --silent clean && jest --config ./jest.config.js --watch",
-    "test-update-snapshot": "tsc && npm run --silent clean && jest --config ./jest.config.js --updateSnapshot",
+    "lint": "npx eslint \"src/**/*.{js,ts,tsx}\"",
+    "test": "tsc --noEmit && npm run --silent clean && npm run jest && npm run lint",
```
Rheeseyb, line 16: Oooooh, if we include the `--noemit` does that mean we can bin off the `clean` step?

editor/package.json
```diff
@@ -12,11 +12,13 @@
...
+    "jest": "jest --config ./jest.config.js",
```
Rheeseyb, line 17: Thank you for doing this  :)

editor/package.json
balazsbajorics: I think so, but at this point we had the clean step in for so long, I'm terrified of removing it

Rheeseyb (approved):

editor/package.json
Rheeseyb: The clean step was just there to bin off the result of the `tsc` step. I _think_ however it would make more sense to have the clean step before it, since most of us run `tsc` separately during dev anyway, or just not include it :shrug:

PR #542, opened 2020-09-21
Separate requested and resolved dependencies.

### LINKED ISSUE #535: Projects using `latest` in a package dependency end up failing in a jsdelivr call.
In this project there's a dependency on `@geist-ui/react`, which has a version of `latest`:
https://github.com/geist-org/react-getting-started/blob/master/package.json

In testing the Github import, it was observed that the editor makes a request to this URL:
`https://data.jsdelivr.com/v1/package/npm/@geist-ui/react@latest/flat`
Which fails with an error indicating that jsdelivr can only handle specific versions.

In conversation with Balazs he mentioned that this may be related to our support for a version of `*`, which if it is handled in the editor may provide a route to implementing a fix.

seanparsons:
Fixes #535

**Problem:**
Internally the editor treats dependencies without distinguishing between requested dependencies which may include bounds or keywords, but some aspects like the property controls can only handle resolved version numbers.

**Fix:**
Removed the old type for representing a dependency and replaced it with two representing the two cases and used those as appropriate.

**Commit Details:**
- Fixes #535.
- Get rid of the `NpmDependency` type and replace it with
  `RequestedNpmDependency` and `ResolvedNpmDependency` as appropriate.
  Which is to say the former is used "earlier" in the flow and represents
  a dependency taken directly from the user's `package.json` file.
  The latter is often then derived from that by finding the `package.json`
  for a given dependency and retrieving the version number within.
- Created `immediatelyResolvableDependenciesWithEditorRequirements`
  to help in `editorModelFromPersistentModel` so that tests can still
  more easily work without needing to resolve the dependencies.
- `fetchPackagerResponse` first gets the resolved version before pulling
  anything from JsDelivr.

### DIFF
--- a/editor/src/components/code-editor/script-editor.tsx
+++ b/editor/src/components/code-editor/script-editor.tsx
@@ -30,7 +30,7 @@ import { runtimeErrorInfoToErrorMessage } from './monaco-wrapper'
 import { EditorPanel, setFocus } from '../common/actions'
 import { EditorAction } from '../editor/action-types'
 import * as EditorActions from '../editor/actions/actions'
-import { dependenciesFromPackageJson } from '../editor/npm-dependency/npm-dependency'
+import { dependenciesFromPackageJson, usePossiblyResolvedPackageDependencies } from '../editor/npm-dependency/npm-dependency'
 import {
   ConsoleLog,
   EditorStore,
@@ -348,8 +348,7 @@ export const ScriptEditor = betterReactMemo('ScriptEditor', (props: ScriptEditor
   const newErrorsForFile = errors.filter((error) => error.fileName === filePath)
   const errorsForFile = useKeepReferenceEqualityIfPossible(newErrorsForFile)
 
-  const newNpmDependencies = dependenciesFromPackageJson(packageJsonFile)
-  const npmDependencies = useKeepReferenceEqualityIfPossible(newNpmDependencies)
+  const npmDependencies = usePossiblyResolvedPackageDependencies()
 
   if (filePath == null || openFile == null) {
     return null
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -113,10 +113,9 @@ import {
 import type {
   RequireFn,
   TypeDefinitions,
-  npmDependency,
-  NpmDependency,
   PackageStatus,
   PackageStatusMap,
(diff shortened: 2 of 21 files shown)


editor/src/components/code-editor/script-editor.tsx
```diff
@@ -348,8 +348,7 @@ export const ScriptEditor = betterReactMemo('ScriptEditor', (props: ScriptEditor
   const newErrorsForFile = errors.filter((error) => error.fileName === filePath)
   const errorsForFile = useKeepReferenceEqualityIfPossible(newErrorsForFile)
 
-  const newNpmDependencies = dependenciesFromPackageJson(packageJsonFile)
-  const npmDependencies = useKeepReferenceEqualityIfPossible(newNpmDependencies)
+  const npmDependencies = usePossiblyResolvedPackageDependencies()
```
balazsbajorics, line 351: was this `useKeepReferenceEqualityIfPossible` redundant here? why was this added in the first place? was there a reason or was it just unnecessary caution?

editor/src/components/editor/actions/actions.ts
```diff
@@ -4049,13 +4048,13 @@ export const UPDATE_FNS = {
...
   UPDATE_PACKAGE_JSON: (action: UpdatePackageJson, editor: EditorState): EditorState => {
     const dependencies = action.dependencies.reduce(
-      (acc: Array<NpmDependency>, curr: NpmDependency) => {
+      (acc: Array<RequestedNpmDependency>, curr: RequestedNpmDependency) => {
         return {
           ...acc,
           [curr.name]: curr.version,
         }
       },
-      {} as Array<NpmDependency>,
```
balazsbajorics, line 4058: ` {} as Array<NpmDependency>,`

omg

editor/src/components/code-editor/script-editor.tsx
seanparsons: Not sure about the original case, but this _should_ be fine as there's some `useMemo` inside it for that kind of reason.

editor/src/components/editor/actions/actions.ts
seanparsons: Yeah...

balazsbajorics (approved):

Rheeseyb (approved):

PR #544, opened 2020-09-21
Fix editor crash on code error

### LINKED ISSUE #541: Errors in the code editor crash the editor
e.g. starting to type `const` at the top level of the `app.js` file of a new project, the editor crashed with:

```
ReferenceError: c is not defined
    at eval (eval at exports.SafeFunctionCurriedErrorHandler (code-exec-utils.ts:NaN), <anonymous>:3:1)
```

This is easy to reproduce on Staging, but I can't reproduce it when running the editor locally.

Rheeseyb:
Fixes #541 

**Problem:**
Code errors are currently bringing down the editor. This was because of the way we were handling error reporting, combined with the way React deals with error bounds. Effectively we were repeatedly re-rendering code that would capture errors, then throw them at a later point in time so that our error bounds could then catch them. However, since moving the error bounds up a level we no longer needed this, but still needed to make sure we weren't excessively running the code that caused the errors.

**Fix:**
Removed all of the threading of `reportError` through `ui-jsx-canvas.tsx`, since we were calling that in the error bounds anyway, and updated the error bounds to stop re-rendering of the children unless the `topLevelElements` had changed.

**Commit Details:**
- Added `getDerivedStateFromError` to `CanvasErrorBoundary`, which we would use to prevent the `UiJsxCanvas` from being rendered repeatedly if there was an existing runtime error. Also added a `componentDidUpdate` to clear the state if the `topLevelElements` had changed (since in that case we _would_ want to try rendering the `UiJsxCanvas` again)
- Binned off all excessive (and incorrect) error reporting that we were threading through, since we can now just leave the executed user code to throw
- Fixed up some of the test utils around generating metadata, since that was silently failing and squashing errors before
- Fixed up some default test data that was incorrectly interpolating strings, resulting in `props.[object Object]` :facepalm: 
- Fixed `jsx-attributes.spec.ts`, which was incorrectly passing in the wrong props (another silent failure from us squashing errors)

### DIFF
--- a/editor/src/components/canvas/canvas-component-entry.tsx
+++ b/editor/src/components/canvas/canvas-component-entry.tsx
@@ -38,14 +38,14 @@ export const CanvasComponentEntry = betterReactMemo(
       ),
     }))
 
-    if (canvasProps.uiFilePath == null) {
+    if (canvasProps == null) {
       return null
     } else {
       return (
         <CanvasErrorBoundary
-          uiFilePath={canvasProps.uiFilePath}
+          fileCode={canvasProps.uiFileCode}
+          filePath={canvasProps.uiFilePath}
           reportError={props.reportError}
-          topLevelElementsIncludingScenes={canvasProps.topLevelElementsIncludingScenes}
         >
           <UiJsxCanvas {...canvasProps} clearErrors={props.clearErrors} />
         </CanvasErrorBoundary>
--- a/editor/src/components/canvas/canvas-component-entry.tsx
+++ b/editor/src/components/canvas/canvas-component-entry.tsx
@@ -42,7 +42,11 @@ export const CanvasComponentEntry = betterReactMemo(
       return null
     } else {
       return (
-        <CanvasErrorBoundary uiFilePath={canvasProps.uiFilePath} reportError={props.reportError}>
+        <CanvasErrorBoundary
+          uiFilePath={canvasProps.uiFilePath}
+          reportError={props.reportError}
+          topLevelElementsIncludingScenes={canvasProps.topLevelElementsIncludingScenes}
+        >
           <UiJsxCanvas
             {...canvasProps}
             clearErrors={props.clearErrors}
--- a/editor/src/components/canvas/canvas-component-entry.tsx
+++ b/editor/src/components/canvas/canvas-component-entry.tsx
@@ -47,11 +47,7 @@ export const CanvasComponentEntry = betterReactMemo(
           reportError={props.reportError}
           topLevelElementsIncludingScenes={canvasProps.topLevelElementsIncludingScenes}
         >
-          <UiJsxCanvas
-            {...canvasProps}
-            clearErrors={props.clearErrors}
-            reportError={props.reportError}
-          />
+          <UiJsxCanvas {...canvasProps} clearErrors={props.clearErrors} />
         </CanvasErrorBoundary>
       )
     }
--- a/editor/src/components/canvas/ui-jsx-canvas.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas.tsx
@@ -160,11 +160,15 @@ export interface CanvasReactReportErrorCallback {
   reportError: (editedFile: string, error: Error, errorInfo?: React.ErrorInfo) => void
 }
 
-export interface CanvasReactErrorCallback extends CanvasReactReportErrorCallback {
+export interface CanvasReactClearErrorsCallback {
   clearErrors: () => void
 }
 
-export type UiJsxCanvasPropsWithErrorCallback = UiJsxCanvasProps & CanvasReactErrorCallback
+export interface CanvasReactErrorCallback
+  extends CanvasReactReportErrorCallback,
+    CanvasReactClearErrorsCallback {}
+
+export type UiJsxCanvasPropsWithErrorCallback = UiJsxCanvasProps & CanvasReactClearErrorsCallback
 
 const emptyImports: Imports = {}
 const emptyTopLevelElements: Array<TopLevelElement> = []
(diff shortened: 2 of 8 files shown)


editor/src/components/canvas/canvas-component-entry.tsx
```diff
@@ -42,12 +42,12 @@ export const CanvasComponentEntry = betterReactMemo(
...
-        <CanvasErrorBoundary uiFilePath={canvasProps.uiFilePath} reportError={props.reportError}>
-          <UiJsxCanvas
-            {...canvasProps}
-            clearErrors={props.clearErrors}
-            reportError={props.reportError}
-          />
+        <CanvasErrorBoundary
+          uiFilePath={canvasProps.uiFilePath}
+          reportError={props.reportError}
+          topLevelElementsIncludingScenes={canvasProps.topLevelElementsIncludingScenes}
```
seanparsons, line 48: I wonder if anything else needs to go in there like `imports` and `jsxFactoryFunction`, might be worth picking through `UiJsxCanvasProps`.

editor/src/utils/test-utils.ts
```diff
@@ -182,6 +183,13 @@ export function createFakeMetadataForParseSuccess(success: ParseSuccess): Array<
       elementMetadata = createFakeMetadataForJSXElement(
         component.rootElement,
         TP.scenePath([BakedInStoryboardUID, createSceneUidFromIndex(index)]),
+        {
+          props: {
+            style: props[PP.toString(PathForResizeContent)] ? props.style : undefined,
```
seanparsons, line 188: Truthiness alert.

seanparsons (approved):

editor/src/components/canvas/canvas-component-entry.tsx
Rheeseyb: I was thinking we'd be best off using the raw code from the file here tbh

editor/src/utils/test-utils.ts
Rheeseyb: Hmmm yes, good point. Will pull this out and make it not truthy

editor/src/components/canvas/canvas-component-entry.tsx
seanparsons: Or that.

balazsbajorics: +1 to raw code

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -160,11 +160,15 @@ export interface CanvasReactReportErrorCallback {
   reportError: (editedFile: string, error: Error, errorInfo?: React.ErrorInfo) => void
```
balazsbajorics, line 160: shouldn't this be deleted then?

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -160,11 +160,15 @@ export interface CanvasReactReportErrorCallback {
...
 }
 
-export interface CanvasReactErrorCallback extends CanvasReactReportErrorCallback {
+export interface CanvasReactClearErrorsCallback {
   clearErrors: () => void
 }
 
-export type UiJsxCanvasPropsWithErrorCallback = UiJsxCanvasProps & CanvasReactErrorCallback
+export interface CanvasReactErrorCallback
```
balazsbajorics, line 167: could you use the same type annotation with the `&` instead of this empty interface extending a list of interfaces?

editor/src/components/canvas/ui-jsx-canvas.tsx
Rheeseyb: It's still used elsewhere (e.g. in the `CanvasErrorBoundary` itself)

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -1500,12 +1436,44 @@ function asErrorObject(e: unknown): Error {
...
+      hasError: true,
+    }
+  }
+
+  componentDidUpdate(
+    prevProps: CanvasErrorBoundaryProps,
+    _prevState: CanvasErrorBoundaryState,
+  ): void {
+    if (prevProps.topLevelElementsIncludingScenes !== this.props.topLevelElementsIncludingScenes) {
+      // eslint-disable-next-line react/no-did-update-set-state
```
balazsbajorics, line 1463: now that componentDidUpdate is the safe recommended way of reacting to prop changes, does this eslint rule still help us? https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-did-update-set-state.md

editor/src/components/canvas/ui-jsx-canvas.tsx
Rheeseyb: Done

editor/src/utils/test-utils.ts
Rheeseyb: Done (though it's still truthy, but it's a bit cleaner now)

editor/src/components/canvas/canvas-component-entry.tsx
Rheeseyb: Done

editor/src/components/canvas/ui-jsx-canvas.tsx
Rheeseyb: It's still a useful rule to have because the recommendation is to _only_ set the state _immediately_, and it must be wrapped in a condition. Having the rule will at the very least flag that the code has the potential to cause serious issues.

balazsbajorics (approved):

PR #545, opened 2020-09-22
Separate out checking vs testing scripts

Rheeseyb:
Separated out a pair of `check` and `check-ci` scripts that run `tsc`, `lint` and `test`


editor/package.json
```diff
@@ -12,13 +12,14 @@
   },
   "scripts": {
     "clean": "rm -rf lib/",
+    "check": "tsc --noEmit && npm run lint && npm run test",
```
balazsbajorics, line 15: these could run lint with a mute flag, so the warnings don't pollute the output

balazsbajorics: or maybe the `npm run lint` should already mute the warnings! the warnings IMO are useful inside your code editor, not really useful as terminal output

Rheeseyb: Good shout. I think it's worth still having a non-mute version for when you want to explicitly lint the whole code so I've also added `lint-loud` :D

balazsbajorics: :D

balazsbajorics (approved):
I  forgot to approve!

PR #546, opened 2020-09-22
experiment/flex-element-move

enidemi:
Canvas Experiment

#### How to try it?

Move flex child elements on cross axis to set `alignSelf` style property, keep your cursor on the target alignment area to set `alignItems` style prop on the parent. You should see highlight around the parent when the parent style change happens. The timer is set to 2 seconds.

Link: https://utopia.pizza/p/f1535dc3-humble-carpenter/?branch_name=experiment/flex-element-move

Sep-21-2020 18-30-58

### DIFF
--- a/editor/src/components/canvas/canvas-types.ts
+++ b/editor/src/components/canvas/canvas-types.ts
@@ -19,6 +19,7 @@ import { Mode } from '../editor/editor-modes'
 import { EditorState, OriginalCanvasAndLocalFrame } from '../editor/store/editor-state'
 import { isFeatureEnabled } from '../../utils/feature-switches'
 import { xor } from '../../core/shared/utils'
+import { FlexAlignment } from 'utopia-api'
 
 export const enum CSSCursor {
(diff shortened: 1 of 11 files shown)


maltenuhn:
**What I think this does** Let me change flex properties for alignment: `align-self`, and `align-items`, through effectively  "gestures".

**Overall reaction** Wow, I _really_ like this. It feels natural, and the 'wait-for-affecting-parent' feels very promising to me. I didn't read the code or description carefully, and found myself trying this with all sorts of other resizing and moving actions. Sadly, none of them worked

**Frustrations**  "Modify parent instead" isn't apparent unless you know about it - and it can be hard to understand what's going on.  And - two seconds feels too long :) 

**Thoughts** This is a _very_ neat and useful pattern. I immediately started wondering if "hold-to..." could / should also work with e.g. reparent? And also how we could extend it to resize (hold-over-snap), and possibly even pins (move-onto-edge pins to edge, resize-and-hold-to-edge pins to edge). 

One thing I really am stumped by is what this means for along-axis. Also it probably could do well if combined with something that tells you what's going to happen / happened. E.g. the toolbar, we could make the affected setting(s) blink before they apply, etc

PR #548, opened 2020-09-22
fix(pragma) skip style prop update for react fragments

enidemi:
**Problem:**
There were errors on the ui-jsx-canvas on every render when you had a text element inside a div.

**Fix:**
Style props are updated for the pragma, even for react fragments. It skips fragments now.

**Commit Details:**
- updated the pragma

### DIFF
--- a/utopia-api/src/pragma/pragma.tsx
+++ b/utopia-api/src/pragma/pragma.tsx
@@ -119,10 +119,15 @@ export const jsx = (type: any, ...pragmaParams: any[]) => {
             ...childStyleProps,
             ...childStylesThatOverwriteStyle[index],
           }
-          return React.cloneElement(child as any, {
-            ...removeLayoutPropFromReactBuiltins,
-            style: layoutEnhancedStyleProp,
-          })
+
+          const updatedProps =
+            (child.type as any).theOriginalType.toString() === 'Symbol(react.fragment)'
+              ? (child.props as any)
+              : {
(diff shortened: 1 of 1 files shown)


utopia-api/src/pragma/pragma.tsx
```diff
@@ -119,10 +119,15 @@ export const jsx = (type: any, ...pragmaParams: any[]) => {
             ...childStyleProps,
             ...childStylesThatOverwriteStyle[index],
           }
-          return React.cloneElement(child as any, {
-            ...removeLayoutPropFromReactBuiltins,
-            style: layoutEnhancedStyleProp,
-          })
+
+          const updatedProps =
+            (child.type as any)?.theOriginalType?.toString() === 'Symbol(react.fragment)'
```
balazsbajorics, line 124: in the monkey code file, I used `type == React.Fragment` and that seemed to work for me

enidemi: > in the monkey code file, I used `type == React.Fragment` and that seemed to work for me

I wrote that first, but it didn't work! The type is React.Element here, only the `theOriginalType` shows that it's a fragment.

seanparsons (approved):

utopia-api/src/pragma/pragma.tsx
balazsbajorics: and `theOriginalType == React.Fragment` does not work?

balazsbajorics (approved):

PR #549, opened 2020-09-22
Much speedier editor internals.

### LINKED ISSUE #537: File browser performance is abysmal
Performance reaches breaking point when loading a particularly large project (e.g. when importing https://github.com/utopia-test/mega-big-utopia). Since the file browser is open by default when editing a code file, this has a huge affect on the editor.
This is also very noticeable on projects that are much smaller than the above repo, but that repo shows just how bad it gets.

seanparsons:
Fixes #537

**Problem:**
With large projects (on the order of 1,000 files was tested), a couple of functions showed up as being terribly non-performant.

**Fix:**
For both `codeCacheToBuildResult` and `contentsToTree`, the underlying performance issue was repeated `Object.assign` calls triggered by use of object spreads either in a reduce or by merging trees containing objects respectively.

**Commit Details:**
- Fixes #537.
- Reworks the entirety of `contentsToTree` to construct the result mutably.
- Fixed an issue in `pathArbitrary` where empty paths ended up being
  constructed which then broke some specs.
- `codeCacheToBuildResult` now constructs the result mutably.

### DIFF
--- a/editor/src/components/assets.ts
+++ b/editor/src/components/assets.ts
@@ -11,6 +11,7 @@ import { isDirectory, directory, isImageFile } from '../core/model/project-file-
 import Utils from '../utils/utils'
 import * as R from 'ramda'
 import { dropLeadingSlash } from './filebrowser/filepath-utils'
+import { fastForEach } from '../core/shared/utils'
 
 interface ImageAsset {
   assetPath: string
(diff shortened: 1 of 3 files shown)


editor/src/components/assets.ts
```diff
@@ -76,87 +77,78 @@ export function projectContentFile(
...
 export function getProjectContentKeyPathElements(projectContentKey: string): Array<string> {
   return dropLeadingSlash(projectContentKey)
     .split('/')
     .filter((s) => s.length > 0)
 }
 
-export function pathElementsToProjectContentKey(elements: Array<string>): string {
-  return '/' + elements.join('/')
+export function pathElementsToProjectContentKey(
+  pathElements: Array<string> | ReadonlyArray<string>,
```
Rheeseyb, line 87:
```suggestion
  pathElements: ReadonlyArray<string>,
```
No need for the union type here. You can pass in an argument of type `Array<T>` to a parameter of type `ReadonlyArray<T>`, but not vice versa.

seanparsons: So I did have `ReadonlyArray<string>` in there originally, but TypeScript had a wobbly about my passing a regular `Array<string>` into it at one point.

seanparsons: I just shrugged my shoulders, added that in and got on with my day.

Rheeseyb (approved):

editor/src/components/assets.ts
Rheeseyb: Weird. I just tested re-writing something similar and it worked fine. It might have been that during one iteration you were doing something that relied on the type of the param matching the type passed in and that's no longer a requirement :shrug:

seanparsons: Yeah, it surprised me, but yeah shouldn't need to be a thing.

Rheeseyb:
I'm going to merge this since it's blocking me a bit

PR #551, opened 2020-09-23
Fix code editor performance with large files

### LINKED ISSUE #539: Code editor can't handle large files
When trying to open particularly large files in the code editor the entire editor will lock up. e.g. when importing https://github.com/utopia-test/react-three-flex-examples and trying to open https://github.com/utopia-test/react-three-flex-examples/blob/master/public/draco-gltf/draco_wasm_wrapper.js the file will open but the editor will very quickly grind to a halt

Rheeseyb:
Fixes #539 

**Problem:**
Monaco executes tokenization and syntax highlighting on the main UI thread (discussed in https://github.com/microsoft/monaco-editor/issues/2054), and for some reason absolutely struggles with files with long lines (the tokenization is supposed to yield after short amounts of time, but that doesn't seem to happen when the lines are too long). Even setting the [`maxTokenizationLineLength`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.iglobaleditoroptions.html#maxtokenizationlinelength) to a much lower number doesn't fully solve this (it seems to trigger more yielding, but still blocks the main thread for ~500ms at a time).

**Fix:**
Disable all language support if any lines are determined to be "too long" (I have set the max line length arbitrarily to 200 characters, so we can tweak this as necessary).

**Commit Details:**
- Updated `monaco-editor`, `monaco-editor-textmate` and `monaco-editor-webpack-plugin` to the latest versions in the hope that would have helped out here
- When creating a new `ITextModel` (the model used by monaco), perform a simple check on the code to determine if any lines are deemed too long, in which case we set the language to `plaintext` to prevent tokenization and syntax highlighting.

**Note:** This is only performed on initial file open, so if any lines exceed that limit during editing we won't disable syntax highlighting (likewise, we won't enable it on an open file if the offending lines are shortened). I could implement that behaviour if there's demand for it, but tbh it felt a bit overkill here.


seanparsons (approved):

balazsbajorics (approved):

balazsbajorics:
I wonder if it's worth warming up the question whether we could outsource monaco to an iframe

Rheeseyb:
> I wonder if it's worth warming up the question whether we could outsource monaco to an iframe

Hahaha I was going to bring this up, but the problem is that we'd still have these issues when running locally. It's probably worth doing that though.

Rheeseyb:
I've had to revert the monaco upgrade because there was a bit of a code completion regression around jsx. I updated and re-applied the patch, which works, but for some reason closing jsx tags were showing up as being deprecated. I'd already spent a fair chunk of time ensuring the patch was still needed and looking into submitting it as a PR to relevant repo, and I don't want to submit that PR until I know where this issue lies, which might take a while, so I'm going to postpone that for the time being.

PR #553, opened 2020-09-24
Fix Code Editor Saving

### LINKED ISSUE #534: cmd+s doesn't seem to reliably save the project
From Slack:

> I press cmd + s
> then I wait a bit
> then I press reload
> warning dialog comes up
> I press reload anyway
> data is lost
> what I usually do is I copy my app ui js contents
> and keep pressing cmd + s + reload until the dialog does not come up

Rheeseyb:
Fixes #534 

**Problem:**
Saving when the code editor was focused wasn't actually including the updated model in the save. As well as this, we were seemingly always telling the user that their project had unsaved changed.

**Fix:**
The code editor `cmd`+`s` flow was using a poorly named function that went outside of the correct saving flow. This function is actually for forcing through the currently queued save, which would have been queued up before the model had been updated. To fix this part of the issue we just dispatch the correct action instead of trying to circumvent the correct flow.
Also, for some reason changing the cursor position was treated as a non-transient action, meaning that in itself was constantly queuing up saves to the server, so that's no longer the case.

**Commit Details:**
- Dispatch a `SAVE_CURRENT_FILE` action from `script-editor.tsx` when forcibly saving, which triggers the "force save" process when that set of actions are handled
- Rename (and no longer export) the poorly named `forceServerSave` to `forceQueuedSave`, since that is an internal only function for forcing through the last queued save
- Made `SAVE_CURSOR_POSITION` a transient action

### DIFF
--- a/editor/src/components/editor/actions/action-utils.ts
+++ b/editor/src/components/editor/actions/action-utils.ts
@@ -80,6 +80,7 @@ export function isTransientAction(action: EditorAction): boolean {
     case 'SET_SHORTCUT':
     case 'UPDATE_PROPERTY_CONTROLS_INFO':
     case 'PROPERTY_CONTROLS_IFRAME_READY':
+    case 'SAVE_CURSOR_POSITION':
       return true
 
     case 'NEW':
@@ -131,7 +132,6 @@ export function isTransientAction(action: EditorAction): boolean {
     case 'UNWRAP_LAYOUTABLE':
     case 'UPDATE_JSX_ELEMENT_NAME':
     case 'SET_ASPECT_RATIO_LOCK':
-    case 'SAVE_CURSOR_POSITION':
     case 'SET_CODE_EDITOR_THEME':
     case 'INSERT_DROPPED_IMAGE':
     case 'RESET_PROP_TO_DEFAULT':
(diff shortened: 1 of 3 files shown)


enidemi (approved):

editor/src/components/editor/actions/action-utils.ts
```diff
@@ -131,7 +132,6 @@ export function isTransientAction(action: EditorAction): boolean {
     case 'UNWRAP_LAYOUTABLE':
     case 'UPDATE_JSX_ELEMENT_NAME':
     case 'SET_ASPECT_RATIO_LOCK':
-    case 'SAVE_CURSOR_POSITION':
```
balazsbajorics, line 134: oops?

editor/src/components/editor/actions/action-utils.ts
```diff
@@ -131,7 +132,6 @@ export function isTransientAction(action: EditorAction): boolean {
...
     case 'SET_CODE_EDITOR_THEME':
```
balazsbajorics, line 135: while we are here... this also sounds like it shouldn't be in the undo flow? :)

balazsbajorics (approved):

editor/src/components/editor/actions/action-utils.ts
Rheeseyb: Yeah... this is part of project "Fix the damn undo flow". The problem is "Transient" functions are used to label things that we want to add to the undo stack and that we want to trigger a save, both of which should definitely be separate things. I'm going to say we leave this until we finally get a chance to tackle that as a whole.

editor/src/components/editor/actions/action-utils.ts
Rheeseyb: I can see why we want to store the cursor position when saving, but having _every single_ cursor change trigger a save was nightmare fuel.

PR #557, opened 2020-09-24
Safer lookup of the main html and js files from the package.json

### LINKED ISSUE #554: Preview fails because of hard coded reliance on 'utopia' fields in 'package.json'

Rheeseyb:
Fixes #554 

**Problem:**
The preview throws an error if the user's project doesn't contain a `'utopia'` key in their `package.json`.

**Fix:**
Provide a much safer way for looking up the main HTML and JS filepaths in the project

**Note:**
This treads quite heavily into the work that Sean is currently undertaking around performance in the File Browser (since there is more work to do there around fixing up the shape of `ProjectContents`), so I have deliberately taken the decision to **not** thread other code through the new code in `project-contents-utils.ts`. For the same reason I haven't spent the time adding tests, because I strongly suspect that the code written here will change significantly or even go in the bin pretty soon (if not I'll add them :) ).

**Commit Details:**
- Added `project-contents-utils.ts`, which right now only provides a safe lookup for specific project file paths, but will be expanded for more general `ProjectContents` functionality
- Updated `moize`, and fixed the types used when exposing its functionality in `memoize.ts`
- Use the new functions in `preview-pane.tsx` and `preview.tsx` to prevent them failing for imported projects


editor/src/core/shared/project-contents-utils.ts
```diff
@@ -0,0 +1,58 @@
+import { objectKeyParser, parseString } from '../../utils/value-parser-utils'
+import { defaultEither, eitherToMaybe, isLeft } from './either'
+import { is } from './equality-utils'
+import { memoize } from './memoize'
+import { isCodeFile, ProjectContents, ProjectFile } from './project-file-types'
+
+function parsePackageJsonInner(packageJson: ProjectFile | null): Record<string, any> | null {
```
seanparsons, line 7: Strictly I think the parameter type should include `| undefined`, as that's what you get if you key into an object.

seanparsons (approved):

editor/src/core/shared/project-contents-utils.ts
Rheeseyb: Done

PR #564, opened 2020-09-25
killing the silly nonstatic layout system

balazsbajorics:
**Problem:**
Special Size Measurements detected layout systems, `layoutSystemForChildren` and `parentLayoutSystem`. The options were `'flex' | 'grid' | 'flow'` and a rather badly named `'nonfixed'` which was meant to basically mean "pinSystem". This was totally wrong for two reasons: 1. it should have been called "nonstatic" and it should have detected whether the position is static of an element or not 2. even if something is "nonfixed" it still actually makes a flow layout.

**Fix:**
Kill nonfixed layout system type. Added a new boolean flag called `providesBoundsForChildren`. If providesBoundsForChildren is true AND the layoutSystem is `flow`, it means it is a "pinned system", ie you can expect to use pins on the children and they will work properly.

**Commit Details:**
- remove `nonfixed` from layout system options
- rename and fix behavior of `isElementNonFixed` to `isElementNonStatic` which is now used to tell if the element provides a coordinate system for its children.
- fixed the Inspector's "Layout System" section which was unable to show Pins as selected for months now. Now it shows Pins if the detected layout system is flow && `providesBoundsForChildren === true`


enidemi (approved):

editor/src/components/editor/actions/actions.ts
```diff
@@ -686,7 +686,7 @@ function switchAndUpdateFrames(
       case 'flex':
         return 'flex'
       case LayoutSystem.PinSystem:
-        return 'nonfixed'
+        return 'flow'
```
seanparsons, line 689: This makes my eye twitch! :)

editor/src/components/canvas/dom-walker.ts
```diff
@@ -47,22 +47,19 @@ function elementLayoutSystem(computedStyle: CSSStyleDeclaration | null): Detecte
...
   return positionValues.includes(valueAsAny) ? valueAsAny : null
 }
 
-function isElementNonFixed(computedStyle: CSSStyleDeclaration | null) {
+function isElementNonStatic(computedStyle: CSSStyleDeclaration | null) {
   if (computedStyle == null) {
     return false
   }
-  if (computedStyle.position != null && computedStyle.position !== 'fixed') {
+  if (computedStyle.position != null && computedStyle.position !== 'static') {
```
Rheeseyb, line 62: Seems we don't need the `!= null` checks at the start of these

seanparsons (approved):
As long as the tests are fixed, looks good to me!

Rheeseyb:
> As long as the tests are fixed, looks good to me!

Looking at the failing tests it seems like there is a bug in the logic and not the tests?

seanparsons:
> Looking at the failing tests it seems like there is a bug in the logic and not the tests?

That's what I meant, wasn't saying the tests themselves were specifically wrong.

balazsbajorics:
I forgot about this PR! also, the tests!

balazsbajorics:
turns out the tests were previously incorrect, the new snapshot is the correct one!

PR #566, opened 2020-09-28
Preview only processes the latest update

### LINKED ISSUE #563: Preview lags badly on chained updates from large projects
When importing https://github.com/utopia-test/mega-big-utopia and making changes to the canvas, all updates are queued up and processed in turn by the preview, rather than just the last update.

Rheeseyb:
Fixes #563 

**Problem:**
There is no queuing system in the preview, meaning it process every single model update in turn. As well as that, we're updating the preview with every model change even though we only care about changes to the `projectContents`

**Fix:**
Only send updates to the preview on changes to the `projectContents`, and queue up the latest model update if the preview is currently being updated with a previous model.

**Commit Details:**
- Added an interface for sending project contents update messages to the floater preview
- `dispatch.tsx` only updates the floater preview if there are project contents changes (rather than any change that would trigger a save), or an explicit request to send the latest project
- Track whether a model is being loaded by the preview, and if so queue up the latest model change (replacing the previously queued change each time)
- Ensure that we handle any queued model change after a preview render, regardless of if there was an error or not.

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -13,6 +13,7 @@ import {
   UIJSFile,
   isCodeOrUiJsFile,
   isParseSuccess,
+  ProjectContents,
 } from '../../../core/shared/project-file-types'
 import {
   codeNeedsParsing,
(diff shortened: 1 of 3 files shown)


maltenuhn:
Perhaps it's the time to rename the thing from "Floater"? It's not been afloat for a while now...

Rheeseyb:
> Perhaps it's the time to rename the thing from "Floater"? It's not been afloat for a while now...

Ha yeah! It's only called that in one place but yep will do

seanparsons (approved):

enidemi (approved):

PR #577, opened 2020-09-30
experiments collection

enidemi:
Almost all previous Canvas Experiments in one place behind feature switches. All of them are enabled by default.

https://utopia.pizza/p/971e8889-cord-cowl/?branch_name=experiments/collected


Rheeseyb:
Tagged and closed

PR #580, opened 2020-10-01
Add in support for ES module system content.

### LINKED ISSUE #508: Support ESModule import
**Problem**
We initially assumed that only a v small number of React projects use ESModules. Unfortunately, that turned out to be somewhat too optimistic.

 The R3F community  in particular likes and uses them, and they the projects there are snacks we'd like to support.

Related / affected: #86, #426, #427

seanparsons:
Fixes #508

**Problem:**
Previously any dependency which was packaged up using ES modules would fail on use/import as everything is ordinarily packaged up in our editor with CommonJS.

The fix to this problem then highlighted a separate issue where modules were being re-executed in the property controls iframe, which in development runs in the main thread. With a large dependency the transpiling of that dependency would cause the editor to lag reasonably often.

**Fix:**
For the first problem, the editor now on a ES module error transpiles the code to be CommonJS packaged.

For the second problem, updates to the property controls iframe are now sent as partial updates if they only include changes to the project files.

**Commit Details:**
- Fixes #508.
- Added `@babel/plugin-transform-modules-commonjs` dependency.
- Refactored out `clearNodeModules` as it's useful to be able to
  produce `NodeModules` without the eval cache.
- Added `onlyProjectFiles` parameter to `generateCodeResultCache`.
- `UPDATE_NODE_MODULES_CONTENTS` action handler scans the content
  to add to determine the value of `onlyProjectFiles` to pass to
  `generateCodeResultCache`.
- Added `transformToCommonJS` function which will rewrite the module
  system of a file to that of CommonJS.
- `evaluateJs` on a first failure checks to see if the error stems
  from the code being an ES module, then transforms it to CommonJS
  and tries evaluating it a second time only.
- Removed previous ES module handling which would throw an error
  when encountering an ES module style package.json file.
- `NodeModulesUpdate` type used instead of `NodeModules` when communicating
  to the property controls iframe, such that partial updates can
  be sent as well as the full updates that were previously always sent.
- `sendPropertyControlsInfoRequest` combines queued updates logically
  so that partial updates are added in on top of full updates that have
  yet to be sent. Which is primarily used during startup of the editor
  as the iframe takes a short while to be available.


editor/src/core/es-modules/evaluator/evaluator.ts
```diff
@@ -18,38 +19,62 @@ function isEsModuleError(error: Error) {
...
-    [],
-    (error) => {
-      // we throw the error here, the require fn will catch it
-      if (isEsModuleError(error)) {
-        error.message = createEsModuleError(filePath, error.message)
-      }
-      throw error
-    },
-  )()
+  function secondErrorHandler(error: Error): void {
```
Rheeseyb, line 50: Looks like this can just be inlined?

editor/src/core/es-modules/package-manager/module-resolution.ts
```diff
@@ -177,7 +154,8 @@ function processPackageJson(
       const mainEntry: string | null = packageJson.main ?? null
       const moduleEntry: string | null = packageJson.module ?? null
       if (moduleEntry != null && mainEntry == null) {
-        return resolveESModuleFailure(moduleName)
+        return resolveSuccess(normalizePath([...containerFolder, ...pathToElements(moduleName)]))
+        //return resolveESModuleFailure(moduleName)
```
Rheeseyb, line 158:
```suggestion
```

Rheeseyb (approved):
Nice work!

editor/src/core/es-modules/evaluator/evaluator.ts
seanparsons: Could do, I just found it looking quite tidy with the other error handler in there.

editor/src/core/es-modules/package-manager/module-resolution.ts
seanparsons: Fixed in 1935a80e.

editor/src/core/es-modules/evaluator/evaluator.ts
```diff
@@ -18,38 +19,62 @@ function isEsModuleError(error: Error) {
...
 
-  // https://nodejs.org/api/process.html#process_process_env
-  // This is a hacky solution, ideally we'd want a transpiler / loader that replaces process.env.NODE_ENV with a user-defined value
-  let process = {
-    env: {
-      NODE_ENV: 'production',
-    },
+  function firstErrorHandler(error: Error): void {
+    if (isEsModuleError(error)) {
+      const transpiledCode = transformToCommonJS(filePath, moduleCode)
```
balazsbajorics, line 43: can Babel's `@babel/plugin-transform-modules-commonjs` run for a file that is already commonjs without destroying it? if so, we could pass every file through babel here.

that way, we could fix my `// This is a hacky solution, ideally we'd want a transpiler...` by also including this transformation: https://babeljs.io/docs/en/babel-plugin-transform-node-env-inline/

editor/src/core/es-modules/test-cases/file-with-es-module.json
```diff
@@ -1,7 +1,7 @@
 {
   "contents": {
     "/node_modules/mypackage/index.js": {
-      "content": "export const cica = 'hello';",
+      "content": "export const hello = 'hello!';",
```
balazsbajorics, line 4: awwww

editor/src/core/es-modules/evaluator/evaluator.ts
seanparsons: I'm a little dubious of everything going through it given that the big blob `three` file takes 2-3 seconds to transpile. Which could _really_ stuff start up times.

editor/src/core/es-modules/test-cases/file-with-es-module.json
seanparsons: Just made it more consistent with another file that looks the same. :)

editor/src/core/es-modules/evaluator/evaluator.ts
balazsbajorics: life in the browser is hard! :)

balazsbajorics (approved):

PR #583, opened 2020-10-01
Support dependencies with tagged versions

### LINKED ISSUE #582: Bundler: Support tagged dependency versions
NPM supports tagged versions (e.g. 'latest', 'next', etc.), so we should too.

Rheeseyb:
Fixes #582 

**Problem:**
When trying to resolve the dependency version, we weren't checking for tagged versions.

**Fix:**
Check what the type of the "version" is, and if it is a tag we use the package metadata to resolve the actual version.

**Commit Details:**
- Add a function `getVersionType` to distinguish between the different types of "versions" a user can specify
- Use that function to determine how we should attempt to resolve the concrete version
- Tests


balazsbajorics (approved):
thanks for the tests!

enidemi (approved):

PR #585, opened 2020-10-02
experiment/do-not-move-while-reorder

balazsbajorics:
https://utopia.pizza/p/44a911cb-spark-persimmon/?branch_name=experiment/do-not-move-while-reorder

**Problem:**
The flex drag-to-reorder interaction updates the index of the dragged element live. This makes it intuitive to see what the result of the drag will be. However, it makes our lives very hard if the layout wraps into multiple lines. (flex-wrap or _flow_). Live updating in a multiline layout means that as you move around, the rows shift wildly around, and depending on the situation, it can cause sitations where you cannot move your mouse to indicate the index position correctly. (A too large sibling might shift around in a way that results in a jump of 2-3 indexes at once, and if you try to move the mouse backwards to correct for it, the jump happens again)

**Fix:**
In this experiment the dragged flex child remains in its original position, and we draw a (big ugly red) indicator line to where it will fall if you let go of the mouse. It tries to mimic the `|` cursor of text editors, when you are drag-to-moving a selected snippet of text.

image

### DIFF
--- a/editor/src/components/canvas/canvas-types.ts
+++ b/editor/src/components/canvas/canvas-types.ts
@@ -181,6 +181,11 @@ export type HigherOrderControl = SvgControl | DivControl
 
 export type ControlOrHigherOrderControl = SvgFragmentControl | HigherOrderControl
 
+export type ReparentTargetIndicatorPosition = {
+  parent: TemplatePath | null
+  drawBeforeChildIndex: number
+}
+
 export interface FrameAndTarget<C extends CoordinateMarker> {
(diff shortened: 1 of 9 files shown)


maltenuhn:
@balazsbajorics is there 🔺.🍕 yet?

balazsbajorics:
@maltenuhn  https://utopia.pizza/p/44a911cb-spark-persimmon/?branch_name=experiment/do-not-move-while-reorder

PR #587, opened 2020-10-05
Make the server return placeholders for filtered files when downloading dependencies

Rheeseyb:
**Problem:**
When discussing the current dependency resolution flow with Sean this morning it occurred to me that there was redundancy in the current work flow and possibly a significant bug hiding in that. The current flow is to:

1. Have the server install a dependency and then serialise the result as JSON, removing all but the `package.json`, `*.js` and `*.d.ts` files
2. Then request the list of files for a dependency from jsdelivr and use that response on the client to create placeholders for all files, merging the result with the server's response from 1
3. Download the actual file contents for those placeholders when they are actually imported

Since the server can simply replace the unwanted files with placeholders, rather than filtering them completely, we can remove step 2 and any chance of missing out files for transitive dependencies (which I believe we were not creating as part of step 2).

**Fix:**
Have the server replace unwanted files with `'PLACEHOLDER_FILE'`, and then turn that into an actual placeholder object on the client.

**Commit Details:**
- Updated `NPM.hs` to replace unwanted files with `'PLACEHOLDER_FILE'`, including the logic for encoding the new file types as JSON
- Created a `PackagerServerFile` type on the client side for handling both types of file
- Updated `fetch-packages.ts` to use this to create either code files or placeholder files, and removed the now unnecessary calls to jsdelivr to create those placeholders
- Added a comment to explain what is happening where this is unclear
- Updated the tests and test data to take into account the new flow
- Committed the updated `package-lock.json` because it seems that hasn't been committed from a previous PR?


seanparsons (approved):

editor/package-lock.json
```diff
@@ -1199,15 +1199,184 @@
       }
     },
     "@babel/plugin-transform-modules-commonjs": {
-      "version": "7.10.1",
-      "resolved": "https://registry.npmjs.org/@babel/plugin-transform-modules-commonjs/-/plugin-transform-modules-commonjs-7.10.1.tgz",
-      "integrity": "sha512-AQG4fc3KOah0vdITwt7Gi6hD9BtQP/8bhem7OjbaMoRNCH5Djx42O2vYMfau7QnAzQCa+RJnhJBmFFMGpQEzrg==",
-      "dev": true,
+      "version": "7.10.4",
```
balazsbajorics, line 1202: what's up with this guy? you didn't change the package.json

Rheeseyb: #580 came with a `package.json` change but the corresponding `package-lock.json` wasn't committed

PR #588, opened 2020-10-05
fix(actions) change propertycontrols info update

enidemi:
I couldn't find the ticket for this.

**Problem:**
Sometimes Views couldn't be resized because the propertyControlsInfo for style properties was missing. Sometimes propertyControls defined in the app.js were missing from the inspector.

**Fix:**
Updating files sets the new propertycontrols but deletes all old ones at the same time, including controls coming from third party components, like View from utopia-api.

Screenshot of the issue where the controls are added to the code but missing from the inspector:
Screenshot 2020-10-05 at 16 59 57

### DIFF
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -4142,7 +4142,10 @@ export const UPDATE_FNS = {
   ): EditorState => {
     return {
       ...editor,
-      propertyControlsInfo: action.propertyControlsInfo,
+      propertyControlsInfo: {
+        ...editor.propertyControlsInfo,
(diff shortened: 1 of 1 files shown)


seanparsons:
That's weird, that should surely be coming from `getControlsForExternalDependencies`? Which is called initially in the property controls info lookup, so maybe that's failing there instead?

It _should_ be a full update coming from the iframe, if we constantly merge we run the risk of irrelevant data living until we reload the editor.

Rheeseyb:
I've spoken to Eni and I'm going to pick this up

Rheeseyb:
Closing as this is now superseded by #606

PR #589, opened 2020-10-05
Update readme.md

alecmolloy:
Updates our readme.

- nix install instructions are now simplified thanks to the nixos team. a simple flag is needed to make it work on macOS Catalina and up.
- added some language to explain we're happy to pair up with contributors
- removed recommended VS Code `"prettier.eslintIntegration": true,` setting as it is no longer functioning

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -134,7 +134,7 @@ Limitations:
 # Troubleshooting
 
 ## fsevents
-If you have fsevents trouble, try nuking your node_modules and reinstalling with `npm install`
+If you notice that 1 or more CPU cores are running 100% because of `node` processes, it is probably webpack-dev-server having trouble with `fsevents` on MacOS. To fix it, run `npm install fsevents` in the `utopia/editor` directory. see https://github.com/webpack/webpack/issues/701#issuecomment-216082726
 
 ## Running this without Nix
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -139,33 +133,10 @@ Limitations:
...
-
-```
-curl -L https://nixos.org/nix/install | sh
-```
-
-Add the `. /Users/{{USERNAME}}/.nix-profile/etc/profile.d/nix.sh` line to your shell's config file (e.g. ~/.zshrc, ~/.bashrc) as directed by the end of the nix install.
-
-And while you're at it, this as well, which will supress the error that you're using a symlink instead of a real directory: `export NIX_IGNORE_SYMLINK_STORE=1`
+## fsevents
+If you have fsevents trouble, try nuking your node_modules and reinstalling with `npm install`
```
balazsbajorics, line 137:
```suggestion
If you notice that 1 or more CPU cores are running 100% because of `node` processes, it is probably webpack-dev-server having trouble with `fsevents` on MacOS. To fix it, run `npm install fsevents` in the `utopia/editor` directory. see https://github.com/webpack/webpack/issues/701#issuecomment-216082726
```

balazsbajorics (approved):

seanparsons (approved):

PR #590, opened 2020-10-05
fix react-select patch-package

alecmolloy:
Patch package had a mismatched react-select version number with the actual package, and was giving a warning. this simply updates the version target.

### DIFF
--- a/editor/patches/react-select+3.1.0.patch
+++ b/editor/patches/react-select+3.1.0.patch
@@ -1,7 +1,6 @@
 diff --git a/node_modules/react-select/src/.DS_Store b/node_modules/react-select/src/.DS_Store
 new file mode 100644
-index 0000000..dbd07c1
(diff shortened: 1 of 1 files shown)


editor/patches/react-select+3.1.0.patch
```diff
@@ -1,7 +1,6 @@
 diff --git a/node_modules/react-select/src/.DS_Store b/node_modules/react-select/src/.DS_Store
 new file mode 100644
-index 0000000..dbd07c1
-Binary files /dev/null and b/node_modules/react-select/src/.DS_Store differ
```
balazsbajorics, line 4: what's up with this line? is this noise?

alecmolloy: ahh the DS_Store is something that must have been added by macOS!

alecmolloy: but that is getting deleted, so… i think we're good here. prob shouldn't have been there in the first place.

PR #591, opened 2020-10-07
Error on undefined components

### LINKED ISSUE #468: Error: "Span" is not part of the THREE namespace!
Code: (with three, react-three-fiber packages installed)

n.b. the error comes from importing `Box` from a file that has no export under that name, but the error is nonsensical.

```jsx
/** @jsx jsx */
import { Canvas } from 'react-three-fiber'
import { Scene, Storyboard, jsx } from 'utopia-api'
import { Box } from './index.js'

export var App = (props) => {
  return (
    <Canvas style={{ left: 38, top: 0 }}>
      <Box />
    </Canvas>
  )
}

export var storyboard = (
  <Storyboard>
    <Scene component={App} />
  </Storyboard>
)
```

Error: 
```
Error: "Span" is not part of the THREE namespace! Did you forget to extend it? See: https://github.com/react-spring/react-three-fiber#using-3rd-party-non-three-namespaced-objects-in-the-scene-graph
```

seanparsons:
Fixes #468 

**Problem:**
Previously when a reference to a component was `undefined`, the canvas would insert into the containing element an error component in its place. However this conflicted with Three in quite a strong way because it assumed that the lowercased `span` was one of its own components.

**Fix:**
Instead of streamlining in the error element, now an exception is thrown which brings it a little more inline with real React. With the message of the exception indicating where the undefined component originates from.

**Commit Details:**
- Fixes #468.
- Changed `createMissingElement` to instead throw an exception with some level of information to indicate where it originated.
- Added `imports` and `topLevelElementNames` to `RerenderUtopiaContextProps` and propagated those values through the render.
- Implemented `findPossibleImport` which attempts to identify which import a given name relates to.


editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -1291,7 +1314,7 @@ function renderJSXElement(
     elementIsIntrinsic && !elementIsBaseHTML ? filterDataProps(elementProps) : elementProps
 
   if (FinalElement == null) {
-    return createMissingElement(jsx)
+    return throwForMissingElement(jsx, imports, topLevelNames)
```
balazsbajorics, line 1317: what would be the error message if we wouldn't handle this case and call createElement with a `null` or `undefined`? would there be an error, or just a silent missing component?

seanparsons: Let me try!

seanparsons: @balazsbajorics Oddly it seems to be really variable, testing around editing a file and reloading the editor I've seen the debug version of the undefined error that React normally gives, the production version of that with the minified error URL and a `ReferenceError`.

balazsbajorics: hah, that's... not the best. the undefined error (if reliable) would be the best, but I don't think it's worth more of your time to try to chase it now

balazsbajorics (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
Rheeseyb: Sorry for the late review here...

In looking at this, I'm wondering if we're better off removing this custom handling and replacing the line below with 
```if (FinalElement != null && TP.containsPath(templatePath, validPaths)) {```
Doing that would lead to us trying to use the standard React handling of invalid components, which gives an error:
```Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.```
Though granted the "Check the render method of " part of that error message will be wrong because it is inside the canvas's custom rendering (it should read "Check the render method of App"):

Screenshot_20201009_173047

seanparsons: I'm not quite sure what you're describing position-wise, but if it's effectively similar to @balazsbajorics's suggestion I noticed that it wasn't very consistent in behaviour when it came to reloads and changes.

Rheeseyb: Hmmm I've just been playing with this and I'm not seeing that weird inconsistent behaviour. OK I'll play about with this a bit more and come back to you because I'd much rather we could rely on the default React behaviour than have custom behaviour if possible.

seanparsons: I'm all for that, entirely possible that a change to the logic makes it work more effectively. It _sounds_ like your logic makes this function just return an `undefined` instead of invoking `React.createElement` with `undefined.

PR #592, opened 2020-10-07
Experiment/removing layout pinsystem

enidemi:
Canvas experiment

- removing 'pinSystem' layoutsystem from utopia-api
- inspector menu to set pinned is removed too
- no more magic style prop from utopia-api: `position: relative` to pin parents and `position: absolute` to pinned children
- default view insertion has `position: absolute` instead of pinSystem


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -410,6 +410,12 @@ export const MetadataUtils = {
   isPositionAbsolute(instance: ElementInstanceMetadata | null): boolean {
     return instance?.specialSizeMeasurements.position === 'absolute'
   },
+  isFlowElement(instance: ElementInstanceMetadata | null): boolean {
+    return (
+      instance?.specialSizeMeasurements.immediateParentProvidesLayout === false &&
```
balazsbajorics, line 415: hmm do we need to check for `instance?.specialSizeMeasurements.immediateParentProvidesLayout === false`? why not `instance?.specialSizeMeasurements.parentLayoutSystem === 'flow'` ?

enidemi: fixed! I had some confusion about the specialSizeMeasurements

balazsbajorics (approved):

Issue #506, opened 2020-09-07
Canvas doesn't show the React "invalid hook call" errors.


seanparsons:
The fix in this PR: https://github.com/concrete-utopia/utopia/pull/505
unearthed an issue where the canvas doesn't display the user friendly "invalid hook call" error that React displays. The triggering exception is displayed instead which gives just the exception message only.

Rheeseyb:
To add a bit more here - the issue is that we're evaluating the user's project inside a functional React component (the canvas), meaning that if they add a hook at the top level of the file that will still be evaluated in a hooks-legal manner. I had been toying with the idea of the Canvas becoming a separate `ReactDOM.render()` call (along with / as opposed to the "canvas as an iframe" idea), though I'm not sure quite how tricky that would make everything...

balazsbajorics:
> To add a bit more here - the issue is that we're evaluating the user's project inside a functional React component (the canvas), meaning that if they add a hook at the top level of the file that will still be evaluated in a hooks-legal manner. I had been toying with the idea of the Canvas becoming a separate `ReactDOM.render()` call (along with / as opposed to the "canvas as an iframe" idea), though I'm not sure quite how tricky that would make everything...

oh this is interesting problem to have! I think it's worth spiking if a separate ReactDOM.render would solve things here. one more reason for more spearation between the editor and the Canvas!

Rheeseyb:
> one more reason for more spearation between the editor and the Canvas!

Did you literally just copy and paste what I said to you there?! ;) But yes, absolutely +1 here, though I'm nervous about the outcomes of this...

seanparsons:
I'm going to mark this as blocked by whatever work creates an isolated canvas.

PR #595, opened 2020-10-08
Remount the canvas on incrementing the mount count.

seanparsons:
**Problem:**
While investigating #506, it was observed that if hooks are added/removed to the code via the code editor the React errors for too many or too few hook calls could be triggered.

**Fix:**
Set the `key` for the `CanvasErrorBoundary` so that the hierarchy is recreated thereby side-stepping the error.

**Commit Details:**
- Increment the `key` of `CanvasErrorBoundary` used in the canvas
  so as to prevent changes in the number of hooks in the root from
  surfacing as the React too many/few hook calls invoked calls error.


seanparsons:
@Rheeseyb Yeah, I thought much the same which is why this surprised me.

PR #597, opened 2020-10-08
Support git dependencies

### LINKED ISSUE #574: Bundler: Support dependencies specified using a direct github URL
e.g. the dependency `"hooks-by-ruvkr": "https://github.com/ruvkr/hooks-by-ruvkr.git"`

### LINKED ISSUE #598: Replace our use of yarn with npm on the server
After hitting against a pretty major yarn bug (https://github.com/yarnpkg/yarn/issues/8238, PR [here](https://github.com/yarnpkg/yarn/pull/8239)), it seems our only choice for supporting git dependencies is to switch to npm. Our only use of yarn is in `NPM.hs`, but the dependency download relies on the format matching that of `yarn add`, so we'll need to replicate that.

Rheeseyb:
Fixes #574 
Fixes #598 

**Problem:**
We don't support most types of npm dependency listed [here](https://docs.npmjs.com/files/package.json#dependencies). ~Unfortunately, this PR doesn't reach the point of solving that, because there is still an outstanding issue with yarn (https://github.com/yarnpkg/yarn/issues/8238), which has an open PR (https://github.com/yarnpkg/yarn/pull/8239) that has unfortunately been awaiting feedback since July.~ - yes it does now reach the point of solving the github part :smiley: 

**Fix:**
Most of the problem here comes from the way in which we resolve versions, how we fetch those dependencies via the server, and then how we blank out certain files and fetch those later.

To fix the version resolution, I've used [`npm-package-arg`](https://www.npmjs.com/package/npm-package-arg) for determining the type of the dependency (this is used by `npm install` itself, so this is far more reliable than trying to write something ourselves for this). I've also changed the way the server fetches the possible matching versions, by reading the `version` field rather than the `versions` field from `npm view <package>@<version>`, which returns either a concrete version or an array of matching versions. We then use `semver` to find the best match if the specified version was a semver range.

For improving the dependency fetching, I've removed the separation of version from package name when making the packager request to the server, since for github dependencies we don't want to concat them in the same way as when specifying npm dependencies.

For the final issue (fetching blanked files), I've used [`hosted-git-info`](https://www.npmjs.com/package/hosted-git-info), which provides functionality for looking up specific files in a git dependency. ~However, I've been unable to verify that this works fully because of the above yarn issue.~ I have now verified that this does work.

**Follow Up:**
~Given there appears to be no movement on the above yarn issue, I think the only way to finish this is to either create our own fork, or tweak the process to use npm instead. I'll create a ticket for this.~ Ok that ended up being super quick so I've committed directly to this branch rather than creating a new PR specifically for that change.


server/src/Utopia/Web/Packager/NPM.hs
```diff
@@ -16,17 +16,31 @@ import           System.Process
...
 
-withInstalledProject :: QSem -> Text -> Text -> (FilePath -> IO a) -> IO a
-withInstalledProject semaphore jsPackageName jsPackageVersion withInstalledPath = do
+withInstalledProject :: QSem -> Text -> (FilePath -> IO a) -> IO a
+withInstalledProject semaphore versionedPackageName withInstalledPath = do
   -- Create temporary folder.
   withSystemTempDirectory "packager" $ \tempDir -> do
     -- Run `npm install "packageName@packageVersion"`.
-    let baseProc = proc "yarn" ["add", "--prefer-offline", "--silent", toS jsPackageName <> "@" <> toS jsPackageVersion]
+    let baseProc = proc "npm" ["install", "--silent", toS versionedPackageName]
```
seanparsons, line 43: To the bin with yarn.

seanparsons (approved):
Does this mean we can remove yarn from shell.nix?

Rheeseyb:
> Does this mean we can remove yarn from shell.nix?

I believe it does :) I will make the change and push (along with resolving the conflicts)

balazsbajorics (commented):
👍

PR #599, opened 2020-10-08
Experiment/canvas pin controls

enidemi:
Canvas Experiment

#### How to try it?
Select an element and you should see guidelines about what position props it uses and how far it is from its containing block.

🍕 https://utopia.pizza/p/85906fbb-hurricane-recess/?branch_name=experiment/canvas-visible-pins


maltenuhn:
**What I think this does** It shows me the distance of an element to its parent (or is it origin container?), for dimensions that are pinned. Or Its width / height, if those are pinned.

**First reaction** Feels useful, at least with the prototype here. It's heavy though, so progressive or contextual disclosure might be help.

**Ideas** Not much that's not cosmetic - in a good way! Interestingly I didn't find myself trying to interact with it; I'll comment more on the PR that adds that functionality.

PR #603, opened 2020-10-09
fix(fileitem.tsx) Handle file names with trailing slashes

### LINKED ISSUE #602: Regression: Importing a Github project now clears folder names

Rheeseyb:
Fixes #602 

**Problem:**
Folder names with a trailing slash were being rendered as empty strings

**Fix:**
Filter empty values from the result of the string split we were using. I've also removed a comment questioning the allowance of folders with a dot in their names since folders prefixed with a dot are definitely allowed.

### DIFF
--- a/editor/src/components/filebrowser/fileitem.tsx
+++ b/editor/src/components/filebrowser/fileitem.tsx
@@ -26,7 +26,6 @@ import * as EditorActions from '../editor/actions/actions'
 import { openFileTab } from '../editor/store/editor-state'
 import { ExpandableIndicator } from '../navigator/navigator-item/expandable-indicator'
 import { FileBrowserItemInfo, FileBrowserItemType } from './filebrowser'
-import { dropLeadingSlash } from './filepath-utils'
 import { PasteResult } from '../../utils/clipboard-utils'
 import { codeFile } from '../../core/model/project-file-utils'
 import { dragAndDropInsertionSubject, EditorModes } from '../editor/editor-modes'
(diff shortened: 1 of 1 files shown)


editor/src/components/filebrowser/fileitem.tsx
```diff
@@ -231,8 +230,7 @@ class FileBrowserItemInner extends React.PureComponent<
         isRenaming: props.renamingTarget === props.path,
       }
     }
-    // todo this is not very safe. should it be not allowed to use . in folder names?
-    const pathParts = dropLeadingSlash(props.path).split('/')
+    const pathParts = props.path.split('/').filter((s) => s !== '')
```
balazsbajorics, line 233: hmm this is interesting. so if I name a folder `something/like/this` it becomes three nested folders?

balazsbajorics: is that in line with how folders work on "unix" ?

seanparsons: I think you can escape them with a backslash...

balazsbajorics (approved):

enidemi (approved):

editor/src/components/filebrowser/fileitem.tsx
Rheeseyb: Homer Disappearing

Rheeseyb: I've just tested with `mkdir` and inside my file browser (Dolphin) and it doesn't seem to be possible to make non-nested directories with forward slashes (even when escaped)

Rheeseyb: https://stackoverflow.com/a/9847573

seanparsons (approved):

PR #604, opened 2020-10-09
Fall back to standard React handling of undefined elements

Rheeseyb:
Follow up to the conversation on #591, specifically [here](https://github.com/concrete-utopia/utopia/pull/591#discussion_r501040300)

This removes our custom handling of undefined elements, allowing us to fall through to using whatever factory function is in scope to attempt (and fail) to create them, resulting in an error that is _almost_ in line with what you'd expect from React:

Screenshot_20201009_173047
(note the `ForwardRef`, which should actually be `App`)

From my testing this appears to give consistently the same error, but maybe @seanparsons you were trying something that I've missed? 

Happy to discuss this approach further because I'm not attached to it, but it does feel closer to what we should be doing than the current custom logic.


balazsbajorics:
I haven't tried it but this sounds like the good approach to me

seanparsons (approved):

PR #607, opened 2020-10-12
show different outlines for different layouttypes

enidemi:
Canvas experiment

#### How to try it?

Select different elements on the canvas in this pizza link: https://utopia.pizza/p/eed9f3d1-tame-couch/?branch_name=experiment/layouttype-outline

### DIFF
--- a/editor/src/components/canvas/controls/outline-control.tsx
+++ b/editor/src/components/canvas/controls/outline-control.tsx
@@ -23,6 +23,8 @@ export function getSelectionColor(
   imports: Imports,
   createsYogaLayout: boolean,
   anySelectedElementIsYogaLayouted: boolean,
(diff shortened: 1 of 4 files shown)


maltenuhn:
**What I think this does** Err, it seems like  selection boxes come in different hues depending on the layout mode for self. I think. TBH I had to read the PR description a few times to make sure.

**First impressions** Welp. So I guess it's a good idea, but it's _way_ too subtle. I ended up activating the `layout info box` mod hoping the colors would match, but nope. 

**Ideas** The outline makes this tricky, mainly because it's already subtle; in some colours I don't see it at all :| that said, I got used to it!  Maybe we could use colour on the resize controls (if any)? Or add an indicator to the mouse pointer?

PR #608, opened 2020-10-12
experiment/floating-menu-1

enidemi:
Canvas Experiment

#### How to try it?
Select elements on the canvas and you should see a floating warning menu when a flow element has top/left/right/bottom/centerX/centerY.

Pizza: https://utopia.pizza/p/1b317322-foam-apogee/?branch_name=experiment/floating-menu-1


maltenuhn:
**What I think this does** Tells me off I write position-y things into my `style` object that won't do anything.

**First impression** Well, it's nothing if not in your face :) Perhaps too much so even, for once - in particular since no canvas operation should _result_ in this state.

**Ideas**  I was kinda expecting there to be a linter issue in the code as well though, and something in the navigator. More forward looking, this made me think of the vscode errors with actions to fix them (eg "add absolute positioning"), or possibly via the context menu?

PR #609, opened 2020-10-12
only collect computedStyles metadata for the selected views

balazsbajorics:
**Problem:**
Performance significantly regressed in the last 6 months.
One of the most glaring problems was a function called `getComputedStyle` in the DOM-walker. It took 6-10ms for a single element, and runs for every element every frame.

**Fix:**
Only run getComputedStyle for elements which are currently selected. Since we only show the inspector for selected elements, we only need to have computedStyles for elements which are selected. This means we can skip evaluating computedStyles for every other element. With projects with 1000 elements, this means we can skip evaluating 999 elements :) 

**Commit Details:**
- Getting selected views inside the `useDomWalker` hook
- To make it obvious that something's amiss and avoid a silent error, I prevent the inspector from showing values if one of the computedStyles are missing
- Unfortunately it makes one of the performance tests measure 100 more rerenders now, I don't know if I can help that. I didn't investigate it too much as I intend to keep on working on performance, I'll get back to this if it is indeed indicative of a real problem, or just an artifact


enidemi (approved):

editor/src/components/canvas/dom-walker.ts
```diff
@@ -177,7 +179,11 @@ export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElem
         )
       }
 
-      function getComputedStyle(element: HTMLElement): ComputedStyle {
+      function getComputedStyle(element: HTMLElement, path: TemplatePath): ComputedStyle | null {
+        if (selectedViews.find((sv) => TP.pathsEqual(sv, path)) == null) {
```
seanparsons, line 183: If you enclose the main bulk with this condition and then `else` the `return null`, then this condition will be a little clearer I think.
```suggestion
        if (selectedViews.find((sv) => TP.pathsEqual(sv, path)) == null) {
        if (selectedViews.some((sv) => TP.pathsEqual(sv, path))) {
```

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -116,7 +116,7 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThanOrEqual(630) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(640)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThanOrEqual(740) // if this breaks, GREAT NEWS but update the test please :)
```
seanparsons, line 119: That seems surprising.

seanparsons (approved):

Rheeseyb:
> With projects with 1000 elements, this means we can skip evaluating 999 elements

What about with projects with 1001 elements?

editor/src/core/performance/performance-regression-tests.spec.tsx
balazsbajorics: I didn't yet spend time to verify, but I think what happens is that we render an inspector before the DOM walker metadata is merged in, and then we render a second time when it does arrive. and I changed the inspector to hide all values if the DOM metadata contains a null for computedStyle, which means we render an empty inspector first, and a full inspector second. If that is indeed the case, then I will change the inspector to render nothing for an empty computed style. 

there's an opportunity to use React Suspense here, and literally not render an inspector until we have the data ready, I'll see if I can use that API for our benefit

editor/src/components/canvas/dom-walker.ts
Rheeseyb: Hmmm yeah it definitely took me a few reads to understand what was going on here. Alternatively if you didn't want to do that you could pull out the condition:
```const isSelected = selectedViews.some((sv) => TP.pathsEqual(sv, path))```

File: editor/src/components/inspector/inspector.tsx [hunk 3]
```diff
@@ -805,6 +805,11 @@ export const InspectorContextProvider = betterReactMemo<{
     } else {
       const elementMetadata = MetadataUtils.getElementByInstancePathMaybe(jsxMetadataKILLME, path)
       if (elementMetadata != null) {
+        if (elementMetadata.computedStyle == null) {
```
Rheeseyb, line 808: Interesting... Why do we stop here and not do everything except the line:
```newComputedStyles.push(elementMetadata.computedStyle)```
?

balazsbajorics: maybe I should make the comment above the return statement more clear. If we continue and only skip newComputedStyles, it is very very easy to make a silent error, which 100% means we would have regressions around the computedStyles stuff

editor/src/core/performance/performance-regression-tests.spec.tsx
Rheeseyb: Ooooh ok that would absolutely make sense, since the DOM walker doesn't run until after the first Canvas render

editor/src/components/inspector/inspector.tsx
Rheeseyb: Ah ok yes please then. It concerns me that this logic is going to live here, but I don't have a better proposal tbh. The only alternative I can think of is to make the metadata a union type to represent if it is the full runtime metadata or not, but I'm not sure that would be worth the huge refactor it would require.

Rheeseyb (approved):

editor/src/components/inspector/inspector.tsx
balazsbajorics: I think I'll come back to this later this week, to prevent those 100 extra rerenders

editor/src/components/canvas/dom-walker.ts
balazsbajorics: fixed

balazsbajorics: @seanparsons in general I prefer an early return of null as opposed to "the main body of the function wrapped in `if`s" style. I think it helps clear out the edge cases at the top so you can focus on the "meat" of the function

PR #610, opened 2020-10-13
Kicking a problem a little further down the line

Rheeseyb:
**Problem:**
The first interaction with the canvas open causes an entire re-render of `EditorComponent`, because of the `UiJsxCanvas` calling callbacks that update state in that component. This is particularly insidious because it causes us to re-create the property controls iframe, which then means re-evaluating the bundle that iframe runs.

**Fix:**
The real fix would be to completely remove this pattern we have used for storing and updating the runtime errors and console logs, but in trying that out I soon discovered that trying to store those in `EditorState` causes all sorts of performance issues because `EditorCanvas` uses the entire state as one of its props. I started trying to fix that but it turned out to be 10x more work than I had anticipated. So for now I've just moved the state inside `OpenFileEditor` to reduce the impact of the illegal calls.


editor/src/components/editor/editor-component.tsx
```diff
@@ -517,20 +442,74 @@ const OpenFileEditor = betterReactMemo('OpenFileEditor', (props: OpenFileEditorP
     }
   })
 
+  const [runtimeErrors, setRuntimeErrors] = React.useState<Array<RuntimeErrorInfo>>(EmptyArray)
```
balazsbajorics, line 445: could you take this code big chunk of code and turn it into a hook? I feel bad for increasing this function's size by 50 lines here

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -59,8 +59,8 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(750) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(760)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(695) // if this breaks, GREAT NEWS but update the test please :)
+    expect(renderCountAfter - renderCountBefore).toBeLessThan(705)
```
balazsbajorics, line 63: yay

balazsbajorics (approved):

PR #612, opened 2020-10-13
Experiment/canvas set pins 1

enidemi:
Canvas Experiment based on #599 

#### How to try it?

Select elements and try dragging the TLRB pin controls on the canvas to change them. It is possible to change a left pin to right, top to bottom, bottom to top etc. The cursor shows which pin control can be changed.
Doesn't work on center pins and width/height.

https://utopia.pizza/p/4d07b7a3-rebel-yam/?branch_name=experiment/canvas-set-pins-1


balazsbajorics:
I like the control dragging! but I wouldn't discover it if I wouldn't know what to look for.

do we have any design that would help distinguish these pins from "distance indicators"?

PR #615, opened 2020-10-14
Support context elements.

### LINKED ISSUE #560: Navigator is empty in the attached project
Trying to import this project https://utopia.pizza/p/4acf68b0-nova-cairnsmore/ results in an empty navigator

seanparsons:
Fixes #560

**Problem:**
Context elements are troublesome for our spy and dom walking because they effectively create gaps in the paths used to identify elements in the hierarchy. The spy will take account of a path `aaa/bbb/ccc` for example where the middle element represents a context, but the dom walker will only see something like `aaa/ccc` which results in most of that data being culled and ending up with very little element metadata in the end.

**Fix:**
This fix attempts to resolve the issue by carrying forwards onto the child elements of a context the parts of the path that relate to the context. Then later reconstructing the path using those values so that the dom walker gets the same paths that the spy sees.

**Commit Details:**
- Fixes #560.
- `mangleExoticType` called from `patchedCreateReactElement` now inserts
  a key that represents the missing parts of the hierarchy into the child
  elements of a context element.
- `renderCoreElement` adds in the value for the `data-utopia-original-parents`
  key.
- `walkElements` inside `useDomWalker` will reconstruct the missing path hierarchy
  using the keys added in via the above two functions.
- Various cases of the constants inserted for keys like `data-uid` in and
  around the logic updated across this work.

### DIFF
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -29,8 +29,15 @@ import {
 } from '../inspector/common/css-utils'
 import { CanvasContainerProps } from './ui-jsx-canvas'
 import { camelCaseToDashed } from '../../core/shared/string-utils'
-import { UTOPIA_ORIGINAL_ID_KEY } from '../../core/model/utopia-constants'
 import { useEditorState } from '../editor/store/store-hook'
+import {
+  UTOPIA_DO_NOT_TRAVERSE_KEY,
+  UTOPIA_LABEL_KEY,
+  UTOPIA_ORIGINAL_ID_KEY,
+  UTOPIA_UID_KEY,
+  UTOPIA_UID_ORIGINAL_PARENTS_KEY,
+  UTOPIA_UID_PARENTS_KEY,
+} from '../../core/model/utopia-constants'
 
 function isValidPath(path: TemplatePath | null, validPaths: Array<string>): boolean {
   return path != null && validPaths.indexOf(TP.toString(path)) > -1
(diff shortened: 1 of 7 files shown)


editor/src/components/canvas/dom-walker.ts
```diff
@@ -316,12 +328,28 @@ export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElem
...
+          )
+
+          // Build the unique path for this element.
+          let uniquePath: TemplatePath = uniqueParentPath
+          if (parentUIDsAttribute != null) {
+            uniquePath = TP.appendToPath(uniquePath, parentUIDsAttribute.split('/'))
+          }
+          uniquePath = TP.appendToPath(uniquePath, pathElement)
+
+          // Build the original path for this element.
```
Rheeseyb, line 344: At this point would it make more sense to just store the full (stringified) original path on the element itself via something like `data-utopia-original-path`?

editor/src/utils/canvas-react-utils.ts
```diff
@@ -5,10 +5,14 @@ import { omitWithPredicate } from '../core/shared/object-utils'
 import { MapLike } from 'typescript'
 import { firstLetterIsLowerCase } from '../core/shared/string-utils'
 import { isIntrinsicHTMLElementString } from '../core/shared/element-template'
-import { UtopiaKeys } from '../core/model/utopia-constants'
+import { UtopiaKeys, UTOPIA_UID_KEY, UTOPIA_UID_PARENTS_KEY } from '../core/model/utopia-constants'
 
 const realCreateElement = React.createElement
 
+const fragmentSymbol = Symbol.for('react.fragment')
```
balazsbajorics, line 12: can't we import these symbols from React somehow? I'm a bit worried about silent errors if they change a symbol

editor/src/components/canvas/dom-walker.ts
seanparsons: I don't _think_ we can necessarily construct that in our `patchedCreateReactElement` function, as it wont know necessarily enough of the chain above it.

editor/src/utils/canvas-react-utils.ts
seanparsons: These are just moved from where they were originally, I've not added them.

editor/src/components/canvas/dom-walker.ts
Rheeseyb: Ah yeah, of course. Urgh.

editor/src/core/model/utopia-constants.ts
```diff
@@ -3,11 +3,15 @@ export const UTOPIA_LABEL_KEY = 'data-label'
 export const UTOPIA_ORIGINAL_ID_KEY = 'data-utopia-original-uid'
 export const UTOPIA_DO_NOT_TRAVERSE_KEY = 'data-utopia-do-not-traverse'
 export const UTOPIA_SCENE_ID_KEY = 'data-utopia-scene-id'
+export const UTOPIA_UID_PARENTS_KEY = 'data-utopia-parents'
+export const UTOPIA_UID_ORIGINAL_PARENTS_KEY = 'data-utopia-original-parents'
```
balazsbajorics, line 7: I forgot what's the difference between `data-uid` and `data-utopia-original-uid`, can you help me here?

seanparsons: The first one may have an index added to it if it's generated in a loop, the second one is the uid that is in the model.

balazsbajorics (approved):

PR #617, opened 2020-10-14
experiment/canvas-set-pins-2

enidemi:
Canvas Experiment

#### How to try it?

Select a pinned element and move it to the edge of the containingblock to change a `style.left` to `style.right`.  The timer is set to 1.5 seconds. You should see the containing block highlighted before the property change happens.
Please keep width/height in style props, it's a bit more limited than the inspector pin controls.

🍕 https://utopia.pizza/p/2d4da2ed-sleepy-reminder/?branch_name=experiment/canvas-set-pins-2

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -317,9 +317,13 @@ export function clearDragState(
   let result: EditorState = model
   clearTimeout((window as any)['flexParentTimer'])
   clearTimeout((window as any)['flexParentHighlightTimer'])
+  clearTimeout((window as any)['elementPinTimer'])
+  clearTimeout((window as any)['elementContainerHighlightTimer'])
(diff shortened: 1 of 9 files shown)


enidemi:
> interesting! I feel like this feature is very hidden though. in Framer the snapping to right AFAIK happens if you cross to the right half of the frame, instead of only when you are snapping to the very right edge of the parent,.

that sounds very aggressive to me, that's why I added only the edge

maltenuhn:
I agree with both of you -  It's basically undiscoverable, but it's also delightfully unaggressive :) As an experiment I feel it needs more thought, but I think the pattern we have here is a winner, because it's still faster than the inspector, and doesn't require you to context switch. 

 If we put a bit of thought into how to design it (eg make it more discoverable, and give you a sense of what will happen after the delay) we can use it in a bunch of situations.

PR #620, opened 2020-10-15
Placeholder dependency resolution callback

Rheeseyb:
**Problem:**
During a discussion we realised that our resolving of placeholder dependencies was actually very broken, and would not support importing assets to use directly in the Preview would not work, since we were only supporting side effects when importing placeholder dependencies.

**Fix:**
Support a callback to be used whenever the running code requires a placeholder dependency, allowing us to explicitly force a re-render or re-evaluation after that dependency has been downloaded. This also gives us the option to extend the behaviour to display something to signify to the user that the editor is downloading further resources (I have left 2 `// FIXME` s in the code to mark that this has _not_ been done as part of this work).

I've also refactored some of the uses of this code to support better test coverage.

**Commit Details:**
- `fetchMissingFileDependency` now returns a `Promise` containing the newly downloaded modules
- `getEditorRequireFn` now takes a callback `onRemoteModuleDownload` to allow us to explicitly deal with running code triggering a download, as well as the completion of that download
- Split out the core of `property-controls-info.tsx` into `property-controls-processor.ts` to allow testing, specifically so that I could ensure the above change was correctly implemented in the property controls processing
- Updated the preview to now trigger a re-render when a placeholder dependency has been finished downloading
- Tests (and some other refactoring to support them)!

### DIFF
--- a/editor/src/components/editor/npm-dependency/npm-dependency.ts
+++ b/editor/src/components/editor/npm-dependency/npm-dependency.ts
@@ -333,20 +333,6 @@ export function dependenciesWithEditorRequirements(
   return [...userDefinedDeps, ...EditorTypePackageDependencies]
 }
 
-export function orderedDependenciesEqual(
-  l: Array<RequestedNpmDependency>,
-  r: Array<RequestedNpmDependency>,
-): boolean {
-  return (
-    l.length === r.length &&
-    l.reduce(
-      (matches: boolean, next: RequestedNpmDependency, index: number) =>
-        matches && shallowEqual(next, r[index]),
-      true,
-    )
-  )
-}
-
 export function immediatelyResolvableDependenciesWithEditorRequirements(
   projectContents: ProjectContentTreeRoot,
 ): Array<PossiblyUnversionedNpmDependency> {
--- a/editor/src/components/editor/npm-dependency/npm-dependency.ts
+++ b/editor/src/components/editor/npm-dependency/npm-dependency.ts
@@ -26,8 +26,8 @@ import {
   packageJsonFileFromProjectContents,
   updatePackageJsonInEditorState,
 } from '../store/editor-state'
-import { objectMap } from '../../../core/shared/object-utils'
-import { mapArrayToDictionary, pluck } from '../../../core/shared/array-utils'
+import { pluck } from '../../../core/shared/array-utils'
+import { shallowEqual } from '../../../core/shared/equality-utils'
 import { useEditorState } from '../store/store-hook'
 import * as React from 'react'
 import { resolvedDependencyVersions } from '../../../core/third-party/third-party-components'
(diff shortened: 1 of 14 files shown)


editor/src/components/editor/npm-dependency/npm-dependency.ts
```diff
@@ -333,6 +333,20 @@ export function dependenciesWithEditorRequirements(
...
 
+export function orderedDependenciesEqual(
+  l: Array<RequestedNpmDependency>,
+  r: Array<RequestedNpmDependency>,
+): boolean {
+  return (
+    l.length === r.length &&
+    l.reduce(
+      (matches: boolean, next: RequestedNpmDependency, index: number) =>
+        matches && shallowEqual(next, r[index]),
```
seanparsons, line 344: Why `shallowEqual`? Does that run the risk of a change to `RequestedNpmDependency` breaking unexpectedly?

Rheeseyb: Hmmm tbh we could probably just use `fastDeepEquals` and bin off this function

seanparsons (approved):

editor/src/components/editor/npm-dependency/npm-dependency.ts
Rheeseyb: Fixed in a647bdbb

editor/src/core/es-modules/package-manager/package-manager.spec.ts
```diff
@@ -134,13 +122,14 @@ describe('ES Dependency Manager — Real-life packages', () => {
       fail(`Expected successful nodeModules fetch`)
     }
     const nodeModules = fetchNodeModulesResult.nodeModules
-    const req = getRequireFn(moduleUpdater(nodeModules), nodeModules)
+    const onRemoteModuleDownload = jest.fn()
```
balazsbajorics, line 125: you don't seem to be passing `onRemoteModuleDownload` anywhere, so the `expect(onRemoteModuleDownload).toBeCalledTimes(0)` is not really testing anything

Rheeseyb: Ooops

editor/src/core/property-controls/property-controls-processor.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+import * as antdPackagerResponse from '../es-modules/test-cases/antd-packager-response.json'
+import { PackagerServerResponse, requestedNpmDependency } from '../shared/npm-dependency-types'
+import { initPropertyControlsProcessor } from './property-controls-processor'
+import { fullNodeModulesUpdate, getThirdPartyPropertyControls } from './property-controls-utils'
+
+require('jest-fetch-mock').enableMocks()
+
+const simpleCssContent = '.utopiaClass { background-color: red; }'
+
+beforeEach(() => {
```
balazsbajorics, line 18: I wonder if we should make this beforeEach a project-level jest configuration

editor/src/core/es-modules/package-manager/package-manager.spec.ts
Rheeseyb: Fixed

editor/src/core/property-controls/property-controls-processor.spec.ts
Rheeseyb: Yeah I wonder about this too. The question is where do you draw the line here? We don't want every test performing a load of side effects before each run but at the same point we don't want to have to deal with forgetting to run the specific required side effects in the test files that need them to run. I think with something like this it should be lightweight enough to not matter if we run it before every test :shrug:

balazsbajorics: I think `resetDepPackagerCache` passes the lightweightness test, it sets a variable to `{}`

balazsbajorics (approved):

PR #622, opened 2020-10-15
Memoized Scene Root v1

balazsbajorics:
**Problem:**
The entire canvas with all its elements rerender for every single change, including scroll events.

**Fix:**
This is a first PR. I am trying to come up with a new way to replace/wrap the canvas renderer functions (such as `renderCoreElement`) with react components so they can be memoized. As a first, kind of exploratory step I replaced the SceneRoot by a new component which is memoized. It uses way less props (1) and uses context to get everything it needs. (A lot of the props were kinda hanging around just because the code was old)

**Commit Details:**
- Made new component called `SceneRootRenderer`. This is memoized aggressively.
- turned `RerenderUtopiaContext` into a use-context-selector type context so we can select a certain top level element without needing to rerender if an unrelated top level element changes
- added `ParentLevelUtopiaContext` which contains the parent template path. I use this to avoid having to pass the parent template path as parameter to the scene root. I'm not yet sure about this but I think this will be one important tool in my toolbelt when I start working on `renderCoreElement` since that function has 14 properties and most of them are things like scope and information about the parent and other context.
- I had to do ugly things in `sceneStyle` to be still compatible with our Jsx Pragma. More about the pragma later.

**Notes:**
- The JSX Pragma screws with memoized components because it creates a new style object reference for every render. Even for components that had no style object previously. This is something I want to look into tomorrow.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
@@ -1,80 +1,74 @@
 import * as React from 'react'
-import { MapLike } from 'typescript'
-import { NormalisedFrame, View } from 'utopia-api'
+import { View } from 'utopia-api'
+import { useContextSelector } from 'use-context-selector'
+import * as fastDeepEquals from 'fast-deep-equal'
 import { getValidTemplatePaths } from '../../../core/model/element-template-utils'
 import { left } from '../../../core/shared/either'
 import {
   emptySpecialSizeMeasurements,
   emptyComputedStyle,
+  JSXElement,
 } from '../../../core/shared/element-template'
-import { InstancePath, TemplatePath, SceneContainer } from '../../../core/shared/project-file-types'
+import {
+  InstancePath,
+  ScenePath,
+  ScenePinnedContainer,
+} from '../../../core/shared/project-file-types'
 import { colorTheme, UtopiaStyles } from '../../../uuiui'
-import { UIFileBase64Blobs } from '../../editor/store/editor-state'
 import { UiJsxCanvasContextData, UiJsxCanvasContext } from '../ui-jsx-canvas'
-import { ComponentRendererComponent } from './ui-jsx-canvas-component-renderer'
-import { RerenderUtopiaContext, SceneLevelUtopiaContext } from './ui-jsx-canvas-contexts'
+import {
+  ComponentRendererComponent,
+  isComponentRendererComponent,
+} from './ui-jsx-canvas-component-renderer'
+import {
+  MutableUtopiaContext,
+  ParentLevelUtopiaContext,
+  RerenderUtopiaContext,
+  SceneLevelUtopiaContext,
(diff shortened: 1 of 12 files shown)


editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
```diff
@@ -83,57 +77,107 @@ export const SceneRoot: React.FunctionComponent<SceneRootProps> = (props) => {
...
+          style={sceneStyle}
+          layout={sceneProps.layout}
+        >
+          {rootElement}
+        </View>
+      </SceneLevelUtopiaContext.Provider>
+    )
+  },
+  (prevProps, nextProps) => {
+    // TODO BEFORE MERGE remove me, this is only needed until I fix the pragma
```
enidemi, line 176: 👀

enidemi (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
balazsbajorics: fixed hahaha

Rheeseyb:
> I am trying to come up with a new way to replace/wrap the canvas renderer functions (such as `renderCoreElement`) with react components so they can be memoized.

So that was actually the first solution we came up with, but we discovered that by introducing our own components into the hierarchy of the user's app it meant that we were changing their app in a way that wouldn't be clear to them, and that we couldn't guarantee it wouldn't completely break the behaviour of their app, because e.g. a simple app:
```
<App>
  <ComponentA>
    <ComponentB />
  </ComponentA>
</App>
```
would previously become something like:
```
<App>
  <CoreElement>
    <ComponentA>
      <CoreElement>
        <ComponentB />
      </CoreElement>
    </ComponentA>
  </CoreElement>
</App>
```
You might remember we discovered how much of an issue this would be when trying to make sure we were correctly passing down props, and realised that it meant we wouldn't e.g. support any of the React functions for checking or manipulating children (since those would always be using our `<CoreElement/>` s rather than the actual children).

Rheeseyb:
...and I've just realised we've ended up doing that anyway with our `SpyWrapper` component. Crud.

balazsbajorics:
> You might remember we discovered how much of an issue this would be when trying to make sure we were correctly passing down props, and realised that it meant we wouldn't e.g. support any of the React functions for checking or manipulating children (since those would always be using our `<CoreElement/>` s rather than the actual children).

I remember the struggling but not all the details. My memory is that the solution was that we make sure that the props are in the `props` of the elements and not passed in via an other method. So if a parent inspects the props of the children, it will find all the expected props. And if the parent alters the props, the element will render using the new props.

I _believe_ the SpyWrapper is well-behaved in this sense.

Rheeseyb (approved):

Rheeseyb:
> I remember the struggling but not all the details.
Yeah... same here :grimacing:

seanparsons (approved):

PR #624, opened 2020-10-19
Support adding a storyboard file.

### LINKED ISSUE #619: Add a UI element that adds a storyboard file.
- Should choose a component called `App` or `Application` or otherwise the first component that has been found.

seanparsons:
Fixes #619

**Problem:**
For an imported project, a user needs a way of bringing their project into the fold of Utopia so that they can see their work.

**Fix:**
Added functionality that searches a project for components, prioritising certain key names but otherwise defaulting to the first component found. If anything has been found, a file containing a `Storyboard` definition is generated pointing at that component.

**Commit Details:**
- Fixes #619.
- Adds `AddStoryboardFile` editor action.
- Deleted `setSceneContainerValueAtPath` because it was unused.
- Added triggering button to the `FileBrowser` component.
- Adds `createSceneFromComponent` and `createStoryboardElement`
  utility functions.
- Changed `addUtopiaCanvasComponentOrDefault` to not insert an empty
  canvas component.
- Implemented `addStoryboardFileToProject` which does the work of
  searching for potential "main" components and then builds a storyboard
  file for one.

### DIFF
--- a/editor/src/core/model/__snapshots__/storyboard-utils.spec.ts.snap
+++ b/editor/src/core/model/__snapshots__/storyboard-utils.spec.ts.snap
@@ -1,0 +1,79 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`addStoryboardFileToProject adds storyboard file to project that does not have one 1`] = `
+Array [
+  Object {
+    "arbitraryJSBlock": null,
+    "isFunction": false,
+    "name": "storyboard",
+    "param": null,
+    "propsUsed": Array [],
+    "rootElement": Object {
+      "children": Array [
+        Object {
+          "children": Array [],
+          "metadata": null,
+          "name": Object {
+            "baseVariable": "Scene",
+            "propertyPath": Object {
+              "propertyElements": Array [],
+            },
+          },
+          "props": Object {
+            "component": Object {
+              "definedElsewhere": Array [
+                "App",
+              ],
+              "javascript": "App",
+              "sourceMap": null,
+              "transpiledJavascript": "return App",
+              "type": "ATTRIBUTE_OTHER_JAVASCRIPT",
+              "uniqueID": "",
+            },
+            "data-uid": Object {
+              "type": "ATTRIBUTE_VALUE",
+              "value": "scene-1",
+            },
+            "props": Object {
+              "type": "ATTRIBUTE_VALUE",
+              "value": Object {},
+            },
+            "style": Object {
+              "type": "ATTRIBUTE_VALUE",
+              "value": Object {
+                "height": 812,
+                "left": 0,
+                "position": "absolute",
+                "top": 0,
+                "width": 375,
+              },
+            },
+          },
+          "type": "JSX_ELEMENT",
+        },
+      ],
+      "metadata": null,
+      "name": Object {
+        "baseVariable": "Storyboard",
+        "propertyPath": Object {
+          "propertyElements": Array [],
+        },
+      },
+      "props": Object {
+        "data-uid": Object {
+          "type": "ATTRIBUTE_VALUE",
+          "value": "utopia-storyboard-uid",
+        },
+        "layout": Object {
+          "type": "ATTRIBUTE_VALUE",
+          "value": Object {
+            "layoutSystem": "pinSystem",
+          },
+        },
+      },
+      "type": "JSX_ELEMENT",
+    },
+    "type": "UTOPIA_JSX_COMPONENT",
+  },
+]
+`;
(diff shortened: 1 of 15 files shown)


editor/src/core/model/__snapshots__/storyboard-utils.spec.ts.snap
```diff
@@ -0,0 +1,79 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`addStoryboardFileToProject adds storyboard file to project that does not have one 1`] = `
```
balazsbajorics, line 3: could you make this an inline snapshot? I think we should consider moving away from sidecar snapshot files, because they make it too easy to capture too much information – meaning that the test will more likely fail for unrelated things. 
Also, inline snapshots are way more human readable, even if they are long, they are colocated with the test itself, no need to hunt them down by the name.

seanparsons: That sounds like a wider decision though that we probably shouldn't make in a PR.

balazsbajorics (approved):

editor/src/core/model/storyboard-utils.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+export function addStoryboardFileToProject(editorModel: EditorModel): EditorModel | null {
+  const storyboardFile = getContentsTreeFileFromString(
+    editorModel.projectContents,
+    StoryboardFilePath,
+  )
+  if (storyboardFile == null) {
+    let namedComponentToImport: ComponentToImport | null = null as any
```
Rheeseyb, line 47: Why the `as any` casts here?

editor/src/core/model/storyboard-utils.ts
```diff
@@ -0,0 +1,129 @@
...
+    let firstComponentToImport: ComponentToImport | null = null as any
+    walkContentsTree(editorModel.projectContents, (fullPath, file) => {
+      // Just in case we ever find a main looking component, skip everything else.
+      if (namedComponentToImport == null) {
+        if (isUIJSFile(file)) {
```
Rheeseyb, line 52: I suspect this check is likely to break when we separate storyboard files from regular files in the near future. Right now this check is effectively checking that the file path is `src/app.js`, which _is_ what we want to check for this piece of code, but there's a question around whether we'll still consider that file to be `'UI_JS_FILE'` after we split out storyboard files. Not something to address right now, but something to keep in mind as we might need to come back and change this in the near future.

I'm wondering if what we'll end up with is a type that represents any parsed JSX files, and a type that represents storyboard files :thinking:

editor/src/core/model/storyboard-utils.ts
seanparsons: Because otherwise TypeScript inferred the real type of the two fields as `null`, which then means the else clause of the null check thinks the type is `never` and nothing works.

editor/src/core/model/storyboard-utils.ts
seanparsons: If the import treats every piece of code as a stock code file that's definitely a big problem as nothing will be parsed for JSX elements. There's not really much in `UIJSFile` that is storyboard specific as we peer inside that representation to see if there is a storyboard.

editor/src/core/model/storyboard-utils.ts
Rheeseyb: Oh I remember you saying about this. Urgh.

Rheeseyb:
What's the deal with the removal of `EmptyUtopiaCanvasComponent` everywhere?

seanparsons:
> What's the deal with the removal of `EmptyUtopiaCanvasComponent` everywhere?

It's an empty storyboard component which doesn't really confer any benefit and was confusing matters when looking to see if a file had a storyboard in it because the logic streamlined it into the parse result.

editor/src/core/model/storyboard-utils.ts
Rheeseyb: Yeah this is what I mean - right now the project import uses a check to see if the file should be a `'UI_JS_FILE'` based on that same file path (the assumption comes from [this line in `fileTypeFromFileName`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/model/project-file-utils.ts#L392)). I can see us ending up with a `JSX_FILE` type or similar which parses all code that includes the JSX pragma, probably in the very near future tbh as we move on to supporting canvas manipulation of arbitrary projects.

Rheeseyb:
> > What's the deal with the removal of `EmptyUtopiaCanvasComponent` everywhere?
> 
> It's an empty storyboard component which doesn't really confer any benefit and was confusing matters when looking to see if a file had a storyboard in it because the logic streamlined it into the parse result.

Ok got it, thanks. :+1: to the PR

PR #625, opened 2020-10-20
experiment/mouse-pointers

enidemi:
Canvas experiment

#### How to try it?
Canvas controls are showing different cursors for layout types.

https://utopia.pizza/p/57c20d14-surf-random/?branch_name=experiment/mouse-pointers


maltenuhn:
**What I think this does** Shows me a flex mouse pointer for flex things, and others for... others. 

**First impression** My pointer designs suck but the combo of dot and colour at least makes it super obvious what's _not_ absolute. I had a temporary moment of wondering "is this about the parent, or about self?", but it went away pretty quickly. The principle really works for me (I've not tried to click-and-drag a single flex child!), and holy shit there's a lot of flex elements in the sample projects :D 

**Improvements** Better pointer designs for sure (the circle suffers from not knowing where the target is). Same pointers in navigator maybe. And some meditation on how to sidestep the temporary confusion about self-or-parent.

PR #626, opened 2020-10-20
experiment/flow-margin-padding

enidemi:
Canvas experiment

#### How to try it?
Mouseover on the edge of an element to see the Resize Menu that shows option to set margin and padding. Cycle through the menu using the shift modifier key.
It uses the Element Resize Menu feature flag and please uncheck the Flow Resize feature.

https://utopia.pizza/p/644f217d-reminiscent-place/?branch_name=experiment/flow-margin-padding

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -1688,7 +1688,18 @@ export function produceResizeSingleSelectCanvasTransientState(
         x: roundedFrame.width - originalFrame.width,
         y: roundedFrame.height - originalFrame.height,
       } as CanvasVector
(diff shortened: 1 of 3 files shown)


maltenuhn:
**What I think this does** Give me additional options to control padding and margin, reachable with `Left Shift`
**First reactions** Yep, this feels good: Intuitive, in the right place, and fast. I'm missing visual feedback during the adjustment though, it feels weird that the controls for padding/margin show up _after_ setting the values.
**Ideas for improvement** (1) Visual feedback during the adjustment (or 'while padding / margin are selected') would be more helpful than after; we could even _only_ show those controls when you've got the appropriate setting selected. (2) If padding or margin are already set for that edge, I'd love to default to that setting. (3) dealing with `padding: x` (maybe just another entry in the menu?).

PR #631, opened 2020-10-21
Webpack File Loader

### LINKED ISSUE #611: Support basic asset importing
Webpack and CRA projects support asset importing via `import icon from './icon.jpg`. This works by using a [`url-loader`](https://github.com/webpack-contrib/url-loader), which falls back to a [`file-loader`](https://github.com/webpack-contrib/file-loader), which creates a module that simply exports the URL of the asset to load.

In order to support this form of imports, we need a way of supporting loaders that create modules for non-js code.

Rheeseyb:
Fixes #611 

**Problem:**
Webpack and CRA support importing asset files directly via loader plugins (which are configured by default for CRA projects).

**Fix:**
Provide a way for hooking in "loaders" to create the modules that the above mentioned plugins would otherwise create on the fly. ~These loaders create source files which are written to BrowserFS, so that the TS compiler can then compile them into actual modules, allowing the imports to work both in the canvas and in the preview.~

This work exposed an issue that we weren't previously aware of, in that we our canvas's import resolution logic differed from the previews, since the preview supported esModuleInterop, whereas the canvas didn't (which is required for the file loader logic to mirror that used by webpack and CRA). I have fixed that as part of this PR by following TS's lead with the helper functions it defines. I can pull that out into a separate PR if needed.

**Edit:** This work has been extended to apply the loaders when loading the file (so it is no longer written to BrowserFS), and also to ensure that the loaders are applied during the module resolution that happens inside our `getRequireFn`, meaning that they will now also be applied to imports from 3rd party modules in both the Canvas and Preview.

**Commit Details:**
- Created `webpack-loaders/loader-types.ts` for defining the interface that should be used by any loaders we implement
- Defined `file-loader.ts` which is based on webpack's [file-loader](https://github.com/webpack-contrib/file-loader), for turning asset imports into their urls
- Updated `ts-worker.ts` to check if a loader exists for a given file, and use that to load the file ~*at the point of writing the file to BrowserFS*, falling back to the default behaviour if not. We do it here because to create it on the fly causes discrepancies between our Canvas and Preview logic (since the resultant module wouldn't be in the build result, and so wouldn't enter the code result cache)~ **Edit:** this behaviour has now changed, and the loaders are now applied on the fly when attempting to import the file.
- Wrote helper functions inspired by those created by TS in `commonjs-interop.ts` for supporting the common importing styles (required for the file loader)
- Updated `npm-dependency.ts` to use those helpers
- Wrote tests for all of the above
- Also a bonus test (provided by Mr Parsons) for testing the CSS evaluation - this will likely be replaced / relocated when we replace the CSS custom evaluation with a loader plugin.

**Edit: Further Commit Details:**
- Added `loaders.ts` for applying all matching loaders to a given file, and (if necessary) returning an updated filename (this is required since we explicitly prevent evaluation of filenames that don't end with `'.js'` or `'.css'`, something we share with the TS compiler, a behaviour that maybe should go out of the window after replacing the custom CSS evaluator with a loader).
- Changed `module-resolution.ts` to actually resolve the module (not just the path to it) from _either_ the `nodeModules` or the `projectContents`. This includes applying the loaders to the result.
- Updated `getRequireFn` to also use the `projectContents` for module resolution - this blows a use of memoization that I'm not entirely sure we even care about?
- Added more tests


editor/src/core/shared/file-utils.ts
```diff
@@ -167,3 +167,8 @@ export function getFileExtension(filename: string): string {
     return '' // TODO How do we handle files with no extension?
   }
 }
+
+export function directoryForFilePath(path: string): string {
+  const finalSlashIndex = path.lastIndexOf('/')
```
seanparsons, line 172: Would be nice if we could use `path.dirname` in here rather than rolling our own...

editor/src/core/shared/project-file-types.ts
```diff
@@ -107,9 +107,9 @@ export function importAliasEquals(first: ImportAlias, second: ImportAlias): bool
 }
 
 export interface ImportDetails {
-  importedWithName: string | null
-  importedFromWithin: Array<ImportAlias>
-  importedAs: string | null
+  importedWithName: string | null // import name from './place'
```
seanparsons, line 110: Good idea adding the comments.

editor/src/core/webpack-loaders/file-loader.ts
```diff
@@ -0,0 +1,15 @@
+import { ProjectFile } from '../shared/project-file-types'
+import { LoadModule, MatchFile, ModuleLoader } from './loader-types'
+
+const matchFile: MatchFile = (filename: string) => {
+  return [/\.avif$/, /\.bmp$/, /\.gif$/, /\.jpe?g$/, /\.png$/].some((regxp) => regxp.test(filename))
```
seanparsons, line 5: Since these always appear to be suffixes you might want to make these straight strings and use `.endsWith` instead as regexes are sloooooow.

editor/src/core/workers/ts/ts-worker.ts
```diff
@@ -498,6 +545,12 @@ function isNodeExported(node: TS.Node): boolean {
   )
 }
 
+const matcher = (importPath: string) => {
+  return [/\.avif$/, /\.bmp$/, /\.gif$/, /\.jpe?g$/, /\.png$/].some((regxp) =>
```
seanparsons, line 549: This is duplicated from elsewhere where I made the comment about these being suffixes right? Would be nice to re-use that function or at least the array of values.

seanparsons (approved):

editor/src/core/shared/file-utils.ts
Rheeseyb: I didn't end up needing this it seems (it was used in the two transform iterations I attempted prior), so I'll bin it off.

editor/src/core/webpack-loaders/file-loader.ts
Rheeseyb: Good shout

editor/src/core/workers/ts/ts-worker.ts
Rheeseyb: Oops, this is left over duplicate code. Deleting.

editor/src/core/shared/file-utils.ts
Rheeseyb: Fixed in 828a4a3

editor/src/core/webpack-loaders/file-loader.ts
Rheeseyb: Fixed in 828a4a3

editor/src/core/workers/ts/ts-worker.ts
Rheeseyb: Fixed in 828a4a3

editor/src/components/editor/npm-dependency/npm-dependency.spec.ts
```diff
@@ -122,3 +124,49 @@ describe('Checking the version type of a dependency', () => {
...
+  })
+
+  it('Handles all forms of imports for a common js module', () => {
+    // import cake, { icing as doIWantIcing } from './cake'
+    // import * as cakeStuff from './cake'
+    const imports = importDetails('cake', [importAlias('icing', 'doIWantIcing')], 'cakeStuff')
+
+    const defaultExport = 'A fully fledged cake'
+
+    // module.exports = 'A fully fledged cake'
```
balazsbajorics, line 160: nitpicky, but is this also called `default` export in commonJS parlance?

editor/src/core/webpack-loaders/file-loader.ts
```diff
@@ -0,0 +1,17 @@
...
+import { LoadModule, MatchFile, ModuleLoader } from './loader-types'
+
+const matchFile: MatchFile = (filename: string) => {
+  return ['.avif', '.bmp', '.gif', '.jpg', '.jpeg', '.png'].some((extension) =>
+    filename.endsWith(extension),
+  )
+}
+
+const loadModule: LoadModule = (filename: string, _: ProjectFile) => {
+  return `module.exports = '.${filename}'`
```
balazsbajorics, line 11: why the `.`? what will it be relative to?

editor/src/core/webpack-loaders/file-loader.ts
```diff
@@ -0,0 +1,17 @@
+import { ProjectFile } from '../shared/project-file-types'
+import { LoadModule, MatchFile, ModuleLoader } from './loader-types'
+
+const matchFile: MatchFile = (filename: string) => {
+  return ['.avif', '.bmp', '.gif', '.jpg', '.jpeg', '.png'].some((extension) =>
+    filename.endsWith(extension),
+  )
+}
+
+const loadModule: LoadModule = (filename: string, _: ProjectFile) => {
```
balazsbajorics, line 10: you can omit `_: ProjectFile` and it will still be correctly typed – unless you wanted to leave this here as an example

editor/src/core/es-modules/package-manager/module-resolution.ts
```diff
@@ -84,33 +106,107 @@ function normalizePath(path: string[]): string[] {
...
+
+const nodeModulesFileLookup: (nodeModules: NodeModules) => FileLookupFn = (
   nodeModules: NodeModules,
-  path: string[],
-): ESCodeFile | ESRemoteDependencyPlaceholder | null {
-  return nodeModules[path.join('/')] ?? null
+) => {
+  return fallbackLookup((path: string[]) => {
+    const filename = path.join('/')
+    return fileLookupResult(filename, nodeModules[filename])
```
Rheeseyb, line 160: I need to call `applyLoaders` here

editor/src/core/webpack-loaders/loader-types.ts
```diff
@@ -0,0 +1,21 @@
+import { ProjectFile } from '../shared/project-file-types'
+
+export type MatchFile = (filename: string) => boolean
```
Rheeseyb, line 3: Add a comment about how this can be extended in future to support https://webpack.js.org/configuration/module/#rule-conditions

editor/src/core/es-modules/package-manager/package-manager.ts
```diff
@@ -24,14 +25,15 @@ export function createDependencyNotFoundError(importOrigin: string, toImport: st
 }
 
 export const getEditorRequireFn = memoize(
-  (nodeModules: NodeModules, dispatch: EditorDispatch) => {
+  // FIXME De we even _care_ about memoization here? I don't think we do...
```
Rheeseyb, line 28: Bin off memoization completely here

editor/src/core/webpack-loaders/default-loader.ts
```diff
@@ -0,0 +1,16 @@
+import { LoadModule, loadModuleResult, MatchFile, ModuleLoader } from './loader-types'
+
+const matchFile: MatchFile = (filename: string) => {
+  return ['.js', '.jsx', '.ts', '.tsx', '.d.ts', '.json'].some((extension) =>
+    filename.endsWith(extension),
+  )
+}
+
+const loadModule: LoadModule = (filename: string, contents: string) => {
+  return loadModuleResult(filename, contents)
```
Rheeseyb, line 10: Add a FIXME me here to investigate whether we should run the Babel transform for converting to Common JS

Rheeseyb (commented):
Note: With this work we are getting closer to making the ts-worker redundant. We should investigate replacing that completely with a Babel transform.

editor/src/core/es-modules/commonjs-interop.ts
```diff
@@ -0,0 +1,28 @@
...
+export function importStar(mod: unknown): unknown {
+  if (isESModule(mod)) {
+    return mod
+  } else {
+    let result: any = { default: mod }
+
+    if (typeof mod === 'object' && mod != null) {
+      for (var k in mod) {
+        if (k !== 'default' && mod.hasOwnProperty(k)) {
+          result[k] = (mod as any)[k]
```
Rheeseyb, line 17: Replace with https://github.com/microsoft/TypeScript/blob/master/src/compiler/factory/emitHelpers.ts#L782 for better mutation support (I never ever ever thought I would want to say that)

editor/src/components/editor/npm-dependency/npm-dependency.spec.ts
```diff
@@ -122,3 +124,49 @@ describe('Checking the version type of a dependency', () => {
...
+
+    const importResult = importResultFromModule(imports, cakesModule)
+    expect(importResult).toEqual({
+      cake: defaultExport,
+      doIWantIcing: icingExport,
+      cakeStuff: cakesModule,
+    })
+  })
+
+  it('Handles all forms of imports for a common js module', () => {
```
Rheeseyb, line 153: Add FIXME to include other export types here (if necessary)

balazsbajorics (approved):

enidemi (approved):

PR #636, opened 2020-10-22
experiment/insertion-plus-button-1

enidemi:
Canvas Experiment

#### How to try it?

Select flex children or flow children, you should see markers on the side of the parent element. On hover the marker becomes a plus sign, clicking on it inserts a new element with a fixed 50x50 size.
The insert controls are not shown for absolute elements and doesn't work well with inline siblings (it works with only 1 span element between divs for example)

Make sure the feature flag `Insertion Plus Button` is enabled.

pizza: https://utopia.pizza/project/57c20d14/?branch_name=experiment/insertion-plus-button-1

### DIFF
--- a/editor/src/components/canvas/controls/insertion-plus-button.tsx
+++ b/editor/src/components/canvas/controls/insertion-plus-button.tsx
@@ -1,0 +1,239 @@
+import { ControlProps } from './new-canvas-controls'
+import * as React from 'react'
+import { MetadataUtils } from '../../../core/model/element-metadata-utils'
+import * as TP from '../../../core/shared/template-path'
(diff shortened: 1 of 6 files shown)


maltenuhn:
**What I think this does** For flow contexts, let's me insert things before things, and after things, and at the start and end of things!
**First Impression** Sweet! It gets me just before I try to insert things by moving stuff out of the way, and gives me an easier way of dealing with that. It's quite in your face, but even on top of a busy layout I found it worked pretty well.
**Ideas for improvement** Nothing major except the sizing / styling of the thing, and I find myself wanting to change its sizing and element type (eg turn it into a `<button/>` via the context menu.

PR #637, opened 2020-10-22
experiment/text-editing-for-simple-text

enidemi:
Canvas Experiment

#### How to try it?
With this experiment you should see an input box on the top area of the canvas. When having a span/div element selected that has a simple text OR it's empty you can type text into the input field. If a div has multiple child elements it is non-interactive. Doesn't work with expressions or more complex things.
Uses the feature flag `Edit Simple Text`.

⚠️ It has horrible performance, please type slowly ⚠️

https://utopia.pizza/project/57c20d14/?branch_name=experiment/text-editing-for-simple-text

### DIFF
--- a/editor/src/components/canvas/controls/insertion-plus-button.tsx
+++ b/editor/src/components/canvas/controls/insertion-plus-button.tsx
@@ -23,7 +23,7 @@ interface ButtonControlProps {
   indexPosition: IndexPosition
 }
 
-export const InsertionControls = (props: ControlProps): (JSX.Element | null)[] | null => {
+export const InsertionControls = (props: ControlProps): JSX.Element | null => {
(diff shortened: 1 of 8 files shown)


maltenuhn:
**What I...** Well, 
**First reaction** Oh wow, the idea is definitely a bit formulaic in its origins, but it's not shit, and I could see this approach spreading. I can even type `{props.userName}`. Being able to select a microscopic span and then softly editing the text inside _without WYSIWYG nightmare fuel_ truly is excellent. One to add to the league table of successful experiments IMO! Even in its visually pretty basic application right now.
**Ideas for improvement** The one thing I found myself trying to do a few times is double-click a leaf span and waiting / hoping / expecting to the bar. Also, maybe updating on Enter / cmd+enter is adequate?

PR #639, opened 2020-10-23
refactor(editor) Unify the two text file types.

seanparsons:
**Problem:**
Currently there is a somewhat arbitrary split between `CODE_FILE` and `UI_JS_FILE` which ends up being problematic if something is the wrong kind and for imports ideally we want everything to be capable of having exportable components.

**Fix:**
Create a new file type based on `UI_JS_FILE` and unify the two file types against that.

**Bugs:**
- Currently there's no distinction between differing types of files so the canvas logic will currently attempt to run a `package.json` file for instance.
- For some reason after a Github project import, the files in the project are all being flagged as having been modified.

**Commit Details:**
- Replace both `UI_JS_FILE` and `CODE_FILE` types with `TEXT_FILE`.
- Created `ParsedTextFile` which replaces the previous `ParseResult`
  type which was an `Either`, so that it can represent possibly
  unparseable text files.
- Added some additional utility files to help smooth the transition
  away from the `ParseResult` type.
- Reshape the content such that the new `TextFile` type consists of
  one or two `TextFileContents` instances.
- Applied the migration earlier in the load flow so that it happens
  before some calls that need the new model.


editor/src/core/shared/project-file-types.ts
```diff
@@ -132,31 +131,84 @@ export interface HighlightBounds {
...
-  return isLeft(result)
+export function isParseFailure(parsed: ParsedTextFile): parsed is ParseFailure {
+  return parsed.type === 'PARSE_FAILURE'
+}
+
+export interface Unparsed {
+  type: 'UNPARSED'
+}
+
+export function unparsed(): Unparsed {
```
Rheeseyb, line 162: This could be replaced with a `const`

editor/src/components/code-editor/monaco-wrapper.tsx
```diff
@@ -360,8 +359,8 @@ export class MonacoWrapper extends React.Component<MonacoWrapperProps, MonacoWra
                 filename !== this.props.filename &&
                 file != null &&
                 (isDirectory(file) ||
-                  isUIJSFile(file) ||
-                  (isCodeFile(file) && isJsOrTsFile(filename)))
+                  isTextFile(file) ||
```
Rheeseyb, line 362: Looks like this part of the condition should be removed :thinking:

editor/src/components/code-editor/monaco-wrapper.tsx
```diff
@@ -374,7 +373,7 @@ export class MonacoWrapper extends React.Component<MonacoWrapperProps, MonacoWra
                     : monaco.languages.CompletionItemKind.File
 
                   const textToInsert =
-                    isUIJSFile(file) || isCodeFile(file)
+                    isTextFile(file) || isTextFile(file)
```
Rheeseyb, line 376: Ha!
```suggestion
                    isTextFile(file)
```

editor/src/components/editor/editor-component.tsx
```diff
@@ -505,7 +505,7 @@ const OpenFileEditor = betterReactMemo('OpenFileEditor', () => {
     const openEditorTab = getOpenEditorTab(store.editor)
     return {
       noFileOpen: openEditorTab == null,
-      isUiJsFileOpen: selectedFile != null && isUIJSFile(selectedFile),
+      isUiJsFileOpen: selectedFile != null && isTextFile(selectedFile),
```
Rheeseyb, line 508: I wonder if this (and other places that were assuming a `UIJSFile` previously) should also check that the parse status isn't `Unparsed` :thinking:

editor/src/components/editor/global-shortcuts.ts
```diff
@@ -726,7 +726,7 @@ export function handleKeyDown(
       return []
     } else {
       switch (openFile.type) {
-        case 'UI_JS_FILE':
+        case 'TEXT_FILE':
```
Rheeseyb, line 729: This might need rethinking

editor/src/core/model/project-file-utils.ts
```diff
@@ -376,14 +359,14 @@ export function fileTypeFromFileName(
     return null
   }
   if (filename.endsWith('src/app.js')) {
-    return 'UI_JS_FILE'
+    return 'TEXT_FILE'
```
Rheeseyb, line 362: This first branch can go in the bin now

editor/src/core/model/project-import.ts
```diff
@@ -132,31 +129,25 @@ export async function importZippedGitProject(
...
           } else {
-            if (expectedFileType === 'UI_JS_FILE') {
-              // TODO We really need a way to represent unparsed files, or a way to separate the parsed file from the contents
-              loadedProject = addFileToProjectContents(
-                loadedProject,
-                shiftedFileName,
-                uiJsFile(
-                  left(parseFailure(null, null, null, [], loadedFile)),
-                  null,
+            // TODO We really need a way to represent unparsed files, or a way to separate the parsed file from the contents
```
Rheeseyb, line 137: Now we don't :grinning:

editor/src/core/model/project-import.ts
```diff
@@ -132,31 +129,25 @@ export async function importZippedGitProject(
...
+            loadedProject = addFileToProjectContents(
+              loadedProject,
+              shiftedFileName,
+              textFile(
+                textFileContents(
+                  loadedFile,
+                  parseFailure(null, null, null, []),
```
Rheeseyb, line 144: Now this can become `unparsed`, and we could probably update the revision state to `CodeAhead`

editor/src/core/workers/bundler-bridge.ts
```diff
@@ -426,7 +426,7 @@ function sendIdGuardedinitializeWorkerPromise(
 function sendIdGuardedUpdateFilePromise(
   worker: BundlerWorker,
   filename: string,
-  content: string | UIJSFile | CodeFile,
+  content: string | TextFile | TextFile,
```
Rheeseyb, line 429:
```suggestion
  content: string | TextFile,
```

editor/src/core/workers/ts/ts-worker.ts
```diff
@@ -74,7 +69,7 @@ export type OutgoingWorkerMessage =
 interface UpdateFileMessage {
   type: 'updatefile'
   filename: string
-  content: string | UIJSFile | CodeFile
+  content: string | TextFile | TextFile
```
Rheeseyb, line 72:
```suggestion
  content: string | TextFile
```

editor/src/core/workers/ts/ts-worker.ts
```diff
@@ -146,7 +141,7 @@ export function filterOldPasses(errorMessages: Array<ErrorMessage>): Array<Error
 
 export function createUpdateFileMessage(
   filename: string,
-  content: string | UIJSFile | CodeFile,
+  content: string | TextFile | TextFile,
```
Rheeseyb, line 144:
```suggestion
  content: string | TextFile,
```

editor/src/components/editor/store/editor-state.ts
```diff
@@ -434,7 +438,7 @@ export function getOpenUIJSFileKey(model: EditorState): string | null {
   const openEditorTab = getOpenEditorTab(model)
   if (
     openFile != null &&
-    isUIJSFile(openFile) &&
+    isTextFile(openFile) &&
```
Rheeseyb, line 441: This definitely feels like a candidate for some kind of `isParsedTextFile` or something along those lines (something to signify that it's not just a regular text file, but rather a code file that we can parse)

Rheeseyb (commented):
Ok I've left some comments (mostly for myself) and taken some notes, and I'll pick this up tomorrow morning.

balazsbajorics (commented):
I like the changes in project-file-types

editor/src/core/shared/project-file-types.ts
Rheeseyb: Done in adc65a8

editor/src/components/code-editor/monaco-wrapper.tsx
Rheeseyb: Done in adc65a8

editor/src/components/code-editor/monaco-wrapper.tsx
Rheeseyb: Done in adc65a8

editor/src/components/editor/editor-component.tsx
Rheeseyb: Done in adc65a8

editor/src/core/model/project-file-utils.ts
Rheeseyb: Done in adc65a8

editor/src/core/model/project-import.ts
Rheeseyb: Done in adc65a8

editor/src/core/workers/bundler-bridge.ts
Rheeseyb: Done in adc65a8

editor/src/core/workers/ts/ts-worker.ts
Rheeseyb: Done in adc65a8

editor/src/core/workers/ts/ts-worker.ts
Rheeseyb: Done in adc65a8

editor/src/components/editor/store/editor-state.ts
Rheeseyb: Done in adc65a8

Rheeseyb:
I have updated this PR to address my feedback as well as:

- Include a new function `isParsedTextFile` for checking if a text file also contains a parse result (success or failure), so that we can still distinguish between files with contents that we can parse and files that we can't
- Used this check for the functions used for getting or updating "UI JS" files
- Added new functions `getOpenTextFileKey` and `getOpenTextFile` as sibling functions to `getOpenUIJSFileKey` and `getOpenUIJSFile` functions, with the difference being that the former don't care if the contents is parseable.

editor/src/components/editor/store/editor-state.ts
balazsbajorics: I loved getting all these emails saying done 😀

editor/src/core/shared/project-file-types.ts
```diff
@@ -190,30 +240,60 @@ export const enum RevisionsState {
...
-export function isCodeFile(projectFile: ProjectFile | null): projectFile is CodeFile {
-  return projectFile != null && projectFile.type === 'CODE_FILE'
+export function isTextFile(projectFile: ProjectFile | null): projectFile is TextFile {
+  return projectFile != null && projectFile.type === 'TEXT_FILE'
 }
 
-export function isCodeOrUiJsFile(file: ProjectFile): file is CodeFile | UIJSFile {
-  return file.type === 'CODE_FILE' || file.type === 'UI_JS_FILE'
+export function isParsedTextFile(projectFile: ProjectFile | null): projectFile is TextFile {
+  return isTextFile(projectFile) && !isUnparsed(projectFile.fileContents.parsed)
```
balazsbajorics, line 296: what if I have a parseable text file that is not parsed yet? can we use this to distinguish files that are parseable from files we don't intend to parse? is that even an important distinction?

balazsbajorics: ping Rheese

Rheeseyb: This was something we thought about but realised that in doing so we'd end up in a situation where we would be marking files as "never parse", which would end up with us also having to build in logic to handle those files changing in a way that would allow us to parse them.

So for example, say a user creates a file `thing.json`, we would want to mark that as "never parse". But they named it that by accident, because really they wanted to call it `thing.js`, so now we have to make sure that renaming to a supported file extension triggers the parsing. It's not a huge issue, but means we have to build in logic to cover edge cases that would otherwise be handled for free.

balazsbajorics (approved):

Rheeseyb:
Ok this PR actually brings about a new problem, in how we determine when to show the `SplitViewCanvasRoot` vs a regular `ScriptEditor`, which is pretty tricky since we currently have no way of knowing whether a file is intended to be editable via the canvas. The seemingly obvious answer here is to say that only files that export the magic `storyboard` element fit the bill, but that relies on the `ParsedTextFile` being a `ParseSuccess` (as well as having to iterate through the top level components), meaning that a code change that would result in successful parse flipping to a failed parse would remove the canvas and replace it with a regular code editor, meaning the failure message wouldn't be shown. We could alternatively use a regex to find that element in the raw code, but that sounds a bit brittle too, *and* on top of that we're overdue a conversation about what should actually trigger the canvas opening (and whether we should bin off the `SplitViewCanvasRoot` in favour of better tab management).

To prevent this PR from being blocked by this, I have made [this change](https://github.com/concrete-utopia/utopia/pull/639/commits/1ad92002c7e12570bba02bc447a96e2db58b8089), which for now means we will only show the canvas for specific file names.

maltenuhn:
> ... how we determine when to show the SplitViewCanvasRoot vs a regular ScriptEditor ...

@Rheeseyb can we sidestep the issue you describe make this an entry in `package.json` or wherever we'll store config settings for a project (`.utopia`)? Or a user choice, similar to how vscode lets you edit settings in their "settings editor" or in a code editor?

PR #640, opened 2020-10-26
bump zustand + add performance measurements

balazsbajorics:
**Commit Details:**
- Bumping Zustand, and fixing the breaking API changes
- Added a new way to performance.measure the zustand selectors by wrapping them in a utility function

### DIFF
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -156,7 +156,7 @@ export const defaultUserState: UserState = {
   shortcutConfig: {},
 }
 
-export interface EditorStore {
+export type EditorStore = {
   editor: EditorState
   derived: DerivedState
   history: StateHistory
--- a/editor/package-lock.json
+++ b/editor/package-lock.json
@@ -28041,9 +28041,9 @@
       }
     },
     "zustand": {
-      "version": "2.2.3",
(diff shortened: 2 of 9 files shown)


editor/src/components/editor/store/editor-state.ts
```diff
@@ -156,7 +156,7 @@ export const defaultUserState: UserState = {
   shortcutConfig: {},
 }
 
-export interface EditorStore {
+export type EditorStore = {
```
Rheeseyb, line 159: Why the change from `interface` to `type` here?

Rheeseyb (approved):

enidemi (approved):

editor/src/components/editor/store/editor-state.ts
balazsbajorics: This is a bit unclear to me.

There was a compile error that went like this:
```
Type 'EditorStore' does not satisfy the constraint 'Record<string | number | symbol, unknown>'.
  Index signature is missing in type 'EditorStore'.ts
```
Changing EditorStore from interface to type fixed this error. I don't quite know why!
The change itself means EditorStore is now a type alias to the `typeof` of the object literal type. 
Maybe that means it also gets some object-like features like an index signature?

PR #642, opened 2020-10-26
enhanced zustand selectors that can early return

balazsbajorics:
**Problem:**
Some Zustand selectors can be very expensive (sometimes in the 2-3ms range!!!), but they are still cheaper than re-rendering the components. If only there were a way to memoize the selected results.

**Fix:**
Added a second (optional!) parameter to the selector functions that can be used to compare the previous state to the new state, and bail out from recalculating the selector.

**Commit Details:**
- Selectors are now called with a second paramter containing `{ previousState, previousResult}`. if previousState === state, they can just return previousResult
- useEditorState now takes the name of the selector as an optional argument, used for debugging performance.

### DIFF
--- a/editor/src/components/canvas/controls/new-canvas-controls.tsx
+++ b/editor/src/components/canvas/controls/new-canvas-controls.tsx
@@ -214,20 +214,30 @@ const NewCanvasControlsClass = (props: NewCanvasControlsClassProps) => {
     return 'enabled'
   }
 
-  const elementsThatRespectLayout = useEditorState((store) => {
-    return flatMapArray((view) => {
-      if (TP.isScenePath(view)) {
-        const scene = MetadataUtils.findSceneByTemplatePath(store.editor.jsxMetadataKILLME, view)
-        if (scene != null) {
-          return [view, ...scene.rootElements.map((e) => e.templatePath)]
+  const elementsThatRespectLayout = useEditorState<TemplatePath[]>(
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -214,20 +214,30 @@ const NewCanvasControlsClass = (props: NewCanvasControlsClassProps) => {
...
-  const elementsThatRespectLayout = useEditorState((store) => {
-    return flatMapArray((view) => {
-      if (TP.isScenePath(view)) {
-        const scene = MetadataUtils.findSceneByTemplatePath(store.editor.jsxMetadataKILLME, view)
-        if (scene != null) {
-          return [view, ...scene.rootElements.map((e) => e.templatePath)]
+  const elementsThatRespectLayout = useEditorState<TemplatePath[]>(
+    (store, { previousResult, previousState }) => {
+      if (
+        previousState?.editor.jsxMetadataKILLME === store.editor.jsxMetadataKILLME &&
```
balazsbajorics, line 220: this is the first (and only so far) use of `previousState`

balazsbajorics:
as per our phone conversation, I'm closing this PR in favor of experimenting with reselect selectors. the reference implementation contains a mistake which kind of proves that this `shouldComponentUpdate`-style dependency checking is dangerous because it doesn't break the build if you miss a dependency and enables silent errors

PR #643, opened 2020-10-27
introducing reselect

balazsbajorics:
**Problem:**
We have some Zustand selectors that are very expensive to run (3ms per frame). We also have a _ton_ of 0.1ms-sized selectors that add up to a significant time!

**Fix:**
This PR introduces [reselect](https://github.com/reduxjs/reselect) and uses it in 1 place, in the fattest selector we had in NewCanvasControls.

**Commit Details:**
- Removing the editor dependency from `getPropertyControlsForTarget`, created a new `getPropertyControlsForTargetFromEditor` that extracts properties from the editor for the callsites I didn't want to touch now
- created the first reselect selector `selectElementsThatRespectLayout` 
- added optional `selectorName` argument to `useEditorState` for easier debugging


editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -155,6 +159,45 @@ interface NewCanvasControlsClassProps {
   windowToCanvasPosition: (event: MouseEvent) => CanvasPositions
 }
 
+const selectElementsThatRespectLayout = createSelector(
```
balazsbajorics, line 162: this is our new reselect selector! here's how it works: `createSelector` is a variadic function, the first n parameters are selectors each getting a part of the editor store, and the last parameter is a combiner which receives as arguments the values that were selected from the EditorStore.

The idea is that if the values returned by the individual selectors is `===` equal to the previously selected value, the reselect selector will return with the previously selected value instead of calling the combiner. Reselect allows us to combine selectors into larger selectors and in turn combine those into even larger selectors.

The big benefit this gives us is that if we previously had a Zustand selector that had to run an expensive computation, now we can put smaller selectors in front of it, just like in this example. If none of the navigatorTargets, propertyControlsInfo, openImports, openFilePath, rootComponents and jsxMetadataKILLME changes, we simply never call the combiner (which takes 3ms!) which saves us 3ms.

This memoized reselect selector takes 0.01ms to return if those parts of the state did not change, which is very promising as most of our current zustand selectors seem to eat around 0.05ms each.

balazsbajorics: if the variadic style with the combiner being the _last_ parameter irks you, there's an alternative form:
`createSelector([reducer1, reducer2, ...], combiner)` where you put the reducers in an array. Has the same typescript support and everything

Rheeseyb (approved):
This is pretty much exactly what I was hoping for after the call this morning, so I'm pretty happy with this! I think we could use this to provide a library of selectors for common parts of the model to make the shift towards this a bit easier.

I had a look into the reselect library and what's very weird is that the repo itself appears to be under active(?) development, with commits to master as recent as this week, but there have been no releases since 2018, so I think we'll have to use a specific commit hash for this dependency.

enidemi (approved):

balazsbajorics:
> appears to be under active(?) development

the approved PRs seem to be very minor things, like update README and them moving the repo and bumping lodash

PR #644, opened 2020-10-28
giving a debug name to every useEditorState selector

balazsbajorics:
This PR takes the optional `selectorName` in `useEditorState` and makes it non-optional. This makes debugging selector performance easier, because you don't need to dig into the stack trace to catch a glimpse of the call site (which is both inconvenient, and surprisingly unreliable (I get a lot of `anonymous@anonymous` functions which is not helpful))


enidemi (approved):

Rheeseyb (commented):
:+1:

PR #645, opened 2020-10-28
Use re-reselect and reselect on the NavigatorItemWrapper's horribly slow selector

balazsbajorics:
**Problem:**
Each NavigatorItemWrapper has a selector function that would take ~0.2-0.3ms to finish, with the number of NavigatorItems this number can grow pretty significant!

**Fix:**
Use a reselect selector to make the selector return a cached result if the relevant editor state stayed the same.

**Commit Details:**
- `navigatorItemWrapperSelector` is the first selector that takes a property. each selector is of the type `(state, props) => S`. the combiner does not get the props, so one of the selectors have to provide it. 
- **important** the cache size of a reselect selector by default is 1!! if you have more than 1 navigator items, the selectors are going to re-calculate for the various items, so it needs to be mitigated somehow
- I introduce re-reselect that is designed to take care of this problem


editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
```diff
@@ -25,16 +30,92 @@ interface NavigatorItemWrapperProps {
...
+      isSelected: TP.containsPath(templatePath, transientState.selectedViews),
+      isHighlighted: TP.containsPath(templatePath, transientState.highlightedViews),
+      noOfChildren: noOfChildrenInner,
+      supportsChildren: supportsChildren,
+      imports: imports,
+      elementOriginType: elementOriginType,
+      elementWarnings: elementWarningsInner ?? defaultElementWarnings,
+    }
+  },
+)((store, templatePath: TemplatePath) => TP.toString(templatePath))
```
balazsbajorics, line 110: this is the re-reselect part. you need to pass in a function that generates a cache key. here I use the one property (templatePath) that distinguishes the NavigatorWrapperItems from each other.

editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
```diff
@@ -25,16 +30,92 @@ interface NavigatorItemWrapperProps {
   getSelectedViewsInRange: (index: number) => Array<TemplatePath>
 }
 
+const navigatorItemWrapperSelector = createCachedSelector(
+  (store: EditorStore) => store.editor.jsxMetadataKILLME,
+  (store: EditorStore) => store.derived.canvas.transientState,
+  (store) => store.derived.navigatorTargets,
+  (store) => store.derived.elementWarnings,
+  (store, templatePath) =>
```
balazsbajorics, line 38: the second parameter of every selector is the "props", which in our case is `templatePath`.

balazsbajorics (commented):
now that I look at the result, I kinda feel like we would be better off creating selector factories and putting them in a React.useMemo. because re-reselect does not read very well and it's not intuitive if you made a mistake or not

Rheeseyb (approved):
For a bit of context here for anyone reading this PR later - we spoke about this during one of our calls yesterday and agreed that the syntax of `re-reselect` was pretty confusing, in that it requires a function that generates cache keys, but that is not clearly communicated at all to anyone reading the code, so could very easily end up in changes to the selectors that aren't reflected in the cache key generation. Instead of that, Balazs proposed using `React.useMemo` to memoize the selector creation based on the extra params it needed (in this case the template path), which is _so much clearer_ to anyone that will want to make a change to the selectors.

enidemi (approved):

PR #648, opened 2020-10-30
Introduce caching around TemplatePath creation

Rheeseyb:
**Problem:**
We are repeatedly creating new `TemplatePath` s over and over in our editor.

**Fix:**
I have copied the style set out by [`property-path.ts`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/property-path.ts) for caching, but adjusted it to support the more complex nature of `TemplatePath` s.

### DIFF
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -3354,6 +3354,10 @@ export const UPDATE_FNS = {
         break
       }
       case 'Model': {
+        // Clear any cached paths since UIDs will have been regenerated and property paths may no longer exist
+        PP.clearPropertyPathCache()
+        TP.clearTemplatePathCache()
+
         // we use the new highlightBounds coming from the action
         code = existing.fileContents.code
         const highlightBounds = getHighlightBoundsFromParseResult(action.file.fileContents.parsed)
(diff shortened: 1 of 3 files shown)


seanparsons:
Doesn't this run the risk of the cache expanding infinitely?

balazsbajorics:
> Doesn't this run the risk of the cache expanding infinitely?

+1 to this.

when the code file is re-parsed, every uid is regenerated. 
1. that means that if you edit the code file a lot, we'll end up with a ton of unused template paths
2. it also means that the code file parsing is a good point to flush this cache

Rheeseyb:
> Doesn't this run the risk of the cache expanding infinitely?

Hmmm good point. We have the same issue with the property path cache too. Seems both should be cleared whenever we trigger a parse of the code. I'll see how dangerous that is...

Rheeseyb:
@balazsbajorics @seanparsons OK I have updated the flow to trigger a cache clear (for both the `TemplatePath` and `PropertyPath`) after each parse (as part of the handling of the `UPDATE_FROM_WORKER` action).

seanparsons (approved):

balazsbajorics (approved):

PR #649, opened 2020-11-02
vastly simplify dynamicPathToStaticPath

balazsbajorics:
**Problem:**
`MetadataUtils.dynamicPathToStaticPath` required the entire jsxMetadataKILLME and performed a fairly costly tree walk to extract the static path for a given dynamic path. This caused `getPropertyControlsForTarget` to take 3ms on the Majestic Broker test project. The problem was mitigated by us keeping the reference equality for jsxMetadataKILLME, but if any change was made on the canvas, this extra 3ms-5ms "penalty" would pile on top of the other necessary extra renders.

**Fix:**
Since the dynamic path element contains the static path element (for example, for the uid `aaa`, the dynamic path element is something like `aaa-1`), this PR uses a string splitting to get the static part of the path.

**Commit Details:**
- Use simple string splitting to get the static path
- `dynamicPathToStaticPath` no longer depends on jsxMetadataKILLME, which means I got to remove jsxMetadataKILLME from a lot of places that only required it to feed it to dynamicPathToStaticPath

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
@@ -79,13 +79,6 @@ export function renderCoreElement(
       if (originalIDForProps != null) {
         passthroughProps[UTOPIA_ORIGINAL_ID_KEY] = originalIDForProps
       }
-      // const originalParentIDForProps = Utils.defaultIfNull(
-      //   assembledProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY],
-      //   parentComponentInputProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY],
-      // )
-      // if (originalParentIDForProps != null) {
-      //   originalParentIDForProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY] = originalParentIDForProps
-      // }
 
       return renderJSXElement(
         TP.toString(templatePath),
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
@@ -34,6 +34,7 @@ import { objectMap } from '../../../core/shared/object-utils'
 import { cssValueOnlyContainsComments } from '../../../printer-parsers/css/css-parser-utils'
 import { filterDataProps } from '../../../utils/canvas-react-utils'
 import { buildSpyWrappedElement } from './ui-jsx-canvas-spy-wrapper'
+import { createIndexedUid } from '../../../core/shared/uid-utils'
 
 export function renderCoreElement(
   element: JSXElementChild,
@@ -78,13 +79,13 @@ export function renderCoreElement(
       if (originalIDForProps != null) {
         passthroughProps[UTOPIA_ORIGINAL_ID_KEY] = originalIDForProps
       }
-      const originalParentIDForProps = Utils.defaultIfNull(
-        assembledProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY],
-        parentComponentInputProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY],
-      )
-      if (originalParentIDForProps != null) {
-        originalParentIDForProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY] = originalParentIDForProps
-      }
+      // const originalParentIDForProps = Utils.defaultIfNull(
+      //   assembledProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY],
+      //   parentComponentInputProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY],
+      // )
+      // if (originalParentIDForProps != null) {
+      //   originalParentIDForProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY] = originalParentIDForProps
+      // }
 
       return renderJSXElement(
         TP.toString(templatePath),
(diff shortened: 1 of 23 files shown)


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -551,52 +550,15 @@ export const MetadataUtils = {
...
-        return eitherToMaybe(
-          mapEither((elems) => {
-            return TP.staticInstancePath(scenePath, elems)
-          }, pathElements),
-        )
-      } else {
-        return TP.asStatic(path)
-      }
-    }
+  dynamicPathToStaticPath(path: InstancePath): StaticInstancePath | null {
```
balazsbajorics, line 560: this is the root of the refactor, I removed the `metadata` parameter, and I agressively removed it from all the places where it was not used anymore

editor/src/core/shared/uid-utils.ts
```diff
@@ -33,6 +33,15 @@ export function generateUID(existingIDs: Array<string>): string {
   return generateUID(existingIDs)
 }
 
+export const GeneratedUIDSeparator = `-`
+export function createIndexedUid(originalUid: string, index: string | number): string {
```
balazsbajorics, line 37: to make things a little safer, I moved the creation of the dynamic uid to this dedicated function, and made an `extractOriginalUidFromIndexedUid` so if we change the format of the dynamic uid, we should be able to control this choke point here

balazsbajorics:
there's two main benefits of this PR:
1. when editing, the jsxMetadataKILLME cannot be referentially equal, so a lot of selectors regress in speed
2. I am working on an experimental branch where the jsxMetadataKILLME is never referentially stable, and dynamicPathToStaticPath was a third of the performance regressions, because a lot of selectors use it transitively

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -78,13 +79,13 @@ export function renderCoreElement(
...
         passthroughProps[UTOPIA_ORIGINAL_ID_KEY] = originalIDForProps
       }
-      const originalParentIDForProps = Utils.defaultIfNull(
-        assembledProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY],
-        parentComponentInputProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY],
-      )
-      if (originalParentIDForProps != null) {
-        originalParentIDForProps[UTOPIA_UID_ORIGINAL_PARENTS_KEY] = originalParentIDForProps
-      }
+      // const originalParentIDForProps = Utils.defaultIfNull(
```
enidemi, line 82: 👀

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -254,8 +254,8 @@ export const UiJsxCanvas = betterReactMemo(
     } = props
 
     // FIXME This is illegal! The two lines below are triggering a re-render
-    clearConsoleLogs()
-    proxyConsole(console, addToConsoleLogs)
+    // clearConsoleLogs()
```
enidemi, line 257: 👀

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1099,7 +1061,7 @@ export const MetadataUtils = {
   getJSXElementName(
     path: TemplatePath,
     components: Array<UtopiaJSXComponent>,
-    metadata: ComponentMetadata[],
+    metadata: ComponentMetadata[], // TODO BEFORE MERGE maybe we could remove metadata as a dependency from here
```
enidemi, line 1064: is this for this PR or the next?

enidemi (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
balazsbajorics: ouch! thanks for catching it

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
balazsbajorics: so I noticed that this is not used, but I wanted to ask if anyone knows if this is an accident or not. @Rheeseyb  or @seanparsons  maybe you remember something about this?

editor/src/core/model/element-metadata-utils.ts
balazsbajorics: for the next actually

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
seanparsons: It's still used here right? https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/dom-walker.ts#L323

seanparsons (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
balazsbajorics: ouch! so `const originalParentIDForProps` was not read anywhere in this function. I guess this means that this probably regressed sometime in the past and we didn't notice it?

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1099,7 +1061,7 @@ export const MetadataUtils = {
   getJSXElementName(
     path: TemplatePath,
     components: Array<UtopiaJSXComponent>,
-    metadata: ComponentMetadata[],
+    metadata: ComponentMetadata[], // TODO maybe we could remove metadata as a dependency from here if we change findSceneByTemplatePath
```
Rheeseyb, line 1064: Have you investigated whether or not these are possible?

editor/src/core/shared/uid-utils.ts
Rheeseyb: :+1:

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
seanparsons: Or it was written wrong in the first place.

editor/src/core/model/element-metadata-utils.ts
balazsbajorics: not yet, but I think it should be fine, just didn't want to bloat this PR

Rheeseyb (approved):

PR #650, opened 2020-11-02
Updated TS from 4.0.2 to 4.0.5

Rheeseyb:
Our current version of TS messes up auto import when existing imports have a trailing comma (which they always do). The TS bug ticket is https://github.com/microsoft/TypeScript/issues/40561

Updating to the latest version fixes that and breaks nothing, so win win :+1:

### DIFF
--- a/editor/package-lock.json
+++ b/editor/package-lock.json
@@ -26189,9 +26189,9 @@
       }
     },
     "typescript": {
-      "version": "4.0.2",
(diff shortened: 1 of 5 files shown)


enidemi (approved):

balazsbajorics (approved):
nice!!!! this was so annoying

seanparsons (approved):

PR #652, opened 2020-11-02
Assorted selector and other minor performance improvements

balazsbajorics:
This PR is the result of my investigation into skipping keepDeepReferenceEquality for the jsxMetadata. But it also was a good way to find which selectors are performing worst when you drag an element (dragging updates the metadata which means we loose reference equality to it even with keepDeepReferenceEquality... running). This PR fixes a few of the worst offenders.

**Commit Details:**
- Fixes the inspector settings screen
- Improves the inspector root selecting the entire jsxMetadata to later use small parts of it
- Removes the need to pass along the ElementInstance in the NavigatorItems
- Removes the use of immer produce from dispatch.tsx which incurred a 1.5-2ms extra overhead


editor/src/components/editor/store/dispatch.tsx
```diff
@@ -481,29 +481,29 @@ function editorDispatchInner(
...
+            jsxMetadataKILLME: reconstructJSXMetadata(result.editor),
+          },
         }
-        // TODO Re-enable this once we have addressed the root cause of the false positives
-        // (these were firing frequently even on elements that remained > 0x0 dimensions)
-        //
-        // const allLostElements = lostElements(r.editor.selectedViews, r.editor.jsxMetadataKILLME)
-        // const newLostElements = TP.filterPaths(allLostElements, r.editor.warnedInstances)
-        // if (newLostElements.length > 0 && isBrowserEnvironment) {
-        //   // FIXME The above `isBrowserEnvironment` check is required because this is tripped by tests that don't update the metadata
```
balazsbajorics, line 496: I can undelete this commented out code, but it felt like dead weight we are bringing along for the ride

editor/src/components/inspector/inspector.tsx
```diff
@@ -448,26 +448,36 @@ const DefaultStyleTargets: Array<CSSTarget> = [cssTarget(['style'], 0), cssTarge
...
+          TP.pathsEqual(m.scenePath, selectedViews[0]),
+        )
+        if (possibleRootComponent != null) {
+          return possibleRootComponent.rootElements.map((e) => e.templatePath)
+        } else {
+          return []
+        }
+      },
+      'InspectorEntryPoint',
+      (oldTemplatePaths, newTemplatePaths) => {
```
balazsbajorics, line 467: hmm, maybe I can ditch this now that @Rheeseyb  's TemplatePath pr is merged?

editor/src/components/editor/store/dispatch.tsx
Rheeseyb: I'm fine with this being binned off

editor/src/components/inspector/inspector.tsx
Rheeseyb: It would still create a new array, so probably worth including a variant of this, but it could be simplified to use reference equality for each element in the array.

editor/src/components/inspector/inspector.tsx
```diff
@@ -615,6 +625,7 @@ export const SingleInspectorEntryPoint: React.FunctionComponent<{
   // in order to prevent a big rerender of the inspector
 
   const inspectorModelMemoized = useKeepReferenceEqualityIfPossible(inspectorModel)
+  const targetsMemoized = useKeepReferenceEqualityIfPossible(targets)
```
Rheeseyb, line 628: We shouldn't be calling these `*Memoized`. Also it looks like that `FIXME` above is dead.

Rheeseyb (approved):

editor/src/components/inspector/inspector.tsx
balazsbajorics: ah, actually TP.pathsEqual starts by checking === equality

editor/src/components/inspector/inspector.tsx
balazsbajorics: renamed!

editor/src/components/editor/store/store-hook.ts
```diff
@@ -28,7 +28,7 @@ export const useEditorState = <U>(
   if (context == null) {
     throw new Error('useStore is missing from editor context')
   }
-  return context.useStore(wrappedSelector, equalityFn)
+  return context.useStore(wrappedSelector, equalityFn as EqualityChecker<U>)
```
seanparsons, line 31: This looks like a dubious lie, given that the second parameter of `EqualityChecker` is `unknown`

editor/src/components/inspector/inspector.tsx
seanparsons: We have the `arrayEquals` function, better to use that here.

seanparsons (approved):

enidemi (approved):

editor/src/components/editor/store/store-hook.ts
balazsbajorics: in the case of our useEditorState selectors, since the selectors are enforced by typescript to always return `U` it makes our lives easier. if you manually write an equality function, and the second type is `unknown`, that equality function you are writing probably for performance reasons will have to spend time and effort on verifying that the type is indeed `U` even though typescript already did that for us.

editor/src/components/inspector/inspector.tsx
balazsbajorics: thanks! I fixed it on master in 79594140

PR #653, opened 2020-11-02
Testing flattening all of the metadata

Rheeseyb:
I've created multiple structures to store the metadata separately rather than in a tree, and tested out the time taken to run `keepDeepReferenceEqualityIfPossible` on all of those vs the tree structure after merging the DOM and Spy reports.


Rheeseyb:
The overall result after fixing the issues with `keepDeepReferenceEqualityIfPossible` is a 2x improvement, which we believe can be achieved with a much simpler form of flattening, so I'm going to close this PR and do that.

PR #655, opened 2020-11-02
fix(Editor) Use rounding and support Maps to improve ref equality

Rheeseyb:
**Problem:**
DOM element measurements weren't rounding, so ref equality was often lost in the metadata due to floating point precision. Also, `Map`s were completely truncated when run through `keepDeepReferenceEqualityIfPossible`

**Fix:**
Round all DOM element measurements to the nearest 0.5, and extend `keepDeepReferenceEqualityIfPossible` to support `Map`s.
The rounding here has reduced the time to render each frame when scrolling in the Majestic Broker project by roughly 5-8ms!

### DIFF
--- a/editor/src/utils/react-performance.ts
+++ b/editor/src/utils/react-performance.ts
@@ -209,86 +209,6 @@ function failSafeMemoEqualityFunction(componentDisplayName: string, severity: 's
   }
 }
 
-function keepDeepReferenceEqualityOfObject(
-  oldObject: any,
-  possibleNewObject: any,
-  stackSizeInner: number,
-  valueStackSoFar: Array<any>,
-): any {
-  const keys = Object.keys(possibleNewObject)
-  const length = keys.length
-
-  let newObjectToReturn: any = {}
-  let canSaveOldObject = true
-
-  const oldKeys = Object.keys(oldObject)
-  if (length !== oldKeys.length) {
-    canSaveOldObject = false
-  }
-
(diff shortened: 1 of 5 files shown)


editor/src/utils/react-performance.ts
```diff
@@ -209,6 +209,86 @@ function failSafeMemoEqualityFunction(componentDisplayName: string, severity: 's
   }
 }
 
+function keepDeepReferenceEqualityOfObject(
```
balazsbajorics, line 212: https://github.com/epoberezkin/fast-deep-equal/ has been updated to support ES6 Maps. since this code here is taken from them, I think we should adopt the current version instead of deviating from it, AFAIK they are still the fastest deep equals out there

maltenuhn:
@Rheeseyb amazing find. I forgot what's the latest here in terms of browser support, but: do we need rounding to 0.5, or should we round to 1 here?

editor/src/utils/react-performance.ts
Rheeseyb: For the benefit of anyone else reading this - we discussed this on the phone and Balazs is going to pick this up. I have removed the change from this PR

Rheeseyb:
> @Rheeseyb amazing find. I forgot what's the latest here in terms of browser support, but: do we need rounding to 0.5, or should we round to 1 here?

I think keeping it at 0.5 is fine tbh, because this should have little affect elsewhere. It's purely to prevent the floating point imprecision.

PR #657, opened 2020-11-03
Improvement/modernizing keep deep reference equality

balazsbajorics:
**Problem:**
The `keepDeepReferenceEqualityIfPossible` utility function did not support ES6 maps.

**Fix:**
A new version of the original https://github.com/epoberezkin/fast-deep-equal/ have been released since then. I copied the new source code and adopted it the same way we did with the original `keepDeepReferenceEqualityIfPossible`

**Commit Details:**
- keepDeepReferenceEqualityIfPossible now supports ES6 Maps and Sets
- I changed the valueStackSoFar to be a Set to avoid having Array.prototpye.includes iterate over the entire array. I also mutate the valueStackSoFar to avoid having to clone the object for every recursion


editor/src/utils/react-performance.ts
```diff
@@ -209,141 +209,188 @@ function failSafeMemoEqualityFunction(componentDisplayName: string, severity: 's
   }
 }
 
+// this function has been adopted from https://github.com/epoberezkin/fast-deep-equal/tree/a33d49ab5cc659e331ff445109f35dd323230d41
```
seanparsons, line 212: Where in that code does it come from?

editor/src/utils/react-performance.ts
```diff
@@ -209,141 +209,188 @@ function failSafeMemoEqualityFunction(componentDisplayName: string, severity: 's
...
-    // for objects we do a deep recursion
-    var keys = keyList(possibleNewValueInner)
+    if (oldValue instanceof Map && possibleNewValue instanceof Map) {
+      let canSaveOldMap = true
+      let newMapToReturn: Map<any, any> = new Map()
+
+      if (oldValue.size !== possibleNewValue.size) {
+        canSaveOldMap = false
+      }
+      for (i of possibleNewValue.entries()) {
```
seanparsons, line 276: For the sake of clarity, given how often `i` is usually used for an index in a for loop.
```suggestion
      for (entry of possibleNewValue.entries()) {
```

editor/src/utils/react-performance.ts
```diff
@@ -209,141 +209,188 @@ function failSafeMemoEqualityFunction(componentDisplayName: string, severity: 's
...
+    }
+
+    if (oldValue instanceof Set && possibleNewValue instanceof Set) {
+      /**
+       * Sets use reference equality to determine if something is in the set or not,
+       * which makes salvaging sub-values very hard. We don't attempt to do that here
+       * */
+
+      if (oldValue.size !== possibleNewValue.size) return possibleNewValue
+      for (i of oldValue.entries()) if (!possibleNewValue.has(i[0])) return possibleNewValue
```
seanparsons, line 303: That's some pretty nasty inlining!
```suggestion
      for (entry of oldValue.entries()) {
        if (!possibleNewValue.has(entry[0])) {
          return possibleNewValue
        }
      }
```

editor/src/utils/react-performance.ts
```diff
@@ -209,141 +209,188 @@ function failSafeMemoEqualityFunction(componentDisplayName: string, severity: 's
...
-        return oldValueInner
+        return oldValue
       } else {
         return newArrayToReturn
       }
     }
 
-    if (arrA != arrB) return possibleNewValueInner
-
-    var dateA = oldValueInner instanceof Date,
```
seanparsons, line 289: Since this has gone, does that mean values of type `Date` are handled by the regular object handling now?

editor/src/utils/react-performance.ts
```diff
@@ -209,141 +209,188 @@ function failSafeMemoEqualityFunction(componentDisplayName: string, severity: 's
...
-  valueStackSoFar: Array<any>,
-): any {
-  // most of the actual structure is copy-pasted from https://github.com/epoberezkin/fast-deep-equal/blob/v2/index.js
-  // I've made it return oldValue or newValue instead of true and false
-  // and the recursion only stops if we get to return oldValue
-  // otherwise we still drill into non-equal objects to try and find
-  // keys that can be made equal
-
-  // If we're more than 100 frames deep, let's just call it a day
+  valueStackSoFar: Set<any>,
```
seanparsons, line 217: `Set<any>` scares me but also it's potentially not a stack of values if stored in a `Set` because it's not really designed to keep things in order.

seanparsons (approved):

editor/src/utils/react-performance.ts
balazsbajorics: there's only one source file, https://github.com/epoberezkin/fast-deep-equal/blob/a33d49ab5cc659e331ff445109f35dd323230d41/src/index.jst

should I link that instead?

editor/src/utils/react-performance.ts
balazsbajorics: I couldn't come up with a better name. I guess `valueSetSoFar` would be uhmm maybe better?

I changed it to Set because `Set.has` is much faster than `Array.includes` according to some googling, and it sounds commmon sense to me because Array.includes must iterate over the array

https://github.com/concrete-utopia/utopia/pull/657

editor/src/utils/react-performance.ts
seanparsons: That makes sense, even if it's got some crazy code generation magic behind it. :/

editor/src/utils/react-performance.ts
seanparsons: Surely if `Set` is just using `===` it must also be iterating over some internal array? Unless there's some internal magic optimisation that we're not privy to.

balazsbajorics: I believe it to be optimized, but I don't know the details.

https://dev.to/arnaud/using-array-prototype-includes-vs-set-prototype-has-to-filter-arrays-41fg

seanparsons: It would be worth confirming if that holds for things other than primitive values (and or those containers with arbitrary types in them). This has the whiff of `Set` using hashed buckets to hold the values which is super easy for numbers as you can just use the number as a hash.

balazsbajorics: good point, I'm checking it! I'll post a CSB here

balazsbajorics: https://codesandbox.io/s/arrayincludes-vs-sethas-lr63q?file=/src/index.ts

balazsbajorics: okay this is actually a wow result. with a very large array (100000 elements) the difference is 1000x!

```
Array.includes for 1000 elements took 3.1350000062957406/1000 ms 
Set.has for 1000 elements took 0.5050000036135316/1000 ms 
Array.includes for 10000 elements took 26.700000002165325/1000 ms 
Set.has for 10000 elements took 0.6600000051548705/1000 ms 
Array.includes for 100000 elements took 260.9450000018114/1000 ms 
Set.has for 100000 elements took 0.28500000189524144/1000 ms 
```

It seems like set has an O(1) lookup time. For use cases like this, when we want to maintain a short lived set of elements for filtering out stuff, Set's downsides (=== equality of elements, cannot serialize, etc) doesn't seem to be problematic at all. and the upside is _huge_.

editor/src/utils/react-performance.ts
Rheeseyb: +1 to this question

editor/src/utils/react-performance.ts
Rheeseyb: +1

Rheeseyb (approved):

editor/src/utils/react-performance.ts
balazsbajorics: I'll write a test case

balazsbajorics: I added two unit tests to test Dates. everything seems to work correctly!

editor/src/utils/react-performance.ts
balazsbajorics: fixed

editor/src/utils/react-performance.ts
balazsbajorics: fixed

balazsbajorics:
I believe I fixed all outsanding requests

Rheeseyb (approved):

PR #660, opened 2020-11-04
Perf/dom-walker add mutation and resize observer

enidemi:
This PR improves the performance of the dom-walker.

It adds the mutation and resize observers, the observer callbacks help tracking which element will need fresh metadata. In `walkElements` the `invalidatedTemplatePathsRef` are checked (with descendants) to return the store value or to recalculate the metadata.
For this `data-template-path` is added to the html elements.

### DIFF
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -359,6 +359,7 @@ export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElem
         validPaths: Array<string>,
       ) {
         if (
+          ObserversAvailable &&
           invalidatedSceneIDsRef.current?.length === 0 &&
           rootMetadataInStateRef.current.length > 0
         ) {
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -386,6 +386,7 @@ export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElem
         if (
           ObserversAvailable &&
           invalidatedSceneIDsRef.current?.length === 0 &&
+          rootMetadataInStateRef.current.length > 0 &&
           initCompleteRef.current === true
         ) {
           // no mutation happened on the entire canvas, just return the old metadata
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -127,17 +127,8 @@ function lazyValue<T>(getter: () => T) {
   }
 }
 
-export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElement> {
-  const containerRef = React.useRef<HTMLDivElement>(null)
-  const rootMetadataInStateRef = useRefEditorState((store) => store.editor.domMetadataKILLME)
-  const invalidatedSceneIDsRef = React.useRef<Array<string>>([])
-  const initCompleteRef = React.useRef<boolean>(false)
-  const selectedViews = useEditorState(
-    (store) => store.editor.selectedViews,
-    'useDomWalker selectedViews',
-  )
-
-  const resizeObserver = React.useMemo(() => {
+function useResizeObserver(invalidatedSceneIDsRef: React.MutableRefObject<Array<string>>) {
+  const resizeObserver: ResizeObserver = React.useMemo(() => {
     if (ObserversAvailable) {
       return new ResizeObserver((entries: any) => {
         for (let entry of entries) {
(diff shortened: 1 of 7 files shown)


editor/src/components/canvas/dom-walker.ts
```diff
@@ -94,17 +104,68 @@ function isScene(node: Node): node is HTMLElement {
...
+      }
+    })
+  }, [])
+
+  const mutationObserver = React.useMemo(() => {
+    return new MutationObserver((mutations: MutationRecord[]) => {
+      for (let mutation of mutations) {
+        if (mutation.attributeName === 'style') {
+          if (mutation.target instanceof HTMLElement) {
+            const pathAsString = getDOMAttribute(mutation.target, UTOPIA_TEMPLATE_PATH)
```
balazsbajorics, line 141: on my machine pathAsString is null for elements that have been resized on the canvas. possibly React deleting these mutations from the HTMLElement? 
@enidemi says she saw these template paths there, so maybe there's inconsistent behavior?
this is the key to the changes working.

right now this data-utopia-template-path is appended as a mutation inside the dom walker. however, if we could put them on during canvas render time, it would mean we don't need to worry about loosing these attributes!

@seanparsons I remember we talked about putting the entire template path on the elements, but you said you are dubious if we could ensure that those are correct. do you remember this?

seanparsons (approved):

editor/src/components/canvas/dom-walker.ts
```diff
@@ -94,24 +101,111 @@ function isScene(node: Node): node is HTMLElement {
...
+      return findParentScene(target.parentElement)
+    } else {
+      return null
+    }
+  }
+}
+
+const LogDomWalkerPerformance = !PRODUCTION_ENV && typeof window.performance.mark === 'function'
+
+function lazyValue<T>(getter: () => T) {
```
balazsbajorics, line 119: I made this new helper. if we like this, we can move it to Utils

editor/src/components/canvas/dom-walker.ts
```diff
@@ -94,24 +101,111 @@ function isScene(node: Node): node is HTMLElement {
...
+                invalidatedSceneIDsRef.current.push(sceneID)
+              }
+            }
+          }
+        }
+      })
+    } else {
+      return null
+    }
+  }, [invalidatedSceneIDsRef])
```
balazsbajorics, line 184: I think we should have these as `[]` empty arrays, and even leave a comment here saying that this should only evaluate once. because we are missing the teardown functions so if someone accidentally starts making multiple resize observers, we are in trouble!

balazsbajorics: btw AFAIK the useMemo lint rule will not be angry if you omit refs because it knows that refs are referentially stable, so including them doesn't matter

balazsbajorics (approved):

editor/src/components/canvas/dom-walker.ts
```diff
@@ -94,24 +100,136 @@ function isScene(node: Node): node is HTMLElement {
...
 
   React.useLayoutEffect(() => {
     if (containerRef.current != null) {
+      if (LogDomWalkerPerformance) {
+        performance.mark('DOM_WALKER_START')
+      }
       // Get some base values relating to the div this component creates.
       const refOfContainer = containerRef.current
+      if (ObserversAvailable) {
+        Array.from(document.querySelectorAll('#canvas-container *')).map((elem) => {
```
Rheeseyb, line 218: Please can you make `'canvas-container'` a `const` that is used here and on the actual canvas container

editor/src/components/canvas/dom-walker.ts
```diff
@@ -232,6 +341,27 @@ export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElem
...
+            } else {
+              const elementFromCurrentMetadata = MetadataUtils.findElementMetadata(
+                scenePath,
+                rootMetadataInStateRef.current,
+              )
+              if (elementFromCurrentMetadata != null) {
+                // early return for cached scenes
+                rootMetadata.push(...elementFromCurrentMetadata.children)
+                rootMetadata.push(elementFromCurrentMetadata)
+                return
```
Rheeseyb, line 362: Please can you restructure this code to avoid having this early return in the middle of the function

editor/src/components/canvas/dom-walker.ts
enidemi: fixed

editor/src/components/canvas/dom-walker.ts
enidemi: fixed

balazsbajorics (approved):

PR #662, opened 2020-11-05
Handle exports similar to imports in the parser.

seanparsons:
**Problem:**
Currently we don't handle exports in any way and exposing those to the storyboard creation would mean we could handle React components that aren't straight exposed with the `export` modifier.

**Fix:**
Have the parser process and handle exports similarly to exports conglomerating the various ways something can be exported.

**Limitations:**
This does not currently handle things like inline anonymous components that are exposed with `export default` and the `export default Thing` style which can expose an existing constant or variable. Thought it better to checkpoint this work here for now as it covers the overall flow.

**Commit Details:**
- Created `ExportsDetail` along with associated types to represent
  the detail of what is exported, attempting to pull together the
  individual representations of the exports.
- Added `exportsDetail` to `ParseSuccess`.
- Implemented `isDefaultExport`.
- Removed `TopLevelElementAndCodeContext` as it wasn't really used.
- Printer now applies modifiers to elements based on the export detail.
- Printer adds in an export expression as appropriate if values are
  exported that way.
- Parser now also processes export declarations and includes the values
  from those along with the modifiers from JSX components parsed.
- Changed some import sorting to use our comparison code instead of
  a hand rolled function.

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer-arbitrary-elements.spec.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer-arbitrary-elements.spec.ts
@@ -13,8 +13,14 @@ import {
   defaultPropsParam,
 } from '../../shared/element-template'
 import { setJSXValueAtPath } from '../../shared/jsx-attributes'
-import { foldEither, forEachRight, right } from '../../shared/either'
-import { foldParsedTextFile, ParseSuccess } from '../../shared/project-file-types'
+import { forEachRight } from '../../shared/either'
+import {
+  addModifierExportToDetail,
+  addNamedExportToDetail,
+  EmptyExportsDetail,
+  foldParsedTextFile,
+  ParseSuccess,
+} from '../../shared/project-file-types'
 import { parseSuccess } from '../common/project-file-utils'
 import { applyPrettier } from './prettier-utils'
 import {
@@ -200,6 +206,7 @@ export var whatever = props => (
       expect.objectContaining({}),
       null,
       null,
+      addModifierExportToDetail(EmptyExportsDetail, 'whatever'),
     )
     expect(actualResult).toEqual(expectedResult)
   })
@@ -291,6 +298,7 @@ return { arr: arr };`
       expect.objectContaining({}),
       null,
       null,
+      addModifierExportToDetail(EmptyExportsDetail, 'whatever'),
     )
     expect(actualResult).toEqual(expectedResult)
   })
@@ -384,6 +392,7 @@ return { arr: arr };`
       expect.objectContaining({}),
       null,
       null,
+      addModifierExportToDetail(EmptyExportsDetail, 'whatever'),
     )
     expect(actualResult).toEqual(expectedResult)
   })
@@ -478,6 +487,7 @@ return { arr: arr };`
       expect.objectContaining({}),
       null,
       null,
+      addModifierExportToDetail(EmptyExportsDetail, 'whatever'),
     )
     expect(actualResult).toEqual(expectedResult)
   })
@@ -554,6 +564,7 @@ export var whatever = (props) => {
       expect.objectContaining({}),
       null,
       null,
+      addModifierExportToDetail(EmptyExportsDetail, 'whatever'),
     )
     expect(actualResult).toEqual(expectedResult)
   })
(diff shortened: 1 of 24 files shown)


editor/src/core/workers/parser-printer/parser-printer-test-utils.ts
```diff
@@ -484,6 +493,36 @@ export function topLevelElementArbitrary(): Arbitrary<TopLevelElement> {
...
+    exportDetailModifierArbitrary(),
+  )
+}
+
+export function exportsDetailArbitrary(possibleNames: Array<string>): Arbitrary<ExportsDetail> {
+  if (possibleNames.length === 0) {
+    return FastCheck.constant(EmptyExportsDetail)
+  } else {
+    return FastCheck.tuple(
+      FastCheck.constant(null), //FastCheck.option(exportDetailNamedArbitrary(possibleNames)),
```
Rheeseyb, line 518: What's with the commented code here?

seanparsons: Ah yes, this is something I need to follow up on, it relates to `export default Thing = (props) => {...}` being invalid.

editor/src/core/workers/parser-printer/parser-printer.spec.ts
```diff
@@ -1907,44 +2010,14 @@ export var whatever = (props) => <View data-uid={'aaa'}>
     const exported = utopiaJSXComponent('whatever', true, defaultPropsParam, ['color'], view, null)
     const imports = addImport('cake', null, [importAlias('cake')], null, sampleImportsForTests)
     const expectedResult = clearParseResultUniqueIDs(
-      parseSuccess(imports, [exported], expect.objectContaining({}), null, null),
-    )
-    expect(actualResult).toEqual(expectedResult)
-  })
-  it('parses the code when it is a var', () => {
```
Rheeseyb, line 1914: What happened to this test?

website/package-lock.json
```diff
@@ -15496,9 +15496,9 @@
       "integrity": "sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c="
     },
     "typescript": {
-      "version": "4.0.2",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.0.2.tgz",
-      "integrity": "sha512-e4ERvRV2wb+rRZ/IQeb3jm2VxBsirQLpQhdxplZ2MEzGvDkkMmPglecnNDfSUBivMjP93vRbngYYDQqQ/78bcQ==",
+      "version": "4.0.5",
```
Rheeseyb, line 15499: Oops, I must have not committed this before :grimacing:

editor/src/core/workers/parser-printer/parser-printer.spec.ts
seanparsons: It's a duplicate:
https://github.com/concrete-utopia/utopia/blob/a9c7edc3ffe8b8a7e0b04d5a7dd6dea9d97ed1a9/editor/src/core/workers/parser-printer/parser-printer.spec.ts#L1914
https://github.com/concrete-utopia/utopia/blob/a9c7edc3ffe8b8a7e0b04d5a7dd6dea9d97ed1a9/editor/src/core/workers/parser-printer/parser-printer.spec.ts#L1951

Rheeseyb: Ha, seems like that was my fault when removing old dead code!

Rheeseyb (approved):

editor/src/core/workers/parser-printer/parser-printer.spec.ts
seanparsons: I totally went "Wait, I fixed this test..." when changing the test.

editor/src/core/workers/parser-printer/parser-printer-arbitrary-elements.spec.ts
```diff
@@ -200,6 +206,7 @@ export var whatever = props => (
       expect.objectContaining({}),
       null,
       null,
+      addModifierExportToDetail(EmptyExportsDetail, 'whatever'),
```
balazsbajorics, line 209: does this mean `whatever` is exported like so: `export var whatever = (props)...` ? or a separate line exporting it?

seanparsons: The former.

balazsbajorics (approved):

PR #663, opened 2020-11-05
never generate the same UID twice during a session

balazsbajorics:
**Problem:**
We keep hitting a duplicate uid error which stops the project from loading.

**Fix:**
Without taking the time to investigate the root cause of the problem, I hardened the uid generator. it now compares the generated UID to a global set of already-generated UIDs, meaning it will never generate the same uid during a session. it still respects the array of forbidden uids that can be passed in as parameter.

### DIFF
--- a/editor/src/core/shared/uid-utils.ts
+++ b/editor/src/core/shared/uid-utils.ts
@@ -20,12 +20,15 @@ import { objectMap } from './object-utils'
 
 export const UtopiaIDPropertyPath = PP.create(['data-uid'])
 
+const GeneratedUIDs: Set<string> = new Set()
+
 export function generateUID(existingIDs: Array<string>): string {
   const fullUid = UUID().replace(/\-/g, '')
   // trying to find a new 3 character substring from the full uid
   for (let i = 0; i < fullUid.length - 3; i++) {
     const id = fullUid.substring(i, i + 3)
-    if (!existingIDs.includes(id)) {
+    if (!existingIDs.includes(id) && !GeneratedUIDs.has(id)) {
+      GeneratedUIDs.add(id)
       return id
     }
   }


editor/src/core/shared/uid-utils.ts
```diff
@@ -20,12 +20,15 @@ import { objectMap } from './object-utils'
 
 export const UtopiaIDPropertyPath = PP.create(['data-uid'])
 
+const GeneratedUIDs: Set<string> = new Set()
```
seanparsons, line 23: How big will this conceivably get in a session?

seanparsons (approved):

enidemi (approved):

editor/src/core/shared/uid-utils.ts
balazsbajorics: number of elements times number of times the jsx file is parsed

seanparsons: "thanks"

Rheeseyb: Yep, right here we have a memory leak. If we're going to use this we need to clear the `Set` on each parse or at some other sensible time.

editor/src/core/shared/uid-utils.ts
```diff
@@ -20,12 +20,15 @@ import { objectMap } from './object-utils'
...
+
 export function generateUID(existingIDs: Array<string>): string {
   const fullUid = UUID().replace(/\-/g, '')
   // trying to find a new 3 character substring from the full uid
   for (let i = 0; i < fullUid.length - 3; i++) {
     const id = fullUid.substring(i, i + 3)
-    if (!existingIDs.includes(id)) {
+    if (!existingIDs.includes(id) && !GeneratedUIDs.has(id)) {
```
Rheeseyb, line 30: If we're using the `GeneratedUIDs` we don't need to use `existingIDs`

balazsbajorics: my assumption was that existingIDs might come from a loaded file, not generated in this session

editor/src/core/shared/uid-utils.ts
balazsbajorics: too bad Set doesn't support a "max size" option, in a FIFO kind of way

balazsbajorics: I talked to Sean and created a follow-up ticket: https://github.com/concrete-utopia/utopia/issues/668

in the meantime I'll merge this as-is to unblock us from testing the editor.

Rheeseyb: Sounds good to me :+1:

PR #664, opened 2020-11-05
performance mark for React renders

balazsbajorics:
**Problem:**
It's hard to see what React is up to when looking at performance measurements.

**Fix:**
I've realized that we control the react render wrappers. This means we can bring at least a bit of functionality back from the good old days of react. It's not a flame graph like the official React one was, rather you can see a flat stream of performance.marks as the various components' render functions are called.

image

But it helps a lot making sense of a larger frame, what is happening during various parts:
image

### DIFF
--- a/editor/src/utils/canvas-react-utils.ts
+++ b/editor/src/utils/canvas-react-utils.ts
@@ -6,6 +6,8 @@ import { MapLike } from 'typescript'
 import { firstLetterIsLowerCase } from '../core/shared/string-utils'
 import { isIntrinsicHTMLElementString } from '../core/shared/element-template'
 import { UtopiaKeys, UTOPIA_UID_KEY, UTOPIA_UID_PARENTS_KEY } from '../core/model/utopia-constants'
+import { v4 } from 'uuid'
+import { PRODUCTION_ENV } from '../common/env-vars'
 
 const realCreateElement = React.createElement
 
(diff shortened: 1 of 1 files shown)


seanparsons (approved):

editor/src/utils/canvas-react-utils.ts
```diff
@@ -143,11 +145,25 @@ function attachDataUidToRoot(
   }
 }
 
+const MeasureRenderTimes = !PRODUCTION_ENV && typeof window.performance.mark === 'function'
+
 const mangleFunctionType = Utils.memoize(
   (type: unknown): React.FunctionComponent => {
+    const uuid = v4()
```
Rheeseyb, line 152:
```suggestion
    const uuid = MeasureRenderTimes ? v4() : ''
```

editor/src/utils/canvas-react-utils.ts
```diff
@@ -163,11 +179,24 @@ const mangleFunctionType = Utils.memoize(
 
 const mangleClassType = Utils.memoize(
   (type: any) => {
+    const uuid = v4()
```
Rheeseyb, line 182:
```suggestion
    const uuid = MeasureRenderTimes ? v4() : ''
```

editor/src/utils/canvas-react-utils.ts
balazsbajorics: hmm, now you are saynig that, we should make a todo to measure the memoization of these mangled functions

enidemi (approved):

PR #665, opened 2020-11-05
Render the canvas first, then run the zustand selectors after Doom Walking finished

balazsbajorics:
this is how a scroll looks on master:
image

you can see how all the selectors are running before the dispatch. (highlighted in red)

if I remove the keepDeepReferenceEqualities, you can see that the zustand selectors run once before the SAVE_DOM_WALKER dispatch, and once more after that:
image

---------

This is what happens in the spike:
image
you can see that only the canvas selectors run before the dispatch, and then the rest of the selectors run after dispatch

this is what happens if I remove EVERY keepDeepReferenceEquality from the spy flow:
image

you can see that the same selectors run after dispatch, but we skipped spending 3-5ms on reference equalities!

**Note:**
However, right now the selectors do run for a longer time, so I need to verify if this finding is valuable.

----------

In any case, even if we can't entirely remove keepDeepReferenceEquality, if the user is editing something, the keepDeepReferenceEquality returns a new value because the measurements do change. So all these findings are very relevant to editing scenarios!


balazsbajorics:
did anyone have a chance to check this out on their computers? I think it's still a viable path forward to shave off a few milliseconds from drag/resize frame times

Rheeseyb:
Oh no not yet, I thought this was still WIP tbh. Will check it today

Rheeseyb:
Heads up - I have merged latest master into this.

Ok I've tested this locally a bit, dragging to resize in the Majestic Broker project, and found that with this change and without removing `keepDeepReferenceEquality`, the frame times when dragging to resize were now increased when compared to latest master (this was measured after merging latest master into this branch).

What I found very interesting however is that if we remove the 3 `keepDeepReferenceEquality` calls in the spy / dom metadata flows (2 when saving the report [here](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/editor/actions/actions.ts#L3622), and 1 when merging the data [here](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/editor/store/editor-state.ts#L1766)) the frame times return to being pretty much identical to current master. The only real difference I can see in the profile tab of the devtools is that there seems to be more CPU activity on background threads with this branch vs master (I have no idea what is causing that).

balazsbajorics:
> The only real difference I can see in the profile tab of the devtools is that there seems to be more CPU activity on background threads with this branch vs master (I have no idea what is causing that).

very interesting!

thanks for looking at it Rheese :) 

tomorrow I will look into why the scrolling frame times increase – I probably missed something about our update flow that works against this spike

PR #670, opened 2020-11-09
spike: a separate canvas iframe

balazsbajorics:
this spike repurposes the property controls iframe to render a Canvas.

at the current stage, the biggest issue seems to be that there's a delay between sending an update message from the editor until the iframe receives it.

probably I made a mistake by piggybacking on the (special-cased) `sendPropertyControlsInfoRequest` function which does some form of throttling.

I'll pick this up today and try to make a more responsive version!

so far the spike seems promising, I think a few more hours of trial and error and I will have an answer whether canvas in a separate iframe is a viable performance improvement.

-------

**Note:**
Even if this is not good enough for the canvas, this is definitely a viable direction to improve the Preview which now takes _seconds_ to respond to user actions.

### DIFF
--- a/editor/src/components/custom-code/code-file.spec.ts
+++ b/editor/src/components/custom-code/code-file.spec.ts
@@ -307,6 +307,7 @@ describe('Generating codeResultCache', () => {
       'incremental',
       null,
       false,
+      null,
     )
 
     expect(codeResultCache).toMatchSnapshot()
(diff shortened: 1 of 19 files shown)


balazsbajorics:
sadly, I think there's strong indication that we shouldn't spend more time on this for the Canvas (the preview will probably need a solution like this, if we keep it around)

image

the main hurdle seems to be that user events (like mouse move or wheel) have higher priority than interprocess communication. the messages are send via window.postMessage, and they only arrive at the iframe side in batches, very suspiciously precisely when there's a little break between mouse events

this means that this iframe lags behind the main canvas and I don't know about a way to remove this lag.

PR #671, opened 2020-11-09
Flattened Metadata

Rheeseyb:
**Problem:**
We were holding the element metadata in a deep tree-like structure, meaning plenty of tree walking and reference equality blowing.

**Fix:**
Replace that structure with a combination of an array for scene metadata and a simple keyed object for storing element metadata (keyed on the stringified `TemplatePath`). The vast majority of the changes are in `element-metadata-utils.ts`, with 90% of the remaining changes being a result of refactoring the type. The new types introduced are:
```
type ElementInstanceMetadataMap = { [key: string]: ElementInstanceMetadata }
interface JSXMetadata {
  components: Array<ComponentMetadata>
  elements: ElementInstanceMetadataMap
}
```
where `ComponentMetadata` and `ElementInstanceMetadata` now only contain the `TemplatePath`s of their children, rather than the children themselves.

On my machine this reduced the frame times by roughly 5-8ms when dragging to resize in the majestic broker project (averaging 18-22ms after vs 23-30ms before, which equates to about a 20-25% reduction)

**Key Files:**
The main files with model or logic changes to review here are:
- `element-template.tsx`
- `element-metadata-utils.ts`
- `editor-state.ts`
- `actions.ts`
- `dom-walker.ts`
All other files can be skimmed.

**Commits:**
- Introduced the above types in `element-template.tsx`. This also meant we could bin off `MetadataWithoutChildren` since that type is now identical to `ElementInstanceMetadata` (hence some changes from `childrenTemplatePaths` to `children`)
- Vastly simplified a lot of the functionality in `element-metadata-utils.ts` as part of having to rewrite it for supporting those types:
  - Some functions in `element-metadata-utils.ts` now only take either the `Array<ComponentMetadata>` (e.g. `findSceneByTemplatePath`) or the `ElementInstanceMetadataMap` as opposed to the full metadata (e.g. `getElementByInstancePathMaybe`) so that we can reduce our reliance on the full metadata. A lot more could be refactored to only take the relevant part of the metadata, but this refactoring already grew out of control.
  - Added a comment to the `getAllPaths` function because it was previously unclear that it was required to return the paths in a specific order (hence the awkward way it is now written, rather than using a simple `Object.values(elements).map()`)
  - Rolled `mergeComponentMetadata` into a single function since it is now somewhat less complex (somewhat)
  - When merging the metadata in the above I found we have a strange hack in place that requires the `element` of the `ElementInstanceMetadata` of a scene to remain a `left('Scene')`, which seemed to work by coincidence because of the way the previous merging worked. I don't fully understand why we need this - I think it came out of the Storyboards work? To work around this there is an explicit check now in place.
  - Some functions which were split across multiple top level functions (e.g. duplication) have been inlined using inner functions so they can have access to the full context (e.g. to save them creating multiple `ElementInstanceMetadataMap`s and merging them all together)
- Updated `dom-walker.ts` to support this change, including the "root" metadata we fetch from the cache when a scene hasn't changed (previously this relied on only the scene and the root child of a scene being pushed to `rootMetadata` since that would implicitly push the cached metadata for all descendants, whereas now it requires all descendants to be explicitly pushed)

**Note:** Although this PR looks to be *huge*, it includes a single snapshot change that is completely unwieldy and thousands of lines long, so the types changing produced a huge diff for that one file.


seanparsons (approved):

enidemi (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1202,176 +1145,254 @@ export const MetadataUtils = {
...
         },
       ]
+      return jsxMetadata(updatedScenes, metadata.elements)
     } else if (
       isRight(originalMetadata) &&
       TP.isInstancePath(oldPath) &&
       TP.isInstancePath(newPath)
     ) {
-      const duplicatedMetadata = this.duplicateElementMetadata(
+      const duplicatedElementPath = duplicateElementMetadata(
```
balazsbajorics, line 1303: this const doesn't seem to be used

balazsbajorics (approved):

editor/src/core/model/element-metadata-utils.ts
Rheeseyb: Oops good catch. I should strictly have been using this rather than `newPath` in the function call immediately below (the two end up being equal anyway, but I have fixed the code below to correctly use `duplicatedElementPath` in case we change anything internal here later).

PR #678, opened 2020-11-11
hidden button to re-parse project

balazsbajorics:
**Problem:**
It is cumbersome to run performance tests testing when the model is re-parsed because it involves editing text in monaco while recording, and some hangups here and there

**Fix:**
A secret button to re-parse the current code file! That blue 'R' button is what you are looking for

image

**Details:**
I made a nice new hook that takes the open code file and bumps the revision state so the parser starts working on it immediately.

**Note:**
image
you need to enable the feature switch 'Re-parse Project Button' and reload the editor to see the effect


Rheeseyb (commented):
Can you please hide this behind a feature switch (something about enabling dev tools), and also rename `project-file-test-utils.ts` to something that doesn't contain the word "test" (since we have implicitly gone with the pattern of reserving that word in file names for files that are only imported by tests or other test utils files)

balazsbajorics:
> Can you please hide this behind a feature switch (something about enabling dev tools), and also rename `project-file-test-utils.ts` to something that doesn't contain the word "test" (since we have implicitly gone with the pattern of reserving that word in file names for files that are only imported by tests or other test utils files)

yeah, good shouts

enidemi (approved):

Rheeseyb (approved):

PR #679, opened 2020-11-11
Experiment/components style warning

enidemi:
Component Experiment #672 

#### How to try it?
Open a project and you should see warnings in the navigator for components that doesn't support style prop. The warning has a tooltip too, but it's not interactive.
It is using the feature flag `Navigator Style Warning`.

Pizza: https://utopia.pizza/p/4f360db7/?branch_name=experiment/components-style-warning


maltenuhn:
**What I think this does** - shows me a warning if a component doesn't support the `style` prop.

**First reaction** - Showing this information helps. But showing it like this, doesn't: I got confused in at least _four_ different ways, and that's _after_ skimming the PR description. first, why don't these things have heigh/width (since we use the same style for the error). Second, what does the crossed-out `style` mean (there's no style _set_ on the thing?). Third (because we don't show the component icon) - are these components? And finally - if we don't support `style`, why do we show the inspector for it?

**Ideas for improvement** - Unsure if "improvement" is the right word here. The information hierarchy for this feels wrong, so in the current navigator I'd be inclined to not make it take the place of the icon, de-emphasize it in colour, maybe shrink it down. But overall, I think we need a slightly different approach to the navigator to not make this awkward. I'll add a couple of ideas to the doc.

Rheeseyb:
**Thoughts:** Having the crossed out "Style" and the tool tip are somewhat useful, but it's very misleading. I find it gives the impression that the style is broken somehow, but I have no way of knowing in what way it is broken (and the fact the styling section of the inspector shows up adds to that confusion). The major issue is that this doesn't inform me *why* this is an issue (specifically the ramifications it has for the editor), or how to fix it.

**Further work:** If we were to continue with this idea, we'd need to have a call to action for fixing the problem that is being flagged, and a way of providing more info as to why this is an issue. Tbh I'm now questioning whether the Navigator is even somewhere that this should be shown - it seems to me like the Inspector would be a better place to show warnings like this, especially since for most components you probably don't want to give them a style prop.

balazsbajorics:
I tried it as well. I agree witht the previously raised points: the alerts are communicating that something is broken. For a lot of third parties we will not be able to offer an actionable fix, so certainly we should think of this mode less as "broken" and more as "degraded capacity". I would try some visual cues that are pointing to that direction: maybe the navigator items themselves could be shown differently (sublty lighter color, one step towards disabled, but not as grayed out as disabled), or the selection color could be different (again, showing you that the editor has less capacity than usual)

alecmolloy:
So this is something that absolutely needs to be communicated, but this doesn't sound like an error, or even "wrong". components don't need to support a style prop. There should perhaps be a warning when a user tries to add a style prop via canvas or inspector action. Is that right?

i could still see an argument for highlighting this in the navigator, but this company does seem to try and throw all the metadata into the navigator and i'm not sure this is useful, nor easy to communicate. I'm struggling to think of a time when I need to understand this fact when I'm not operating directly on a component which doesn't accept a style prop.

PR #680, opened 2020-11-11
Experiment/components navigator props

enidemi:
Component Experiment #676

#### How to try it?
There is a component property list in the navigator showing style, children and other props.
It is using the feature flag `Navigator Component Props`.

Pizza: https://utopia.pizza/p/4f360db7/?branch_name=experiment/components-navigator-props

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
@@ -64,6 +64,7 @@ export interface NavigatorItemDragAndDropWrapperProps {
   renamingTarget: TemplatePath | null
   imports: Imports
   elementWarnings: ElementWarnings
+  properties: { [key: string]: string }
 }
 
 function canDrop(props: NavigatorItemDragAndDropWrapperProps, dropSource: TemplatePath): boolean {
@@ -269,6 +270,7 @@ export class NavigatorItemDndWrapper extends PureComponent<
           imports={this.props.imports}
           elementOriginType={this.props.elementOriginType}
           elementWarnings={this.props.elementWarnings}
+          properties={this.props.properties}
         />
         <NavigatorHintTop
           isOver={this.props.isOver}
--- a/editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
@@ -25,6 +25,7 @@ import { betterReactMemo } from 'uuiui-deps'
 import { getValueFromComplexMap } from '../../../utils/map'
 import { createSelector } from 'reselect'
 import { UtopiaKeys } from '../../../core/model/utopia-constants'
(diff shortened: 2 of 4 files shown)


maltenuhn:
**What I think this does** - It shows me all of the props of a component, _in the navigator_. Including things like style, and also seemingly the _count_ of children.

**First reactions** - I like the idea of it, because it makes me feel a bit more confident about what's going on. At least with the sample project, the props of things have pretty significant bearing on _what I actually see_ on the screen, so I don't mind the additional real estate here so much. In fact, the additional effective row height for components _itself_ is valuable to spot them without parsing the icons

**Ideas for improvement / iteration** The UX / design here needs some work :) More substantively, I'm not sure how valuable showing all of the the _values_ is. I suspect we'll probably want to special-case `style` and `css`, and only show values for text and JSX and (maybe) numbers? This one requires some discussion I think.

Rheeseyb:
**Thoughts:** I like the prospect but I think there's a problem here caused by this using the runtime props rather than the props defined by code. I'm not sure which of the two we should use tbh. I say it's a problem because it adds confusion when props are showing that aren't in the code - e.g. in https://utopia.pizza/p/89cff853-spice-chord/?branch_name=experiment/components-navigator-props you can see the `style {"position": "absolute"}` and `children 2`. Also, if a component is passed as a prop we just show the text "REACT_COMPONENT".

**Further work:** I think it's worth trying this same experiment and still showing the runtime values, but filtering props that aren't explicitly defined in the code (so e.g. the `children` and `style` props in that example shouldn't be shown since nothing was explicitly set for those).

enidemi:
> **Thoughts:** I like the prospect but I think there's a problem here caused by this using the runtime props rather than the props defined by code. I'm not sure which of the two we should use tbh. I say it's a problem because it adds confusion when props are showing that aren't in the code - e.g. in https://utopia.pizza/p/89cff853-spice-chord/?branch_name=experiment/components-navigator-props you can see the `style {"position": "absolute"}` and `children 2`. Also, if a component is passed as a prop we just show the text "REACT_COMPONENT".
> 
> **Further work:** I think it's worth trying this same experiment and still showing the runtime values, but filtering props that aren't explicitly defined in the code (so e.g. the `children` and `style` props in that example shouldn't be shown since nothing was explicitly set for those).

I realized after creating the PR that `css` prop is not in the runtime props, it's only available from the jsx element props. In a next version I would use only the jsx element props.

balazsbajorics:
I like this, I would also like to see the same for component roots. (now apparently I only see it for components that are child elements inside the tree) I'll check out the updated version

alecmolloy:
Wow! This is great. Some thoughts:

1. It needs some more ux experimentation, but fundamentally this is a good idea.
2. This enters the danger zone as technically _any_ prop that accepts a react child (`string | number | JSX.Element …`) should be visible here. Do we throw all string props in here? Do we put none of them? No good answer here really.
3. Curious what this means for code view "children" vs generated elements.

Issue #673, opened 2020-11-10
Basic (non-interactive) split pane Navigator for components


Rheeseyb:
Create a split-pane Navigator, with the regular Navigator at the top, and a new Navigator for the Component of a currently selected instance. Bottom Navigator is not interactive in anyway.

PR #682, opened 2020-11-11
experiment(Editor) Creating a basic split pane navigator experiment

Rheeseyb:
Experiment for #673 

This is a first pass of an experiment for a Navigator pane that shows info about the component of a selected instance. It's not at all interactive, so purely serves to show a representation of the component itself.

You can try the experiment here: https://utopia.pizza/p/b2796b7d-short-hole/?branch_name=experiment/split-pane-navigator
(or alternatively with this project: https://utopia.pizza/p/4f360db7/?branch_name=experiment/split-pane-navigator)

When selecting the `Card` instance you'll see the Component Navigator appear below the regular Navigator like so:
Screenshot_20201111_163601


maltenuhn:
**What I think this does** Sometimes adds an entry to the navigator with the currently open component
**First reaction** I don't get it. Things randomly appear there, and I don't have any sense of why that is, and then you click on something and they go away
**ideas for improvement** I suspect this will only make sense with two fixes: first making the navigator a split pane, so that the additional component always shows up in the same place (and if there isn't one, it's empty - it doesn't just disappear). Second, by giving it some kind of 'label' and segmented appearance (like a section in the inspector), so there's a sense of what this does.

enidemi:
I like that it shows the component structure inside the editor somewhere! It's useful because I can see that there is a layer of div wrapper around the props.children inside the component so it tells me I should look at that div if I want to change some style prop.
It looks too much like the navigator so I want to click on the elements there and also makes me wonder as a user why is this not part of the regular navigator?

Rheeseyb:
I strongly agree that this feels like it is too confusing to add any value at this point. A lot of that confusion seems to come from the styling here - it looks exactly like the existing Navigator so I expect it to behave exactly like it. Even with distinct styling I'd expect to be able to click rows to select things, and I'd be very confused about why it disappears.

To improve this I'd suggest:

1. Completely different styling
2. Always show the Component Navigator, using the nearest component instance ancestor if no instance is selected, so that in the first example code:
  - If I select the first `div` element it renders the Component Navigator with the `App` component
  - If I select the `Card` element it renders the Component Navigator with the `Card` component
  - If I select the `View` element that is a child of the `Card` element it renders the Component Navigator with the `Card` component
3. Selection of a row in the Component Navigator triggers the Isolation Mode from #699

balazsbajorics:
> 2. Always show the Component Navigator, using the nearest component instance ancestor if no instance is selected, so that in the first example code

I vote to try this first.

Actually, to take it further, we could make a variation of this experiment which lets you toggle the navigator (or show two views below each other) between DOM Outline Mode and Code Outline Mode. (dom outline mode being the vanilla navigator, and code outline mode being an altered version of this new navigator)

the Code Outline Mode could show all components in a file (with the selected one highlighted), in file appearance order.

alecmolloy:
I have to say I am confused by this one. I think the suggested full design helps a bit, but can I get an understanding of what this is solving?

Naïvely I wonder why I cannot click through to the component, where I certainly would want a navigator. What use does this information have to me otherwise?

enidemi:

PR #683, opened 2020-11-12
Window the navigator.

seanparsons:
**Problem:**
The navigator renders an entry for everything, even if only some small part of the navigator is visible. Which results in the render performance degrading linearly with the number of elements in the project.

**Fix:**
Used `react-window` to only render a part of the navigator elements that is visible.

**Notes:**
There is a weird kludge in `navigator.tsx` around how `AutoSizerComponent` is imported, this is some weird interaction between the two module systems it is bundled up in and the TypeScript types. One approach works for the browser and the other works in NodeJS.

**Commit Details:**
- Added `react-virtualized-auto-sizer` to the dependencies.
- `DerivedState` now includes `visibleNavigatorTargets` which is similar to
  `navigatorTargets` but does not include the children of collapsed
  elements to help with the windowing logic of the navigator.
- Fixed an odd issue in `useInspectorInfoForPropertyControl` where two
  arrays were indexed into without checking if values exist first.
- Removed `ancestorCollapsed` and added `windowStyle` to
  `NavigatorItemDragAndDropWrapperProps`.
- `NavigatorItemDndWrapper` now uses the `windowStyle` prop to position
  entries.
- Reworked `NavigatorComponent` to use `react-window` with the added
  auto-sizing component to limit the number of components that need to be
  rendered.
- Refactored `createOrderedTemplatePathsFromElements` to allocate less
  intermediate arrays and generate both `navigatorTargets` and
  `visibleNavigatorTargets`.
- Improved the performance of `extractOriginalUidFromIndexedUid` by
  a factor of 120 according to a micro benchmark.


editor/src/components/editor/actions/actions.ts
```diff
@@ -965,6 +965,7 @@ export function restoreDerivedState(history: StateHistory): DerivedState {
 
   return {
     navigatorTargets: poppedDerived.navigatorTargets,
+    visibleNavigatorTargets: poppedDerived.visibleNavigatorTargets,
```
balazsbajorics, line 968: I raised my eyebrow, and then I realized collapsedViews is in the undo history. eek!
 `collapsedViews: poppedEditor.navigator.collapsedViews` 

now that I noticed it, I _think_ we should change this to not be undoable, unless I am missing something?

editor/src/components/inspector/common/property-controls-hooks.ts
```diff
@@ -77,7 +77,10 @@ export function useInspectorInfoForPropertyControl(
 
   const parserFn = unwrapperAndParserForPropertyControl(control)
   const printerFn = printerForPropertyControl(control)
-  const parsedValue = eitherToMaybe(parserFn(rawValues[0], realValues[0])) // TODO We need a way to surface these errors to the users
+  let parsedValue: unknown = null
+  if (0 in rawValues && 0 in realValues) {
```
balazsbajorics, line 81: did you hit a new bug with this being unchecked? is this related to the navigator at all?

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -236,16 +236,15 @@ export class NavigatorItemDndWrapper extends PureComponent<
...
   }
 
-  render() {
+  render(): React.ReactElement {
     const props = this.props
 
     const navigatorItemContainer = (
       <div
         key='navigatorItem'
+        id={`navigator-item-${TP.toVarSafeComponentId(this.props.templatePath)}`}
```
balazsbajorics, line 245: does react-window rely on the `id`?

editor/src/components/editor/actions/actions.ts
seanparsons: I guess it's only an issue if the action triggers a checkpoint in the undo history right?

editor/src/components/inspector/common/property-controls-hooks.ts
seanparsons: Not related to the navigator, but it was blocking me from testing things.

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
seanparsons: Nope, this was just helpful in debugging.

editor/src/components/navigator/navigator.tsx
```diff
@@ -17,6 +17,14 @@ import { useEditorState, useRefEditorState } from '../editor/store/store-hook'
 import { ElementContextMenu } from '../element-context-menu'
 import { createDragSelections } from '../../templates/editor-navigator'
 import { betterReactMemo } from 'uuiui-deps'
+import { FixedSizeList, ListChildComponentProps } from 'react-window'
+import { Size } from 'react-virtualized-auto-sizer'
+// There's some weirdness between the types and the results in the two module systems.
+// This is to effectively massage the result so that if it is loaded in the browser or in
+// node it should end up with the right thing.
+const AutoSizer = require('react-virtualized-auto-sizer')
```
balazsbajorics, line 25: this is about the 4th time we encounter this problem. Last time I tried searching for a solution I didn't find anything, but that was a long time ago. It feels like we shouldn't be the only people who has this problem? surely other teams use `react-virtualized-auto-sizer` and also write tests, right? surely

editor/src/components/navigator/navigator.tsx
```diff
@@ -140,6 +133,41 @@ export const NavigatorComponent = betterReactMemo('NavigatorComponent', () => {
     dispatch([EditorActions.togglePanel('navigator')])
   }, [dispatch])
 
+  const Item = React.memo(({ index, style }: ListChildComponentProps) => {
```
balazsbajorics, line 136: please always use `betterReactMemo` instead of `React.memo`

editor/src/components/navigator/navigator.tsx
seanparsons: I guess if you make sure both sides are using the same working module system this becomes a non-problem.

editor/src/components/navigator/navigator.tsx
seanparsons: Good shout, fixed in edb21800.

editor/src/components/navigator/navigator.tsx
```diff
@@ -148,28 +176,18 @@ export const NavigatorComponent = betterReactMemo('NavigatorComponent', () => {
...
-              targetComponentKey={componentKey}
-              templatePath={targetPath}
-              getMaximumDistance={getDistanceFromAncestorWhereImTheLastLeaf}
-              getDragSelections={getDragSelections}
-              getSelectedViewsInRange={getSelectedViewsInRange}
-            />
-          )
-        })}
+        <div style={{ flex: '1 1 auto' }}>
+          <AutoSizerComponent disableWidth={true}>{ItemList}</AutoSizerComponent>
```
balazsbajorics, line 189: that's a weird kind of HOC! 

I wonder if we support this style on the canvas?

https://utopia.pizza/p/24153547-perpetual-action/

wow we do! :D

seanparsons: Yeah, they expect a certain parameter in their case rather than having an arity of 0.

balazsbajorics:
did you have a chance to run some performance measurements? I expect this to help with the zustand selectors too, because unmounted elements should unsubscribe their selectors, which should _significantly_ reduce our 120-or-so selectors, most of which are navigator items

seanparsons:
@balazsbajorics In dev mode the big block that used to eat ~400ms on my machine when editing code was down to ~230ms.

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -116,7 +116,7 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThanOrEqual(660) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(670)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThanOrEqual(570) // if this breaks, GREAT NEWS but update the test please :)
```
balazsbajorics, line 119: hmm, I presume this tests runs with a 0-high window, which means if we make a regression in the navigator render count, that will not be visible in a change here

editor/src/core/shared/uid-utils.ts
```diff
@@ -45,7 +45,12 @@ export function createIndexedUid(originalUid: string, index: string | number): s
 }
 
 export function extractOriginalUidFromIndexedUid(uid: string): string {
-  return uid.split(GeneratedUIDSeparator)[0]
+  const separatorIndex = uid.indexOf(GeneratedUIDSeparator)
```
balazsbajorics, line 48: a factor of 120! nice

balazsbajorics (approved):
I think we should make sure there's rendered navigator items in the render count tests, otherwise looks great, and _massive_ improvement! :)

PR #687, opened 2020-11-13
[WIP] Spike/code editor iframe

balazsbajorics:
instead of bulldozing over everything and breaking the build for days, I've opted to go with an iterative refactor approach. 

I moved the code editor to its own react rook (which means it doesn't have access to the zustand state hooks) and I removed every use of dispatch from inside the code editor files.

the props are passed down JSON stringified (to emulate message passing) and the dispatch is replaced with a window.message based channel.

the code editor is not in an actual iframe yet, the way I'm going with this refactor, that will be the last step.


balazsbajorics:
right now where I stopped, there is an interesting behavior: when editing the canvas, for a flash we can see an older parsed project rendered and then we see the new one rendered. I probably made an asynchronicity bug in the code editor, but that is going to be Monday Balazs's task to fix

balazsbajorics:
Monday update: the code editor is in an iframe!!!!!! it works!

...welllll, almost :)

image

I see a load of these fontfaceobserver `3000ms timeout exceeded` errors

also, when I type to change the code, it is still pretty buggy and laggy.

also, at this step I'm serializing _all_ of the monaco props and sending them down (including the entire project structureand the type definitions), which seems to take a staggering 17 ms!

image

but, baby steps, so far I did not encounter anything that would be a blocker, so I'm still very optimistic about this spike

alecmolloy:
sounds pretty rad!

balazsbajorics:
I've updated the PR. it's untestable unfortunately until it's merged. I put the actual iframe part behind a feature switch, so I think it's safe to merge

PR #688, opened 2020-11-13
Experiment/components navigator style css buttons

enidemi:
Component Experiment

#### How to try it?
Next version for https://github.com/concrete-utopia/utopia/pull/680
When opening a project with components you should see component properties in the navigator.
- It is using the jsx element props instead of the metadata props.
- The props are selectable, switching between css and style props changes the inspector target too.
- The props are deletable with the little X when selected.
- When a component doesn't support layouting style and css props are not visible.
- Children props are not interactive.

Pizza link: https://utopia.pizza/p/f8054bba-panoramic-condor/?branch_name=experiment/components-navigator-style-css-buttons


maltenuhn:
**What I think this does** Gives components more space in the navigator, and uses that space to show props
**First reaction** There's something in this concept that resonates with me,  but its current execution doesn't seem to help much
**Ideas for improvement** I'd be interested to see if any of these four enhancements make a difference: (1) showing props that contain JSX as trees in the navigator, (2) showing props that are set but _don't_ contain JSX on one row (but without their values), (3) showing the props in a visually distinct way, so they don't look "the same" as the text labels., and (4) making the props themselves somehow 'selectable' (even if nothing _happens_ when you select them)

enidemi:
I like that I can see the properties somewhere, but I think the properties should be in the inspector.

Rheeseyb:
Seeing the properties clearly in the navigator is definitely informative, but this pass feels wrong and I'm not 100% sure why. I don't think it's a million miles off, so maybe some tweaking makes sense. I'm thinking that maybe the Navigator should only show props that are components (e.g. the `icon` prop) since those relate to the _structure_ of the components, and that if we want to include all other props, they should be hidden until explicitly exposed by the user via some interaction (along the lines of the arrow to show / hide children). I agree that it probably makes most sense to only show all other properties in the Inspector.

balazsbajorics:
I think the values are less interesting here, the names of the props themselves contain more information to me (mostly that this selected entity is a Component, but also gives me some feeling of understanding about its shape.)

Putting the children-as-props into the tree would makes sense to me in a code-outline navigator. As things are now AFAIK, children-as-props do show up in the navigator if they are rendered into the DOM. Or am I mistaken?

alecmolloy:
So I think this gets at a very interesting idea, which is of a turbo-mode code outline. I get excited by this because I feel like this (and other) nav proposals are starting to hint at a compositional workflow where the code editor is able to be hidden. This is worthy of exploration but I feel we run the risk of overrunning the navigator with data, making scanning hard. Do we really have two things here? Should we give this turbo navigator legs and see how it works? Is it modals on top of the existing navigator? Not sure. But I would like to discuss the broader workflows that these are supposed to enable and see if we can come up with bigger themes.

PR #689, opened 2020-11-13
Made parsing produce consistent UID values.

seanparsons:
**Problem:**
Each time the code is parsed it will produce entirely new UIDs for the model which means everything will get re-rendered.

**Fix:**
Now the parse produces consistent UIDs based on a hash of what is parse and otherwise rotating the UID until an unused one is reached instead of randomly generating them.

**Commit Details:**
- Added `object-hash` dependency.
- Switched to `keepReferenceIfDeepEqualSLOW` for navigator targets.
- Implemented `clearAttributesSourceMaps` and associated functions
  to null out the `sourceMap` property.
- Implemented `generateConsistentUID` which attempts to generate
  the same UID value given the same starting point.
- `forciblyUpdateDataUID` now uses `generateConsistentUID` to have
  deterministic results for the UIDs in elements.
- `keepReferenceIfDeepEqualSLOW` uses `fast-deep-equal` instead of
  `R.equals`.


Rheeseyb:
Do you have a performance profile comparing these changes with the current master for things like scrolling, selection, dragging (ideally on the majestic broker project)? I ask mainly because of that new use of `keepReferenceIfDeepEqualSLOW`

editor/src/core/shared/uid-utils.ts
```diff
@@ -20,6 +20,67 @@ import { objectMap } from './object-utils'
...
+
+// Assumes possibleStartingValue consists only of characters that are valid to begin with.
+export function generateConsistentUID(
+  existingIDs: Set<string>,
+  possibleStartingValue: string,
+): string {
+  if (possibleStartingValue.length >= 3) {
+    const maxSteps = Math.floor(possibleStartingValue.length / 3)
+    for (let step = 0; step < maxSteps; step++) {
+      const possibleUID = possibleStartingValue.substring(step * 3, (step + 1) * 3)
```
Rheeseyb, line 60: So, thinking about it, we only went with the 3 character restriction on UIDs because they were previously included in the code. I wonder if it would make more sense to either remove this restriction entirely, or increase the allowed length to reduce the chance of collisions :thinking:

editor/src/components/editor/store/editor-state.ts
```diff
@@ -1259,11 +1259,11 @@ export function deriveState(
   )
 
   const derived: DerivedState = {
-    navigatorTargets: Utils.keepReferenceIfShallowEqual(
+    navigatorTargets: Utils.keepReferenceIfDeepEqualSLOW(
```
Rheeseyb, line 1262: Given the change to `Utils.keepReferenceIfDeepEqualSLOW` to now use `fast-deep-equal`, I think it probably makes sense to just switch these to use `keepDeepReferenceEqualityIfPossible`, since that is written based on `fast-deep-equal` anyway (assuming this switch was only because we were previously using `Utils.keepReferenceIfShallowEqual`)

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1392,11 +1400,16 @@ function buildHighlightBounds(
...
   props: JSXAttributes,
   existingHighlightBounds: Readonly<HighlightBoundsForUids>,
   alreadyExistingUIDs: Set<string>,
 ): WithParserMetadata<JSXAttributes> {
-  const uid = generateUID(alreadyExistingUIDs)
+  const hash = Hash({
+    name: name,
+    props: clearAttributesSourceMaps(clearAttributesUniqueIDs(props)),
+  })
+  const uid = generateConsistentUID(alreadyExistingUIDs, hash)
```
Rheeseyb, line 1412: Nice!

editor/src/core/shared/uid-utils.ts
```diff
@@ -20,6 +20,67 @@ import { objectMap } from './object-utils'
...
+  'u',
+  'v',
+  'w',
+  'x',
+  'y',
+  'z',
+]
+
+// Assumes possibleStartingValue consists only of characters that are valid to begin with.
+export function generateConsistentUID(
```
Rheeseyb, line 53: Should we completely replace `generateUID` with this now? Or is there a reason to keep the old one about?

editor/src/components/editor/store/editor-state.ts
seanparsons: So I might revert the change to the `SLOW` suffixed function and switch over to that.

editor/src/core/shared/uid-utils.ts
seanparsons: I think it still potentially makes sense to keep that for things like action changes, as that's got no consistency benefits/requirements. At least for now.

seanparsons:
@Rheeseyb Performance-wise, those things seem fine with the majestic broker project as far as I can see. Primarily this should only affect situations where the elements are regenerated and since `navigatorTargets` should only change to a shallow level the cost of doing the deep equals should be identical to before.

editor/src/components/editor/store/editor-state.ts
seanparsons: Changes made in e52217f4.

editor/src/core/shared/uid-utils.ts
seanparsons: It crossed my mind, but it seemed like a good idea to try and maintain the existing behaviour without changing too much and it makes it easier to eyeball things with shorter UIDs.

Rheeseyb (approved):

seanparsons:
Odd, switching the function for maintaining deep reference equality back has actually broken the fix for navigator performance...

Rheeseyb:
> Odd, switching the function for maintaining deep reference equality back has actually broken the fix for navigator performance...

Errrrr :grimacing: The only difference between what you had in the previous version and this is that the previous version would check if the entire objects deeply equalled, and return the previous object if they did, whereas using `keepDeepReferenceEqualityIfPossible` can potentially retain some of the original innards of the object without having to retain the entire object reference

seanparsons:
> Errrrr  The only difference between what you had in the previous version and this is that the previous version would check if the entire objects deeply equalled, and return the previous object if they did, whereas using `keepDeepReferenceEqualityIfPossible` can potentially retain some of the original innards of the object without having to retain the entire object reference

Which is confusing as you'd expect the former to be _more_ sensitive to changes than the latter. Since relates to an array of `TemplatePath` instances as well, it's even more baffling.

editor/src/core/shared/uid-utils.ts
balazsbajorics: I think increasing it to 4 or 5 characters would keep them debuggable, but make the collision space much larger

editor/src/components/editor/store/editor-state.ts
balazsbajorics: hmm, since the template paths are memoized, keepReferenceIfShallowEqual should be sufficient here. but I think there shouldn't be a significant difference because fast-deep-equals starts by doing a `===` check.

balazsbajorics:
@seanparsons did you consider running a keepDeepReferenceEqualityIfPossible on the ParseResult? how would that change the the big rerender after a new model is parsed?

seanparsons:
@balazsbajorics Things like source maps and the code itself will completely nerf any benefits of that.

balazsbajorics (approved):

PR #690, opened 2020-11-13
Experiment/text cursor focuses element

alecmolloy:
Completes the following two experiments: 
    1. Selected element follows cursor focus: as you move your text insertion cursor from node to node, the inspector targets the element, and navigator shows the selected element.
    2. Inspector edits mode: The inspector targets the element that has the text insertion cursor’s focus, and editing properties in the inspector edits that element’s style prop target (usually `props.style`, but whatever the inspector target is)

I've created another canvas mode: 'text', this gets rid of the canvas controls and just lets you click on the canvas to select text, text selection changes are monitored, and any change looks to see what element is selected, which then sets the selection for the editor, which lets you change properties for the element selected in the inspector. 

N.b. this was supposed to include basic contentEditable support, so that you could insert a text insertion cursor, but that just doesn't work. and there is so much bullshit going on with CSS that i didn't want to figure that out. There is a loooooot of keyboard shortcuts things that would need to be refactored to make text editing work properly.

Also: we need to get rid of all shotgun-style css selectors (i.e. `body`, `div`, or `*`). We can use cascading things for all the parts of the editor that are not the canvas, but we shouldn't be using any sort of `all`, because that seems to not work.


enidemi:
🍕 https://utopia.pizza/p/91f5a24e-alpine-pump/?branch_name=experiment/text-cursor-focuses-element

Use this button the active text editing mode (assets are unfortunately not included in the PR deploys)
Screenshot 2020-11-16 at 12 18 14

maltenuhn:
**What I think this does** in "text mode", changes the mouse pointer to the text caret when over text, and selects elements on click (not cmd-click), and updates selection on drag

**First reactions** It's quite interesting  - less so for the text editing, more so for the Chrome Devtools-like targeting mode. The ability to _select_ text (and the caret, maybe?) feels misleading (particularly since the navigator doesn't reflect multiselection, and because the "text selected on the canvas" presumably isn't the only thing that's selected.

**Ideas** There are three angles I'd explore further here: first, a 'targeting' or drawing-to-pick mode; this feels very satisfying text aside. I'd probably turn off the ability to actually _select_ text. Second, giving feedback on what exactly is selected here - in particular if there is > 1 node, or > 1 element selected. And third, combining this with the formulabar experiment to see if we can make the jump from 'selecting text' to editing text.

Rheeseyb:
I don't know if this is broken now, but I can't trigger text mode and the code editor doesn't render for `app.js` now :grimacing:

balazsbajorics:
somehow feels very fast to select elements this way. I really like that we use the browser text editor (something most computer users have an intuition for) to figure out what element to select when multiple elements are overlapping / containing each other. This behavior alone would be kind of hard to reproduce in a 100% intuitive way without reimplementing much of the core text rendering capabilities of the DOM / webkit.

But when I select something the selection outline is missing from the canvas – is this intentional? 
I miss some **strong** visual way that would tell me _how_ to edit the text. This can be contentEditable enabled with actual WYSIWYG, or a text editing box, or even a stronger animation inside Monaco that draws my attention to the selected text in the source code.

alecmolloy:
So some feedback for the feedback:
1. Contenteditable is broken in the editor. The intended experiment is "text cursor focuses element", but drag-selecting selection is really all you can do so that is somewhat misleading.
2. Override mode: because selection is the default experience, i might direct you to the experiment which first touches on the styled text overrides concept. https://github.com/concrete-utopia/utopia/pull/691. Selecting text shows the current style, and styling it in the inspector should add a span with the new styles.
3. Multiselect: as mentioned in the "text experiment" doc, selecting text across elements is weird, but when you look at it as an extension of how canvas multiselection works × text overriding works, it comes together nicely. If you have two adjacent span tags, each with a sentence, and you select the end text of the first span and the beginning text of the second span, the inspector should show the style props that match for each span, and "mixed" for those that don't. Styling in the inspector should add spans to each block of text in each element.

I think what is most interesting to discuss here is whether it feels right to click into a text block and have the inspector show the styles of that element. If there are other interesting things coming besides that, please do share them though.

PR #691, opened 2020-11-13
Experiment/wrap text in span action

alecmolloy:
This completes this text editing experiment:
    3. Override Mode: Select a portion of a text node, and the editor gets turned into an “override mode”. Any changes to the inspector, such as changing the font weight, or pressing CMD + B will add a span and apply the styling changes to that new span. Selecting all the text in an element will not enter override mode, but will change that element’s styles.

What it actually does is something even more basic: you can select text, and press the Bold icon in the right panel. That wraps whatever you have selected in a span with a style prop.


enidemi:
🍕 https://utopia.pizza/p/91f5a24e-alpine-pump/?branch_name=experiment/wrap-text-in-span-action

enidemi:
I tried using the bold shortcut with a partial text selection in Text Editing Mode, nothing happened. What am I doing wrong?

Rheeseyb:
I have tried this out too and I'm confused as to what is happening. I can see it creating new nodes but it doesn't seem to be creating them correctly (I don't think)? I'm not sure if this is because I can only see the changes in the Navigator though since the code editor appears to be broken in this branch.

balazsbajorics:
+1, I could see it wrapping the entire selected element in a span, but if I made a selection range, I couldn't make it create an inner span

alecmolloy:
so it looks like you all didn't see the bold icon in the right menu.
i mean, how did you miss it:

Screenshot 2020-11-20 at 4 44 10 PM

So the description is a bit confusing, because it first describes the experiment, and then the actual thing i made, which was much simpler. All this is is a custom bold button that should simulate what it would be like to make changes in the inspector. And it doesn't seem to load or show up as even being unloaded on the provided link. But i promise you there is a button there.

PR #694, opened 2020-11-16
experiment/component-canvas-highlights

enidemi:
Component experiment https://github.com/concrete-utopia/utopia/issues/693

#### How to try it?
Select Card/Photo/Comment/User components on the canvas/navigator, you should see all other components using the same component highlighted.
Selecting these components requires the cmd key, because they might not be siblings and the highlight is only visually there, but cannot catch the mousedown event.

🍕 https://utopia.pizza/p/4f360db7/?branch_name=experiment/component-canvas-highlights


Rheeseyb:
I've just tried this and quite honestly found it very confusing, but I think a lot of that is down to the styles chosen (my budget eyes can't see which element is selected and which is being highlighted as an instance of the component). TBH though, I suspect that even with the right styling here it would be too subtle, and that I would still find it confusing and not understand what the editor was trying to communicate to me without further (perhaps more explicit) visual indicators.

maltenuhn:
**What I think this does** Highlights other components based on the one I've got selected
**Reaction** This feels pretty confusing to me. I had to actually read the description to figure this out, playing with the editor left me confused. More importantly - I feel like it's not really helping me much. The main reason _I think_ isn't that knowing that these components are being used isn't helpful, but that _this information only really helps me if I'm already visualizing where they're being used on the canvas_.

There definitely have been instances where "show me all of the places where this component is being used" or even just the special case of "is this only being used in one place, right here - or also elsewhere?" are helpful, but to me that requires surfacing the info somewhere else - maybe the navigator, maybe the inspector, maybe the file browser.

But most importantly - I feel like this problem doesn't really arise until you've got large-enough projects.

enidemi:
This would be more helpful if it work differently only with instance showing where the original Component is to give information about if I want to edit the Component, but only if they fit on one page in the navigator. 
The same thing could be solved with a simple context menu.

Rheeseyb:
I've thought about this a bit more, and I think maybe a more subtle solution (changing the font styling of instances in the navigator when the component is selected, or vice versa) would be perhaps more suitable. Or alternatively holding a specific key would temporarily highlight instances of the component.

balazsbajorics:
image

one accidental thing I do like about this is that it highlights elements that are overflow hidden. I think _that_ is something worth thinking about a bit in the context of components, because there's lot of situations (like this very example) when a component potentially only shows a subset of its children

alecmolloy:
The most concrete I can get to a definition of the problem this is solving is this: "a bigger project helps you keep track where you use a component instance". This doesn't really feel like a problem I have when I selecting a component instance. There is some variation of showing all instances of a component that could be useful, but this feels too confusing to be of general use except for solving this very specific case.

PR #695, opened 2020-11-16
Added some custom deep reference keeping.

seanparsons:
**Problem:**
Deep equality checks to maintain references can be quite slow, so it seemed worth investigating if they can be done in a more typed way, without the introspection. When testing the fix it could be observed that two further things could do with fixing:
- The selected views were being cleared.
- Changing the focused panel was causing the navigator to re-render unnecessarily.

**Fix:**
Starting with just making a typed way to keep deep references for arrays of `TemplatePath` which eliminated them from the performance profile.

For the two further issues, the first just required a bunch of previous logic to be stripped out which either cleared or attempted to rebuild the selected views. In relation to the second, the `EditorState.focusedPanel` property was checked in a hook to prevent changes if no actual changes were made, but that means the hooks change each time the focused panel changes making the entire navigator re-render. That check was moved to the action handler instead.

**Commit Details:**
- Implemented some functions in `deep-equality.ts` as a typed
  alternative to our introspective deep reference keeping code
  that should be a lot faster.
- Implemented instances of `KeepDeepEqualityCall` to be used
  elsewhere for some of our types.
- `deriveState` now uses `TemplatePathArrayKeepDeepEquality` for
  `navigatorTargets` and `visibleNavigatorTargets`. Also no longer
  back-calculates the selected views heuristically against the
  highlight bounds.
- Moved handling of unnecessary focused panel changing to the
  `SET_FOCUS` action.
- `UPDATE_FROM_WORKER` action no longer clears the selected views.
- Removed checking of `EditorState.focusedPanel` from the navigator.


editor/src/utils/utils.ts
```diff
@@ -900,7 +900,7 @@ function defer<T>(): Promise<T> & {
   return promise as any
 }
 
-export function keepReferenceIfDeepEqualSLOW<T>(original: T, maybeNew: T, measure = false) {
+export function keepReferenceIfDeepEqualSLOW<T>(original: T, maybeNew: T, measure = false): T {
   if (measure) {
     performance.mark('before eq')
   }
```
balazsbajorics, line 906: the R.equals from here should be changed to fast-deep-equals

seanparsons: I did do this at one point in a previous, but reverted it as I didn't actually want the change connected to it.

editor/src/core/shared/deep-equality-instances.ts
```diff
@@ -0,0 +1,188 @@
...
+    oldDerivedState.elementWarnings,
+    newDerivedState.elementWarnings,
+  )
+  const elementWarnings = elementWarningsResult.value
+  areEqual = areEqual && elementWarningsResult.areEqual
+
+  if (areEqual) {
+    return keepDeepEqualityResult(oldDerivedState, true)
+  } else {
+    const derivedState: DerivedState = {
```
balazsbajorics, line 180: this is probably gross over engineering, but if the derived state changes, whoever will touch this might not see that they need to mutate `areEqual`. if  we'd have a combiner that takes `KeepDeepEqualityResult`s and map them into an object, it could automatically read the `KeepDeepEqualityResult.areEqual`s

seanparsons: Ideally I'd implement an applicative-like set of functions as a I did for `Either`. This is just a first swing at the problem.

balazsbajorics (approved):

editor/src/core/shared/deep-equality-instances.ts
```diff
@@ -0,0 +1,188 @@
...
+  areEqual = areEqual && propertyPathResult.areEqual
+
+  if (areEqual) {
+    return keepDeepEqualityResult(oldName, true)
+  } else {
+    const elementName: JSXElementName = {
+      baseVariable: baseVariable,
+      propertyPath: propertyPath,
+    }
+    return keepDeepEqualityResult(elementName, areEqual)
```
Rheeseyb, line 61: Might as well replace `areEqual` with `false` here for clarity

editor/src/core/shared/deep-equality-instances.ts
Rheeseyb: Hmmm I agree here. It might make sense in this case to move `DerivedStateKeepDeepEquality` next to the `deriveState` function to clear that up (and possibly likewise with the other specific deep equality functions)

editor/src/core/shared/deep-equality.ts
```diff
@@ -0,0 +1,98 @@
+import { keepDeepReferenceEqualityIfPossible } from '../../utils/react-performance'
```
Rheeseyb, line 1: Again this is importing from outside the `core/shared` folder. Please can you move this or the file it imports or _something_ to fix that

Rheeseyb:
I really like the work here but I have 2 concerns:

1. Like Balazs I'm a bit concerned that there's a risk someone will update the definition of an interface that has a corresponding `KeepDeepEqualityCall`, and not realise they have to update that too. I think moving the various `KeepDeepEqualityCall` implementations closer to the interfaces might help there, but then we lose the benefit of keeping them in a central place (which is that we know which interfaces have implementations of them and which don't). Maybe the answer is simply to ensure we have tests for each of them which explicitly handroll instances of the interfaces, ensuring those tests would break if the interface changes?
2. The back calculation of the selected views was required because making a code change could result in new UIDs (e.g. when adding an element), and we'd want to ensure the currently selected line still matches up with the selected element on the canvas - is that behaviour still the case here?

editor/src/core/shared/deep-equality-instances.ts
```diff
@@ -0,0 +1,188 @@
+import { HigherOrderControl } from '../../components/canvas/canvas-types'
+import {
+  DerivedState,
+  transientCanvasState,
+  TransientCanvasState,
+  TransientFileState,
+} from '../../components/editor/store/editor-state'
```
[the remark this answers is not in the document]
seanparsons: This should probably be moved instead, as it doesn't contain the general functions for these bits of deep equality logic.

editor/src/core/shared/deep-equality.ts
seanparsons: See previous comment.

seanparsons:
> 1. Like Balazs I'm a bit concerned that there's a risk someone will update the definition of an interface that has a corresponding `KeepDeepEqualityCall`, and not realise they have to update that too. I think moving the various `KeepDeepEqualityCall` implementations closer to the interfaces might help there, but then we lose the benefit of keeping them in a central place (which is that we know which interfaces have implementations of them and which don't). Maybe the answer is simply to ensure we have tests for each of them which explicitly handroll instances of the interfaces, ensuring those tests would break if the interface changes?

Yeah, this is the balance point between having something custom which is fast and something general which is slow. It's also a strong reason to not use default parameters for the factory functions of our types.

> 2. The back calculation of the selected views was required because making a code change could result in new UIDs (e.g. when adding an element), and we'd want to ensure the currently selected line still matches up with the selected element on the canvas - is that behaviour still the case here?

The consistent generation _should_ do most of that work, but something in this logic was causing the selected views to be lost even if the UIDs didn't change, so I deleted it given that what we have now should cover _most_ cases.

editor/src/core/shared/deep-equality-instances.ts
seanparsons: Sure.

editor/src/core/shared/deep-equality-instances.ts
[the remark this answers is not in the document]
seanparsons: Moved everything out of /core/shared and split them further in ff58d632.

Rheeseyb (approved):

editor/src/core/shared/deep-equality-instances.ts
seanparsons: Made this changes in f1b06eb0.

PR #697, opened 2020-11-17
Fix broken workflows

Rheeseyb:
**Problem:**
Workflows are now broken because of a command that has been disabled

**Fix:**
The command was being used by a now very outdated version of one of our actions, so I have updated the version used.


seanparsons (approved):

Rheeseyb:
Ok this looks to have fixed the issue here, so I'm going to merge straight away

PR #696, opened 2020-11-16
Experiment/component isolation mode

Rheeseyb:
Experiment for #686 

Double clicking an instance of a component **on the canvas** will switch to an "isolation" mode, which allows you to directly edit that component. Very rough around the edges, but this works by slipping a new scene into the transient state of the canvas.

Basic sample project: https://utopia.pizza/p/2d8b43ad-pitch-ankle/?branch_name=experiment/component-isolation-mode

PR #699, opened 2020-11-17
Experiment/Component Isolation Mode 2nd Pass

Rheeseyb:
Covers #701 - Further improvements to the previous component isolation mode experiment from #696 

The following changes have been made:
- Overlay displayed over everything else when in isolation mode
- Prevent selection of other elements via the canvas when in isolation mode
- Selection of other elements elsewhere (e.g. navigator) exits isolation mode
- `Isolated ${componentName}` label (there's a bug in clearing this, so after exiting isolation mode you might need to scroll for it to disappear)
- Dragging the root element or scene updates the instance that was isolated
- Support insert mode when component is isolated

Try it out at https://utopia.pizza/p/2d8b43ad-pitch-ankle/?branch_name=experiment/component-isolation-mode-pt2

### DIFF
--- a/editor/src/components/canvas/controls/component-area-control.tsx
+++ b/editor/src/components/canvas/controls/component-area-control.tsx
@@ -12,6 +12,7 @@ import * as TP from '../../../core/shared/template-path'
 import { ControlFontSize } from '../canvas-controls-frame'
 import { CanvasPositions } from '../canvas-types'
 import { calculateExtraSizeForZeroSizedElement } from './outline-utils'
+import { IsolatedComponent } from '../../editor/store/editor-state'
 
 interface ComponentAreaControlProps {
   target: TemplatePath
@@ -39,6 +40,7 @@ interface ComponentAreaControlProps {
   selectedViews: TemplatePath[]
   imports: Imports
   showAdditionalControls: boolean
+  isolatedComponent: IsolatedComponent | null
   testID?: string
 }
 
@@ -160,13 +162,23 @@ class ComponentAreaControlInner extends React.Component<ComponentAreaControlProp
     selectedViews: Array<TemplatePath>,
     event: React.MouseEvent<HTMLDivElement>,
   ): void {
+    // If operating on the isolated scene itself we should apply that change to the original target
+    let targets = selectedViews
+    let actualTarget = this.props.target
+    if (this.props.isolatedComponent != null) {
+      const { instance: isolatedTarget, scenePath: isolatedScene } = this.props.isolatedComponent
+      const willActOnIsolatedScene = (path: TemplatePath) =>
+        TP.pathsEqual(path, isolatedScene) || TP.isChildOf(path, isolatedScene)
+      if (selectedViews.length === 1 && willActOnIsolatedScene(selectedViews[0])) {
+        targets = [isolatedTarget]
+      }
+      if (willActOnIsolatedScene(this.props.target)) {
+        actualTarget = isolatedTarget
+      }
+    }
+
     const cursorPosition = this.props.windowToCanvasPosition(event.nativeEvent)
-    this.props.onMouseDown(
-      selectedViews,
-      this.props.target,
-      cursorPosition.canvasPositionRaw,
-      event,
-    )
+    this.props.onMouseDown(targets, actualTarget, cursorPosition.canvasPositionRaw, event)
   }
 
   getComponentAreaControl = (canShowInvisibleIndicator: boolean) => {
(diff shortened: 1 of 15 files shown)


maltenuhn:
**What I think this does** Crudely overlays a scene onto a scene, so that if you double-click into a component we can edit, it shows this.
**First reaction** Promising. Happy path, gesture, and escape hatch feel pretty aligned here
**Ideas** Two areas: first, the UX needs ~fine~ tuning to not be annoying. In particular: does this show up in the navigator as a separate scene, or "fold out" in place? how can you get out of isolation mode? How do we make it visually more clear what's going on (split-pane navigator focus?)

Much more importantly, we need to come up with ideas to solve the problem of "we have two trees to navigate now", like so:
```
const IconButton = <button><span><img ... /></span><span>{props.children}</span></button>

const MyComponent = <div>Hello world, <IconButton imgSrc={...}><span style={{...}>Hello how are you</span></IconButton>
```

if you keep double-clicking, we'd let you eventually select the span containing "Hello how are you". If you enter isolation mode, can you still get to this? Does selecting it pop you _back out of_ isolation mode? And what happens if "Hello how are you" is replaced with another set of nested components?

enidemi:
There are 2 things I like about this experiment:
1. it's very effortless to edit a component, compared to having a component directly added as a Scene, the doubleclick takes you inside instead of manually searching in the navigator to click on the Scene that you want to edit
2. I don't need to go through a current not so user friendly way of Scene creation for a Card Component if I want to edit it on the canvas

Rheeseyb:
So I've made it very clear elsewhere that I believe this experiment has a lot promise. However I definitely agree that there are a number of UX issues to tackle before it is truly valuable without being confusing. I think a lot of this can be achieved by combining this with the Component Navigator from #682 , with some tweaks in how that would work, so I'll propose some further experiments around that. Roughly, my thinking is this:

- The transient scene should not show up in the regular navigator at all
- When isolating a component, the transient scene should feed the bottom pane of the split pane navigator
- When not in Isolation mode, selecting something from the bottom navigator pane (the Component Navigator) enters Isolation mode with that element selected
- Both panes of the Navigator are fed using a combination of runtime data and the parsed code (with clear distinctions which data is coming from the runtime and which isn't)

I'll add separate experiments for each of these

balazsbajorics:
I only focused on the Canvas when trying this.

whoa this feels amazing! I think this takes out a lot of the resistance from using utopia "properly" with react components. I think without a feature like this, we are pushing people to create one megacomponent too much.

I'm not sure what and how the navigator should behave, but just the canvas experience is 👌

Rheeseyb:
> I'm not sure what and how the navigator should behave

I've proposed a few experiments around this in the Paper doc based on this experiment. Basically I think the best way for this to work is to combine it with the Split Pane navigator experiment, with the extension proposed around always showing the Component pane.

maltenuhn:
Does anyone have responses to the problem statement I posed [above](https://github.com/concrete-utopia/utopia/pull/699#issuecomment-730411780)? Or am I seeing an issue where there isn't one?

Regardless: for me this direction overall is the most promising, because it has the least amount of cognitive overhead - and I think in combination with the split pan navigator and perhaps a way to toggle between "I'm editing the component" and "I'm editing children passed into it" we can unlock a lot of utility here.

To progress this, I suggest y'all do what I just did and see what happens when with multiple levels of nesting, e.g. `<div><Component><div><Component2><div><Component3>` If we can work out a good behaviour for that scenario, I think we're onto something rather valuable here!

Rheeseyb:
> Does anyone have responses to the problem statement I posed above? Or am I seeing an issue where there isn't one?

Nope, there is definitely an issue there, but it's currently masked by a bug in that you can't add children to scenes. I believe with our current implementation of scenes you should be able to without too much trouble, so I might roll that into it's own experiment that branches off this.

Now assuming that we fix the above, IMO double clicking the "Hello how are you" span whilst in Isolation mode should pop you back out to the previous context (i.e. exit isolation mode and select the span), since that element is coming from the _instance_, and not the component. In this situation, with the second example (`<div><Component><div><Component2><div><Component3>`), you'd see the following behaviour:

1. Start by selecting the top most `div`, which keeps you out of isolation mode and selects that element in the `App` scene (assuming this is in a Scene rendering `App`)
2. Double clicking `Component` enters isolation mode for `Component`
3. Double clicking `Component2` exists isolation mode and selects that `Component2` instance in the `App` scene
4. Double clicking `Component2` again now enters isolation mode for `Component2`
5. Same behaviour for `Component3`

In these situations I'm very unsure how we should handle single click behaviour of children that came from the instance.

enidemi:

PR #700, opened 2020-11-17
WIP: ternary parsing

alecmolloy:
Currently is broken because of parser-printer-parsing.ts, specifically in the flatMapEither check that looks for a JSX_ARBITRARY_BLOCK, but in a TS context. Our JSX types are not possible here in a TS node

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer-parsing.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer-parsing.ts
@@ -1488,6 +1488,67 @@ function createJSXElementAllocatingUID(
   )
 }
 
+function produceConditionalFromExpression(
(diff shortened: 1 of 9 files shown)


editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1265,8 +1268,6 @@ function pullOutElementsToParse(nodes: Array<TS.Node>): Either<string, ElementsT
       TS.isJsxFragment(node)
     ) {
       result.push(node)
-    } else {
```
alecmolloy, line 1268: removed the else condition here because ternaries can have arbitrary nodes in here. this doesn't feel right.

PR #703, opened 2020-11-19
experiment/canvas-component-to-scene

enidemi:
Component Experiment #698 

#### How to try it?
Select a Component, open the right click menu. The first item should be `Extract Scene`. Clicking it will create a Scene using the current component and runtime properties from the metadata. The Scene positioning is not so smart, it's hardcoded to appear at (400, 0).
Please note that it doesn't work with component as props, like props.icon.

🍕 https://utopia.pizza/p/ddab49af-clumsy-license/?branch_name=experiment/canvas-component-to-scene

### DIFF
--- a/editor/src/components/context-menu-items.ts
+++ b/editor/src/components/context-menu-items.ts
@@ -11,6 +11,7 @@ import {
   deleteView,
   duplicateSelected,
   toggleHidden,
+  extractSceneFromComponent,
 } from './editor/actions/actions'
 import {
   toggleBackgroundLayers,
@@ -36,6 +37,15 @@ export interface CanvasData {
 
 export function requireDispatch(dispatch: EditorDispatch | null | undefined): EditorDispatch {
   return Utils.forceNotNull('Dispatch not supplied.', dispatch)
+}
+
+export const extractScene: ContextMenuItem<CanvasData> = {
+  name: 'Extract Scene',
+  enabled: true,
+  shortcut: '',
+  action: (data, dispatch?: EditorDispatch) => {
+    requireDispatch(dispatch)([extractSceneFromComponent(data.selectedViews)] as any, 'everyone')
+  },
 }
 
 export const duplicateElement: ContextMenuItem<CanvasData> = {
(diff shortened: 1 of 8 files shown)


maltenuhn:
@enidemi  I'm getting this after I click. Am I using it wrong? image

enidemi:
> @enidemi I'm getting this after I click. Am I using it wrong?

Try clicking and saving inside the code editor 🐛 I've seen this happen sometimes too

Rheeseyb:
I love being able to quickly create a scene based on an instance. However I'm very conflicted because I find that this flow instantly creates 2 expectations that I didn't previously expect:

1. I expect **all** changes to the created scene to also change the instance, meaning changes to the props provided. So in this case I expect to somehow update the text and that update both the Scene and the instance. That expectation is extremely confusing for me, because I'm not sure then how I would expect those changes to be applied - should they change the actual component or should there be some way of mapping changes onto _either_ the instance _or_ the component depending on what I have changed? I'm not sure if this expectation has come about as a result of the isolation mode experiment (the only difference there is it blurs the lines when changing the frame of the scene).
2. I also suddenly expect to be able to extract any selection into a scene, which I guess then means what I'm really expecting is to extract the entire thing into a new component. That then completely negates the original intention, which is to supply a way of editing the original component with a given context.

I think this then leads to a few possible paths:

- We provide an option to jump to an existing scene for the selected component, or create one (without those props) if one doesn't exist (or just always provide the latter option)
- We provide an option for extracting a new component from the current selection (something we had in the old editor)
- We do neither of the above because the functionality is fine as it is and only needs tweaking to the positioning of the scene

enidemi:
In general I like this one because it makes it easy to access and edit a component that is not added to the Canvas yet.
Improvement could be to make it more distinct that this creates a copy with the current props, but doesn't change the original one.

maltenuhn:
I like this idea, and think it'll make for a useful feature if we develop it towards that. However, I don't get the impression that it's particularly helpful towards the top end of the loop (you already need to know that something _is_ a component, etc), and I'm not really clear how to expose the potential issues here to a user - e.g. that we hardcode things, where we put them (which file), and how to change / update them etc.

alecmolloy:
i would echo basically what's been said already. i think this is to be considered alongside a bunch of extraction / autorefactor features, and i am not convinced this is a pressing thing for us to work on just now.

however this does get me excited about the idea of extracting a jsx element into its own component with a scene.

Issue #704, opened 2020-11-20
Always show the split pane navigator


Rheeseyb:
Extends #673

Proposed changes:

1. Always show both the regular and the component Navigators, using the closest component instance ancestor to feed the Component navigator
2. Change background colours of the title row to signify that this is a different section
3. Consider trying an alternative title for the Component Navigator

PR #705, opened 2020-11-20
Experiment: Always show the split pane navigator

Rheeseyb:
Experiment for #704

This is a second pass at the Split Pane Navigator that includes the following changes:

- The Component Navigator is always displayed, and uses the nearest component instance ancestor to populate its rows
- The Component Navigator's title now states what it is, rather than using the component's name as its title. The component's name is now the first row (This can be toggled using the feature switch `'Component Navigator Component Outline Title'`)
- I've used grey backgrounds to signify both that this is a separate Navigator, and that it isn't interactive (other than clicking the title to minimise it)

Test here: https://utopia.pizza/p/5f1b752b-abstract-snowman?branch_name=experiment/split-pane-always-present
Or with the Majestic Broker: https://utopia.pizza/p/6f386199-glowing-homburg/?branch_name=experiment/split-pane-always-present


maltenuhn:
For me this fixes a number of problems I had previously. I think the navigator stopped auto-scrolling to the selected element, but if delete the other scenes, I get this which works for me. Only two major frustrations are not seeing the name of the 'Card' component in the second navigator, and the navigator by default also showing the component of the main scene. But none of them change my feeling about this being a good direction to pursue!

image

balazsbajorics:
interesting, definitely an improvement! I miss not being able to select elements though

enidemi:
I like this because showing the nearest component instance gives context about what is being edited. Improvement idea: the `Component Outline` label can be scrolled out of view, I would keep that visible.

alecmolloy:
basically comment from the first PR stands, curious to hear about the thinking behind this one!

Rheeseyb:
> basically comment from the first PR stands, curious to hear about the thinking behind this one!

In a nutshell, this is about helping to more clearly communicate the underlying structure of the part of the application that is selected w.r.t. components vs instances, in order to answers the user's questions around why something is being rendered the way it is, and what part of that model they are currently manipulating.

enidemi:

PR #707, opened 2020-11-20
Prevent use of `AutosizeInput` in `react-select`.

seanparsons:
**Problem:**
`react-select` uses `react-input-autosize` internally which puts hooks in for when elements are mounted or updated which trigger reflows during the React render call chain, which needlessly puts strain on various updates like changing selection.

**Fix:**
`react-select` supports replacing the `Input` component, which has been done in all the places it is used.

**Commit Details:**
- Use a custom implementation of the `Input` component wherever
  `react-select` is used.
- Couple of tiny changes in `SelectModeControlContainer` which are
  mostly tiny fine tuning.


balazsbajorics:
how does this change the frame times? I have high hopes because AutoSizeInput crops up a lot of places...

seanparsons:
> how does this change the frame times? I have high hopes because AutoSizeInput crops up a lot of places...

Well it _appears_ to have chopped a bunch of time out of the mouse down/click event that changes selection, which consisted of up to 15ms.

balazsbajorics (approved):

PR #714, opened 2020-11-23
Trying to reduce the amount of work the DOM walker does.

seanparsons:
**Problem:**
The DOM walker has to walk the DOM of the canvas elements and as that's currently linear to the number of elements anything that can be shaved off is beneficial.

**Fix:**
Mostly this has attempted to avoid the costly calls into the browser DOM, by either caching or sharing values. As well as avoiding a regex that was run many times.

**Commit Details:**
- Created a cache that `globalFrameForElement` consults to reduce the number of calls to `getCanvasRectangleFromElement`.
- Moved around calls to `window.getComputedStyle` such that the results are shared with the result for a parent being passed down into a child.
- `getComputedStyle` (our function) now only collects a limited number of values and doesn't invoke `camelCaseToDashed` while doing it, which was particularly slow.
- More fully fleshed out the types of `forEachValue` and `clearTemplatePathCache`.


seanparsons:
Taking a different approach as limiting the properties so drastically is causing bugs.

PR #717, opened 2020-11-24
Experiment/components instances highlight

enidemi:
Component Experiment #706

#### How to try it?
Try highlighting and selecting component children on the canvas. Uses feature flag `Component Children Highlights`.
Editing is not included in this PR, the elements can be only selected.

To try it on a different project please add `?branch_name=experiment/components-instances-highlight` to the end of the URL. 

🍕 https://utopia.pizza/p/4c23662a-hissing-repair/?branch_name=experiment/components-instances-highlight

### DIFF
--- a/editor/src/components/canvas/controls/new-canvas-controls.tsx
+++ b/editor/src/components/canvas/controls/new-canvas-controls.tsx
@@ -360,9 +360,18 @@ const NewCanvasControlsClass = (props: NewCanvasControlsClassProps) => {
           if (frame == null) {
             return null
           }
-          const color = TP.isScenePath(path)
-            ? colorTheme.canvasSelectionSceneOutline.value
-            : colorTheme.canvasSelectionPrimaryOutline.value
+          let color: string
+          if (TP.isScenePath(path)) {
+            color = colorTheme.canvasSelectionSceneOutline.value
+          } else {
+            const element = MetadataUtils.getElementByInstancePathMaybe(
+              componentMetadata.elements,
+              path,
+            )
+            color = element?.internalChildOfComponent
+              ? '#FFA500'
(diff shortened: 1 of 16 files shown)


maltenuhn:
Ok, so I've just been playing with this and extended the sample project in a slightly evil manner - see [this link](https://utopia.pizza/p/7bd5867c-my-custom-experiment/?branch_name=experiment/components-instances-highlight).

**First reaction** is overwhelmingly positive - this addresses a huge part of the problem, _certainly_ on the canvas.

Beyond that, I struggled with these things the most

1. Even just bringing in one component makes the navigator rather unwieldy. With multiple, the navigator quickly becomes pretty an endless (and endlessly indented) list. In itself perhaps not _such_ a big deal (have you seen. my Sketch files?) but...
2. ... it makes it hard to understand when you're going from "inside the main component" to "inside elsewhere".
3. It's similarly hard to understand "where you are", especially if the component you're editing is rather small and well-structured.
4. I haven't yet worked out what the difference is between selecting a component vs the first rendered child of one but... it doesn't seem like a big deal?

However! I think the split-pane navigator could address a huge part of this, by making me "jump" between the top pane (main component I'm editing, i.e. the one in the scene) that only renders what's in the render function of `MainComponent` and bottom pane (secondary component I've clicked inside of, including _its_ render function outline). Empty otherwise. Focus of "selected element" moves from one panel to the other.

Alternatively, perhaps a more sophisticated "unfolding / folding" approach could help here, where we keep one navigator. If you select something from a secondary component directly on the canvas, we could _then_ show only the render tree of the component in question (styled differently in the navigator).

enidemi:
Merged into https://github.com/concrete-utopia/utopia/pull/727

PR #720, opened 2020-11-24
Use our deep equality handling for element metadata.

seanparsons:
**Problem:**
There's a lot of deep equality checks that happen every frame, which are currently handled by introspective code.

**Fix:**
Mostly attempt to reduce the introspection by building more manual deep equality checking.

**Results:**
This appears to be upper bounded to be as good as the existing introspective calls or in a best case scenario was twice as fast.

**Commit Details:**
- Created `KeepDeepEqualityCall` instances for everything up to `JSXMetadata`.
- Added some more utility functions for deep equality.
- Fixed a bug in `arrayDeepEquality`.
- Used the new deep equality functions when reconstructing the metadata
  and when the DOM report is pushed.
- Fixed some typing issues around `SceneProps` that were uncovered when
  running the deep equality checker.

### DIFF
--- a/editor/src/components/editor/store/store-deep-equality-instances.ts
+++ b/editor/src/components/editor/store/store-deep-equality-instances.ts
@@ -1,13 +1,85 @@
+import { Sides } from 'utopia-api'
+import {
+  ComponentMetadata,
+  ElementInstanceMetadata,
+  elementInstanceMetadata,
+  ElementInstanceMetadataMap,
+  ElementsWithin,
+  isArraySpread,
+  isArrayValue,
+  isJSXArbitraryBlock,
+  isJSXAttributeFunctionCall,
+  isJSXAttributeNestedArray,
+  isJSXAttributeNestedObject,
+  isJSXAttributeOtherJavaScript,
+  isJSXAttributeValue,
+  isJSXElement,
+  isJSXFragment,
+  isJSXTextBlock,
+  isPropertyAssignment,
+  isSpreadAssignment,
+  JSXArbitraryBlock,
+  JSXArrayElement,
+  jsxArraySpread,
+  JSXArraySpread,
+  JSXArrayValue,
+  jsxArrayValue,
+  JSXAttribute,
+  jsxAttributeFunctionCall,
+  JSXAttributeFunctionCall,
+  jsxAttributeNestedArray,
+  JSXAttributeNestedArray,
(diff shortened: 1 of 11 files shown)


balazsbajorics:
>  in a best case scenario

does it mean the performance varied, or you mean for a certain shape of jsxMetadata

seanparsons:
> does it mean the performance varied, or you mean for a certain shape of jsxMetadata

Yes and impossible to know without creating lots of test cases.

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -64,3 +136,508 @@ export function DerivedStateKeepDeepEquality(): KeepDeepEqualityCall<DerivedStat
...
+      layoutSystemForChildrenResult &&
+      providesBoundsForChildrenResult &&
+      positionResult &&
+      marginResult.areEqual &&
+      paddingResult.areEqual &&
+      naturalWidthResult &&
+      naturalHeightResult &&
+      clientWidthResult &&
+      clientHeightResult &&
+      parentFlexDirectionResult
```
balazsbajorics, line 518: `SpecialSizeMeasurements` is one of those objects which we change the shape pretty often (mostly to extend it, to be honest), it would be important to use some sort of type enforcement that makes sure we are not leaving this code behind. why is this not using the `combineNEqualityCalls` function?

seanparsons: Because I didn't want to make a `combine16EqualityCalls` function and most of the values in here are primitives so they don't need any particularly special handling.

balazsbajorics: I worry about friction making things bad here. how about writing a variadic combiner function? 

my problem is that if we add new parameters to the list of SpecialSizeMeasurements, not only the code will compile without letting us know that this needs to be updated, the chance that we catch a bug before the PR is merged is very low, because those new parameters need to be the only ones changing.

seanparsons: The code will only compile if a default parameter is used in `specialSizeMeasurements`, which I think is a worse anti-pattern that catches us out surprisingly often.

balazsbajorics (approved):
other than my sadness about type safety, we can merge this!

Rheeseyb (commented):
:+1:

Issue #719, opened 2020-11-24
Implement High Priority Selection State in the canvas controls


balazsbajorics:
#716 seems like a winner, perception-wise

PR #721, opened 2020-11-24
fix(DragState) only enter dragstate after mouse moved >2 px

balazsbajorics:
Part of #719

**Problem:**
When performing click-to-select on any element, we'd immediately dispatch an CREATE_DRAG_STATE on mouse down, and we'd follow up with a CLEAR_DRAG_STATE on mouse up. This caused two problems: it incurred some unnecessary spinning of the editor update functions, and it meant that we wouldn't show the resize rectangles on a selected element until we processed CLEAR_DRAG_STATE, because those are hidden during a drag.

This meant that perceptually click-to-drag could take 100-200 ms to take effect on the screen (an expensive 50-100ms action to change selected view + enter drag state, and a second expensive action to clear drag state). 

**Fix:**
`onControlMouseDown` does not immediately dispatch a CREATE_DRAG_STATE, instead it sets mouse move listeners which wait for a certain treshold to be crossed.

**Commit Details:**
- added mouse move and mouse up listeners inside `onControlMouseDown`
- only dispatch `CREATE_DRAG_STATE` if mouse is further than 2 pixels away from the mouse down point

### DIFF
--- a/editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
+++ b/editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
@@ -99,6 +99,24 @@ describe('moving a scene/rootview on the canvas', () => {
       const domFinished = renderResult.getDomReportDispatched()
       const dispatchDone = renderResult.getDispatchFollowUpactionsFinished()
       fireEvent(
+        areaControl,
+        new MouseEvent('mousemove', {
+          bubbles: true,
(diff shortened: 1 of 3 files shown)


seanparsons (approved):

enidemi (approved):

Rheeseyb (commented):
:+1:

PR #725, opened 2020-11-24
Performance/canvas selection updates canvas first

### LINKED ISSUE #719: Implement High Priority Selection State in the canvas controls

balazsbajorics:
Fixes #719

**Problem:**
The actions dispatched when selecting a View on the canvas take a long time. 

**Fix:**
We can fix the perceptual speed of the editor by first drawing the new selection controls _then_ dispatching the actions in the next frame.

**Commit Details:**
- renamed `NewCanvasControlsClass` to `NewCanvasControlsInner` because it's not even a class anymore
- added `usePropControlledStateV2` which does not render twice if the prop changes.
- added `localHighlightedViews` and `localSelectedViews` to `NewCanvasControlsInner` using `usePropControlledStateV2`
- added a new callback `setSelectedViewsLocally` which is called from `select-mode-control-container`
- `select-mode-control-container`'s `selectComponent` only calls the dispatch after _two_ animation frames are fired (the first one is the current one, the second one is the next one relative to the event.)

**Notes:**
This only works for mouse-based selection on the canvas. Keyboard-based selection and selection in the navigator is still slow as ever.


editor/src/components/canvas/controls/select-mode-control-container.tsx
```diff
@@ -104,13 +106,28 @@ export class SelectModeControlContainer extends React.Component<
...
       }
-      let selectActions = [
-        EditorActions.clearHighlightedViews(),
-        EditorActions.selectComponents(updatedSelection, false),
-        EditorActions.setLeftMenuTab(LeftMenuTab.UINavigate),
-        EditorActions.setRightMenuTab(RightMenuTab.Inspector),
-      ]
-      this.props.dispatch(selectActions, 'canvas')
+
+      ;(ReactDOM as any).flushSync(() => {
```
seanparsons, line 110: If this is secret functionality could you add a comment indicating what this is up to please?

seanparsons (approved):

enidemi (approved):

editor/src/components/canvas/controls/select-mode-control-container.tsx
balazsbajorics: good idea!

balazsbajorics: I removed this altogether instead of explaining it :)

Rheeseyb (commented):
:+1:

PR #728, opened 2020-11-25
Improve attribute manipulation performance.

seanparsons:
**Problem:**
Making changes or looking up values from the props is done quite a lot and any wasted time gets magnified a lot.

**Fix:**
Two main types of change, one was to avoid making the `object-path` library re-parse a property path. The other is to only use the `tail` function for property paths when it is needed, as property paths tend to be short but numerous and the tail of a single element property path will never be used for example.

**Commit Details:**
- Instead of producing a string and then calling the `object-path`
  library with that (which no doubt then has to parse that),
  we pass the internal property path array directly to it.
- Defer invoking the `tail` function for property paths until it's
  needed in various places.


balazsbajorics (approved):

Rheeseyb (commented):
:+1:

PR #729, opened 2020-11-25
Simplify traverseForPreparsedLayers.

seanparsons:
**Problem:**
`traverseForPreparsedLayers` kept showing up in profiles because it calls itself recursively potentially a dangerous number of times.

**Fix:**
Changed the function to loop instead of recursively invoking itself.

**Commit Details:**
- Change `traverseForPreparsedLayers` to loop rather than invoking
  itself recursively without shifting chars from one string to another.


balazsbajorics (approved):
did this code have test coverage?

seanparsons:
> did this code have test coverage?

It does indeed, at the usual filename.

Rheeseyb (commented):
:+1:

PR #732, opened 2020-11-26
Dependency cruiser

balazsbajorics:
image

**Problem:**
We have a lot of problems with the shape of the codebase. Circular dependencies, our different editor compile targets (editor, preview, monaco-editor, all the workers) accidentally pulling in each other's code, bloating the editor download size unchecked.

**Fix:**
Introducing dependency-cruiser, a tool which takes pretty readable configuration and "lints" the dependencies of code files!

**Commit Details:**
- Adding dependency-cruiser
- Renaming *-test-utils.ts files to *.test-utils.ts to make it very obvious that these are special files (not to be imported from production code!)
- Moving some dependencies from devDependencies to real dependencies as the editor was actually using them


editor/.dependency-cruiser.js
```diff
@@ -0,0 +1,371 @@
...
+    //   comment:
+    //     'This dependency is part of a circular relationship. You might want to revise ' +
+    //     'your solution (i.e. use dependency inversion, make sure the modules have a single responsibility) ',
+    //   from: {},
+    //   to: {
+    //     circular: true,
+    //   },
+    // },
+    // {
+    //   name: 'no-orphans',
```
balazsbajorics, line 16: no-orphans was problematic because I told dependency-cruiser to discard type-only ts imports. but that bugged out no-orphans as it pointed to some files which contained types only. bummer!

editor/.dependency-cruiser.js
```diff
@@ -0,0 +1,371 @@
+module.exports = {
+  forbidden: [
+    /* rules from the 'recommended' preset: */
+    // {
+    //   name: 'no-circular',
```
balazsbajorics, line 5: it would be _so good_ if we could enable this! however, right now it produces 253 errors :)

editor/.dependency-cruiser.js
```diff
@@ -0,0 +1,371 @@
...
+    {
+      name: 'not-to-unresolvable',
+      comment:
+        "This module depends on a module that cannot be found ('resolved to disk'). If it's an npm " +
+        'module: add it to your package.json. In all other cases you likely already know what to do.',
+      severity: 'error',
+      from: {},
+      to: {
+        couldNotResolve: true,
+        pathNot: 'domtoimage|monaco-editor',
```
balazsbajorics, line 102: I added these two as exceptions, domtoimage I understand because we load it in a funky way, monaco-editor was more surprising, but anyways, we know it _is_ resolved :)

enidemi (approved):

balazsbajorics:
I'm merging this as it's a pre-requirement for my other PR

editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
```diff
@@ -4,7 +4,7 @@ import {
   TestScenePath,
   makeTestProjectCodeWithSnippet,
   getTestParseSuccess,
-} from './ui-jsx-test-utils' // IMPORTANT - THIS IMPORT MUST ALWAYS COME FIRST
+} from './ui-jsx.test-utils' // IMPORTANT - THIS IMPORT MUST ALWAYS COME FIRST
```
seanparsons, line 7: Why is this renamed?

balazsbajorics: see the description

> Renaming *-test-utils.ts files to *.test-utils.ts to make it very obvious that these are special files (not to be imported from production code!)

balazsbajorics: I've made a dependency cruiser rule that fails if a non-spec file imports from a .test-utils file. But I didn't want to make the rule depend on `-test-utils`, I felt like that might be too unexpected. So I renamed these to .test-utils.ts to mark that something special is going on

editor/.dependency-cruiser.js
Rheeseyb: How many did this flag up? It might be worth adding exceptions for those files to the config

balazsbajorics: at least two files were false positives

Rheeseyb (commented):
:+1:

PR #733, opened 2020-11-26
Add storyboard file now uses parsed exports.

### LINKED ISSUE #629: For storyboard file, handle components which are exported but not marked directly as such.
This broadly comes down to the difference between

```export const App = ...```

and

```
const App = ...
export default App
```

or

```export default function App() {...```

seanparsons:
Fixes #629

**Problem:**
The add storyboard for file functionality looked at components directly marked as exported which appears to be a less commonly used approach in projects we have been testing the Github import with.

**Fix:**
Building on top of some recently added export parsing, the above functionality has been modified to use that to form the way the storyboard file imports from the component it picks.

**Commit Details:**
- `createSceneFromComponent` now uses a text component name.
- Reworked the type `ComponentToImport` to be a union type to
  represent the two cases that would be imported.
- Implemented `betterImportCandidate` as a heuristic way of
  determining what should be put into the storyboard.
- `addStoryboardFileToProject` now rewritten to work with the heuristic
  and the new types.
- `addStoryboardFileForComponent` more correctly does the right
  kind of imports and handles the new type of the import candidate.
- Added `detailsFromExportAssignment` to handle the `export default App`
  type of exports.
- Removed the slightly faulty way of handling named exports exported
  in an object named `default`.

### DIFF
--- a/editor/src/core/model/scene-utils.ts
+++ b/editor/src/core/model/scene-utils.ts
@@ -155,12 +155,12 @@ export function convertScenesToUtopiaCanvasComponent(
   )
 }
 
-export function createSceneFromComponent(component: UtopiaJSXComponent, uid: string): JSXElement {
+export function createSceneFromComponent(componentImportedAs: string, uid: string): JSXElement {
   const sceneProps = {
     component: jsxAttributeOtherJavaScript(
-      component.name,
-      `return ${component.name}`,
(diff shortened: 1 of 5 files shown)


balazsbajorics (approved):

Rheeseyb (commented):
:+1:

PR #734, opened 2020-11-26
Experiment/code outline ternaries

enidemi:
Component Experiment #730

#### How to try it?
Open a project with ternaries inside components. Look at the the Code Outline navigator on the left side. It's not selectable. 
Feature flag: `Code Outline Navigator`

TODO before creating a link:
- [ ] showing which condition is currently active
- [ ] selection? highlights?

🍕 https://utopia.pizza/p/8b108d72-probable-mandible/?branch_name=experiment/code-outline-ternaries


maltenuhn:
**First impression** I like this, even in its bare bones minimalism here. I think it comes down to feeling substantially more "in control" over code. Try it - type a ternary, _hide the code editor_, and see what you think!

Not leaving any comments beyond that because of the bare-bones nature of the implementation, just to call out that I tried
- adding nested ternaries
- adding ternaries with jsx and then more ternaries
- ternaries inside of mapped contents

and immediately missed the tree structure here that would have been required for success :)

PR #735, opened 2020-11-26
Fix/monaco do not pull editor

balazsbajorics:
**Problem:**
The monaco-editor-iframe was linking the entire editor and thus it needed to pull in 30 megabytes to load.

**Fix:**
Using dependency cruiser I banned the (transitive) linking of editor-state.ts and editor.tsx for any dependency path starting from `monaco-editor-iframe.tsx` (the entry point)

**Commit Details:**
- split `code-editor-container` which contained both the utopia side and the iframe side into `code-editor-container` (the utopia editor side) and `code-editor-iframe-entry-point` (the iframe side) into two files
- I duplicated a couple of action creators. I will make a follow up PR that moves all action creators to their own file, so that importing an action creator does not pull in the entire editor. the diff is massive tho, so I don't want to do it here.
- some react deepReferenceEquality hooks were living in `property-path-hooks`, which means they involved massive transitive dependencies. I moved them to react-performance.
- pulling in `uuiui/index.ts` or `uuiui-deps/index.ts` leads to the editor, so in the critical code paths I've changed the imports to be more precise, and point to certain UI files. the long term solution is to create a dependency-cruiser rule that prevents uuiui from reading files that are not in the uuiui folder. that will involve killing uuiui-deps (since the point of uuiui-deps was literally to pull in dependencies from the rest of the editor) so it's a larger undertaking.


balazsbajorics:
btw the number of components involved in rendering the code editor are way out of hand:
this is because the way I refactored `ScriptEditor` into an iframe involved wrapping it in a component which translated its props, which eventually became the iframe entry point, but I needed a component that loads the iframe, right now this is the chain of components that eventually render Monaco:

`CodeEditorWrapper` -> `CodeEditorIframeContainer` ----> iframe ----> `CodeEditorEntryPoint` -> `ScriptEditor` -> `CodeEditor` -> `MonacoWrapper`  

the names are all horrible! please help!

editor/.dependency-cruiser.js
```diff
@@ -32,6 +32,19 @@ module.exports = {
...
+    {
+      name: 'not-from-monaco-to-editor',
+      comment:
+        'The code editor files should not accidentally import the main editor, as that pulls in 10+ megabytes of dependencies',
+      severity: 'error',
+      from: {
+        path: 'monaco-editor-iframe\\.tsx$',
+      },
+      to: {
+        path: ['src/templates/editor\\.tsx', 'src/components/editor/store/editor-state\\.ts'],
```
Rheeseyb, line 44: Probably worth extending the `'src/templates/editor\\.tsx'` to become everything except `src/templates/monaco-editor-iframe.tsx`

editor/src/components/code-editor/code-problems.tsx
```diff
@@ -4,20 +4,26 @@ import { Console } from 'console-feed'
 import * as R from 'ramda'
 import { Resizable, ResizeCallback } from 're-resizable'
 import * as React from 'react'
-import { colorTheme, FlexRow, Icons, TabComponent, UtopiaTheme, SimpleFlexColumn } from 'uuiui'
-import { betterReactMemo } from 'uuiui-deps'
+// import { colorTheme, FlexRow, Icons, TabComponent, UtopiaTheme, SimpleFlexColumn } from 'uuiui'
+// import { betterReactMemo } from 'uuiui-deps'
```
Rheeseyb, line 8: Delete me

Rheeseyb (approved):

editor/.dependency-cruiser.js
balazsbajorics: you mean everyting inside /templates?

Rheeseyb: Ha yeah!

seanparsons (approved):

PR #739, opened 2020-11-27
move action creators to separate file

balazsbajorics:
**Problem:**
All of our action creators live in actions.ts. importing actions.ts brings most of the editor with it.

**Fix:**
Move the action creators to a dedicated file. Almost all imports are `import type`s


editor/src/components/editor/actions/action-creators.ts
```diff
@@ -0,0 +1,1208 @@
+import { LayoutSystem } from 'utopia-api' // TODO fixme this imports utopia-api
```
seanparsons, line 1: Why is this an issue?

balazsbajorics: since this file imports almost nothing else, I felt bad for having to pull in utopia-api (and thus emotion, and other such stuff we are using there). probably not a massive problem, but since we only use it to get an Enum (which are soft deprecated anyways), it wouldn't be hard to switch that type to a string literal union, and then we wouldn't need to import it, we could just `import type` it

seanparsons (approved):

Rheeseyb (approved):
I know Sean said something to the same effect, but I really wish the diffs would reflect that code has moved from file A to file B rather than having to trust that there have been no actual changes

balazsbajorics:
> I know Sean said something to the same effect, but I really wish the diffs would reflect that code has moved from file A to file B rather than having to trust that there have been no actual changes

yeah, it's not too fun to read this PR, that's why I made it a separate PR from the monaco-editor one.

Rheeseyb:
> that's why I made it a separate PR from the monaco-editor one.
Yep, fully appreciate that :)

PR #740, opened 2020-11-27
Component isolation mode 3rd pass

Rheeseyb:
**What this adds**:

- Scenes pass children to the components they render
- Component Isolation mode now correctly renders children and props that are components
- Children can now be selected in scenes
- Isolation mode supports drilling deeper (e.g. isolating `ComponentA` which renders `ComponentB` would now support double clicking `ComponentB` to isolate that)

**What this doesn't include**:

The intented behaviour was that double clicking a child passed in from the parent when isolating a component would pop back out of isolation mode and select the child in the original context. However, I'm cutting this experiment now because even hacking that together became extremely complicated based on the current hacks to get to this stage, mainly because of some issues around selection of children in general, how we determine the target of a selection, nested isolations would then require maintaining a stack of previously isolated scenes, and metadata would be lost unless we render all isolated scenes in the stack. I'm fully confident we could solve all of these issues with a full implementation of the feature, but for an experiment it feels like we've reached the point of diminishing returns here.

:pizza: https://utopia.pizza/p/119ab8a6-auspicious-bergamot/?branch_name=experiment/component-isolation-mode-pt3
or (Majestic broker, which isn't really a great project for testing this feature): https://utopia.pizza/p/6f386199-glowing-homburg/?branch_name=experiment/component-isolation-mode-pt3

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -1557,66 +1557,51 @@ function createTransientSceneForSelectedComponent(
   if (isolatedComponent == null) {
     return null
   } else {
-    const metadata = editorState.canvas.dragState?.metadata ?? editorState.jsxMetadataKILLME
-    const instanceMetadata = MetadataUtils.getElementByInstancePathMaybe(
-      metadata.elements,
-      isolatedComponent.instance,
+    const { componentName, frame, element, scenePath } = isolatedComponent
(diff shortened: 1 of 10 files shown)


Rheeseyb:
My main issue with this now is that isolating a scene that renders children or props is pretty tricky, since those elements will capture the mouse events. It definitely requires different styling on the highlight & selection controls, and probably another way of entering isolation mode (perhaps via the Navigator).

PR #741, opened 2020-11-30
fix(Dom walker) re-run dom-walker if the canvas mount count increases

### LINKED ISSUE #718: When the code is changed by the user, the canvas is rerendered, but the jsxMetadata seems like stays old, at least some stray highlights linger on the screen
https://paradise-engineering.slack.com/files/U0EVD7VFX/F01FK15CMRS/the_mystery_of_the_missing_code.mp4

video from 0:24

(MN: screenshot of similar issue:

image

balazsbajorics:
Fixes #718

**Problem:**
Changing the code did not immediately re-run the DOM walker, leaving traces of old highlights behind.

**Fix:**
The DOM walker now obeys the canvas mount counter, which we use to forcibly update the canvas in situations like a code update.

**Commit Details:**
- Added mount count to dom walker props
- replaced the `initCompleteRef` with a hook that resets initComplete to false if the mount count increases

### DIFF
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -208,11 +208,28 @@ function useInvalidateScenesWhenSelectedViewChanges(
   )
 }
 
+function useInvalidateInitCompleteOnMountCount(mountCount: number): [boolean, () => void] {
+  const initCompleteRef = React.useRef<boolean>(false)
+  const previousMountCountRef = React.useRef<number>(mountCount)
+
+  const setInitComplete = React.useCallback(() => {
+    initCompleteRef.current = true
+  }, [])
+
+  if (previousMountCountRef.current !== mountCount) {
+    // mount count increased, re-initialize dom-walker
+    initCompleteRef.current = false
+    previousMountCountRef.current = mountCount
+  }
+
+  return [initCompleteRef.current, setInitComplete]
+}
+
 export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElement> {
   const containerRef = React.useRef<HTMLDivElement>(null)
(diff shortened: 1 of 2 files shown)


Rheeseyb (approved):

editor/src/components/canvas/dom-walker.ts
```diff
@@ -373,7 +390,7 @@ export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElem
               }
             }
 
-            if (cachedMetadata == null) {
+            if (cachedMetadata == null || initComplete === false) {
```
seanparsons, line 393:
```suggestion
            if (cachedMetadata == null || !initComplete) {
```
*6 million dollar man eye zoom*

seanparsons (approved):

editor/src/components/canvas/dom-walker.ts
balazsbajorics: Fixed.

But I think it's worth a quick vote. 
The dom walker function became rather complicated and harder to maintain in the last few months, and while changing it I felt (type) safer with the `=== false` check because if you change the type from boolean to something else (for example you wrap it in an array and forget to destructure) typescript will break on `=== true` but it will let you test the truthyness if you omit the `===`. 

OTOH, our (unwritten) style guide says that boolean checks should not compare to true or false.

What do you think?

seanparsons: Do we still have no way to prevent truthiness? As that is the real underlying issue here.

balazsbajorics: AFAIK there's no way to disallow truthiness in typescript

Rheeseyb: Looks like we could do with using https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md

PR #727, opened 2020-11-25
Experiment/components

enidemi:
This PR contains previous component experiments. 
- Isolation Mode
- Split Pane Navigator
- Component Children Selection/Highlights
- Second Canvas for Isolated Component (disabled by default) - you need to disable 'Component Isolation Mode' before enabling this since the implementation of two experiments clash

To try it on different project please add `?branch_name=experiment/components`.

🍕 https://utopia.pizza/p/02cdda47-my-custom-experiment/?branch_name=experiment/components

PR #743, opened 2020-12-01
experiment(Canvas) Enable second canvas experiment

Rheeseyb:
A very very hacky attempt at simulating the "Second Canvas" proposal, whereby we always show a second canvas for isolating the currently selected component. This is built using the existing isolation mode, with the second canvas just being a new scene that's always rendered in the same place.

When selecting anything in the left "Canvas", we update the isolated component used to feed the right "Canvas". I've added a divider `div` to the actual example project to separate the canvii.

:pizza: https://utopia.pizza/p/119ab8a6-auspicious-bergamot/?branch_name=experiment/second-canvas

### DIFF
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -1753,16 +1753,29 @@ export const UPDATE_FNS = {
           componentsByName,
         )
         if (nearestComponentResult != null) {
-          const { element, component: componentToIsolate } = nearestComponentResult
+          const { element, component: componentToIsolate, path } = nearestComponentResult
+
           const elementName = componentToIsolate.name
           const components = getOpenUtopiaJSXComponentsFromState(editor)
           const storyBoardPath = getStoryboardTemplatePath(components)
           if (elementName !== newIsolatedComponent?.componentName && storyBoardPath != null) {
+            const instanceFrame = MetadataUtils.getFrameInCanvasCoords(
+              path,
+              editor.jsxMetadataKILLME,
+            )
+            const normalisedInstanceFrame =
+              instanceFrame == null
+                ? {
+                    left: 0,
+                    top: 0,
+                    width: 375,
+                    height: 812,
+                  }
+                : canvasFrameToNormalisedFrame(instanceFrame)
+
             const normalisedFrame: NormalisedFrame = {
-              left: 430,
-              top: 0,
-              width: 375,
-              height: 812,
+              ...normalisedInstanceFrame,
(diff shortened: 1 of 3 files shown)


enidemi:
I like how it's easy to track which Component I'm editing.

maltenuhn:
My word, this _is_ hacky :D 

*First impression* (Well, actually, third or fourth more like, once I got my head around it and resized / hid various UI pieces: this is interesting. But I don't know if it's an improvement over in-place editing or in-place isolating. Let me esplain...

This "mode" only makes sense if I don't get "reach into any component" (and select things being output by its render fn) by default. So it needs me to select a component, realize it's a component, try to target something I see, realize it's probably something in the render function of the component (vs something I pass in), then incur the wrath of Fitt and his laws, and then... traverse into the thing, make the changes, etc etc.

Which is all fine, but if I could eg "drag an instance onto the canvas and make a scene out of it that renders that component", or some such mechanism, I'd get the same result, without Fitt's penalty, and don't have a chunk of my canvas taken up by valuable real estate that _most of the time_ doesn't really help me much.

 I think the calculus _would_ change if the "right pane" actually gave me new superpowers: seeing props and state, being able to move nodes with things like `{props.children}` around, etc. But in this mode as it stands, all I can do is what I can already do on a canvas - select/reparent/position/style - if I'm able to target components.

*Other things* 
- There _is_ something strangely alluring about a detail view pane for me. Also, seeing more clearly that "ah, this thing I've selected _is_ a component" -- with this view components become _very_ hard to miss once you have selected them. 
- I wonder if "show me props", "show me state" would make sense to put in the same pane. An optional "detail view", if you like. Maybe even with the detail navigator?
- Similarly, we haven't yet touched things like 'styled components'. They probably require some sort of special mode to edit (eg to quickly visualize things like hover states), which would similarly fit into an extra pane.
- I wonder how dimensionally "big" most components are - my guess is that most are _rather_ small, or at least potentially rather small, because they are are really driven by the size of their children (or the available space). If so, that means actually rendering the component isn't all that helpful.

Rheeseyb:
As extremely hacky as this is, I somewhat prefer it over the Isolation Mode. I think with some stylistic changes and some tying into the component navigator it would go a long way to solving confusion around what you're editing, and how to edit components rather than instances.

The main strength that the Isolation Mode has over this is the ability to drill down into components (isolate Component A, then select its instance of Component B to isolate that etc.), though I think we could find an interaction to tackle that in the 2 Canvas approach here.

enidemi:

PR #751, opened 2020-12-03
feature(editor) remove layoutsystem from defaults for insertion, default project, updating frames

### LINKED ISSUE #750: Remove layout property from default elements/project
- [ ] layouttype pins

enidemi:
Fixes #750 

It is the first step to removing the layout property, this PR removes layoutSystem: pinSystem from:
- the default project
- for inserted elements it's replaced with `position: absolute` for non-flex elements. because the defaults give style properties, this changes depending on the parent target while in insertion mode.
- reparenting adds or removes `position: absolute` depending on flex or non-flex parent

### DIFF
--- a/editor/src/components/canvas/controls/insert-mode-control-container.tsx
+++ b/editor/src/components/canvas/controls/insert-mode-control-container.tsx
@@ -13,7 +13,7 @@ import {
   JSXElement,
   JSXElementChild,
 } from '../../../core/shared/element-template'
-import { setJSXValueAtPath, unsetJSXValueAtPath } from '../../../core/shared/jsx-attributes'
+import { setJSXValueAtPath } from '../../../core/shared/jsx-attributes'
 import { PropertyPath, TemplatePath, Imports } from '../../../core/shared/project-file-types'
 import { Either, eitherToMaybe, isLeft, right } from '../../../core/shared/either'
 import { KeysPressed } from '../../../utils/keyboard'
@@ -443,7 +443,11 @@ export class InsertModeControlContainer extends React.Component<
           ...element,
           props:
             eitherToMaybe(
-              unsetJSXValueAtPath(element.props, createLayoutPropertyPath('position')),
+              setJSXValueAtPath(
+                element.props,
+                createLayoutPropertyPath('position'),
+                jsxAttributeValue('relative'),
+              ),
             ) ?? element.props,
         }
       }
--- a/editor/src/components/canvas/controls/insert-mode-control-container.tsx
+++ b/editor/src/components/canvas/controls/insert-mode-control-container.tsx
@@ -13,9 +13,9 @@ import {
   JSXElement,
   JSXElementChild,
 } from '../../../core/shared/element-template'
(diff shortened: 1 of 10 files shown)


alecmolloy (approved):

alecmolloy:
>:)

editor/src/components/canvas/controls/insert-mode-control-container.tsx
```diff
@@ -437,7 +438,16 @@ export class InsertModeControlContainer extends React.Component<
       let { element } = this.props.mode.subject
       if (this.isTextInsertion(element, insertionSubject.importsToAdd)) {
         element = this.getTextElementAutoSizeWithPosition(element, snappedMousePoint)
+      } else if (this.parentIsFlex(this.props.highlightedViews[0])) {
+        element = {
+          ...element,
+          props:
+            eitherToMaybe(
+              unsetJSXValueAtPath(element.props, createLayoutPropertyPath('position')),
```
balazsbajorics, line 446: if I understand correctly, this here means that inserting an element to flex will not give it a position prop. 

if we want to keep a transitional behavior (allowing for any child to be pinned), we want to insert a `position: relative` here

editor/src/core/layout/layout-helpers.ts
```diff
@@ -278,11 +278,19 @@ export const LayoutHelpers = {
   setLayoutAttribute: (newValue: LayoutSystem) => (
     props: JSXAttributes,
   ): Either<string, JSXAttributes> => {
-    return setJSXValueAtPath(
-      props,
-      createLayoutPropertyPath('LayoutSystem'),
-      jsxAttributeValue(newValue),
-    )
+    if (newValue === LayoutSystem.PinSystem) {
```
balazsbajorics, line 281: hah sneaky

balazsbajorics (commented):
my only qualm is that inserting into a flex element should default to `position: relative`, otherwise 👍

editor/src/core/layout/layout-helpers.ts
Rheeseyb: Looks like the only other value this can be called with is `LayoutSystem.Group`, which we don't support anyway, so please can you remove this and update the code that calls it to reflect that.

enidemi: fixed!

enidemi:
> position: relative

I am not convinced about that. Relative positioning gives the option to have top/left style properties too, that is not reflected on the canvas controls.

Rheeseyb:
> > position: relative
> 
> I am not convinced about that. Relative positioning gives the option to have top/left style properties too, that is not reflected on the canvas controls.

The problem is that if we don't include `position: relative` any children of that element won't be able to position themselves with respect to that element.

Rheeseyb (approved):

maltenuhn:
I have been saving this screenshot for a **long** time :) Thank you @enidemi for exorcising the demon ~I asked to adopt a year ago~ we've always hated, for ever and ever.

image

PR #752, opened 2020-12-04
Support comments against components.

seanparsons:
**Problem:**
Currently comments are completely obliterated when the code is regenerated after a model based update of a text file.

**Fix:**
Comments are currently tracked for components, which is to say anything parsed into a `UtopiaJSXComponent` will have any comments preceding it tracked and carried over when the code is regenerated.

**Commit Details:**
- Added a couple of types to represent single or multi-line comments.
- `UtopiaJSXComponent` now includes `leadingComments` as a property.
- Added `getComments` to parse the comments from a TypeScript node.
- Added `addCommentsToNode` to reintroduce comments back into a TypeScript node.
- Slightly reworked `printStatements` to print each node individually
  as the previous code was introducing an indent of 4 characters for no
  reason whatsoever.


balazsbajorics (commented):
does it mean it only stores comments leading to a component? what about comments in the component body or inside the jsx?

balazsbajorics (approved):

editor/src/core/workers/parser-printer/parser-printer-comments.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export function getComments(sourceText: string, node: TS.Node): CommentsResult {
+  let leadingComments: Array<Comment> = []
+  TS.forEachLeadingCommentRange(
+    sourceText,
+    node.getFullStart(),
+    (pos: number, end: number, commentKind: TS.CommentKind, hasTrailingNewLine: boolean) => {
+      switch (commentKind) {
+        case TS.SyntaxKind.SingleLineCommentTrivia:
```
Rheeseyb, line 16: "Trivia"? Ha!

editor/src/core/workers/parser-printer/parser-printer-comments.ts
```diff
@@ -0,0 +1,64 @@
+import * as TS from 'typescript'
+import { Comment, multiLineComment, singleLineComment } from '../../shared/element-template'
+import { fastForEach } from '../../shared/utils'
+
+export interface CommentsResult {
+  leadingComments: Array<Comment>
+}
+
+export function getComments(sourceText: string, node: TS.Node): CommentsResult {
```
Rheeseyb, line 9: So is the idea then that we could effectively run this against all of the nodes we parse to progressively add more and more comment support?

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -564,19 +572,10 @@ function printStatements(statements: Array<TS.Node>, shouldPrettify: boolean): s
...
-    TS.ListFormat.PreferNewLine |
-    TS.ListFormat.MultiLineFunctionBodyStatements |
-    TS.ListFormat.MultiLineTypeLiteralMembers
-  const typescriptPrintedResult = printer.printList(
-    printFlags,
-    TS.createNodeArray(statements),
-    resultFile,
-  )
+  const printedParts = statements.map((statement) => {
+    return printer.printNode(TS.EmitHint.Unspecified, statement, resultFile)
```
Rheeseyb, line 576: What's the deal with removing all of the flags here?

Rheeseyb (approved):

seanparsons:
> does it mean it only stores comments leading to a component? what about comments in the component body or inside the jsx?

Yes to the first bit.

JSX comments appear to be a `JsxExpression` which might just work as is, I'd have to look.

For everything else the complexity is that _anything_ can have comments leading or trailing it so a simple component with no arbitrary block might have a comment attached to what we effectively treat as the root element. Arbitrary blocks will need some customisation because it's entirely likely they'll consist of something like comment -> code -> comment -> code, so the nodes that make it would need to be stored along with their respective comments.

The potentially painful aspect of this will be comments in the midst of expressions, like one in the middle of a JavaScript object. For our parsed attributes I think we can wrap the values with something that holds the comments and then the value becomes a nested object or similar.

editor/src/core/workers/parser-printer/parser-printer.ts
seanparsons: See last commit detail bullet point.

editor/src/core/workers/parser-printer/parser-printer-comments.ts
seanparsons: Yes, there's some varied complexity around this as described in a larger comment block, but that should work, also for eventually adding trailing comments.

editor/src/core/workers/parser-printer/parser-printer-comments.ts
seanparsons: Trivia is what the TypeScript compiler treats anything that could be elided, so comments and whitespace for example.

PR #753, opened 2020-12-07
chore(Editor) Remove electron

Rheeseyb:
Remove old Electron resources since they're causing an issue with Windows


enidemi (commented):
👍🏻

PR #756, opened 2020-12-07
fix(editor) Fixed invalid DOM nesting in Getting Started page

### LINKED ISSUE #754: Getting Started page has a invalid DOM nesting
**Describe the bug**
The ["Getting Started" page](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/documentation/getting-started.tsx) has some invalid DOM nesting, which shows up as a series of warnings in the console when launching the editor.

**To Reproduce**

1. Load the editor
2. Open the Chrome DevTools

Result: Near the top of the console output will be 4 warnings mentioning invalid DOM nesting.

**Expected behaviour**
No invalid DOM nesting warnings in the console

OmarDaSilva:
Fixes #754 

**Problem:**
DOM invalid nest 

**Fix:**
replaced all instances of  `<p>` with `<div>`

### DIFF
--- a/editor/src/components/documentation/getting-started.tsx
+++ b/editor/src/components/documentation/getting-started.tsx
@@ -222,7 +222,7 @@ export const GettingStarted = betterReactMemo('Getting Started', () => {
       <FeaturedProjects />
 
       <H1>Keyboard Shortcuts</H1>
-      <div>
+      <P>
         <ul>
           <li>
             Showing / hiding the code editor: use <PrettyKeys shortcut='⌘⎇.' />
@@ -236,17 +236,17 @@ export const GettingStarted = betterReactMemo('Getting Started', () => {
           You don't need an account to get started, but you'll need to sign up to use assets in your
           products, and share them with others.
         </Callout>
-      </div>
-      <div>
+      </P>
+      <P>
         <H2>Using controls</H2>
         You can add custom controls to your components, and configure them in the inspector. To try
         it, copy and paste this code at the end of your project.
         <SyntaxHighlighter language='jsx' style={syntaxTheme}>
           {codeString}
         </SyntaxHighlighter>
-      </div>
+      </P>
 
-      <div>
+      <P>
         <H2>Known Issues</H2>
         Utopia is in developer preview. You should expect your code to be safe, your projects
         recoverable, and code errors being reported well and consistently. However, some features -
@@ -266,19 +266,19 @@ export const GettingStarted = betterReactMemo('Getting Started', () => {
           </li>
           <li>Class-based components, and top-level code mutation, don't yet work reliably.</li>
         </ul>
-      </div>
+      </P>
 
       <H1>Change Log - June 2020</H1>
       <div>
-        <div>
+        <P>
             <EM>Highlights of June</EM>
             <ul style={{ paddingLeft: 30 }}></ul>
-        </div>
+        </P>
       </div>
 
       <H1>Change Log - May 2020</H1>
       <div>
-        <div>
+        <P>
           <EM>Highlights of May</EM>
           <ul style={{ paddingLeft: 30 }}>
             <li>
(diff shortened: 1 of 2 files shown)


Rheeseyb (approved):

editor/src/components/documentation/getting-started.tsx
```diff
@@ -300,12 +300,12 @@ export const GettingStarted = betterReactMemo('Getting Started', () => {
...
             </li>
           </ul>
-        </p>
+        </div>
       </div>
 
       <H1>Change Log - April 2020</H1>
       <div>
-        <p style={{ marginTop: 36 }}>
+        <div>
```
enidemi, line 308: why are these marginTops removed?

enidemi (approved):

editor/src/components/documentation/getting-started.tsx
```diff
@@ -222,7 +222,7 @@ export const GettingStarted = betterReactMemo('Getting Started', () => {
       <FeaturedProjects />
 
       <H1>Keyboard Shortcuts</H1>
-      <P>
```
alecmolloy, line 225: the styled component ([link](https://github.com/concrete-utopia/utopia/blob/c773a0b171374d43aa0c3b0d5d062b330d14b26f/editor%2Fsrc%2Fcomponents%2Fdocumentation%2Fdocumentation-components.tsx#L51-L53)) has a marginTop. We could just make sure _all_ the "paragraph" (`p`, `P`, et. al.) use `P`, and change the `P` styled component to extend a `div` instead, which would fix the invalid nesting problem.

alecmolloy, line 225: the styled component ([link](https://github.com/concrete-utopia/utopia/blob/c773a0b171374d43aa0c3b0d5d062b330d14b26f/editor%2Fsrc%2Fcomponents%2Fdocumentation%2Fdocumentation-components.tsx#L51-L53)) has a marginTop. We could just make sure _all_ the "paragraph" (`p`, `P`, et. al.) use `P`, and change the `P` styled component to extend a `div` instead, which would fix the invalid nesting problem.

alecmolloy (changes requested):

editor/src/components/documentation/getting-started.tsx
```diff
@@ -270,15 +270,15 @@ export const GettingStarted = betterReactMemo('Getting Started', () => {
 
       <H1>Change Log - June 2020</H1>
       <div>
-        <p style={{ marginTop: 36 }}>
-          <EM>Highlights of June</EM>
-          <ul style={{ paddingLeft: 30 }}></ul>
-        </p>
+        <P>
+            <EM>Highlights of June</EM>
```
alecmolloy, line 274: looks like prettier isn't set up properly 🤔. it should have automatically removed these extra indents, perhaps someone on @concrete-utopia/maintaineers can help set up tomorrow.

editor/src/components/documentation/getting-started.tsx
```diff
@@ -270,15 +270,15 @@ export const GettingStarted = betterReactMemo('Getting Started', () => {
 
       <H1>Change Log - June 2020</H1>
       <div>
```
alecmolloy, line 272: don't think this one is necessary.

alecmolloy (changes requested):
looks almost ready to me. just need to get you set up with prettier to make sure the code is properly formatted. could do with removing extra `div` tags with no styling that directly wrap a `P`, but otherwise i'm happy to merge.

alecmolloy (approved):

balazsbajorics (approved):

editor/src/components/documentation/getting-started.tsx
Rheeseyb: Ok good catch, seems we want to have all of these using that `marginTop`, but making this change means we're losing any possible future legitimate use of `P` as a paragraph tag. If we're going to make this change, we should just change the name of that styled component too so that it's not misleading (I don't care what name we go with, as long as it's not "P").

editor/src/components/documentation/getting-started.tsx
Rheeseyb: Ah I know what the issue here was - when Omar committed this change he was doing it from a different shell that wasn't correctly set up, so the commit hooks didn't run.

PR #762, opened 2020-12-08
Prefer the execution scope over the global object.

### LINKED ISSUE #747: DOM object constructor cannot be called as a function
These two basic projects cause an error observed in a much more complicated project. Not sure why they cause it:

```
TypeError: Failed to construct 'Comment': Please use the 'new' operator, this DOM object constructor cannot be called as a function.
```

- https://utopia.pizza/p/e45dccdc-charming-tibia/
- https://utopia.pizza/p/2226d6d1-shard-tugboat/

seanparsons:
Fixes #747

**Problem:**
When creating a file with a component called `Comment` a weird type error about the DOM constructor being used is triggered. This was because a function called `Comment` was being picked up from the global object over the component in the file.

**Fix:**
The handling of values defined "elsewhere" was overwriting the current scope with those from the global with the same name, so the ordering was reversed.

**Commit Details:**
- Fixes #747.
- Reversed the order of the various clauses inside `resolveDefinedElsewhere`
  such that values in `scope` supersede values in `global`.


editor/src/components/canvas/__snapshots__/ui-jsx-canvas-bugs.spec.tsx.snap
```diff
@@ -0,0 +1,33 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`UiJsxCanvas #747 - DOM object constructor cannot be called as a function 1`] = `
```
balazsbajorics, line 3: I think this test is small enough to be inlined!

editor/src/components/canvas/ui-jsx-canvas-bugs.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <Scene
+      component={App}
+      style={{ position: 'absolute', left: 0, top: 0, width: 375, height: 812 }}
+    />
+  </Storyboard>
+)
+    `,
+    )
+  })
+})
```
balazsbajorics, line 35: thanks for making a new file :)

balazsbajorics (approved):

editor/src/components/canvas/__snapshots__/ui-jsx-canvas-bugs.spec.tsx.snap
seanparsons: Can't easily be done given that `testCanvasRender` checks a bunch of different things.

balazsbajorics: there's `testCanvasRenderInline` to solve that problem

Rheeseyb (approved):

enidemi (approved):

PR #763, opened 2020-12-08
feature(editor) removing layout property (scenes)

### LINKED ISSUE #755: Remove layout property: layouttype PinSystem part 2
Scenes

enidemi:
This time it's Scenes. Closes #755 

- Removed Scene layout property: `layout={{ layoutSystem: 'pinSystem' }}` was always set on Scene elements, now the SceneRoot ignores it. It was also used in a lot of testcases.
- Changed Scene Metadata to not have the container prop anymore.
- Some tests used Scene with layout prop to set TLWH, this is updated to use style prop instead.


editor/src/components/canvas/dom-walker.spec.browser.tsx
```diff
@@ -143,10 +143,9 @@ describe('DOM Walker tests', () => {
           <Storyboard data-uid={'${BakedInStoryboardUID}'}>
             <Scene
               static
-              style={{ left: 0, top: 0, width: 375, height: 812 }}
+              style={{ position: 'relative', left: 0, top: 0, width: 375, height: 812 }}
```
Rheeseyb, line 146: Shouldn't these all be `position: 'absolute'`?

seanparsons (approved):

Rheeseyb (changes requested):
I'm pretty sure this needs a migration (and I think we're missing one for the previous pass at this too)

editor/src/components/canvas/dom-walker.spec.browser.tsx
enidemi: I thought so too, but it turns out utopia-api added relative position to pinsystem parents, and absolute position to all children. I tried to match the current tests with how the pragma works.

enidemi:
> I'm pretty sure this needs a migration (and I think we're missing one for the previous pass at this too)

@Rheeseyb I've updated the PR with a migration step, adding `position: 'absolute'` where it had PinSystem set.  Without the parent layout it is not known if an element is in a flex container or not.

maltenuhn:
Re migrations - do we need to bother at this stage? We're not breaking anything that's not layout, and the layout we break would have never worked outside of Utopia anyway. Just a consideration.

enidemi:
> Re migrations - do we need to bother at this stage? We're not breaking anything that's not layout, and the layout we break would have never worked outside of Utopia anyway. Just a consideration.

Sample projects would look differently, but without errors.

maltenuhn:
> Sample projects would look differently, but without errors.

Yep, and I think that's ok, because they'd look _the same they'd look outside of Utopia_ - which is the whole point (well, some point) of "any code works" anyway.

enidemi:
I actually reverted the migration commit after having a discussion about it today. The main goal is to have project code that can work without using the utopia-api.

Currently there are only a few test projects and those would not throw errors with this change, the only inconvenience is that on Scene elements positioning needs to be set manually as a result of to the container prop removal. Other places that use that layout prop won't break because the utopia-api still supports it.

editor/src/components/canvas/dom-walker.spec.browser.tsx
balazsbajorics: hm I also think it makes more sense as position: 'absolute' because otherwise sibling scenes will muck things up

balazsbajorics (approved):

editor/src/components/canvas/dom-walker.spec.browser.tsx
enidemi: fixed :) 

Maybe in the past even storyboard had the pinSystem set so all Scenes were absolute?

PR #767, opened 2020-12-09
Migrated the old css-style-loader to the webpack styled loaders

### LINKED ISSUE #635: Replace our current CSS custom evaluation with a loader
Since we now have a framework for supporting loaders, we should replace the current CSS custom evaluation logic with a loader (following the example in #631)

Rheeseyb:
Fixes #635 

**Problem:**
Previously we were using a special custom evaluator for handling CSS imports. Since implementing that, we have introduced a system for supporting webpack styled loaders, so we wanted to port the CSS import handling to use that.

**Fix:**
Move the custom evaluator code to the loader instead, and introduce a function for unloading CSS files when the import is removed (since the previous method had regressed)

**Commit Details:**
- Moved the result of calling `transformCssNodeModule` to `css-loader.ts` to ensure the side-effecting code is executed every time the module is imported
- Created `unimportAllButTheseCSSFiles` for unloading previously imported CSS files, with a note that we need a more general way of handling this
- No new tests because this is already covered by existing tests in `package-manager.spec.ts`


balazsbajorics:
could we somehow test the css unloader functionality?

editor/src/core/webpack-loaders/css-loader.ts
```diff
@@ -1,15 +1,68 @@
...
+  return loadModuleResult(filename + '.js', loadedContents)
 }
 
 export const CSSLoader: ModuleLoader = {
   match: matchFile,
   load: loadModule,
 }
+
+// FIXME This doesn't quite fit in here. We need a generalised strategy for unloading files when they're removed from the list of imports
+export function unimportAllButTheseCSSFiles(filesToKeep: Array<string>): void {
```
balazsbajorics, line 48: beautiful name!

balazsbajorics (approved):

Rheeseyb:
> could we somehow test the css unloader functionality?

I've just added a test for that

editor/src/core/webpack-loaders/css-loader.ts
```diff
@@ -1,15 +1,68 @@
...
+  const head = document.querySelector('head')
+  const shouldBeRemoved = (elemId: string) =>
```
seanparsons, line 50: There's a danger (albeit a small one) of this deleting something with a filename that ends the same as another.
`/style.css` versus `/home/style.css` for example.

seanparsons (approved):

editor/src/core/webpack-loaders/css-loader.ts
Rheeseyb: I've just double checked and this isn't an issue because it uses the full path to the file, not just the file name.

PR #768, opened 2020-12-10
Updated the nix-pkgs and package.yaml for the server

Rheeseyb:
Updated the pinned nix packages from 19.09 to 20.09 (the latest release). This broke a couple of things on the server so we had to update the frozen dependencies there too, but chose to pin some of them down (protolude and servant-websockets) in `package.yaml` rather than fix the issues that updating them introduced, since this is mainly about trying to unblock other issues.

As part of this I have also updated the node version from 12.x to 14.x and checked that doesn't break anything


balazsbajorics (approved):

balazsbajorics:
works on my machine! I amended the readme.md too

PR #774, opened 2020-12-11
Fix/upgrade emotion

### LINKED ISSUE #765: Fix Emotion 11.x in the canvas.

seanparsons:
Fixes #765 

**Problem:**
When a user added `@emotion/react` 11.x to their project and imported it in staging/production the styling for the editor ends up getting broken. The underlying issue was that the changes to the Emotion context provider in 11.x meant that it would create a new cache of the classes which coincidentally deletes any existing classes, reaching directly up to `document.head` to find those and eliminate them.

**Fix:**
By using the editor's handle on `@emotion/react` to shim into the package manager, similarly to how we handle `react` and` react-dom` the underlying caching works as is. `@emotion/core` is also shimmed in as that is what `@emotion/react` was in version 10.x and below.

**Commit Details:**
- Bumped Emotion to 11.x in every project.
- Shimmed in the editor's `@emotion/react` for any `@emotion/react` or `@emotion/core` import.


maltenuhn:
> as that is what @emotion/react was in version 10.x and below

... we sincerely hope :)

balazsbajorics (approved):

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -1,3 +1,5 @@
+/** @jsx jsx */
+import { jsx } from '@emotion/react'
```
Rheeseyb, line 2: Does this imply that we now need to include these two lines at the top of every `.tsx` file in the editor? :grimacing:

seanparsons: If we want to use the magical `css` property, yeah.

editor/src/components/navigator/dependency-list-item.tsx
```diff
@@ -1,14 +1,18 @@
...
 import { NpmDependencyVersionAndStatusIndicator } from './dependecy-version-status-indicator'
 import { ContextMenuItem } from '../context-menu-items'
 import { MomentumContextMenu } from '../../uuiui-deps'
-import { MenuProvider } from 'react-contexify'
+import { MenuProvider as ReactContexifyMenuProvider } from 'react-contexify'
 import { NO_OP } from '../../core/shared/utils'
 
+// FIXME: For some reason we've been able to use this incorrectly
+// according to the types, but following the types causes the code to fail.
+const MenuProvider = ReactContexifyMenuProvider as any
```
Rheeseyb, line 14: Well that's concerning!

seanparsons: Agreed, it's used in 3 places and out of the blue TypeScript started complaining about this one.

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -59,8 +59,8 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(565) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(580)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(475) // if this breaks, GREAT NEWS but update the test please :)
```
Rheeseyb, line 62: Winner!

Rheeseyb (approved):

editor/src/components/canvas/controls/new-canvas-controls.tsx
balazsbajorics: but this file was not using it, seems like an unused import to me?

seanparsons: It's definitely using the `css` property, where it renders the `TextEditor` element.

balazsbajorics: interesting, so this is broken on master I guess

maltenuhn: > Does this imply that we now need to include these two lines at the top of every .tsx file in the editor? 😬

that was the case before IIRC?

enidemi (approved):

PR #783, opened 2020-12-15
Make Emotion use globalThis.

seanparsons:
**Problem:**
The `@emotion/react` library references `window` which detonates the web workers killing them on initialisation.

**Fix:**
Use `patch-package` to "fix" the code so that `globalThis` is used in place of `window`.

**Commit Details:**
- Added in patches for the editor and utopia-api that replaces
  `window` with `globalThis` in the `@emotion/react` code.

### DIFF
--- a/editor/package-lock.json
+++ b/editor/package-lock.json
@@ -26883,6 +26883,11 @@
             "@babel/types": "^7.12.5"
           }
         },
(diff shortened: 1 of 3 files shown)


Rheeseyb (approved):
Very sad face

PR #785, opened 2020-12-15
Direct URL for all project files

### LINKED ISSUE #777: Support serving all files directly via URL
On the server we only support loading files that are treated as "assets" via URL (i.e. binary files that have been uploaded to S3). However, users need to be able to reference project text files by URL too (e.g. in the case of rendering an SVG via an `img` tag).

Rheeseyb:
Fixes #777

**Problem:**
Assets and text files are treated separately since we don't want to fill up our database with binary blobs. However, this meant we had no way of providing a URL for directly referencing a project file if that file was a text file.

**Fix:**
Provide an endpoint on the server for attempting to dig into the project's contents. If there is a text file for the provided file path we now return the contents of that, otherwise we fall back to trying to treat it as an asset and checking on S3.

This fix introduces a rather unfortunate coupling of the server and editor, since the server now needs to know enough information about the structure of the project model to read an individual file. To try to protect us against this coupling being broken, I have added a "test" on the JS side that generates the JSON model of the default project, and a test on the server side which uses that model to ensure the server can still read a specific file from it.


seanparsons (approved):

editor/src/sample-projects/sample-project-utils.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { defaultProject } from './sample-project-utils'
+import * as fs from 'fs'
+
+describe('Project model integrity check', () => {
```
balazsbajorics, line 4: is this test here to help us commit the latest SampleProject.json to the repository? or is the idea that there's a committed SampleProject for fallback, but ideally this test will run before the server tests execute?

Rheeseyb: The former. This reminds me though, I need to fix up the workflows that run on PR pushes to ensure that server tests are run on changes that affect the server, since right now we're only running editor tests on PR pushes.

wrt the ordering of the tests - the editor tests and server tests run in parallel as two separate jobs, meaning that even if the editor tests ran first the updated file wouldn't be used in the server tests since it wouldn't be committed to the repo by the CI. We can try to fix that but I think a better fix right now is to just add a check in this `.spec` that the new file matches the old one, then just overwrite if running locally or fail the test if running in the CI.

balazsbajorics: hmm. could we instead make a `ts-node` script that is ran by the server tests to generate this file on-the-fly?

Rheeseyb: I ended up spending 2 hours trying different solutions here and have given up for now. The problem is that we create the default project through code, and in a way that (inconsequential) parts of it are auto-generated and will change each time. That means having a test that checks if the file has changed will always break without a really brittle way of removing those parts, and that a `ts-node` script needs to pull in the editor project to get the dependencies it needs around generating a default `PersistentModel`. I tried adding one to our project but that fails to compile because it struggles to resolve the dependencies, and I also tried adding a separate webpack config to build an executable bundle for a script to generate that model, but had problems there too. I'll come back to this when I have a chance, but until then I have `xit`ed the test to save us from committing a change to that sample project JSON file with every run of the tests.

balazsbajorics: oh, it's too bad `ts-node` couldn't handle the editor codebase! thank you for looking into this :)

balazsbajorics (approved):

editor/src/sample-projects/sample-project-utils.spec.ts
Rheeseyb: I think it most likely can, but needs some time spent figuring out the exact config required.

PR #791, opened 2020-12-18
Fix/inspector calculated values

enidemi:
Fixing inspector hook `useInspectorLayoutInfo` a little for showing calculated values.

The issue was that in one part is was using specialSizeMeasurements for padding only and added a special propertyStatus to have different colors, the other part was using calculated style but didn't set the propertyStatus for the colors. The specialSizeMeasurements is also limited to padding and margin only, so it's not a good fit here.

Changed the text color to grey from light blue for these detected values.

### DIFF
--- a/editor/src/components/inspector/common/property-path-hooks.ts
+++ b/editor/src/components/inspector/common/property-path-hooks.ts
@@ -377,47 +377,35 @@ function parseFinalValue<PropertiesToControl extends ParsedPropertiesKeys>(
   isUnknown: boolean
   usesComputedFallback: boolean
 } {
-  const valueAsMaybe = eitherToMaybe(simpleValue)
+  const simpleValueAsMaybe = eitherToMaybe(simpleValue)
   const rawValueAsMaybe = eitherToMaybe(rawValue)
 
-  function finalValueFromReal(
-    isUnknown: boolean,
-  ): {
-    finalValue: ParsedPropertiesValues
-    isUnknown: boolean
-    usesComputedFallback: boolean
-  } {
-    const parsedRealValue = parseAnyParseableValue(property, realValue, null)
-    if (isRight(parsedRealValue)) {
-      return {
-        finalValue: parsedRealValue.value,
-        isUnknown: isUnknown,
-        usesComputedFallback: false,
-      }
+  const parsedValue = parseAnyParseableValue(property, simpleValueAsMaybe, rawValueAsMaybe)
+  const parsedRealValue = parseAnyParseableValue(property, realValue, null)
+  const parsedComputedValue = parseAnyParseableValue(property, computedValue, null)
+  if (isRight(parsedValue)) {
+    return {
+      finalValue: parsedValue.value,
+      isUnknown: isCSSUnknownFunctionParameters(parsedValue.value),
+      usesComputedFallback: false,
     }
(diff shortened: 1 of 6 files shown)


editor/src/components/inspector/common/property-path-hooks.ts
```diff
@@ -376,31 +375,49 @@ function parseFinalValue<PropertiesToControl extends ParsedPropertiesKeys>(
...
+  } {
+    const parsedRealValue = parseAnyParseableValue(property, realValue, null)
+    if (isRight(parsedRealValue)) {
+      return {
+        finalValue: parsedRealValue.value,
+        isUnknown: isUnknown,
+        usesComputedFallback: false,
+      }
     }
+    const parsedComputedValue = parseAnyParseableValue(property, computedValue, null)
```
Rheeseyb, line 398: Can you turn these into `else if` statements please so it's immediately clear to see where the logic forks

editor/src/components/inspector/common/property-path-hooks.ts
```diff
@@ -768,60 +788,9 @@ export function useInspectorLayoutInfo<P extends LayoutProp | StyleLayoutProp>(
...
 
-function getValueFromSpecialSizeMeasurements<P extends LayoutProp | StyleLayoutProp>(
-  property: P,
-  specialSizeMeasurements: SpecialSizeMeasurements,
-): ParsedProperties[P] | null {
-  // TODO: are there other properties in special size measurementes to extract?
-  let value: number | undefined = undefined
-  switch (property) {
-    case 'paddingLeft':
-      value = specialSizeMeasurements.padding.left
```
balazsbajorics, line 794: do we use specialSizeMeasurements.padding and specialSizeMeasurements.margin anywhere else? if not, we could remove them. but maybe some canvas controls use them?

editor/src/components/inspector/common/property-path-hooks.ts
```diff
@@ -376,31 +375,49 @@ function parseFinalValue<PropertiesToControl extends ParsedPropertiesKeys>(
...
   const rawValueAsMaybe = eitherToMaybe(rawValue)
 
-  function finalValueFromReal(): ParsedPropertiesValues {
-    if (realValue == null) {
-      const parsedComputedValue = parseAnyParseableValue(property, computedValue, null)
-      return defaultEither(emptyValues[property], parsedComputedValue)
-    } else {
-      const parsedRealValue = parseAnyParseableValue(property, realValue, null)
-      return defaultEither(emptyValues[property], parsedRealValue)
+  function finalValueFromReal(
```
balazsbajorics, line 383: this is very hard to read :|

editor/src/components/inspector/common/property-path-hooks.ts
Rheeseyb: We have `margin-controls.tsx` and `padding-controls.tsx` that use them

editor/src/components/inspector/common/property-path-hooks.ts
Rheeseyb: I think extracting the return type would help since it's used above too

enidemi: fixed in effda55

seanparsons (approved):

balazsbajorics (approved):

PR #793, opened 2020-12-18
Remove the paths option from tsconfig.json.

seanparsons:
**Problem:**
The `paths` setting causes problems with (at least) `ts-node`: https://github.com/TypeStrong/ts-node/issues/138

**Fix:**
Removed the option and fixed up all the imports in the project because we currently derive no benefit from it.

**Commit Details:**
- Removed the `paths` value from the tsconfig.json file and
  fixed all the imports that refer to the module name rather than
  the path.


Rheeseyb (approved):

enidemi (approved):

balazsbajorics (commented):
:+1:

Issue #795, opened 2020-12-18
Preserve Comments in JSX


Rheeseyb:
Follow up from #788

Preservation of comments in and around JSX (as outlined in `parser-printer-comments.spec.ts`)

maltenuhn:
From reading `parser-printer-comments.spec.ts` I wasn't clear if this also includes comments inside of JSX attributes, e.g.
```
<div style={{
     color: 'mauve',
     // ^ ^ looks the way it sounds
    background: 'burlywood'
  }}>
```

seanparsons:
@maltenuhn It would appear so given this list of items: https://github.com/concrete-utopia/utopia/pull/796/files?file-filters%5B%5D=.snap&file-filters%5B%5D=.ts#diff-39b9306c4d59d87d304dd47156be3ea8fab4399792c2f861354d80a602740ffcR91-R98

PR #796, opened 2020-12-18
Comment Support Part 2

### LINKED ISSUE #788: Preserve Comments Part 2
Preserve comments:
- [x] in and around arbitrary blocks at the top level.
- [x] in and around arbitrary blocks inside components.
- [ ] in parsed attributes - now covered in #795 

(Follow on work from #752)

Rheeseyb:
Fixes #788 

**Problem:**
Comments are lost during the parsing since the TS AST doesn't care about them explicitly.

**Fix:**
Continue the work outlined in #752 to capture the comments during parsing and store them on the parsed model. This includes capturing trailing comments too. The work is unfortunately piecemeal purely because this is the only way to capture them.

The bulk of the work is in `parser-printer.ts`, `parser-printer-parsing.ts`, and `parser-printer-comments.ts`, with most of the remaining changes being for supporting the passing around of the extra parameters to function calls.

**Commit Details:**
- Added tests to `parser-printer-comments.spec.ts` for capturing the various places comments can be added - this is a useful checklist of the level of support we now have for comments
- `parser-printer-comments.ts` now captures trailing comments as well as leading comments
- Handle the printing of comments (adding them back when creating the AST) inside `parser-printer.ts`
- When parsing, separate the top level arbitrary JS blocks rather than combining them all into one, so that we can keep the captured comments in the right place when printing back out
- No longer blindly create the jsx pragma when printing, since this is now captured as a comment against the first import statement
- Updated a bazillion places that now expect some form of comments to be passed in

A follow up ticket has been created (#795) for retaining the remaining comments.


editor/src/core/workers/parser-printer/parser-printer-comments.spec.ts
```diff
@@ -1,10 +1,15 @@
...
+import {
+  clearParseResultUniqueIDs,
+  parseThenPrint,
+  testParseCode,
+} from './parser-printer.test-utils'
 import { applyPrettier } from './prettier-utils'
 
 describe('parseCode', () => {
-  it('should parse a component with comments in front of it', () => {
+  xit('should parse a component with comments in front of it', () => {
```
seanparsons, line 12: This is now disabled?

editor/src/core/workers/parser-printer/parser-printer-comments.spec.ts
```diff
@@ -35,10 +44,112 @@ describe('parseCode', () => {
...
+          */",
           "trailingNewLine": true,
           "type": "MULTI_LINE_COMMENT",
         },
       ]
     `)
   })
 })
+
+describe('Parsing and printing code with comments', () => {
```
seanparsons, line 60: Nice!

editor/src/core/workers/parser-printer/parser-printer-comments.ts
```diff
@@ -2,63 +2,131 @@ import * as TS from 'typescript'
...
+    result.push(tsComment)
   })
+  return result
+}
+
+export function addCommentsToNode(node: TS.Node, comments: ParsedComments): void {
+  const leadingTSComments = createTSComments(comments.leadingComments)
+  const trailingTSComments = createTSComments(comments.trailingComments)
+
   if (leadingTSComments.length > 0) {
```
seanparsons, line 126: Tiny thing, but probably better to check `comments.leadingComments` here to avoid spawning mountains of empty arrays that are immediately thrown away.

editor/src/core/workers/parser-printer/parser-printer.test-utils.ts
```diff
@@ -77,12 +77,14 @@ import { getUtopiaIDFromJSXElement } from '../../shared/uid-utils'
...
 import { addUniquely, flatMapArray } from '../../shared/array-utils'
 import { optionalMap } from '../../shared/optional-utils'
+import { emptyComments, parsedComments, ParsedComments } from './parser-printer-comments'
 
 export const singleLineCommentArbitrary: Arbitrary<SingleLineComment> = lowercaseStringArbitrary().map(
   (text) => {
     return {
       type: 'SINGLE_LINE_COMMENT',
       comment: text,
+      rawText: text,
```
seanparsons, line 87: This looks a little suspicious, `comment` would from what I can tell never be equal to `rawText`.

seanparsons (approved):

editor/src/core/workers/parser-printer/parser-printer-comments.spec.ts
Rheeseyb: Oops, this was unintentionally committed

PR #798, opened 2020-12-21
Debugging functionality for the parse-print-analysis script.

### LINKED ISSUE #797: Script to debug the results of the parse-print cycle.
When trying to interpret the result from the original work to address #787, it can be troublesome to find what has changed.

seanparsons:
Fixes #797

**Problem:**
The `parse-print-analysis` script produces a lot of output, but it is not particularly easy to then get some information on what is necessary to improve the results that are seen.

**Fix:**
Created another script which can produce some debug like information about a particular file from those listed that can then be used to work through what needs fixing/improving. There are two types of output the script can produce:

The first command looks like this:
`npm run parse-debug -- print 'react-three-flex-examples' '/src/App.js'`
Which prints out the parsed model in full for that particular file, akin to what is produced by the currently available debugging button in the right hand pane.

The second command is very similar:
`npm run parse-debug -- structure 'react-three-flex-examples' '/src/App.js'`
That prints out a more readable and terse hierarchy from parsing the given file, see the `elementsStructure` function as that is the crux of what is used.

**Commit Details:**
- Fixes #797.
- Generalise the earlier work for grabbing projects from Github.
- Added a `parse-debug.ts` script with `print` and `structure` commands.
- Moved `getStoryboardUID` and `getStoryboardTemplatePath` to `scene-utils.ts`.
- `walkElement` better supports handling the depth.
- `elementsStructure` and `forceParseSuccessFromFileOrFail` moved to
  `parser-printer.test-utils.tsx`.
- `elementsStructure` now includes the top level elements and their types,
  as well as the names of JSX elements.
- Added the `ignore-styles` plugin so that ts-node doesn't attempt to process
  CSS files.
- Reworked the ts-node scripts slightly to ensure that we can get the stack
  trace when something fails.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,5 +120,6 @@ server/utopia-web
 utopia-api/index.d.ts
 editor/src/components/editor/npm-dependency/utopia-api-typings.ts
 
-# Parse print analysis results.
+# Cache of projects for scripts.
 .parse-print-analysis
+.github-test-projects
--- a/editor/.dependency-cruiser.js
+++ b/editor/.dependency-cruiser.js
@@ -153,7 +153,7 @@ module.exports = {
       severity: 'error',
       from: {
         pathNot:
-          '\\.(spec|test|spec.browser|test-utils)\\.(js|mjs|cjs|ts|tsx|ls|coffee|litcoffee|coffee\\.md)$',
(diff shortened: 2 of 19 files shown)


Rheeseyb (commented):
Nice!

PR #800, opened 2020-12-22
Dom-based Selection and Highlights

balazsbajorics:
**Problem:**
The selection and highlight system uses AABB (axis-aligned bounding box) bounds to determine where the user clicked. This is an issue with elements that are rotated or otherwise not rectangular. This makes selection feel unreliable/unpredictable.

**Fix:**
This PR fixes the click-to select and mouse-based highlight for elements. It *does not* fix the following, these will be fixed in follow up PRs:
- the actual drawn highlights still use the old AABB
- the spacebar-based selection still uses the old AABB
- insert mode still uses the old AABBs
- on master, when the user clicks to select something, we first dispatch that as a local action (effectively giving it a priority to render) and then in the next animation frame we dispatch the select action. This behavior is not yet working with the new selection hooks.
- if the user has event listeners in the user code, those are *not yet disabled!*

**Commit Details:**
- Created a new family of hooks in `select-mode-hooks`
- useSelectAndHover returns mouse event handlers that the `CanvasContainer` attaches to the canvas root
- the hover and mouseDown detection relies on bubbling events
- fixed a bunch of tests that relied on the old canvas controls

### DIFF
--- a/editor/src/components/canvas/controls/deselect-control.tsx
+++ b/editor/src/components/canvas/controls/deselect-control.tsx
@@ -2,32 +2,37 @@ import * as React from 'react'
 import * as EditorActions from '../../editor/actions/action-creators'
 import { Mode } from '../../editor/editor-modes'
 import { EditorDispatch } from '../../editor/action-types'
+import { betterReactMemo } from '../../../uuiui-deps'
+import { useEditorState, useRefEditorState } from '../../editor/store/store-hook'
+import { pickSelectionEnabled } from './select-mode/select-mode-hooks'
 
-interface DeselectControlProps {
-  mode: Mode
-  dispatch: EditorDispatch
-}
+interface DeselectControlProps {}
 
-export class DeselectControl extends React.Component<DeselectControlProps> {
-  onMouseDown = () => {
-    if (this.props.mode.type !== 'insert') {
-      this.props.dispatch([EditorActions.clearSelection()], 'canvas')
-    }
-  }
+export const DeselectControl = betterReactMemo<DeselectControlProps>('DeselectControl', (props) => {
+  const editorStateRef = useRefEditorState((store) => store.editor)
+  const dispatch = useEditorState((store) => store.dispatch, 'DeselectControl dispatch')
 
-  render() {
-    return (
-      <div
-        className='deselect-control'
-        style={{
-          position: 'absolute',
-          top: 0,
-          left: 0,
-          width: '100%',
-          height: '100%',
-        }}
-        onMouseDown={this.onMouseDown}
-      />
+  const onMouseDown = React.useCallback(() => {
+    const selectionEnabled = pickSelectionEnabled(
+      editorStateRef.current.canvas,
+      editorStateRef.current.keysPressed,
     )
-  }
-}
+    if (selectionEnabled && editorStateRef.current.mode.type !== 'insert') {
+      dispatch([EditorActions.clearSelection()], 'canvas')
+    }
+  }, [dispatch, editorStateRef])
+
+  return (
+    <div
+      data-testid='deselect-control'
+      style={{
+        position: 'absolute',
+        top: 0,
+        left: 0,
+        width: '100%',
+        height: '100%',
+      }}
+      onMouseDown={onMouseDown}
+    />
+  )
+})
--- a/editor/src/components/canvas/controls/deselect-control.tsx
+++ b/editor/src/components/canvas/controls/deselect-control.tsx
@@ -6,9 +6,7 @@ import { betterReactMemo } from '../../../uuiui-deps'
 import { useEditorState, useRefEditorState } from '../../editor/store/store-hook'
 import { pickSelectionEnabled } from './select-mode/select-mode-hooks'
 
-interface DeselectControlProps {}
-
-export const DeselectControl = betterReactMemo<DeselectControlProps>('DeselectControl', (props) => {
+export const DeselectControl = betterReactMemo('DeselectControl', () => {
   const editorStateRef = useRefEditorState((store) => store.editor)
   const dispatch = useEditorState((store) => store.dispatch, 'DeselectControl dispatch')
 
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
@@ -36,11 +36,11 @@ import { selectElementsThatRespectLayout } from '../new-canvas-controls'
 
 const DRAG_START_TRESHOLD = 2
 
-export function pickIsResizing(dragState: DragState | null): boolean {
+export function isResizing(dragState: DragState | null): boolean {
   return dragState != null && dragState.type === 'RESIZE_DRAG_STATE' && dragState.drag != null
 }
 
-export function pickIsDragging(dragState: DragState | null): boolean {
+export function isDragging(dragState: DragState | null): boolean {
   return dragState != null && dragState.type === 'MOVE_DRAG_STATE' && dragState.drag != null
 }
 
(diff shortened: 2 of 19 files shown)


enidemi (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
```diff
@@ -0,0 +1,407 @@
...
+  point,
+  WindowPoint,
+  windowPoint,
+} from '../../../../core/shared/math-utils'
+import { ScenePath, TemplatePath } from '../../../../core/shared/project-file-types'
+import * as TP from '../../../../core/shared/template-path'
+import { NO_OP } from '../../../../core/shared/utils'
+import { KeysPressed } from '../../../../utils/keyboard'
+import { useKeepShallowReferenceEquality } from '../../../../utils/react-performance'
+import utils from '../../../../utils/utils'
```
seanparsons, line 18: This is normally imported with an uppercase `Utils`. :)

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
```diff
@@ -0,0 +1,407 @@
...
+    })
+    const allRoots = MetadataUtils.getAllCanvasRootPaths(componentMetadata).filter((rootPath) => {
+      return !rootElementsToFilter.some((path) => TP.pathsEqual(rootPath, path))
+    })
+    let siblings: Array<TemplatePath> = []
+    utils.fastForEach(selectedViews, (view) => {
+      utils.fastForEach(TP.allPaths(view), (ancestor) => {
+        const ancestorChildren = MetadataUtils.getImmediateChildren(componentMetadata, ancestor)
+
+        siblings.push(...ancestorChildren.map((child) => child.templatePath))
```
seanparsons, line 134: Should probably for each these rather than create an array we immediately bin off.

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
```diff
@@ -0,0 +1,407 @@
...
+    [maybeHighlightOnHover, findValidTarget],
+  )
+
+  const onMouseOut = React.useCallback(() => {
+    maybeClearHighlightsOnHoverEnd()
+  }, [maybeClearHighlightsOnHoverEnd])
+
+  const onMouseDown = React.useCallback(
+    (event: React.MouseEvent<HTMLDivElement>) => {
+      const foundTarget = findValidTarget(event.target as HTMLDivElement, event.metaKey)
```
seanparsons, line 336: Is this cast necessary?

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
```diff
@@ -0,0 +1,407 @@
...
+          DRAG_START_TRESHOLD,
+          startDragState(
+            foundTarget.templatePath,
+            windowToCanvasCoordinates(windowPoint(point(event.clientX, event.clientY)))
+              .canvasPositionRounded,
+          ),
+        )
+
+        if (!foundTarget.isSelected) {
+          const doubleClick = event.detail > 1 // we interpret a triple click as two double clicks, a quadruple click as three double clicks, etc  // TODO TEST ME
```
seanparsons, line 349: TODO for later work?

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (canvasWrapper != null) {
+    const canvasWrapperRect = canvasWrapper.getBoundingClientRect()
+    const canvasDivCoords = {
+      x: screenPoint.x - canvasWrapperRect.left,
+      y: screenPoint.y - canvasWrapperRect.top,
+    } as RawPoint
+    const inverseOffset = negate(canvasOffset)
+    const inverseScale = 1 / canvasScale
+    const pagePosition = scaleVector(canvasDivCoords, inverseScale)
+    const canvasPositionRaw = offsetPoint(pagePosition as any, inverseOffset)
```
seanparsons, line 86: `any` alert!

seanparsons (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
balazsbajorics: yes, all TODOs are meant for follow-up tasks

editor/src/components/canvas/controls/deselect-control.tsx
```diff
@@ -2,32 +2,37 @@ import * as React from 'react'
...
 import { EditorDispatch } from '../../editor/action-types'
+import { betterReactMemo } from '../../../uuiui-deps'
+import { useEditorState, useRefEditorState } from '../../editor/store/store-hook'
+import { pickSelectionEnabled } from './select-mode/select-mode-hooks'
 
-interface DeselectControlProps {
-  mode: Mode
-  dispatch: EditorDispatch
-}
+interface DeselectControlProps {}
```
Rheeseyb, line 9: erm... bin me?

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
```diff
@@ -0,0 +1,407 @@
...
+import CanvasActions from '../../canvas-actions'
+import { DragState, moveDragState } from '../../canvas-types'
+import { createDuplicationNewUIDs, getOriginalCanvasFrames } from '../../canvas-utils'
+import { findFirstParentWithValidUID } from '../../dom-lookup'
+import { useWindowToCanvasCoordinates } from '../../dom-lookup-hooks'
+import { selectElementsThatRespectLayout } from '../new-canvas-controls'
+
+const DRAG_START_TRESHOLD = 2
+
+export function pickIsResizing(dragState: DragState | null): boolean {
```
Rheeseyb, line 39: I understand the choice for the `pick` prefix here, but we're not using it anywhere else in the code base, so please can we remove it? Otherwise we're just going to end up with competing conventions.

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
Rheeseyb: I believe this function was cut and pasted without change from `editor/src/components/canvas/controls/select-mode-control-container.tsx`

editor/src/core/shared/dom-utils.ts
```diff
@@ -202,7 +202,8 @@ export const intrinsicHTMLElementNamesThatSupportChildren: Array<string> = [
   'section',
 ]
 
-export function getDOMAttribute(element: HTMLElement, attributeName: string): string | null {
+// eslint-disable-next-line no-restricted-globals
+export function getDOMAttribute(element: Element, attributeName: string): string | null {
```
Rheeseyb, line 206: Why the change from `HTMLElement` to `Element`?

Rheeseyb (approved):
LGTM! Before merging please can you make sure all follow up work is tracked either in the Paper doc or in tickets

editor/src/core/shared/dom-utils.ts
balazsbajorics: yeah that's a good question. some code used HTMLElement, some used Element. I think Element covers everything we need, so that's why I changed to the broader type. I can be talked into swapping everything to HTMLElement though

Rheeseyb: I'm fine with either, just wanted to make sure this was a deliberate change

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
balazsbajorics: @Rheeseyb I renamed it to isResizing (and the consts in the code to `resizing` so they don't shadow each other)

I now remembered that the reason for pickIsResizing was to be able to call the consts isResizing. could you check if this new updated version is better?

Rheeseyb: Works for me :+1:

PR #801, opened 2020-12-22
Fix/ticount-test

OmarDaSilva:
Fixes #[742]

**Problem:**
Broken Import 
**Fix:**
Renamed Import

**Commit Details:**
- renamed 'move-testFile' to 'move-file'


balazsbajorics:
@OmarDaSilva it looks like you committed the fix to the old branch and opened the PR from that. please instead check out master, make sure it's pulled from the latest, and create a new branch for the fix. and commit the fix on that new branch.

PR #802, opened 2020-12-22
Fixed broken import

OmarDaSilva:
Fixes #[ticket_number]

**Problem:**
Broken import

**Fix:**
Renamed broken import

**Commit Details:**
- renamed import, move-testfIle to move-file


balazsbajorics (approved):
it works again! now we need to figure out why sometimes only half of the image is uploaded :)

seanparsons (approved):

PR #803, opened 2020-12-22
X axis starts at 0, Y range increased to 50, image resized to 800x600, Image still cut when uploading

OmarDaSilva:
Fixes #[802]

**Fix:**
Changed the look of the graph

**Commit Details:**
- x axis values set to 0


performance-test/src/puppeteer-test.ts
```diff
@@ -326,7 +326,7 @@ async function createTestPng(
 }
 
 async function uploadPNGtoAWS(testFile: string) {
-  AWS.config.update({
+  AWS.config.update(
```
balazsbajorics, line 329: looks like you accidentally removed a { here that broke the script

enidemi (approved):

seanparsons (approved):

PR #805, opened 2020-12-30
fix(toasts) - make toasts dismissable, disappear in correct order

### LINKED ISSUE #423: "File saved and formatted" toasts seem to stay up forever, no way to close them

maltenuhn:
Fixes #423. 

**Problem:**
Our toasts today sometimes stay up indefinitely, sometimes close and then re-open, and also can't be dismissed by the user

**Fix:**
Toasts are now uniquely identified, can optionally be user-dismissed, and by default still close after 5.5s (unless `persistent` is set to true)

**Commit Details:**
- Remove `PUSH_TOAST` and `POP_TOAST`
- Replaced them with `ADD_TOAST` and `REMOVE_TOAST`
- Toasts get a unique ID using `UUID`
- UI supports user dismissal of tests
- Removed the tests for `PUSH_TOAST` and `POP_TOAST`
- Inadvertently fixed a bug that meant the last test in the file was unreachable


editor/src/components/code-editor/code-editor.tsx
```diff
@@ -182,7 +182,7 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
             () => {
               this.save(this.state.code, true, {
                 message: 'File saved and formatted',
-                level: 'SUCCESS',
+                level: 'INFO',
```
maltenuhn, line 185: The idea is to use `SUCCESS` only for things where the outcome (or timing of outcome) is unexpected, e.g. adding a dependency / uploading an image that takes an unpredictable amount of time.

editor/src/components/common/notices.tsx
```diff
@@ -17,14 +19,16 @@ export interface Notice {
   message: React.ReactChild
   persistent?: boolean
   level?: NoticeLevel
+  id?: string
```
maltenuhn, line 22: probably not the best pattern but I was unclear whether it's better to make the ID non-optional and force each caller to provide it (but then how to enforce uniqueness), or to make something like a InnerNotice type (without ID), and an OuterNotice prop (with ID).

editor/src/components/editor/actions/actions.ts
```diff
@@ -1789,18 +1790,26 @@ export const UPDATE_FNS = {
...
-    setTimeout(() => dispatch([popToast()], 'everyone'), 5500)
+  ADD_TOAST: (action: AddToast, editor: EditorModel, dispatch: EditorDispatch): EditorModel => {
+    const uniqueToastId = UUID()
+
+    if (!action.toast.persistent) {
+      setTimeout(() => dispatch([removeToast(uniqueToastId)], 'everyone'), 5500)
+    }
+    const newToast: Notice = {
+      ...action.toast,
+      id: uniqueToastId,
```
maltenuhn, line 1801: See comment above; here I'm basically nuking whatever ID has been set, but that then requires two types (or alternatively not enforcing uniqueness, unless there's a type _for that_?)

seanparsons (approved):

editor/src/components/code-editor/code-editor.tsx
Rheeseyb: Sounds good but in that case we should rename it to reflect that (e.g. to `ASYNC_COMPLETED` or something highlighting that it was triggered by a successful async side effect). Probably something for a separate PR

editor/src/components/common/notices.tsx
Rheeseyb: It seems like ID should be made non-optional. Using a random (v4) UUID is enough to enforce uniqueness since the likelihood of clash is so small.

editor/src/components/editor/actions/actions.ts
Rheeseyb: If we make ID non-optional then we can create the UUID in the function that creates `AddToast` actions, removing the need for this new object

editor/src/components/common/notices.tsx
maltenuhn: Can you suggest a way to do this that addresses my comment from above? Issue is unless there's a type that enforces or correlates with uniqueness, we rely on _anywhere in our codebase_ to actually _use_ UUID.

maltenuhn:
@Rheeseyb thank you for your comments. I've tried adding in the changes you suggested but don't seem to get anywhere that doesn't cause a large number of undesired changes - e.g. forcing each place that can fire a toast to also provide an `id` in the form of a string (and hoping that whoever uses it in the future will remember to use `uuid` to create the string). I assume the real answer is to use a separate type for the notice content (without `id`) and notice-in-our-model (with `id`), but that, too, will cause a large number of changes.

Re: tests, the previous tests IMO are no longer useful because they relied on a particular way of toasts working (i.e. the stack) that this PR removes. They were also not correct in the first place, in that they didn't catch the real-life problems with "stuck" toasts, and also didn't correctly test that toasts with `persistent: true` stayed up and instead expected them to close. 

As a result of this, I will make this PR a draft and add it to the backlog, since with the additional changes now required it'd become a larger endeavour than its worth.

Rheeseyb:
@maltenuhn gimme a min, I'll take a look now because it should be a quick fix

Rheeseyb:
Ok I have replaced all direct creations of a `Notice` object with calls to the `notice` factory function, which now also generates a unique ID. I have also restored and updated the tests for the `ADD_TOAST` and `REMOVE_TOAST` to reflect the new behaviour.

enidemi (approved):

PR #806, opened 2020-12-30
Fix/change font size editor

OmarDaSilva:
(edit MN) This PR introduces a feature to change the font size of the code editor, as a relatively straightforward feature (and a showcase for our redux-y architecture)


maltenuhn:
I'm going to remove draft status here since this is feature-complete and ready for review. Unfortunately it also had my branch merged into it (unnecessarily; afaik no functionality in this PR depends on it)

editor/src/components/editor/actions/actions.ts
```diff
@@ -479,6 +482,8 @@ import {
 } from './action-creators'
 import { EditorTab, isOpenFileTab, openFileTab } from '../store/editor-tabs'
 import { emptyComments } from '../../../core/workers/parser-printer/parser-printer-comments'
+import { number } from 'prop-types'
+import { Editor } from 'draft-js'
```
Rheeseyb, line 486: These imports are definitely wrong. Likely accidentally added by autocomplete.
```suggestion
```

editor/src/components/editor/action-types.ts
```diff
@@ -287,6 +287,19 @@ export type ToggleInterfaceDesignerCodeEditor = {
...
 
+export type IncrementCodeEditorFontSize = {
+  action: 'INCREMENT_CODE_EDITOR_FONT_SIZE'
+}
+
+export type DecrementCodeEditorFontSize = {
+  action: 'DECREMENT_CODE_EDITOR_FONT_SIZE'
+}
+
+export type UpdateCodeEditorFontSize = {
```
Rheeseyb, line 298: Since we always know the current font size when dispatching these actions, we can remove the `Increment` and `Decrement` actions and just use the `Update` action in all cases.

OmarDaSilva:
This PR is ready for review

maltenuhn:
(I suspect this PR won't work anymore once the switch to VSCode is complete. I think @seanparsons and @Rheeseyb are going to look at the settings part of VSCode imminently. If we're lucky, we will get "changeable font size" free with VSCode settings - that would make this PR obsolete however.)

Rheeseyb:
> If we're lucky, we will get "changeable font size" free with VSCode settings 

Yep, we do indeed

This PR is broken btw, given it has 52 files changed. It seems you've somehow managed to install some node dependencies at the root level and committed those. Since the feature has since become obsolete, I recommend we close this PR and if there are no issues with it I'll delete the branch to save bloat on the repo.

maltenuhn:
Yep. A sad ending for this PR, but some learnings here @OmarDaSilva: 

1 - avoid accidentally committing node modules. The way they work is that we commit _the requirement to have them_ (in package.json).
2 - It's totally not a competition, but it pays to be faster than Rheese and Sean :D

PR #807, opened 2020-12-31
LeftPanelAnimation

OmarDaSilva:
Problem:
The left panel from editor-component.tsx does not animate in and out

FIx:
For the animation to work on the left panel it would have to always be present and not replaced by null when we want it closed so I removed the ternary operator, and thus the left panel will always be present. To hide the panel when we want it to close I set the width to 0. These changes allowed the CSS transition property to correctly animate the left panel.


OmarDaSilva:
Can you please check this PR?

OmarDaSilva:
Can you please check this PR?

editor/src/components/editor/hook-utils.ts
```diff
@@ -26,3 +26,22 @@ export function useForceUpdate() {
   const [, forceUpdate] = React.useReducer((c) => c + 1, 0)
   return forceUpdate
 }
+
+export function useDelayedValueHook(inputValue: boolean, delayMs: number): boolean {
+  const [returnValue, setReturnValue] = React.useState(inputValue)
+  React.useEffect(() => {
+    let timerID: any = undefined
+    if (inputValue) {
+      // we do not delay the toggling if the input value is true
```
Rheeseyb, line 35: This custom logic is specific to our use of the value in the `editor-component`, and so shouldn't live here. For the sake of this PR, I'd say we should move this code from here into `editor-component`, otherwise this code is presenting itself as being far more generic than it actually is.

seanparsons (approved):

editor/src/components/menubar/menubar.tsx
```diff
@@ -119,6 +119,7 @@ export const Menubar = betterReactMemo('Menubar', () => {
   )
 
   const onDoubleClick = React.useCallback(() => {
+    // TODO KILL ME
```
Rheeseyb, line 122: Follow up work?

maltenuhn: @OmarDaSilva as I mentioned, this _entire_ callback isn't necessary anymore and can be removed

editor/src/components/editor/hook-utils.ts
maltenuhn: @Rheeseyb next up is using this for the right pane (eg inspector) and then things that should similarly animate (eg sections in the left menu). What about the behaviour is specific, and how can we make it more generic?

editor/src/components/menubar/menubar.tsx
OmarDaSilva: Sorted

editor/src/components/editor/hook-utils.ts
OmarDaSilva: @Rheeseyb Hey Rheese any suggestions to make it more generic as @maltenuhn mentioned?

balazsbajorics: I think we shouldn't try to make it more generic than it is. when you work on the right pane animation, you should start with a copy-pasted version, and tailor it for the inspector's needs. once that is done, we can look at the two hooks, and see what are the similarities, and how (and if) it can be merged into one hook.

Rheeseyb: Yeah, as Balazs says we don't need to make this more generic right now. I'm just suggesting that we please move the code into the source file where it is used so that it isn't being presented as being directed.

balazsbajorics (approved):

editor/src/components/editor/hook-utils.ts
OmarDaSilva: Done!

OmarDaSilva:
Added the last changes and ready for review

Rheeseyb (approved):
After removing the commented import this is good to go :+1:

PR #809, opened 2021-01-04
data(any-code-works) - add more repos to the automated testing script

maltenuhn:
This adds a number of repositories for automated testing, to catch more cases, complete with additional hashes. Libraries were selected from a list of the most popular React-related repos, with a focus on components / component libraries (rather than, e.g., frameworks):

* react-select (jed w)
* react-dates (airbnb)
* react-big-calendar (jquense)
* antd (ant)
* react-responsive (contra)
* react-resizable (resizable)
* react-data-grid (adazzle)

Commit in each case was "latest merged to master". I'm not sure what the rules are for adding things to the `pathsToIgnore` property. Please advise.

### DIFF
--- a/editor/src/scripts/github-projects.ts
+++ b/editor/src/scripts/github-projects.ts
@@ -29,6 +29,55 @@ export const githubProjects: Array<GitRepoWithRevision> = [
     revision: 'd3eeb96536d9e7a78435b967a41d656164535528',
     pathsToIgnore: [],
   },
+  {
+    name: 'ant-design',
+    username: 'utopia-test',
+    repositoryName: 'ant-design',
+    revision: '739f87ed3dcf613214e7504cf712ac9f1c03d315',
+    pathsToIgnore: [],
+  },
+  {
+    name: 'react-select',
+    username: 'JedWatson',
+    repositoryName: 'react-select',
+    revision: 'd1e660c6b261d7fd60a85a6eca2ee9e3e0348ea2',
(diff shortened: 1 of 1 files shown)


Rheeseyb:
> I'm not sure what the rules are for adding things to the pathsToIgnore property. Please advise.

In some cases repos will include entire forks / copies of other projects, which might have undergone minification etc. This is typically because the repo has a dependency on that code which isn't resolvable via normal means (maybe the version or even library isn't available on a public npm registry or maybe it includes some patching or has some non-compatible code removed for specific reasons). For those repos we want to exclude those folders since we're only interested in testing the project itself and not its dependencies.

Rheeseyb (approved):

seanparsons:
> > I'm not sure what the rules are for adding things to the pathsToIgnore property. Please advise.
> 
> In some cases repos will include entire forks / copies of other projects, which might have undergone minification etc. This is typically because the repo has a dependency on that code which isn't resolvable via normal means (maybe the version or even library isn't available on a public npm registry or maybe it includes some patching or has some non-compatible code removed for specific reasons). For those repos we want to exclude those folders since we're only interested in testing the project itself and not its dependencies.

`react-three-flex-examples` ignores a bundled `three.js` file which is half a megabyte IIRC. That alone makes the output really unwieldy on top of the above. I could also see it being used to exclude files which are similarly massive and not necessarily React related for example as well.

seanparsons (approved):

PR #811, opened 2021-01-05
Retain the declaration syntax used for function components

### LINKED ISSUE #808: Preserve function vs var style for parsed components
Currently we treat all parsed functional components as being defined using `var` syntax (e.g. `var App = (props) => `). If a component has been defined using `function` syntax we should honour that when re-generating the code.

Rheeseyb:
Fixes #808 

**Problem:**
We were previously treating all function components as `var` arrow functions, meaning any re-printing of the code would replace `let` or `const` with a `var`, and would switch `function` to a `var` arrow function (also breaking `export default function` style components)

**Fix:**
Capture the declaration style at the point of parsing by checking the flags on the variable declaration list, or just retaining the fact that it was in fact a regular `function`.

**Commit Details:**
- Introduced the types `VarLetOrConst` and `FunctionDeclarationSyntax` to track which type of declaration was used
- Store that information in the `UtopiaJSXComponent` object
- Ensure we use that information to determine how to print the component in `parser-printer.ts`
- Added some tests to a new test file `parser-printer-code-preservation.spec.ts `


seanparsons:
Testing a comment.

seanparsons (approved):

PR #817, opened 2021-01-05
feature: disable user events on canvas in edit mode

balazsbajorics:
**Problem:**
The new dom-based mouse down selection means we need to let the mouse events reach the user's content. This has the side effect that if the user has any active mouse event listeners, those will trigger.

**Fix:**
I disable all user dom event listeners while the canvas is in edit mode. I do it by filtering out every prop that is listed in React Typings' `DOMAttributes`

**Commit Details:**
- Added exhaustive list of `DOMAttributes` to filter out
- using this new list to filter the rendered props in `renderJSXElement`

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
@@ -35,6 +35,7 @@ import { cssValueOnlyContainsComments } from '../../../printer-parsers/css/css-p
 import { filterDataProps } from '../../../utils/canvas-react-utils'
 import { buildSpyWrappedElement } from './ui-jsx-canvas-spy-wrapper'
 import { createIndexedUid } from '../../../core/shared/uid-utils'
+import { filterPropsForUserEvents } from './ui-jsx-canvas-event-handlers'
 
 export function renderCoreElement(
   element: JSXElementChild,
@@ -52,6 +53,7 @@ export function renderCoreElement(
   jsxFactoryFunctionName: string | null,
   codeError: Error | null,
   shouldIncludeCanvasRootInTheSpy: boolean,
+  canvasIsLive: boolean,
 ): React.ReactElement {
   if (codeError != null) {
     throw codeError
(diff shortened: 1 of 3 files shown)


enidemi (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -248,8 +255,7 @@ function renderJSXElement(
   const elementIsIntrinsic = ElementFromScopeOrImport == null && isIntrinsicElement(jsx.name)
   const elementIsBaseHTML = elementIsIntrinsic && isIntrinsicHTMLElement(jsx.name)
   const FinalElement = elementIsIntrinsic ? jsx.name.baseVariable : ElementFromScopeOrImport
-  const finalProps =
-    elementIsIntrinsic && !elementIsBaseHTML ? filterDataProps(elementProps) : elementProps
+  const finalProps = filterProps(canvasIsLive, elementIsIntrinsic, elementIsBaseHTML, elementProps)
```
Rheeseyb, line 258: This will only disable those handlers on something defined in this file, so an imported component with handlers will still react to mouse events, e.g.:
https://utopia.pizza/p/c27666a4-thorn-summer/?branch_name=feature/disable-user-events-on-canvas

maltenuhn:
Does this also cover css-based responses to events, eg `:hover` or similar?

seanparsons (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
balazsbajorics: oh, you are right. this means that my entire approach is wrong here :D

balazsbajorics:
> Does this also cover css-based responses to events, eg `:hover` or similar?

good shout! no this does not cover that. well, that's the second reason why this approach is wrong 😭

balazsbajorics:
I'm closing this as this doesn't prevent css and third party code from receiving mouse events

PR #818, opened 2021-01-06
Don't wrap JSX string literals in braces

### LINKED ISSUE #816: Don't add or remove braces around JSX property values
During parsing and printing we shouldn't replace `<div data-something='something' />` with `<div data-something={'something'} />` and vice versa

Rheeseyb:
Fixes #816 

**Problem:**
When printing a JSX attribute that is a string literal we were always wrapping it in braces.

**Fix:**
When printing the attribute, if the value is a string literal we don't wrap in braces. The ideal fix would be to track whether the value was or wasn't originally wrapped in braces, but doing so would require a huge chunk of refactoring for a very little gain (especially since the braces could be added again via https://github.com/yannickcr/eslint-plugin-react/blob/HEAD/docs/rules/jsx-curly-brace-presence.md)

**Commit Details:**
- The actual functional change is in `editor/src/core/workers/parser-printer/parser-printer.ts`, where we now check if the value is a string literal, and in that case don't wrap it in a JSX expression
- Fixed a bajillion tests that were using `data-uid={'blah'}` since the braces are now removed :upside_down_face:

### DIFF
--- a/editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
+++ b/editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
@@ -28,24 +28,24 @@ describe('moving a scene/rootview on the canvas', () => {
             <View
               style={{ width: 375, height: 812 }}
               layout={{ layoutSystem: 'pinSystem' }}
-              data-uid={'aaa'}
-              data-testid={'aaa'}
+              data-uid='aaa'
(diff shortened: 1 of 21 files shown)


enidemi (approved):

seanparsons (approved):

balazsbajorics (commented):
:+1:

PR #819, opened 2021-01-06
Feature/canvas top menu formula bar

### LINKED ISSUE #810: Canvas topmenu with formula bar

enidemi:
Fixes #810 
Added a topmenu for the canvas that contains the button to toggle the navigator and has a formula bar, based on the canvas experiment #637

**Commit Details:**
- the filetabs no longer takes full width, because the area is shared with the canvas topmenu
- fixed a bug where closing/opening the code editor made it appear as 0 width code editor
- created components for TopMenu and FormulaBar
- moved the navigator button from the right menu to the top menu
- action and update functions to set child text content
- model update from typing into the formula bar is debounced for better performance (similarly to the code editor)

Screenshot 2021-01-06 at 17 27 45


editor/src/components/canvas/controls/formula-bar.tsx
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+      return null
+    }
+  }, 'Formula Bar selectedElement')
+
+  const [simpleText, setSimpleText] = React.useState('')
+  const [disabled, setDisabled] = React.useState(false)
+
+  React.useEffect(() => {
+    if (saveTimerRef.current != null) {
```
seanparsons, line 38: What's this for as it doesn't look like it's used in this block?

seanparsons (approved):

editor/src/components/canvas/controls/formula-bar.tsx
enidemi: it is used for keeping the inputfield state when a user is typing while the save action is still waiting inside the timer

editor/src/components/editor/actions/actions.ts
```diff
@@ -4171,6 +4173,29 @@ export const UPDATE_FNS = {
       return UPDATE_FNS.OPEN_EDITOR_TAB(openTab, updatedEditor)
     }
   },
+  UPDATE_CHILD_TEXT: (action: UpdateChildText, editor: EditorModel): EditorModel => {
```
Rheeseyb, line 4176: Looks to me like this will always remove any regular children on the selected element, rather than just replacing a text block child.

maltenuhn: That should be the case; the experiment was deliberately permissive. You could even enter things like `<h1>Hello World</h1>` there. Perhaps we need a better name for the action, eg `UPDATE_CHILD_CONTENT` ?

editor/src/components/canvas/controls/formula-bar.tsx
balazsbajorics: it looks like if the user changes the element while the timer is still running, the formula bar is going to remain unchanged, showing the text of the previously selected element.

This is beacuse you use a ref for the timer, which is tied to the lifecycle of the FormulaBar component, so if the selected element changes, the ref is still pointing to an active timer, which we do not want to cancel if the user types more, because that would lead to a loss of the data of the last few keystrokes on the previously selected element.

I think the best solution to solve this would be to remount the FormulaBar when selection changes. Then the old ref will remain intact while the timer runs out, and we create a new ref for the newly selected element.
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key

editor/src/components/canvas/design-panel-root.tsx
```diff
@@ -149,42 +151,80 @@ export const DesignPanelRoot = betterReactMemo('DesignPanelRoot', (props: Design
...
+            <CodeEditorWrapper />
+          </Resizable>
+        </SimpleFlexColumn>
+        {props.isUiJsFileOpen ? (
+          <SimpleFlexColumn style={{ flexGrow: 1 }}>
+            <SimpleFlexRow
+              className='topMenu'
+              style={{
+                minHeight: 30,
+                height: 30,
```
balazsbajorics, line 208: could you make this height a const here, and use it for the NavigatorComponent's top too?

editor/src/components/documentation/release-notes.tsx
```diff
@@ -3,33 +3,50 @@ import { jsx } from '@emotion/react'
...
-        paddingBottom: 18,
-        width: '100%',
-        cursor: 'text',
-        userSelect: 'text',
-        WebkitUserSelect: 'text',
-        overflow: 'scroll',
-        whiteSpace: 'pre-wrap',
-        '&  *': {
+    <SimpleFlexColumn>
+      <SimpleFlexRow
```
balazsbajorics, line 15: this SimpleFlexRow with the same style object is used in two places – could you make it part of FileTabs to avoid copy-pasted stlye code?

editor/src/components/editor/actions/actions.ts
Rheeseyb: Doing that would just result in that exact text appearing on the Canvas until the code file is then re-parsed, e.g.:
Screenshot_20210107_144159

maltenuhn: Yep. We'll likely want to enhance this so that only hitting `Enter` actually confirms the entry (and then also causes re-parsing).

editor/src/components/documentation/release-notes.tsx
enidemi: fixed

editor/src/components/canvas/design-panel-root.tsx
enidemi: fixed

editor/src/components/canvas/controls/formula-bar.tsx
enidemi: fixed

editor/src/components/editor/actions/actions.ts
enidemi: The formula bar is limited now because it is not interactive for non-text children, with future improvements this updater function will change too.

balazsbajorics (approved):

PR #820, opened 2021-01-06
Handle comments correctly either side of JSX expressions.

### LINKED ISSUE #795: Preserve Comments in JSX

seanparsons:
Fixes #795

**Problem:**
Currently we don't carry over comments properly from arbitrary JSX blocks.

**Fix:**
This has involved all sorts of weird trickery trying to navigate around all the bizarre things with comments that the TypeScript compiler API does:
- Duplicates comment ranges across multiple levels of an expression.
- A ternary (for example) has leading comments attached normally when just a raw one, if it's inside a JSX expression the leading comments are treated as trailing comments attached to the open brace of the JSX expression.
- Since we have to generate a big string to eval, have to carefully slip calls to `addCommentsToNode` as little as necessary so that the string doesn't massively blow up in size.

**Notes:**
This includes some changes to support comments around attributes (mostly adding fields where appropriate) as that was starting out to be the bigger piece of work until the nightmare of JSX arbitrary blocks loomed large.

**Commit Details:**
- Preparatory work for properly supporting comments in and around parts of JSX attributes.
- `BaseComment` now includes `pos` to help with deduplication.
- Removed a bunch of unused code and dependencies from `json-to-expression.ts`.
- Added `addCommentsToCode` as a helper to prefix and suffix a string
  with some parsed out comments.
- Added `getJSXExpressionLeadingComments` as a helper to help cater for
  one of the problems with comments in JSX expressions having their leading
  comments actually attached to the open brace.
- Slightly tweaked the parameters of `parseJSXArbitraryBlock` so that we
  have access to the whole expression, not just the expression within that.
- In the above, lift the code for the block by snipping the entire text
  contents out of it.
- Use multi-line comments only in the arbitrary generators because single-line
  comments can be inserted in such a way that TypeScript generates nonsense
  code that will fail a subsequent parse.
- `transformer.ts` reworked in a few ways to try and carry over comments in a
  sensible way which involves mostly dancing around the various issues like
  passing in some comments to add on and de-duplicating those comments found
  because they can re-appear on multiple levels of an expression.
- Renamed a function called `BigIntLiteral` because that's shadowing a core class constructor.


balazsbajorics:
> Use multi-line comments only in the arbitrary generators because single-line
comments can be inserted in such a way that TypeScript generates nonsense
code that will fail a subsequent parse.

does this mean we turn single-line comments into multiline before passing it to the generator?

editor/src/components/canvas/__snapshots__/ui-jsx-canvas-bugs.spec.tsx.snap
```diff
@@ -11,7 +11,7 @@ exports[`UiJsxCanvas #747 - DOM object constructor cannot be called as a functio
   \\"
 >
   <div
-    data-utopia-scene-id=\\"fc3/12b\\"
+    data-utopia-scene-id=\\"32d/0f1\\"
```
balazsbajorics, line 14: interesting – these changed without a change to ui-jsx-canvas-bugs.spec. is this because our hash generation changed?

seanparsons:
> does this mean we turn single-line comments into multiline before passing it to the generator?

This is explicitly for the arbitrary generators used in the testing, not in the core code.

editor/src/components/canvas/__snapshots__/ui-jsx-canvas-bugs.spec.tsx.snap
seanparsons: Hmmm, the hash generation itself hasn't changed, I would surmise this is due to comments being included in the representation of the JSX attributes.

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -951,23 +952,21 @@ export const disabledFunctionName = UtopiaUtils.disabled.name
...
   const indexOfLastEnabledLayer = findLastIndex(isLayerEnabled, boxShadows)
-  return jsxAttributeValue(
-    [...boxShadows]
+  return jsxAttributeValue([...boxShadows]
       .map((boxShadow, i) => {
-        const comma = indexOfLastEnabledLayer > i && boxShadow.enabled
-        const { inset, offsetX, offsetY, blurRadius, spreadRadius, color } = boxShadow
-        const parts = Utils.stripNulls([
+      const comma = indexOfLastEnabledLayer > i && boxShadow.enabled
+      const {inset, offsetX, offsetY, blurRadius, spreadRadius, color} = boxShadow
```
balazsbajorics, line 958: this is weird! why is the formatting different here? shouldn't Prettier put a space between the `{` and `i`?

seanparsons: Your guess is as good as mine, answers on a postcard please.

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -169,6 +169,7 @@ import { fastForEach } from '../../core/shared/utils'
 import { UiJsxCanvasContextData } from './ui-jsx-canvas'
 import { addFileToProjectContents, contentsToTree } from '../assets'
 import { openFileTab } from '../editor/store/editor-tabs'
+import {emptyComments} from "../../core/workers/parser-printer/parser-printer-comments";
```
Rheeseyb, line 172: I have a horrible feeling we're going to see all of these import statements change in a future PR as a result of something prettier or eslint related, because of the missing space inside the braces, and the quotation marks :grimacing:

balazsbajorics: oh I didn't notice these! didn't I fix the tests to break if something is not formatted according to prettier? or did we just talk about this?

in any case, you can run `prettier` from the command line and tell it to go over a list of files, or every file

editor/src/components/inspector/common/css-utils.ts
Rheeseyb: It seems like something strange has happened with either prettier or eslint, because a bunch of indenting has changed in this file too

balazsbajorics: maybe the Jetbrains auto refactor re-printed the entire file, and then it skipped prettier?

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -3242,48 +3237,46 @@ export function printBackgroundImage(
...
-        const comma = indexOfLastEnabledLayer > i && bgSize.enabled
-        switch (bgSize.size.value.type) {
+      const comma = indexOfLastEnabledLayer > i && bgSize.enabled
+      switch (bgSize.size.value.type) {
           case 'keyword': {
-            return printEnabled(
-              printComma(printCSSKeyword(bgSize.size.value), comma),
-              bgSize.enabled,
-            )
+              return printEnabled(
```
Rheeseyb, line 3251: e.g. from here onwards it seems the indenting has changed to 4 spaces rather than 2

editor/src/components/inspector/common/css-utils.ts
seanparsons: We run prettier from a commit hook though right?

balazsbajorics: > We run prettier from a commit hook though right?

AFAIK yes, unless that broke recently?

editor/src/components/inspector/common/css-utils.ts
seanparsons: That does look wrong.

editor/src/components/inspector/common/css-utils.ts
Rheeseyb: Yeah we still run it in a commit hook, and the config doesn't appear to have changed at all

balazsbajorics: however, I think many of us have their IDE also run prettier, so we might not see it if it silently breaks

Rheeseyb: I believe if it failed to run it would fail the commit

editor/jest.config.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     {
       testEnvironment: 'jest-environment-jsdom-global',
       preset: 'ts-jest',
-      testPathIgnorePatterns: ['/lib/', '/node_modules/'],
+      testPathIgnorePatterns: ['/lib/', '/node_modules/', '/.github-test-projects/'],
```
Rheeseyb, line 7: Is this still needed / did this even make a difference?

seanparsons: I guess there's no harm in keeping it.

editor/src/core/workers/parser-printer/ts-creator/index.ts
```diff
@@ -32,7 +32,7 @@ export default function create(code: string): string {
     ScriptKind.TSX,
   )
 
-  const factoryFile = transformSourceFile(file)
+  const factoryFile = transformSourceFile(code, file)
```
Rheeseyb, line 35: What was the reason for this change?

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -118,67 +118,90 @@ function buildPropertyCallingFunction(
...
+        )()
+        return newExpression
+      case 'ATTRIBUTE_FUNCTION_CALL':
+        return buildPropertyCallingFunction(attribute.functionName, attribute.parameters)
+      default:
+        const _exhaustiveCheck: never = attribute
+        throw new Error(`Unhandled prop type ${JSON.stringify(attribute)}`)
+    }
+  }
+  // Slide the comments onto the expression.
```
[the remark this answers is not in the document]
seanparsons: That is how I enter every social situation.

editor/src/core/workers/parser-printer/ts-creator/index.ts
seanparsons: Need the original text to get the comments from it.

editor/src/core/workers/parser-printer/ts-creator/transformer.ts
```diff
@@ -168,1359 +180,1447 @@ import {
...
   exclamationToken?: ExclamationToken
 }
 
-function generateEnumMember(node: EnumMember) {
-  return createTsCall('createEnumMember', [
-    transformVisitor(node.name),
-    transformVisitor(node.initializer),
-  ])
-}
+function produceNodeFromNode(
```
Rheeseyb, line 191: Well this name is confusing!

editor/jest.config.js
Rheeseyb: Potential confusion if it does nothing

Rheeseyb (approved):

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -237,20 +247,32 @@ function parseObjectLiteralExpression(
   let highlightBounds = existingHighlightBounds
   let propsUsed: Array<string> = []
   let definedElsewhere: Array<string> = []
+
+  // Get comments attached to the open brace.
+  let openBraceComments: ParsedComments = emptyComments
+  const firstChild = literal.getChildAt(0, sourceFile)
+  if (firstChild != null && firstChild.kind === TS.SyntaxKind.OpenBraceToken) {
```
balazsbajorics, line 254: isn't this the same code as `getJSXExpressionLeadingComments` ?

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -274,30 +296,35 @@ function parseObjectLiteralExpression(
...
-        contents.push(jsxSpreadAssignment(subExpressionValue))
+        contents.push(jsxSpreadAssignment(subExpressionValue, emptyComments))
         highlightBounds = subExpression.value.highlightBounds
         propsUsed.push(...subExpression.value.propsUsed)
         definedElsewhere.push(...subExpression.value.definedElsewhere)
       }
     }
+
+    // First prop reset after everything has been handled.
+    firstProp = false
```
balazsbajorics, line 323: :| I think it'd be worth turning this into a const and changing the loop to another syntax which gets the index

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
seanparsons: Very similar, I'll likely be revisiting this when finishing off the work around including comments in JSX attributes.

balazsbajorics (commented):
all I can offer is an `emptyComment`

PR #821, opened 2021-01-07
Assorted Select Mode Hook Fixes

balazsbajorics:
**Commit Details:**
- I re-enabled the "select on the canvas, and one frame later dispatch the selection" behavior
  - for this, I added Recoil to the editor to be able to trigger a state refresh from across the hierarchy without having to re-render the canvas itself or mess with layers of passing through props
- I unified mouseDown behavior of the Scene labels and the actual elements
- I fixed a bug when a dragStarting in a multiselection across scenes would select an unexpected element
- I removed some unused code from select-mode-control-container

### DIFF
--- a/editor/package.json
+++ b/editor/package.json
@@ -165,7 +165,6 @@
     "react-virtualized-auto-sizer": "1.0.2",
     "react-vtree": "2.0.0",
     "react-window": "1.8.5",
-    "recoil": "0.1.2",
     "reselect": "4.0.0",
     "resize-observer-polyfill": "1.5.0",
     "scheduler": "0.17.0",
--- a/editor/package.json
+++ b/editor/package.json
@@ -165,6 +165,7 @@
     "react-virtualized-auto-sizer": "1.0.2",
     "react-vtree": "2.0.0",
     "react-window": "1.8.5",
+    "recoil": "0.1.2",
     "reselect": "4.0.0",
     "resize-observer-polyfill": "1.5.0",
     "scheduler": "0.17.0",
--- a/editor/src/components/canvas/controls/new-canvas-controls.tsx
+++ b/editor/src/components/canvas/controls/new-canvas-controls.tsx
@@ -35,7 +35,6 @@ import { flatMapArray } from '../../../core/shared/array-utils'
 import { targetRespectsLayout } from '../../../core/layout/layout-helpers'
 import { createSelector } from 'reselect'
 import { PropertyControlsInfo } from '../../custom-code/code-file'
-import { usePropControlledStateV3 } from '../../inspector/common/inspector-utils'
 import { colorTheme } from '../../../uuiui'
 import { betterReactMemo } from '../../../uuiui-deps'
 import {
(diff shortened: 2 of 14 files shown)


editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -204,12 +207,32 @@ export const selectElementsThatRespectLayout = createSelector(
...
-const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
-  const [localSelectedViews, setLocalSelectedViews] = usePropControlledStateV2(
-    props.derived.canvas.transientState.selectedViews,
+/**
+ * these local states are deliberately stored as atoms!
+ * this makes it easier to create a shortcut so we can update the selected views from ui-jsx-canvas
+ * this ONLY works with a single canvas!
+ * overhaul this for multi-canvas support
+ */
+const localSelectedViewsAtom = atom<Array<TemplatePath>>({ key: 'localSelectedViews', default: [] })
```
balazsbajorics, line 216: We have a perceived performance trick: if you click to select something on the canvas, we only redraw the canvas controls on the first frame, _then_ dispatch the selection action to the editor. This is so that we can visually react to the user interaction the next possible frame without having to worry about the editor update's speed. This hack vastly improves the percieved performance of the editor.

The problem I need to solve now is that select-mode-hooks are used from the UiJsxCanvas component, but I need to change the state of NewCanvasControls when the selection changes. And I don't want to re-render the UiJsxCanvas, because that would jeopardize the reaction time. So I needed a way to be able to update the canvas controls without having to rerender the canvas, or having to rerender a common ancestor and dealing with memoization and/or refs to achieve the same effect.

You can think of a Recoil Atom as a "global variable". If you have reference to the _same_ Atom, it will hold the same value for all the places that are accessing it. If any one place updates the value of the atom, every subscriber (places that are calling the `useRecoilState` hook) will be re-rendered.
Recoil also supports write-only access to the Atom, which means you can have some components which can update the value of the atom, but do not need to rerender if the atom value changes. This feature made me pick Recoil over Jotai, which does not support write-only access.

editor/package.json
```diff
@@ -165,6 +165,7 @@
     "react-virtualized-auto-sizer": "1.0.2",
     "react-vtree": "2.0.0",
     "react-window": "1.8.5",
+    "recoil": "0.1.2",
```
seanparsons, line 168: Might be worth mentioning this and talking it through a bit more than sneaking it in via a PR!

editor/src/components/inspector/common/inspector-utils.ts
```diff
@@ -48,26 +49,20 @@ export function usePropControlledState<T>(
   return [localState, setLocalState, forceUpdate]
 }
 
-export function usePropControlledStateV2<T>(propValue: T): [T, React.Dispatch<T>] {
+export function usePropControlledStateV3<T>(
```
seanparsons, line 52: Do we need the version suffix?

seanparsons (approved):

balazsbajorics:
I broke the performance regression tests. The problem is that I was calling a setState on the atom in propControlledStateV3 – this unfortunately caused a double-rendering for the canvas controls. I couldn't come up with a clean way to use atoms – and not have to setState them. So I reverted to propControlledStateV2, and moved the localSelectedViews to the common ancestor (`CanvasWrapperComponent`).

In the end this causes an improvement of the render count and uses less magic, and so I removed recoil altogether.

enidemi (approved):

PR #827, opened 2021-01-07
Fixed incorrect handling of export default statements

### LINKED ISSUE #826: Preserve trailing `export default`
Currently the parser printer will remove the trailing `export default` statement in the below example:
```
const Thing = 1
export default Thing
```

Rheeseyb:
Fixes #826 

**Problem:**
`export default` statements were being either (incorrectly) added as modifiers to components defined elsewhere in the file, or removed completely.

**Fix:**
This PR adds a new type for `export default` modifiers (rather than named assignments) so that we can ensure we are only adding those modifiers back on to components that they were originally present on. It also updates the function that produces the export statements to include export default statements.

**Commit Details:**
- Introduced the type `ExportDefaultDetailModifier` for differentiating between `export default` modifiers vs assignments
- Updated `produceExportStatements` to ensure we print `export default` statements
- Added two test cases to capture default exports for a component and a regular constant

### DIFF
--- a/editor/src/scripts/github-projects.ts
+++ b/editor/src/scripts/github-projects.ts
@@ -76,7 +76,7 @@ export const githubProjects: Array<GitRepoWithRevision> = [
     username: 'contra',
     repositoryName: 'react-responsive',
     revision: 'b6ffef5c4a4b89c69299c2b04c9196dde59b3ef4',
-    pathsToIgnore: [],
+    pathsToIgnore: ['/samples/sandbox/dist/sample.js'],
   },
   {
     name: 'context-provider-hooks-sample',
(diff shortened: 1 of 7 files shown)


editor/src/scripts/github-projects.ts
```diff
@@ -76,7 +76,7 @@ export const githubProjects: Array<GitRepoWithRevision> = [
     username: 'contra',
     repositoryName: 'react-responsive',
     revision: 'b6ffef5c4a4b89c69299c2b04c9196dde59b3ef4',
-    pathsToIgnore: [],
+    pathsToIgnore: ['/samples/sandbox/dist/sample.js'],
```
Rheeseyb, line 79: Forgive me for this sneaky sneak. It was either slip this in here or push it straight to master.

balazsbajorics (approved):

editor/src/scripts/github-projects.ts
seanparsons: There is no forgiveness.

seanparsons (approved):

PR #831, opened 2021-01-08
Feature/dom based insert mode

### LINKED ISSUE #829: Canvas Selection: make insert mode use the dom-based selection

balazsbajorics:
Fixes #829 

**Problem:**
The insert mode was still using the old hittarget divs.

**Fix:**
Use the dom-based highlight functionality of the select-mode-hooks.

**Commit Details:**
- I kept most of the actual logic in insert-mode-control-container to avoid a massive PR
- I refactored the highlight hook so it can take a getter function for potential highlight targets. implemented two functions (`gethHiglightableViewsForInsertMode` for insert mode, and `getSelectableViewsForSelectMode` for select mode). the two functions are wrapped in hooks so they have access to ref editor state.

### DIFF
--- a/editor/src/components/canvas/controls/insert-mode-control-container.tsx
+++ b/editor/src/components/canvas/controls/insert-mode-control-container.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-expressions */
 import * as React from 'react'
 import { FramePoint } from 'utopia-api'
 import { LayoutHelpers, TopLeftWidthHeight } from '../../../core/layout/layout-helpers'
@@ -36,7 +37,7 @@ import { LeftMenuTab } from '../../navigator/left-pane'
 import * as PP from '../../../core/shared/property-path'
 import * as TP from '../../../core/shared/template-path'
 import CanvasActions from '../canvas-actions'
-import { InsertDragState, insertDragState } from '../canvas-types'
+import { CanvasContainerID, InsertDragState, insertDragState } from '../canvas-types'
 import { GuidelineWithSnappingVector } from '../guideline'
 import { ComponentAreaControl, ComponentLabelControl } from './component-area-control'
 import { GuidelineControl } from './guideline-control'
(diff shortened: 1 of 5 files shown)


editor/src/components/canvas/controls/insert-mode-control-container.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-expressions */
```
Rheeseyb, line 1: What is this being used for? Please can we only use this around the actual expressions we want to ignore.

Rheeseyb (approved):

editor/src/components/canvas/controls/select-mode/insert-mode-hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+  const storeRef = useRefEditorState((store) => {
+    return {
+      componentMetadata: store.editor.jsxMetadataKILLME,
+      mode: store.editor.mode,
+      imports: getOpenImportsFromState(store.editor),
+    }
+  })
+  return React.useCallback(
+    // we don't use allElementsDirectlySelectable and childrenSelectable in insert mode, but they are needed for select mode
+    (_allElementsDirectlySelectable: boolean, _childrenSelectable: boolean) => {
```
seanparsons, line 25: Surely just don't include the parameters if none of them are used?

editor/src/components/canvas/controls/insert-mode-control-container.tsx
balazsbajorics: eslint has a problem if you call a function that involves `?.` the no-unused-expressions rule is useful to catch the mistake of a line of code that does nothing, but I wonder if I should just disable this rule, given that we don't usually make that kind of mistake.

alternatively we should bump eslint, but that involves changing the eslint we embed in the editor. or we should have two eslint dependencies, one aliased for bundling, and one for our own actual use.

editor/src/components/canvas/controls/select-mode/insert-mode-hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+import {
+  insertionSubjectIsDragAndDrop,
+  insertionSubjectIsJSXElement,
+  isInsertMode,
+} from '../../../editor/editor-modes'
+import { getOpenImportsFromState } from '../../../editor/store/editor-state'
+import { useRefEditorState } from '../../../editor/store/store-hook'
+import { useHighlightCallbacks } from './select-mode-hooks'
+
+function useGethHiglightableViewsForInsertMode() {
```
seanparsons, line 15: `Geth`? :)

seanparsons (approved):

editor/src/components/canvas/controls/select-mode/insert-mode-hooks.ts
balazsbajorics: geth!

PR #838, opened 2021-01-11
Adjusted JSX boolean attribute styling

### LINKED ISSUE #837: For boolean JSX attributes, use the shorthand style if true, or the explicit style if false
e.g. when printing code we should go with the style of:
`<div data-trueprop data-falseprop={false} />`

Rheeseyb:
Fixes #837 

**Problem:**
We always go with the shorthand boolean attribute style, meaning false props are removed.

**Fix:**
Use the shorthand style for true values, and the explicit style for false values

**Commit Details:**
- Updated the printing logic in `parser-printer.ts`
- Added a test and updated 2 existing tests for the boolean styling


seanparsons (approved):

enidemi (approved):

balazsbajorics (commented):
👍

PR #839, opened 2021-01-11
Tests/performance resize select

enidemi:
Created performance 2 test buttons for element resize and selection next to the other performance test button, hidden behind the `Performance Test Triggers` feature flag.

Resizing test uses the selectedview and resizes it from a corner.
Selection test selects a view then deselects a view and measures both performance.

These are not used in a performance tests yet.

### DIFF
--- a/editor/src/components/menubar/menubar.tsx
+++ b/editor/src/components/menubar/menubar.tsx
@@ -4,7 +4,11 @@ import styled from '@emotion/styled'
 import * as React from 'react'
 import { FLOATING_PREVIEW_BASE_URL } from '../../common/env-vars'
 import { LoginState } from '../../common/user'
-import { useTriggerScrollPerformanceTest } from '../../core/model/performance-scripts'
+import {
+  useTriggerScrollPerformanceTest,
+  useTriggerResizePerformanceTest,
+  useTriggerSelectionPerformanceTest,
+} from '../../core/model/performance-scripts'
 import { useReParseOpenProjectFile } from '../../core/model/project-file-helper-hooks'
 import { shareURLForProject } from '../../core/shared/utils'
 import { isFeatureEnabled } from '../../utils/feature-switches'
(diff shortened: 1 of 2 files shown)


editor/src/core/model/performance-scripts.ts
```diff
@@ -33,3 +45,123 @@ export function useTriggerScrollPerformanceTest(): () => void {
   }, [dispatch])
   return trigger
 }
+
+export function useTriggerResizePerformanceTest(): () => void {
+  const dispatch = useEditorState(
+    (store) => store.dispatch as DebugDispatch,
+    'useTriggerResizePerformanceTest dispatch',
+  )
+  const metadata = useEditorState(
```
balazsbajorics, line 54: since you are only using these in the callback, I think it's worth replacing this with a useRefEditorState, so the component using this hook doesn't rerender when the selected views or the metadata changes

balazsbajorics (approved):

editor/src/core/model/performance-scripts.ts
enidemi: fixed

editor/src/components/menubar/menubar.tsx
```diff
@@ -174,9 +180,17 @@ export const Menubar = betterReactMemo('Menubar', () => {
...
+          <Tile style={{ marginTop: 12, marginBottom: 12 }}>
+            <a onClick={onTriggerScrollTest}>P S</a>
+          </Tile>
+          <Tile style={{ marginTop: 12, marginBottom: 12 }}>
+            <a onClick={onTriggerResizeTest}>P R</a>
+          </Tile>
+          <Tile style={{ marginTop: 12, marginBottom: 12 }}>
+            <a onClick={onTriggerSelectionTest}>P E</a>
+          </Tile>
+        </React.Fragment>
```
Rheeseyb, line 193: If we're only using these for puppeteer, we could presumably style these to have zero width and height so that they're not visible to an actual user

Rheeseyb (approved):

editor/src/components/menubar/menubar.tsx
seanparsons: They're not rendered unless the trigger is enabled right?

seanparsons (approved):

editor/src/components/menubar/menubar.tsx
Rheeseyb: Yeah but it's enabled by default

PR #840, opened 2021-01-11
Parser Printer blank line insertion

### LINKED ISSUE #843: Apply heuristics to determine when to insert blank lines whilst printing code
Since we currently are unable to track blank lines when parsing and printing code, we should apply a series of heuristics to determine when we should insert them.

Rheeseyb:
Fixes #843 

**Problem:**
We print code without any blank lines between top level statements because we're unable to track their existence during parsing.

**Fix:**
Apply a series of rules to determine when to insert them

**Commit Details:**
- Updated the printing logic in `parser-printer.ts` to apply rules and optionally insert a blank line
- Added a test and fixed all existing tests that were broken by this

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer.ts
@@ -672,7 +672,7 @@ function isFunctionDeclaration(line: string): boolean {
   return withoutDefault.startsWith('function')
 }
 
-function createBlanklineTracker() {
+function createBlanklineTracker(): (statement: TS.Node, printedCode: string) => boolean {
   // When inserting blank lines we apply the following series of rules
   // 1) we shouldn't insert blank lines in amongst import statements or exports declarations (not the modifier keyword)
   // 2) variable declarations shouldn't insert a blank line unless a single declaration is spread over separate lines
--- a/editor/src/core/workers/parser-printer/parser-printer.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer.ts
@@ -102,7 +102,7 @@ import {
 import { getBoundsOfNodes, guaranteeUniqueUidsFromTopLevel } from './parser-printer-utils'
 import { ParserPrinterResultMessage } from './parser-printer-worker'
 import creator from './ts-creator'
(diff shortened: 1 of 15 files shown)


editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -654,6 +655,74 @@ export const printCode = memoize(printCodeImpl, {
   maxSize: 1,
 })
 
+function isVariableDeclaration(line: string): boolean {
```
seanparsons, line 658: This looks really brittle to me! It's probably fine for this use case but I can certainly imagine this is going to break on slightly different code.

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -654,6 +655,74 @@ export const printCode = memoize(printCodeImpl, {
...
+
+function isFunctionDeclaration(line: string): boolean {
+  const withoutExport = line.startsWith('export ') ? line.slice(7) : line
+  const withoutDefault = withoutExport.startsWith('default ')
+    ? withoutExport.slice(8)
+    : withoutExport
+  return withoutDefault.startsWith('function')
+}
+
+function createBlanklineTracker() {
```
seanparsons, line 675: Would it be possible to have a return type on this function please?

editor/src/core/workers/parser-printer/parser-printer.ts
Rheeseyb: Yeah, I think this whole PR is one of those "it's not great but better than nothing" situations unfortunately :grimacing:

editor/src/core/workers/parser-printer/parser-printer.ts
Rheeseyb: Done

editor/src/core/workers/parser-printer/parser-printer.ts
seanparsons: That's cool.

seanparsons (approved):

balazsbajorics (commented):
:+1:

PR #841, opened 2021-01-11
Dom-based select hooks use canvas control overlay div

balazsbajorics:
**Problem:**
The dom-based select mode hooks removed the canvas control overlay. This meant that they could directly subscribe to mouseDown events on the canvas elements. However, it had the unfortunate side effect that the user's code got every mouse event that was previously conveniently prevented by the canvas controls overlay div.

**Fix:**
This PR reinstates the canvas control overlay div with pointerEvents enabled. The hover and select detection now uses the `document.elementsFromPoint(point.x, point.y)` API instead of relying on mouseDown and mouseOver events.

**Commit Details:**
- Moved the hook entry point to new-canvas-controls
- Changed the hooks to use `getValidTargetAtPoint`
- Moved the `localSelectedViews` performance-hack to new-canvas-controls as now that is the highest point in the tree that needs them. 
- Made `canvas-component-entry` and `ui-jsx-canvas` clean again

### DIFF
--- a/editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
+++ b/editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
@@ -17,11 +17,19 @@ import { determineElementsToOperateOnForDragging } from './controls/select-mode/
 import { BakedInStoryboardUID } from '../../core/model/scene-utils'
 import { CanvasControlsContainerID } from './controls/new-canvas-controls'
 
-// we need to set the Electron window to a larger size so document.elementsUnderPoint works correctly!
-const currentWindow = require('electron').remote.getCurrentWindow()
-currentWindow.setSize(2200, 1000)
-
 describe('moving a scene/rootview on the canvas', () => {
+  beforeAll((done) => {
+    // we need to set the Electron window to a larger size so document.elementsUnderPoint works correctly!
+    const currentWindow = require('electron').remote.getCurrentWindow()
+    const size = currentWindow.getSize()
+    if (size.width !== 2200) {
+      currentWindow.once('resize', () => {
+        done()
+      })
+      currentWindow.setSize(2200, 1000)
+    }
+  })
+
   it('dragging a dynamic scene’s root view sets the scene position', async () => {
     const testCode = Prettier.format(
       `
--- a/editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
+++ b/editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
@@ -15,6 +15,11 @@ import { PrettierConfig } from '../../core/workers/parser-printer/prettier-utils
 import { createFakeMetadataForParseSuccess, wait } from '../../utils/test-utils'
 import { determineElementsToOperateOnForDragging } from './controls/select-mode/move-utils'
 import { BakedInStoryboardUID } from '../../core/model/scene-utils'
+import { CanvasControlsContainerID } from './controls/new-canvas-controls'
+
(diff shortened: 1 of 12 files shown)


enidemi (approved):

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -46,6 +46,27 @@ export function findFirstParentWithValidUID(
...
+export function getValidTargetAtPoint(
+  validTemplatePaths: Array<string>,
+  point: WindowPoint | null,
+): TemplatePath | null {
+  if (point == null) {
+    return null
+  }
+  const elementsUnderPoint = document.elementsFromPoint(point.x, point.y)
+  return (
+    stripNulls(
```
seanparsons, line 58: This would probably be better formulated as a regular for...of loop that returns the not null case with a default return null after it. Otherwise this is creating two arrays that are immediately thrown away for that first result, after going through two arrays fully.

seanparsons (approved):

editor/src/components/canvas/dom-lookup.ts
Rheeseyb: Or just an [`Array.prototype.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)

Rheeseyb (approved):
Nice!

editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
```diff
@@ -15,6 +15,11 @@ import { PrettierConfig } from '../../core/workers/parser-printer/prettier-utils
 import { createFakeMetadataForParseSuccess, wait } from '../../utils/test-utils'
 import { determineElementsToOperateOnForDragging } from './controls/select-mode/move-utils'
 import { BakedInStoryboardUID } from '../../core/model/scene-utils'
+import { CanvasControlsContainerID } from './controls/new-canvas-controls'
+
+// we need to set the Electron window to a larger size so document.elementsUnderPoint works correctly!
+const currentWindow = require('electron').remote.getCurrentWindow()
+currentWindow.setSize(2200, 1000)
```
balazsbajorics, line 22: this was _very_ painful to debug :( elementsUnderPoint apparently does not work outside of an electron window's visible area… :| 

to make things worse, even though this function is synchronous, it apparently takes a moment for this to take effect. so it took me some time to realize that the resizing _is_ the solution, but it needs to happen some time before the test runs.

I feel like this is a house of cards, but I don't have a better solution than to set it at the root of the module here. this way by the time the test actually runs, the window will be resized

balazsbajorics: hmm, now that I wrote it out "loud" , I think I probably should do something like wait for the electron resize event: https://www.electronjs.org/docs/api/browser-window#event-resize

balazsbajorics: fixed! I pushed a much improved setup which uses `beforeAll` and waits for the resize event

editor/src/components/canvas/dom-lookup.ts
seanparsons: It's applying some logic inside this loop however, which is why that can't be used alone.

balazsbajorics: fixed

editor/src/components/canvas/canvas-utils-scene.spec.browser.ts
Rheeseyb: :+1:

PR #842, opened 2021-01-11
Still support modifying elements in arbitrary blocks.

### LINKED ISSUE #834: Fix being able to manipulate elements inside arbitrary JSX blocks.
Appears the recent work to support comments in arbitrary blocks has broken the ability to reproduce the code after manipulating the elements.

Sample project: [this project](https://utopia.pizza/p/4a1b2e66-elegant-wallaby/) - you are able to select the blue rectangle on the canvas and drag it or update it via the Inspector, but doing so no longer updates the code. This is a very recent regression, possibly caused by a data-uid fix that was included in #820

seanparsons:
Fixes #834

**Problem:**
In an earlier piece of work the printing of code was changed to pull across the comments, but inadvertently stopped some generated `data-uid` keys from coming through. This resulted in elements embedded in arbitrary blocks not getting updated correctly.

**Fix:**
The core of this was to carry over the correct variant of the JavaScript code, which then caused a bunch of fallout problems mostly around the start position as that shifts with the leading comments as they're earlier in the file than the code they relate to.

**Commit Details:**
- Fixes #834.
- `parseOtherJavaScript` supports a more complicated type instead of
  just raw `Node` elements so that the comments and the start position
  including those comments can be carried through.
- `parseJSXArbitraryBlock` does some juggling to get the right values
  passed through into `parseOtherJavaScript` including some surrounding
  whitespace.
- Switched `jsxElementToExpression` from using the `originalJavascript`
  property to the `javascript` property as that includes the comments
  and the `data-uid`.
- Additional mitigation in `transformSourceFile` to cater for expressions
  that now have an ending semicolon which trailing comments can be set on
  as leading comments.

### DIFF
--- a/editor/src/core/workers/parser-printer/ts-creator/transformer.ts
+++ b/editor/src/core/workers/parser-printer/ts-creator/transformer.ts
@@ -158,9 +158,6 @@ import {
   updateSourceFileNode,
   createExpressionStatement,
   JsxText,
-  getSyntheticLeadingComments,
-  setSyntheticLeadingComments,
-  getSyntheticTrailingComments,
   createCall,
   getTrailingCommentRanges,
 } from 'typescript'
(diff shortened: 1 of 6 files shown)


balazsbajorics (approved):

seanparsons:
Test added to cover the code rewriting when `elementsWithin` are updated.

editor/src/core/workers/parser-printer/ts-creator/transformer.ts
```diff
@@ -1609,15 +1606,29 @@ function produceNodeFromNode(
...
-  const updatedStatements = sourceFile.statements.map((statement) => {
-    const lastToken = sourceFile.getLastToken(sourceFile)
-    let additionalTrailingComments: Array<ParserPrinterComment> = []
-    if (lastToken != null && lastToken.kind === SyntaxKind.EndOfFileToken) {
-      // A sneaky way to take comments attached to the EndOfFileToken and put them on the node
-      // that is created. As this code can be run for segments of an actual file, it doesn't
-      // even make sense for there to be a parallel of the EndOfFileToken in what we're handling.
-      const lastTokenComments = getComments(sourceText, lastToken)
-      additionalTrailingComments = lastTokenComments.leadingComments
+  const fileLastToken = sourceFile.getLastToken(sourceFile)
```
Rheeseyb, line 1609: Does the below fix mean we can now remove this unsupported case? https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/workers/parser-printer/parser-printer-comments.spec.ts#L103

Rheeseyb (approved):

editor/src/core/workers/parser-printer/ts-creator/transformer.ts
seanparsons: Nope, as that handling is for arbitrary blocks, potentially it could be co-opted/refactored to be used to fix that though.

Rheeseyb: Got ya :+1:

PR #845, opened 2021-01-12
Tests/performance resize select part 2

enidemi:
Second part of the performance tests.

- 2 new tests in `puppeteer-test.ts` for resizing and selecting elements
- moved common functionality to smaller functions, like image upload and creating the browser
- updated the result string to have the extra images and result data


performance-test/src/puppeteer-test.ts
```diff
@@ -127,7 +239,16 @@ async function createTestPng(
...
+  isSelectionTest: boolean,
+): Promise<string | null> {
+  if (
+    process.env.PERFORMANCE_GRAPHS_PLOTLY_USERNAME == null ||
+    process.env.PERFORMANCE_GRAPHS_PLOTLY_API_KEY == null
+  ) {
+    console.info('Plotly graph generation skipped because of missing username or API key')
+    return null
+  }
+  const xAxisRange = isSelectionTest ? { min: 100, max: 200 } : { min: 0, max: 134 }
```
balazsbajorics, line 251: I guess min and max range could be a parameter, and then this wouldn't be special cased to the selection test

balazsbajorics (approved):

performance-test/README.md
```diff
@@ -0,0 +1,8 @@
+## How to run the performance tests locally
+
+1. npm install
+2. Create a project on localhost and copy its URL.
+3. run the performance test like this in the terminal: `EDITOR_URL=<project-url-you-just-copied> npm run`
+4. you will see console output with the frame numbers.
+5. for the plotly graph generation, you need a Plotly account username and password
+6. to upload images to S3, you need an S3 access key
```
Rheeseyb, line 8: Where / how do you provide these values to the test script?

performance-test/src/puppeteer-test.ts
```diff
@@ -9,8 +9,18 @@ const yn = require('yn')
 
 const BRANCH_NAME = process.env.BRANCH_NAME
 const PROJECT_ID = '5596ecdd'
-// const EDITOR_URL = `http://localhost:8000/p/39c427a7-hypnotic-king/`
-const EDITOR_URL = `https://utopia.pizza/project/${PROJECT_ID}/?branch_name=${BRANCH_NAME}`
+const EDITOR_URL =
+  process.env.EDITOR_URL ?? `https://utopia.pizza/project/${PROJECT_ID}/?branch_name=${BRANCH_NAME}`
```
Rheeseyb, line 13: This variable should support calling the staging url without a branch too

performance-test/src/puppeteer-test.ts
```diff
@@ -47,33 +64,123 @@ export const testScrollingPerformance = async function () {
...
   await page.tracing.stop()
   await browser.close()
   let traceData = fs.readFileSync('trace.json').toString()
   const traceJson = JSON.parse(traceData)
 
+  return getFrameData(traceJson, 'scroll_step_')
+}
+
+export const testResizePerformance = async function (): Promise<FrameResult> {
+  const { page, browser } = await setupBrowser()
```
Rheeseyb, line 131: All of these tests need to be wrapped in a `try` `finally` block to ensure that `browser.close()` is still called if they throw an error. I know Omar made that change locally last week whilst I was pairing up with him, but it looks like it wasn't merged.

Rheeseyb (approved):
I've made a couple of comments that should be easy to fix up. Once they're done this looks good to me.

seanparsons (approved):

performance-test/src/puppeteer-test.ts
maltenuhn: I just realised what this does. Not a fan of axis cutting in public, we should _always_ start at zero.

If we use the box plot, we'd automatically dimension the x-axis (showing the values) anyway, and there's an option in the graphing library to hide true outliers.

balazsbajorics: I agree with starting from 0, I think there's a benefit to manually cutting on the 'right' side: it makes it easier to compare a test run with a one month old one, since you don't need to reinterpret what you see based on markings on the x axis

performance-test/src/puppeteer-test.ts
enidemi: fixed

performance-test/src/puppeteer-test.ts
```diff
@@ -25,19 +46,27 @@ function defer() {
   return promise
 }
 
-function consoleDoneMessage(page: puppeteer.Page) {
-  return new Promise<void>((resolve, reject) => {
+function consoleDoneMessage(
+  page: puppeteer.Page,
+  expectedConsoleMessage: string,
+  errorMessage?: string,
+) {
```
balazsbajorics, line 53: I know it's not new code, but could you put a return value on the top level functions to bring it in line with the editor coding style?

balazsbajorics:
I think the selection test should be on a separate box chart, because it's so wildly slower

PR #847, opened 2021-01-13
fixing multiselect dragging

balazsbajorics:
you can try me here: http://utopia.pizza/p/25647216-cypress-emperor/?branch_name=fix/multiselect-dragging

**Problem:**
Click to drag on an element that was multiselected accidentally led the canvas controls thinking only that one element was selected. This rendered an incorrect selection outline, and applied the wrong snapping.

**Fix:**
It's a one liner basically, we already had code to check if an element is already selected or not, but it was not applied for the local-to-canvas-controls localSelectedViews state.


seanparsons (approved):

Rheeseyb (approved):
This is my favourite kind of PR :)

PR #848, opened 2021-01-13
Spike Capture the text between nodes when parsing

Rheeseyb:
The overall approach here is to diff the result of `node.getText()` with `node.getFullText()` when parsing nodes at each level, and use that to either create a new `ArbitraryJSBlock` or include it in an existing one. By doing so we're able to capture blank lines as well as comments.

The implementation here is (obviously) extremely hacky in places.


Rheeseyb:
Closing as the results of this spike have now been implemented in #861

PR #852, opened 2021-01-13
fix(canvas) remove storyboard filtering to allow direct canvas selection

enidemi:
Fixes a bug where it is not possible to select elements that are directly inserted to the canvas (without Scenes).

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -555,9 +555,7 @@ export const MetadataUtils = {
     return Utils.flatMapArray(getChildrenInner, immediateChildren)
   },
   getAllScenePaths(scenes: ComponentMetadata[]): ScenePath[] {
-    return scenes
-      .map((s) => s.scenePath)
-      .filter((s) => !TP.pathsEqual(s, EmptyScenePathForStoryboard))
+    return scenes.map((s) => s.scenePath)
   },
   getCanvasRootScenesAndElements(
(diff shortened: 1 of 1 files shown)


balazsbajorics (approved):

seanparsons (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -555,9 +555,7 @@ export const MetadataUtils = {
     return Utils.flatMapArray(getChildrenInner, immediateChildren)
   },
   getAllScenePaths(scenes: ComponentMetadata[]): ScenePath[] {
-    return scenes
-      .map((s) => s.scenePath)
-      .filter((s) => !TP.pathsEqual(s, EmptyScenePathForStoryboard))
+    return scenes.map((s) => s.scenePath)
```
Rheeseyb, line 558: This feels like this is going to introduce a regression somewhere else. The empty scene path is quite literally an empty path (https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/model/scene-utils.ts#L46), so if we were previously filtering that as part of this function it probably means we need two versions of the function - one where we want to include that path and one where we explicitly don't. Have you checked that this change makes sense in all of the places where this function is used?

balazsbajorics: that's a good shout. I split it into two functions – one that keeps the old name and functionality, and a new one for our new use case

Rheeseyb (approved):

PR #853, opened 2021-01-14
Handle comments inside arrays and objects of attributes.

### LINKED ISSUE #844: Capture comments in and around attributes.
With some code like this:

```
<div style={{backgroundColor: 'red' /* Not blue */}} />
```

We want it to be the case that these comments aren't lost.

seanparsons:
Fixes #844

**Problem:**
Currently comments held within arrays and objects used as attribute values are lost during the parse-print cycle.

**Fix:**
The rest of the comment handling has been added for array and objects, along with a field to capture comments around keys of an object.

**Notes:**
- Turns out that array and object separators are actually a nightmare for comments, as not only do the comments sit against a comma token that has to be searched for. However it's not readily apparent how to reinsert those, as adding them to the property assignment just results in them being invisibly lost.
- There is some test cases for comments between attributes, as those came up looking at the results of the analysis script.

**Commit Details:**
- Fix #844.
- Added `keyComments` to `JSXPropertyAssignment` to capture comments around
  the key.
- Implemented `simplifyAttributeIfPossible` to handle trying to create the
  most minimal case of a given attribute taking into account both the
  attribute types and if comments are present.
- Added `isParsedCommentsEmpty` to cover the common case of that check.
- `parseArrayLiteralExpression` and `parseObjectLiteralExpression` now do
  more carrying over of comments.
- Handling of simplifying the model removed from `parseArrayLiteralExpression`
  and `parseObjectLiteralExpression`.
- `jsxAttributeToExpression` adds in the key comments for a given object property.
- Added snapshot test for the comments test code.
- Added additional test cases which are disabled for now.


editor/src/core/shared/element-template.tsx
```diff
@@ -320,6 +328,89 @@ export function clearJSXAttributeOtherJavaScriptUniqueIDs(
...
 
+export function simplifyAttributeIfPossible(attribute: JSXAttribute): JSXAttribute {
+  switch (attribute.type) {
+    case 'ATTRIBUTE_VALUE':
+    case 'ATTRIBUTE_OTHER_JAVASCRIPT':
+    case 'ATTRIBUTE_FUNCTION_CALL':
+      return attribute
+    case 'ATTRIBUTE_NESTED_ARRAY':
+      let simpleArray: Array<unknown> = []
+      let notSoSimpleArray: Array<JSXArrayElement> = []
```
Rheeseyb, line 339: :grinning:

editor/src/core/shared/element-template.tsx
```diff
@@ -320,6 +328,89 @@ export function clearJSXAttributeOtherJavaScriptUniqueIDs(
...
+            notSoSimpleArray.push(jsxArraySpread(simplifiedAttribute, elem.comments))
+            break
+          case 'ARRAY_VALUE':
+            notSoSimpleArray.push(jsxArrayValue(simplifiedAttribute, elem.comments))
+            break
+          default:
+            const _exhaustiveCheck: never = elem
+            throw new Error(`Unhandled elem ${JSON.stringify(elem)}`)
+        }
+        if (isParsedCommentsEmpty(elem.comments) && isJSXAttributeValue(simplifiedAttribute)) {
```
Rheeseyb, line 354: Very minor, but we could skip this if `isSimpleArray` is false

editor/src/core/shared/element-template.tsx
```diff
@@ -320,6 +328,89 @@ export function clearJSXAttributeOtherJavaScriptUniqueIDs(
...
+      let simpleObject: MapLike<unknown> = {}
+      let notSoSimpleObject: Array<JSXProperty> = []
+      let isSimpleObject: boolean = true
+      for (const elem of attribute.content) {
+        const simplifiedAttribute = simplifyAttributeIfPossible(elem.value)
+        switch (elem.type) {
+          case 'SPREAD_ASSIGNMENT': {
+            const noComments = isParsedCommentsEmpty(elem.comments)
+            notSoSimpleObject.push(jsxSpreadAssignment(simplifiedAttribute, elem.comments))
+            if (isJSXAttributeValue(simplifiedAttribute) && noComments) {
```
Rheeseyb, line 375: Same here if `isSimpleObject` is false

Rheeseyb (approved):

editor/src/core/shared/element-template.tsx
seanparsons: Changed in 6fafae98.

editor/src/core/shared/element-template.tsx
seanparsons: Changed in 6fafae98.

balazsbajorics (approved):

Issue #851, opened 2021-01-13
Use node.getFullText to capture text (whitespace and comments) between nodes during parsing


Rheeseyb:
This is for the implementation of the spike from https://github.com/concrete-utopia/utopia/pull/848

To fix this up we need to:
- Remove the hack used to capture the text prefixing a return statement in `parseOutFunctionContents`
- Finish applying the approach inside `parseJSXArbitraryBlock`
- Continue to use the previous style of comment capturing for all other JSX
- Fix up the tests

This is currently blocked on #850 as it will move `/** @jsx jsx */` after import statements

PR #854, opened 2021-01-14
Use raw import and export statements where possible

### LINKED ISSUE #850: Preserve arbitrary blocks interspersed amongst import and export statements
Currently if an arbitrary piece of code appears between import or export statements it will be printed in a different place because of our printing logic for import and export statements

Rheeseyb:
Fixes #850 

**Problem:**
We parse the import and export statements separately, and then convert those parsed objects back into statements when printing the model. Because of this, all imports will be printed together (regardless of where they exist in the code), as will all export statements, meaning any other statements mixed in with those will be moved outside of them. 

**Fix:**
Include the raw import and export statements in the `TopLevelElement` type, meaning we can print them back out exactly as they were when first parsed. With exports we're always able to do that (as we don't automatically generate or modify the exports). With imports, because we insert new import statements as the result of some insert actions, we now have to reconcile the `Imports` object with the raw statements to determine if new imports were added, and if they were we print them after the last import statement.

**Commit Details:**
- Introduced the types `ImportStatement` and `UnparsedCode`, which are included in the union type `TopLevelElement`
- `ImportStatement` captures the raw code, as well as the module name and the imported symbols for the sake of reconcilliation
- Push these new elements to the top level elements captured when parsing
- When printing the `Imports` object, check each import to see if it already exists in a one of the captured `ImportStatement`s
- When printing, find the last `ImportStatement`, and use this as the point after which to insert new import statements
- When printing an `ImportStatement` or `UnparsedCode`, we create a `TS.UnparsedSourceFile` using the raw code to ensure we print it back out exactly as it was when parsed
- Fixed up a whole bunch of tests that were reliant on a specific indexed `TopLevelElement`
- Fixed other tests using `expect.arrayContaining` to replace explicitly created `topLevelElements`


balazsbajorics (commented):
what is the difference between a ArbitraryJSBlock and an UnparsedCode? I thought we already put every unparsed code into an ArbitraryJSBlock - I guess I was wrong.

Does the introduction of UnparsedCode help us preserve some more whitespace and newlines etc?

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -685,7 +697,10 @@ function createBlanklineTracker(): (statement: TS.Node, printedCode: string) =>
   let lastLineWasVariable = false
   let lastLineWasMultiLineStatement = false
 
-  return (statement: TS.Node, printedCode: string): boolean => {
+  return (statement: TS.Node, rawPrintedCode: string): boolean => {
+    // Awful hack until we implement #851
+    const printedCode = replaceAll(rawPrintedCode, '/** @jsx jsx */', '').trim()
```
balazsbajorics, line 702: what behavior are we altering here? is this to prevent a newline below the pragma? or was this tripping up the test cases and throwing an error?

Rheeseyb:
> what is the difference between a ArbitraryJSBlock and an UnparsedCode? I thought we already put every unparsed code into an ArbitraryJSBlock - I guess I was wrong.
> 
> Does the introduction of UnparsedCode help us preserve some more whitespace and newlines etc?

We actually parse an `ArbitraryJSBlock` (because they might contain elements within), which also means we need to perform some Babel transpiling to insert UIDs. We have no need to do any of that with this code since it has already been parsed separately, so we're basically flagging it as code that we just don't need to interpret or care about.

editor/src/core/workers/parser-printer/parser-printer.ts
Rheeseyb: Yeah this is because of the blank line insertion implemented in #840 that is about to get binned anyway as part of #851 (which I'm working on right now). It was mainly because this was tripping test cases that weren't expecting the extra blank line it was adding into the code.

balazsbajorics (approved):

balazsbajorics:
nicely done!

seanparsons (approved):

PR #856, opened 2021-01-15
Fix/canvas reparent orphans

balazsbajorics:
**Problem:**
We want to be able to reparent elements directly to the canvas. These elements become the children of the Storyboard element, without existing inside a component. These are called Orphans. We had a half-baked support for Orphans for the last year, and we made some significant regressions in the last six months, and most recently, last week.

**Fix:**
Allow Orphans to be reparented to - from the Canvas root. Selecting them is already supported on master.

**Commit Details:**
- Added data-utopia-valid-paths to the Storyboard, containing orphan paths – this allows the dom-based select mode hooks to find the elements
- The reparent target lookup now uses the DOM, which meant that now we need to feed the _transient_ canvas state's selected views to the reparent functions. (because the transient canvas state is rendered to the DOM, that's what the dom lookup will see, that's what we need to filter / validate against)

### DIFF
--- a/editor/src/core/shared/template-path.ts
+++ b/editor/src/core/shared/template-path.ts
@@ -149,17 +149,6 @@ export function toString(target: TemplatePath): string {
   }
 }
 
-export function toStringOptional(target: TemplatePath | null | undefined): string | null {
-  if (target == null) {
-    return null
-  }
-  if (isScenePath(target)) {
-    return scenePathToString(target)
-  } else {
-    return instancePathToString(target)
-  }
-}
-
 function newScenePath(elements: StaticElementPath): ScenePath {
   return {
     type: 'scenepath',
--- a/editor/src/core/shared/template-path.ts
+++ b/editor/src/core/shared/template-path.ts
@@ -149,6 +149,17 @@ export function toString(target: TemplatePath): string {
   }
 }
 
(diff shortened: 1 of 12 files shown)


editor/src/core/shared/template-path.ts
```diff
@@ -149,6 +149,17 @@ export function toString(target: TemplatePath): string {
   }
 }
 
+export function toStringOptional(target: TemplatePath | null | undefined): string | null {
```
balazsbajorics, line 152: unfortunately I couldn't extend toString to be able to return an optional null without having to change too much irrelevant code – so I made a second variant of the function instead

seanparsons: I'd argue for using `optionalMap` with the regular `toString` function personally, but this doesn't appear to be being used?

seanparsons (approved):

editor/src/core/shared/template-path.ts
balazsbajorics: I meant to write that I wrote this for debugging convenience, not used in the code. I didn't think about optionalMap, I guess I can just remove this then!

balazsbajorics: removed!

enidemi (approved):

PR #860, opened 2021-01-18
Re-enable zero-sized elements selection

balazsbajorics:
**Problem:**
We lost the ability to select zero-sized elements (using a 5px threshold) when I implemented the DOM-based canvas selection. Unfortunately since these elements have one or two missing dimensions, we cannot rely on the DOM apis to be able to select them.

**Fix:**
Resurrect `Canvas.getAllTargetsAtPoint`. `Canvas.getAllTargetsAtPoint` is the primary source of information and ordering of elements. We use the list from `document.elementsFromPoint` to filter out elements which are _not_ under the mouse, unless they have zero size in either dimension.

**Commit Details:**
- Resurrect `Canvas.getAllTargetsAtPoint`.
- make `dom-lookup@getValidTargetAtPoint` use `dom-lookup@getAllTargetsAtPoint`
- make `dom-lookup@getAllTargetsAtPoint` cross-reference `Canvas.getAllTargetsAtPoint` and `document.elementsFromPoint`.

### DIFF
--- a/editor/src/components/canvas/dom-lookup.ts
+++ b/editor/src/components/canvas/dom-lookup.ts
@@ -1,6 +1,7 @@
 import * as R from 'ramda'
 import { last, stripNulls } from '../../core/shared/array-utils'
 import { getDOMAttribute } from '../../core/shared/dom-utils'
+import { JSXMetadata } from '../../core/shared/element-template'
 import {
   canvasPoint,
   CanvasVector,
@@ -13,6 +14,7 @@ import {
 } from '../../core/shared/math-utils'
 import { TemplatePath } from '../../core/shared/project-file-types'
 import * as TP from '../../core/shared/template-path'
+import Canvas, { TargetSearchType } from './canvas'
 import { CanvasPositions } from './canvas-types'
 
 export function findParentSceneValidPaths(target: Element): Array<string> | null {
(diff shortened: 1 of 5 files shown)


enidemi (approved):

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -50,39 +55,76 @@ export function findFirstParentWithValidUID(
...
+        return null
       }
-      return null
     }),
   )
+
+  return getElementsUnderPointFromAABB
+    .filter((foundElement) => {
+      if (!foundElement.canBeFilteredOut) {
+        return true
```
Rheeseyb, line 122: I'm finding this very hard to read because of the combination of `filter` removing elements that return `false` combined with the `!foundElement.canBeFilteredOut` returning `true` inside the `if` statement :confused:

Rheeseyb (commented):
Doesn't this effectively undo the benefits of the DOM based selection, and revert us back to using AABB selection?

balazsbajorics:
> Doesn't this effectively undo the benefits of the DOM based selection, and revert us back to using AABB selection?

The problem I had to solve is how to interleave the zero-sized elements with the results of the DOM query. I think using the AABB query as the source of truth wrt the ordering of the elements and then culling the list using the DOM query was the easiest solution.

Since the AABB can never be smaller than the actual bounds of an element, we can say that the results of the AABB-based query will always contain a strict superset of the dom-based query. Thus if we filter the AABB list using the DOM list (except for zero sized elements), we effectively have the results from the DOM-based query, ordered by the AABB query, interspersed  with the zero-sized elements.

Rheeseyb (approved):
> Since the AABB can never be smaller than the actual bounds of an element

I was mistakenly under the impression that our AABB calculations weren't taking into account the result of transforming the element for some reason. I think my memory was stuck in a very old (pre-dom-walker) version of the editor. I just double checked and it indeed does use the result of transforms, so I'm an idiot :smile:

PR #861, opened 2021-01-18
Capture raw code between nodes

### LINKED ISSUE #851: Use node.getFullText to capture text (whitespace and comments) between nodes during parsing

Rheeseyb:
Fixes #851 

**Problem:**
We wanted a way to capture the whitespace and comments between parsed nodes so that we could retain that when printing the code back out after making changes.

**Fix:**
We can capture this by comparing a `TS.Node`'s `getText` with its `getFullText`, since the former will strip off any leading whitespace and comments whereas the latter won't. At the top level of the file we'll use the result of this to introduce `UnparsedCode` elements to capture this. When parsing a component, we use the `getFullText` to capture the arbitrary code at the start, and then also use it to capture the prefix to a return statement. Inside JSX expressions we just use `getFullText` everywhere.

**Commits:**
- Added a function `extractPrefixedCode` for capturing the raw lines of code leading up to a TS Node
- In `parser-printer.ts`:
  - Before parsing each top level node we use the above to capture the code before the node, and push it to our array of `TopLevelElement`s as an `UnparsedCode` element
  - Removed any explicit capturing of comments at the top level because that functionality is now covered in the above
  - Rather than collecting arbitrary nodes before calling `applyAndResetArbitraryNodes` (which triggers the parsing of them all) at specific instances, I have rolled that functionality into the `pushArbitraryNode` function, meaning that we're no longer trying to merge them together, since that was never actually happening anymore anyway. This fixed a bug around the order of `UnparsedCode` elements and `ArbitraryJSBlock` elements in a very specific case caught by an existing test.
  - Removed heuristic blank line printing since that is no longer required.
  - Added an option for explicitly inserting line breaks between statements when printing, specifically for use in tests that were defining the parsed objects (and so wouldn't have been including the leading whitespace)
  - When printing JSX Arbitrary blocks we do a bit of a switcharoo, creating a TS source file, extracting and updating what we need, printing the result to a string, and then creating a `JsxText` element that wraps the string in braces. This allows us to directly craft the resulting block of code and insert it into the AST, since the `JsxText` element will be printed as is.
  - Removed now unneeded calls to a seemingly defunct library, which was generating the factory code for creating TS AST nodes, since we no longer need it and can instead create a TS source file and directly use that
- `parseOtherJavaScript` now includes a `tailingCode` param to allow us to capture the lines before a `return` statement when parsing a component (since the return statement is handled by separate logic for parsing out the elements)
- `parseArbitraryNodes` provides the option to use either the full text (which you'd want in most cases) or the regular text (which we use when parsing top level nodes, since we'll have already captured the preceding lines as `UnparsedCode`)
- All of these changes broke a lot of tests (obviously), since e.g. a lot of them were crafting the parsed objects directly. To get around this, I updated some of the existing test helper functions (which were focused on clearing UIDs from the code) to also trim the whitespace around code blocks and remove empty code blocks (since the alternative would have been to manually add that whitespace in to many, *many* tests)


seanparsons (approved):

editor/src/core/workers/parser-printer/parser-printer-comments.spec.ts
```diff
@@ -209,18 +149,19 @@ describe('Parsing and printing code with comments', () => {
...
+
         // Comment before return statement
         return (
           <div /* Comment before all attributes */
             data-uid='aaa' /* Comment inside attributes */
             someProp={/* Comment at start of JSX attribute */ 1000 /* Comment at end of JSX attribute */}
             someProp2={{
               /* Comment before object key */ someKey /* Comment after object key */:
-                /* Comment before object value */ 'someValue' /* Comment after object value */,
+                /* Comment before object value */ 'someValue' /* Comment after object separator */ /* Comment after object value */,
```
Rheeseyb, line 160: The comments after an object or array separator is moved in front of the comment after the object or array value by prettier for some reason.

editor/src/core/workers/parser-printer/parser-printer-comments.spec.ts
```diff
@@ -209,18 +149,19 @@ describe('Parsing and printing code with comments', () => {
...
               someKey2: 'someValue2',
             }}
             someProp3={[
-              /* Comment before array value */ 100 /* Comment after array value */,
+              /* Comment before array value */ 100 /* Comment after array separator */ /* Comment after array value */,
```
balazsbajorics, line 164: `/* Comment after array separator */` is before the array separator here

editor/src/core/workers/parser-printer/parser-printer-comments.spec.ts
balazsbajorics: oh I missed your comment when I wrote mine 1 minute ago :) 

eek! PR to prettier? ;)

editor/src/core/workers/parser-printer/parser-printer-comments.spec.ts
Rheeseyb: Ha please see my comment above :grinning:

editor/src/core/workers/parser-printer/parser-printer.spec.ts
```diff
@@ -4566,9 +4570,8 @@ export var App = props => {
 
     const position = consumer.getOriginalPosition(transpiledLine, transpiledCharacter)
 
-    // TODO BALAZS we should test that the code is in col 9, but I just couldn't make it work :(
-    // expect(position).toEqual(expect.objectContaining({ line: 22, column: 9 }))
-    expect(position).toEqual(expect.objectContaining({ line: 16 }))
+    // FIXME This should be line 16 char 9 but I cannot get this to work
+    expect(position).toEqual(expect.objectContaining({ line: 17, column: 10 }))
```
balazsbajorics, line 4574: btw what is wrong with this test? :D

Rheeseyb: Honestly I have no idea and gave up at this point trying to fix it. I tested inserting an error in that position in the actual editor and the error overlay mapped to the correct line and character so :shrug:

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -111,19 +111,19 @@ import {
   isExported,
   isDefaultExport,
   expressionTypeForExpression,
+  extractPrefixedCode,
 } from './parser-printer-parsing'
 import { getBoundsOfNodes, guaranteeUniqueUidsFromTopLevel } from './parser-printer-utils'
 import { ParserPrinterResultMessage } from './parser-printer-worker'
-import creator from './ts-creator'
```
balazsbajorics, line 117: does this mean Sean doesn't have to sacrifice a goat once every week anymore?

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -195,20 +213,10 @@ function jsxAttributeToExpression(attribute: JSXAttribute): TS.Expression {
         })
         return TS.createArrayLiteral(arrayExpressions)
       case 'ATTRIBUTE_OTHER_JAVASCRIPT':
-        // SP: This is quite truly the most spectacular fudge I've ever had to create in my entire career.
-        // Creates a representation of the AST that is code which executes the TS factory functions...
-        // ...Then evals it to produce the AST and drills into the result a little.
```
balazsbajorics, line 200: can we please save this to the Utopian Archives for posterity? :)

balazsbajorics (approved):

editor/src/core/workers/parser-printer/parser-printer.ts
Rheeseyb: It indeed means exactly that! This dependency will go in the bin. I didn't remove it yet for fear of merge conflicts with Sean's ongoing work, but that has been merged to master now so I will remove it as part of this PR

PR #862, opened 2021-01-19
fix: do not change insertion target while dragging

balazsbajorics:
**Problem:**
During drag-to-insert, if the mouse left the target parent and went over a new element, we changed the highlight to the new element, and we thus changed the target parent to the new element. 

**Fix:**
Extend `maybeHighlightOnHover` to not switch highlight targets during an insertion drag.

**Commit Details:**
- added `!isInserting` clause to `maybeHighlightOnHover`


enidemi (approved):

Rheeseyb (approved):
If this file is now being used during insertion mode we should definitely rename it. Other than that, the code looks fine to me

seanparsons (approved):

PR #864, opened 2021-01-19
introduce jest-clean-console-reporter

balazsbajorics:
**Problem:**
Running `npm run test` produced 80,000 lines of console output, making it impossible to query the logs from GitHub.

**Fix:**
Introducing jest-clean-console-reporter, which helped me categorize most of those console output into a few groups. This reduces the entire console output to less than 2000 lines!


Rheeseyb (approved):
Thank you!

enidemi (approved):

PR #865, opened 2021-01-19
Feature/canvas contextmenu submenu

enidemi:
Extending the canvas context menu to have a submenu listing all elements under the cursor. Clicking on the element list item changes the selection.
- updated the css file, this should have been in the previous PR as it was changed with the 5.0.0 update
- creating the element context menu adds the array of all submenu element items with an extra detail about their templatepaths
- when opening the menu the elements under the cursor array is passed in as a prop to the contextmenu open
- the new hidden prop on contextmenu items help with hiding the non-matching elements, with comparing the templatepaths
- `MetadataUtils.getAllPaths` had some duplicated values since the orphan changes

### DIFF
--- a/editor/resources/editor/ReactContexify.css
+++ b/editor/resources/editor/ReactContexify.css
@@ -11,7 +11,8 @@
 }
 .react-contexify .react-contexify__submenu {
   position: absolute;
-  top: 0;
+  /* negate padding */
+  top: -4px;
   pointer-events: none;
   transition: opacity 0.275s;
 }
@@ -24,7 +25,6 @@
 }
 
 .react-contexify__separator {
-  float: left;
   width: 100%;
   height: 1px;
   cursor: default;
@@ -36,6 +36,7 @@
   font-size: 10px;
   font-weight: 500; /* medium */
   cursor: default;
+  position: relative;
 }
 .react-contexify__item:not(.react-contexify__item--disabled):hover
   > .react-contexify__item__content {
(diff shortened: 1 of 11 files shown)


editor/resources/editor/ReactContexify.css
```diff
@@ -142,42 +123,71 @@
...
-  }
-  60% {
-    transform: perspective(400px) rotate3d(1, 0, 0, 10deg);
+  to {
+    transform: perspective(800px);
   }
-  80% {
-    transform: perspective(400px) rotate3d(1, 0, 0, -5deg);
+}
+@keyframes react-contexify__flipOutX {
```
balazsbajorics, line 132: did you write all this new CSS? or rather, just updated this css file from the new version of react-contexify? I _think_ now we have better tools to include a css from node_modules, no need to copy paste the css to the editor. or maybe we only introduced css loaders for _our_ users? :)

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -78,17 +88,30 @@ export class MomentumContextMenu<T> extends ReactComponent<ContextMenuProps<T>>
     return splitItems
   }
 
+  isHidden = (
```
balazsbajorics, line 91: this feels to speceific to live in the generic MomentumContextMenu – isn't there a more specific place to put it for the canvas context menu?

editor/src/components/element-context-menu.tsx
```diff
@@ -58,6 +62,40 @@ const ElementContextMenuItems: Array<ContextMenuItem<CanvasData>> = [
   toggleShadowItem,
 ]
 
+function useCanvasContextMenuItems(
+  contextMenuInstance: ElementContextMenuInstance,
+  dispatch: EditorDispatch,
+): Array<ContextMenuItem<CanvasData>> {
+  const metadata = useEditorState(
```
balazsbajorics, line 69: does the context menu render (offscreen) when it is closed?

editor/resources/editor/ReactContexify.css
enidemi: this css file was changed to match the utopia style, the original contexify css is not the same

editor/src/components/editor/actions/actions.ts
```diff
@@ -3146,7 +3148,14 @@ export const UPDATE_FNS = {
   },
   SHOW_CONTEXT_MENU: (action: ShowContextMenu, editor: EditorModel): EditorModel => {
     // side effect!
-    openMenu(action.menuName, action.event)
+    if (action.menuName === 'context-menu-canvas') {
```
balazsbajorics, line 3151: I don't like this here. it either should be a separate special action, or a parameter in the action payload, it definitely should not switch behavior based on the `menuName` ID.

enidemi: fixed, I moved this to the action payload

editor/src/components/context-menu-wrapper.tsx
enidemi: this could be similar to how action callbacks are done for context menu, I'm not sure if that's better?

balazsbajorics: I think that sounds better – something like an optional `isHidden?: (props: ContextMenuInnerProps) => void` property on the `ContextMenuItem` would mean we only need to provide it for the Canvas context menu, so we could remove the specific hiding code from here

balazsbajorics (approved):

Rheeseyb (approved):

PR #866, opened 2021-01-19
prevent highlights when resize controls are hovered

balazsbajorics:
**Problem:**
The resize controls have a click catcher area that sticks out from the element. This is deliberate to make it easier for the user to grab them. In the previous setup, the resize control divs were simply above the highlight catcher divs, so we didn't have an issue with the user accidentally highlighting the parent element when they want to start a resize. However, since I moved the highlight logic to an `onMouseMove` handler, these divs were no longer preventing the highlight on the parent from happening.

**Fix:**
This is a rather unorthodox fix to prevent the highlights from happening. 
The resize controls are a children element of the `new-canvas-controls-container` div where the `onMouseMove` controlling the highlights is attached. Using this fact, I could catch the bubbling mouseMove event in the resize control divs, and append an extra property to the event object itself. This event object then bubbles to `new-canvas-controls-container`'s highlight handler hook, and if we see this extra property on the event, we treat it as a command to remove any highlights.

**Commit Details:**
- `ResizeControl` catches the mouseMove event and adds a new property on it with the key `PreventHighlightsForThisEvent` 
- the `onMouseMove` event handler inside `useHighlightCallbacks` checks for the existence of this special property. if the property is there, it bails early and calls `maybeClearHighlightsOnHoverEnd`

### DIFF
--- a/editor/src/components/canvas/controls/multiselect-resize-control.tsx
+++ b/editor/src/components/canvas/controls/multiselect-resize-control.tsx
@@ -172,6 +172,7 @@ export class MultiselectResizeControl extends React.Component<
               metadata={this.props.componentMetadata}
               onResizeStart={this.onResizeStart}
               testID={'component-resize-control-0'}
+              maybeClearHighlightsOnHoverEnd={this.props.maybeClearHighlightsOnHoverEnd}
             />
             {...guidelineElements}
           </>
@@ -224,6 +225,7 @@ export class SingleSelectResizeControls extends React.Component<SingleselectResi
               metadata={this.props.componentMetadata}
               onResizeStart={this.props.onResizeStart}
               testID={`component-resize-control-${TP.toComponentId(view)}-${index}`}
+              maybeClearHighlightsOnHoverEnd={this.props.maybeClearHighlightsOnHoverEnd}
             />
           </>
         )
--- a/editor/src/components/canvas/controls/new-canvas-controls.tsx
+++ b/editor/src/components/canvas/controls/new-canvas-controls.tsx
@@ -94,6 +94,7 @@ export interface ControlProps {
   cmdKeyPressed: boolean
   showAdditionalControls: boolean
   elementsThatRespectLayout: Array<TemplatePath>
+  maybeClearHighlightsOnHoverEnd: () => void
 }
 
 interface NewCanvasControlsProps {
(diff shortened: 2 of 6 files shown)


Rheeseyb (commented):
I'm confused by this solution. Why can't we `preventDefault` on the event in the resize box to prevent it from triggering the highlighting?

balazsbajorics:
> I'm confused by this solution. Why can't we `preventDefault` on the event in the resize box to prevent it from triggering the highlighting?

because that prevents detecting the mouse move on `window` which means it prevents the actual dragging from happening

balazsbajorics:
> > I'm confused by this solution. Why can't we `preventDefault` on the event in the resize box to prevent it from triggering the highlighting?
> 
> because that prevents detecting the mouse move on `window` which means it prevents the actual dragging from happening

@enidemi proposed what if I would put special `data-utopia-dont-highlight` attribute on the resize controls, then in `getAllTargetsAtPoint` if I encounter those attributes bail out.

what do you think @Rheeseyb  ? would that be a better solution? I can't see how one is better than the other, both of them have to rely on out-of-band communication. I guess Eni's solution is more declarative and doesn't rely on the event being caught and mutated

Rheeseyb:
> > > I'm confused by this solution. Why can't we `preventDefault` on the event in the resize box to prevent it from triggering the highlighting?
> > 
> > 
> > because that prevents detecting the mouse move on `window` which means it prevents the actual dragging from happening
> 
> @enidemi proposed what if I would put special `data-utopia-dont-highlight` attribute on the resize controls, then in `getAllTargetsAtPoint` if I encounter those attributes bail out.
> 
> what do you think @Rheeseyb ? would that be a better solution? I can't see how one is better than the other, both of them have to rely on out-of-band communication. I guess Eni's solution is more declarative and doesn't rely on the event being caught and mutated

For visibility - the two of us have had a conversation about having the drag handling logic (mentioned above) capture and process the event first, which would then allow the resize controls to stop any further bubbling of the event.

balazsbajorics:
@Rheeseyb I pushed the changes – it seems to work allright, no more event mutation involved. I had to remove one performance optimization but I measured it and it saved us less than a few microseconds, so all is fine

enidemi (approved):

Rheeseyb (approved):

seanparsons (approved):

PR #868, opened 2021-01-19
Fix/better invisible element handling

balazsbajorics:
**Problem:**
My previous PR https://github.com/concrete-utopia/utopia/pull/860 had to be reverted because it broke selection for out-of-bounds children and Orphans (elements inserted to the Storyboard root)

**Fix:**
This PR re-reverts #860, and fixes the two mentioned issues

**Commit Details:**
- Revert reverting #860 
- Fix selection of out-of-bounds children by flipping the `useBoundingFrames` flag to true when calling `getAllTargetsAtPoint`
- Give a zero-sized rectangle as `localFrame` and `globalFrame` for the Storyboard element instance metadata, this makes `getFramesInCanvasContext` be able to digest Orphans


editor/src/components/canvas/canvas.ts
```diff
@@ -131,16 +131,10 @@ const Canvas = {
...
-      } else {
-        const nonNullGlobalFrame = rootComponent.globalFrame
-        const rootElements = MetadataUtils.getImmediateChildren(metadata, rootComponent.scenePath)
-        return flatMapArray(
-          (root) => recurseChildren(nonNullGlobalFrame, root).frames,
-          rootElements,
-        )
-      }
+      const nonNullGlobalOffset: CanvasVector =
+        rootComponent.globalFrame ?? ({ x: 0, y: 0 } as CanvasVector)
```
balazsbajorics, line 135: I spent a bit of time trying to fix the actual `globalFrame` of the Storyboard `rootComponent`, but the potential fix involves touching quite a bit of the scene/storyboard hacky code in `useGetStoryboardRoot` https://github.com/concrete-utopia/utopia/blob/4d7507afd73dfb0c7175befa7d286318b284327a/editor/src/components/canvas/ui-jsx-canvas.tsx#L416-L424

so instead I just hacked a fallback in here and added a TODO there.

I think it's not worth touching as the potential risk for regressions is too large, and I don't mind this 0;0 vector as a fallback value here

editor/src/components/canvas/dom-walker.ts
```diff
@@ -436,8 +437,8 @@ export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElem
             canvasRootPath as InstancePath,
             left('Storyboard'),
             {},
-            null,
-            null,
+            { x: 0, y: 0, width: 0, height: 0 } as CanvasRectangle,
```
balazsbajorics, line 440: x and y are correct - they are indeed 0 for the canvas root. width and height could be infinity or NaN, but 0 felt appropriate as well – no code should rely on it anyways

enidemi (approved):

Rheeseyb (approved):

PR #869, opened 2021-01-19
Chore: run prettier --check on CI

balazsbajorics:
**Problem:**
We broke our pre-commit hook that was supposed to run prettier and we didn't notice it for a few days. This led me to realize that if someone uses `git commit --no-verify` they can sneak unprettified code into our codebase.

**Fix:**
Run `prettier --check` as part of running the tests on Github CI. This fails the pull request that attempts to add unprettified code to our codebase.

**Commit Details:**
- The meat of the PR is in `package.json`
- Added new npm script called `prettier-check`, running this as part of `check` and `check-ci`
- Fixing the pre-existing prettier issues in our codebase so this PR can pass (sorry for the ugly diff – feel free to ignore anything that is not `package.json`, it was auto-written by Prettier, I pinky swear I didn't alter the files :) )


editor/package.json
```diff
@@ -12,8 +12,8 @@
   },
   "scripts": {
     "clean": "rm -rf lib/",
-    "check": "tsc --noEmit && npm run lint && npm run check-dependency-cruiser && npm run test",
-    "check-ci": "tsc --noEmit && npm run lint && npm run check-dependency-cruiser && npm run test-ci",
+    "check": "tsc --noEmit && npm run lint && npm run check-dependency-cruiser && npm run prettier-check && npm run test",
```
balazsbajorics, line 15: I'm very confident that `check-ci` should run `prettier-check`, I'm less sure about `check` here. Does anyone have a workflow that would be hurt by this inclusion here? maybe @seanparsons  ?

enidemi (approved):

editor/package.json
Rheeseyb: `check` should absolutely be running the same checks that `check-ci` does

Rheeseyb (approved):

Rheeseyb:
What actually broke the pre-commit hook?

Rheeseyb:
> What actually broke the pre-commit hook?

After writing this I immediately switched tabs to #870 and found the answer there :D

seanparsons (approved):

PR #870, opened 2021-01-20
remove precommit hook from performance test

enidemi:
The issue is that since adding the precommit hook for the performance test the editor precommit hook is not doing anything.  Because the editor is more important I'm removing the performance test hook.

Next step is to set up lint-staged to work on both projects. According to the documentation it should be added to the git root folder.

### DIFF
--- a/performance-test/package-lock.json
+++ b/performance-test/package-lock.json
@@ -4,60 +4,10 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):
we might need to pull the plug on a `lerna` setup

Rheeseyb (approved):

seanparsons (approved):

PR #871, opened 2021-01-20
Fix: recalculate highlighted views when mode changes or cmd toggles

balazsbajorics:
**Problem:**
When the user presses or releases cmd, but does not move the mouse, we left the (outdated) highlights on the screen. When the user enters insert mode, they had to move the mouse a bit to select the target parent, instead of it being already selected when entering insert mode.

**Fix:**
Add a new useEffect which recalculates (and updates) the highlighted views if the active mode changes or if the cmd is toggled.

**Commit Details:**
- add new flags `active` and `cmdPressed` to the `useSelectAndHover` hooks
- moved highlight logic to `calculateHighlightedViews`
- it is called from `onMouseMove` just like before
- it is also called from a useEffect when `active` or `cmdPressed` changes, triggering a refresh of what is highlighted

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
@@ -13,6 +13,7 @@ import {
 import { ScenePath, TemplatePath } from '../../../../core/shared/project-file-types'
 import * as TP from '../../../../core/shared/template-path'
 import { fastForEach, NO_OP } from '../../../../core/shared/utils'
+import { WindowMousePositionRaw } from '../../../../templates/editor-canvas'
 import { KeysPressed } from '../../../../utils/keyboard'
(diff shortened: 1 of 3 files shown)


enidemi (approved):

Rheeseyb (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
```diff
@@ -13,6 +13,7 @@ import {
 import { ScenePath, TemplatePath } from '../../../../core/shared/project-file-types'
 import * as TP from '../../../../core/shared/template-path'
 import { fastForEach, NO_OP } from '../../../../core/shared/utils'
+import { WindowMousePositionRaw } from '../../../../templates/editor-canvas'
```
seanparsons, line 16: Be warned, this is the route of some _really_ big circular reference problems.

seanparsons (approved):

PR #879, opened 2021-01-22
tests(canvas-utils) added tests for orphans, reparent and insert

enidemi:
Added 3 missing tests for the canvas selection changes. These tests are all for orphan elements.

- reparent from the canvas root
- reparent to the canvas root
- insert an element to the canvas root

### DIFF
--- a/editor/src/components/canvas/canvas-utils.spec.browser.ts
+++ b/editor/src/components/canvas/canvas-utils.spec.browser.ts
@@ -20,6 +20,9 @@ import {
 } from './canvas-types'
 import { wait } from '../../utils/test-utils'
 import { CanvasControlsContainerID } from './controls/new-canvas-controls'
+import { PrettierConfig } from '../../core/workers/parser-printer/prettier-utils'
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/canvas-utils.spec.browser.ts
```diff
@@ -1401,6 +1404,336 @@ describe('moveTemplate', () => {
...
+          clientX: areaControlBounds.left - 5,
+          clientY: areaControlBounds.top + 5,
+          buttons: 1,
+        }),
+      )
+      await domFinished
+      await dispatchDone
+    })
+
+    await act(async () => {
```
balazsbajorics, line 1494: is this the same mouse move event as above? do we need two of these?

enidemi: I've seen this for other reparent tests, when I tried it with only one mousemove event the reparent is not happening and the element stays where it was.

balazsbajorics: okay :)

balazsbajorics (approved):

Rheeseyb (approved):

PR #886, opened 2021-01-25
Chore: fix illegal react setstate canvas console updates

balazsbajorics:
**Problem:**
The ui-jsx-canvas component was calling setStates that lived in EditorComponent. These setstates triggered a re-render of EditorComponent. This caused two problems: 
1. React complaining endlessly about this being _illegal_.
2. if any one of us put a console log inside any of the components re-rendered by the rerendering EditorComponent, it would trigger a rerender of the EditorComponent, potentially trigger a rerender in the component triggering the rerender... creating an infinite feedback loop.

**Fix:**
Instead of `canvasConsoleLogs` and `runtimeErrors` living in EditorComponent as state, I created write-only setters for it that are used in `ui-jsx-canvas` (but these are write only, so they can no way trigger a re-render loop). 
And instead of reading the values as state in `ErrorOverlayComponent` and `CodeEditor` – I created read-only getters that are called asynchronously (as part of the event loop, think `setTimeout(0)`) so they cannot trigger a cascading rerender preventing React from being able to complain.

To achieve this, I added an ancient library called [PubSub](https://www.npmjs.com/package/pubsub-js). This is a very simple publish-subscribe system but it let me avoid having to create an `EventEmitter` or send horrible message events on the window.
I created a new construct (`atomWithPubSub`) that mimics Recoil Atoms, you can make them, subscribe to their value being changed as a callback, access them read only, write only, or read and write mode. The update always happens via PubSub, so they are not for generic usage, however I use them for two atoms in this PR, so it was worth refactoring the read and write code into helper functions instead of having them duplicated.

Finally, if the iframe code editor is enabled (and it should be!) we are not even triggering a react rerender: we use the write-only hooks to update the values, and use the callback to send messages via the iframe Code Editor Bridge to the code editor directly, without having to update its wrapper component.

**Commit Details:**
- Added PubSub
- Removed the `canvasConsoleLogs` and `runtimeErrors` hooks from `EditorComponent`, moving them to a dedicated file
- instead of passing the setState function across the tree hierarchy, use the write-only hooks directly in `CanvasComponentEntry`
- Instead of passing the errors and logs values down across the tree hierarchy, use the read-only hooks directly in `ErrorOverlayComponent` and `CodeEditorNonIframeEntryPoint`
- Instead of triggering re-renders in `CodeEditorWrapper` , I use the new subscription-based callbacks in `CodeEditorIframeContainer` passing the values directly to `useBridgeTowardsIframe`


enidemi (approved):

editor/package.json
```diff
@@ -145,6 +145,7 @@
     "platform-detect": "3.0.0",
     "prettier": "2.0.5",
     "prop-types": "15.6.2",
+    "pubsub-js": "^1.9.2",
```
Rheeseyb, line 148:
```suggestion
    "pubsub-js": "1.9.2",
```

editor/package.json
```diff
@@ -217,6 +218,7 @@
     "@types/path-parse": "1.0.19",
     "@types/prettier": "2.0.1",
     "@types/prop-types": "15.5.6",
+    "@types/pubsub-js": "^1.8.2",
```
Rheeseyb, line 221:
```suggestion
    "@types/pubsub-js": "1.8.2",
```

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -117,7 +117,7 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThanOrEqual(455) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(465)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThanOrEqual(460) // if this breaks, GREAT NEWS but update the test please :)
```
Rheeseyb, line 120: Interesting that this one has gone up...

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -131,6 +92,21 @@ const ErrorOverlayComponent = betterReactMemo(
       return getAllCodeEditorErrors(store.editor, true, true)
     }, 'ErrorOverlayComponent fatalCodeEditorErrors')
 
+    const runtimeErrors = useReadOnlyRuntimeErrors()
+
+    const overlayErrors = React.useMemo(() => {
+      return runtimeErrors.map((runtimeError) => {
+        const stackFrames =
+          runtimeError.error.stackFrames != null ? runtimeError.error.stackFrames : []
```
seanparsons, line 100: `defaultIfNull`?

editor/src/core/shared/atom-with-pub-sub.ts
```diff
@@ -0,0 +1,76 @@
...
+import * as React from 'react'
+
+export interface AtomWithPubSub<T> {
+  key: string
+  currentValue: T
+}
+
+const GlobalAtomMap: { [key: string]: AtomWithPubSub<any> } = {}
+
+export function atomWithPubSub<T>(options: { key: string; defaultValue: T }): AtomWithPubSub<T> {
```
Rheeseyb, line 11: Why have you gone with `options: { key: string; defaultValue: T }` rather than two separate params here?

Rheeseyb (approved):

seanparsons (approved):

editor/src/components/canvas/canvas-wrapper-component.tsx
balazsbajorics: this is just moved code, I thought it's best if I keep it as-is for easier brain processing of PR

editor/src/core/performance/performance-regression-tests.spec.tsx
balazsbajorics: now that you mention – I didn't implement setState's default behavior that if `currentState === nextState` then `setState(nextState)` will be a NO_OP. I think I should amend this!

editor/package.json
balazsbajorics: I am the worst. is there a way to configure npm to avoid ^s by default?

PR #895, opened 2021-01-27
adding testId to every string and number control

balazsbajorics:
The inspector controls already had an `id` property, but that was also doubling as the html/css `id` html attribute. 
The react-testing-library explicitly disallows querying elements by their `id` and forces you to use a `data-testid`. Instead of just printing the `id` value to `data-testid`, I introduced a new `testId` prop, to carry on the spirit of the testing library, and not tie the test code to css ids.

The PR introduces no other changes to any functionality, this is just preparation.


balazsbajorics:
hello! I am a lonely PR, missing reviews

enidemi (approved):

seanparsons (approved):

Issue #888, opened 2021-01-26
empty (or fresh inserted) scenes do not show the Scene Inspector


balazsbajorics:
When I select a scene which has `component={null}`. 
First off, I can only select it via the navigator.
Second, the inspector shows this:
<img width="299" alt="image" src="https://user-images.githubusercontent.com/2226774/105856805-e6da0a80-5fe9-11eb-9038-1a6baf9b3f5a.png">
which is definitely not the Scene inspector which I should be seeing:
<img width="320" alt="image" src="https://user-images.githubusercontent.com/2226774/105856828-ee011880-5fe9-11eb-8150-a8e062f4c821.png">

This means that if I insert a new scene, I cannot use the UI to give it a target component, which means scene insertion is totally broken!

PR #898, opened 2021-01-29
fix(scenes) componentmetadata should keep empty scenes

enidemi:
Fixes https://github.com/concrete-utopia/utopia/issues/888

When you insert a new scene on the canvas it shows an element inspector instead of a scene inspector and the canvas control label is missing.

Since the storyboard+Scene changes the scene metadata (and other direct children of the storyboard, like orphans) use instance path as a templatepath, this was easily filtered from real scenes with the spy cleaner.

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -77,6 +77,7 @@ import {
   foldParsedTextFile,
   textFile,
   textFileContents,
+  ScenePath,
 } from '../../core/shared/project-file-types'
 import {
   getOrDefaultScenes,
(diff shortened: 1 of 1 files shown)


balazsbajorics (approved):

Rheeseyb (approved):

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -2507,7 +2508,12 @@ export function cullSpyCollector(
   })
   // Eliminate the scene paths which are invalid.
   fastForEach(Object.keys(spyCollector.current.spyValues.scenes), (scenePath) => {
-    if (!scenePaths.has(scenePath)) {
+    if (
+      !scenePaths.has(scenePath) &&
+      !elementPaths.has(
+        TP.toString(TP.instancePath([], (TP.fromString(scenePath) as ScenePath).sceneElementPath)),
```
seanparsons, line 2514: I _believe_ you can pull the scene path object from `spyCollector.current.spyValues.scenes`, then invoke `sceneElementPath` on that. We _really_ don't want to be running `TP.fromString` if we can help it.

seanparsons (approved):

PR #899, opened 2021-02-01
Fix/inspector layout pin displayed values

enidemi:
Fixes a layout/pin section bug in the inspector where unset values are shown even if they are unset.

Fix:
- use the hook result pointInfo to get the display value
- remove the fullFrame calculated fallback, because in general we use computedStyle as fallback with different controlstatus/style, that should come from the hook.
- 3 tests that show the expected result checking the controlstatus and the input value.

### DIFF
--- a/editor/src/components/inspector/sections/layout-section/self-layout-subsection/gigantic-size-pins-subsection.tsx
+++ b/editor/src/components/inspector/sections/layout-section/self-layout-subsection/gigantic-size-pins-subsection.tsx
@@ -49,11 +49,7 @@ export const PinsLayoutNumberControl = betterReactMemo(
   (props: PinsLayoutNumberControlProps) => {
     const framePoint = framePointForPinnedProp(props.prop)
     const pointInfo = useInspectorLayoutInfo(props.prop)
-    const fullFrame = Utils.defaultIfNull(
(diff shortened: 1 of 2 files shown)


editor/src/components/inspector/sections/layout-section/self-layout-subsection/gigantic-size-pins-subsection.tsx
```diff
@@ -49,11 +49,7 @@ export const PinsLayoutNumberControl = betterReactMemo(
   (props: PinsLayoutNumberControlProps) => {
     const framePoint = framePointForPinnedProp(props.prop)
     const pointInfo = useInspectorLayoutInfo(props.prop)
-    const fullFrame = Utils.defaultIfNull(
-      {} as FullFrame,
-      Utils.optionalMap(getFullFrame, props.frame),
```
balazsbajorics, line 54: is `props.frame` used for other things too? if not, please also remove it from props to declutter the code

enidemi: good idea, fixed

Rheeseyb (commented):
:+1:

PR #900, opened 2021-02-01
Fix: Padding control shows 'px' at the end of values

balazsbajorics:
**Problem:**
The padding control was showing the `px` unit even though most of the inspector hides `px` and only shows a unit if it is different from `px`.

**Fix:**
I realized that the problem was that defaultUnitToHide was not provided for paddingControl

**Commit Details:**
- Made defaultUnitToHide an obligatory prop
- Tried to put either `'px'` or `null` as defaultUnitToHide everywhere. **Note:** I tried to put nulls in the various background gradient controls, please catch me if I used a `px` in there somewhere
- Added a new test that makes sure the padding control prints the number without `px`


enidemi (approved):

Rheeseyb (approved):

editor/src/uuiui/inputs/number-input.tsx
```diff
@@ -124,7 +124,7 @@ export interface NumberInputOptions {
   height?: number
   roundCorners?: BoxCorners
   numberType: CSSNumberType
-  defaultUnitToHide?: CSSNumberUnit
+  defaultUnitToHide: CSSNumberUnit | null
```
seanparsons, line 127: Yay!

seanparsons (approved):

PR #901, opened 2021-02-01
Fix inspector layout parsers

enidemi:
This is a fix to unify the inspector controls a little. FramePin parsing was not using the css parser, but something custom simple version of it. I couldn't use the CSSNumber directly because that's a more complex type and effects more things (like canvas-utils frame updating), this is a smaller step to at least use the same parser.

Fixes:
- use `parseCSSNumber`
- for px and unitless cases return the number value
- other cases use it as a string
- tests are updated too, because this introduces new behavior in the layout section

### DIFF
--- a/editor/src/components/inspector/common/css-utils.ts
+++ b/editor/src/components/inspector/common/css-utils.ts
@@ -3799,13 +3799,15 @@ function parseFramePin(
   simpleValue: unknown | null,
   _: ModifiableAttribute | null,
 ): Either<string, FramePin> {
-  if (
-    typeof simpleValue === 'number' ||
-    (typeof simpleValue === 'string' && isPercentPin(simpleValue))
-  ) {
-    return right(simpleValue)
+  const parsedValue = parseCSSNumber(simpleValue, 'Length')
+  if (isRight(parsedValue)) {
+    if (parsedValue.value.unit === 'px') {
+      return right(parsedValue.value.value)
+    } else {
+      return right(`${parsedValue.value.value}${parsedValue.value.unit}`)
+    }
   } else {
-    return left('Value is not a valid frame pin.')
+    return parsedValue
   }
 }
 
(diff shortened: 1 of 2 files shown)


editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser.tsx
```diff
@@ -87,15 +87,15 @@ describe('inspector tests with real metadata', () => {
       leftControl.attributes.getNamedItemNS(null, 'data-controlstatus')?.value,
     ).toMatchInlineSnapshot(`"simple"`)
 
-    expect(bottomControl.value).toMatchInlineSnapshot(`""`)
+    expect(bottomControl.value).toMatchInlineSnapshot(`"178"`)
```
balazsbajorics, line 90: this is interesting. why does the bottomControl changed from empty to 178? is it because the `detected` value was 178px and the `px` part screwed with the parsing and the logic fall back to unset?

enidemi: yes, now all layout controls show computedStyle with light grey color. this is good because now it is matching how other controls like minwidth/padding/margin works.

balazsbajorics (approved):

balazsbajorics:
just a note that while I think this PR is okay, we shouldn't forget that eventually we want to _not_ show detected values for layout controls which are not explicitly set. (ie do not show `detected` right and bottom if TLWH are set)

Rheeseyb (approved):

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -3799,13 +3799,15 @@ function parseFramePin(
   simpleValue: unknown | null,
   _: ModifiableAttribute | null,
 ): Either<string, FramePin> {
-  if (
-    typeof simpleValue === 'number' ||
-    (typeof simpleValue === 'string' && isPercentPin(simpleValue))
-  ) {
-    return right(simpleValue)
+  const parsedValue = parseCSSNumber(simpleValue, 'Length')
+  if (isRight(parsedValue)) {
```
seanparsons, line 3803: `mapEither` looks like it would make this a bunch simpler.

seanparsons (approved):

PR #905, opened 2021-02-04
Fix/inspector layout percents

### LINKED ISSUE #904: inspector: % values in layout controls should work

enidemi:
Fixes #904 

**Problem:**
Element styles with percent values are not editable in the inspector, `width: '100%'` is shown as unknown type in non-editable control.

**Fix:**
Update the `parseFramePin` to use `LengthPercent` parser.
Also switched other layout parsers, margin/padding/min-maxwidth/flexgrow to their correct type based on the mdn docs.

**Commit Details:**
- update parser in css-utils
- fixed the new tests

### DIFF
--- a/editor/src/components/inspector/common/css-utils.ts
+++ b/editor/src/components/inspector/common/css-utils.ts
@@ -3799,13 +3799,15 @@ function parseFramePin(
   simpleValue: unknown | null,
   _: ModifiableAttribute | null,
 ): Either<string, FramePin> {
-  if (
-    typeof simpleValue === 'number' ||
-    (typeof simpleValue === 'string' && isPercentPin(simpleValue))
-  ) {
-    return right(simpleValue)
+  const parsedValue = parseCSSNumber(simpleValue, 'Length')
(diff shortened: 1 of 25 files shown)


editor/src/components/inspector/common/css-utils.ts
```diff
@@ -4113,27 +4113,27 @@ const cssParsers: CSSParsers = {
...
-  flexShrink: parseCSSLength,
+  minWidth: parseCSSLengthPercent,
+  maxWidth: parseCSSLengthPercent,
+  minHeight: parseCSSLengthPercent,
+  maxHeight: parseCSSLengthPercent,
+  marginTop: parseCSSLengthPercent,
+  marginRight: parseCSSLengthPercent,
+  marginBottom: parseCSSLengthPercent,
+  marginLeft: parseCSSLengthPercent,
+  flexGrow: parseCSSUnitless,
```
balazsbajorics, line 4135: nice catch!

balazsbajorics (approved):
nice!

PR #907, opened 2021-02-04
Detect and show if an inspector value is coming from a css stylesheet

balazsbajorics:
**Problem:**
We have these "detected" values in the inspector that are drawn as if they are defaults. We want to be able to show to the user if a value is not coming from the style prop, but is also not a default.

**Fix:**
Detect if a value is coming from a css stylesheet **Warning: this includes Emotion rules too**

**Commit Details:**
- Added `getAttributesComingFromStyleSheets` to DOM-walker. 
- **Note:** `getAttributesComingFromStyleSheets` takes ~40ms for an element to run, so we only want to evaluate it if the selection changes.
- added `attributeMetadatada` to `ElementInstanceMetadata`. right now the only metadata it stores: `{ fromStyleSheet: boolean }`. 
- adding a new controlStatus `detected-fromcss`, this is like `detected` but we also know the data is coming from css
- setting a placeholder controlStyle for `detected-fromcss` which is the same as a regular value but with green font color
- weaving `attributeMetadatada` through the inspector hooks
- **Note** the inspector hooks are really messy and I didn't feel great extending them, but I decided not to change anything about it, this PR only extends the existing code
- updating the inspector tests with the new controlStyle


balazsbajorics:
image
this is how the inspector looks like when `padding` is set via css

seanparsons:
> * `getAttributesComingFromStyleSheets` takes ~40ms for an element to run, so we only want to evaluate it if the selection changes.

*sharp intake of breath*

balazsbajorics:
> > * `getAttributesComingFromStyleSheets` takes ~40ms for an element to run, so we only want to evaluate it if the selection changes.
> 
> _sharp intake of breath_

actually approved by @maltenuhn with the proviso that it shall only run on selection change (which is slow anyways and we deploy visual tricks to make it feel less slow)

editor/src/components/canvas/dom-walker.ts
```diff
@@ -135,6 +137,50 @@ function lazyValue<T>(getter: () => T) {
   }
 }
 
+function getAttributesComingFromStyleSheets(element: HTMLElement): Set<string> {
+  let appliedAttributes = new Set<string>()
+  const sheets = document.styleSheets
+  try {
+    for (var i in sheets) {
```
Rheeseyb, line 144: These both should probably run using `fastForEach`

seanparsons, line 144:
```suggestion
    for (const i in sheets) {
```

maltenuhn:
@balazsbajorics if we did want to further perceived-performance optimise things here, could we run it first for a list of attributes we care about a lot?

editor/src/components/canvas/dom-walker.ts
```diff
@@ -135,6 +137,50 @@ function lazyValue<T>(getter: () => T) {
...
+      var rules = sheets[i].rules || sheets[i].cssRules
```
seanparsons, line 145: Can we not just `for...of` this instead if we don't care about the index otherwise? As it's indexing here twice.

editor/src/components/canvas/dom-walker.ts
```diff
@@ -231,14 +279,18 @@ export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElem
   const containerRef = React.useRef<HTMLDivElement>(null)
   const rootMetadataInStateRef = useRefEditorState((store) => store.editor.domMetadataKILLME)
   const invalidatedSceneIDsRef = React.useRef<Set<string>>(emptySet())
+  const invalidatedPathsForStylesheetCacheRef = React.useRef<Set<string>>(new Set())
```
seanparsons, line 282: Just as a safeguard.
```suggestion
  const invalidatedPathsForStylesheetCacheRef = React.useRef<Set<string>>(emptySet())
```

editor/src/utils/deep-equality.ts
```diff
@@ -358,6 +358,72 @@ export function combine9EqualityCalls<A, B, C, D, E, F, G, H, I, X>(
     }
   }
 }
+
+export function combine10EqualityCalls<A, B, C, D, E, F, G, H, I, J, X>(
```
Rheeseyb, line 362: So much equality!

editor/src/components/canvas/dom-walker.ts
balazsbajorics: I _think_ I tried that in codesandbox and it crashed because IIRC `sheets` is a thing that has keys but is not iterable or something horrifying like that

Rheeseyb (approved):

editor/src/utils/deep-equality.ts
balazsbajorics: a few weeks ago I spent a day trying to use new typescript type features to annotate a combineNEqualityCalls function but I failed

maltenuhn:
> yep. a lot of it is because of Emotion's thousands of little stylesheets. which we can maybe remove in the future. but as it is now, at least it gives us a stress test so we know we are prepared for thousands of stylesheets

Do you mean 'rule' when you say 'little stylesheet' - or do we have thousands of stylesheets attached by emotion? And do you know why we have so many of them?

editor/src/components/canvas/dom-walker.ts
Rheeseyb: Ooooh... so maybe `fastForEach` wouldn't work here then :/

seanparsons (approved):

balazsbajorics:
> @balazsbajorics if we did want to further perceived-performance optimise things here, could we run it first for a list of attributes we care about a lot?

@maltenuhn we could try to discard stylesheets which only set irrelevant properties before running `match` against the element. but I think it'd be not much gain because our emotion stylesheets usually set properties we also care about in the inspector.

Rheeseyb:
> > > * `getAttributesComingFromStyleSheets` takes ~40ms for an element to run, so we only want to evaluate it if the selection changes.
> > 
> > 
> > _sharp intake of breath_
> 
> actually approved by @maltenuhn with the proviso that it shall only run on selection change (which is slow anyways and we deploy visual tricks to make it feel less slow)
> 
> > I'm guessing this is just a cost we have to eat and there's not much we can do about it?
> 
> yep. a lot of it is because of Emotion's thousands of little stylesheets. which we can maybe remove in the future. but as it is now, at least it gives us a stress test so we know we are prepared for thousands of stylesheets

It's expensive now but tbh I have no doubt we'll find ways to optimise this in the future, so looks good to me

enidemi (commented):
👍

PR #911, opened 2021-02-08
Inspector: hide trivial default values

balazsbajorics:
**Problem:**
Our "empty" inspector is full of greyed-out default numbers: half a dozen 0s for padding, margin, etc.

**Fix:**
Only show default numbers if those are not trivial

**Commit Details:**
- Created a new dictionary named `trivialDefaultValues`. For every property, we store if the default value is trivial (and what it is). If something is nontrivial, we store `nontrivial`
        **Rules:**
            -- if the value is set in the style prop, show it plainly as ‘set’
            -- otherwise, if the value is coming from a css rule, show it as ‘detected-fromcss’ 
            -- otherwise, compare the detected value with the value stored in `trivialDefaultValues`. If it doesn't match or `trivialDefaultValues` has a `nontrivial` entry for the given prop , show it as `detected`
            -- othwerwise, do not show the value

- When calculating the control style, if the value is not set in the style prop, and not set via css stylesheet (so it is either inherited or a browser default) AND the value matches the stored trivial default, we set the control style to `showContent: false`
- Added a new end-to-end unit test for an element that has only default style.

### DIFF
--- a/editor/src/components/inspector/common/css-utils.ts
+++ b/editor/src/components/inspector/common/css-utils.ts
@@ -1,5 +1,6 @@
 // FIXME This file shouldn't live under the inspector, and shouldn't be defining types
 import * as Chroma from 'chroma-js'
+import * as fastDeepEqual from 'fast-deep-equal'
 import { Property } from 'csstype'
 import {
   FlexAlignment,
@@ -1553,13 +1554,13 @@ function printTransform(cssTransforms: CSSTransforms): JSXAttributeValue<Propert
   return jsxAttributeValue(cssTransforms.map(printCSSTransformItem).join(' '), emptyComments)
 }
 
-enum CSSTransformOriginStringValueX {
+export enum CSSTransformOriginStringValueX {
   Left = 'left',
   Center = 'center',
   Right = 'right',
 }
 
-enum CSSTransformOriginStringValueY {
+export enum CSSTransformOriginStringValueY {
   Top = 'top',
   Center = 'center',
   Bottom = 'bottom',
@@ -2614,7 +2615,9 @@ export function cssSolidColor(color: CSSColor, enabled = true): CSSSolidColor {
   }
 }
 
-const emptyBackgroundColor: CSSSolidColor = cssSolidColor(cssColorRGB(0, 0, 0, 0, false, false))
+export const emptyBackgroundColor: CSSSolidColor = cssSolidColor(
+  cssColorRGB(0, 0, 0, 0, false, false),
+)
 
 export function isCSSSolidColor(
   value: CSSBackground | CSSUnknownArrayItem,
@@ -4378,7 +4381,7 @@ export interface ParsedElementProperties
 
 export type ParsedElementPropertiesKeys = keyof ParsedElementProperties
 
-const DOMEventHandlerEmptyValues = DOMEventHandlerNames.reduce((current, item) => {
+export const DOMEventHandlerEmptyValues = DOMEventHandlerNames.reduce((current, item) => {
   current[item] = jsxAttributeValue(undefined, emptyComments)
   return current
 }, {} as DOMEventAttributeProperties)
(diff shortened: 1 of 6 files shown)


editor/src/components/inspector/common/css-utils.ts
```diff
@@ -4746,3 +4749,163 @@ const LayoutPropertyList = [
...
+  trivial: false
+}
+
+const nontrivial: NonTrivialKeyword = { trivial: false }
+
+type ParsedPropertiesWithNonTrivial = {
+  [Property in keyof ParsedProperties]: ParsedProperties[Property] | NonTrivialKeyword
+}
+
+export const trivialDefaultValues: ParsedPropertiesWithNonTrivial = {
```
balazsbajorics, line 4763: @maltenuhn  this dictionary will need some QA. for every trivial value we want to hide (like maxWidth: none) the entry here must match the value found in `computedStyle`. 

We can't use the default useragent stylesheet because at least in Chrome, the length units for example are computed as `0px` instead of `0`

editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser.tsx
```diff
@@ -519,10 +519,10 @@ describe('inspector tests with real metadata', () => {
       'opacity-number-control',
     )) as HTMLInputElement
     const minWidthControl = (await renderResult.renderedDOM.findByTestId(
-      'position-Width-number-input',
```
balazsbajorics, line 522: @enidemi I found that the endtoend tests were using the wrong controls for minWidth and maxWidth, I fixed them here

editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser.tsx
```diff
@@ -569,10 +569,10 @@ describe('inspector tests with real metadata', () => {
     expect(minWidthControl.value).toMatchInlineSnapshot(`"0"`)
     expect(
       minWidthControl.attributes.getNamedItemNS(null, 'data-controlstatus')?.value,
-    ).toMatchInlineSnapshot(`"simple-unknown-css"`)
```
balazsbajorics, line 572: TBH I don't know the reason for this change

editor/src/components/inspector/common/css-utils.ts
maltenuhn: I've taken a look through, and will leave a couple of annotations, but overall this looks good to me and we can merge it as is. (I take it this dictionary implies that every property can have zero or one values we can hide?)

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -4746,3 +4749,163 @@ const LayoutPropertyList = [
...
+  },
+  marginBottom: {
+    value: 0,
+    unit: 'px',
+  },
+  marginLeft: {
+    value: 0,
+    unit: 'px',
+  },
+  flexGrow: cssUnitlessLength(0),
```
maltenuhn, line 4865: for `flexGrow` and `flexShrink`, the default values at least to me seem meaningful enough that I'd not hide them - in other words, I'd always show these.

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -4746,3 +4749,163 @@ const LayoutPropertyList = [
...
+  textSizing: 'fixed',
+  ...DOMEventHandlerEmptyValues,
+  className: '',
+
+  // ParsedLayoutProperties
+  layoutSystem: undefined,
+  pinLeft: undefined,
+  pinRight: undefined,
+  centerX: undefined,
+  width: undefined,
```
maltenuhn, line 4881: I assume this PR doesn't address this, but in case it does: what do we show if width is not explicitly set (e.g. `<span>Hello!</span>`)?

editor/src/components/inspector/common/css-utils.ts
balazsbajorics: > I've taken a look through, and will leave a couple of annotations, but overall this looks good to me and we can merge it as is. 

> (I take it this dictionary implies that every property can have zero or one values we can hide?)

yes. if that is not good enough, we can change it later to some other format

editor/src/components/inspector/common/css-utils.ts
balazsbajorics: image

we see an empty Flow section

enidemi (approved):

editor/src/components/inspector/common/css-utils.ts
maltenuhn: > yes. if that is not good enough, we can change it later to some other format

Nah,  fine as is. If we truly have multiple values that could each be meaningless, that fact in itself probably makes them meaningful :)

seanparsons (approved):

editor/src/components/inspector/common/css-utils.ts
maltenuhn: Interesting, I thought we'd show [auto] here... but anyway, that's, out of scope!

PR #912, opened 2021-02-09
Fix: hide min-max width-height controls if they are empty

balazsbajorics:
**Problem:**
The min-max row was showing up with default computedStyle (showing 0 for min, nothing for max, to make things more interesting) instead of being hidden when unset.

**Fix:**
The regression happened with the introduction of computedStyle measurements. The 0 for min was interpreted as a not-unset value and that would make the subsection show up. 

**Commit Details:**
- max-width and max-height uses a new parser called `parseCSSLengthPercentNone` which can interpret a `<length> | <percentage> | none`. **Note:** `none` is parsed as `undefined` as that makes it easier to work with the number input.
- change `hasMinMaxValues` to hide minmax if the value is `unset` or if it is `trivial-default` (the new control style)
- update the tests


balazsbajorics:
the previous test misfire was probably a glitch – this PR is ready for reviews 🥰

seanparsons (approved):

Rheeseyb (approved):

PR #913, opened 2021-02-10
Feature/inspector padding shorthands

enidemi:
**Problem:**
Shorthand css values are not supported in the inspector.

**Fix:**
This PR introduces padding shorthand parsing and printing, based on [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties) using the 1-to-4-value syntax. Now the css-parser is capable of parsing values like `padding: "2px 4px"`.
It is not used anywhere in the inspector yet, that will be the next PR.

**Commit Details:**
- update css-tree to 1.1.2
- added helper to print cssnumber with default unit, as the shorthand syntax needs units
- interface for CSSPadding containing all sides
- fix an error in css-background-size where the error message was wrong
- extending `getLexerPropertyMatches` to add a default unit, this fixes the missing number type support
- fixed `parseLengthPercentage` in css-parser, this part is untyped using the css-tree, and in some cases the length-percentage contains a nested block (for background-size), some cases it is directly storing the value (for padding)
- unit tests for parser and printer

### DIFF
--- a/editor/src/printer-parsers/css/css-parser-padding.ts
+++ b/editor/src/printer-parsers/css/css-parser-padding.ts
@@ -1,0 +1,59 @@
+import { CSSNumber, CSSPadding, printCSSNumber } from '../../components/inspector/common/css-utils'
+import { Either, isRight, left, right } from '../../core/shared/either'
+import { JSXAttributeValue, jsxAttributeValue } from '../../core/shared/element-template'
+import { emptyComments } from '../../core/workers/parser-printer/parser-printer-comments'
+import { getLexerPropertyMatches, parseLengthPercentage, parseCSSArray } from './css-parser-utils'
+
+export const parsePadding = (value: unknown): Either<string, CSSPadding> => {
+  const lexer = getLexerPropertyMatches('padding', value)
+  if (isRight(lexer)) {
+    const parseResult = parseCSSArray([parseLengthPercentage])(lexer.value)
+    if (
+      isRight(parseResult) &&
+      Array.isArray(parseResult.value) &&
+      parseResult.value.every(isRight)
+    ) {
+      const resultArray = parseResult.value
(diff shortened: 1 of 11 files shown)


balazsbajorics (approved):

editor/src/printer-parsers/css/css-parser-padding.ts
```diff
@@ -0,0 +1,91 @@
...
+import { emptyComments } from '../../core/workers/parser-printer/parser-printer-comments'
+import { getLexerPropertyMatches, parseLengthPercentage, parseCSSArray } from './css-parser-utils'
+
+export const parsePadding = (value: unknown): Either<string, CSSPadding> => {
+  const lexer = getLexerPropertyMatches('padding', value, 'px')
+  if (isRight(lexer)) {
+    const parseResult = parseCSSArray([parseLengthPercentage])(lexer.value)
+    if (isRight(parseResult)) {
+      const resultArray = parseResult.value
+      let paddingTop, paddingRight, paddingBottom, paddingLeft
```
seanparsons, line 19: Ideally it would be nice if these had types assigned to them.

editor/src/printer-parsers/css/css-parser-utils.ts
```diff
@@ -393,6 +403,19 @@ export function parseDoubleBar<T>(
   }
 }
 
+export function parseCSSArray<T>(parsers: Array<Parser<T>>): Parser<Array<T>> {
+  return (match: unknown): ParseResult<Array<T>> => {
+    if (Array.isArray(match) && match.length > 0) {
+      return sequenceEither(
```
seanparsons, line 409: `traverseEither` is equivalent to a `map` followed by a `sequenceEither`.

seanparsons (approved):

PR #914, opened 2021-02-11
Fix/inspector keep style props in place

enidemi:
**Problem:**
Element style can contain longhand and shorthand values that affect the same style props, like padding, border, margin. For these style props the order matters, but the setProp update always pushes the new value to the end.

**Fix:**
Update `setJSXValueInAttributeAtPath` for nested objects to update in place instead of filtering and adding to the end of the content array.

**Commit Details:**
- update `setJSXValueInAttributeAtPath`
- added one test
- updated lots of canvas tests

### DIFF
--- a/editor/src/core/shared/jsx-attributes.ts
+++ b/editor/src/core/shared/jsx-attributes.ts
@@ -470,17 +470,30 @@ export function setJSXValueInAttributeAtPath(
           const attributeKey = PP.firstPart(path)
           const tailPath = PP.tail(path)
           const key = `${attributeKey}`
-          const newProps = dropKeyFromNestedObject(attribute, key).content
           if (lastPartOfPath) {
-            return right(
-              jsxAttributeNestedObject(
-                newProps.concat(
-                  jsxPropertyAssignment(key, newAttrib, emptyComments, emptyComments),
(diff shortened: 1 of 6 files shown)


balazsbajorics (approved):
it's so good to see the fixed tests – no human would write code like `style={{height; width;}}` :D

editor/src/core/shared/jsx-attributes.ts
```diff
@@ -470,17 +470,30 @@ export function setJSXValueInAttributeAtPath(
...
           if (lastPartOfPath) {
-            return right(
-              jsxAttributeNestedObject(
-                newProps.concat(
-                  jsxPropertyAssignment(key, newAttrib, emptyComments, emptyComments),
+            let updatedExistingProperty = false
+            let updatedContent = attribute.content.map((attr) => {
+              if (attr.type === 'PROPERTY_ASSIGNMENT' && attr.key === key) {
+                updatedExistingProperty = true
+                return jsxPropertyAssignment(key, newAttrib, emptyComments, emptyComments)
```
Rheeseyb, line 478: Looks like this will bin off comments if a property is updated. We should probably copy the comments across from the old attribute

PR #920, opened 2021-02-15
JSX Attributes: getAllPathsFromAttributes

balazsbajorics:
The inspector needs to understand shorthand and longhand properties shadowing each other.

We made this new utility function to be able to list all property paths of a given `JSXAttributes` props object, in order of appearance.  

**Commit Details:**
- we modified `walkAttributes` to include the attribute's property path in the callback
- created a new `getAllPathsFromAttributes` that uses `walkAttributes` 
- wrote tests

### DIFF
--- a/editor/src/core/model/jsx-attributes.spec.ts
+++ b/editor/src/core/model/jsx-attributes.spec.ts
@@ -24,6 +24,7 @@ import {
 } from '../shared/element-template'
 import {
   dropKeyFromNestedObject,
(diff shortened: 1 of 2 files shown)


enidemi (approved):

balazsbajorics:
I'm gonna go ahead and merge this, I need to use it in our main PR branch

Rheeseyb (commented):
:+1:

PR #921, opened 2021-02-15
Embedded VS Code Part 1

### LINKED ISSUE #902: Run VS Code in Utopia behind a feature flag
- [x] Fork VS Code (Make our spike a bit more production ready)
  - [x] Tidy up the work around removing parts of the editor we don’t want
  - [x] Incorporate the build scripts for bundling the web version, including it into GH workflow
- [x] Create Extension for providing FS via BrowserFS
  - [x] Create an up-to-date extension (write / rewrite) for r/w files to BrowserFS (using our spike) in our main Utopia repo
  - [x] Update our VS Code fork to communicate with the extension via BrowserFS
- [x] Serve VS Code in an iframe [1PD]
  - [x] Ensure we can serve up the VS Code bundle as an asset via the existing endpoints or create a new route if required
  - [x] Run VS Code in an iframe (wrapped in a React Component) inside Utopia (behind a deactivated feature flag)
  - [x] Run it with our extension
  - [x] Potentially remove Utopia tabbing system at this point, or leave as is until we’re ready to remove the feature flag

seanparsons:
Fixes #902

**Problem:**
We want to run full VS Code inside Utopia rather than Monaco.

**Fix:**
This solution is a combination of a few parts:

**1. Bundling VS Code**
VS Code doesn't come with a way of bundling / minifying the web build - the tasks for this did exist at some point but were later moved or removed for some reason. There's an issue [here](https://github.com/microsoft/vscode/issues/101087) about it but it seems there is no plan to expose those again. To work around this, we have used [this repo](https://github.com/Felx-B/vscode-web) to seed a patch that we have created. The result of that all lives under `vscode-build`, which includes a script for checking out a specific version of VS Code, applying a patch to allow us to bundle the web version, and then actually building it.

**2. Patching VS Code**
Within the `vscode-build` sub-project there are two scripts specifically for this process:

- `build` - Which pulls/updates the source code to `vscode-build/vscode`, applies the patch held in `vscode.patch` and runs the customised build flow dropping all the relevant files into `vscode-build/dist`.
- `make-patch` - This takes the source code as it currently is in `vscode-build/vscode` and reproduces the patch file `vscode.patch` from that. After making additional modifications to the VS Code source this would be run to record the changes into that patch file.

**3. Using a Custom Extension**
We make use of a VS Code extension to provide as much of the functionality as possible like bridging files and their content between the main editor and VS Code. Where possible changes should go into the extension rather than by modifying the core source with a patch.

**4. Serving it all up**
The VS Code files are separately served from the `/vscode` endpoint in the server which in development relates to the `vscode-build/dist` folder in the project.

**5. Including it Utopia**
You may be asking why we have an outer and inner iframe for loading VS Code into the editor. This is so that we can use a second domain for the outer iframe (to take advantage of Chrome running that on a separate process), and then switch back to the main domain for the inner iframe, allowing Utopia and the VS Code extension to both directly read from and write to the same underlying IndexedDB store.

**Limitations:**
- This currently just hosts the modified VS Code in place of our current code editor, but the editor itself and VS Code aren't currently communicating beyond VS Code being project ID aware which keeps files edited within it separate by project.

**Notes:**
- Dependencies have been added to the `shell.nix` for VS Code to allow it to build on Linux, it's likely there will be similar dependencies that need adding for a Mac to build it.
- This needs further work for production as the VS Code parts are not included in the Dockerfile produced bundle and it doesn't get included as part of a regular build currently.


utopia-vscode-extension/src/browserfs-utils.ts
```diff
@@ -0,0 +1,295 @@
...
+  onDeleted: (path: string) => void
+}
+
+let watchTimeout: number | null = null
+let watchedPaths: Map<string, WatchConfig> = new Map()
+let lastModifiedTSs: Map<string, number> = new Map()
+const POLLING_TIMEOUT = 100
+
+// Immediately attempt to initialise BrowserFS
+new Promise<void>((resolve, reject) => {
```
balazsbajorics, line 25: why is this wrapped in a Promise? I thought that unhandled Promise rejections cause a big warning saying they are not supported

balazsbajorics (approved):

PR #923, opened 2021-02-16
Chore/support shorthands longhands

balazsbajorics:
**Problem:**
The `useInspectorInfo` hook would lie about the existence of a prop (it would return a default value instead of nothing). Furthermore, the new longhand-shorthand aware hooks will need information about the order of the property keys.

**Fix:**
I made a new hook, `useInspectorInfoNoDefaults`. I also wrote new tests. The hook is not yet used for anything in production.

**Commit Details:**
- Added new hook, `useInspectorInfoNoDefaults` which is a "fork" of the original `useInspectorInfo`. @enidemi  already started work in replacing all usage of `useInspectorInfo` with `useInspectorInfoNoDefaults` to keep the codebase manageable and avoid having multiple "forks" of the same function laying around
- wrote tests that explore padding and paddingLeft shorthands in various scenarios. this hook is just a pre-requisite for our understanding of longhands and shorthands, not full support for it 
- changed `InspectorInfo<T>` to `InspectorInfo<T, P>` and added a new prop: `orderedPropKeys: Array<Array<P>>` which contains the keys in order of appearance in the code (to be able to determine of paddingLeft shadows padding)
- for `InspectorInfo<T, P>` I had to include some `InspectorInfo<T, unknown>`s in the codebase, sorry for the mess


balazsbajorics:
oh my god I'm closing this PR, we realized that from this entire PR we only _really_ need the new `useMemoizedPropKeys` function utilized in a clever way...

PR #924, opened 2021-02-16
Feature/inspector hooks emptyvalues

enidemi:
The goal is to use the new hook which has no emptyValues inside the hook. Since cssdefaults and computedstyles is used to provide fallback where it's possible the emptyValues are causing only confusion.
But most of the control types rely on having a real value instead of undefined, so this fixes that.

- changed layout, style, and element hooks to use `useInspectorInfoNoDefaults`
- Changed all other places that directly use `useInspectorInfo` to  have `useInspectorInfoNoDefaults`


balazsbajorics:
I'm closing this PR for now because we changed the origin branch too much :( 

Sorry about this @enidemi  !

PR #925, opened 2021-02-18
css shorthand longhand support hook

balazsbajorics:
**Problem:**
We want to support css properties which can be expressed as either shorthand or longhand. To make things harder, the property which is defined _later_ in the style object shadows the other. 
So these are various cases with different meaning:
```javascript
{ padding: 5, paddingLeft: 10 }  ===> paddingLeft is 10
{ paddingLeft: 10, padding: 5 } ===> paddingLeft is 5
{ padding: 5, paddingLeft: undefined } ===> paddingLeft is 0!!!!!!!!!!!!!!!!!! 
```

**Fix:**
We made a new hook that can support any css property which can be expressed as part of a shorthand, or in a single longhand. The main use case (and the tests focus on) is padding and paddingLeft. We expect the hook to be easily generalizable to margin, borderRadius. 

**Commit Details:**
- created hook `useGetOrderedPropertyKeys` which returns the shorthand and longhand keys in order of appearance. we later use this information to figure out shadowing
- created new super-hook `useInspectorInfoLonghandShorthand` that uses `useInspectorInfo` and `useGetOrderedPropertyKeys` internally
- written extensive test coverage for both `useGetOrderedPropertyKeys` and `useInspectorInfoLonghandShorthand`

**Notes:**
For this first iteration, the hooks bail out in case of a multiselect where the selected elements have non identical property shadowing situation. (so if I select two elements: one with { paddingLeft: 5, padding: 5 }, and another with { padding: 5 } we bail out unceremoniously.) We can improve on this later and widen the happy path, but the PR is already large as it is.

### DIFF
--- a/editor/src/components/inspector/common/longhand-shorthand-hooks.ts
+++ b/editor/src/components/inspector/common/longhand-shorthand-hooks.ts
@@ -38,7 +38,7 @@ function getShadowedLonghandShorthandValue<
   longhandPropertyStatus: PropertyStatus,
   shorthandPropertyStatus: PropertyStatus,
   longhandValue: ParsedProperties[LonghandKey],
-  shorthandValue: ParsedProperties[ShorthandKey],
+  shorthandValueObject: ParsedProperties[ShorthandKey], // the shorthand value has to be an object where the keys are longhand property names and the types are same as the longhand values
   orderedPropKeys: (LonghandKey | ShorthandKey)[][], // multiselect
 ): { value: ParsedProperties[LonghandKey]; propertyStatus: PropertyStatus } {
   const allPropKeysEqual = orderedPropKeys.every((propKeys) => {
@@ -67,11 +67,11 @@ function getShadowedLonghandShorthandValue<
             },
       }
     } else {
-      // Important: we ASSUME that the transformed shorthand value is an object,
+      // Important: we assume that shorthandValue is an object
       // where the keys are the longhand keys and the values are the individual longhand values
-      if (longhand in shorthandValue) {
+      if (longhand in shorthandValueObject) {
         return {
-          value: (shorthandValue as any)?.[longhand] as ParsedProperties[LonghandKey],
+          value: (shorthandValueObject as any)?.[longhand] as ParsedProperties[LonghandKey],
           propertyStatus: allPropKeysEqual
             ? shorthandPropertyStatus
             : {
--- a/editor/src/components/inspector/common/longhand-shorthand-hooks.ts
+++ b/editor/src/components/inspector/common/longhand-shorthand-hooks.ts
@@ -1,22 +1,27 @@
-import { last } from '../../../core/shared/array-utils'
+import * as deepEqual from 'fast-deep-equal'
+import { useContextSelector } from 'use-context-selector'
+import { flatMapArray, last } from '../../../core/shared/array-utils'
 import { jsxAttributeValue } from '../../../core/shared/element-template'
 import { create } from '../../../core/shared/property-path'
-import { instancePath } from '../../../core/shared/template-path'
+import { instancePath, isInstancePath } from '../../../core/shared/template-path'
 import { arrayEquals } from '../../../core/shared/utils'
 import { emptyComments } from '../../../core/workers/parser-printer/parser-printer-comments'
+import { useKeepReferenceEqualityIfPossible } from '../../../utils/react-performance'
 import {
   getControlStatusFromPropertyStatus,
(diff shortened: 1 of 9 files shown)


editor/src/components/inspector/common/longhand-shorthand-hooks.ts
```diff
@@ -0,0 +1,236 @@
...
+  const inspectorTargetPath = useKeepReferenceEqualityIfPossible(
+    useContextSelector(InspectorPropsContext, (contextData) => contextData.targetPath, deepEqual),
+  )
+  const { selectedViewsRef } = useInspectorContext()
+  const orderedPropKeys = useGetOrderedPropertyKeys(pathMappingFn, [longhand, shorthand])
+  const longhandInfo = useInspectorInfo(
+    [longhand],
+    (v) => v[longhand],
+    () => {
+      return null as any
```
balazsbajorics, line 117: this is fine because we don't use `useInspectorInfo`'s builtin `onSubmitValue`, which means we don't need this untransformValue neither

editor/src/components/inspector/common/property-path-hooks.test-utils.tsx
```diff
@@ -0,0 +1,103 @@
...
+  isUtopiaJSXComponent,
+  JSXAttributes,
+  StyleAttributeMetadata,
+} from '../../../core/shared/element-template'
+import { isParseSuccess, TemplatePath } from '../../../core/shared/project-file-types'
+import { NO_OP } from '../../../core/shared/utils'
+import { testParseCode } from '../../../core/workers/parser-printer/parser-printer.test-utils'
+import { InspectorCallbackContext, InspectorPropsContext } from './property-path-hooks'
+
+export const makeInspectorHookContextProvider = (
```
balazsbajorics, line 14: this has been copy pasted from property-path-hooks.spec.tsx

maltenuhn:
> We can improve on this later and widen the happy path, but the PR is already large as it is.

👍  - definitely not one for here. FWIW I think we'll end up with a much more opinionated multiselect ('alert-and-nuke'). The price of convenience, so to say. But that's defo one for the canvas track, for later.

enidemi (approved):

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
```diff
@@ -0,0 +1,236 @@
...
+              ...longhandPropertyStatus,
+              set: true,
+              controlled: true,
+              overwritable: false,
+              identical: false,
+            },
+      }
+    } else {
+      // Important: we ASSUME that the transformed shorthand value is an object,
+      // where the keys are the longhand keys and the values are the individual longhand values
```
Rheeseyb, line 71: By "transformed shorthand value" here do you mean `shorthandValue`, i.e. the param to this function? It looks like you do, in which case please can you rename that param to something that hints that it should be an object that is the same shape as the longhand value (and move this comment to the function signature)

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
Rheeseyb: Can you make this a comment please

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
```diff
@@ -0,0 +1,236 @@
...
+    shorthandInfo.propertyStatus,
+    longhandInfo.value,
+    shorthandInfo.value,
+    orderedPropKeys,
+  )
+
+  const onSubmitValue = (
+    newTransformedValues: ParsedProperties[LonghandKey],
+    transient?: boolean | undefined,
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
```
Rheeseyb, line 142: Looks like this eslint comment can be removed?

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
balazsbajorics: yeah sure!

Rheeseyb (approved):

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
balazsbajorics: I left this comment here (with some modification) but also fixed the variable name and left a smaller comment there

Rheeseyb: Works for me! Thanks!

PR #931, opened 2021-02-23
Feature/longhand shorthand unset

balazsbajorics:
**Problem:**
The user might want to `unset` their longhand properties – even if there's no longhand property, only a shorthand property!

**Fix:**
This means we need to support breaking up the shorthand property into 4 longhands – and delete one longhand.

**Commit Details:**
- Mapped out various cases when an unset might happen
- wrote tests for these cases
- implemented a new `onUnsetValues` function that looks at the problem case-by-case and satisfies the unit tests
- did try it IRL and it seems fine


editor/src/components/inspector/common/longhand-shorthand-hooks.ts
```diff
@@ -245,14 +231,92 @@ export function useInspectorInfoLonghandShorthand<
...
+      const allPropKeysEqual = allOrderedPropKeys.every((propKeys) => {
+        return arrayEquals(propKeys, allOrderedPropKeys[0])
+      })
+      if (!allPropKeysEqual) {
+        // we do nothing for now. we cannot ensure that we can make a sensible update and surface it on the UI as well
+        return
+      }
+      const propkeysToUse = allOrderedPropKeys[0]
+
+      const actionsToDispatch = (() => {
```
seanparsons, line 256: Does this really need to be an immediately invoked anonymous function?

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
```diff
@@ -245,14 +231,92 @@ export function useInspectorInfoLonghandShorthand<
...
+        const shorthandExists = propkeysToUse.includes(shorthand)
+        const longhandPropertyPath = pathMappingFn(longhandToUnset, inspectorTargetPath)
+        const shorthandPropertyPath = pathMappingFn(shorthand, inspectorTargetPath)
+
+        if (!shorthandExists) {
```
seanparsons, line 261: Double negative alert.

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
```diff
@@ -245,14 +231,92 @@ export function useInspectorInfoLonghandShorthand<
...
+              }
+            }, selectedViewsRef.current)
+          } else {
+            // we have to split the shorthand into longhands 😭
+            return flatMapArray((selectedView) => {
+              if (isInstancePath(selectedView)) {
+                return [
+                  unsetProperty(selectedView, longhandPropertyPath),
+                  unsetProperty(selectedView, shorthandPropertyPath),
+                  ...longhands
```
seanparsons, line 294: Could this be broken out into a variable first for clarity?

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
```diff
@@ -245,14 +231,92 @@ export function useInspectorInfoLonghandShorthand<
...
     (result) => result,
   )
+
+  const longhandResultsWithUnset = objectMap((longhandResult, longhandToUnset) => {
+    const onUnsetValues = () => {
+      const allPropKeysEqual = allOrderedPropKeys.every((propKeys) => {
+        return arrayEquals(propKeys, allOrderedPropKeys[0])
+      })
+      if (!allPropKeysEqual) {
+        // we do nothing for now. we cannot ensure that we can make a sensible update and surface it on the UI as well
```
seanparsons, line 251: Does this mean a silent failure?

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
Rheeseyb, line 256: Please can you pull out this inlined function and if possible break up the individual parts so it's easier to trace which actions will be fired under which conditions

Rheeseyb (approved):

seanparsons (approved):

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
balazsbajorics: I'm testing the water with this one. when we were refactoring the `useInspectorInfo` hook, it was made harder by the pattern we use when we make a mutable variable at the top of the function, and then assign a value to it in various conditional branches. I have two problems: the mutable variable remains mutable even after the conditions because unfortunately typescript doesn't have an "assign once" const, the other problem is that you can't `break` or `return` in a condition so you cannot guarantee someone accidentally overwriting your value later down the line.

we talked about using an immediately executed function to simulate the missing javascript feature of `if`s being expressions.

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
```diff
@@ -245,14 +231,92 @@ export function useInspectorInfoLonghandShorthand<
...
+                      return setProp_UNSAFE(selectedView, pathTouse, printedValue)
+                    }),
+                ]
+              } else {
+                return []
+              }
+            }, selectedViewsRef.current)
+          }
+        }
+      })()
```
enidemi, line 311: 👀

enidemi (approved):

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
balazsbajorics: fixed

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
balazsbajorics: fixed

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
balazsbajorics: yes. this is code I copy pasted from here: https://github.com/concrete-utopia/utopia/blob/7a97a87663b77a7573614a3b42b0af5081913400/editor/src/components/inspector/common/longhand-shorthand-hooks.ts#L166

we agreed to not solve the not-identical multiselect problem for now

balazsbajorics: more info in the **Note:** section of this PR https://github.com/concrete-utopia/utopia/pull/925 and here https://github.com/concrete-utopia/utopia/pull/925#issuecomment-781590033

editor/src/components/inspector/common/longhand-shorthand-hooks.ts
seanparsons: I'd be tempted to just make it an inner function and return/use the value after invoking it.

balazsbajorics: I turned it into a regular function in the module scope, it was on the verge of being too big to be inline

Issue #813, opened 2021-01-05
Code editor remounts shortly after editor load finishes


balazsbajorics:
steps to reproduce: 
when the editor is fully loaded, select an element. upon the first selection change, the code editor will turn white and reload. 

this is caused by some parent component remounting. I don't know why it remounts!

maltenuhn:
Fixed by #933

PR #933, opened 2021-02-24
Feature/embedded vscode pt2

### LINKED ISSUE #932: Connect Utopia and VSCode
- [x] Provide communication between Utopia and VS Code
  - [x] Utopia sends full project contents to VS Code via iframe message post at start up
  - [x] VS Code sends back project file changes to Utopia
  - [x] Utopia sends individual file changes to VS Code as they happen
  - [x] Implement file watch functionality in the extension
  - [x] Trigger file opening in VS Code from Utopia
  - [x] Trigger scrolling / highlighting changes
    - [x] From Utopia to VS Code
    - [x] From VS Code to Utopia
  - [x] ~Support dirty / unsaved files~ Autosave files
    - [x] ~Flag in VS Code~
- [x] Deal with VS Code cached state / settings
  - [x] Prevent or support properly, including restoring it
    - [x] Try using project ID at the root of the path
  - [x] Make sure it doesn’t conflict across projects

Rheeseyb:
Fixes #932 

**Problem:**
We want to connect Utopia and VS Code to reproduce the functionality we had already with Monaco. This includes:

- Providing VS Code access to the full project contents at start up
- Sending file changes in both directions
- Triggering file opening in VS Code
- Triggering selection / highlighting in both directions
- Ensuring VS Code understands the difference between different projects, e.g.:
  - It doesn't try to restore a cached state of project A when opening project B
  - It won't return search results from project A whilst project B is open
  - It won't allow you to open or change a file from project A whilst project B is open

**Implementation Details:**

- **Introduced a new module `utopia-vscode-common`** for providing functionality that is required in both projects for connecting the two
  - BrowserFS was replaced with Localforage due to an issue with the former silently failing during a file write resulting in the file being truncated
  - `src/fs/*` for providing wrappers around Localforage to simulate the filesystem functionality required by VS Code
    - Each project creates its own IndexedDB store, allowing complete separation of projects
    - We use a `storesLastAccessedStore` for explicitly tracking the stores that have been created and when they were used, since IndexedDB doesn't provide a way of querying the stores that exist. This allows us to clear up stale project stores.
    - File watching functionality via polling to check when the files were last modified.
  - A mailbox feature for sending and receiving messages between Utopia and VS Code, which polls every 8ms to ensure minimal lag without blocking anything
    - Processed messages aren't deleted, since that would lead to a race condition if the user had the same project open in two tabs
    - Both mailboxes are emptied when starting Utopia so that we don't accidentally process old messages from a previous session
    - We hold off on initialising VS Code until the mailboxes are empty because we want it to be able to process messages sent from Utopia before it had finished initialising, but not process messages from a previous session.
    - The mailbox feature is built using the above fs system.
  - Both Utopia and the `utopia-vscode-extension` have dependencies on this common project, defined as local dependencies for the sake of simpler developer sanity
- In **Utopia** we have:
  - Introduced the `src/core/vscode/vscode-bridge.ts` file for providing the bridging between the two
    - This initialises the Utopia side of the mailboxes and clears both mailboxes
    - It initialises the IndexedDB project store and writes the entire project contents to it
      - This required the introduction of `walkContentsTreeAsync` and `zipContentsTreeAsync` in `src/components/assets.ts`, which is ~probably~ definitely the wrong place for these but they were added there because of the existing `walkContentsTree` in that file.
    - It watches for changes to files in that store, and updates the editor on those
    - It writes any changes to files coming from Utopia to the store
    - Provides the functions for sending messages to VS Code for triggering file opening and selection / highlighting
  - We initialise the bridge at the first available opportunity where we know the project ID, meaning that it happens at the beginning of a project load, or if creating a new project it happens as part of the handling of the `SET_PROJECT_ID` action
  - Added actions for: marking the bridge as ready; updating Utopia based on changes from VS Code to a file; selecting something based on a cursor position change.
  - We apply project content and decoration (selection / highlighting) changes as part of the dispatch function (meaning this is where we send those changes to VS Code)
  - We send the file opening message as part of the handling of the file opening action (`OPEN_EDITOR_TAB` because that action covers the case of switching tabs as well as opening a file)
  - For the sake of the tests, we mock out the entire bridging functionality. We also have to provide a custom Jest environment `editor/jest-environment.js` because Jest uses `jsdom`, which doesn't include `TextEncoder` or `TextDecoder` which prevented the tests from running
- In **VS Code** we have:
  - Changed the open folder's scheme from "utopia" to the project's ID, meaning the path can now be the root path. This allows us to provide the complete separation between projects inside VS Code whilst also ensuring that from the user's perspective it feels like a regular project in VS Code
    - To support this change we also had to make a change in the way VS Code activates `FileSystemProvider` extensions, since the logic there is used by VS Code to block it from accessing the file until the relevant extension has been activated, and there is no wildcard support for `onFileSystem:` activation events (which uses the scheme for the event)
  - Explicitly removed the config option to disable autosaving as that causes issues between Utopia and VS Code if making changes via both
- In **`utopia-vscode-extension`** we have:
  - Replaced the previous BrowserFS use with the new Localforage API from `utopia-vscode-common`
  - Updated the registration calls in `utopia-fs.ts` to use the project's ID as the scheme (so that VS Code knows to use this extension for that project)
  - Hooked in the messaging system, with a handler for each message type
    - The open file message will trigger a command in VS Code to open that file
    - Update decorations messages will update the highlight and selection decorations by calling `setDecorations` on the editor, which replaces the existing decorations with the new ones
    - Selected element changed messages will cause VS Code to reveal that range
  - Also subscribes to text selection changes in VS Code, and sends those as a message back to Utopia

**Note:**
As part of this, we discovered that because we were applying the monkey patching when first rendering the Canvas, it meant that the rest of the editor was first rendered as regular React components, and then after any change was re-rendered as patched components, and so were being treated as entirely new components, leading to the iframe being completely remounted. To fix this (as it was causing issues with the VS Code work) we moved the call to apply the monkey patching to `editor/src/templates/editor-entry-point.tsx`. This change caused the render counts in the tests to drop a bit.

**TODO Before Merging:**

- [x] Including `utopia-vscode-common` in build and test scripts in `shell.nix`
- [x] Update deploy scripts to also include that module


balazsbajorics:
> first rendered as regular React components, and then after any change was re-rendered as patched components, and so were being treated as entirely new components, leading to the iframe being completely remounted.

omg so _that's why_ the iframe remounted!!! you just fixed https://github.com/concrete-utopia/utopia/issues/813

maltenuhn:
> Processed messages aren't deleted, since that would lead to a race condition if the user had the same project open in two tabs

Nice! I can totally see how that would have caught us out at some point

utopia-vscode-extension/package.json
```diff
@@ -25,7 +25,8 @@
     "postinstall": "npm run download-api"
   },
   "dependencies": {
-    "browserfs": "1.4.3"
+    "browserfs": "2.0",
```
Rheeseyb, line 28:
```suggestion
```

utopia-vscode-extension/src/extension.ts
```diff
@@ -1,14 +1,205 @@
...
+  function handleMessage(message: ToVSCodeMessage): void {
+    if (isOpenFileMessage(message)) {
+      openFile(utopiaFS, vscode.Uri.joinPath(workspaceRootUri, message.filePath))
+    } else if (isUpdateDecorationsMessage(message)) {
+      currentDecorations = message.decorations
+      updateDecorations(currentDecorations)
+    } else if (isSelectedElementChanged(message)) {
+      revealRangeIfPossible(message.boundsInFile)
+    } else {
+      console.error(`Unhandled message type ${JSON.stringify(message)}`)
```
balazsbajorics, line 69: can you have our usual `const _exhaustiveCheck: never = message` here?

editor/src/core/vscode/vscode-bridge.ts
```diff
@@ -0,0 +1,230 @@
...
+    await clearBothMailboxes()
+    await writeProjectContents(projectID, projectContents)
+    initMailbox(UtopiaInbox, parseFromVSCodeMessage, (message: FromVSCodeMessage) => {
+      if (isEditorCursorPositionChanged(message)) {
+        dispatch(
+          [selectFromFileAndPosition(message.filePath, message.line, message.column)],
+          'everyone',
+        )
+      } else {
+        console.error(`Unhandled message type ${JSON.stringify(message)}`)
```
balazsbajorics, line 132: same question about `const _exhaustiveCheck: never = message`

utopia-vscode-extension/src/extension.ts
seanparsons: Probably can, originally the to and from messages shared a type.

utopia-vscode-extension/package.json
Rheeseyb: Done

utopia-vscode-extension/src/extension.ts
Rheeseyb: Done

editor/src/core/vscode/vscode-bridge.ts
Rheeseyb: We actually can't here because there is only one type for this message, so I've removed the `if` `else` statement since now if a new message type is added it will cause a compile error there instead now.

balazsbajorics (approved):

PR #943, opened 2021-02-26
fix: only update from code editor if the code file actually changed

balazsbajorics:
**Problem:**
After updating the model, a few moments later we would loose selection. 

Turns out the problem was `vscode-bridge.ts@watchForChanges` dispatches an `UPDATE_FROM_CODE_EDITOR` even if the update actually came _from_ the editor.

**Fix:**

Added a simple if to `UPDATE_FROM_CODE_EDITOR` to check if the action's code file matches the editorState's code file. in that case we don't need to update the editor state and we can just bail out.

### DIFF
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -3485,6 +3485,12 @@ export const UPDATE_FNS = {
     dispatch: EditorDispatch,
   ): EditorModel => {
     const existing = getContentsTreeFileFromString(editor.projectContents, action.filePath)
+
+    if (isTextFile(existing) && existing.fileContents.code === action.fileContents) {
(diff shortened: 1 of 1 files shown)


enidemi (approved):

Rheeseyb (approved):
Ok so this is then actually an issue in the watch functionality, in that the thing that writes the file shouldn't be informed of changes to it. Seems I need to fix it there, but this works as a quick fix in the mean time

balazsbajorics:
> Ok so this is then actually an issue in the watch functionality, in that the thing that writes the file shouldn't be informed of changes to it. Seems I need to fix it there, but this works as a quick fix in the mean time

I'm not sure. I think if you `fs.watch` in node for changes and _you_ also make some changes from the same process for example, you still get notifications

PR #945, opened 2021-03-01
Fix/prevent losing selected view

balazsbajorics:
**Problem:**
When the user edits the source code in the code editor, we'd lose the selections.

This started happening when we introduced hash-based (not printed to the code) UIDs. When the source code of a JSX Element changes, their hash changes, which means the uid will also change, which means their TemplatePath changes.
This would lead to selections (still pointing to the old TemplatePath) become invalid and deleted.

**Fix:**
We already had a mechanism to filter out nonexistent TemplatePaths from selections and highlights. I've changed that mechanism so that instead of filtering out the lists, I try to update the old TemplatePath to the new TemplatePath. This only works for changes where the parent element doesn't lose its TemplatePath. If the parent path can still be found in the source code, I try to find a template path at the same child index as the previous selection. So if you are editing the 2nd child of a View, the code will try to select the second child using the newly parsed uids instead of the old ones.

**Commit Details:**
- Added `findMatchingTemplatePath` that uses the childIndex of the old parent to try to find the edited element in the new parent
- it only works if the parentPath remains the same
- using `TemplatePathArrayKeepDeepEquality` to not change the selected views array if I don't have to (the performance regression tests caught me blowing the template paths, yay for tests)
- Instead of `stripNulls`, I use `flatMap` and a helper function that returns an optional result wrapped in an Array. I am also utilizing the new Typescript tuple types, so instead of `Array<TemplatePath>` (which can have any number of elements), I could type  it as `[] | [TemplatePath]` (so empty array or an array with a single element) which to me reads much better as an optional, and doesn't allow code which would return an array with multiple elements. I actually quite like this! 

**Note:**
I started out by looking at using the cursor position and highlight the edited code range, but unfortunately the highlight ranges use `uid`s and not template paths. For the full template paths we need the spy run at the end of the dispatch function. I didn't want to add temporary state to the dispatch function (something along the lines of "select this uid once the spy finishes") and I looked for an alternative solution instead.

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -799,5 +799,6 @@ function findMatchingTemplatePath(
     const newChildren = newAllPaths.filter((p) => TP.isChildOf(p, parentPath))
     const potentialNewPath = newChildren[oldChildIndex]
     return potentialNewPath == null ? [] : [potentialNewPath]
-  } else return []
+  }
+  return []
 }
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -83,6 +83,7 @@ import {
   DecorationRangeType,
 } from 'utopia-vscode-common'
 import { TemplatePathArrayKeepDeepEquality } from '../../../utils/deep-equality-instances'
+import { mapDropNulls } from '../../../core/shared/array-utils'
 
 export interface DispatchResult extends EditorStore {
   nothingChanged: boolean
@@ -758,20 +759,23 @@ function removeNonExistingViewReferencesFromState(
   const allPaths = MetadataUtils.getAllPaths(rootComponents)
   const updatedSelectedViews = TemplatePathArrayKeepDeepEquality(
     editorState.selectedViews,
-    editorState.selectedViews.flatMap((selectedView) =>
-      findMatchingTemplatePath(oldAllPaths, allPaths, selectedView),
+    mapDropNulls(
+      (selectedView) => findMatchingTemplatePath(oldAllPaths, allPaths, selectedView),
+      editorState.selectedViews,
     ),
   ).value
   const updatedHighlightedViews = TemplatePathArrayKeepDeepEquality(
     editorState.highlightedViews,
-    editorState.highlightedViews.flatMap((highlightedView) =>
-      findMatchingTemplatePath(oldAllPaths, allPaths, highlightedView),
+    mapDropNulls(
+      (highlightedView) => findMatchingTemplatePath(oldAllPaths, allPaths, highlightedView),
+      editorState.highlightedViews,
     ),
   ).value
   const updatedHiddenInstances = TemplatePathArrayKeepDeepEquality(
     editorState.hiddenInstances,
-    editorState.hiddenInstances.flatMap((hiddenInstance) =>
-      findMatchingTemplatePath(oldAllPaths, allPaths, hiddenInstance),
+    mapDropNulls(
+      (hiddenInstance) => findMatchingTemplatePath(oldAllPaths, allPaths, hiddenInstance),
+      editorState.hiddenInstances,
     ),
   ).value
   return {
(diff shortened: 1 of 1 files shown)


editor/src/components/editor/store/dispatch.tsx
```diff
@@ -747,12 +748,32 @@ function notifyTsWorker(
   }
 }
 
-function removeNonExistingViewReferencesFromState(editorState: EditorState): EditorState {
+function removeNonExistingViewReferencesFromState(
+  oldEditorState: EditorState,
+  editorState: EditorState,
+): EditorState {
+  const oldRootComponents = oldEditorState.jsxMetadataKILLME
```
Rheeseyb, line 755: I think at this point we should probably just rename `jsxMetadataKILLME` to `jsxMetadata` :smile:

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -761,14 +782,22 @@ function removeNonExistingViewReferencesFromState(editorState: EditorState): Edi
...
   }
+  const parentPath = TP.parentPath(pathToUpdate)
+  const parentPathStillExists = newAllPaths.findIndex((p) => TP.pathsEqual(p, parentPath)) > -1
+  if (parentPathStillExists) {
+    const oldChildren = oldAllPaths.filter((p) => TP.isChildOf(p, parentPath))
+    const oldChildIndex = oldChildren.findIndex((p) => TP.pathsEqual(p, pathToUpdate))
+    const newChildren = newAllPaths.filter((p) => TP.isChildOf(p, parentPath))
+    const potentialNewPath = newChildren[oldChildIndex]
+    return potentialNewPath == null ? [] : [potentialNewPath]
+  } else return []
```
Rheeseyb, line 802: Can you wrap this all in an `else` block please

balazsbajorics: whoa who wrote this code

seanparsons (approved):

Rheeseyb (approved):
This looks good to me but potentially expensive - have you checked that this doesn't cause a performance regression when typing?

balazsbajorics:
> This looks good to me but potentially expensive - have you checked that this doesn't cause a performance regression when typing?

FYI `removeNonExistingViewReferencesFromState` (which I modified) runs at the end of the dispatch block – for every single action! 

However, I don't think performance is a concern, I essentially replaced a single array walkthrough (filter) with ~6 walkthroughs (a flatmap, two filters, two findIndexes and the keep equality call). But these arrays are usually very short (selectedViews is usually 0 or 1 long, allPaths is almost never longer than 100) so I'm not worried. The actual comparisons are always TP.pathsEquals which is pretty fast as well.

seanparsons:
> However, I don't think performance is a concern, I essentially replaced a single array walkthrough (filter) with ~6 walkthroughs (a flatmap, two filters, two findIndexes and the keep equality call). But these arrays are usually very short (selectedViews is usually 0 or 1 long, allPaths is almost never longer than 100) so I'm not worried. The actual comparisons are always TP.pathsEquals which is pretty fast as well.

That's quite a lot of intermediate arrays, we could probably avoid some of that by using `mapDropNulls` instead of `Array.flatmap` because it's designed for the case of filtering and transforming in one step.

maltenuhn:
> This looks good to me but potentially expensive - have you checked that this doesn't cause a performance regression when typing?

Could we put a performance test in place so we can establish a baseline and also reason about it with (at least some) data?

enidemi (approved):

balazsbajorics:
I've done some performance measurements with our standard large stress test project, and based on the measurements I've pushed a new version of the function that removes all array filtering. This new version +-5% takes the same amount of time than the original function on `master` (whereas my previous array filtering method was about 2x slower)

enidemi (approved):

Rheeseyb:
> I've done some performance measurements with our standard large stress test project, and based on the measurements I've pushed a new version of the function that removes all array filtering. This new version +-5% takes the same amount of time than the original function on `master` (whereas my previous array filtering method was about 2x slower)

Thanks for doing this!

PR #946, opened 2021-03-01
fix(canvas) use cssparser to have parsed result of a prop while resizing/moving

enidemi:
**Problem:**
The canvas breaks when resizing/moving an element with an expression or string value. It is possible that the target frameProp is a string like `5px`, where it just wanted to add the resize delta to the string. 

**Fix:**
The frame prop calculation uses the cssparser to have a correct parsed result of a prop. Only %, px, and unitless number is supported for now. Dragging with other units and expressions keeps the element in place.

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -817,13 +817,17 @@ function updateFrameValueForProp(
         value: jsxAttributeValue(delta, emptyComments),
       }
     }
-    const parsedProp = parseCSSLengthPercent(existingProp)
-    if (isRight(parsedProp)) {
-      const pinIsPercentage = parsedProp.value.unit === '%'
-      const pinIsUnitlessOrPx = parsedProp.value.unit == null || parsedProp.value.unit === 'px'
+    const parsedProp = foldEither(
+      (_) => null,
+      (r) => r,
+      parseCSSLengthPercent(existingProp),
+    )
+    if (parsedProp != null) {
+      const pinIsPercentage = parsedProp.unit === '%'
+      const pinIsUnitlessOrPx = parsedProp.unit == null || parsedProp.unit === 'px'
       let valueToUse: string | number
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -810,23 +811,38 @@ function updateFrameValueForProp(
...
+        value: jsxAttributeValue(delta, emptyComments),
       }
-    } else {
-      valueToUse = existingProp == null ? delta : Number(existingProp) + delta
     }
-    return {
-      path: createLayoutPropertyPath(pinnedPropForFramePoint(framePoint)),
-      value: jsxAttributeValue(valueToUse, emptyComments),
+    const parsedProp = parseCSSLengthPercent(existingProp)
+    if (isRight(parsedProp)) {
```
seanparsons, line 821: `foldEither` here would save you from have to write `.value` in a bunch of places.

seanparsons (approved):

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -810,23 +811,42 @@ function updateFrameValueForProp(
...
-      value: jsxAttributeValue(valueToUse, emptyComments),
+    const parsedProp = foldEither(
+      (_) => null,
+      (r) => r,
+      parseCSSLengthPercent(existingProp),
+    )
+    if (parsedProp != null) {
+      const pinIsPercentage = parsedProp.unit === '%'
+      const pinIsUnitlessOrPx = parsedProp.unit == null || parsedProp.unit === 'px'
+      let valueToUse: string | number
```
balazsbajorics, line 828: this valueToUse could be moved into the `if (pinIsPercentage)` body to limit scope

enidemi: fixed

balazsbajorics (approved):

PR #950, opened 2021-03-03
Support unsaved files in VS Code

### LINKED ISSUE #939: Support unsaved files with vs code

Rheeseyb:
Fixes #939 

**Problem:**
We want to support unsaved changes to files in Utopia, and keep those in sync with VS Code.

**Fix:**
Since VS Code doesn't write unsaved changes to file (only to its own cached user state), and since there is no way to instruct VS Code to discard the unsaved changes to a specific file, this meant we had to get a bit inventive. The fix can be broken down into a few parts:

**Changes to the core FS:**
- Include unsaved content in the stored file:
  - `FSFile` type now includes a nullable `unsavedContent` field
  - `FSNode` now includes `lastSavedTime` so that we can provide this to VS Code when it requests the `stats` of a file
  - `FSNode` now includes `sourceOfLastChange`, which is used so that Utopia and VS Code can determine if they were the ones to last write to a file
  - Initialising the file system now takes a `FSUser` to track which user (Utopia or VS Code) is using this filesystem instance, and uses that when writing files
  - The `watch` functionality's `onModified` callback now also includes whether a file modification came from this user
  - Added functions for reading and writing the entire file (saved and unsaved content), or either the saved or unsaved content individually
  - The mailbox just uses the saved content field in all cases, and so acts as before

**Changes to Utopia**
- Update the `UPDATE_FROM_CODE_EDITOR` action to include saved and unsaved content changes
- The `DELETE_FILE` action had an early return which has been moved even earlier, so that we don't waste time when the fs watch function tells Utopia a file has been deleted (there's no way of tracking which fs user instigated a deletion)
- Before calling `applyProjectContentChanges` in the bridge (which writes project file changes to the fs), we check if the action that resulted in the change came from VS Code (to save us re-writing the change back into the fs)
  - We also check if actions from VS Code are mixed with actions from Utopia, and log an error if that's the case, since this shouldn't be the case ever but we don't want to hard fail it yet
- In `vscode-bridge` we:
  - Write both the saved and unsaved content to the fs
  - Ignore file creation or modification changes coming from the fs watch callback if the file was last modified by Utopia
  - Ensure that we only write changes to the fs if the saved or unsaved content was changed, carefully handling an edge case where the project file's unsaved content has a change in the parsed model (i.e. it was updated via the Canvas or Inspector), but that change hasn't yet generated the new code

**Changes to VS Code**
- Since VS Code doesn't provide a way to forcibly discard unsaved changes in a specific file and replace them with the latest version "on disk", we had to create a new command in our VS Code fork for doing exactly this. This command is closely based on an existing command that VS Code provides for reverting the currently active text editor, but instead takes the file's path (the resource's URI)
- Switched the default autosave behaviour to "off" and restored the ability for user's to change this setting

**Changes to The VS Code Extension**
- In `utopia-fs` we:
  - Provide event queues for changes to saved or unsaved content coming from Utopia only (i.e. we filter out any changes that came from VS Code)
  - Only read and write saved content, since this is the way in which VS Code expects to use this API
  - Use a file's last saved time (rather than it's `mtime`) in the `stat` call made by VS Code, so that VS Code doesn't wrongly think a file has changed when just the unsaved content had changed (otherwise this would cause VS Code to fail any saves it tries to make)
- In `extension.ts` we:
  - Subscribe to changes coming from Utopia, via the above new event queues in `utopia-fs`
    - If the saved content has changed, it means the user had manually saved the file in Utopia, which means we need to force VS Code to bin off its record of unsaved changes, and load the saved version from the filesystem. We do this by telling VS Code to execute the new command we added into our VS Code fork, which "reverts" the file (meaning it loads the version on disk)
    - If the unsaved content has changed, we need to create a `WorkspaceEdit`, which contains instructions for VS Code to replace a given range in a given file with new text. We provide this with a range that will cause the entire file to be replaced, and give it the new unsaved content to replace it with.
  - Subscribe to changes coming from VS Code, via VS Code's `onDidChangeTextDocument` and `onDidCloseTextDocument` events, in both cases filtering to only the documents relevant to utopia:
    - We are only interested in document changes if the document is dirty (i.e. it has unsaved content). In this case we add the dirty document to a `Set` tracking all open dirty documents, so that we can also react to them being reverted in the `onDidChangeTextDocument` event (in which case we must clear the unsaved changes from the FS and Utopia)
    - If a dirty document changes, we write the full text content to the `unsavedContent` field in that file in the fs
    - To prevent duplicated writes as a result of changes coming from Utopia, we use a `Set` of file names (`incomingFileChanges`), which we add a file path to when processing a change from Utopia, and remove the file path in our handling of the corresponding `onDidChangeTextDocument` event.
      - In the theoretical case where the workspace edit for an update from Utopia failed, we clear this flag also


editor/src/components/editor/actions/actions.ts
```diff
@@ -3486,21 +3486,19 @@ export const UPDATE_FNS = {
   ): EditorModel => {
     const existing = getContentsTreeFileFromString(editor.projectContents, action.filePath)
 
-    if (isTextFile(existing) && existing.fileContents.code === action.fileContents) {
```
Rheeseyb, line 3489: This check (introduced in https://github.com/concrete-utopia/utopia/pull/943) is no longer necessary, since we now only dispatch this action on changes that actually came from VS Code (and therefore we _know_ that the code changed)

editor/src/core/vscode/vscode-bridge.ts
```diff
@@ -71,7 +71,9 @@ async function writeProjectFile(
       return ensureDirectoryExists(toFSPath(projectPath))
     }
     case 'TEXT_FILE': {
-      return writeFileAsUTF8(toFSPath(projectPath), file.fileContents.code)
+      const savedContent = file.lastSavedContents?.code ?? file.fileContents.code
```
enidemi, line 74: minor suggestion: this could use the `getSavedCodeFromTextFile` helper

editor/src/components/editor/actions/actions.ts
```diff
@@ -3584,6 +3582,12 @@ export const UPDATE_FNS = {
     userState: UserState,
   ): EditorModel => {
     const file = getContentsTreeFileFromString(editor.projectContents, action.filename)
+
+    // Don't delete package.json, otherwise it will bring about the end of days.
```
balazsbajorics, line 3586: 😹

enidemi (approved):

balazsbajorics (approved):

editor/src/core/vscode/vscode-bridge.ts
Rheeseyb: Good catch! Done :+1:

editor/src/components/editor/actions/actions.ts
Rheeseyb: Ha this comment was already in the code, we had just moved the associated line up a few lines

PR #951, opened 2021-03-03
Fix/canvas element context menu

enidemi:
This PR adds more functionality to the contextmenu submenu that lists the elements under cursor.
- moved the submenu to the top of the context menu
- rename submenu to 'Select element'
- mouseover highlights for elements
- icons

Changes:
- New hook to get element label, type and icon
- inner parts from the navigator item-preview moved to a common function to get the icon type
- in a future PR TODO is cleaning up navigator props, because it could use this nice new hook

### DIFF
--- a/editor/src/components/inspector/common/name-and-icon-hook.ts
+++ b/editor/src/components/inspector/common/name-and-icon-hook.ts
@@ -1,0 +1,171 @@
+import { MetadataUtils } from '../../../core/model/element-metadata-utils'
+import {
+  isAnimatedElementAgainstImports,
+  isEllipseAgainstImports,
+  isHTMLComponent,
+  isImg,
+  isRectangleAgainstImports,
+  isTextAgainstImports,
+  isViewAgainstImports,
+} from '../../../core/model/project-file-utils'
+import {
+  JSXElementName,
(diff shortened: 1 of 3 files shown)


seanparsons (approved):

Rheeseyb (approved):

editor/src/components/inspector/common/name-and-icon-hook.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export interface NameAndIconResult {
+  path: TemplatePath
+  name: JSXElementName | null
+  label: string
+  iconProps: IcnProps
+}
+
+export function useNameAndIcon(path: TemplatePath): NameAndIconResult {
+  const metadata = useEditorState(
```
balazsbajorics, line 33: this will trigger a re-render of the user component any time the metadata or components change. I think a better solution would be to move `getNameAndIconResult` into the selector function (given that it has acceptable performance), and return the result as the selection. that would mean the user component is only rerendered if the selected NameAndIconResult changed. (there's an optional equality function you can provide to useEditorState to avoid blowing equality when two NameAndIconResults are identical)

PR #952, opened 2021-03-04
VS Code message triggered to get data on startup.

seanparsons:
**Problem:**
At load or reload, VS Code doesn't get sent the current state of things (for selection and the like), but is instead only sent that information on changes.

**Fix:**
Have the VS Code extension trigger a message which causes the main part of Utopia to send the current state of affairs back to the extension.

**Commit Details:**
- Added `SEND_CODE_EDITOR_INITIALISATION` action, which sends
  a messages to VS Code for the decorations and range that should
  be revealed.
- Refactored out `sendCodeEditorDecorations` and `sendSelectedElement`
  from `dispatch.tsx` so they can be used elsewhere.
- VS Code requests the initialisation data with the `SendInitialData`
  message.

### DIFF
--- a/editor/src/core/vscode/vscode-bridge.ts
+++ b/editor/src/core/vscode/vscode-bridge.ts
@@ -165,6 +165,9 @@ export async function initVSCodeBridge(
               'everyone',
             )
             break
+          default:
+            const _exhaustiveCheck: never = message
+            throw new Error(`Unhandled message type${JSON.stringify(message)}`)
         }
       })
       watchForChanges(dispatch)
--- a/editor/src/core/vscode/vscode-bridge.ts
+++ b/editor/src/core/vscode/vscode-bridge.ts
@@ -12,6 +12,7 @@ import {
   selectFromFileAndPosition,
   markVSCodeBridgeReady,
   updateFromCodeEditor,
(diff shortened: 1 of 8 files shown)


editor/src/core/vscode/vscode-bridge.ts
```diff
@@ -146,10 +155,17 @@ export async function initVSCodeBridge(
...
+        switch (message.type) {
+          case 'SEND_INITIAL_DATA':
+            dispatch([sendCodeEditorInitialisation()], 'everyone')
+            break
+          case 'EDITOR_CURSOR_POSITION_CHANGED':
+            dispatch(
+              [selectFromFileAndPosition(message.filePath, message.line, message.column)],
+              'everyone',
+            )
+            break
```
balazsbajorics, line 167: could you add our usual catchall `_exhaustiveCheck` here?

balazsbajorics (approved):

Rheeseyb (approved):

editor/src/core/vscode/vscode-bridge.ts
seanparsons: Fixed in 65d34bc6.

PR #955, opened 2021-03-05
Remove Monaco from the editor.

### LINKED ISSUE #887: Remove the feature toggle for non-iframe monaco
The iframe code editor is enabled by default, but I left in the feature switch to disable it, reverting to the non-iframe version. 
The two options increase complexity and cost of maintenance of the editor, so if we all agree that the iframe version is the one true version, we should remove the other one.

seanparsons:
Fixes #887

**Problem:**
Monaco is in the codebase.

**Fix:**
Remove Monaco from the codebase.

**Commit Details:**
- Fixes #887.
- Removed all the Monaco related dependencies.
- Removed the webpack support for Monaco.
- Removed the feature flags for iframe Monaco and to
  turn on VS Code.
- Removed the `SetCodeEditorTheme` action.
- Removed `codeEditorTheme` from `EditorState`.
- Fixed an issue with `parseFromVSCodeMessage` where it didn't
  handle the `SendInitialData` message.
- Removed Monaco themes, the override CSS and supporting source
  for Monaco.


Rheeseyb:
I'm assuming this still leaves the utopia tabbing system in?

Rheeseyb:
Also

> Problem:
Monaco is in the codebase.

LOL!

seanparsons:
@Rheeseyb Yup, this just removes Monaco as a thing.

Rheeseyb (approved):
This is just a thing of beauty

enidemi (approved):

PR #958, opened 2021-03-05
Ensure we set the webview endpoint for VS Code

Rheeseyb:
**Problem:**
Whilst figuring out how to display the welcome page, I discovered that loading webviews would hit MS servers.

**Fix:**
Set the webview endpoint so that our app is the one handling those requests. Also included in this is some work to support loading utopia urls using those webviews (as a precursor to the welcome page)

### DIFF
--- a/vscode-build/vscode.patch
+++ b/vscode-build/vscode.patch
@@ -36,11 +36,24 @@ index 9e26dfeeb6..0000000000
 @@ -1 +0,0 @@
 -{}
 \ No newline at end of file
+diff --git a/extensions/simple-browser/src/simpleBrowserView.ts b/extensions/simple-browser/src/simpleBrowserView.ts
+index 104acfe955..123e8e79f5 100644
+--- a/extensions/simple-browser/src/simpleBrowserView.ts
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):
popups!

enidemi (approved):

Rheeseyb:
> popups!

Ha! The popup support is so that we can open new browser tabs from webviews, so that we can support the links to sample apps in the welcome tab

vscode-build/vscode.patch
```diff
@@ -186,13 +199,14 @@ index 9cddae242a..35400d5d59 100644
 +    "applicationName": "code-web",
 +    "dataFolderName": ".vscode-web",
 +    "version": "1.53.0"
-+  }
++  },
++  "webviewEndpoint": "http://localhost:8000/vscode/vscode/vs/workbench/contrib/webview/browser/pre"
```
Rheeseyb, line 203: Oops this line needs to go because it was moved to the `workbench.ts` file just below

seanparsons (approved):

PR #959, opened 2021-03-09
Fix to selection retention logic

### LINKED ISSUE #940: Selection is always lost after any model change
Believed to be caused by UIDs now being a hash of the code

Rheeseyb:
Fixes #940 and fixes a `forceNotNull` bug in the Inspector

**Problem:**
When making a code change we were trying to keep the selected template path, but the path we were keeping was now non-existent. The logic also failed for root elements of scenes.

**Fix:**
Fix the check that would return the original path, and use existing functions to get all children of the existing parent so that we also handle root elements.

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -776,7 +776,11 @@ function findMatchingTemplatePath(
 
   const parentPath = TP.parentPath(pathToUpdate)
   if (parentPath == null) {
-    return null
+    const oldRootPaths = MetadataUtils.getAllCanvasRootPaths(oldComponents)
+    const newRootPaths = MetadataUtils.getAllCanvasRootPaths(oldComponents)
+    const oldRootIndex = oldRootPaths.findIndex((p) => TP.pathsEqual(p, pathToUpdate))
+    const newRootPath = newRootPaths[oldRootIndex]
+    return newRootPath ?? null
   } else {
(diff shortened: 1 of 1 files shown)


seanparsons (approved):

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -770,24 +770,20 @@ function findMatchingTemplatePath(
...
-  const newParent = MetadataUtils.getElementByInstancePathMaybe(
-    newComponents.elements,
-    parentPath as InstancePath,
-  )
-  if (oldParent != null && newParent != null) {
-    const oldChildIndex = oldParent.children.findIndex((p) => TP.pathsEqual(p, pathToUpdate))
-    const potentialNewPath = newParent.children[oldChildIndex]
-    return potentialNewPath ?? null
+  if (parentPath == null) {
+    return null
```
balazsbajorics, line 779: maybe it's worth leaving a comment here that this means we will not even attempt to salvage selection for root elements

```suggestion
    return null // TODO this means we don't attempt to salvage the selection of root elements
```

balazsbajorics (approved):

editor/src/components/editor/store/dispatch.tsx
Rheeseyb: Good catch! I've updated the logic now to handle root paths

PR #960, opened 2021-03-09
Removed Utopia tabs.

seanparsons:
**Problem:**
There are too many tab bars in the editor.

**Fix:**
Mostly this removes the editor tabs (visual and model) and then fixes the fallout from those changes. In doing that the seeds of the model split between separate code editors and canvas designer has been laid so that the two things are independent rather than being tied directly in what is shown.

**Commit Details:**
- Renamed `OpenEditorTab` action to be `OpenCodeEditorFile`.
- Renamed `CloseEditorTab` action to be `CloseDesignerFile`.
- Removed `ReorderEditorTabs` action.
- Moved the call to send an open file message to the newly renamed
  `OPEN_CODE_EDITOR_FILE` action.
- Removed `openFiles` and `selectedFile` from `EditorState`.
- Deleted `editor-tabs.ts` and `file-tabs.tsx`.
- Hardcoded the storyboard path for the canvas to be `/utopia/storyboard.js`.
- Removed migrations test as it was creating a modern model and forcing
  it through the migration steps which is completely incompatible with
  the logic.
- Added a migration step to move the first storyboard containing file it finds to
  `/utopia/storyboard.js`.


editor/src/components/editor/actions/migrations/migrations.spec.ts
```diff
@@ -8,11 +8,4 @@ describe('Migrations', () => {
     const migrated = applyMigrations(emptyProject)
     expect(migrated.projectVersion).toEqual(CURRENT_PROJECT_VERSION)
   })
-
-  it('returns with the latest project version for a non-versioned project', () => {
```
Rheeseyb, line 12: Hmmmm this test is definitely wrong, but this also means we have no test to ensure that the migrations actually run for an old project :/ Is there any chance you can grab an old model and hardcode that in here? I'm concerned that there's a chance this PR will destroy prior projects.

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -60,8 +60,8 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(400) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(410)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(280) // if this breaks, GREAT NEWS but update the test please :)
+    expect(renderCountAfter - renderCountBefore).toBeLessThan(290)
```
Rheeseyb, line 64: OMG!

editor/src/components/editor/actions/migrations/migrations.spec.ts
seanparsons: The problem is that I'd need to get one from before all the migrations were implemented, which is a long time ago now.

editor/src/core/performance/performance-regression-tests.spec.tsx
seanparsons: Yeah, that, almost suspiciously low.

editor/src/components/editor/actions/migrations/migrations.spec.ts
seanparsons: I did manage to find a project from a couple of months ago and that seems to work with the migration I added in this PR.

PR #963, opened 2021-03-10
Feature/component button

OmarDaSilva:
*Problems*
- we need a prominent way to indicate if something you're editing is (or can be) "focused" (per terminology of the spike), even if the navigator is hidden
- we have component-related functionality that's currently buried in the UI in the header section, but that - in a component-centric world - becomes more important; notably being able to change what a component is being "rendered as" (e.g. swapping out a `FlexRow` for a `FlexColumn`, a div for a `Button` ...)

*Solution:* The Component Unibutton.

image

- introduces a prominent button atop the canvas, next to the formula bar
- The button (a) indicates whether you're editing is an element, a component instance, or an "isolated" / "focused" component child of a component instance, (b) lets you isolate / focus a component - similar to the context menu entry in the spike
- The button also shows a drop-down menu with the breadcrumb trail, and the option to change the component to another component (the "Render as" capability formerly shown in the inspector)

Other changes 
- clean up inspector remove the `header-section` that formerly housed these
- introduces element icons in an orange colorway for children-of-focused-components (to align with Sketch's treatment of these)
- UX improvements to the PopupList used here (was using wrong mouse pointers, and not opening when clicking on the label)
- Clean up the formula bar and container, since they didn't behave well with siblings / having > 1 child respectively

NB: the button currently does not "do anything", since that will require functionality provided by the spike. Plus, the button has a nice background animation.


OmarDaSilva:
This PR is ready and I would like feedback!

Rheeseyb:
> Plus, the button has a nice background animation.

:grinning: love it!

editor/src/components/canvas/controls/breadcrumb-trail.tsx
```diff
@@ -0,0 +1,95 @@
...
+  useKeepReferenceEqualityIfPossible,
+} from '../../../utils/react-performance'
+import { Utils } from '../../../uuiui-deps'
+import * as TP from '../../../core/shared/template-path'
+import { getOpenUtopiaJSXComponentsFromState } from '../../editor/store/editor-state'
+import { MetadataUtils } from '../../../core/model/element-metadata-utils'
+import { selectComponents } from '../../../components/editor/actions/action-creators'
+import { FlexRow, Icons } from '../../../uuiui'
+
+export const BreadcrumbTrail = () => {
```
Rheeseyb, line 15: Please use `betterReactMemo` on all components you define. It improves performance and also gives them a nice label to make debugging easier! (e.g. https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/canvas-component-entry.tsx#L21)

editor/src/components/canvas/controls/breadcrumb-trail.tsx
```diff
@@ -0,0 +1,95 @@
...
+            elements.push({
+              name: scene.label,
+              path: path,
+            })
+          }
+        }
+      })
+      return elements
+    }, [selectedViews, jsxMetadataKILLME]),
+  )
```
Rheeseyb, line 65: I see this is copy pasted code from `inspector.tsx`. Please can you extract it into a hook so that both files use the same code path. The only rule with custom hooks is that they have to have their names must have the `use` prefix, otherwise they're just regular functions.

editor/src/components/canvas/controls/breadcrumb-trail.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const ElementPathButtons = betterReactMemo('ElementPathButtons', () => {
```
Rheeseyb, line 67: Please pull this component up to the top level as otherwise we're creating inner components here that will cause issues with performance and memoisation.

editor/src/components/canvas/controls/namerowcrumbs.tsx
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    const onElementTypeChange = React.useCallback(
+        (value: JSXElementName) => {
+            const actions = refElementsToTargetForUpdates.current.map((path) => {
+                return EditorActions.updateJSXElementName(path, value)
+            })
+            dispatch(actions, 'everyone')
+        },
+        [dispatch, refElementsToTargetForUpdates],
```
Rheeseyb, line 51: You don't need to include `refElementsToTargetForUpdates` here since it is a ref.

editor/src/components/canvas/controls/namerowcrumbs.tsx
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+    const label: string = 'Test'
+    const type: string = 'TestType'
+    const hookResult = useNamesAndIconsSelectedViews()
+
+    const constrolStatus = 'simple'
+    const controlStyles = getControlStyles(constrolStatus)
+
+    const NameRow = betterReactMemo('NameRow', () => {
```
Rheeseyb, line 61: Again, please avoid inner components and move this to the top.

editor/src/components/editor/component-button.tsx
```diff
@@ -0,0 +1,202 @@
...
+    enum editContext {
+      ELEMENT = 0,
+      COMPONENT_INSTANCE,
+      COMPONENT,
+      COMPONENT_CHILD,
+    }
+
+    const [currentEditContext, setCurrentEditContext] = React.useState(editContext.COMPONENT)
+
+    const TODOREPLACEMEcycleEditContext = () => {
```
Rheeseyb, line 41: I see you come from the @balazsbajorics school of naming conventions :grinning:

editor/src/components/editor/top-menu.tsx
```diff
@@ -46,15 +46,19 @@ export const TopMenu = betterReactMemo('TopMenu', () => {
...
       <Tooltip title={'Toggle outline'} placement={'bottom'}>
-        <RightMenuTile
-          selected={navigatorPosition !== 'hidden'}
-          highlightSelected={false}
-          icon={<MenuIcons.Navigator />}
+        <SquareButton
+          spotlight={false}
+          highlight={true}
+          // selected={navigatorPosition !== 'hidden'}
+          // highlightSelected={false}
```
Rheeseyb, line 55: Can these commented lines be deleted?

editor/src/components/inspector/sections/style-section/style-section.tsx
```diff
@@ -8,6 +8,14 @@ import { ShadowSubsection } from './shadow-subsection/shadow-subsection'
...
 import { AutosizingTextSubsection } from './text-subsection/text-subsection'
 import { TransformSubsection } from './transform-subsection/transform-subsection'
+import {
+  FlexRow,
+  InspectorSectionHeader,
+  InspectorSubsectionHeader,
+  Section,
+} from '../../../../uuiui'
+import { GridRow } from '../../widgets/grid-row'
+import { InspectorContextMenuWrapper } from '../../../../uuiui-deps'
```
Rheeseyb, line 18: Seems these imports weren't used in the end so can go in the bin

editor/src/uuiui/button.tsx
```diff
@@ -36,25 +37,29 @@ export const Button = styled.div<ButtonProps>((props: ButtonProps) => ({
...
   opacity: props.disabled ? 0.5 : 1,
   pointerEvents: props.disabled ? 'none' : 'initial',
   boxShadow: props.outline ? 'inset 0px 0px 0px 1px hsl(0,0%,90%)' : undefined,
   color: props.primary ? 'white' : 'inherit',
-  background: props.primary ? UtopiaStyles.backgrounds.blue : 'transparent',
+  background: props.primary
+    ? UtopiaStyles.backgrounds.blue
+    : props.spotlight
+    ? UtopiaTheme.color.secondaryBackground.value
+    : undefined,
```
Rheeseyb, line 48: Nested ternaries?! You monsters!

editor/src/uuiui/widgets/popup-list/popup-list.tsx
```diff
@@ -321,7 +323,7 @@ const MenuPortal = (props: MenuPortalProps<SelectOption>) => {
           minWidth: 150,
           maxWidth: 250,
           boxShadow: 'rgba(0, 0, 0, 0.1) 0px 0px 0px 1px, rgba(0, 0, 0, 0.1) 0px 4px 11px',
-          zIndex: 1,
+          zIndex: 999999,
```
Rheeseyb, line 326: erm...

editor/src/uuiui/widgets/popup-list/popup-list.tsx
```diff
@@ -388,15 +390,24 @@ const MenuList = (props: MenuListComponentProps<SelectOption>) => {
...
+    components.DropdownIndicator && (
+      <components.DropdownIndicator {...indicatorProps}>
+        <SmallerIcons.ExpansionArrowDown />
+      </components.DropdownIndicator>
+    )
+  )
 }
 
 const SingleValue = (props: SingleValueProps<SelectOption>) => {
+  console.log('single value props', props)
```
Rheeseyb, line 406: Delete me please

editor/src/uuiui/widgets/popup-list/popup-list.tsx
```diff
@@ -544,13 +556,24 @@ export const PopupList = betterReactMemo<PopupListProps>(
...
+            display: 'flex',
             alignItems: 'center',
             justifyContent: 'center',
             flexShrink: 0,
+            '&:hover': {
+              filter: 'brightness(.98)',
+            },
+            '&:active': {
+              filter: 'brightness(.95)',
+            },
```
Rheeseyb, line 576: Have you checked that the changes in this file work with all other cases of the popup-list in the editor?

editor/src/components/canvas/controls/breadcrumb-trail.tsx
maltenuhn: The inspector _I think_ will not need this after this change, because it's only used for the breadcrumbs. @OmarDaSilva did you already remove the now-unused `header-section` etc from the inspector?

editor/src/components/editor/component-button.tsx
maltenuhn: (this function is the one that will need to be replaced entirely once we're able to "focus" components)

editor/src/uuiui/button.tsx
maltenuhn: @OmarDaSilva this is my doing, I'll fix it. We have a coding convention against nested ternaries which I always ~choose to~ forget :)

editor/src/uuiui/widgets/popup-list/popup-list.tsx
maltenuhn: @Rheeseyb Yep, the changes are actually bug fixes.

maltenuhn:
@Rheeseyb thanks for your feedback! Some of the things you spotted were actually present in the code before (eg the nested components), so hopefully making the changes will further help with performance / prevent unnecessary re-rendering :) 

This PR **moves** some functionality out of the inspector, specifically pretty much the entire contents of the `header-section` (as well as the props that we pass down into the various nested inspector, inner-inner-inspector etc components). I don't think that's been completed yet, which accounts for some of the duplication here. @OmarDaSilva did you already remove some inspector code for this PR?

editor/src/components/canvas/controls/breadcrumb-trail.tsx
Rheeseyb: Ah ok if the other is being binned off then it's fine to leave this as is

editor/src/components/canvas/controls/breadcrumb-trail.tsx
balazsbajorics: since the main component seems to be only rendering this, I think it's even easier to just "inline" this component. ie just render it without making it a component first.

editor/src/components/canvas/controls/namerowcrumbs.tsx
balazsbajorics: unfortunately eslint doesn't know that, and it doesn't cause any problems if it is included. I've done this to pseudo-refs myself too

editor/src/components/canvas/controls/namerowcrumbs.tsx
balazsbajorics: again, why is this an inner component if the outer component's sole job is rendering the inner one?

editor/src/components/canvas/controls/breadcrumb-trail.tsx
maltenuhn: This was legacy inspector code that I just asked Omar to copy over. Good occasion to clean it up!

editor/src/components/canvas/controls/namerowcrumbs.tsx
maltenuhn: Legacy code, but we'll fix it :D

balazsbajorics: https://github.com/concrete-utopia/utopia/blob/d4c971fc0587cda5c880220a270e09e0191172ee/editor/src/components/inspector/sections/header-section/name-row.tsx#L18

on master this is a top level component, so this is definitely not a problem with legacy code ;)

editor/src/components/editor/component-button.tsx
balazsbajorics: this is the way.

editor/src/components/canvas/controls/namerowcrumbs.tsx
maltenuhn: I stand corrected!

editor/src/components/inspector/common/name-and-icon-hook.ts
```diff
@@ -53,6 +50,19 @@ export function useNameAndIcon(path: TemplatePath): NameAndIconResult {
...
 }
 
+export function useNamesAndIconsSelectedViews(): NameAndIconResult[] {
+  return useEditorState(
+    (store) => {
+      const metadata = store.editor.jsxMetadataKILLME
+      const components = getOpenUtopiaJSXComponentsFromState(store.editor)
+      const imports = getOpenImportsFromState(store.editor)
+      const selectedViews = store.editor.selectedViews
+      return selectedViews.map(path => getNameAndIconResult(path, components, metadata, imports))
```
balazsbajorics, line 60: by default `useEditorState` uses a shallow equality comparison. so if you return an array of stuff, it will use `===` comparison for each array item. 

however `getNameAndIconResult` will always return a new object, so even if you call it with the same parameters,
it will never satisfy `===` equality, and it will always trigger a re-render of the component using this hook, for every single editor state change.

One trick is while developing a new component or new hook, I recommend turning on React Devtool's "Highlight updates when components render" feature.



That will flash your component whenever it re-renders. If you see your component flashing while you are scrolling the canvas for example, that means that your component is re-rendering too often. _Ideally_ it should only flash when it actually changes what is displayed. It is sometimes too complicated to achieve perfection, so you can use deliberation and decide how much optimization is necessary. But we always want to avoid a component which re-renders on every change!

There's two potential solutions:
one is to use memoization on this selector (the function passed into useEditorState is called a selector). There's some selector memoization helpers called `reselect` that we use in some places in the codebase: https://github.com/concrete-utopia/utopia/blob/d4c971fc0587cda5c880220a270e09e0191172ee/editor/src/components/canvas/controls/new-canvas-controls.tsx#L211-L248

The second (less good, but sometimes necessary) option is to roll your own equality function. You can provide a third (optional) parameter to `useEditorState` which is an equality function in the form of `(oldSlice, newSlice) => boolean`, if this equality function returns true, we consider the old state slice and the new state slice to be unchanged, and we do not trigger a re-render of the element. Here's an example for that: https://github.com/concrete-utopia/utopia/blob/d4c971fc0587cda5c880220a270e09e0191172ee/editor/src/components/inspector/common/name-and-icon-hook.ts#L34-L54

editor/src/components/inspector/sections/header-section/header-section.tsx
```diff
@@ -22,9 +20,9 @@ export interface HeaderSectionCoreProps extends ElementPathProps {
 }
 export interface HeaderSectionProps
   extends HeaderSectionCoreProps,
-    NameRowInnerProps,
-    LayoutWrapperCoreProps,
-    LayoutWrapperRowProps {}
+  NameRowInnerProps,
```
balazsbajorics, line 23: this is weird. was the previous code not formatted or is the new code not formatted by Prettier? or is this a prettier version change?

editor/src/uuiui/widgets/popup-list/popup-list.tsx
balazsbajorics: interesting. how could you push this? is someone not running the pre-commit hook?

maltenuhn: I think we bypassed it in an early commit because I needed to add some logging to show Omar. Does that mean that branch is poisoned, and _all_ future commits don't get run through the hook? How do you re-enable it?

OmarDaSilva:
Commit #34c670e fixes the error "src/components/editor/component-button.tsx(115,13): error TS2322: Type 'string' is not assignable to type '"white" | "black" | "blue" | "purple" | "red" | "orange" | "darkgray" | "gray" | "lightgray" | undefined'." from

editor/src/components/canvas/controls/breadcrumb-trail.tsx
OmarDaSilva: commit #c5a34d1 addresses this

editor/src/components/canvas/controls/breadcrumb-trail.tsx
OmarDaSilva: Commit #8b5b7b1 address this

editor/src/components/canvas/controls/breadcrumb-trail.tsx
OmarDaSilva: Commit #930332e address this

editor/src/components/canvas/controls/namerowcrumbs.tsx
OmarDaSilva: Commit #a75137b addresses this

editor/src/components/canvas/controls/namerowcrumbs.tsx
OmarDaSilva: Commit #a77209e addresses this 👍🏾

editor/src/components/editor/top-menu.tsx
OmarDaSilva: #30ee747

editor/src/components/inspector/sections/style-section/style-section.tsx
OmarDaSilva: #30ee747

editor/src/uuiui/button.tsx
OmarDaSilva: #ace17cf

editor/src/uuiui/widgets/popup-list/popup-list.tsx
OmarDaSilva: #ace17cf

editor/src/uuiui/widgets/popup-list/popup-list.tsx
OmarDaSilva: #ace17cf

editor/src/components/inspector/common/name-and-icon-hook.ts
OmarDaSilva: Todo!

Rheeseyb: Extremely helpful comment Balazs, thanks for taking the time to write this out!

Rheeseyb:
This PR is failing to compile because it seems the contents of a bunch of source files under the `header-section` directory have been truncated, but code elsewhere is trying to import from them.

editor/src/uuiui/widgets/popup-list/popup-list.tsx
maltenuhn: Did we need to fix this? I think there was a reason I made this change :D

editor/src/components/inspector/common/name-and-icon-hook.ts
enidemi: fixed this issue, so the PR can be merged today

editor/src/components/canvas/controls/breadcrumb-trail.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { selectComponents } from '../../../components/editor/actions/action-creators'
+import { FlexRow, Icons } from '../../../uuiui'
+import { TemplatePath } from 'src/core/shared/project-file-types'
+
+interface ElementPathElement {
+  name?: string
+  path: TemplatePath
+}
+
+const { dispatch, jsxMetadataKILLME, rootComponents } = useEditorState((store) => {
```
seanparsons, line 20: These React hooks should be inside a React component, not at the root level of the file.

editor/src/components/canvas/controls/namerowcrumbs.tsx
```diff
@@ -0,0 +1,184 @@
...
+import * as TP from '../../../core/shared/template-path'
+import * as EditorActions from '../../editor/actions/action-creators'
+import { GridRow } from '../../../components/inspector/widgets/grid-row'
+import { PopupList } from '../../../uuiui'
+import { JSXElementName, jsxElementName } from '../../../core/shared/element-template'
+import { useNamesAndIconsSelectedViews } from '../../../components/inspector/common/name-and-icon-hook'
+
+const constrolStatus = 'simple'
+const controlStyles = getControlStyles(constrolStatus)
+const hookResult = useNamesAndIconsSelectedViews()
```
seanparsons, line 16: These calls to hooks should also be inside a React component.

editor/src/components/canvas/controls/namerowcrumbs.tsx
```diff
@@ -0,0 +1,184 @@
+import * as React from 'react'
```
seanparsons, line 1: It looks like this was moved as well as modifications made to it, would it be possible for the original path to be used  please? Just because I've also been making changes to this file and it'll potentially make any merge I have to do quite painful.

Rheeseyb:
I've converted this to a draft because it seems like a bunch of changes have been made and pushed but haven't been verified yet.

balazsbajorics:
> Performance: Scroll Canvas: 9.7ms | Resize: 6.2ms | Selection: 59.4ms SummaryChart

hah, the scroll canvas is moving to the right!

maltenuhn:
Have at it, folks!

editor/src/components/canvas/controls/render-as.tsx
```diff
@@ -0,0 +1,120 @@
+import * as React from 'react'
```
maltenuhn, line 1: This moves the functionality from `nameRow`, but substitutes props with hooks

editor/src/components/editor/component-button.tsx
```diff
@@ -0,0 +1,214 @@
+/** @jsx jsx */
```
maltenuhn, line 1: This file contains both the _actual_ button and the popup since it's "attached" to the button. If anyone has a better naming suggestion...

editor/src/components/inspector/inspector.tsx
```diff
@@ -62,15 +61,14 @@ import {
...
+import { useEditorState } from '../editor/store/store-hook'
 import { CSSPosition } from './common/css-utils'
 import { InspectorCallbackContext, InspectorPropsContext } from './common/property-path-hooks'
 import { ComponentSection } from './sections/component-section/component-section'
 import { EventHandlersSection } from './sections/event-handlers-section/event-handlers-section'
-import { ElementPathElement } from './sections/header-section/element-path'
-import { HeaderSection, HeaderSectionCoreProps } from './sections/header-section/header-section'
-import { LayoutWrapperCoreProps } from './sections/header-section/layout-wrapper-section'
-import { NameRowProps } from './sections/header-section/name-row'
+// import { HeaderSection } from './sections/header-section/header-section'
```
maltenuhn, line 69: Delete these pls

editor/src/components/inspector/inspector.tsx
```diff
@@ -687,10 +667,10 @@ export const SingleInspectorEntryPoint: React.FunctionComponent<{
     [dispatch, refElementsToTargetForUpdates],
   )
 
-  const onSelect = React.useCallback(
-    (path) => dispatch([selectComponents([path], false)], 'everyone'),
-    [dispatch],
-  )
+  // const onSelect = React.useCallback(
```
maltenuhn, line 670: delete these pls

editor/src/components/inspector/inspector.tsx
```diff
@@ -759,15 +739,15 @@ export const SingleInspectorEntryPoint: React.FunctionComponent<{
...
-  const onElementTypeChange = React.useCallback(
-    (newElementName: JSXElementName, importsToAdd: Imports) => {
-      const actions = refElementsToTargetForUpdates.current.flatMap((path) => {
-        return EditorActions.updateJSXElementName(path, newElementName, importsToAdd)
-      })
-      dispatch(actions, 'everyone')
-    },
-    [dispatch, refElementsToTargetForUpdates],
-  )
+  // const onElementTypeChange = React.useCallback(
```
maltenuhn, line 742: into the bin with these

editor/src/components/inspector/inspector.tsx
Rheeseyb: :+1: please delete

utopia-api/src/primitives/positionable.tsx
```diff
@@ -11,6 +11,10 @@ export interface PositionableProps
   wrappedComponent: React.JSXElementConstructor<any>
 }
 
+const tempFix = (props: any) => {
+  return <div />
+}
```
Rheeseyb, line 16: I have a feeling this temporary fix isn't going to remain temporary, so please can you rename this accordingly.

Rheeseyb (approved):

editor/src/components/editor/component-button.tsx
```diff
@@ -0,0 +1,216 @@
...
+          ...getEditContextStyle(),
+          '&:hover': {
+            filter: 'brightness(1.02  )',
+          },
+          '&:active': {
+            filter: 'brightness(1.03)',
+          },
+        }}
+      >
+        {/* TODO replace me with the real icon */}
```
balazsbajorics, line 121: is this TODO for the next PR ?

editor/src/uuiui/button.tsx
balazsbajorics: it is still not fixed, and #ace17cf is not highlighted by github – was a commit lost somewhere?

utopia-api/src/primitives/positionable.tsx
```diff
@@ -11,6 +11,10 @@ export interface PositionableProps
   wrappedComponent: React.JSXElementConstructor<any>
 }
 
+const wrappedComponentProp = (props: any) => {
```
balazsbajorics, line 14:
```suggestion
// TODO Delete the entire Positionable component and remove it from the UI
const wrappedComponentProp = (props: any) => {
```

balazsbajorics (dismissed):
I left a comment here: https://github.com/concrete-utopia/utopia/pull/963#discussion_r598641908

this is my only remaining question about the PR

editor/src/components/editor/component-button.tsx
maltenuhn: Yes. We basically need the info from the focus mode work, so we can show the correct icon, label, colour, and status based on the select element(s).

PR #964, opened 2021-03-11
Feature/navigator new icons

enidemi:
**Description:**
This PR is part of the navigator redesign, adding new layout/element type icons before the label and component icons shown after the label.
There are actually 3 types of icons now: layout, element, component.
Icons before the element show layout information or if that's missing it showing element type icon. Component icons are different, they indicate when an element is a react component/emotion/animated component.
This changes the original icon hook result, because that returns component-element icon now, as component is optional.

**Commit Details:**
- new icons
- extended specialsizemeasurements with style display information and html element name, the element name is used for buttons now
- extended the spied data, to know if something is an emotion component or not
- snapshot updates for the metadata changes
- helpers in `element-metadata-utils` for grid and button elements

Screenshot 2021-03-11 at 15 27 03


editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -620,6 +620,7 @@ export function SpecialSizeMeasurementsKeepDeepEquality(): KeepDeepEqualityCall<
         newSize.parentLayoutSystem,
         newSize.layoutSystemForChildren,
         newSize.providesBoundsForChildren,
+        newSize.display,
```
balazsbajorics, line 623: I think display and htmlElementName are missing from the gigantic `areEqual` expression from line 594

enidemi: fixed

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -591,6 +591,8 @@ export function SpecialSizeMeasurementsKeepDeepEquality(): KeepDeepEqualityCall<
     const clientWidthResult = oldSize.clientWidth === newSize.clientWidth
     const clientHeightResult = oldSize.clientHeight === newSize.clientHeight
     const parentFlexDirectionResult = oldSize.parentFlexDirection === newSize.parentFlexDirection
+    const display = oldSize.display === newSize.display
```
balazsbajorics, line 594: I think `displayEquals` would be better. or go with the pattern and `displayResult` but I like equals better because this is not a comparison result object

editor/src/components/navigator/layout-element-icons.ts
```diff
@@ -0,0 +1,248 @@
...
+    },
+  )
+}
+
+export function useComponentIcon(path: TemplatePath): IcnProps | null {
+  return useEditorState((store) => {
+    const metadata = store.editor.jsxMetadataKILLME
+    const components = getOpenUtopiaJSXComponentsFromState(store.editor)
+    const imports = getOpenImportsFromState(store.editor)
+    return createComponentIconProps(path, components, metadata, imports)
```
balazsbajorics, line 51: this looks like it's going to be busting the shallow equality.

enidemi: why? it's returning an object with strings and numbers

balazsbajorics: ah yeah you are right! sorry I thought there's something more complex in there

editor/src/utils/deep-equality.ts
```diff
@@ -359,7 +359,7 @@ export function combine9EqualityCalls<A, B, C, D, E, F, G, H, I, X>(
   }
 }
 
-export function combine10EqualityCalls<A, B, C, D, E, F, G, H, I, J, X>(
+export function combine11EqualityCalls<A, B, C, D, E, F, G, H, I, J, K, X>(
```
balazsbajorics, line 362: I think it'd be better if we'd also keep `combine10EqualityCalls` alive, we might need it in the future

balazsbajorics (approved):

editor/src/components/navigator/layout-element-icons.ts
enidemi: I changed the type here to a simple one

editor/src/utils/deep-equality.ts
enidemi: fixed

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.tsx
```diff
@@ -33,6 +33,11 @@ export function buildSpyWrappedElement(
   }
   const childrenElementsOrNull = childrenElements.length > 0 ? childrenElements : null
   const spyCallback = (reportedProps: any) => {
+    /** This is not so nice, but the way to know if something is an emotion component is
+     * that it adds some extra properties to the Element itself, like __emotion_base,
+     * TODO move this out of metadata once we have syledcomponent editing
+     */
+    const isEmotionComponent = Element['__emotion_base'] != null
```
Rheeseyb, line 40: Does emotion export a const for this key (or any other key we can rely on)? If not, please can you add a test to ensure we're correctly recording `isEmotionComponent: true` for a styled component so that we can catch if they change this key.

editor/src/components/navigator/layout-element-icons.ts
```diff
@@ -0,0 +1,248 @@
...
+export function createComponentOrElementIconProps(
+  path: TemplatePath,
+  components: UtopiaJSXComponent[],
+  metadata: JSXMetadata,
+  imports: Imports,
+): IcnPropsBase {
+  return (
+    createComponentIconProps(path, components, metadata, imports) ??
+    createElementIconProps(path, components, metadata)
+  )
```
Rheeseyb, line 64: I ... actually quite like this!

editor/src/components/navigator/navigator-item/item-preview.tsx
```diff
@@ -18,12 +19,21 @@ export const ItemPreview: React.FunctionComponent<ItemPreviewProps> = betterReac
     // 3 - if it's a component or not
     // 4 - if it's a placeholder or not
     // 5 if it's generated or not
```
Rheeseyb, line 21: `// 6 - if we can generate a preview for it`?
I don't actually understand what the 6th 3rd thing is here - if we return `null` do we just not have a preview?

Rheeseyb (approved):

editor/src/components/navigator/navigator-item/item-preview.tsx
enidemi: I removed these comments because this is not how the icons work now

editor/src/core/model/project-file-utils.ts
```diff
@@ -164,6 +165,13 @@ function isHTMLComponentFromBaseName(baseName: string, imports: Imports): boolea
   }
 }
 
+export function isImportedComponent(jsxElementName: JSXElementName, imports: Imports): boolean {
+  return Object.keys(imports).some((importKey) => {
```
seanparsons, line 169: I don't believe this covers all the cases for how something may be imported, like if a component is the default export.

seanparsons (approved):

PR #966, opened 2021-03-12
Feature/scene paths array

### LINKED ISSUE #965: Scene Paths should be made of an array of Element Paths
Right now a Template Path consists of a Scene Path and an Element Path, where the former represents the logical path to the scene that an element is being rendered in. In order to support drilling into components, we need to be able to support concatenation in the Scene Path part of the Template Path in a way that allows us to capture the contexts at each point.

Rheeseyb:
Fixes #965 

**Problem:**
Scene Paths only contain an Element Path, which is an array of strings. We want them to contain multiple Element Paths so that we can use them to drill into components, so e.g. if the user selects `storyboard/scene:app/card` and wants to then drill into the card component, we could represent the root level element of the Card component as `storyboard/scene:app/card:card-rood`

**Fix:**
Change the `ScenePath` definition from:

```
export type ScenePath = {
  type: 'scenepath'
  sceneElementPath: StaticElementPath
}
```

To:

```
export type ScenePath = {
  type: 'scenepath'
  sceneElementPaths: StaticElementPath[]
}
```

**Commit Details:**
- Reduced explicit use of `ScenePath.sceneElementPaths` throughout
- Changed the various factory functions to support an array of `ElementPath`s as part of the creation
- Introduced the function `instancePathForElementAtScenePath` for taking a Scene Path and creating the Instance Path that points to the instance of that scene - right now we record metadata for the scene `storyboard/scene` at `:storyboard/scene`, so we were previously creating this by using `TP.instancePath([], TP.elementPathForPath(scenePath))`, but with this new shape of Scene Paths in the above Card example the path for the Card "scene" would be `storyboard/scene:app/card` (so we want to take the Scene Path, pop the last Element Path, and use that to create the new Instance Path)
- Introduced a `splitAt` function for splitting an array into two to support the above (definitely could have used `Array.pop()` in hindsight, but you live and you learn - update: I used `Array.pop`, but kept the new useful function)
- Also introduced `scenePathForElementAtInstancePath` which is the inverse of `instancePathForElementAtScenePath`, pushing the `InstancePath`'s element path onto its `ScenePath`, and returning that
- Updated the logic for looking up a Scene Path from the cache, using the `SceneSeparator` in a rather nasty and unacceptable way but I thought that any other way of forking the caching logic there would have been more complex - happy to reevaluate though if anyone wants to challenge that
- Updated all of the tests to reflect these changes

**Note:** The vast majority of the logic changes are inside `template-path.ts`, with the model change in `project-file-types.ts`, so I'd recommend starting there an then referring to either the snapshots or other tests to see how those changes are reflected in practice.

### DIFF
--- a/editor/src/core/shared/template-path.ts
+++ b/editor/src/core/shared/template-path.ts
@@ -165,7 +165,7 @@ function newScenePath(elementPaths: StaticElementPath[]): ScenePath {
   }
 }
 
-export const emptyScenePath: ScenePath = newScenePath(([] as any) as StaticElementPath[])
+export const emptyScenePath: ScenePath = newScenePath([])
 
 function newInstancePath(scene: ScenePath, elementPath: ElementPath): InstancePath {
   return {
--- a/editor/src/core/shared/template-path.ts
+++ b/editor/src/core/shared/template-path.ts
@@ -10,7 +10,7 @@ import {
 } from './project-file-types'
 import { arrayEquals, longestCommonArray, identity, fastForEach } from './utils'
 import { replaceAll } from './string-utils'
-import { last, dropLastN, drop } from './array-utils'
+import { last, dropLastN, drop, splitAt } from './array-utils'
 import { extractOriginalUidFromIndexedUid } from './uid-utils'
 
 // KILLME, except in 106 places
@@ -75,7 +75,7 @@ function getScenePathCache(sceneElementPath: StaticElementPath): ScenePathCache 
 }
 
 function getScenePathCacheForScenePath(scene: ScenePath): ScenePathCache {
-  return getScenePathCache(scene.sceneElementPath)
+  return getScenePathCache(scene.sceneElementPaths)
 }
 
 function getInstancePathCacheFromScenePathCache(
@@ -109,7 +109,7 @@ const ElementSeparator = '/'
 function scenePathToString(path: ScenePath): string {
   const pathCache = getScenePathCacheForScenePath(path)
   if (pathCache.cachedToString == null) {
-    const result = elementPathToString(path.sceneElementPath)
+    const result = elementPathToString(path.sceneElementPaths)
     pathCache.cachedToString = result
     return result
   } else {
@@ -153,26 +153,26 @@ export function toString(target: TemplatePath): string {
 function newScenePath(elements: StaticElementPath): ScenePath {
   return {
     type: 'scenepath',
-    sceneElementPath: elements,
+    sceneElementPaths: elements,
   }
 }
 
 export const emptyScenePath: ScenePath = newScenePath(([] as any) as StaticElementPath)
 
-function newInstancePath(scene: ScenePath | ElementPath, elementPath: ElementPath): InstancePath {
+function newInstancePath(scene: ScenePath, elementPath: ElementPath): InstancePath {
   return {
-    scene: scenePathFromElementPathOrScenePath(scene),
+    scene: scene,
     element: elementPath,
   }
 }
 
 export const emptyInstancePath: InstancePath = newInstancePath(emptyScenePath, [])
 
-export function scenePath(elements: ElementPath): ScenePath {
-  const staticElements = elements as StaticElementPath
-  const pathCache = getScenePathCache(staticElements)
+export function scenePath(elementPaths: ElementPath[]): ScenePath {
+  const staticElementPaths = elementPaths as StaticElementPath[]
+  const pathCache = getScenePathCache(last(staticElementPaths)!)
   if (pathCache.cached == null) {
-    const newPath = newScenePath(staticElements)
+    const newPath = newScenePath(last(staticElementPaths)!)
     pathCache.cached = newPath
     return newPath
   } else {
@@ -180,16 +180,8 @@ export function scenePath(elements: ElementPath): ScenePath {
   }
 }
 
-function scenePathFromElementPathOrScenePath(scene: ScenePath | ElementPath): ScenePath {
-  if (Array.isArray(scene)) {
-    return scenePath(scene)
-  } else {
-    return scene
-  }
-}
-
 export function instancePath(scene: ScenePath, elementPath: ElementPath): InstancePath {
-  if (scene.sceneElementPath.length === 0 && elementPath.length === 0) {
+  if (scene.sceneElementPaths.length === 0 && elementPath.length === 0) {
     return emptyInstancePath
   } else {
     const pathCache = getInstancePathCache(scene, elementPath)
@@ -232,9 +224,14 @@ export function scenePathForPath(path: TemplatePath): ScenePath {
 }
 
 export function instancePathForScenePath(path: ScenePath): StaticInstancePath {
-  const lastElementPath = path.sceneElementPath
+  const lastElementPath = path.sceneElementPaths
   const targetScenePath = emptyScenePath
   return staticInstancePath(targetScenePath, lastElementPath)
+}
+
+export function scenePathFromInstancePath(path: InstancePath): ScenePath {
+  // FIXME Rename? The use of this is a bit confusing
+  return scenePath([path.scene.sceneElementPaths, path.element])
 }
 
 export function elementPathForPath(path: StaticInstancePath): StaticElementPath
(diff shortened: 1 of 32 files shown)


editor/src/core/shared/template-path.ts
```diff
@@ -262,18 +269,26 @@ export function filterScenes(paths: TemplatePath[]): StaticInstancePath | Instan
...
-  const scene = sceneStr.split(ElementSeparator).filter((e) => e.length > 0)
+  const elementPathStrs = path.split(SceneSeparator)
+  const [sceneStrs, elementStrs] =
+    elementPathStrs.length > 1
+      ? splitAt(elementPathStrs.length - 1, elementPathStrs)
+      : [elementPathStrs, []]
+  const elementStr = elementStrs[0] ?? null
+  const sceneElementPaths = sceneStrs.map((scene) =>
+    scene.split(ElementSeparator).filter((e) => e.length > 0),
+  )
```
Rheeseyb, line 280: I'm totally going to change this to use `Array.pop` because I was an idiot for not doing that in the first place

editor/src/core/shared/template-path.ts
```diff
@@ -241,18 +238,28 @@ export function scenePathForPath(path: TemplatePath): ScenePath {
...
+    return emptyInstancePath
   } else {
-    return path.element
+    const targetSceneElementPaths = dropLast(path.sceneElementPaths)
+    const targetScenePath = scenePath(targetSceneElementPaths)
+    return staticInstancePath(targetScenePath, lastElementPath)
   }
 }
 
+export function scenePathFromInstancePath(path: InstancePath): ScenePath {
```
Rheeseyb, line 252: This function is the inverse of `instancePathForScenePath`, but we also have `scenePathForPath` which has completely different functionality, so the naming is becoming a bit confusing. Right now the use of the 3 different functions are as follows (just pretend equality of objects is a thing :grinning:):

```
const a = { 
  type: 'scenepath',
  sceneElementPaths: [
    ['storyboard', 'scene'],
    ['app', 'card'],
  ]
}

const b = {
  type: 'instancepath',
  scene: {
    type: 'scenepath',
    sceneElementPaths: [
      ['storyboard', 'scene']
    ]
  },
  element: ['app', 'card']
}

const c = { 
  type: 'scenepath',
  sceneElementPaths: [
    ['storyboard', 'scene'],
  ]
}

instancePathForScenePath(a) == b // true
scenePathFromInstancePath(b) == c // true
scenePathForPath(b) == c // true
```

editor/src/core/shared/template-path.ts
Rheeseyb: Done

editor/src/core/shared/template-path.ts
```diff
@@ -149,71 +158,59 @@ export function toString(target: TemplatePath): string {
...
+function newScenePath(elementPaths: StaticElementPath[]): ScenePath {
   return {
     type: 'scenepath',
-    sceneElementPath: elements,
+    sceneElementPaths: elementPaths,
   }
 }
 
-const emptyScenePath: ScenePath = newScenePath(([] as any) as StaticElementPath)
+export const emptyScenePath: ScenePath = newScenePath(([] as any) as StaticElementPath[])
```
balazsbajorics, line 168: is the `as any) as StaticElementPath[]` still needed? maybe typescript became smarter since then

editor/src/core/shared/template-path.ts
balazsbajorics: I think it's worth adding some comments to these functions, to help us navigate the code in the future

editor/src/core/shared/template-path.ts
Rheeseyb: Ha, it actually did become smarter! Nice :+1:

Rheeseyb: Fixed

editor/src/core/shared/template-path.ts
balazsbajorics: uhmm actually, shouldn't `scenePathFromInstancePath(b) == c` be instead `scenePathFromInstancePath(b) == a`?

Rheeseyb: Yeah agreed. I was hoping someone would pitch me some much clearer names :grinning: 
I'll also add tests for each of the functions to help aid in the documentation of them.

Rheeseyb: > uhmm actually, shouldn't `scenePathFromInstancePath(b) == c` be instead `scenePathFromInstancePath(b) == a`?

Dunno what you're talking about, that's exactly what it says

editor/src/core/shared/template-path.ts
```diff
@@ -241,18 +238,28 @@ export function scenePathForPath(path: TemplatePath): ScenePath {
   }
 }
 
-export function elementPathForPath(path: ScenePath): StaticElementPath
-export function elementPathForPath(path: StaticInstancePath): StaticElementPath
-export function elementPathForPath(path: InstancePath): ElementPath
-export function elementPathForPath(path: TemplatePath): ElementPath
-export function elementPathForPath(path: TemplatePath): ElementPath {
-  if (isScenePath(path)) {
```
balazsbajorics, line 249: this seems like a hard to test change – I guess nothing fell on you after removing the scenePath from here? :)

editor/src/core/shared/template-path.ts
```diff
@@ -262,18 +269,25 @@ export function filterScenes(paths: TemplatePath[]): StaticInstancePath | Instan
 // FIXME: This should be retired, it's just plain dangerous.
```
balazsbajorics, line 269: note to self: if we'd move the valid path out of the DOM, we could kill the fromString, that is the only place it is used at!

editor/src/core/shared/template-path.ts
Rheeseyb: Ok I have renamed the 3 functions to add clarity, inserted a comment at the start of them each, and added tests for them in a single `describe` block that should also help demonstrate their differences.

editor/src/core/shared/template-path.ts
Rheeseyb: In the case of scenes it was mostly used for the functionality provided by (the now beautifully named) `instancePathForLastPartOfScenePath`, and is covered in various tests so we should be good here

editor/src/core/shared/template-path.ts
Rheeseyb: Unless we have to added stringified template paths to the DOM for some sort of upcoming drilling work :grimacing: ...

Rheeseyb: (I'm not saying we do, just that that's a possibility)

balazsbajorics (approved):

editor/src/core/shared/template-path.ts
```diff
@@ -60,9 +60,17 @@ export function clearTemplatePathCache() {
   globalScenePathCache = emptyScenePathCache()
 }
 
-function getScenePathCache(sceneElementPath: StaticElementPath): ScenePathCache {
+function getScenePathCache(sceneElementPaths: StaticElementPath[]): ScenePathCache {
+  let joinedPaths: string[] = []
+  for (var i = 0, len = sceneElementPaths.length; i < len; i++) {
```
seanparsons, line 65: Why is `len` initialised as part of the `for` loop? I thought for a moment it was being changed in here.

Rheeseyb: I've inlined our `fastForEach` here because I needed access to the index

editor/src/core/shared/template-path.ts
```diff
@@ -262,18 +272,25 @@ export function filterScenes(paths: TemplatePath[]): StaticInstancePath | Instan
 // FIXME: This should be retired, it's just plain dangerous.
 // Right now this is only used for SpecialNodes (in CanvasMetaData)
 function fromStringUncached(path: string): TemplatePath {
-  const [sceneStr, elementStr] = path.split(SceneSeparator)
-  const scene = sceneStr.split(ElementSeparator).filter((e) => e.length > 0)
+  let elementPathStrs = path.split(SceneSeparator)
+  const lastElementStr = elementPathStrs.pop()! // We know this is safe because ''.split(SceneSeparator).pop() returns ''
```
seanparsons, line 276: I _think_ latest TypeScript has some stuff that allows you to do a spread like this: `const [...elementPathStrs, lastElementStr] = ...`.

editor/src/core/shared/template-path.ts
Rheeseyb: As for why I didn't use `fastForEach` since that also provides the index, I don't know...

seanparsons (approved):

editor/src/core/shared/template-path.ts
Rheeseyb: I just checked this and it appears to not be a thing (though I know you can do `const [head, ...tail] = [1,2,3,4]`)

PR #967, opened 2021-03-12
Groundwork before Spy changes

balazsbajorics:
This PR contains a new test suite that will help us figuring out if we are on the right path while working on the Spy. It also contains a (not yet used) new editor state variable that Eni can use in the navigator and we can use in the Spy.

**Commit Details:**
- new Spy Wrapper Template Path Tests that only test what template paths the spy uses as keys and as children
- new editor state entry `focusedElementPath: ScenePath | null` that in the future will point to the focused element
- action boilerplate for changing `focusedElementPath`


Rheeseyb (approved):

Rheeseyb:
Since this doesn't touch anything existing, this should be fine to merge straight to master (assuming the tests pass)

seanparsons (approved):

PR #968, opened 2021-03-12
fix(fonts) - replace badly rendered variable font with non-variable

maltenuhn:
**Problem** The Inter variable font compresses oddly horizontally by default. Also, we were no longer using the monospaced font in our repo.

**Solution** Replace variable Inter with the non-variable variants we're actually using. Remove the monospaced font entirely.

Thanks @enidemi for spotting the root cause, I had this down to an un-debuggable Chrome text rendering weirdness.

Before (left) and after (right)
image


enidemi (approved):

editor/resources/editor/fonts.css
```diff
@@ -1,9 +1,36 @@
+@font-face {
+  font-family: 'utopian-inconsolata';
+  src: url('./fonts/Inconsolata-VariableFont_wdth\,wght.ttf');
+}
+
```
Rheeseyb, line 5: Looks like this should be deleted then?

Rheeseyb (approved):

editor/resources/editor/fonts.css
maltenuhn: Fixed!

PR #971, opened 2021-03-15
Support static and dynamic scene paths

Rheeseyb:
**Problem:**
Scene paths were assumed to be static, but with the work to drill into Components we can no longer make this assumption.

**Fix:**
Create a `StaticScenePath` and explicitly support both


editor/src/core/shared/template-path.ts
```diff
@@ -166,7 +167,7 @@ function newScenePath(elementPaths: StaticElementPath[]): ScenePath {
   }
 }
 
-export const emptyScenePath: ScenePath = newScenePath([])
+export const emptyScenePath: StaticScenePath = newScenePath([]) as StaticScenePath
```
seanparsons, line 170: Isn't this cast dangerous? If the definition of `newScenePath` changes for instance?

editor/src/core/shared/template-path.ts
```diff
@@ -196,6 +197,10 @@ export function scenePath(elementPaths: ElementPath[]): ScenePath {
   }
 }
 
+export function staticScenePath(elementPaths: ElementPath[]): StaticScenePath {
+  return scenePath(elementPaths) as StaticScenePath
```
seanparsons, line 201: This is also dangerous right?

seanparsons: I'd advocate for a tiny bit of duplicated code over these kinds of casts.

seanparsons (approved):

editor/src/core/shared/template-path.ts
Rheeseyb: Oops, this function was supposed to take `StaticElementPath[]`. Good catch.

editor/src/core/shared/template-path.ts
Rheeseyb: Fair point. This was based on the `emptyInstancePath` definition, but I'll hand roll it for both.

editor/src/core/shared/project-file-types.ts
```diff
@@ -14,9 +14,15 @@ import { arrayEquals, objectEquals } from './utils'
 export type id = string
 
 export type ScenePath = {
+  type: 'scenepath'
+  sceneElementPaths: ElementPath[]
+}
+
+export type StaticScenePath = {
   type: 'scenepath'
```
balazsbajorics, line 22: for extra super duper type safety, we could name it staticscenepath too

editor/src/core/shared/template-path.ts
```diff
@@ -856,6 +865,40 @@ export function isFromSameSceneAs(a: TemplatePath, b: TemplatePath): boolean {
...
+  return scene.sceneElementPaths.some((sceneElementPath) =>
+    elementPathsEqual(sceneElementPath, elementPath),
+  )
+}
+
+export function staticScenePathContainsElementPath(
+  scene: ScenePath,
+  elementPath: ElementPath,
+): boolean {
+  return scene.sceneElementPaths.some((sceneElementPath) => {
```
balazsbajorics, line 900: this could use `dynamicScenePathToStaticScenePath` instead of manually calling map to extractOriginalUidFromIndexedUid

editor/src/core/shared/template-path.ts
Rheeseyb: Done

editor/src/core/shared/template-path.ts
Rheeseyb: Done

editor/src/core/shared/project-file-types.ts
Rheeseyb: We could but I'm not going to make that change as part of this as it requires even more refactoring

editor/src/core/shared/template-path.ts
Rheeseyb: Done and I also introduced a `dynamicElementPathToStaticElementPath` since we were now doing this same map a few times.

PR #972, opened 2021-03-15
Improved Spy Wrapper

balazsbajorics:
**Problem:**
As part of our work introducing "Component Focus Mode" we introduced the concept of a focused element path: this is a scenePath that points to an element instance which we want to be able to "drill into" and edit the component definition itself as if it would be a scene's target. If a component instance is focused, we want the spy to wrap it and all its children. We also want to spy on every parent along the path. 

If you have a component hierarchy like this:

```javascript
const Button = (props) => {
  return <div data-uid="button-root">{props.children}</div>;
};
const Card = () => {
  return (
    <Button data-uid="button-instance">
        <div data-uid="hi-element">Hi!</div>
    </Button>
  );
};
export var App = (props) => {
  return (
    <div
      data-uid="app-root"
    >
      <div data-uid="inner-div">
        <Card data-uid="card-instance" />
      </div>
    </div>
  );
};
```

If you set the `focusedElementPath` to point to `storyboard/scene:app-root/inner-div/card-instance:button-instance` we want to Spy wrap all the elements along this path: App and its elements, Card, Button and their divs too.

**Fix:**
We changed `getValidTemplatePathsFromElement` to get a `focusedElementPath: ScenePath | null` parameter. If we encounter a templatePath that is along the `focusedElementPath`, and that element is a component with a name we can find in the `topLevelElements`, we recurse into that component and collect its valid paths as well.
So if in the above example the focusedElementPath is `storyboard/scene:app-root/inner-div/card-instance:button-instance`, the valid paths are going to be: 

```
storyboard/scene:app-root
storyboard/scene:app-root/inner-div
storyboard/scene:app-root/inner-div/card-instance
storyboard/scene:app-root/inner-div/card-instance:button-instance
storyboard/scene:app-root/inner-div/card-instance:button-instance/hi-element
storyboard/scene:app-root/inner-div/card-instance:button-instance:button-root
```

Once we made these changes to the valid paths, we also had to do some tweaking to `renderCoreElement` and the componentRendererComponent to be able to build a path like this `storyboard/scene:app-root/inner-div/card-instance:button-instance:button-root` while recursively rendering children elements. We needed the change because by default an element deep in the hierarchy such as `button-root` would have gotten a half-broken template path such as `storyboard/scene:button-root`. This wasn't a problem previously because we only ever needed to drill into the selected Scene's main Component, and these template paths would be discarded as invalid.

To aid us during this work we made some TDD unit tests, those are disabled in the PR because they are not yet passing. (To seem them passing, you can comment out `cullSpyCollector` but that will break other things in the editor.)

While we were here, we realized that the SpyWrapper was not running for generated elements, a regression masked by the fact that the dom-walker was providing sufficient data to show generated elements in the navigator. We decided to not leave it broken so this PR includes quite some changes regarding generated elements.

**Note:**
There is a hypermassive change to ui-jsx-canvas.spec.snapshot, all of the changes there are coming from the fact that the spy now runs for generated elements, so we include spy metadata for those as well. Sorry about that – I think we should significantly filter the metadata from the snapshot, make it focus on the template paths, props, children. Right now it includes the entire parsed element, including the source javascript - for every single element!


editor/src/core/shared/template-path.ts
```diff
@@ -13,7 +13,7 @@ import { replaceAll } from './string-utils'
 import { last, dropLastN, drop, splitAt, flattenArray, dropLast } from './array-utils'
 import { extractOriginalUidFromIndexedUid } from './uid-utils'
 
-// KILLME, except in 106 places
+// KILLME, except in 28 places
```
balazsbajorics, line 16: 😀

balazsbajorics:
> Hmmm - the "scroll canvas" test takes three times as much time as before, but the others stayed the same. Is that plausible?

that's pretty interesting! let me re-run the tests :D

maltenuhn:
Seems like it’s consistent at least 🙃

balazsbajorics:
> Seems like it’s consistent at least 🙃

this might well be a real life performance regression caught by the test suite :)

editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
```diff
@@ -135,14 +137,19 @@ export const SceneRootRenderer = betterReactMemo(
 
     useRunSpy(scenePath, templatePath, topLevelElementName, sceneProps)
 
+    const propsWithScenePath = {
+      ...sceneProps.props,
+      [UTOPIA_SCENE_PATH]: scenePath,
+    }
```
Rheeseyb, line 143: Previously we were providing the `scenePath` to the `ComponentRendererComponent` via the context when creating those `ComponentRendererComponent`s for each top level element. The reason being that the concept of nested "scene"s was never really a thing for us. This meant that when calling `renderCoreElement` later on down the line, the template path being passed in (`ownTemplatePath`) would only be correct for top level scenes. To support drilling into components we needed to address this, so now we pass in the actual `scenePath` via a prop here.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -34,7 +36,14 @@ export function isComponentRendererComponent(
 export function createComponentRendererComponent(params: {
   topLevelElementName: string
 }): ComponentRendererComponent {
-  const Component = (realPassedProps: any) => {
+  const Component = (realPassedPropsIncludingUtopiaSpecialStuff: any) => {
+    const {
+      [UTOPIA_SCENE_PATH]: scenePathAny, // TODO types?
```
Rheeseyb, line 41: This `TODO` can go in the bin now since we have the `isScenePath` below

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -34,7 +36,14 @@ export function isComponentRendererComponent(
...
+      ...realPassedProps
+    } = realPassedPropsIncludingUtopiaSpecialStuff
+
+    const scenePath: ScenePath | null = TP.isScenePath(scenePathAny) ? scenePathAny : null
```
Rheeseyb, line 45: We still appear to have a case where the scene path prop isn't set, which is captured in the test [here](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/ui-jsx-canvas-bugs.spec.tsx). To get around that we support null template paths now where we didn't before, but it's still a bit confusing whether or not that is actually a legitimate case or a bug elsewhere in the logic... (the commit for supporting those null template paths is https://github.com/concrete-utopia/utopia/pull/972/commits/afab4a59099262bf059d0dfcede95cd7658ed6c8)

editor/src/core/shared/template-path.ts
```diff
@@ -219,8 +219,8 @@ export function asStatic(path: InstancePath): StaticInstancePath {
   return path as StaticInstancePath
 }
 
-export function isScenePath(path: TemplatePath): path is ScenePath {
-  return (path as any).type === 'scenepath'
+export function isScenePath(path: unknown): path is ScenePath {
+  return (path as any)?.type === 'scenepath'
```
Rheeseyb, line 223: Turns out this magical question mark saves us from having to write
```
typeof path === 'object' && path != null && (path as any).type ===
```

Who knew?! Balazs apparently :grinning:

editor/src/core/shared/template-path.ts
```diff
@@ -856,6 +856,34 @@ export function isFromSameSceneAs(a: TemplatePath, b: TemplatePath): boolean {
   return scenePathsEqual(scenePathPartOfTemplatePath(a), scenePathPartOfTemplatePath(b))
 }
 
+function dynamicScenePathToStaticScenePath(scene: ScenePath): ScenePath {
```
Rheeseyb, line 859: Some of the below is duplicated in #971 - we decided to separate that work into its own PR and will fix the merge conflicts since this was required for this to work.

seanparsons (approved):

enidemi (approved):

balazsbajorics:
> > Seems like it’s consistent at least 🙃
> 
> this might well be a real life performance regression caught by the test suite :)

found the performance regression, working on a fix. tldr the canvas started rerendering the entire content for every scroll move. that explains the doubled frame times

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -331,7 +334,7 @@ export const UiJsxCanvas = betterReactMemo(
         jsxFactoryFunctionName: jsxFactoryFunction,
       })
 
-      const topLevelElementsMap = new Map(topLevelJsxComponents)
+      const topLevelElementsMap = useKeepReferenceEqualityIfPossible(new Map(topLevelJsxComponents))
```
balazsbajorics, line 337: this was the one liner to keep the canvas root from rerendering. However, I don't remember why was `topLevelJsxComponents` here copied into a `new Map()`? Is this a feature we still rely on? I'm too sleepy to figure out but maybe @Rheeseyb remembers something...

balazsbajorics: https://github.com/concrete-utopia/dystopia/pull/3457/files#r595547751

balazsbajorics:
I'm merging this now as I am blocked on this

PR #975, opened 2021-03-16
fix(hook) - update when any values are not the same

maltenuhn:
**Problem:**
The `useNameAndIcon` hook did not update when the underlying information changed. We discovered this in the `RenderAs` component that lets you change the currently selected element or component to a different one.

**Fix:**
Changed the equality function to use `and` not `or`


balazsbajorics:
@enidemi it might be worth to replace this equality function with a more structured `combine5EqualityCalls` one

balazsbajorics (approved):

enidemi (approved):

Rheeseyb (commented):
:+1:

PR #977, opened 2021-03-17
Support identifying the underlying file and path of a given element.

seanparsons:
**Problem:**
We want to be able to edit a component that is shown in the canvas for a given file but is not from that file, but have no way to "find" what element it is in which particular file.

**Fix:**
Provide functionality for identifying the originating file for a given element along with the component it relates to in that file.

**Commit Details:**
- Copied some path related functions and added `absolutePathFromRelativePath`.
- Added `findElementWithUID` to lookup a given element given the `data-uid`
  that relates to it.
- Added `dropFirstScenePathElement` to assist with walking a path down
  the various scenes that are included.
- Added `importedFromWhere` to find the import that a variable name relates to.
- Implemented `normalisePathToUnderlyingTarget` which performs the work
  of jumping around the project contents of a project finding the underlying
  components as it goes.

### DIFF
--- a/editor/src/components/custom-code/code-file.spec.ts
+++ b/editor/src/components/custom-code/code-file.spec.ts
@@ -467,11 +467,22 @@ function defaultProjectContentsForNormalising(): ProjectContentTreeRoot {
 import * as React from 'react'
 import { Scene, Storyboard, jsx } from 'utopia-api'
 import { App } from '/src/app.js'
+
+export var SameFileApp = (props) => {
+  return <div data-uid='same-file-app-div' />
+}
+
 export var storyboard = (
   <Storyboard data-uid='storyboard-entity'>
     <Scene
-      data-uid='scene-entity'
+      data-uid='scene-1-entity'
       component={App}
+      props={{}}
+      style={{ position: 'absolute', left: 0, top: 0, width: 375, height: 812 }}
+    />
+    <Scene
+      data-uid='scene-2-entity'
+      component={SameFileApp}
       props={{}}
       style={{ position: 'absolute', left: 0, top: 0, width: 375, height: 812 }}
     />
@@ -514,7 +525,7 @@ import { App } from '/src/app.js'
 export var storyboard = (
   <Storyboard data-uid='storyboard-entity'>
     <Scene
-      data-uid='scene-entity'
+      data-uid='scene-1-entity'
       component={App}
       props={{}}
       style={{ position: 'absolute', left: 0, top: 0, width: 375, height: 812 }}
@@ -558,12 +569,25 @@ function staticInstancePathFromString(path: string): StaticInstancePath {
 
 describe('normalisePathToUnderlyingTarget', () => {
   const projectContents = defaultProjectContentsForNormalising()
+  it('handles finding the target within the same file', () => {
+    const actualResult = normalisePathToUnderlyingTarget(
+      projectContents,
+      StoryboardFilePath,
+      staticInstancePathFromString('storyboard-entity/scene-2-entity:same-file-app-div'),
+    )
+    const expectedResult = normalisePathSuccess(
+      staticInstancePathFromString(':same-file-app-div'),
+      StoryboardFilePath,
+      getTextFileByPath(projectContents, StoryboardFilePath),
+    )
+    expect(actualResult).toEqual(expectedResult)
+  })
   it('jumps across multiple files to reach the actual target', () => {
     const actualResult = normalisePathToUnderlyingTarget(
       projectContents,
       StoryboardFilePath,
       staticInstancePathFromString(
-        'storyboard-entity/scene-entity:app-outer-div/card-instance:card-outer-div/card-inner-div',
+        'storyboard-entity/scene-1-entity:app-outer-div/card-instance:card-outer-div/card-inner-div',
       ),
     )
     const expectedResult = normalisePathSuccess(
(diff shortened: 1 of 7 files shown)


editor/src/components/custom-code/code-file.ts
```diff
@@ -254,3 +273,169 @@ export function codeCacheToBuildResult(cache: {
...
+}
+
+export function normalisePathError(errorMessage: string): NormalisePathError {
+  return {
+    type: 'NORMALISE_PATH_ERROR',
+    errorMessage: errorMessage,
+  }
+}
+
+export interface NormalisePathUnableToProceed {
```
Rheeseyb, line 321: Seems like this could do with coming with an error message too?

editor/src/components/custom-code/code-file.ts
```diff
@@ -254,3 +273,169 @@ export function codeCacheToBuildResult(cache: {
...
+      const staticPath = TP.dynamicPathToStaticPath(elementPath)
+      const potentiallyDroppedFirstSceneElementResult = TP.dropFirstScenePathElement(staticPath)
+      if (potentiallyDroppedFirstSceneElementResult.droppedScenePathElements == null) {
+        // As the scene path is empty, there's no more traversing to do, the target is in this file.
+        return normalisePathSuccess(staticPath, currentFilePath, currentFile)
+      } else {
+        const droppedPathPart = potentiallyDroppedFirstSceneElementResult.droppedScenePathElements
+        if (droppedPathPart.length === 0) {
+          return normalisePathError(
+            `Unable to handle empty scene path part for ${TP.toString(elementPath)}`,
```
Rheeseyb, line 372: Is this the difference between `scenePath([])` and `scenePath([[]])`?

editor/src/components/custom-code/code-file.ts
```diff
@@ -254,3 +273,169 @@ export function codeCacheToBuildResult(cache: {
...
+            }
+            // Handle things like divs.
+            if (isIntrinsicElement(targetElement.name)) {
+              return normalisePathSuccess(
+                TP.dynamicPathToStaticPath(potentiallyDroppedFirstSceneElementResult.newPath),
+                currentFilePath,
+                currentFile,
+              )
+            } else {
+              return lookupElementImport(targetElement.name.baseVariable)
```
Rheeseyb, line 432: This needs to be extended to support the fully qualified element name (I think this would break right now on `animated.div`)

Rheeseyb: Oh sorry, I'm an idiot. I just realised what `baseVariable` is :facepalm:

editor/src/core/shared/template-path.ts
```diff
@@ -914,3 +914,35 @@ export function staticScenePathContainsElementPath(
...
   })
 }
+
+export function isScenePathEmpty(path: TemplatePath): boolean {
+  if (isScenePath(path)) {
+    return path.sceneElementPaths.length === 0
+  } else {
+    return path.scene.sceneElementPaths.length === 0
+  }
+}
```
Rheeseyb, line 924: Nice!

editor/src/core/shared/template-path.ts
```diff
@@ -914,3 +914,35 @@ export function staticScenePathContainsElementPath(
...
+
+interface DropFirstScenePathElementResultType {
+  newPath: InstancePath
+  droppedScenePathElements: StaticElementPath | null
+}
+
+export function dropFirstScenePathElement(
+  path: StaticInstancePath,
+): DropFirstScenePathElementResultType {
+  if (path.scene.sceneElementPaths.length === 0) {
```
Rheeseyb, line 934: Should this use `isScenePathEmpty`?

editor/src/components/custom-code/code-file.spec.ts
```diff
@@ -407,3 +442,179 @@ describe('Creating require function', () => {
...
+      ),
+    )
+    const expectedResult = normalisePathSuccess(
+      staticInstancePathFromString(':card-outer-div/card-inner-div'),
+      '/src/card.js',
+      getTextFileByPath(projectContents, '/src/card.js'),
+    )
+    expect(actualResult).toEqual(expectedResult)
+  })
+  it('returns the same path because there are no hops to take', () => {
```
Rheeseyb, line 576: Please can you add a test case for covering no hops with an instance path that also has a scene path, so that we can ensure that's not a weird potential edge case.

Rheeseyb (approved):

editor/src/components/custom-code/code-file.ts
seanparsons: Yes.

editor/src/components/custom-code/code-file.ts
seanparsons: Shouldn't do, as `animated` is the value in `baseVariable`, which would be the reference to find in the imports.

editor/src/core/shared/template-path.ts
seanparsons: It probably should actually, it did, then this function changed a couple of times and now it probably can again.

editor/src/components/custom-code/code-file.ts
seanparsons: Could probably do with specifying the file that it's halted at so that an error could be constructed from it.

seanparsons: Fixed in 9bf80891.

editor/src/core/shared/template-path.ts
seanparsons: Fixed in 9bf80891.

editor/src/components/custom-code/code-file.spec.ts
seanparsons: Fixed in 9bf80891.

PR #979, opened 2021-03-17
Bring arbitrary JS block component rendering in line with arbitrary JSX blocks

Rheeseyb:
**Problem:**
Due to a discrepancy between how we were transipiling and capturing arbitrary JS blocks, we weren't using the `utopiaCanvasJSXLookup` function to render components used within them, meaning that we weren't actually rendering those components in the same way, and weren't slipping in required props. This was exposed in #972 and was the cause of [this comment](https://github.com/concrete-utopia/utopia/pull/972/files#r595050458)

**Fix:**
Use the `parsedElementsWithin` to ensure we are correctly replacing them with the lookup call when transpiling all arbitrary nodes, and ensure that we have a lookup function in scope (along with those parsed elements within) when evaluating that code. Also, as part of the `babelRewriteJSXArbitraryBlockCode`, since the elements within positions are shifted when dealing with the combined top level arbitrary block, I updated the function `handleByPosition` to also check using the name of the element (updating the name to `handleByPositionOrName`).

**Commit Details:**
- Renamed `handleByPosition` to `handleByPositionOrName` in `parser-printer-transpiling.ts`, and added a function for capturing the name of the Babel element for that
- Include the `elementsWithin` in `ArbitraryJSBlock`
- Pulled apart `innerRender` in our rendering of `JSX_ARBITRARY_BLOCK` to create `createLookupRender`, which could then be used to add a `utopiaCanvasJSXLookup` function to the scope of any arbitrary JS block
- Updated the tests to reflect the use of this lookup (and elements within)

Annoyingly, because I started this as a spike investigation on the branch from #972, the commit log is a bit messy even after merging master. I've double checked that this doesn't undo or replace any of that work.


editor/src/core/workers/parser-printer/parser-printer-transpiling.ts
```diff
@@ -110,15 +114,38 @@ function babelRewriteJSXArbitraryBlockCode(
...
+      return `${nodeName.namespace.name}.${nodeName.name}`
+    }
+  }
+  function handleByPositionOrName(path: BabelTraverse.NodePath<BabelTypes.JSXElement>): void {
     const pathLocation: {
       line: number
       column: number
-    } = path.node.loc.start
-    const foundElementWithin = elementsWithin.find((e) => {
+    } = path.node.loc?.start ?? { line: -1, column: -1 }
```
seanparsons, line 135: Are these being set to `-1` to make the `find` lower down fail?

Rheeseyb: Yeah. So `path.node.loc` is nullable, but we weren't capturing that because we were previously using `any`. I think in practice it won't be `null`, but either way this means the find by location part will fail.

Rheeseyb:
Something is wrong with this PR - it causes the `ui-jsx-canvas.spec` tests to bail with an out of memory error :grimacing: (I have verified that this isn't happening at all on master, so the issue is specific to this branch)

seanparsons (approved):

Rheeseyb:
This PR has now been updated - there was an issue with the mapping of inner elements, which required fixing the location tracking of those, and there was also an issue around our use of `SafeFunction`, in that we were never correctly capturing and passing down `this`, meaning all uses of `SafeFunction` we defaulting to using the top level window's `this` (which this work exposed). To solve that issue the Babel transpile now includes a `callerThis` for passing it in.

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -685,7 +671,7 @@ function parseOtherJavaScript<E extends TS.Node, T>(
         currentScopeInner: Array<string>,
         statement: TS.Statement,
       ): Array<string> {
-        if (TS.isFunctionDeclaration(statement) && statement.name != null) {
+        if (TS.isFunctionLike(statement) && statement.name != null) {
```
seanparsons, line 674: Handy!

seanparsons (approved):

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
Rheeseyb: This caught me about because apparently there is also `TS.isMethodDeclaration`, which of course is completely different to a function declaration since it applies to a class

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
```diff
@@ -21694,7 +21713,7 @@ exports[`UiJsxCanvas render renders fine with two components that reference each
     \\"
     data-uid=\\"utopia-storyboard-uid\\"
   >
-    <div data-uid=\\"BBB\\">great</div>
+    <div data-uid=\\"BBB\\" data-utopia-original-uid=\\"2a2\\">great</div>
```
balazsbajorics, line 21716: this is an interesting change

balazsbajorics (approved):

PR #980, opened 2021-03-19
Focus mode compatible Dom-walker

balazsbajorics:
**Note:**
This PR is still a draft because there's a minor performance regression in the DOM-walker, and there are some unfinished TODOs. However, because it is a _massive_ change and there's some significant change to the TemplatePath/ uid logic, I would like to ask you to take a look at this and give me early feedback. Much appreciated!

-----------

This is the second half of the focus mode work. 
The first half was changing the SpyWrapper  in #972, all that was remaining was "just" changing the dom-walker. 

The main change in this PR is that in the final DOM this is how the data-uid of an element looks like.

Let's take our example project:
```javascript
const HiElement = (props) => {
  return <div data-uid="hi-element-root">hi!</div>;
};

const Button = (props) => {
  return <div data-uid="button-root">{props.children}</div>;
};
const Card = () => {
  return (
    <Button data-uid="button-instance">
      {[0, 1, 2].map((i) => (
        <HiElement data-uid="hi-element" />
      ))}
    </Button>
  );
};
export var App = (props) => {
  return (
    <div data-uid="app-root">
      <div data-uid="inner-div">
        <Card data-uid="card-instance" />
      </div>
    </div>
  );
};

export var storyboard = (
  <Storyboard data-uid="storyboard">
    <Scene data-uid="scene" component={App} />
  </Storyboard>
);
```

Because we only supported editing the target component of a Scene (in our example, that is `App`) we only cared about being able to match the DOM elements with elements inside App. 

App renders Card, which renders Button, which renders a div. If you look at the above JS code, you can see that the div inside Button has a `data-uid: button-root`. 

However, if you look at the rendered DOM, you will not find `button-root` there:
```html
<div data-uid="app-root">
  <div data-uid="inner-div">
    <div data-uid="card-instance"> // <- this div is rendered by Button, original uid button-root. Button is rendered by Card, original uid button-instance. Card is rendered by App, instance uid card-instance.
      <div data-uid="hi-element~~~1" data-utopia-original-uid="hi-element"> // <- this div is actually hi-element-root, our react Monkey Function replacing the UID with the parent's UID
        hi!
      </div>
      <div data-uid="hi-element~~~2" data-utopia-original-uid="hi-element">
        hi!
      </div>
      <div data-uid="hi-element~~~3" data-utopia-original-uid="hi-element">
        hi!
      </div>
    </div>
  </div>
</div>
```
That is because we only cared about editing elements that are inside App, and being able to click on card-instance was our goal. Being able to match card-instance to something rendered in the DOM meant that we had to monkey with the rendered UIDs. To achieve this, we created the monkey patched react createElement, which would make sure that if a rendered component had a data-uid prop, that prop would overwrite any existing data-uid rendered by the actual element.
So Button, rendered by Card, rendered by App would have a data-uid of `card-instance`. 

And this mostly works all right, however there were some edge cases where we had to fight around the limitations (for example there's some code building a `data-utopia-parents` path).

However, in the new world where we also want to be able to focus elements deeper in the hierarchy to highlight, select and manipulate them, it is no longer sufficient to replace the UID `button-root` with the UID `card-instance`. We actually need a way to represent the fact that the `div` in the DOM is there because it is the root element of Button which is the root element of Card which is an element in App. We want to understand that `div` in all three roles: sometimes you want to be able to select Card, sometimes Button, but sometimes you want to focus on the Button component and see that div as a div.

To achieve this, I changed the monkey logic so it appends uids together, and prints them as a space-separated array:
```html
<div data-uid="app-root scene">
  <div data-uid="inner-div">
    <div data-uid="button-root button-instance card-instance"> // <- multiple uids separated by space
      <div
        data-uid="hi-element-root hi-element~~~1" // <- generated uid hi-element~~~1
        data-utopia-original-uid="hi-element"
      >
        hi!
      </div>
      <div
        data-uid="hi-element-root hi-element~~~2"
        data-utopia-original-uid="hi-element"
      >
        hi!
      </div>
      <div
        data-uid="hi-element-root hi-element~~~3"
        data-utopia-original-uid="hi-element"
      >
        hi!
      </div>
    </div>
  </div>
</div>
```

The rest of the PR deals with the rippling effects of this, most notably in the dom-walker.


editor/src/components/canvas/dom-walker.ts
```diff
@@ -53,12 +53,33 @@ import { PRODUCTION_ENV } from '../../common/env-vars'
...
+import { mapDropNulls } from '../../core/shared/array-utils'
+import { optionalMap } from '../../core/shared/optional-utils'
 
 const MutationObserverConfig = { attributes: true, childList: true, subtree: true }
 const ObserversAvailable = (window as any).MutationObserver != null && ResizeObserver != null
 
-function isValidPath(path: TemplatePath | null, validPaths: Array<string>): boolean {
-  return path != null && validPaths.indexOf(TP.toString(path)) > -1
+function findValidPath(uid: string | null, validPathsString: Array<string>): InstancePath | null {
+  const validPaths = validPathsString.map(TP.fromString) // Hi Rheese :)
```
Rheeseyb, line 64: :wave:

editor/src/components/canvas/dom-walker.ts
```diff
@@ -53,12 +53,33 @@ import { PRODUCTION_ENV } from '../../common/env-vars'
...
+  if (uid == null) {
+    return null
+  }
```
Rheeseyb, line 67: Might as well move the early return up

editor/src/components/canvas/dom-walker.ts
```diff
@@ -53,12 +53,33 @@ import { PRODUCTION_ENV } from '../../common/env-vars'
...
+  return (
+    mapDropNulls((validPath) => {
+      if (TP.isScenePath(validPath)) {
+        return null // TODO do we need to care about a scene path here?
```
Rheeseyb, line 71: I don't think so. I'm pretty sure the valid paths are always instance paths anyway

editor/src/components/canvas/dom-walker.ts
```diff
@@ -748,93 +768,82 @@ function walkElements(
...
-
-    // Build the unique path for this element.
-    let uniquePath: TemplatePath = uniqueParentPath
-    if (parentUIDsAttribute != null) {
-      uniquePath = TP.appendToPath(uniquePath, parentUIDsAttribute.split('/'))
-    }
-    uniquePath = TP.appendToPath(uniquePath, pathElement)
+    const uids = getUIDsOnDomELement(element)
+    if (uids != null) {
+      const results = uids.map((uidAttribute) => {
```
Rheeseyb, line 777: Worth adding a `FIXME` here to save accidentally merging - right now this will mean we'll walk the DOM multiple times for each element (I know you mentioned knowing this, I'm just adding the comment in case that gets forgotten about amongst the other work in here)

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1134,12 +1134,10 @@ export const MetadataUtils = {
         let componentInstance = spyElem.componentInstance || domElem.componentInstance
         let jsxElement = alternativeEither(spyElem.element, domElem.element)
 
-        const possibleUID: string | null | undefined = Utils.defaultIfNull(
-          spyElem.props[UTOPIA_UID_KEY],
-          spyElem.props[UTOPIA_ORIGINAL_ID_KEY],
-        )
-        if (possibleUID != null) {
-          const possibleElement = elementsByUID[possibleUID]
+        const possibleUIDs: Array<string> | null = uidsFromString(spyElem.props[UTOPIA_UIDS_KEY]) // TODO what about original-uid???
```
Rheeseyb, line 1137: I think `original-uid` will also have to become a space separated list, but that would mean all elements would need it (since right now we only add it for generated elements, meaning the length of the uid arrays would become out of sync). But then that brings about the question of whether we still need them, since we're already using a function for converting a dynamic uid to a static one. I _think_ we don't.

editor/src/core/shared/template-path.ts
```diff
@@ -899,18 +899,20 @@ export function dynamicPathToStaticPath(path: TemplatePath): StaticTemplatePath
...
 
+// TODO maybe delete me
 export function scenePathContainsElementPath(scene: ScenePath, elementPath: ElementPath): boolean {
   return scene.sceneElementPaths.some((sceneElementPath) =>
     elementPathsEqual(sceneElementPath, elementPath),
   )
 }
 
-export function staticScenePathContainsElementPath(
+export function staticScenePathContainsElementPath( // TODO rename me!
```
Rheeseyb, line 909: `staticScenePathContainingElementPath` maybe?

editor/src/core/model/element-metadata-utils.ts
balazsbajorics: I also think we don't, so long as we stick to the `~~~index` scheme of indicating a generated uid

editor/src/core/shared/uid-utils.ts
```diff
@@ -208,3 +212,33 @@ export function fixUtopiaElement(
...
+
+export function appendToUidString(
+  uidsString: string | null | undefined,
+  uidsToAppend: string | null | undefined,
+): string | null {
+  if (uidsToAppend == null) {
+    return uidsString ?? null
+  } else if (uidsString == null || uidsString.length === 0) {
+    return uidsToAppend
+  } else if (R.difference(uidsFromString(uidsToAppend), uidsFromString(uidsString)).length > 0) {
```
Rheeseyb, line 232: This is probably a very expensive function. The docs mention using "value equality, not reference equality", implying it's probably using the extra expensive `R.equals` function.

editor/src/utils/canvas-react-utils.ts
```diff
@@ -119,28 +120,30 @@ function shouldIncludeDataUID(type: any): boolean {
...
     return originalResponse
   } else {
     if (shouldIncludeDataUID(originalResponse.type)) {
-      return React.cloneElement(originalResponse, { [UTOPIA_UID_KEY]: dataUid })
+      return React.cloneElement(originalResponse, {
+        [UTOPIA_UIDS_KEY]: appendToUidString(originalResponse.props[UTOPIA_UIDS_KEY], dataUids),
+      })
     } else {
-      return React.cloneElement(originalResponse)
+      return React.cloneElement(originalResponse) // This is probably an unnecessary clone
```
Rheeseyb, line 146: Good catch. It looks like it is (in other cases we're returning `originalResponse`)

editor/src/utils/utils.ts
```diff
@@ -921,7 +921,7 @@ function timeLimitPromise<T>(promise: Promise<T>, limitms: number, message: stri
     const timeoutID = setTimeout(() => {
       clearTimeout(timeoutID)
       reject(message)
-    })
+    }, limitms)
```
Rheeseyb, line 924: Hahahahaha omg

balazsbajorics: yup

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -34,17 +35,20 @@ export function findFirstParentWithValidUID(
   validTemplatePathsForLookup: Array<string> | 'no-filter',
   target: Element,
 ): string | null {
-  const uid = getDOMAttribute(target, 'data-uid')
-  const originalUid = getDOMAttribute(target, 'data-utopia-original-uid')
+  const uids = getUIDsOnDomELement(target)
+  const originalUid = getDOMAttribute(target, 'data-utopia-original-uid') // TODO BALAZS What's up with original-uid?
```
Rheeseyb, line 39: Heads up, I'm binning this off right now and will open a separate PR for that to unblock this PR.

editor/src/components/canvas/dom-walker.ts
```diff
@@ -346,23 +367,23 @@ export function useDomWalker(props: CanvasContainerProps): React.Ref<HTMLDivElem
...
   const globalFrame = globalFrameForElement(element, scale, containerRectLazy)
   const localFrame = localRectangle(Utils.offsetRect(globalFrame, Utils.negate(parentPoint)))
 
-  const uidAttribute = getDOMAttribute(element, UTOPIA_UID_KEY)
   const originalUIDAttribute = getDOMAttribute(element, UTOPIA_ORIGINAL_ID_KEY)
   const labelAttribute = getDOMAttribute(element, UTOPIA_LABEL_KEY)
   let elementProps: any = {}
   if (uidAttribute != null) {
-    elementProps[UTOPIA_UID_KEY] = uidAttribute
+    elementProps[UTOPIA_UIDS_KEY] = uidAttribute // TODO Balazs we are making a fake prop with a single UID instead of the UID array – maybe this means changes to mergeComponentMetadata
```
Rheeseyb, line 386: This (and the other instance later down with a similar comment) are also going to be binned in my "other PR"

editor/src/components/canvas/dom-walker.ts
Rheeseyb: This is being worked on by Balazs right now, but will be submitted as a separate PR to unblock this one.

editor/src/core/model/element-template-utils.ts
```diff
@@ -126,31 +128,35 @@ export function getValidTemplatePathsFromElement(
...
-      TP.staticScenePathContainsElementPath(focusedElementPath, TP.elementPathForPath(path))
-    ) {
+    const matchingFocusedPathPart =
+      focusedElementPath == null
+        ? null
+        : TP.staticScenePathContainsElementPath(focusedElementPath, TP.elementPathForPath(path))
+
+    if (matchingFocusedPathPart != null) {
+      // storyboard/scene:app-root/inner-div/card-instance:button-instance/hi-element~~~2
+      // storyboard/scene:app-root/inner-div/card-instance
```
Rheeseyb, line 138: Should probably delete these comments now @balazsbajorics

editor/src/core/shared/template-path.ts
Rheeseyb: I'll take this (and the above `delete me`)

editor/src/core/shared/uid-utils.ts
Rheeseyb: I'll take this

editor/src/utils/canvas-react-utils.ts
Rheeseyb: I'll take this

editor/src/utils/canvas-react-utils.ts
```diff
@@ -211,37 +214,40 @@ const mangleExoticType = Utils.memoize(
...
-          pathParts.push(existingChildUID)
+        if (typeof existingChildUIDs === 'string') {
+          const childUIDArray = uidsFromString(existingChildUIDs)
+          pathParts.push(childUIDArray[0]) // TODO Balazs we are arbitrarily picking the first element here
         }
-        if (typeof dataUid === 'string') {
-          pathParts.push(dataUid)
+        if (typeof dataUids === 'string') {
+          const uidArray = uidsFromString(dataUids)
+          pathParts.push(uidArray[0]) // TODO Balazs we are arbitrarily picking the first element here
```
Rheeseyb, line 235: Looks like this needs addressing. I don't mind pulling it into my spring cleaning work @balazsbajorics ?

balazsbajorics: good catch – actually I think we should remove the pathParts altogether, alongside with `data-utopia-original-parents` – I don't mind if you deal with this in the `data-original-uid` PR

editor/src/core/model/element-template-utils.ts
```diff
@@ -126,31 +128,33 @@ export function getValidTemplatePathsFromElement(
...
+
+    if (matchingFocusedPathPart != null) {
       paths = [
         ...paths,
         ...getValidTemplatePaths(
           topLevelElements,
           focusedElementPath,
           name,
-          TP.staticScenePathForElementAtInstancePath(path),
+          matchingFocusedPathPart, // TP.scenePathForElementAtInstancePath(path),
```
enidemi, line 143: comment! pls delete

enidemi (approved):

PR #982, opened 2021-03-19
Feature/navigator focused elements

enidemi:
This PR changes how focused elements are shown in the navigator.

Changes:
- showing focused element children in the navigator
- 3 new icons
- helpers in `template-path` for the real depth
- style changes

Screenshot 2021-03-19 at 15 04 59


editor/src/core/shared/template-path.ts
```diff
@@ -360,6 +360,26 @@ export function depth(path: TemplatePath): number {
...
+    })
+    return scenePath + path.element.length
+  }
+}
+
+export function isInsideFocusedComponent(path: TemplatePath): boolean {
+  if (isScenePath(path)) {
+    return false
+  } else {
+    return path.scene.sceneElementPaths.length > 1
```
balazsbajorics, line 379: tricky

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -187,45 +256,36 @@ export const NavigatorItem: React.FunctionComponent<NavigatorItemInnerProps> = b
     } = props
 
     const domElementRef = useScrollToThisIfSelected(selected)
+    const isFocusedComponent = useEditorState((store) => {
```
balazsbajorics, line 259: I think it'd be worth making a function (in this file) that calculates `isFocusedComponent`, because we are doing similar but not the same logic in two places and I suspect that we might need to update this (because the dom-walker work is not 100% finished yet)

balazsbajorics (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
Rheeseyb: :+1:

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -794,6 +797,23 @@ export const MetadataUtils = {
...
+      const matchingFocusPath =
+        focusedElementPath == null
+          ? null
+          : TP.staticScenePathContainsElementPath(focusedElementPath, TP.elementPathForPath(path))
+      const focusedRootElementPaths =
+        matchingFocusPath == null
+          ? []
+          : allPaths.filter(
+              (p) =>
+                TP.depth(p) === 2 && // TODO this is actually pretty silly, TP.depth returns depth + 1 for legacy reasons
```
Rheeseyb, line 810: It's probably worth extracting the depth check then to a function in `template-path.ts` that checks if the path is a root element path

Rheeseyb (approved):

PR #998, opened 2021-03-22
Chore/dedupe dom walker

balazsbajorics:
**Problem:**
The dom-walker would repeat dom lookups for elements (and their subtree) that have multiple uids. This is not good for performance.

**Fix:**
Tweak the `walkElements` function so that it only does dom lookups once per element, and creates multiple ElementInstanceMetadata for each valid uid

**Commit Details:**
- Removed unused `uniqueParentPath`
- Split the element lookup parts of `collectMetadata` to `collectMetadataForElement`, only call it once per element
- removed mapping over the uids inside `walkElements`
- fixed an issue that the scene path needs to be used as a scene path for child path validation, but used as an instance path for generating the ElementInstanceMetadata's key


editor/src/components/canvas/dom-walker.ts
```diff
@@ -431,7 +473,7 @@ function getComputedStyle(
   const elementStyle = window.getComputedStyle(element)
   const attributesSetByStylesheet = getCachedAttributesComingFromStyleSheets(
     invalidatedPathsForStylesheetCacheRef,
-    path,
+    paths[0], // TODO is this sufficient to use the first path element for caching?
```
Rheeseyb, line 476: If I have
```
const Cake = () => {
  return <div data-uid='cake-root' />
}
const App = () => {
  return <Cake data-uid='app-root' />
}
```
The paths array would contain the path for `'app-root'` and `'cake-root'`, but which would come first?

If the answer is `app-root`, then this should be fine, but if the answer is `cake-root` then I think that would be a problem because it would mean the first path changes as we dig deeper via focusing, so every focus would blow the cache

balazsbajorics: I _think_ we are fine blowing the cache for every refocus. this cache mostly exists to make scrolling speedier, and only re-calculate relevant things while dragging. Actions such as change selection already blow caches in the dom-walker

enidemi (approved):

Rheeseyb (approved):

PR #1004, opened 2021-03-24
Chore/component renderer no context

enidemi:
**Problem:**
For the multifile focus element feature the canvas context needs to handle different topLevelElements from different files with their own scopes. This PR is the first step, it clears the topLevelElements from the canvas context and replaces it with searching for the element from the editorstate based on filename.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
@@ -9,8 +9,15 @@ import {
   emptyComputedStyle,
   JSXElement,
   emptyAttributeMetadatada,
+  isUtopiaJSXComponent,
+  UtopiaJSXComponent,
 } from '../../../core/shared/element-template'
-import { InstancePath, ScenePath } from '../../../core/shared/project-file-types'
+import {
+  InstancePath,
+  isParseSuccess,
+  isTextFile,
+  ScenePath,
+} from '../../../core/shared/project-file-types'
 import { colorTheme, UtopiaStyles } from '../../../uuiui'
 import { UiJsxCanvasContextData, UiJsxCanvasContext } from '../ui-jsx-canvas'
 import {
@@ -33,6 +40,9 @@ import utils from '../../../utils/utils'
 import { PathForResizeContent } from '../../../core/model/scene-utils'
 import { UTOPIA_SCENE_PATH, UTOPIA_UIDS_KEY } from '../../../core/model/utopia-constants'
 import { optionalMap } from '../../../core/shared/optional-utils'
+import { useEditorState } from '../../editor/store/store-hook'
+import { getFileForName, getOpenUIJSFileKey } from '../../editor/store/editor-state'
+import { fastForEach } from '../../../core/shared/utils'
 
 interface SceneProps {
   component?: React.ComponentType | null
@@ -96,7 +106,21 @@ function useGetValidTemplatePaths(
   topLevelElementName: string | null,
   scenePath: ScenePath,
 ): Array<InstancePath> {
-  const topLevelElements = useContextSelector(RerenderUtopiaContext, (c) => c.topLevelElements)
+  const topLevelElements: Map<string, UtopiaJSXComponent> = useEditorState((store) => {
+    const uiFilePath = getOpenUIJSFileKey(store.editor)
+    const projectFile = uiFilePath ? getFileForName(uiFilePath, store.editor) : null
+    if (isTextFile(projectFile) && isParseSuccess(projectFile.fileContents.parsed)) {
+      let topLevelJsxComponents: Map<string, UtopiaJSXComponent> = new Map()
+      fastForEach(projectFile.fileContents.parsed.topLevelElements, (topLevelElement) => {
+        if (isUtopiaJSXComponent(topLevelElement)) {
+          topLevelJsxComponents.set(topLevelElement.name, topLevelElement)
+        }
+      })
+      return topLevelJsxComponents
+    } else {
+      return {} as Map<string, UtopiaJSXComponent>
+    }
+  }, 'useGetValidTemplatePaths')
   const focusedElementPath = useContextSelector(RerenderUtopiaContext, (c) => c.focusedElementPath)
   return getValidTemplatePaths(
     topLevelElements,
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
@@ -29,6 +29,7 @@ import {
   ParentLevelUtopiaContext,
   RerenderUtopiaContext,
   SceneLevelUtopiaContext,
+  UtopiaProjectContext,
 } from './ui-jsx-canvas-contexts'
 import { renderComponentUsingJsxFactoryFunction } from './ui-jsx-canvas-element-renderer-utils'
 import * as TP from '../../../core/shared/template-path'
@@ -43,7 +44,7 @@ import { optionalMap } from '../../../core/shared/optional-utils'
 import { useEditorState } from '../../editor/store/store-hook'
 import { getFileForName, getOpenUIJSFileKey } from '../../editor/store/editor-state'
 import { fastForEach } from '../../../core/shared/utils'
-import { getTopLevelElements } from './ui-jsx-canvas-top-level-elements'
+import { getTopLevelElements, useGetTopLevelElements } from './ui-jsx-canvas-top-level-elements'
 
 interface SceneProps {
   component?: React.ComponentType | null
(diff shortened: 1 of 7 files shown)


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -49,10 +59,24 @@ export function createComponentRendererComponent(params: {
...
+              return elem
+            } else {
+              return null
+            }
+          }, projectFile.fileContents.parsed.topLevelElements)[0] ?? null
+        )
+      } else {
+        return null
+      }
+    }, 'createComponentRendererComponent utopiaJsxComponent')
```
Rheeseyb, line 79: This block will run every time the component's props change, right? Could we move the block outside of the component and into the `createComponentRendererComponent` (meaning it would have to become a hook) or into the `ui-jsx-canvas`?

balazsbajorics: we need to re-render the component if the related topLevelElement changes. but we want to avoid re-mounting the component, so it needs to be a hook and it needs to live inside the component renderer component. it's not a problem that this hook is called every time the component is rendered, since the useEditorState hooks are designed to be cheap and fast

editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
```diff
@@ -96,10 +107,23 @@ function useGetValidTemplatePaths(
   topLevelElementName: string | null,
   scenePath: ScenePath,
 ): Array<InstancePath> {
-  const topLevelElements = useContextSelector(RerenderUtopiaContext, (c) => c.topLevelElements)
+  const topLevelJSXComponentsMap: Map<string, UtopiaJSXComponent> = useEditorState((store) => {
```
balazsbajorics, line 110: there is a problem with memoization here: since this is a Map, it will not pass shallow equal, because we create a new Map inside every time. I think there's not much point in keeping this a Map, so we should turn this into a MapLike object instead. that would fix the shallow equality, and the breaking render count test

balazsbajorics: actually, it turns out shallowEquals did not at all handle Maps correctly. since that is a potentially huge problem, we decided to fix it for Maps. Now this shallowEquals works fine for this maps case here too, so disregard my previous comment!

balazsbajorics:
this PR is now ready for your review! we believe we found and fixed the problem. I'll write it up in a minute, but I need a few minutes to recover first :D

Rheeseyb (approved):

balazsbajorics:
so here's what we found:
the previous version of this PR used `useEditorState` to get the projectContents from editor. This was working with the exception of a single test – we started debugging it and the reaons nearly broke our brains. 

When we introduced the ui-jsx-canvas, we immediately introduced a form of spying to be able to extract metadata about the rendered elements. At some point in time the way it worked was that it would use ReactDOM.renderToString() to render the canvas off-screen, collect the spy stuff, update the editor state, _then_ render the real canvas with the final editor state. 
When we later introduced the dom-walker, this approach stopped being feasible: we needed the canvas to modify the DOM so we can run measurements and compute the styles. So iteratively, our render process became something like this:
→ User Action `dispatched` (like SCROLL_CANVAS) 
→ dispatch update functions calculate new state 
→ `Canvas renders` (→ the spy wrapper collects spied information to a mutable ref) → DOM Updated 
→ React.useLayoutEffect inside dom-walker gets triggered 
→ UPDATE_DOM_REPORT Action `dispatched` 
→ dispatch update function for DOM_REPORT runs, merges the spy metadata (that was being collected in the mutable ref) and the dom-walker metadata (coming from the action) 
→ `Canvas renders` *again* (→ the spy wrapper collects spied information to a mutable ref) → DOM Updated 
-> UPDATE_DOM_REPORT called second time, this time around it finds no changes, and settles, we leave the editorState unchanged.
If the second render changes something on the DOM, the DOM_REPORT **will** find new changes, it will update the editor state, and the Canvas will render for a third time.

Now, there's two big problems with this approach, and it has been like this for months: 
1. The canvas renders twice for no real reason, only because of the iterative changes we made to the system and no one spotting it until it was on master and too expensive to be dealt with
2. We are at the mercy of the second DOM_REPORT being identical to the first one. If that's not the case, the canvas will keep re-rendering until two subsequent dom reports match
3. **well, actually** it turns out that when you are duplicating Elements that have children...😅it only works by total accident. on Master, when the canvas renders for a second time around, somehow the SAVE_DOM_REPORT action gets dispatched _before_ the spy would be able to run. That means that we compare the old spy results to the old spy results (which will always return equality) and the new dom report to the old dom report (which _thank god_ ends up being the same) and that's why we don't lock up the editor in an infinite loop and that's why we thought that the drag to duplicate feature is fine.


_However_, with the changes we made in this PR, we used `useEditorState` inside the ComponentRendererComponent which slightly changed the order of execution:
→ User Action `dispatched` (like SCROLL_CANVAS) 
→ dispatch update functions calculate new state
**→ all the ComponentRendererComponents and SceneRoots re-render because of their `useEditorState` hook (→ the spy wrapper collects spied information to a mutable ref)**
→ `Canvas renders` (→ the spy wrapper collects spied information to a mutable ref) → DOM Updated 
→ React.useLayoutEffect inside dom-walker gets triggered 
→ UPDATE_DOM_REPORT Action `dispatched` 
→ dispatch update function for DOM_REPORT runs, merges the spy metadata (that was being collected in the mutable ref) and the dom-walker metadata (coming from the action) 
**→ all the ComponentRendererComponents and SceneRoots re-render because of their `useEditorState` hook (→ the spy wrapper collects spied information to a mutable ref)**
→ `Canvas renders` *again* (→ the spy wrapper collects spied information to a mutable ref) → DOM Updated 
-> UPDATE_DOM_REPORT called second time, this time around it finds no changes, and settles, we leave the editorState unchanged.

This change is mostly fine, and we didn't notice any regressions. With the exception of a single breaking test.

The drag-to-duplicate action is buggy: for every single render mid-drag, it will create a new `uid` for the children and grandchildren (etc) of the duplicated element. while it correctly stores and preserves a duplicate UID for the dragged element, that logic does not extend to the children: so long as you are dragging, every frame will yield a new uid for those. It's _fine_ because we don't want to be able to select those elements, it is definitely not aesthetic. But because it caused no harm, we left it like this.
However, the slight change in the order of execution means that the spy metadata gets updated a tiny bit earlier, and by the time the UPDATE_DOM_REPORT gets called, the spy metadata is already fresh: (hilariously, this is the CORRECT behavior, but alas) this means that we see that the uid of the duplicated grandchildren are different: we then create a new editor state, new canvas render. But by creating a new editorState, we also create a new DerivedState, creating a new transient canvas drag state, creating even newer UIDs for the duplicated children... which means the subsequent UPDATE_DOM_REPORT will also be (correctly) seeing a change in the spy metadata, triggering yet another state update... repeat ad infinitum. 

---------

So our slight change in the order of re-render for the various pieces inside ui-jsx-canvas, we accidentally fixed a bug in how the spy and dom metadata is reconciled, which accidentally highlighted a bug in the duplicate logic. 

What should be the solution? Fix the duplicate logic? Make the spying system more robust? No!

For now we decided to not mess with these things because it works mostly fine on the Canvas. We instead decided to revert the useEditorStates and move back to using Contexts created by the canvas to preserve the order of execution as it was on master. 
What we learned today is that if you need to control order of re-rendering for you components because they are built like a hosue of cards, Contexts are a very handy tool to give you more control over what is happening.


I should probably create follow up tickets to fix the duplicate action, only run the Canvas once per dispatch and figure out why does the spying even work on master.

editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
```diff
@@ -96,10 +108,20 @@ function useGetValidTemplatePaths(
   topLevelElementName: string | null,
   scenePath: ScenePath,
 ): Array<InstancePath> {
-  const topLevelElements = useContextSelector(RerenderUtopiaContext, (c) => c.topLevelElements)
+  const uiFilePath = useContextSelector(UtopiaProjectContext, (c) => c.openStoryboardFilePathKILLME)
+
+  const topLevelElements = useGetTopLevelElements(uiFilePath ?? '')
```
seanparsons, line 113: `?? ''` <- This feels a bit wrong, surely the hook should handle the nullable case?

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-top-level-elements.ts
```diff
@@ -0,0 +1,55 @@
...
+import { useEditorState } from '../../editor/store/store-hook'
+import { UtopiaProjectContext } from './ui-jsx-canvas-contexts'
+
+export function getTopLevelElementsFromEditorState(
+  filePath: string,
+  editor: EditorState,
+  derived: DerivedState,
+): TopLevelElement[] {
+  return getTopLevelElements(
+    filePath,
```
seanparsons, line 21: Shouldn't this only rely on `getOpenUIJSXFileKey` and not `filePath` at all if it's for the canvas?

seanparsons: Ah no, I see, this could be doing a lookup for the elements used by other elements.

seanparsons (approved):

PR #1008, opened 2021-03-25
Chore/canvas extract execution scope

enidemi:
This PR is the next step after https://github.com/concrete-utopia/utopia/pull/1004 to enable multifile canvas focused components. This doesn't change any current functionality of the canvas yet.
There are 2 changes:
- the mutableContext stores scopes and results based on filepath
- from ui-jsx-canvas `createComponentRendererComponent` is moved to a function that it can work based on filepath not just for storyboard

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-contexts.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-contexts.tsx
@@ -2,9 +2,10 @@ import * as React from 'react'
 import type { MapLike } from 'typescript'
 import { createContext } from 'use-context-selector'
 import { EmptyScenePathForStoryboard } from '../../../core/model/scene-utils'
-import type { UtopiaJSXComponent } from '../../../core/shared/element-template'
+import type { TopLevelElement, UtopiaJSXComponent } from '../../../core/shared/element-template'
 import type { InstancePath, ScenePath, TemplatePath } from '../../../core/shared/project-file-types'
-import type { UIFileBase64Blobs } from '../../editor/store/editor-state'
+import { ProjectContentTreeRoot } from '../../assets'
+import type { TransientFileState, UIFileBase64Blobs } from '../../editor/store/editor-state'
 
 export interface MutableUtopiaContextProps {
   requireResult: MapLike<any>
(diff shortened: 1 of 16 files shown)


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-contexts.tsx
```diff
@@ -8,19 +8,18 @@ import { ProjectContentTreeRoot } from '../../assets'
 import type { TransientFileState, UIFileBase64Blobs } from '../../editor/store/editor-state'
 
 export interface MutableUtopiaContextProps {
-  requireResult: MapLike<any>
-  fileBlobs: UIFileBase64Blobs
-  rootScope: MapLike<any>
-  jsxFactoryFunctionName: string | null
+  [filePath: string]: {
+    mutableContext: {
```
balazsbajorics, line 12: note: the reason we has this funky object shape is that requireResult and rootScope are effectively `any`s. We originally refactored the code to this: 

```typescript
export interface MutableUtopiaContextProps {
   [filePath: string]: {
     requireResult: MapLike<any>
     fileBlobs: UIFileBase64Blobs
     rootScope: MapLike<any>
     jsxFactoryFunctionName: string | null
  }
}
```

but since the places we use this object looked like this: `mutableContext.requireResult`, these places would not break the TS build, because mutableContext.requireResult is inferred to be of the shape 
```
{
     requireResult: MapLike<any>
     fileBlobs: UIFileBase64Blobs
     rootScope: MapLike<any>
     jsxFactoryFunctionName: string | null
  }
```
which can be passed into an `any`. So we wrapped the whole thing in yet another object with a single key, `mutableContext` to break out of `any` prison and let Typescript help us find all the places that use mutableContext in the old way

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.ts
```diff
@@ -0,0 +1,69 @@
...
+  topLevelJsxComponents: Map<string, UtopiaJSXComponent>
+  requireResult: MapLike<any>
+} {
+  const { topLevelElements, imports } = getTopLevelElements(
+    filePath,
+    projectContents,
+    openStoryboardFileNameKILLME,
+    transientFileState,
+  )
+  const requireResult: MapLike<any> = importResultFromImports(filePath, imports, customRequire)
```
balazsbajorics, line 33: this code is mostly moved verbatim from ui-jsx-canvas, with the exception of the use of `getTopLevelElements` above

seanparsons (approved):

balazsbajorics (approved):

Rheeseyb (commented):
:+1:

PR #1017, opened 2021-03-29
Feature/require parse success

enidemi:
**Problem:**
With multifile project only the main file components can be focused and edited. This PR adds element metadata for components imported from other files in the project.

**Commit Details:**
- for the DOM metadata the Scene's `getValidTemplatePaths` was modified to collect imported component information, using `importedFromWhere` and the projectContents `parseSuccess` to get the correct topLevelElement.
- the `customRequire` is where the file level scope is created containing the correct `ComponentRendererComponents` to have the spied metadata. The previous PRs made it possible to create filebased scopes. Non-exported items are filtered out from this scope.
- there are some TODOs left for next PRs


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-contexts.tsx
```diff
@@ -49,11 +50,14 @@ interface UtopiaProjectContextProps {
   projectContents: ProjectContentTreeRoot
   openStoryboardFilePathKILLME: string | null
   transientFileState: TransientFileState | null
+  resolve: (importOrigin: string, toImport: string) => Either<string, string>
 }
+const EmptyResolve = (): Either<string, string> => left('')
```
seanparsons, line 55: Should there be an error message in that string?

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -267,10 +283,68 @@ export const UiJsxCanvas = betterReactMemo(
...
+              return requireFn(importOrigin, toImport, false)
+            },
+            (scope) => {
+              // Return an artificial exports object that contains our ComponentRendererComponents
+              return scope
+            },
+            resolvedParseSuccess,
+          )
+        },
+        // TODO I don't like projectContents and transientFileState here because that means dragging smth on the Canvas would recreate the customRequire fn
```
seanparsons, line 336: Danger Will Robinson, performance?

seanparsons (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-contexts.tsx
Rheeseyb: This definitely feels wrong to me too.

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -267,10 +283,68 @@ export const UiJsxCanvas = betterReactMemo(
       clearErrors()
     }
 
+    // TODO after merge requireFn can never be null
```
Rheeseyb, line 286: I don't understand what this TODO is signifying?

balazsbajorics: hah! I'll fix the todo. basically we can remove this IF from here, but that would de-indent the entire function making the PR garbage

editor/src/components/canvas/ui-jsx-canvas.tsx
Rheeseyb: I'm assuming this is part of the next load of work?

Rheeseyb (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
balazsbajorics: well this is a placeholder for discussion – right now we do not have a better idea than this, and we are actually re-creating the global requireFn for any file change as well, so I think this is definitely not a performance _regression_ just an area to improve

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-contexts.tsx
enidemi: fixed

PR #1019, opened 2021-03-30
adding new rule to disallow src/core/shared to reach the rest of the editor

balazsbajorics:
the branch has 39 dependency cruiser errors. 
some of these errors are valid and need fixing in the codebase
other errors are not valid and we need to tweak the ruleset a bit


Rheeseyb:
Tagged and closed

PR #1033, opened 2021-03-31
Fix multifile canvas selection

balazsbajorics:
**Problem:**
Focused elements from other files could be selected from the navigator, they would get a proper highlight / selection outline on the canvas, but clicking on them on the canvas could not select them. The problem is that the `select-mode-hooks@getSelectableViews` and `canvas@getAllTargetsAtPoint` would both rely on `MetadataUtils.getImmediateChildren` which only returned real children elements, and could not include the "unfurled" focused component's elements.

**Fix:**
I made a new function called `getAllChildrenIncludingUnfurledFocusedComponents` which returns the children + the focused elements.

**Commit Details:**
- Made new `getAllChildrenIncludingUnfurledFocusedComponents`
- Moved code from `createOrderedTemplatePathsFromElements` to avoid duplication
- Created new `getImmediateChildrenPaths` that returns the instance paths instead of the metadata to speed things up a little
- Also made a `getAllChildrenElementsIncludingUnfurledFocusedComponents` that returns `ElementInstanceMetadata[]` instead of instance paths.

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -1749,6 +1749,7 @@ function getReparentTargetAtPosition(
   componentMeta: JSXMetadata,
   selectedViews: Array<TemplatePath>,
   hiddenInstances: Array<TemplatePath>,
+  focusedElementPath: ScenePath | null,
   canvasScale: number,
   canvasOffset: CanvasVector,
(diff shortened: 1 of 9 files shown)


seanparsons (approved):

Rheeseyb (commented):
:+1:

PR #1035, opened 2021-03-31
New/UI row

OmarDaSilva:
Fix styling inconsistencies and new styled Component 

**Problem:**
Inconsistent styling 

**Fix:**

Created a new styled component UIRow with optional props padded and rowHeight. This makes it easier to maintain consistency in the styling and replaces all instances of <FlexRow height=... padding=...> where we provide a height and/or padding, but not any <FlexRow> with no modifications or with a lot of very unique modifications. 

Amended GridRow. Now is known as UIGridRow, property 'type' renamed to layout for clarity.


OmarDaSilva:
PR now ready for feedback!

editor/src/components/canvas/right-menu.tsx
```diff
@@ -69,8 +69,8 @@ export const RightMenuTile: React.FunctionComponent<RightMenuTileProps> = (props
     <SimpleTile
       style={{
         ...props.style,
-        width: UtopiaTheme.layout.rowHeight.mediumLarge,
-        height: UtopiaTheme.layout.rowHeight.mediumLarge,
+        width: 38,
```
Rheeseyb, line 72: How come these (and later values too) weren't updated in the `UtopiaTheme`? Should they be?

Rheeseyb: Ok looks like they were actually removed from there. Please can you put them re-add them in a sensible place.

editor/src/uuiui/widgets/layout/ui-row.tsx
```diff
@@ -0,0 +1,18 @@
...
+export interface UIRowProp extends React.InputHTMLAttributes<HTMLDivElement> {
+  padded?: boolean
+  rowHeight?: keyof typeof UtopiaTheme.layout.rowHeight
+}
+
+export const UIRow = styled.div<UIRowProp>((props) => ({
+  ...commonSenseUtopiaLayoutShorthands,
+  ...flexRowStyle,
+  padding: props.padded ? UtopiaTheme.layout.rowHorizontalPadding : undefined,
+  height: props.rowHeight ?? UtopiaTheme.layout.rowHeight.normal,
```
Rheeseyb, line 16: This looks wrong. Shouldn't it be something like:
```
height: 
  props.rowHeight == null ? 
    UtopiaTheme.layout.rowHeight.normal :
    UtopiaTheme.layout.rowHeight[props.rowHeight]
```
since `props.rowHeight` will be a string (e.g. `'tall'`)?

seanparsons (approved):

editor/src/components/canvas/right-menu.tsx
OmarDaSilva: There weren't many places of components using deleted values such as UtopiaTheme.lay...mediumLarge, and we considered them to be unusual values so instead, we just set them in style to indicate that. Do you still think I should re-add them in?

balazsbajorics: it sounds like if those are rarely used and unusual, it's better to have them locally 👍

balazsbajorics (approved):

editor/src/components/canvas/right-menu.tsx
Rheeseyb: Ok sure that's fine then

maltenuhn:
@OmarDaSilva could you merge this please? I'm hoping to use it in a few places!

PR #1036, opened 2021-04-01
Merge ComponentMetadata into ElementInstanceMetadata (part 1)

Rheeseyb:
**Problem:**
We no longer want or need the distinction between `ComponentMetadata` and `ElementInstanceMetadata`

**Fix:**
Move everything we need into `ElementInstanceMetadata` and start using that instead.

**Commit Details:**

- Added `rootElements`, `componentName`, and `label` to `ElementInstanceMetadata` so that we can capture the information that is being captured by `ComponentMetadata`
- Updated the metadata capturing in `scene-root.tsx` to include those values, as well as capture other props from the scene (`resizesContent`, `style` and the UID)
- Added functions to `template-path.ts` for checking if an `InstancePath` is the path for the Storyboard's instance, or a descendant of it. Also added functions for converting a given `TemplatePath` to an `InstancePath` or a `ScenePath`, again underlining the fact that really we don't need the distinction anymore.
- Replaced the multitude of ways that we get the root paths we are interested in to only use the `ElementInstanceMetadata`, in `element-metadata-utils.ts`, and added tests to cover the new functions
- Added a helper function for determining if the instance metadata is for a scene or not, based on previous usage (checking that the element is `left('Scene')` unfortunately, but at least this means we can fix this horrible hack that we're using when capturing the metadata)
- Updated the metadata merging (spy + dom walker data) to include the new props added to the instance metadata
- Updated and fixed the tests

I'm pretty convinced after having done this that:

1. `rootElements` should just be merged back with the `children`, since we can distinguish between which is which based on the `TemplatePath`s of them
2. We should absolutely bin off the concept of a separate `ScenePath` and `InstancePath`, but not right now because that will be quite the undertaking :grimacing:

### DIFF
--- a/editor/src/components/canvas/__snapshots__/dom-walker.spec.browser.tsx.snap
+++ b/editor/src/components/canvas/__snapshots__/dom-walker.spec.browser.tsx.snap
@@ -4,35 +4,6 @@ exports[`DOM Walker tests Label carried through for generated elements 1`] = `
 Object {
   "components": Array [
     Object {
-      "component": "storyboard",
-      "globalFrame": null,
-      "label": "Storyboard",
-      "rootElements": Array [
-        Object {
-          "element": Array [
-            "utopia-storyboard-uid",
-          ],
-          "scene": Object {
-            "sceneElementPaths": Array [],
-            "type": "scenepath",
-          },
-        },
-      ],
-      "scenePath": Object {
-        "sceneElementPaths": Array [],
-        "type": "scenepath",
-      },
-      "sceneResizesContent": false,
-      "style": Object {},
-      "templatePath": Object {
-        "element": Array [],
-        "scene": Object {
-          "sceneElementPaths": Array [],
-          "type": "scenepath",
-        },
-      },
-    },
-    Object {
       "component": "App",
       "globalFrame": Object {
         "height": 812,
@@ -714,35 +685,6 @@ exports[`DOM Walker tests Label carried through for normal elements 1`] = `
 Object {
   "components": Array [
     Object {
-      "component": "storyboard",
-      "globalFrame": null,
-      "label": "Storyboard",
-      "rootElements": Array [
-        Object {
-          "element": Array [
-            "utopia-storyboard-uid",
-          ],
-          "scene": Object {
-            "sceneElementPaths": Array [],
-            "type": "scenepath",
-          },
-        },
-      ],
-      "scenePath": Object {
-        "sceneElementPaths": Array [],
-        "type": "scenepath",
-      },
-      "sceneResizesContent": false,
-      "style": Object {},
-      "templatePath": Object {
-        "element": Array [],
-        "scene": Object {
-          "sceneElementPaths": Array [],
-          "type": "scenepath",
-        },
-      },
-    },
-    Object {
       "component": "App",
       "globalFrame": Object {
         "height": 812,
@@ -1108,35 +1050,6 @@ Object {
 exports[`DOM Walker tests Simple Project with divs 1`] = `
 Object {
   "components": Array [
-    Object {
-      "component": "storyboard",
-      "globalFrame": null,
-      "label": "Storyboard",
-      "rootElements": Array [
-        Object {
-          "element": Array [
-            "utopia-storyboard-uid",
-          ],
-          "scene": Object {
-            "sceneElementPaths": Array [],
-            "type": "scenepath",
-          },
-        },
-      ],
-      "scenePath": Object {
-        "sceneElementPaths": Array [],
-        "type": "scenepath",
-      },
-      "sceneResizesContent": false,
-      "style": Object {},
-      "templatePath": Object {
-        "element": Array [],
-        "scene": Object {
-          "sceneElementPaths": Array [],
-          "type": "scenepath",
-        },
-      },
-    },
     Object {
       "component": "App",
       "globalFrame": Object {
@@ -1766,35 +1679,6 @@ exports[`DOM Walker tests Simple Project with flex parent 1`] = `
 Object {
   "components": Array [
     Object {
-      "component": "storyboard",
-      "globalFrame": null,
-      "label": "Storyboard",
-      "rootElements": Array [
-        Object {
-          "element": Array [
-            "utopia-storyboard-uid",
-          ],
-          "scene": Object {
-            "sceneElementPaths": Array [],
-            "type": "scenepath",
-          },
-        },
-      ],
-      "scenePath": Object {
-        "sceneElementPaths": Array [],
-        "type": "scenepath",
-      },
-      "sceneResizesContent": false,
-      "style": Object {},
-      "templatePath": Object {
-        "element": Array [],
-        "scene": Object {
-          "sceneElementPaths": Array [],
-          "type": "scenepath",
-        },
-      },
-    },
-    Object {
       "component": "App",
       "globalFrame": Object {
         "height": 812,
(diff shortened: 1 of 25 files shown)


editor/src/components/canvas/__snapshots__/dom-walker.spec.browser.tsx.snap
```diff
@@ -3,35 +3,6 @@
 exports[`DOM Walker tests Label carried through for generated elements 1`] = `
 Object {
   "components": Array [
-    Object {
```
balazsbajorics, line 6: I assume this is fine because no one relies on the fake storyboard scene anymore?

Rheeseyb: Yep! feelsgoodman.gif

editor/src/components/canvas/canvas.ts
```diff
@@ -137,21 +137,10 @@ const Canvas = {
...
-        metadata,
-        storyboardComponent.scenePath,
-      )
-      return storyboardEntryPoint.flatMap((storyboardRoot) => {
-        return recurseChildren({ x: 0, y: 0 } as CanvasVector, storyboardRoot).frames
-      })
-    } else {
-      return []
-    }
+    const storyboardEntryPoint = MetadataUtils.getAllStoryboardChildren(metadata)
```
balazsbajorics, line 140: this could be `canvasRoots` or `storyboardRoots` or `storyboardChildren`

editor/src/components/canvas/dom-walker.ts
```diff
@@ -408,9 +413,17 @@ function collectMetadata(
   )
 
   return pathsForElement.map((path) => {
-    const filteredChildPaths = unfilteredChildrenPaths.filter((childPath) =>
-      TP.pathsEqual(path, TP.parentPath(childPath)),
-    )
+    let filteredChildPaths: InstancePath[] = []
+    let filteredRootElements: InstancePath[] = []
```
balazsbajorics, line 417: hmmmmmm, hmmmmmmmmmmmmm, `MetadataUtils.getAllChildrenIncludingUnfurledFocusedComponents` now filters through allPaths to find `TP.scenePathsEqual(TP.scenePathPartOfTemplatePath(p), matchingFocusPath),` 

https://github.com/concrete-utopia/utopia/blob/e2e11c462a8ce61508a4b84d2e417db9008a8619/editor/src/core/model/element-metadata-utils.ts#L805

If I understand this correctly, instead of the (slightly expensive) array creationg and array filtering, it could use the rootElements property

editor/src/components/editor/actions/actions.ts
```diff
@@ -1064,7 +1064,7 @@ function deleteElements(targets: TemplatePath[], editor: EditorModel): EditorMod
 
       return element.children.every((childPath) => {
         const child = MetadataUtils.getElementByInstancePathMaybe(metadata.elements, childPath)
-        child == null || isElementToBeDeleted(child) || isEmptyOrContainsDeleted(child)
+        return child == null || isElementToBeDeleted(child) || isEmptyOrContainsDeleted(child)
```
balazsbajorics, line 1067: oh my god

Rheeseyb: Ha! I'll be adding that eslint rule :grinning:

editor/src/components/inspector/inspector.tsx
```diff
@@ -434,36 +434,23 @@ export const InspectorEntryPoint: React.FunctionComponent = betterReactMemo(
...
         return arrayEquals(oldTemplatePaths, newTemplatePaths, TP.pathsEqual)
       },
     )
 
-    const showSceneInspector =
-      selectedViews[0] != null &&
-      TP.depth(selectedViews[0]) === 1 &&
-      TP.isScenePath(selectedViews[0])
-
-    if (showSceneInspector) {
```
balazsbajorics, line 459: I think removing `const showSceneInspector` makes the code readability worse here

editor/src/components/canvas/dom-walker.ts
Rheeseyb: :thinking: that _sounds_ correct

editor/src/components/canvas/canvas.ts
Rheeseyb: `storyboardChildren` sounds right to me

Rheeseyb: (I'm going to roll this change into the next pass, since I'm working on that now)

editor/src/components/canvas/dom-walker.ts
balazsbajorics: interesting! this is actually more useful than I thought – however, I realized just now that you are kinda cheating with Scenes, the root element of a scene is a... div? and not really the "target component" of the scene – this is not a problem as next week we are going to change how a Scene works, but sneaky sneaky

editor/src/components/inspector/inspector.tsx
Rheeseyb: Fair. I'll re-add it

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
```diff
@@ -479,28 +526,33 @@ describe('Spy Wrapper Template Path Tests', () => {
...
           "children": Array [
             "storyboard/scene:app-root/inner-div/card-instance",
           ],
           "name": "div",
+          "rootElements": Array [],
         },
         "storyboard/scene:app-root/inner-div/card-instance": Object {
           "children": Array [],
           "name": "div",
+          "rootElements": Array [],
```
balazsbajorics, line 555: hmm, shouldn't this list `"storyboard/scene:app-root/inner-div/card-instance:button-instance"` as a rootElement?

Rheeseyb: It should eventually yes, but right now the dom walker isn't capturing that information because of the different paths for scenes vs elements, which this work will address.

Rheeseyb: Oh which maybe means that we can't replace the `getAllChildrenIncludingUnfurledFocusedComponents` function just yet

balazsbajorics: yeah

balazsbajorics (approved):

editor/src/components/editor/actions/actions.ts
seanparsons: Well this is terrifying.

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -532,21 +547,25 @@ export const MetadataUtils = {
...
+    return element?.rootElements ?? []
+  },
+  getImmediateChildrenPaths(
+    elements: ElementInstanceMetadataMap,
+    target: TemplatePath,
+  ): Array<InstancePath> {
+    const element = MetadataUtils.findElementByTemplatePath(elements, target)
+    const rootElements = element?.rootElements ?? []
+    const children = element?.children ?? []
+    return [...rootElements, ...children]
```
seanparsons, line 561: Surely return the merged arrays if `element` is not null and return an empty one otherwise?

editor/src/utils/deep-equality.ts
```diff
@@ -494,6 +494,106 @@ export function combine11EqualityCalls<A, B, C, D, E, F, G, H, I, J, K, X>(
   }
 }
 
+export function combine14EqualityCalls<A, B, C, D, E, F, G, H, I, J, K, L, M, N, X>(
```
seanparsons, line 497: OMG, what have I done?

seanparsons (commented):
+1

editor/src/core/model/element-metadata-utils.ts
Rheeseyb: Yeah that's fair :+1:

editor/src/utils/deep-equality.ts
Rheeseyb: This is all your doing!

PR #1037, opened 2021-04-01
fix(dom-walker) invalidate selected parent scene for focused components

enidemi:
**Problem:**
When selecting elements inside a focused component the inspector can be empty because only the focused element scenepath is invalidated on selection change, this causes the dom-walker to return the cached metadata for the scene, skips walking the focused element children.

**Commit Details:**
- adding the first scenepath to `invalidatedSceneIDsRef` instead of the full focused component path

### DIFF
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -265,7 +265,7 @@ function useInvalidateScenesWhenSelectedViewChanges(
     (store) => store.editor.selectedViews,
     (newSelectedViews) => {
       newSelectedViews.forEach((sv) => {
-        const scenePath = TP.scenePathPartOfTemplatePath(sv)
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):

Rheeseyb (commented):
Does this fix also apply to nested focused components? If not, we might need to change the approach here and check if a given path is a parent of one of the invalidated paths (possibly using `scenePathUpToElementPath`)

enidemi:
> Does this fix also apply to nested focused components? If not, we might need to change the approach here and check if a given path is a parent of one of the invalidated paths (possibly using `scenePathUpToElementPath`)

yes, it fixes this problem too because in dom-walker world an element is a scene only when it has the `data-utopia-scene-id`, focused components are not counted as scenes

Rheeseyb (approved):

Issue #1020, opened 2021-03-30
Investigate Missing UID on imported component's root element


balazsbajorics:
in the commit 06b6dee8 I removed this uid from the snapshot for antd's DatePicker, because it is already missing from the running editor on master. but this _is_ a regression that should be investigated.

PR #1038, opened 2021-04-01
Dependency file graph handling for modules.

seanparsons:
**Problem:**
Previously if a file `b.js` was changed and a file `a.js` which depended on it were not, the result of executing the module `a.js` would persist which potentially resulted in some stale data generated from the old `b.js` code sticking around. This might mean that weird behaviour would present to the user like an old component rendering for instance.

**Fix:**
Previously the execution result of each project file was held against the file itself in `ESCodeFile`, now those are in a single shared object which can be easily updated with the new reverse transitive search applied on top of just updating those files which have changed.

In the process of doing this, the multi-file editing was also implemented so that non-storyboard files can be edited and the canvas will update with the results of those changes. As the worker responsible for parsing and printing now handles potentially the entire project in one go, as opposed to just the storyboard file.

There's also a bonus fix of breaking a dependency between `ts-worker.ts` and `tippy` which caused some workers to die on initialisation.

**Commit Details:**
- Added a dependency cruiser rule that should check for the ts worker not ending up with a transitive dependency on tippy.
- Deleted `addUtopiaUtilsImportIfUsed` as that breaks the chain from the ts worker to tippy, along with the usage of it in `actions.ts`.
- Implemented `zipContentsTree` as a non promise based version of `zipContentsTreeAsync`.
- Moved `SampleNodeModules` to `code-file.test-utils.ts`.
- Added `evaluationCache` to `CodeResultCache` as a shared location for the result of executing modules.
- `generateCodeResultCache` now does a reverse transitive dependency search for the files that depend on those which have changed so that they may be removed from `evaluationCache`.
- `UpdateFromWorker` action now takes multiple updates instead of handling just one.
- Reworked the `UPDATE_FROM_WORKER` action to handle multiple updates.
- `maybeRequestModelUpdate` sends potentially many parse or print file requests in one message to the worker.
- `ESCodeFile` now does not include `evalResultCache`.
- `getRequireFn` now takes `EvaluationCache` as a parameter.
- `clearNodeModules` removed as it is not necessary anymore due to the changes to `ESCodeFile`.
- Implemented `getDirectReverseDependencies` to determine what files depend on a given file.
- Implemented `getTransitiveReverseDependencies` to chain up through the reverse dependencies of a given file to find everything that is transitively dependent on some files.
- Implemented `identifyFilesThatHaveChanged` for comparing two instances of `ProjectContentTreeRoot`.
- Moved `getSavedCodeFromTextFile` and `getUnsavedCodeFromTextFile` out of `vscode-bridge.ts`.
- Combined the `sendParseFileMessage` and `sendPrintCodeMessage` worker services into the `sendParsePrintMessage` service.


editor/src/core/shared/project-contents-dependencies.ts
```diff
@@ -0,0 +1,170 @@
...
+  oldContents: ProjectContentTreeRoot,
+  newContents: ProjectContentTreeRoot,
+): Set<string> {
+  let filesWithChanges: Set<string> = emptySet()
+
+  function addFileToChanges(filename: string): void {
+    filesWithChanges.add(filename)
+  }
+
+  function applyChanges(
```
Rheeseyb, line 94: It looks like this function was extracted from `vscode-bridge.ts` but still exists there? Should `vscode-bridge` now be using this instead?

Rheeseyb (approved):
Nice one!

editor/src/core/shared/project-contents-dependencies.ts
seanparsons: It was duplicated from there but is different internally, it would be possible to make a version that identifies _all_ changes as something like Added/Removed/Modified/Etc and then process those, but I was a bit dubious of abstracting that with only 2 uses.

editor/src/core/es-modules/package-manager/package-manager.ts
```diff
@@ -91,11 +105,11 @@ export function getRequireFn(
             // we should extend the module objects so it not only contains the exports,
             // to have feature parity with the popular bundlers (Parcel / webpack)
             // MUTATION
-            resolvedFile.evalResultCache = { module: partialModule }
+            //  resolvedFile.evalResultCache = { module: partialModule }
```
Rheeseyb, line 108: Does this (and the commented line below) need to go in the bin or was this a temporary commenting?

seanparsons: Yes, good spot.

seanparsons: Fixed in 1e264e2b.

editor/src/components/assets.ts
```diff
@@ -248,6 +248,37 @@ export async function walkContentsTreeAsync(
...
+  firstTree: ProjectContentTreeRoot,
+  secondTree: ProjectContentTreeRoot,
+  onElement: (
+    fullPath: string,
+    firstContents: ProjectContentsTree | null,
+    secondContents: ProjectContentsTree | null,
+  ) => boolean,
+): void {
+  let keys: Set<string> = emptySet()
+  Object.keys(firstTree).forEach(keys.add, keys)
```
balazsbajorics, line 261: this syntax is new to me – what happens if you don't include the second parameter `keys` here?

seanparsons: From memory, it has all kinds of issues around `this`, because of reasons.

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
```diff
@@ -1308,6 +1308,7 @@ export var storyboard = (
             <div
               class=\\"ant-picker\\"
               style=\\"width: 123px; height: 51px; left: 113px; top: 395px;\\"
+              data-uid=\\"antd-date-picker\\"
```
balazsbajorics, line 1311: most interesting – do you know how did this start working on this branch?

seanparsons: I didn't look into this very far at all mostly because it seemed like a step up from what we had.

balazsbajorics: you very likely fixed https://github.com/concrete-utopia/utopia/issues/1020

seanparsons: I assumed it was never in there in the first place, so that's actually even better news.

editor/src/core/shared/project-contents-dependencies.ts
```diff
@@ -0,0 +1,170 @@
...
+              throw new Error(`Unhandled case ${JSON.stringify(resolveResult)}`)
+          }
+        }
+      }
+    }
+  })
+  return result
+}
+
+export function getTransitiveReverseDependencies(
```
balazsbajorics, line 60: could you also make a test for this function?

balazsbajorics (approved):

editor/src/core/shared/project-contents-dependencies.ts
seanparsons: Noted, will add some subsequently.

PR #1039, opened 2021-04-01
Complete Removal of ComponentMetadata

Rheeseyb:
**Problem:**
We no longer want a separate array of `ComponentMetadata` for scene only metadata.

**Fix:**
Remove it completely, since now everything is contained in the  `ElementInstanceMetadata`. As a result, we can completely remove `JSXMetadata`, and replace it everywhere with `ElementInstanceMetadataMap`, which I have done, but am submitting in a separate PR.


editor/src/core/model/element-metadata-utils.spec.ts
```diff
@@ -490,26 +470,26 @@ describe('getElementLabel', () => {
...
-        height: 100,
-      }),
-      style: {
-        width: 100,
-        height: 100,
-      },
-    },
-  ]
-  const metadata = jsxMetadata(components, elements)
+  // const components: Array<ComponentMetadata> = [
```
seanparsons, line 473: Delete me?

editor/src/core/shared/template-path.ts
```diff
@@ -441,23 +441,32 @@ export function parentPath(path: InstancePath): TemplatePath
 export function parentPath(path: TemplatePath): TemplatePath | null
 export function parentPath(path: TemplatePath): TemplatePath | null {
   if (isScenePath(path)) {
-    return null
+    return null // FIXME This is wrong!
```
seanparsons, line 444: But fine for now?

seanparsons (approved):
So much simpler.

enidemi (approved):

editor/src/core/shared/template-path.ts
Rheeseyb: Ah yes, the problem here is that since scene paths are also element paths, they can still have parents. I was considering handling this during this work but decided against since that would mean the PR would be addressing separate, unrelated issues, and especially since this work has absolutely highlighted that we should remove the concept of `ScenePath` and `InstancePath`

editor/src/core/model/element-metadata-utils.spec.ts
Rheeseyb: Oops :grinning:

editor/src/core/model/scene-utils.ts
```diff
@@ -303,10 +303,10 @@ export function isSceneChildWidthHeightPercentage(
 }
 
 export function isDynamicSceneChildWidthHeightPercentage(
-  scene: ComponentMetadata,
+  scene: ElementInstanceMetadata,
   metadata: JSXMetadata,
 ): boolean {
-  const isDynamicScene = scene.sceneResizesContent
+  const isDynamicScene = scene.props.resizesContent ?? false
```
balazsbajorics, line 309: I just realized `PathForResizeContent` is `resizeContent` and not `resizesContent`.

I think this "fake prop" should also use the `PathForResizeContent` const.

editor/src/core/model/element-metadata-utils.spec.ts
Rheeseyb: Oh yeah, this was deleted in a later commit, but that commit was moved into #1040, so this has been deleted :)

editor/src/core/shared/template-path.ts
```diff
@@ -441,23 +441,32 @@ export function parentPath(path: InstancePath): TemplatePath
...
   } else {
     return instancePathParent(path)
   }
 }
 
 function elementPathToUID(path: ElementPath): id {
-  return last(path)!
+  return last(path) ?? ''
```
balazsbajorics, line 451: hmmm. I don't like the `''` here because it can lead to more silent errors. at least a surprise `undefined` throws a runtime error, helping us find the problem. but a `'' === ''` comparison can survive in the codebase for years

balazsbajorics (approved):

editor/src/core/model/scene-utils.ts
Rheeseyb: Good catch. Fixing now :+1:

editor/src/core/shared/template-path.ts
Rheeseyb: I _think_ this will still be a problem whilst we have the empty path used in some places for the storyboard root. I'll revert this and check nothing breaks, then make a note to remove all use of the empty path for the storyboard root.

balazsbajorics: or maybe you could throw an error here if the path is empty?

Rheeseyb: :+1:

PR #1046, opened 2021-04-06
Fix: cmd-click to select views

balazsbajorics:
**Problem:**
There were two problems with selection on canvas, this PR fixes both:
1. cmd-click to select was not at all fixed, I somehow tricked my brain into belieiving I fixed it
2. double click to select was flawed, it was adding double offsets to focused component root elements. I didn't find it because I was not testing it with absolute positioned elements (my bad)

**Commit Details:**
- Created a `getAllPathsIncludingUnfurledFocusedComponents` similar to `getAllPaths`. This fixed cmd - click to select
- split the return value of `getAllChildrenIncludingUnfurledFocusedComponents` to `children` and `unfurledComponents` so `Canvas.getFramesInCanvasContext` can apply different offset to these by type


balazsbajorics:
there's one more bug with generated elements, I made an entry for it in the paper doc. I will, however, wait until Rheese merges his PR before attempting to fix that

Rheeseyb (approved):

enidemi (approved):

maltenuhn:
Introducing the New CMD Select. The same functionality you've always trusted, now with 100% more selection power. ~Call now to get a free click-to-zoom~

PR #1047, opened 2021-04-07
Render scene children

### LINKED ISSUE #1048: Render and capture metadata of children of scenes
In preparation for changing `<Scene component={App} />` to `<Scene><App /><Scene>`, we first need to actually render the children of the scene, and also capture their metadata.

Rheeseyb:
Fixes #1048 

**Problem:**
In order to support `<Scene><App /></Scene>` we first need to render scene children.

**Fix:**
Render the scene children using `renderCoreElement`. Since right now the scene children will be the children of the component being rendered by the scene, for the sake of separating this piece of work out from the later work, right now we will pass the rendered children to the component being rendered (as opposed to rendering them as children of the scene, which is what we will be doing in the next iteration).

Also included in this work is a change that includes all valid paths on the storyboard (treating children of the scene as being automatically focused), and then passes them down through the scene (previously that connection wasn't there). That doesn't include the special casing that would be required for fetching the valid paths of the `sceneProps.component`, as this is in preparation for removing that anyway (I've left the existing logic for using the `sceneProps.component` to add extra valid paths to the rendered scene element).

This work doesn't make the children of the scene selectable, and doesn't result in them being included in the navigator. It just means that we will now render them (as children of the component rendered by the scene), and capture their metadata. It should be purely considered to be prep work for the removal of the `component` prop of the `Scene`.


editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
```diff
@@ -197,7 +248,7 @@ export const SceneRootRenderer = betterReactMemo(
             mutableUtopiaContext.jsxFactoryFunctionName,
             sceneProps.component,
             propsWithScenePath,
-            undefined,
+            renderedChildren,
```
Rheeseyb, line 251: In hindsight there was absolutely no point to adding the `renderedChildren` as children of the `rootElement` here. I'm immediately changing that in the next piece of work (that I've started), and instead in the return statement below, I'm replacing line 271:
```
{rootElement}
```
with:
```
<>
  {rootElement}
  {renderedChildren}
</>
```
I can't add that comment on that line because it's not included in the git diff's context.
That change will come with the next PR later today.

seanparsons (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
```diff
@@ -152,6 +149,7 @@ function useGetValidTemplatePaths(
 interface SceneRootRendererProps {
   sceneElement: JSXElement
   filePath: string
+  validPaths: Array<InstancePath>
```
balazsbajorics, line 152: is this a preparation to consolidating the data-utopia-valid-paths to the canvas root element, and removing it from the scene element in the DOM?

Rheeseyb: Yep. We were always adding them there but never passing them into the scene, so the scene would wipe them.

balazsbajorics (approved):

PR #1052, opened 2021-04-07
Fix: dom walker focused generated elements

balazsbajorics:
**Problem:**
If there is an array of generated elements (say 6 Card elements), no matter which element was focused, the last one would always override the metadata for the real focused one. This prevented selection on the canvas because the hit target for the selection would be the last element (and not the focused element). 

**Fix:**
In the dom-walker, when walking a generated UID, filter out the valid paths who's scene path points to the same uid but with a different generated id.

**Commit Details:**
- Refactored dom-walker so valid paths is an `Array<InstancePath>` instead of `Array<string>`
- Added `filterValidFocusedPathsForGeneratedElement` to dom-walker which removes valid paths with non-matching indexed dynamic UIDs
- Use `filterValidFocusedPathsForGeneratedElement` in `walkElements`


Rheeseyb (approved):

balazsbajorics:
I need one more review :)

enidemi (approved):
looks good, please add tests for `filterValidFocusedPathsForGeneratedElement`

balazsbajorics:
> looks good, please add tests for `filterValidFocusedPathsForGeneratedElement`

good shout. I'll add tests in a follow-up pr, merging this to fix master asap

PR #1053, opened 2021-04-08
Multi-file insertion.

seanparsons:
**Problem:**
Insertion has no concept that the result of insertion might be that the change will go into something other than what is shown on the canvas.

**Fix:**
As changes potentially to any file in the entire project those changes are now tracked against the files they need to change. The new infrastructural code around modifying the underlying target of a change was implemented into multiple place that didn't currently handle that.

**Limitations:**
- The canvas is potentially presenting too many targets which results in one level of nesting being applied (changes going into `/src/app.js` instead of `/src/card.js` for example in the dev project). This appears to be an issue with focusing that should be fixed separately.

**Commit Details:**
- Applying the transient state changes in `clearDragState` now handles
  the changes potentially hitting multiple files.
- Functions for handling move and resize transient canvas state shimmed to
  work with the multiple file handling transient state.
- `produceCanvasTransientState` refactored to transform the underlying
  target for insertion.
- Canvas components use `TransientFilesState` instead of `TransientFileState`.
- `normalisePathToUnderlyingTarget` supports returning a null
  element path to allow for root insertion.
- Added storyboard components to some test data, as they previously didn't
  have one and relied on faked metadata which did include data for
  the storyboard component.
- `INSERT_JSX_ELEMENT` action uses `modifyUnderlyingTarget` to perform the
  insertion.
- `modifyOpenJsxElementAtPath` and `modifyOpenJsxElementAtStaticPath` also
  now utilise `modifyUnderlyingTarget`.
- `TransientCanvasState` now holds a `TransientFilesState` instance which
  holds a `TransientFileState` for each file that is changing.
- Widened the type of the `target` parameter for `modifyUnderlyingTarget`.
- `navigatorItemWrapperSelectorFactory` reworked to lookup the underlying
  element so that the correct transient file state can be obtained.
- Added `Scene` to `sampleIncludedElementTypes` and fixed the mountain of
  snapshots and tests which used that.
- `dropFirstScenePathElement` now allows for a null `path` parameter.
- Fixed `createFakeMetadataForComponents` such that fake values are inserted
  for anything in the `definedElsewhere` value and removed the `parentProps`
  parameter which was never used.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-top-level-elements.ts
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-top-level-elements.ts
@@ -2,7 +2,7 @@ import { useContextSelector } from 'use-context-selector'
 import { ArbitraryJSBlock, TopLevelElement } from '../../../core/shared/element-template'
 import { Imports, isParseSuccess, isTextFile } from '../../../core/shared/project-file-types'
 import { getContentsTreeFileFromString, ProjectContentTreeRoot } from '../../assets'
-import { TransientFileState } from '../../editor/store/editor-state'
+import { TransientFilesState, TransientFileState } from '../../editor/store/editor-state'
 import { UtopiaProjectContext } from './ui-jsx-canvas-contexts'
 
 interface GetParseSuccessOrTransientResult {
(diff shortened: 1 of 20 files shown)


editor/src/components/editor/actions/actions.ts
```diff
@@ -1944,41 +1945,49 @@ export const UPDATE_FNS = {
...
-    const withNewElement = modifyOpenJSXElements((elements) => {
-      const targetParent =
-        action.parent == null
-          ? // action.parent == null means Canvas, which means storyboard root element
-            getStoryboardTemplatePath(elements)
-          : action.parent
+    let newSelectedViews: TemplatePath[] = []
+    const withNewElement = modifyUnderlyingTarget(
+      action.parent,
+      forceNotNull('Should originate from a designer', editor.canvas.openFile?.filename),
```
enidemi, line 1951: is this inserting to the currently open file?

seanparsons: No, it's starting the search from the currently open canvas file.

enidemi (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-top-level-elements.ts
```diff
@@ -22,26 +22,28 @@ const EmptyResult: GetParseSuccessOrTransientResult = {
...
-  openStoryboardFileNameKILLME: string | null,
-  transientFileState: TransientFileState | null,
+  transientFilesState: TransientFilesState | null,
 ): GetParseSuccessOrTransientResult {
   const projectFile = getContentsTreeFileFromString(projectContents, filePath)
   if (isTextFile(projectFile) && isParseSuccess(projectFile.fileContents.parsed)) {
     const parseSuccess = projectFile.fileContents.parsed
-    if (openStoryboardFileNameKILLME === filePath && transientFileState != null) {
+    const transientFileState: TransientFileState | null =
+      transientFilesState == null ? null : transientFilesState[filePath] ?? null
```
balazsbajorics, line 31: I think this reads better:

```suggestion
      transientFilesState?.[filePath] ?? null
```

seanparsons: I find that operator really weird.

balazsbajorics: it _is_ a bit weird at first, but once I got used to it, it reads so much better than the conditional operator!

seanparsons: You say that "at first", the sentence doesn't really need it. :)

balazsbajorics (approved):

Rheeseyb (approved):

PR #1054, opened 2021-04-08
fixing the navigator showing unfocused targets as focused

balazsbajorics:
**Problem:**
image
The Navigator would unfurl non-focused instances of a focused component. In the screenshot, LabelRow inside Scene 2 should not be unfurled.

**Fix:**
The problem is that the logic to determine if something is along the focused path (ie a _prefix_ of the focused path) was using `TP.scenePathUpToElementPath` which actually only finds the last elementPath's index inside the focused path, and returns the subpath up to that index. Which means it will match for all template paths with _any_ scene paths so long as the elementPath is in there. We accidentally used this function to also determine if something is along the focused path, but that is simply wrong for this purpose. 
I now wonder if the final usage of `TP.scenePathUpToElementPath` is also wrong and should instead use `longestSharedElementPathArray` 

**Commit Details:**
- Introduced `TP.isPathAlongFocusedElementPath` along with a handful of helper functions that treat a TemplatePath as an `Array<ElementPath>`
- Using `TP. isPathAlongFocusedElementPath` in `getAllChildrenIncludingUnfurledFocusedComponents` 
- Using `TP.isPathAlongFocusedElementPath` in `navigator-item@isFocused`

### DIFF
--- a/editor/src/core/shared/template-path.ts
+++ b/editor/src/core/shared/template-path.ts
@@ -1040,3 +1040,52 @@ export function outermostScenePathPart(path: TemplatePath): ScenePath {
     return asScenePath
   }
 }
+
+function toElementPathArray(path: TemplatePath | null): Array<ElementPath> | null {
+  if (path == null) {
+    return null
+  } else if (isScenePath(path)) {
+    return path.sceneElementPaths
+  } else {
+    return [...path.scene.sceneElementPaths, path.element]
+  }
+}
+
+function longestSharedElementPathArray(
+  l: Array<ElementPath> | null,
+  r: Array<ElementPath> | null,
+): Array<ElementPath> | null {
+  if (l === null || r === null) {
+    return null
+  } else if (l === r) {
+    return l
+  } else {
+    const matchedElements = longestCommonArray(l, r, elementPathsEqual)
+    if (matchedElements.length > 0) {
+      return matchedElements
+    } else {
+      return null
+    }
+  }
+}
+
+function isElementPathPrefixOfElementPath(
+  potentialPrefix: Array<ElementPath> | null,
+  target: Array<ElementPath> | null,
+): boolean {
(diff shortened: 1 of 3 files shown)


enidemi (approved):

editor/src/core/shared/template-path.ts
```diff
@@ -1040,3 +1040,52 @@ export function outermostScenePathPart(path: TemplatePath): ScenePath {
...
+    }
+  }
+}
+
+function isElementPathPrefixOfElementPath(
+  potentialPrefix: Array<ElementPath> | null,
+  target: Array<ElementPath> | null,
+): boolean {
+  const longestSharedArray = longestSharedElementPathArray(potentialPrefix, target)
+  return potentialPrefix != null && longestSharedArray?.length === potentialPrefix.length
```
seanparsons, line 1077: This logic makes me feel slightly queasy, where we avoid the double null case by checking if one side is null and then optional chaining the other.

editor/src/core/shared/template-path.ts
```diff
@@ -1040,3 +1040,52 @@ export function outermostScenePathPart(path: TemplatePath): ScenePath {
...
+}
+
+export function isPathAlongFocusedElementPath(
+  path: TemplatePath,
+  focusedElementPath: ScenePath | null,
+): boolean {
+  const elementPathArrayForTarget = toElementPathArray(path) ?? []
```
seanparsons, line 1084: Should this have the empty array default?

seanparsons (approved):

editor/src/core/shared/template-path.ts
balazsbajorics: probably I should improve the type of `toElementPathArray` to know that not-null input leads to not-null output

balazsbajorics:
I'm closing this in favor of #1059

PR #1055, opened 2021-04-08
feature(canvas) scroll animation on canvas

enidemi:
**Problem:**
There are no animations on the canvas.

**Fix:**
Adding animation for the scroll to element feature. While the canvas animations are enabled the canvas controls are not visible.

**Commit Details:**
- new action, nem variable in editorstate
- hide canvas controls while animations are enabled
- add css `transition` for transform property while animations are enabled to the canvas-container

### DIFF
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -4207,6 +4207,9 @@ export const UPDATE_FNS = {
     dispatch: EditorDispatch,
   ): EditorModel => {
     if (action.value) {
+      if (canvasScrollAnimationTimer != null) {
+        clearTimeout(canvasScrollAnimationTimer)
+      }
       canvasScrollAnimationTimer = window.setTimeout(() => {
         clearTimeout(canvasScrollAnimationTimer)
         canvasScrollAnimationTimer = undefined
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -356,6 +356,7 @@ import {
   SendCodeEditorInitialisation,
   SetFocusedElement,
   AddImports,
+  SetScrollAnimation,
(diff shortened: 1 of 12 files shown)


editor/src/components/editor/actions/actions.ts
```diff
@@ -4197,6 +4201,26 @@ export const UPDATE_FNS = {
       },
     }
   },
+  SET_SCROLL_ANIMATION: (
+    action: SetScrollAnimation,
+    editor: EditorModel,
+    dispatch: EditorDispatch,
+  ): EditorModel => {
+    if (action.value === true) {
```
seanparsons, line 4209: `=== true`?

seanparsons (approved):

editor/src/components/editor/actions/actions.ts
```diff
@@ -4197,6 +4201,26 @@ export const UPDATE_FNS = {
       },
     }
   },
+  SET_SCROLL_ANIMATION: (
+    action: SetScrollAnimation,
+    editor: EditorModel,
+    dispatch: EditorDispatch,
+  ): EditorModel => {
+    if (action.value) {
+      canvasScrollAnimationTimer = window.setTimeout(() => {
```
balazsbajorics, line 4210: it looks like you are missing a line that checks if `canvasScrollAnimationTimer != null` then clearTimeout:

```suggestion
      if (canvasScrollAnimationTimer != null) {
        clearTimeout(canvasScrollAnimationTimer)
      }
      canvasScrollAnimationTimer = window.setTimeout(() => {
```

because if multiple `setScrollAnimation(true)` are dispatched in short order, I think we want to extend the time the value stays true.

enidemi: fixed

balazsbajorics (approved):

PR #1058, opened 2021-04-08
Include root element paths of focused elements in the metadata

### LINKED ISSUE #1049: Treat top level children of any Scene element as being "focused"
In order to replace `<Scene component={App]/>` with `<Scene><App /></Scene>` we need to treat the top level children of any `Scene` element as though it is always a focused element. This means that for each of those children's root elements (and descendants of those root elements) we would want to

 - [x] include them in the valid paths
 - [x] spy wrap them to capture their render time metadata
 - [x] walk them in the DOM walker to capture their render result metadata
 - [x] include them in the list of selectable paths
 - [x] include them in the Navigator

Rheeseyb:
~Partial work towards~ Now fixes #1049 (after merging the follow up PR into this)
Also in turn fixes the same problem that was tackled in #1054 

**Problem:**
In order to fully support children of scenes as being focused elements, we need to capture the root elements of all "focused" elements in their metadata (in the same way that we were doing so for scenes after merging the two types of metadata).

**Fix:**
First I switched to rendering the scene children as children of the actual scene element rather than it's rendered component. After that I used the multiple paths for a given element to determine which paths were descendants of others (in the sense that one is a path to an instance and the other is a path to its root view) in order to capture the root element paths (of valid paths only) during the dom walking.

Since we now capture that information during the dom walking, I have replaced the contents of the two `IncludingUnfurledFocusedComponents` functions to use the captured root element paths rather than having to perform more lookups using the focused element path.

**Commit Details:**
- Updated `scene-root.tsx` to render its own children (rather than passing them to the component it renders)
- Added helper functions to `template-path.ts` for a better `isParentOf` and `isDescendentOf` check
- Use the multiple (valid) paths for a single element in `dom-walker.ts` to capture the root element paths in the metadata
- Updated `element-metadata-utils.ts` to use the captured root element paths in `getAllChildrenIncludingUnfurledFocusedComponents` and `getAllChildrenElementsIncludingUnfurledFocusedComponents`

**Note:** This means that in the case of:
```
    <Scene
      component={null}
      props={{}}
      style={{ position: 'absolute', left: 0, top: 0, width: 375, height: 812 }}
    >
      <App />
    </Scene>
```
~The elements from `App`'s render function are selectable on the canvas, but aren't yet included in the Navigator. I'm going to tackle that problem in the next step.~

Also, it's important to mention that this work exposed a flaw in our mapping from an element to its path in the DOM walker, since we are only using the element's UID rather than its full path. After having undertaken this work, I am fully convinced that the **only** safe way to handle this is to include all of the template paths for an element in a `data-template-paths` (rather than `data-uid`) attribute on the final dom element.


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -812,27 +836,10 @@ export const MetadataUtils = {
...
-        : allPaths.filter(
-            (p) =>
-              TP.depth(p) === 2 && // TODO this is actually pretty silly, TP.depth returns depth + 1 for legacy reasons
-              TP.scenePathsEqual(TP.scenePathPartOfTemplatePath(p), matchingFocusPath),
-          )
-
-    return { children: children, unfurledComponents: focusedRootElementPaths }
+    return {
+      children: MetadataUtils.getChildrenPaths(metadata, path),
+      unfurledComponents: MetadataUtils.getRootViewPaths(metadata, path),
```
balazsbajorics, line 841: 👍

editor/src/core/shared/template-path.ts
```diff
@@ -295,8 +295,19 @@ export function instancePathForElementAtScenePath(path: ScenePath): StaticInstan
   }
 }
 
+function instancePathForElementPaths(elementPaths: ElementPath[]): InstancePath {
+  const lastElementPath = last(elementPaths)
```
balazsbajorics, line 299: this function should share logic with `fromStringUncached`!

fromStringUncached should be the same as `instancePathForElementPaths(string.split(':').map(elementPath => elementPath.split('/'))`

toString should just be `instancePathToElementPaths().join('/').join(':')`

editor/src/core/shared/template-path.ts
```diff
@@ -647,6 +669,38 @@ export function isAncestorOf(
...
 
+function fullElementPathForPath(path: TemplatePath): ElementPath[] {
+  if (isScenePath(path)) {
+    return path.sceneElementPaths
+  } else {
+    return [...path.scene.sceneElementPaths, path.element]
+  }
+}
+
+export function isDescendentOf(
```
balazsbajorics, line 680: nitpick, but I think `descendant` is the more commonly used one

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
```diff
@@ -257,30 +257,11 @@ Object {
       "data-uid": "aaa scene-aaa",
       "skipDeepFreeze": true,
       "style": Object {
-        "alignContent": undefined,
```
balazsbajorics, line 260: where did these undefineds go? it _looks like_ this style object is only populated from the spy and not the dom walker

editor/src/core/shared/template-path.ts
Rheeseyb: Yep, agreed. But I think that would be better placed to happen alongside the removal of `ScenePath` and `InstancePath` otherwise it risks introducing regressions elsewhere.

editor/src/core/shared/template-path.ts
Rheeseyb: Oh ffs, I even said this on a call the other day to you! Yep will fix :facepalm:

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
Rheeseyb: :thinking:

editor/src/core/shared/template-path.ts
balazsbajorics: I didn't run the code, but reading it I think the logic doesn't match fromString now, if the elementPaths is a single array long, that should become a ScenePath and no instance part

balazsbajorics: in that case, I would make it return emptyinstance path if the elementPaths is 0 or 1 long!

Rheeseyb: No, because that's not what this is for. This function is for returning the InstancePath for a given ElementPath array, to help bridge the gap until we remove InstancePath and ScenePath.

editor/src/core/shared/template-path.ts
Rheeseyb: Fixed!

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
Rheeseyb: Fixed by the removal of the fragment in the Scene Root, but oh my do we need to remove that pragma now :grimacing:

editor/src/core/shared/template-path.ts
```diff
@@ -647,6 +669,38 @@ export function isAncestorOf(
...
 
+function fullElementPathForPath(path: TemplatePath): ElementPath[] {
+  if (isScenePath(path)) {
+    return path.sceneElementPaths
+  } else {
+    return [...path.scene.sceneElementPaths, path.element]
+  }
+}
+
+export function isDescendantOf(
```
enidemi, line 680: there is an `isAncestorOf` in the template-path.ts, should that be replaced with this one?

seanparsons:
There's a strikethrough part in the description which doesn't appear to have a replacement text?

Rheeseyb:
> There's a strikethrough part in the description which doesn't appear to have a replacement text?

That was tackled in #1059 which was branched off of this, and has since been merged into it.

enidemi (approved):

editor/src/core/model/element-metadata-utils.spec.ts
```diff
@@ -527,13 +595,79 @@ describe('getting the root paths', () => {
...
+      testComponentSceneChildElementRoot.templatePath,
+      testComponentSceneChildElementRootChild.templatePath,
       testStoryboardChildElement.templatePath,
       testStoryboardGrandChildElement.templatePath,
     ]
     expect(actualResult).toEqual(expectedResult)
   })
 })
 
+describe('createOrderedTemplatePathsFromElements returns all of the ordered navigator targets, visible and not', () => {
```
enidemi, line 608: 👍

balazsbajorics (approved):

PR #1072, opened 2021-04-12
Print template paths

balazsbajorics:
**Problem:**
Only UIDs are printed to the DOM, and we use the validPaths array to backsolve which template path might belong to which UID. There's a lot of edge cases that need dancing around. In the meantime, we realized that the old limitations that prevented printing the full TemplatePath don't exist anymore. 

**Fix:**
Use template paths in the DOM

**Commit Details:**
- Print the full template paths to the DOM, 
- Apply the same monkey patching to the printed template paths that we apply to UIDs
- use those in the dom-walker instead of backsolving based on uids
- make templatePath non-optional in the element-renderer functions (it didn't need to be optional anymore anyways)
- update a bunch of test snapshots


editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -88,6 +88,9 @@ import { JSX_CANVAS_LOOKUP_FUNCTION_NAME } from '../../core/workers/parser-print
 import { getParseSuccessOrTransientForFilePath } from './ui-jsx-canvas-renderer/ui-jsx-canvas-top-level-elements'
 import { ProjectContentTreeRoot, getContentsTreeFileFromString } from '../assets'
 import { createExecutionScope } from './ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope'
+import { applyUIDMonkeyPatch } from '../../utils/canvas-react-utils'
+
+applyUIDMonkeyPatch()
```
enidemi, line 93: why is this changed?

enidemi (approved):
I like it!

editor/src/components/canvas/__snapshots__/dom-walker.spec.browser.tsx.snap
```diff
@@ -262,6 +263,7 @@ Object {
       "y": 0,
     },
     "props": Object {
+      "data-paths": "utopia-storyboard-uid/scene-aaa:aaa",
```
Rheeseyb, line 266: Hmmm a lot of these seem wrong. Where they should have two paths (according to the `data-uid` prop below) they only have one.

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
```diff
@@ -1234,9 +1240,14 @@ export var ${BakedInStoryboardVariableName} = (props) => {
...
             height: 400px;
           \\"
+          data-paths=\\":utopia-storyboard-uid\\"
           data-uid=\\"utopia-storyboard-uid\\"
         >
-          <div id=\\"nasty-div\\" data-uid=\\"2a7~~~1 150~~~2 2f5~~~1 aaa scene-aaa\\">
+          <div
+            id=\\"nasty-div\\"
+            data-uid=\\"2a7~~~1 150~~~2 2f5~~~1 aaa scene-aaa\\"
+            data-paths=\\":150~~~2/2a7~~~1 :150~~~2 :2f5~~~1 utopia-storyboard-uid/scene-aaa:aaa\\"
```
Rheeseyb, line 1249: Hmmm yeah something is definitely wrong here

editor/src/components/canvas/__snapshots__/dom-walker.spec.browser.tsx.snap
balazsbajorics: I pushed a fix!

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
```diff
@@ -21572,21 +22062,30 @@ exports[`UiJsxCanvas render renders fragments correctly 1`] = `
...
         style=\\"width: 100px; height: 100px;\\"
         data-uid=\\"bbb\\"
+        data-paths=\\"eee/fff:aaa/25d80840-fbda-4984-be3b-43838cb78079\\"
       >
         Hello
         <div
           data-label=\\"some-other-div\\"
           style=\\"width: 100px; height: 100px;\\"
           data-uid=\\"ccc\\"
+          data-paths=\\"eee/fff:aaa/25d80840-fbda-4984-be3b-43838cb78079/8e84c2c0-99bd-40dc-bc63-4d8230425414\\"
```
balazsbajorics, line 22079: hmmmmmmmmm, what turns `ccc` into `25d80840-fbda-4984-be3b-43838cb78079` ? 😂 

at least this uid is truly random so this broke the test, helping me find it...

editor/src/components/canvas/__snapshots__/dom-walker.spec.browser.tsx.snap
```diff
@@ -38,6 +38,7 @@ Object {
       "y": 0,
     },
     "props": Object {
+      "data-paths": ":utopia-storyboard-uid",
```
Rheeseyb, line 41: I can't wait until we can remove the leading `:`

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
```diff
@@ -4226,11 +4347,25 @@ exports[`UiJsxCanvas render class component is available from arbitrary block in
...
+      <div
+        data-uid=\\"ccc aaa\\"
+        data-paths=\\"utopia-storyboard-uid/scene-aaa:zzz/aaa\\"
+      >
+        Thing
+      </div>
+      <div
+        data-uid=\\"ccc bbb\\"
+        data-paths=\\"utopia-storyboard-uid/scene-aaa:zzz/bbb\\"
+      >
```
Rheeseyb, line 4366: Looks like something has gone wrong here, because both of these elements should have 2 paths - both are missing a path ending in `:ccc`

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
```diff
@@ -15598,9 +15954,15 @@ exports[`UiJsxCanvas render refs are handled and triggered correctly in a class
...
+    data-paths=\\":utopia-storyboard-uid\\"
     data-uid=\\"utopia-storyboard-uid\\"
   >
-    <div data-uid=\\"ccc aaa scene-aaa\\">Thing</div>
+    <div
+      data-uid=\\"ccc aaa scene-aaa\\"
+      data-paths=\\"utopia-storyboard-uid/scene-aaa:aaa :utopia-storyboard-uid/scene-aaa\\"
+    >
+      Thing
+    </div>
```
Rheeseyb, line 15965: This is another case that seems to be missing a path

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
```diff
@@ -21228,9 +21710,15 @@ exports[`UiJsxCanvas render renders fine with two components that reference each
...
+    data-paths=\\":utopia-storyboard-uid\\"
     data-uid=\\"utopia-storyboard-uid\\"
   >
-    <div data-uid=\\"2a2~~~1 aaa bbb BBB scene\\">great</div>
+    <div
+      data-uid=\\"2a2~~~1 aaa bbb BBB scene\\"
+      data-paths=\\":2a2~~~1 utopia-storyboard-uid/scene:BBB :utopia-storyboard-uid/scene\\"
+    >
+      great
+    </div>
```
Rheeseyb, line 21721: I'm not sure what has happened with the paths here, but something is definitely wrong :grimacing:

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
```diff
@@ -21840,7 +22340,7 @@ export var storyboard = (
       },
     },
   },
-  "eee/fff:aaa/ddd": Object {
+  "eee/fff:aaa/bbb": Object {
```
Rheeseyb, line 22343: This is interesting. Something from here onwards has been parsed or walked completely differently :thinking:

editor/src/components/canvas/dom-walker.ts
```diff
@@ -827,16 +796,21 @@ function walkElements(
   }
   if (element instanceof HTMLElement) {
     // Determine the uid of this element if it has one.
-    const uids = getUIDsOnDomELement(element) ?? []
-    const filteredValidPaths = filterValidFocusedPathsForGeneratedElement(uids, validPaths)
+    const paths = getPathsOnDomElement(element)
+    // const filteredValidPaths = filterValidFocusedPathsForGeneratedElement(uids, validPaths)
```
Rheeseyb, line 800:
```suggestion
```

editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
```diff
@@ -238,6 +242,7 @@ export const SceneRootRenderer = betterReactMemo(
       ...sceneProps.props,
       [UTOPIA_SCENE_PATH]: scenePath,
       [UTOPIA_UIDS_KEY]: sceneUID,
```
Rheeseyb, line 244: Can we bin off the uid key with this change?

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
```diff
@@ -1133,9 +1133,15 @@ export var ${BakedInStoryboardVariableName} = (props) => {
             width: 400px;
             height: 400px;
           \\"
+          data-paths=\\":utopia-storyboard-uid\\"
           data-uid=\\"utopia-storyboard-uid\\"
         >
-          <div data-uid=\\"ccc aaa scene-aaa\\">hello</div>
+          <div
+            data-uid=\\"ccc aaa scene-aaa\\"
+            data-paths=\\"utopia-storyboard-uid/scene-aaa:aaa :utopia-storyboard-uid/scene-aaa\\"
```
Rheeseyb, line 1141: This is missing a path, but I think that's right since the `Thing` isn't focused?

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
```diff
@@ -1222,9 +1228,14 @@ export var ${BakedInStoryboardVariableName} = (props) => {
...
             height: 400px;
           \\"
+          data-paths=\\":utopia-storyboard-uid\\"
           data-uid=\\"utopia-storyboard-uid\\"
         >
-          <div id=\\"nasty-div\\" data-uid=\\"2a7~~~1 150~~~2 2f5~~~1 aaa scene-aaa\\">
+          <div
+            id=\\"nasty-div\\"
+            data-uid=\\"2a7~~~1 150~~~2 2f5~~~1 aaa scene-aaa\\"
+            data-paths=\\":150~~~2/2a7~~~1 :150~~~2 :2f5~~~1 utopia-storyboard-uid/scene-aaa:aaa :utopia-storyboard-uid/scene-aaa\\"
```
Rheeseyb, line 1237: Again, I think since `Thing` isn't focused, the paths should end at `aaa`

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -60,8 +60,8 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(280) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(290)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(320) // if this breaks, GREAT NEWS but update the test please :)
```
Rheeseyb, line 63: Hmmmmmmmmm.....

editor/src/utils/canvas-react-utils.ts
```diff
@@ -128,26 +133,30 @@ function shouldIncludeDataUID(type: any): boolean {
...
     // the response was an array of elements
-    return originalResponse.map((element) => attachDataUidToRoot(element, dataUids))
+    return originalResponse.map((element) => attachDataUidToRoot(element, dataUids, paths))
   } else if (!React.isValidElement(originalResponse as any)) {
     return originalResponse
   } else {
     if (shouldIncludeDataUID(originalResponse.type)) {
       return React.cloneElement(originalResponse, {
         [UTOPIA_UIDS_KEY]: appendToUidString(originalResponse.props[UTOPIA_UIDS_KEY], dataUids),
+        [UTOPIA_PATHS_KEY]: appendToUidString(originalResponse.props[UTOPIA_PATHS_KEY], paths),
```
Rheeseyb, line 159: This should probably not be using `appendToUidString` (or that function should be renamed)

editor/src/utils/canvas-react-utils.ts
```diff
@@ -269,15 +290,18 @@ const mangleExoticType = Utils.memoize(
           const originalFunction = p.children
           children = function (...params: any[]) {
             const originalResponse = originalFunction(...params)
-            return attachDataUidToRoot(originalResponse, uids)
+            return attachDataUidToRoot(originalResponse, uids, paths)
           }
         } else {
           const uidsToPass = uidsWithoutExoticUID(uids)
+          const pathsToPass = uidsWithoutExoticUID(paths)
```
Rheeseyb, line 297: I think this line might be wrong and might be causing some of the weirdness in the tests, so it's worth checking that this is behaving correctly with contexts and the like

editor/src/core/performance/performance-regression-tests.spec.tsx
balazsbajorics: WELL, ACTUALLY this is purely because we are now running the monkey function for every test. but that is fine, as the real editor is also always running the monkey function

balazsbajorics:
Important note: this PR does not (yet) change the selection (I didn't change the code of dom-lookup, that still uses uids), only the dom-walker's behavior! 

I'm not sure if that should be a separate PR or rolled into this one

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
balazsbajorics: this was fine, ccc is coming from an ARBITRARY_JS block

editor/src/components/canvas/ui-jsx-canvas-renderer/scene-root.tsx
balazsbajorics: 🤔 actually I think we could. however, since the data-uid is inserted as a prop, unless we make it a not-prop, we will still print _some_ UIDs. I think in the short term it's better to leave the system of uid printing as-is, and only remove it once we believe we can fully remove it (ie not even pretend it's a prop, attach it to the parsed JSXElement in some other way)

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
balazsbajorics: `eee/fff:aaa/bbb` was previously not in the spy data, and now it is!

balazsbajorics: the spy data for aaa/ddd is intact, it has been moved lower though.

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
balazsbajorics: in this test case "App" is an arbitrary js. this is the last known case where the template path is going to be gibberish, but since the editor does not support editing these projects, this is not a regression, just making an existing problem explicit

editor/src/utils/canvas-react-utils.ts
Rheeseyb: Actually ignore this :point_up: I think this should still result in it working the same way as before...

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
balazsbajorics: I would say this is fine as is, :2a2~~~1 is an irrelevant path, and utopia-storyboard-uid/scene:BBB is valid. aaa and bbb come from arbitrary JS blocks so we can't edit them hence the missing inner template paths

editor/src/components/canvas/dom-walker.ts
```diff
@@ -50,52 +51,20 @@ import { extractOriginalUidFromIndexedUid, getUIDsOnDomELement } from '../../cor
 import { mapDropNulls } from '../../core/shared/array-utils'
 import { optionalMap } from '../../core/shared/optional-utils'
 import { fastForEach } from '../../core/shared/utils'
+import { valid } from 'chroma-js'
```
enidemi, line 54: this looks like an unused import

enidemi (approved):

editor/src/components/canvas/dom-walker.ts
balazsbajorics: indeed it was

PR #1078, opened 2021-04-13
Canvas colour control

OmarDaSilva:
**Problem:**
Selected focused component on the canvas is visually not clear 
 
**Fix:**
Completely changed the function getSelectionColor, so that it can choose the right colour depending on the selected component

Focusable Component (blue): 
Screenshot 2021-04-13 at 17 50 41

Another Component Inside of isolated component highlight (orange):
Screenshot 2021-04-13 at 17 47 49


maltenuhn:
Nice @OmarDaSilva! Can you post an image or two of what these controls look like - ideally with the test project copied-as-code from pizza?

OmarDaSilva:
This PR is now ready for review!

enidemi (approved):

editor/src/components/canvas/controls/outline-control.tsx
```diff
@@ -58,9 +56,9 @@ function isDraggingToMove(
   return dragState != null && dragState?.type === 'MOVE_DRAG_STATE' && targetIsDragged
 }
 
-export class OutlineControls extends React.Component<OutlineControlsProps> {
-  getDragStateFrame = (target: TemplatePath): CanvasRectangle | null => {
-    const { dragState } = this.props
+export const OutlineControls = (props: OutlineControlsProps) => {
+  const getDragStateFrame = (target: TemplatePath): CanvasRectangle | null => {
```
balazsbajorics, line 60: this is a good idea, converting an old class component to a function component! :) 

these functions that used to live on the Class are now turned into inline functions. there's some important differences between how the two kinds work – in the previous version, the `getDragStateFrame` function would be instantiated once when the class lifecycle begins, and we would use the same reference for every new call of the render() function. In this new function component, for every single render this line of code will re-run, creating a new instance of the function `getDragStateFrame`. The most important thing to know is that this means `getDragStateFrame` will have a new _reference_ for every render. Because a lot of React depends on the reference being stable (being the same from render to render), this can very easily lead to performance problems.

The easy solution is to just remember, as a rule, **when you are doing a class conversion, you should always wrap the functions to a `React.useCallback`**, like so:

```javascript
const getDragStateFrame = React.useCallback((target: TemplatePath): CanvasRectangle | null => {
   ....
  }, 
[   /* list the dependencies here */ ])
```

This will bring the behavior of the functions a little bit closer to how it worked in the class!

balazsbajorics: Okay, I think this needs futher elaboration. 

There are many exceptions to this rule, a function actually only needs React.useCallback() wrapping if it is defined inside the Function component and
1. used as a prop in the JSX 
2. listed as a dependency of a hook (like, for example, another useCallback is listing this function as a dependency)

But because it is hard to tell when this rule should be appplied (our `eslint` setup struggles with detecting some mistakes too) I would advise that it is best to just stick to it and wrap everything in a React.useCallback when you are converting a class to a function component 🙂

BTW, when you are creating a new component, it is best if you don't make functions inside other functions unless you actually need to use them as callbacks. But because this is code conversion, it's perfectly acceptable that you did not refactor this function to be in the root level of the file (which would be the nicest, code-quality-wise)

OmarDaSilva: Commit 3420f3d addresses this, thanks for the feedback!

OmarDaSilva:
PR now ready review

editor/src/components/canvas/controls/outline-control.tsx
```diff
@@ -58,154 +56,172 @@ function isDraggingToMove(
...
       } else {
-        return offsetRect(startingFrameAndTarget.frame, dragState.drag)
+        return null
       }
-    } else {
-      return null
-    }
-  }
+    },
+    [props],
```
Rheeseyb, line 80: This callback only changes on the `dragState` (as opposed to the entire `props` object), so you should extract that prop before defining the function, and change this dependency array to just include that one prop (otherwise it will result in a new function being created for every change to any prop in the `props` object)

Rheeseyb: Or in fact you can do what you did below where you added the single prop as a dependency `props.componentMetadata`

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -183,7 +183,7 @@ const computeResultingStyle = (
   return result
 }
 
-function isFocused(focusedElementPath: ScenePath | null, path: TemplatePath): boolean {
+export function isFocused(focusedElementPath: ScenePath | null, path: TemplatePath): boolean {
```
Rheeseyb, line 186: Rather than exporting this function here, it would make more sense to move the function to a more core shared file (`template-path.ts` would work). Otherwise we are creating a direct dependency between the Canvas Controls code and the Navigator code (which isn't something we should do)

utopia-vscode-common/package-lock.json
```diff
@@ -424,6 +424,16 @@
       "dev": true,
       "optional": true
     },
+    "bindings": {
```
Rheeseyb, line 427: Interesting that this file changed. Please can you revert it from this PR since it's unrelated.

Rheeseyb (commented):
Looks good to me! Once those bits of feedback are addressed this will be good to go :+1:

OmarDaSilva:
This branch has now addressed all feedback and is now ready to merge just needs approvals now!

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -183,19 +183,19 @@ const computeResultingStyle = (
...
+//   } else {
+//     return (
+//       TP.scenePathUpToElementPath(
+//         focusedElementPath,
+//         TP.elementPathForPath(path),
+//         'dynamic-scene-path',
+//       ) != null
+//     )
+//   }
+// }
```
Rheeseyb, line 198:
```suggestion
```

Rheeseyb (approved):

balazsbajorics (approved):

PR #1079, opened 2021-04-13
Fix selection that totally regressed

balazsbajorics:
**Problem:**
Selection severely regressed on the Canvas. The problem comes from `Canvas.getAllTargetsAtPoint`. It _seems like_ it is caused by the merging of https://github.com/concrete-utopia/utopia/pull/1058 which broke the offsets calculated by `Canvas.getFramesInCanvasContext`.

**Fix:**
Instead of trying to find exactly what broke the offsets calculated by `Canvas.getFramesInCanvasContext`, I realized that it is old code relying on `localFrame` and in this new world we have access to the calculated `globalFrame`s so there's no need to understand what caused the problem, and I just switched the code to use `globalFrame`. There's still a chance that the missing offset broke something else (likely mouse-driven selection of scenes or some such thing), but if the tests pass for now, I think this should be merged.

### DIFF
--- a/editor/src/components/canvas/canvas.ts
+++ b/editor/src/components/canvas/canvas.ts
@@ -91,21 +91,16 @@ const Canvas = {
     useBoundingFrames: boolean,
   ): Array<FrameWithPath> {
     function recurseChildren(
-      offset: CanvasPoint,
       component: ElementInstanceMetadata,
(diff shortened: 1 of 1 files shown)


Rheeseyb (approved):

balazsbajorics:
I'm going to merge this to unblock the people!

enidemi (approved):

PR #1080, opened 2021-04-13
Feature/navigator component icons multifile

enidemi:
**Problem:**
Navigator icons are not aware if something is focusable or not since multifile editing. And Context-menu enable/disable state will need the same functionality.

**Commit Details:**
- allow context-menu items enable/disable themselves from a callback
- added helper `isFocusableComponent` that will be used later for the context-menu, now it is used for the navigator component icons
- added helper that returns components and imports for instances, like for sample project card-instance will have the `app.js` components
- added helper that returns components for instances + original components and original imports, for the sample project card-instance will have the `card.js` components and imports + `app.js` components. in this version the `app.js` components are needed because otherwise the elementname for path resolve is not working, I think the extra components shouldn't cause problems.


editor/src/components/context-menu-items.ts
```diff
@@ -135,9 +135,10 @@ export const toggleShadowItem: ContextMenuItem<CanvasData> = {
 }
 
 export const setAsFocusedElement: ContextMenuItem<CanvasData> = {
-  name: 'Edit Component',
-  // todo only enable if thing is component
-  enabled: true,
+  name: 'Set As Focused Element',
+  enabled: (data) => {
+    return true
```
balazsbajorics, line 140: I think this is still not going to be good enough here :)

enidemi: that's for the next PR! :D I should have split it into 2 branches :(

balazsbajorics (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1539,6 +1545,36 @@ export const MetadataUtils = {
...
+    path: TemplatePath,
+    components: UtopiaJSXComponent[],
+    metadata: ElementInstanceMetadataMap,
+    imports: Imports,
+  ): boolean {
+    const elementName = MetadataUtils.getJSXElementName(path, components, metadata)
+    const elementInstancePath = TP.isInstancePath(path)
+      ? path
+      : TP.instancePathForElementAtScenePath(path)
+    const element = MetadataUtils.getElementByInstancePathMaybe(metadata, elementInstancePath)
```
Rheeseyb, line 1558: We have a function for doing exactly this `findElementByTemplatePath`, which exists for smoothing the transition between the separate path types and a future where they don't exist

Rheeseyb (approved):

editor/src/core/model/element-metadata-utils.ts
enidemi: fixed!

PR #1085, opened 2021-04-14
Feature/toggles for codeeditor and canvas

maltenuhn:
**Problem:**
1. Our code editor can be toggled, the canvas can't. 
2. Opening a code file doesn't show the code editor
3. The Storyboards "list" in the left menu doesn't open the design tool if it's closed and you click on it
4. There's no empty state if you hide both code editor and design tool

**Fix:**
- Introduce a 'design tool' toggle in the left menu
- Duplicate the 'code editor' toggle to go there
- Introduce a parameter when opening a code file to force the code editor to open (explicit user actions for opening files should use this, implicit ones shouldn't - so that eg if the user has hidden the code editor, clicking on a component that opens a code file doesn't then also show the editor
- Open the design tool from the Storyboards panel, and correctly show no storyboard as selected if the design tool isn't open
- Bonus content: an empty state when neither design tool nor code editor are open

[Prototyped in Utopia](https://utopia.pizza/p/e4e4f2ac-odd-punch/)


editor/src/components/canvas/design-panel-root.tsx
```diff
@@ -29,6 +35,72 @@ interface NumberSize {
...
+        display: 'flex',
+        flexDirection: 'column',
+        alignItems: 'center',
+        justifyContent: 'space-around',
+        borderRadius: 4,
+        backgroundColor: 'white',
+        border: '1px solid lightgrey',
+      }}
+    >
+      <img src={'/editor/icons/light/special/palm4.png'} alt='aesthetic palmtree' />
```
balazsbajorics, line 69: I'm not sure but I think the url should be wrapped in `getPossiblyHashedURL`. @Rheeseyb can you help me out?

editor/src/components/navigator/left-pane.tsx
```diff
@@ -188,18 +188,23 @@ const StoryboardListItem = styled.div<StoryboardListItemProps>((props) => ({
...
-  const { dispatch, openFile, projectContents } = useEditorState((store) => {
+  const { dispatch, openFile, projectContents, isCanvasVisible } = useEditorState((store) => {
     return {
       dispatch: store.dispatch,
       openFile: store.editor.canvas.openFile?.filename,
       projectContents: store.editor.projectContents,
+      isCanvasVisible: store.editor.canvas.visible,
     }
-  }, 'FileBrowser')
+  }, 'StoryboardsPane')
```
balazsbajorics, line 198: thanks!

balazsbajorics (approved):

editor/src/components/canvas/design-panel-root.tsx
seanparsons: Definitely should.

seanparsons (approved):

editor/src/components/canvas/design-panel-root.tsx
maltenuhn: I've just turned it into a proper Icon so it can benefit from the helper / dealing with the hashed URL

PR #1087, opened 2021-04-15
Feature/context menu focus menuitem changes

enidemi:
**Problem:**
Focus component in the context-menu should be hidden for elements.

**Commit Details:**
- change label from `Set As Focused Element` to `Edit Component`
- `isHidden` callback for focus context-menu item
- update existing isHidden for element list, this was using data from props which is unusual because we use the data in other cases

### DIFF
--- a/editor/src/components/canvas/controls/select-mode-control-container.tsx
+++ b/editor/src/components/canvas/controls/select-mode-control-container.tsx
@@ -148,22 +148,8 @@ export class SelectModeControlContainer extends React.Component<
   onContextMenu = (event: React.MouseEvent<HTMLDivElement>) => {
     event.stopPropagation()
     event.preventDefault()
-    const elementsUnderCursor = getAllTargetsAtPoint(
-      this.props.componentMetadata,
-      this.props.selectedViews,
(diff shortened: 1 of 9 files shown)


editor/src/components/canvas/controls/select-mode-control-container.tsx
```diff
@@ -148,22 +148,8 @@ export class SelectModeControlContainer extends React.Component<
   onContextMenu = (event: React.MouseEvent<HTMLDivElement>) => {
     event.stopPropagation()
     event.preventDefault()
-    const elementsUnderCursor = getAllTargetsAtPoint(
```
balazsbajorics, line 151: so we were only using this `elementsUnderCursor` to tell if something should be hidden or not? :O I thought this was also populating the list too!

seanparsons (approved):

editor/src/components/canvas/controls/select-mode-control-container.tsx
enidemi: the list contains all element path!

Rheeseyb (approved):

PR #1088, opened 2021-04-15
Multi-file Movement

seanparsons:
**Problem:**
Previously none of the logic around moving/resizing was multi-file editing aware so would be unable to handle an element defined deeply within other files.

**Fix:**
A lot of functions involved taking arrays of components and component metadata, transforming those and returning them again. Many of those have now been switched to taking `EditorState` as an input and returning it again as an output. This also means that some of the logic produces the transient state from an `EditorState`, instead of from the individual parts as it previously did.

**Limitations:**
- It appears there's still an issue with resizing in that the controls aren't showing up for deeply nested components, but I thought given that pretty much everything is in place at this point I'd rather get the PR done then investigate that afterwards.

**Commit Details:**
- Removed some filtering around the targets of `useStartDragState`.
- Speculatively upgraded electron to 6.1.12 to see if it would result in console logging making an appearance for browser tests on Linux systems.
- Added a `beforeAll` inside a few `describe` calls of browser tests so they run properly when run alone.
- Tests in `canvas-utils-unit-tests.spec.ts` switched to using `EditorState` instead of `ParseSuccess` for their test data.
- Implemented `applyTransientFilesState` to extract out the logic of taking the values held within `TransientFilesState` to the `EditorState`.
- Much of the logic in `canvas-utils.ts` now uses functions like `withUnderlyingTarget` and `modifyUnderlyingForOpenFile` to reach down into or transform the `EditorState`.
- `getTransientCanvasStateFromFrameChanges` implemented to produce the transient state from an `EditorState`.
- `NormalisePathImportNotFound` now includes what it couldn't find.

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
@@ -257,10 +257,6 @@ function useStartDragState(): (
         entireEditorStoreRef.current.editor,
       )
 
-      const elementsThatRespectLayout = selectElementsThatRespectLayout(
-        entireEditorStoreRef.current,
-      )
-
       const duplicate = event.altKey
       const duplicateNewUIDs = duplicate
         ? createDuplicationNewUIDs(selectedViews, componentMetadata, rootComponents)
@@ -269,14 +265,8 @@ function useStartDragState(): (
       const isTargetSelected = selectedViews.some((sv) => TP.pathsEqual(sv, target))
       const selection =
         isTargetSelected && TP.areAllElementsInSameScene(selectedViews) ? selectedViews : [target]
-      const moveTargets = selection.filter(
-        (view) =>
-          TP.isScenePath(view) ||
-          TP.isStoryboardDescendant(view) || // FIXME This must go in the bin when we separate the Scene from the component it renders
-          elementsThatRespectLayout.some((path) => TP.pathsEqual(path, view)),
-      )
 
-      let originalFrames = getOriginalCanvasFrames(moveTargets, componentMetadata)
+      let originalFrames = getOriginalCanvasFrames(selection, componentMetadata)
       originalFrames = originalFrames.filter((f) => f.frame != null)
 
       const selectionArea = boundingRectangleArray(
@@ -300,7 +290,7 @@ function useStartDragState(): (
             duplicateNewUIDs,
             start,
             componentMetadata,
-            moveTargets,
+            selection,
           ),
         ),
       ])
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
@@ -263,6 +263,9 @@ function useStartDragState(): (
         : null
 
       const isTargetSelected = selectedViews.some((sv) => TP.pathsEqual(sv, target))
+
+      // FIXME: Re-establish filtering based on `selectElementsThatRespectLayout`.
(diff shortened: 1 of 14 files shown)


editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
```diff
@@ -257,10 +257,6 @@ function useStartDragState(): (
         entireEditorStoreRef.current.editor,
       )
 
-      const elementsThatRespectLayout = selectElementsThatRespectLayout(
```
enidemi, line 260: `selectElementsThatRespectLayout` prevents moving/resizing elements that can't accept style layout props, dragging should keep them in place to show this behavior. Resizing also uses this helper somewhere near here to show the resize corner on elements that can be really resized.

seanparsons: Seems like the next item on the track is to fix that exact function, so I'll reintroduce this call with that work, just at the moment it's filtering out everything that isn't from the canvas file by default.

balazsbajorics: could you leave a TODO here please?

also, are there any other known regressions made by this PR? it'd be worth tracking them in the paper doc before we loose track of some…

seanparsons: Not known ones no.

enidemi (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
seanparsons: > Why has this filtering been removed?
Because for the dev project it held just two values, which were the same thing twice and that resulted in effectively nothing being movable or resizable.

Rheeseyb: Which is by design since the dev project doesn't define any property controls, which are what's used to determine if an element can be dragged.

seanparsons: Except when I say that I mean even when the Card is focused and there are `Rectangle` elements in the mix.

Rheeseyb: Yep which appears to be a bug on master that affects a focused component imported from another file, which should either be fixed separately, or the filtering removed as a conscious decision in a separate PR. Otherwise this is mixing an unexpected feature change into this PR.

Obviously the bug got in the way of developing this feature, but I'd rather we didn't remove the filtering. We should either provide some (strictly temporary) way to disable it (for the sake of this work and passing the tests) and address the bug immediately after, or discuss removing the filtering completely.

seanparsons: See this comment: https://github.com/concrete-utopia/utopia/pull/1088#discussion_r614031198

balazsbajorics: I'm good with a FIXME + tracking this

balazsbajorics:
Can you please add some tests for multifile setups? You could take some of the existing canvas-utils test cases and create multifile versions of those.

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
seanparsons: Added in c02ad628.

Rheeseyb (approved):

seanparsons:
> Can you please add some tests for multifile setups? You could take some of the existing canvas-utils test cases and create multifile versions of those.

Added some tests in e85beeef.

balazsbajorics (approved):

PR #1083, opened 2021-04-14
Natural Scene Components

Rheeseyb:
**Problem:**
We want to replace `<Scene component={App} />` with `<Scene><App/></Scene>`

**Fix:**
This is the final step of that, but I'm splitting into 2 PRs - one introducing the changes to support natural scenes, and then a second upcoming PR to remove the old style scenes (for the sake of keeping the functional change PR as free from noise as possible).

This PR introduces a very basic `Scene` component into `utopia-api` (so that user's projects can build and run outside of Utopia), and a `SceneComponent` component that wraps the above when running in the canvas (to provide styling based on the canvas state). It also updates the targeting functionality used throughout for selection and insertion to reflect the updated hierarchy, and also updates the indentation depth in the Navigator for the same reason.

**Demo:**
natural-scenes

**Note:**
This work hasn't ported across the dynamic / static scene functionality - I'm not sure if that still makes sense now... (will bring that up in Slack)

PR #1091, opened 2021-04-15
Completely Remove Old Style Scene

Rheeseyb:
Follow up from #1083 

The previous PR introduced all of the functionality to allow us to bin off the old style scenes, but stopped short of removing it from the tests. This PR updates all default projects to use the natural scene style, removes the now dead `componentName` prop from the metadata, and updates all scenes to use the natural scene style. It's a monster because of the number of tests that required updating to finish this off.

The only "functional" changes in here are the removal of `componentName` (and its uses in `element-metadata-utils.ts`), the updating of default scenes to use the new natural scenes, and the removal of `SceneRootRenderer`. Everything is a consequential test update.

Upon approval, this will be merged into #1083 before that PR is merged to master.


enidemi (approved):

editor/src/components/editor/store/editor-state.spec.ts
```diff
@@ -127,9 +132,28 @@ describe('modifyUnderlyingTarget', () => {
...
+                left: 0,
+                top: 0,
+                width: 50,
+                height: 50,
+                backgroundColor: 'red',
+              }}
+              data-thing='a thing'
+            />
+            <Rectangle
+              style={{
```
balazsbajorics, line 148: this style prop is new here, right? where is this coming from?

Rheeseyb: The tests were previously using a different version of the complex multifile project for no apparent reason, so I unified that to ensure we kept the two inline (as I was updating the sample project and wondering why it didn't break multifile tests)

editor/src/test-cases/majestic-broker.ts
```diff
@@ -437,27 +437,18 @@ export var App = () => {
 }
 export var storyboard = (
   <Storyboard>
-    <Scene
-      component={App}
-      props={{}}
-      style={{ position: 'absolute', left: 0, top: 0, width: 522, height: 652 }}
-    />
-    <Scene
-      data-label={'ScrollablePhotoGrid'}
```
balazsbajorics, line 446: did scenes lose their data-label?

balazsbajorics (approved):
my brain is blunt now

editor/src/test-cases/majestic-broker.ts
Rheeseyb: Nope, this one was accidental

PR #1092, opened 2021-04-15
feature(canvas) clicking into element sets focus mode for components without children passed in

enidemi:
**Problem:**
Entering component editing mode is possible only from the context-menu, component's inner children are not selectable without it. This is not matching with the idea for if an element is editable and visible with some doubleclicking it should become selected. It is allowed focusing on instances only with inner children, without direct props.children.

**Commit Details:**
- extending `useSelectModeSelectAndHover` to set focused element when the foundTarget is a focusable leaf component
- a helper in `metadata-utils`

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -1578,11 +1578,7 @@ export const MetadataUtils = {
     metadata: ElementInstanceMetadataMap,
     imports: Imports,
   ): boolean {
-    if (MetadataUtils.getChildrenPaths(metadata, path).length > 0) {
-      return false
-    } else {
-      return MetadataUtils.isFocusableComponent(path, components, metadata, imports)
-    }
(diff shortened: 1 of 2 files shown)


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1572,6 +1572,18 @@ export const MetadataUtils = {
...
+    path: TemplatePath,
+    components: UtopiaJSXComponent[],
+    metadata: ElementInstanceMetadataMap,
+    imports: Imports,
+  ): boolean {
+    if (MetadataUtils.getChildrenPaths(metadata, path).length > 0) {
+      return false
+    } else {
+      return MetadataUtils.isFocusableComponent(path, components, metadata, imports)
+    }
```
balazsbajorics, line 1585:
```suggestion
    return MetadataUtils.getChildrenPaths(metadata, path).length === 0 && MetadataUtils.isFocusableComponent(path, components, metadata, imports)
```

balazsbajorics (approved):

Rheeseyb (approved):

PR #1095, opened 2021-04-16
Tidying up after Natural Scenes work

### LINKED ISSUE #1050: Remove `component` from the `Scene` props completely
Remove the `component` prop from the `Scene` element, so that now Scenes will render a div containing their children (each of which will now act as though it is always a focused element). This is a follow up of #1049 

- [x] Remove `SceneRootRenderer` and replace with a simple `Scene` element
  - [x] Replace all uses of the old style scenes with the `component` prop
- [x] Simplify or completely remove the Scene Inspector (in either case we want to show the regular Inspector for the Scene element, so the most likely choice would be to remove the special Scene Inspector and add property controls for the Scene element)
 - [x] In the Navigator render separate rows for the Scene, its children, and the root elements of those children
    - [x] Rewrite or remove any special casing of "component" entry points in the Navigator
  - [x] ~Implement a right-click command for creating a “New Scene from Selection”~ (Moved to #1094)

For this work it is maybe worth at this point removing the separate `ScenePath` and `InstancePath`, and replacing both with a single path type.

Rheeseyb:
Closes #1050 

**Problem:**
We still had old code in the Navigator and Inspector specifically for handling Scenes. Some of this code needed updating (for the sake of styling in the case of the Navigator), and some just needed placing gently in the bin (e.g. the Scene Inspector).

**Fix:**
The Navigator code for styling was updated using a helper function for roughly guessing if an element is a Scene (based on it being a `Scene` and that's about it). This needs improving by checking against imports to find where it is actually resolved from, but is good enough for now. I've also removed the "*Scene* editing *App*" heading since Scenes can have multiple children, so it uses the default behaviour of using either the `data-label` (if available), or the element tag itself (so the word "Scene").

The Scene Inspector has been removed, though there is likely some residual uncalled code left relating to that which could do with being removed, but I'm opening this PR to switch to the code editor related bugs.

**Note:** I've updated the default dev project to include labels on the Scenes for clarity.


editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -226,6 +229,26 @@ function useStyleFullyVisible(path: TemplatePath): boolean {
...
 }
 
+export function isProbablySceneFromMetadata(
+  jsxMetadata: ElementInstanceMetadataMap,
+  path: TemplatePath,
+): boolean {
+  const elementMetadata = MetadataUtils.findElementByTemplatePath(jsxMetadata, path)
+  return (
+    elementMetadata != null &&
+    isRight(elementMetadata.element) &&
```
seanparsons, line 239: Probably be ever so slightly cleaner with a `foldEither`.

seanparsons (approved):

editor/src/components/navigator/layout-element-icons.ts
```diff
@@ -75,24 +76,22 @@ export function createLayoutOrElementIconResult(
...
+  if (isProbablySceneFromMetadata(metadata, path)) {
     return {
       iconProps: {
         category: 'component',
         type: 'scene',
         width: 18,
         height: 18,
       },
-      hasWidthOrHeight: hasWidthOrHeight,
+      hasWidthOrHeight: false,
```
balazsbajorics, line 94: this seems like an unrelated change?

balazsbajorics (approved):

editor/src/components/navigator/layout-element-icons.ts
Rheeseyb: Not quite. This would previously always be `false` for a `ScenePath` (and therefore any descendant of the storyboard), but that logic was coming from the deleted line above.

editor/src/components/navigator/navigator-item/navigator-item.tsx
Rheeseyb: Yep that's fair. I've taken a note (physically on my notepad in front of me) and will change that when I return to the tidying up work.

enidemi (approved):

PR #1096, opened 2021-04-16
Reinstate resizing filtering.

seanparsons:
**Problem:**
Some element filtering was removed here: https://github.com/concrete-utopia/utopia/pull/1088#pullrequestreview-636631904

**Fix:**
Reinstated the filtering and brought it into the multi-file world of wonder.

**Commit Details:**
- `withUnderlyingTarget` now renamed to `withUnderlyingTargetFromEditorState`
  and `withUnderlyingTarget` operates on the fields of `EditorState`
  individually.
- `getPropertyControlsForTarget` now uses `withUnderlyingTarget` and
  `importedFromWhere`
- Fixed `targetRespectsLayout` to work with the modified
  `getPropertyControlsForTarget`.
- Fixed `selectElementsThatRespectLayout` to work with the modified
  `targetRespectsLayout`.
- Update to `utopia-vscode-common/package-lock.json`
- Added a bunch of return types to `select-mode-control-container.tsx`.


editor/src/components/canvas/controls/select-mode-control-container.tsx
```diff
@@ -96,13 +91,14 @@ export class SelectModeControlContainer extends React.Component<
   static getDerivedStateFromProps(
     props: SelectModeControlContainerProps,
     previousState: SelectModeControlContainerState,
-  ) {
+  ): SelectModeControlContainerState {
```
balazsbajorics, line 94: Unless your explicit intention is resetting lastHovered to null for every props change, I think this here should be `Partial<SelectModeControlContainerState>` and the `lastHovered: null` can be removed then.

enidemi (approved):

editor/src/core/property-controls/property-controls-utils.ts
```diff
@@ -354,65 +361,77 @@ export function getPropertyControlsForTargetFromEditor(
...
+          filenameForLookup = openFilePath.replace(/\.(js|jsx|ts|tsx)$/, '')
+        }
       } else {
+        filenameForLookup = importedFrom
+      }
+      if (filenameForLookup == null) {
         return null
+      } else {
+        // TODO figure out absolute filepath
+        const absoluteFilePath = filenameForLookup.startsWith('.')
```
balazsbajorics, line 420: is this `TODO figure out absolute filepath` still relevant here? if not, I'd remove this `.` removal gibberish altogether.
But if it _is_ relevant, does that have multi-file implications for us?

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
```diff
@@ -264,12 +267,17 @@ function useStartDragState(): (
...
-      // FIXME: Re-establish filtering based on `selectElementsThatRespectLayout`.
-      // https://github.com/concrete-utopia/utopia/pull/1088/files/460e68fb6b4156833744a5441c9e56d662c8b51d#r614029462
       const selection =
         isTargetSelected && TP.areAllElementsInSameScene(selectedViews) ? selectedViews : [target]
 
-      let originalFrames = getOriginalCanvasFrames(selection, componentMetadata)
+      const moveTargets = selection.filter(
+        (view) =>
+          TP.isScenePath(view) ||
+          TP.isStoryboardDescendant(view) || // FIXME This must go in the bin when we separate the Scene from the component it renders
```
Rheeseyb, line 276: These two lines still need replacing, but I don't think that should happen in this PR. The reason for this filtering was to enable dragging of actual `Scene` elements, which this combination no longer captures. I'll add a follow up ticket for that.

Rheeseyb: In fact, I think the `Scene` elements would be included in the `elementsThatRespectLayout` since I added the relevant property controls for them. If that's the case, please can you delete the `TP.isScenePath(view) || TP.isStoryboardDescendant(view)` part of the condition.

editor/src/components/canvas/controls/select-mode-control-container.tsx
seanparsons: Ah, that would make more sense, I thought this might just be breezing through because `getDerivedStateFromProps` is effectively not checked that the types line up.

editor/src/core/property-controls/property-controls-utils.ts
seanparsons: I'd just moved it around as I'd refactored the rest, but given the logic I'd somewhat assume it would be called `relativeFilePath`, as that's what it would imply. But I think now this should always be an absolute path anyway. Will look into it.

seanparsons: Fixed in 3455633c.

editor/src/components/canvas/controls/select-mode/select-mode-hooks.ts
seanparsons: Fixed in 3455633c.

editor/src/components/canvas/controls/select-mode-control-container.tsx
seanparsons: Fixed in 3455633c.

Rheeseyb (approved):

balazsbajorics (approved):

PR #1099, opened 2021-04-16
Restore formatting on saving in the code editor

### LINKED ISSUE #1023: Re-introduce Prettier-on-CMD+S
**Problem**
We previously ran prettier on save, to match the behaviour of other online coding environments. Today, this no longer happens

**Solution**
Reintroduce prettier on save (CMD+S) of the code file

Rheeseyb:
Fixes #1023

**Problem:**
We lost the automatic formatting when saving in the code editor.

**Fix:**
Move the `prettier-utils` source to the shared project `utopia-vscode-common`, and then use it in `utopia-vscode-extension` to apply edits to a text document when saving it (as long as the save is flagged as a manual save by VS Code)

I also noticed that the `install-editor` script wasn't building and pulling across changes to the extension, so I've fixed that.


seanparsons (approved):

balazsbajorics:
what's up with Puppeteer?

Rheeseyb:
> what's up with Puppeteer?

I think it just fell over. It ran fine on #1100 and #1101 which were both branched off this :shrug:

shell.nix
```diff
@@ -23,9 +23,14 @@ let
     (pkgs.writeScriptBin "install-editor" ''
       #!/usr/bin/env bash
       set -e
-      cd $(${pkgs.git}/bin/git rev-parse --show-toplevel)/utopia-vscode-common
+      update-vscode-build-extension
+      cd $(${pkgs.git}/bin/git rev-parse --show-toplevel)/editor
       ${node}/bin/npm --scripts-prepend-node-path=true install
-      ${node}/bin/npm --scripts-prepend-node-path=true run build
+    '')
+    (pkgs.writeScriptBin "install-editor-ci" ''
```
balazsbajorics, line 30: I think it's important information that for first time checkout out the editor, this is the script to run.

Maybe we could make a dedicated `install-editor-first-time` or a more general `setup` script that would also execute on the CI?

shell.nix
```diff
@@ -23,9 +23,14 @@ let
     (pkgs.writeScriptBin "install-editor" ''
       #!/usr/bin/env bash
       set -e
-      cd $(${pkgs.git}/bin/git rev-parse --show-toplevel)/utopia-vscode-common
+      update-vscode-build-extension
```
balazsbajorics, line 26: in https://github.com/concrete-utopia/utopia/pull/1099/files#diff-e53dfbfffe62ae3c0b411b3938ccffa9fb6a2ecc565f55785ef8daa756631a6bL270 

didn't you rename `update-vscode-build-extension` to `pull-extension` ?

balazsbajorics (approved):
I am not entirely clear about the new shell.nix, the rest of the PR is fine

PR #1101, opened 2021-04-16
Basic follow me feature implementation

Rheeseyb:
**Problem:**
We want to support a "follow me" feature, whereby selecting an element in the editor (via the Canvas or Navigator etc.) will open the file containing that element in the code editor, and jump to the corresponding block of code.

**Fix:**
To do this we need to extend the highlight bounds mapping to include the file path. For this iteration I've added a function for mapping over the project to build that, and used that for generating the decorations and for sending the selection changed message.

In the extension, we now open the file for the selection if it isn't already opened, and also update the selection as well as the revealed lines (to prevent undesirable jumping that occurs when re-opening a file, as VSCode tries to reveal the previously selected code from that file).

Also, we filter out `onDidChangeTextEditorSelection` events that weren't a result of a mouse or keyboard interaction in the code editor, again to prevent undesirable jumping (otherwise opening a file triggers that event with the selection position from the last time that file was open, which would then change the selection in the editor).


utopia-vscode-extension/src/extension.ts
```diff
@@ -219,20 +225,36 @@ function cursorPositionChanged(event: vscode.TextEditorSelectionChangeEvent): vo
...
+    if (!alreadySelected) {
+      const selectionRange = getVSCodeRange(boundsInFile)
+      visibleEditor.selection = new vscode.Selection(selectionRange.start, selectionRange.start)
+    }
+
+    const shouldReveal = !(alreadySelected && alreadyVisible)
+    if (shouldReveal) {
+      visibleEditor.revealRange(
+        rangeToReveal,
+        vscode.TextEditorRevealType.InCenterIfOutsideViewport,
```
Rheeseyb, line 256: I changed the reveal type here from the default, as this felt more natural to me when playing with this. The default behaviour is to reveal the code with minimal scrolling, which either opens it at the very top or the very bottom. This behaviour will reveal the code in the center (but only if the code was previously not visible). I don't mind changing this back though if others dislike it.

editor/src/core/model/project-file-utils.ts
```diff
@@ -230,6 +233,24 @@ export function getHighlightBoundsFromParseResult(
   )
 }
 
+export function getHighlightBoundsForProject(
```
Rheeseyb, line 236: If this function ends up being too costly, we can instead add the file path to the highlight bounds during the parsing at no extra cost.

seanparsons: I don't think it should be too costly, albeit it is running for each selected and highlighted view.

seanparsons (approved):

enidemi (approved):

PR #1102, opened 2021-04-19
Fix the UID of changed code

balazsbajorics:
**Problem:**
The UIDs are no longer printed to the code. When we are parsing a code file, we generate UIDs by hashing the jsx elements. The solution is overall pretty elegant and it makes the editor feel like magic. However, there is a big downside: if the user types in the code editor to update an element, that element will also be automatically selected on the Canvas and Navigator and Inspector. But when the selected element's code changes, its hash will change too. And that means its UID will change as well. And that means all the template paths pointing to the selected element, the highlighted element, the spy and dom-walker metadata etc will refer to a nonexistent element. We had an earlier mitigation in place that would salvage the selected view, but it would not save the metadata. Which means that as the user was typing in the code editor, the inspector would flash as it got a new template path with no metadata, and flash again as fresh metadata arrived a frame later. 

**Fix:**
I basically took the same mitigation we had before, but moved it to the end of `parseCode`: 
We zip the old projectFile and new projectFile contents, and we try to match up UIDs by their index among their siblings. If we find a single mismatched UID, (to prevent accidentally matching the wrong element after a delete or insertion or copy-paste or code duplication) we assume that we found the same element but a different hash. We then update the uid to the old one, and also fix the highlightBounds map.

**Details:**
* `parseCode` now takes an `oldParseResultForUIDComparison: ParsedTextFile | null` parameter. I made it explicit instead of optional, so I could breeze through all the places using `parseCode` and decide if I should provide null or a real oldParseResult
* I made a new function `fixParseSuccessUIDs` that runs at the end of `parseCode`. It zips the old parsed file and the new parsed file against each other, and tries to find matching UIDs based on the element's indexes inside their parent.
* Gutted `removeNonExistingViewReferencesFromState` so now it really only removes nonexisting references and does not try to salvage paths
* Moved `applyUtopiaJSXComponentsChanges` from editor-state to project-file-utils to prevent the parser-printer-worker from bombing out (the dependency cruiser actually caught this issue, nice job, dependency cruiser!)


editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -0,0 +1,196 @@
...
+    oldParsed.topLevelElements,
+    newParsed.topLevelElements,
+    (oldElement, newElement, newPath) => {
+      const oldUID = optionalMap(getUtopiaID, oldElement)
+      const newUID = optionalMap(getUtopiaID, newElement)
+
+      if (oldUID != null && newUID != null && oldUID !== newUID) {
+        // we have a UID mismatch
+        newToOldUidMapping[newUID] = { oldUID: oldUID, newUID: newUID, pathToModify: newPath }
+      }
```
Rheeseyb, line 78: I think at this point we really need to check that we're actually matching up the elements correctly, otherwise we're going to be blinding assigning incorrect UIDs during duplications etc. which definitely risks displaying or using incorrect cached data. If we had the before and after code in scope we could probably use the Hamming distance to determine that

balazsbajorics:
url to test: https://utopia.pizza/project/6d2e4030/?branch_name=spike-uid-fix

editor/src/core/model/project-file-utils.ts
```diff
@@ -216,6 +218,37 @@ export const getUtopiaJSXComponentsFromSuccess = Utils.memoize(
   getUtopiaJSXComponentsFromSuccessInner,
 )
 
+export function applyUtopiaJSXComponentsChanges(
```
balazsbajorics, line 221: I moved this verbatim from editor-state.ts
the reason was transitive imports killing the parser-printer-worker

editor/src/components/editor/store/editor-state.ts
```diff
@@ -515,37 +516,6 @@ export function modifyOpenParseSuccess(
   )
 }
 
-export function applyUtopiaJSXComponentsChanges(
```
balazsbajorics, line 518: this function was moved verbatim to project-file-utils.ts

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -0,0 +1,170 @@
...
+          [newTopLevelElement.rootElement],
+          onElement,
+        )
+      }
+    }
+  })
+}
+
+function walkElementChildren(
+  pathSoFar: StaticElementPath,
```
balazsbajorics, line 138: as @seanparsons observed last week, we are kind of using the TemplatePath for two things: to point to something in the file hierarchy, and to point to something in the same file. Here I'm trying to separate those concerns and instead of using TemplatePath, I'm using a `StaticElementPath` to express that this is all local to the one file we are working with. 

Also, using a TemplatePath here would require for me to provide a fake EmptyScenePath which kind of reveals why TemplatePath is not the best tool for this job

editor/src/core/workers/parser-printer/uid-fix.ts
balazsbajorics: I opted for a much simpler heuristic, (in line 55) if we found more than one mismatched UIDs, we just bail out.

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    return {
+      ...newParsed,
+      topLevelElements: fixedTopLevelElements,
+      highlightBounds: fixedHighlightBounds,
+    }
+  }
+}
+
+function zipProjectContents(
```
seanparsons, line 108:
```suggestion
function zipTopLevelElements(
```
Maybe?

balazsbajorics: makes sense

balazsbajorics: fixed

seanparsons (approved):

editor/src/core/workers/parser-printer/uid-fix.ts
Rheeseyb: Does that mean if an element is duplicated it will trip that? I think that intended behaviour makes sense.

balazsbajorics:
> This looks good, but please can you add some tests to `parser-printer-uids.spec.ts` for this

_slaps forehead_ tests! what a great idea!

balazsbajorics:
> > This looks good, but please can you add some tests to `parser-printer-uids.spec.ts` for this
> 
> _slaps forehead_ tests! what a great idea!

I added new tests :)

Rheeseyb (approved):

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -0,0 +1,191 @@
...
+   * • before correcting a UID, make sure the uids of the preceding and proceding siblings are matching! this makes the comparison more robust and makes us less exposed to off-by-one errors
+   * • try to match offsets: if the user deletes or inserts an element, all subsequent uids will be shifted and thus mismatched in an index-to-index comparison
+   */
+
+  newElements.forEach((newElement, index) => {
+    const oldElement: JSXElementChild | null = oldElements[index]
+    const newUid = getUtopiaID(newElement)
+    const path = TP.appendToElementPath(pathSoFar, newUid)
+
+    onElement(oldElement, newElement, path)
```
Rheeseyb, line 180: I think this call should be inside the below `if` statement (or at least should be wrapped in a conditional checking that the new and old elements are `JSXElement`s

balazsbajorics: fixed

balazsbajorics:
we concluded that the performance regression was not caused by this branch, so I'm merging it!

PR #1110, opened 2021-04-19
Feature/UI toggle button

enidemi:
**Problem:**
There is no common simple toggle button in the editor. The goal is to use these in different places, like in the TopMenu. Currently it can have 2 states only, later we can extend it to toggle on a list of values if needed.

**Commit Details:**
- add button component that can show icons for on/off state
- the icon changes already on active state like onmousedown, the toggling happens on onClick


Rheeseyb (approved):

balazsbajorics (commented):
👍

PR #1111, opened 2021-04-20
Fix/component UI grid row minheight

OmarDaSilva:
Slight improvement, UIGridRow has now minHeight property instead of height, also added whitespace property so that the look remains the same.


OmarDaSilva:
Ready for review

editor/src/components/inspector/inspector.tsx
```diff
@@ -373,6 +373,7 @@ export const Inspector = betterReactMemo<InspectorProps>('Inspector', (props: In
       return (
         <React.Fragment>
           <AlignmentButtons numberOfTargets={instancePaths.length} />
+          {anyComponents ? <ComponentSection isScene={false} /> : null}
```
maltenuhn, line 376: This probably shouldn't be in this PR... is there a separate PR open for it?

editor/src/components/inspector/widgets/ui-grid-row.tsx
```diff
@@ -86,7 +86,8 @@ export const UIGridRow: React.FunctionComponent<GridRowProps> = ({
     style={{
       padding: padded ? `0px ${UtopiaTheme.layout.rowHorizontalPadding}px` : undefined,
       display: 'grid',
-      height: tall ? UtopiaTheme.layout.rowHeight.max : UtopiaTheme.layout.rowHeight.normal,
+      minHeight: tall ? UtopiaTheme.layout.rowHeight.max : UtopiaTheme.layout.rowHeight.normal,
```
maltenuhn, line 89: we tested this, and it doesn't seem to cause anything to break visually

editor/src/components/inspector/widgets/ui-grid-row.tsx
```diff
@@ -86,7 +86,8 @@ export const UIGridRow: React.FunctionComponent<GridRowProps> = ({
...
+      whiteSpace: 'normal',
```
maltenuhn, line 90: NB: This is a behaviour change. It allows for longer copy in rows and will make them expand automatically.

 The only place where that's an issue is rows with extra-long labels which before would get truncated, and now can span multiple lines. As that's the exception, happy with this change for the default behaviour.

maltenuhn (approved):

editor/src/components/inspector/inspector.tsx
OmarDaSilva: This was just merged in another PR

enidemi (approved):

PR #1113, opened 2021-04-20
types(icons) - adds 'colourful', uses it

maltenuhn:
**Problem:**
The NPM logo isn't ours, and comes in One Colour Only. This creates an icon for it using the `colourful` colour name, so we can use it for other things that don't neatly fit into the monochromatic colour scheme.

**Fix:**
Added `colourful` to type, added new icon `Icons.NpmLogo`

### DIFF
--- a/editor/src/uuiui/icn.tsx
+++ b/editor/src/uuiui/icn.tsx
@@ -14,6 +14,7 @@ export type IcnColor =
   | 'purple'
   | 'red'
   | 'orange'
+  | 'colourful'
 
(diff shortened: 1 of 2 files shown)


seanparsons (approved):

balazsbajorics:
is the icon asset already merged? I don't see it in this PR.

maltenuhn:
@balazsbajorics it is indeed

balazsbajorics (approved):

PR #1114, opened 2021-04-20
Fix/natural scenes performance

balazsbajorics:
**Problem:**
The memoization of Scenes regressed in #1083 

**Fix:**
Try to resurrect the memoization of scenes

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/scene-component.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/scene-component.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import * as fastDeepEquals from 'fast-deep-equal'
 import { useContextSelector } from 'use-context-selector'
 import { Scene, SceneProps } from 'utopia-api'
 import { colorTheme, UtopiaStyles } from '../../../uuiui'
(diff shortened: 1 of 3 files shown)


balazsbajorics:
> Performance: Scroll Canvas: 14ms | Resize: 11.7ms | Selection: 257.7ms SummaryChart

okay it seems like we managed to fix Scrolling, selection is still busted. Can confirm I see the same on my machine. Looking into it

enidemi (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/scene-component.tsx
```diff
@@ -27,4 +28,35 @@ export const SceneComponent = betterReactMemo(
...
+
+type ReactChild = Exclude<React.ReactNode, boolean | null | undefined>
+
+function childUnchanged(prevChild: ReactChild, nextChild: ReactChild): boolean {
+  if (typeof prevChild === 'string' || typeof prevChild === 'number') {
+    return nextChild === prevChild
+  } else if (React.isValidElement(prevChild)) {
+    return React.isValidElement(nextChild) && prevChild.type === nextChild.type
+  } else {
+    // FIXME Fragments are all that is left
```
Rheeseyb, line 59: We found that actually fragments aren't left over here, since they get squashed in the `React.Children.toArray` call above. However, we should add something here to explicitly capture that in case that changes, as well as any other exotic elements.

Rheeseyb (approved):

PR #1120, opened 2021-04-20
fix scene-level DOM-walker caching

balazsbajorics:
**Problem:**
The new Natural Scenes worked removed the `data-utopia-scene-id` field from Scenes. This had the unintended consequence of breaking the caching in DOM-walker.

**Commit Details:**
- Restore `data-utopia-scene-id` on the scenes.
- Created a sinful `TP.createBackwardsCompatibleScenePath` function which takes the first two elements of the first element path of a scene path of a path. Long story short: for this path: `storyboard/scene-1/app-instance:app-root/card-instance` it will return `storyboard/scene-1`. This is horrible, and really bad, but the alternative was to change the keying of the dom walker caching which has zero test coverage, and I will probably just break it if I touch it.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
@@ -161,9 +161,7 @@ describe('Spy Wrapper Template Path Tests', () => {
     expect(sanitizedDomMetadata).toMatchInlineSnapshot(`
       Object {
         ":storyboard": Object {
-          "children": Array [
-            ":storyboard/scene",
-          ],
+          "children": Array [],
           "name": "Storyboard",
           "rootElements": Array [],
         },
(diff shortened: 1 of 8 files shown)


enidemi (approved):

balazsbajorics:
this looks to significantly improve the selection tests, but it's still not as good as last week.

scroll canvas is also slightly worse than resize, so it needs more attention.

there's a lump of resize performance points around the 116ms mark, I think that is also something worthy of looking into.

maltenuhn:
> scroll canvas is also slightly worse than resize, so it needs more attention.

Is the "scrolling is our canary" / not having a synthetic baseline performance test compounding this? If so, maybe that particular canary should meet its maker (you, IIRC)

Rheeseyb (approved):
Well this is horrifying. Approved.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
```diff
@@ -161,9 +161,7 @@ describe('Spy Wrapper Template Path Tests', () => {
     expect(sanitizedDomMetadata).toMatchInlineSnapshot(`
       Object {
         ":storyboard": Object {
-          "children": Array [
-            ":storyboard/scene",
-          ],
+          "children": Array [],
```
balazsbajorics, line 164: @Rheeseyb heads up, I think this change is "correct" and nothing depends on this behavior (this is how it was 5 days ago).

The reason we don't see children here because the dom-walker now treats scenes as scenes again, and "restarts" the walk when it encounters one.

seanparsons (approved):

PR #1121, opened 2021-04-20
Strip UIDs when printing code in production environments

### LINKED ISSUE #1119: Regression - code printing shouldn't re-introduce UIDs into the code
At some point in the recent past we started printed UIDs back into the code. We should not.

Rheeseyb:
Fixes #1119 

**Problem:**
We were no longer stripping UIDs because of a hard coded flag that was presumably left in from development.

**Fix:**
Use the environment to determine if UIDs should be stripped, meaning they will be removed in all production environments (including staging)

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -250,9 +250,8 @@ function maybeRequestModelUpdate(
         codeNeedsPrinting(file.fileContents.revisionsState) &&
         isParseSuccess(file.fileContents.parsed)
       ) {
-        const stripUIDs = PRODUCTION_ENV
         filesToUpdate.push(
-          createPrintCode(fullPath, file.fileContents.parsed, stripUIDs, file.lastRevisedTime),
+          createPrintCode(fullPath, file.fileContents.parsed, true, file.lastRevisedTime),
         )
       }
     }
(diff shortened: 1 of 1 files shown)


editor/src/components/editor/store/dispatch.tsx
```diff
@@ -250,8 +250,9 @@ function maybeRequestModelUpdate(
         codeNeedsPrinting(file.fileContents.revisionsState) &&
         isParseSuccess(file.fileContents.parsed)
       ) {
+        const stripUIDs = process.env.NODE_ENV === 'production'
```
Rheeseyb, line 253: I'm going to change this to use the `PRODUCTION_ENV` const we declare in `env-vars.ts` actually

balazsbajorics:
I think this is not a good idea! We should only print uids that were written by humans. (ie uid='scene' etc) we should strip autogenerated uids even on development. otherwise we risk not noticing bugs we introduce to the system that only come up if the uids are not printed. This behavior should not change depending on process.env!!

I vote that we either:
1. strip all uids for production and development and staging etc
2. keep "human" uids for production and development and staging alike

Rheeseyb:
Good argument. I've changed it to always strip them for now.

balazsbajorics (approved):

Rheeseyb:
I've had to go back to the approach of only stripping UIDs in production environments, as otherwise there is no simple way to leave them in for `spec.browser` tests. This needs following up.

enidemi (approved):

seanparsons (approved):
I'm marginally terrified of this, but otherwise it's fine.

PR #1140, opened 2021-04-21
Dom-walker: move selectedView from useEditorState to props

balazsbajorics:
**Problem:**
The selected views used by the dom-walker was accessed via useEditorState which is harder to control than having it in the props.

**Fix:**
Move it to the props

**Commit Details:**
- Create a dedicated DomWalkerProps interface
- move selectedViews to props
- add selectedViews as a canvas prop

### DIFF
--- a/editor/package-lock.json
+++ b/editor/package-lock.json
@@ -5123,14 +5123,28 @@
       }
     },
     "@testing-library/react-hooks": {
-      "version": "2.0.1",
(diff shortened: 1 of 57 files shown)


seanparsons (approved):

balazsbajorics:
this is ready for review BTW, and the performance graph looks promising too

Rheeseyb (approved):

enidemi (approved):

PR #1142, opened 2021-04-22
Natural Scenes Tidying Part 1

Rheeseyb:
**Problem:**
Ridding the world of the old style scenes left over a lot of code that is either unreachable or in some cases undesirable.

**Fix:**
The first step in trying to rid us of this code is to remove explicit uses of `ScenePath`s where they shouldn't be used anymore. To tackle this I've been targeting code that forks if a path is an `InstancePath` or `ScenePath` and then attempts to use metadata if the path is an `InstancePath` - I have been replacing these forks with a use of the function `findElementByTemplatePath`, and ~have added a boolean flag to that to~ [updated] have added two versions of the function, one of which will throw an error in cases where it should absolutely not be called with a `ScenePath`.

I have also added a `throw` into the `editorDispatch` if we ever end up somehow selecting a `ScenePath`, since this should now be impossible. I've been testing and have been unable to cause any of these to throw.

### DIFF
--- a/editor/src/templates/editor-navigator.ts
+++ b/editor/src/templates/editor-navigator.ts
@@ -4,29 +4,18 @@ import { DerivedState, EditorState } from '../components/editor/store/editor-sta
 import { LocalNavigatorAction } from '../components/navigator/actions'
 import { DragSelection } from '../components/navigator/navigator-item/navigator-item-dnd-container'
 import * as TP from '../core/shared/template-path'
-import { ElementInstanceMetadataMap } from '../core/shared/element-template'
 import Utils from '../utils/utils'
-import { MetadataUtils } from '../core/model/element-metadata-utils'
 
 export function createDragSelections(
-  components: ElementInstanceMetadataMap,
   templatePaths: TemplatePath[],
   selectedViews: TemplatePath[],
 ): Array<DragSelection> {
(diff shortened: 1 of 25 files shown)


seanparsons (approved):

Rheeseyb:
I've updated this to create two versions of the function (one that throws, one that doesn't, with clear naming on the one that doesn't throw), so that it's easier to track the remaining code paths that use `ScenePath`

editor/src/templates/editor-navigator.ts
```diff
@@ -4,29 +4,18 @@ import { DerivedState, EditorState } from '../components/editor/store/editor-sta
...
   templatePaths: TemplatePath[],
   selectedViews: TemplatePath[],
 ): Array<DragSelection> {
   let selections: Array<DragSelection> = []
   Utils.fastForEach(selectedViews, (selectedView) => {
-    const metadata = TP.isScenePath(selectedView)
-      ? null // TODO Scene Implementation?
-      : MetadataUtils.getElementByInstancePathMaybe(components, selectedView)
-    if (metadata != null) {
-      // FIXME It seems nobody really knows why we have this `null` check here, so
```
balazsbajorics, line 22: haha

balazsbajorics (commented):
👍

Issue #1134, opened 2021-04-21
Fix the preview


maltenuhn:
Preview does not render anything or show error messages

enidemi:
https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/es-modules/package-manager/built-in-dependencies.ts#L47
the `src/index.js` used by the Preview uses `ReactDOM.render` which is empty.

PR #1143, opened 2021-04-22
fix(preview) SafeReactDOM in preview

enidemi:
Fixes https://github.com/concrete-utopia/utopia/issues/1134

**Problem:**
ReactDOM.render was missing from the preview because it uses a safe version as the canvas.

**Fix:**
Extended `fetchNodeModules` and `getRequireFn` to have a isPreview flag

### DIFF
--- a/editor/src/core/es-modules/package-manager/built-in-dependencies.ts
+++ b/editor/src/core/es-modules/package-manager/built-in-dependencies.ts
@@ -42,17 +42,27 @@ function builtInDependency(
 
 // Prevent ReactDOM.render from running in the canvas/editor because it's
 // entirely likely to cause either havoc or just break.
-export const SafeReactDOM = {
-  ...ReactDOM,
-  render: NO_OP,
+export const SafeReactDOM = (isPreview: boolean) => {
+  if (isPreview) {
+    return ReactDOM
+  } else {
+    return {
+      ...ReactDOM,
+      render: NO_OP,
+    }
+  }
 }
 
-const BuiltInDependencies: Array<BuiltInDependency> = [
+const BuiltInDependencies = (isPreview: boolean): Array<BuiltInDependency> => [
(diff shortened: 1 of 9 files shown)


editor/src/core/es-modules/package-manager/built-in-dependencies.ts
```diff
@@ -70,18 +80,21 @@ const BuiltInDependencies: Array<BuiltInDependency> = [
   ),
 ]
 
-function findBuiltInForName(moduleName: string): BuiltInDependency | undefined {
-  return BuiltInDependencies.find((builtIn) => builtIn.moduleName === moduleName)
+function findBuiltInForName(moduleName: string, isPreview: boolean): BuiltInDependency | undefined {
```
balazsbajorics, line 83: I am a bit afraid of boolean flags like this, because where we are using it, you might miss what this means and the consequences. 

I personally like something like this much better: `mode: 'canvas' | 'preview'`. This means that when in the future we are copy pasting someone is more likely to notice that in the preview codebase this should not say 'canvas'

editor/src/core/es-modules/package-manager/fetch-packages.ts
```diff
@@ -196,8 +196,9 @@ function failError(dependency: RequestedNpmDependency): DependencyFetchError {
 export async function fetchNodeModules(
   newDeps: Array<RequestedNpmDependency>,
   shouldRetry: boolean = true,
+  mode: 'canvas' | 'preview' = 'canvas',
```
Rheeseyb, line 199: Please can you make this parameter non-optional

enidemi: fixed

Rheeseyb (approved):

balazsbajorics (approved):

PR #1150, opened 2021-04-23
Chore/dom walker caching tests

balazsbajorics:
This PR adds a first few tests for Dom-walker caching behavior.

Since the dom-walker is deeply embedded in the editor, plus it runs as a useLayoutEffect hook, I was thinking a lot about how best to test it as a black box. 

I realized that the output of the dom-walker is very clean: it is an action called `SAVE_DOM_REPORT` which is dispatched when the walk finished. 

So what I ended up doing is that I extended the `renderTestEditorWithProjectContent` function which creates our end-to-end Electron test editor, to record the dispatched actions. This will make it easy to write tests that for example click on a button and then observe the dispatched actions in the editor!

To be able to easily test the cache behavior, I extended the SAVE_DOM_REPORT action with this field: `cachedTreeRoots: Array<TemplatePath>`. We don't actually (yet) use this field in the update function, however, this makes it really easy to write tests for the behavor.

**Commit Details:**
- Added `getRecordedActions` and `clearRecordedActions` to the test editor API
- include `cachedTreeRoots: Array<TemplatePath>` in SAVE_DOM_REPORT
- fixed a _stupid_ synthetic default export bug with the `ResizeObserver` which prevented the resizeObserver from running in jest tests
- Wrote some simple test cases

**Notes:**
It seems like that the tests highlighted that the dom-walker is one frame behind the dispatched actions. To make the tests work, I had to include a non-action (dispatching a canvas scroll of x: 0; y: 0) to give the dom-walker a second reason for walking. I think this points to two things: first off, this test is a good starting point for editor performance optimalization. For any given user interaction, we can count how many actions are dispatched, and catch things like double dom-walking. Secondly, I think there's a low-ish hanging fruit for performance optimization around the update loop, which today looks kind of like this: `dispatch -> update -> canvas -> spy -> dom-walker -> dispatch -> update -> canvas -> hopefully no dom-walk`


editor/src/components/canvas/dom-walker.ts
```diff
@@ -1,6 +1,9 @@
 import * as React from 'react'
 import * as R from 'ramda'
 import { sides } from 'utopia-api'
+import * as ResizeObserverSyntheticDefault from 'resize-observer-polyfill'
+const ResizeObserver = ResizeObserverSyntheticDefault.default ?? ResizeObserverSyntheticDefault
```
seanparsons, line 5: Oh this burns my eyes.

balazsbajorics: I think the only solution is to cave in and enable synthetic default imports in our tsconfig.json

seanparsons (approved):

enidemi (approved):

editor/src/components/canvas/dom-walker.ts
Rheeseyb: Is there a reason why we haven't enabled it? I actually don't recall

Rheeseyb (commented):
:+1:

PR #1154, opened 2021-04-26
Fix Template Path isDescendantOf check

Rheeseyb:
**Problem:**
When checking if one `TemplatePath` was a descendant of another, we were incorrectly defaulting to also return true if the paths matched. This manifested it in the DOM walker returning cached data that was different to the previous result because it would include duplicated results, triggering a re-render. This is partially responsible for the issue Balazs discovered in #1151 

**Fix:**
Change `isDescendantOf` to return false if the paths match, since it is currently only used in the DOM walker.

### DIFF
--- a/editor/src/core/shared/template-path.spec.ts
+++ b/editor/src/core/shared/template-path.spec.ts
@@ -37,6 +37,56 @@ describe('serialization', function () {
   })
 })
 
+describe('isDescendantOf', () => {
+  it('returns true if it is a descendant from the same instance', () => {
+    const result = TP.isDescendantOf(
+      TP.instancePath(TP.scenePath([[BakedInStoryboardUID, 'scene-aaa']]), ['X', 'Y']),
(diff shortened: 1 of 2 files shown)


enidemi (approved):

editor/src/core/shared/template-path.spec.ts
```diff
@@ -37,6 +37,56 @@ describe('serialization', function () {
...
+
+  it('returns false if not a descendant', () => {
+    const result = TP.isDescendantOf(
+      TP.instancePath(TP.scenePath([[BakedInStoryboardUID, 'scene-aaa']]), ['X']),
+      TP.instancePath(TP.scenePath([[BakedInStoryboardUID, 'scene-aaa']]), ['Y']),
+    )
+    chaiExpect(result).to.be.false
+  })
+
+  it('returns false if it is a parent of the target', () => {
```
balazsbajorics, line 81:
```suggestion
  it('returns false if it is a child of the target', () => {
```

balazsbajorics (approved):

balazsbajorics:
I'm merging this

PR #1157, opened 2021-04-26
Remove some of the single file functions.

seanparsons:
**Problem:**
There was some old functions like `getOpenUtopiaComponentsFromState`, which doesn't handle nested components from other files at all.

**Fix:**
The above function and `getOpenImportsFromState` have been removed altogether and multi-file compatible alternatives have been used. In the process of making these changes the `getAllUniqueUids` logic was also replaced to have it walk the project contents and not just look at one file.

**Commit Details:**
- Removed `getOpenUtopiaComponentsFromState` and
  `getOpenImportsFromState` altogether.
- Renamed `forUnderlyingTarget` to `forUnderlyingTargetFromEditorState`
  and added added a new `forUnderlyingTarget` that take the fields
  individually.
- Added `walkContentsTreeForParseSuccess`.
- Numerous functions to use the new utility functions and often bringing
  in `ProjectContentsTreeRoot instead of just an array of components.
- Moved `getTargetParentForPaste` to `clipboard.ts` as that makes a little
  more sense it being in there.

### DIFF
--- a/editor/src/components/canvas/controls/yoga-control.tsx
+++ b/editor/src/components/canvas/controls/yoga-control.tsx
@@ -3,7 +3,11 @@ import { FlexStretch, Sides } from 'utopia-api'
 import { LayoutHelpers } from '../../../core/layout/layout-helpers'
 import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import { ElementInstanceMetadata } from '../../../core/shared/element-template'
-import { InstancePath } from '../../../core/shared/project-file-types'
+import {
+  ElementOriginType,
+  InstancePath,
+  isUnknownOrGeneratedElement,
+} from '../../../core/shared/project-file-types'
 import { defaultEither, mapEither } from '../../../core/shared/either'
 import Utils from '../../../utils/utils'
 import { CanvasRectangle, canvasRectangle } from '../../../core/shared/math-utils'
@@ -14,6 +18,11 @@ import { getOriginalFrames } from '../canvas-utils'
 import { ControlProps } from './new-canvas-controls'
 import { getSelectionColor } from './outline-control'
 import { ResizeRectangle } from './size-box'
+import {
+  getJSXComponentsAndImportsForPathInnerComponent,
+  withUnderlyingTarget,
+} from '../../editor/store/editor-state'
+import { getUtopiaJSXComponentsFromSuccess } from '../../../core/model/project-file-utils'
 interface YogaResizeControlProps extends ControlProps {
   targetElement: ElementInstanceMetadata
   target: InstancePath
@@ -25,7 +34,7 @@ class YogaResizeControl extends React.Component<YogaResizeControlProps> {
   getTargetStretch = (): FlexStretch => {
     const target = this.props.targetElement
     const parentPath = TP.parentPath(this.props.target)
-    const sceneMetadataOrElementMetadata = MetadataUtils.findElementByTemplatePath(
+    const sceneMetadataOrElementMetadata = MetadataUtils.findElementByTemplatePathDontThrowOnScenes(
       this.props.componentMetadata,
       parentPath,
     )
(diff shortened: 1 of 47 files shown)


editor/src/components/canvas/controls/yoga-control.tsx
```diff
@@ -111,21 +128,45 @@ export class YogaControls extends React.Component<YogaControlsProps> {
         this.props.componentMetadata,
       )
     })
-    const unknownElementsSelected = MetadataUtils.anyUnknownOrGeneratedElements(
```
balazsbajorics, line 114: I see this function is used from two places, wouldn't it be better to keep `MetadataUtils.anyUnknownOrGeneratedElements` alive and update it with the below code?

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
```diff
@@ -304,36 +304,16 @@ describe('Spy Wrapper Template Path Tests', () => {
...
           "name": "div",
           "rootElements": Array [],
         },
-        "storyboard/scene/app:app-root/inner-div/card-instance": Object {
+        "storyboard/scene/app:app-root/inner-div/other-card-instance": Object {
           "children": Array [],
           "name": "Card",
           "rootElements": Array [],
         },
-        "storyboard/scene/app:app-root/inner-div/card-instance:button-instance": Object {
```
balazsbajorics, line 317: 🚨  this is definitely not right, there's no reason why button-instance and the hi-elements would be gone from here

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
```diff
@@ -48,7 +48,7 @@ export var App = (props) => {
       }}
     >
       <div data-uid="inner-div">
-        <Card data-uid="card-instance" />
+        <Card data-uid="other-card-instance" />
```
balazsbajorics, line 51: why the rename?

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
Rheeseyb: This is missing because the focused element path no longer matches after the uid change

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
seanparsons: Because it clashes with the uid in one of the other files (`/src/app.js` from memory), so the LOAD action fails.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
seanparsons: Right, I'll have another look at this.

editor/src/components/canvas/controls/yoga-control.tsx
seanparsons: Yeah, I was a bit dubious of pulling the project contents types and the like into `MetadataUtils`, but there appear to be no mild explosions at doing that.

seanparsons: Fixed in 70316f2.

balazsbajorics (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
balazsbajorics: I approved the PR assuming this will be fixed :)

seanparsons: Fixed up the snapshots in df568e48ce6a3b848d37987eff57315e291b70b6.

balazsbajorics (approved):

PR #1160, opened 2021-04-27
Further tidying after natural scenes work

Rheeseyb:
Most of this is focused on the systematic removal of `ScenePath` from the code base, with a huge percentage of the changes being updated tests.

- Changed `focusedElementPath` to be of type `TemplatePath | null`, and removed a lot of now redundant uses of it
- Removed more `isScenePath` uses, as well as uses of the `*DontThrowOnScenes` variant functions introduced in the previous PR
- Removed some dragging code that gave special behaviour to old style scenes (where dragging the scene would previously update the root view of the component it rendered)
- Removed now unused actions and functions for modifying or deleting scenes (these were firmly rooted in the old world view)
- Added a new function to `template-path.ts` called `appendNewElementPath`, which adds a new part to the element path (so e.g. `appendNewElementPath([['a','b','c']], ['d'])` would return `[['a','b','c'],['d']]`). There is a use of this during DOM walking when moving from an instance to its root view.
- Also in `template-path.ts` I have added `parentTemplatePath` which will always return the parent path as an `InstancePath` (the plan is to replace the existing `parentPath` with this functionality)


editor/src/core/model/utopia-constants.ts
```diff
@@ -5,7 +5,7 @@ export const UTOPIA_DO_NOT_TRAVERSE_KEY = 'data-utopia-do-not-traverse'
 export const UTOPIA_SCENE_ID_KEY = 'data-utopia-scene-id'
 export const UTOPIA_UID_PARENTS_KEY = 'data-utopia-parents'
 export const UTOPIA_UID_ORIGINAL_PARENTS_KEY = 'data-utopia-original-parents'
-export const UTOPIA_SCENE_PATH = 'data-utopia-scene-path'
+export const UTOPIA_INSTANCE_PATH = 'data-utopia-instance-path'
```
balazsbajorics, line 8: should / could this be unified with the `data-paths` key?

editor/src/core/shared/template-path.ts
```diff
@@ -394,7 +398,18 @@ export function fromString(path: string): TemplatePath {
   }
 }
 
-function allElementPaths(path: ElementPath): Array<ElementPath> {
+function allElementPaths(fullPath: ElementPath[]): Array<ElementPath[]> {
```
balazsbajorics, line 401:
```suggestion
function allElementPaths(fullPath: Array<ElementPath>): Array<Array<ElementPath>> {
```

balazsbajorics (approved):

editor/src/core/model/utopia-constants.ts
Rheeseyb: Good question. I'm not sure tbh. This exists purely for the DOM walker caching. I think it probably should be unified but I wouldn't want to tackle that as part of this PR

balazsbajorics: ah okay, makes sense, thanks :)

editor/src/core/shared/template-path.ts
Rheeseyb: Will change as part of the next one

Rheeseyb:
Heads up, all of the changes to tests that introduce a colon at the start of printed template paths are actually going to be reversed immediately in the next PR. That was just part of the detour to remove ScenePaths from the code, and is reversed by a change to template path serialising / deserialising, but I didn't want to further bloat this PR as that comes with other changes too.

seanparsons (approved):

Issue #1128, opened 2021-04-20
[BUG] Highlight bounds aren't correctly captured for all top level elements of this project


Rheeseyb:
In the Majestic Broker project, for some reason we only capture highlight bounds for the main `App` component, meaning that selection following is broken for anything else.

Project: https://utopia.pizza/project/0501e894/

PR #1161, opened 2021-04-27
fix(parser) fix highlightbounds

enidemi:
Fixes https://github.com/concrete-utopia/utopia/issues/1128

**Problem:**
Code following only works for one component in the majestic broker example project. Highlightbounds are missing for some elements from the parsed project contents.
In the parser only the last parsed element's highlighbounds are used. Instead of capturing only the last element it is now extending the current highlightbounds map.


Rheeseyb (approved):

balazsbajorics (approved):

balazsbajorics:
could you write a test for this @enidemi ?

PR #1162, opened 2021-04-27
Prevent stack overflow on insert from same file.

### LINKED ISSUE #1082: [Bug] Module resolution causes a stack overflow when attempting to insert a component from the same file
STR:
1. Paste the following code into your project:
```
/** @jsx jsx */
import { Scene, Storyboard, jsx } from 'utopia-api'
import { View } from 'utopia-api'

export const Card = (props) => {
  return <div data-uid='card-element-root'>hi!</div>
}
export var App = (props) => {
  return (
    <div
      data-uid='app-root'
      style={{
        width: '100%',
        height: '100%',
        backgroundColor: '#FFFFFF',
        position: 'relative'
      }}
    />
  )
}

export var storyboard = (
  <Storyboard data-uid='storyboard'>
    <Scene
      data-uid='scene'
      component={App}
      style={{ position: 'absolute', left: 0, top: 0, width: 375, height: 812 }}
    />
  </Storyboard>
)
```
2. From the Insert menu on the right, select the "Card" component
3. Start dragging to insert the Card as a child of the App

You should see the following stack overflow error:
```
react_devtools_backend.js:2556 RangeError: Maximum call stack size exceeded
    at eval (module-resolution.ts:126)
    at eval (module-resolution.ts:187)
    at failoverResolveResults (module-resolution.ts:32)
    at resolveNonRelativeModule (module-resolution.ts:186)
    at resolveModule (module-resolution.ts:223)
    at Object.resolveModulePath (module-resolution.ts:229)
    at eval (package-manager.ts:21)
    at eval (ui-jsx-canvas.tsx:116)
    at eval (npm-dependency.ts:310)
    at Object.fastForEach (utils.ts:19)
```

seanparsons:
Fixes #1082

**Problem:**
When attempting to insert something from the same file, a stack overflow was triggered pretty much instantly as the mouse drag started.

**Fix:**
When constructing the list of components to export the short file path was used for the originating path instead of the absolute path (`storyboard.js` versus `/utopia/storyboard.js`), which resulted in an import for the current file being included. As a result the canvas require function would stack overflow when attempting to resolve that import.

**Notes:**
I believe there's still some an underlying issue where the import/package resolution can fail in the same way if an import like this has been written in by hand.

**Commit Details:**
- Use the full path when constructing the imports to add to
  a file from the insert menu so that the check for same file
  imports correctly functions.
- Removed a field being constructed for the insert menu props
  as it was not being referenced or used.

### DIFF
--- a/editor/src/components/editor/insertmenu.tsx
+++ b/editor/src/components/editor/insertmenu.tsx
@@ -83,12 +83,6 @@ import {
 } from '../shared/project-components'
 import { ProjectContentTreeRoot } from '../assets'
 
-interface CurrentFileComponent {
-  componentName: string
-  defaultProps: { [prop: string]: unknown }
-  detectedProps: Array<string>
-}
-
 interface InsertMenuProps {
   lastFontSettings: FontSettings | null
   editorDispatch: EditorDispatch
@@ -96,7 +90,6 @@ interface InsertMenuProps {
   mode: Mode
   existingUIDs: Array<string>
   currentlyOpenFilename: string | null
-  currentFileComponents: Array<CurrentFileComponent>
   dependencies: Array<PossiblyUnversionedNpmDependency>
   packageStatus: PackageStatusMap
   propertyControlsInfo: PropertyControlsInfo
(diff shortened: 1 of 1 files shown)


enidemi (approved):

editor/src/components/editor/insertmenu.tsx
```diff
@@ -106,43 +99,15 @@ interface InsertMenuProps {
...
-            const detectedProps = topLevelElement.propsUsed
-            currentFileComponents.push({
-              componentName: componentName,
-              defaultProps: defaultProps,
-              detectedProps: detectedProps,
-            })
-          }
-        })
-      }, openUIJSFile.fileContents.parsed)
-    }
```
Rheeseyb, line 136: Don't delete this code please. I hadn't realised we had regressed and lost the default and detected props used for insertion in #969. Looks like it shouldn't be too tricky to restore that.

seanparsons: I'll put a comment on that issue specifically mentioning this code snippet, because the change I made for this fix affects this code as well so I'll be modifying dead code that is never (currently) being run or tested. Then whenever someone works on that issue they can see what was there at one point.

Rheeseyb: Yep ok works for me

Rheeseyb (approved):

balazsbajorics (approved):

Issue #1156, opened 2021-04-26
Pragma and missing jsx import result in silent failures


maltenuhn:
**Problem**
- [ ] if the user forgets to add `import { jsx } from 'utopia-api'` there are no elements on the canvas and we show no error message
- [ ] kill the pragma, check what still breaks 🔥
- [ ] if there’s a pragma but no imported jsx, we shouldn’t fail silently
- [ ] check that we can still use emotion with a pragma, this notwithstanding

enidemi:
I put this back to TODO, because now I always see the runtime error screen correctly.

Issue #996, opened 2021-03-22
Remove `Positionable` etc utopia-API wrappers, and other places in our codebase that use / rely on them


maltenuhn:
**Problem** `Positionable` was a concept introduced in the olden days, to circumvent the problem of immovable / non-style-or-classname-supporting 3pty components. However, it hasn't worked in a while, is inaccessible, and in the New World _full of_ vanilla React code and 3pty imports, it no longer is an acceptable work-around. Ergo, we should remove it

**Solution**

Terminate with prejudice, including its `utopia-api` alter-ego, any wrapping actions that make use of it, and whatever tests we had that tested this.

maltenuhn:
@enidemi wasn't this something you were tackling as part of another ticket? Free bonus points 😁

enidemi:
yes as part of https://github.com/concrete-utopia/utopia/issues/1156 I'm gonna remove the unused layout components too

PR #1166, opened 2021-04-28
Feature/utopia api pragma removal

enidemi:
Fixes https://github.com/concrete-utopia/utopia/issues/1156.
Fixes https://github.com/concrete-utopia/utopia/issues/996.

**Problem:**
Utopia-api has lots of unused code. One of that is the pragma that was added to support layout props on regular divs. Since we don't use the layout props anymore it should be removed. There are also 3 layout related components: Positionable, Layoutable and Resizable, these also used the layout props to position themselves.

**Fix:**
Removing the utopia-api pragma and the unused components. After this emotion/styled components can be still added manually to projects by the user to support css props.

**Commit Details:**
- updating lots of tests
- removed actions that used the layout wrappers


balazsbajorics:
seems like the tests break because of a merge conflict

balazsbajorics:
@enidemi I merged the conflicts on this branch, please check my commit to make sure I didn't mess it up :)

editor/src/components/custom-code/__snapshots__/code-file.spec.ts.snap
```diff
@@ -1546,11 +1654,47 @@ export var App = (props) => {
       },
       "transpiledCode": "\\"use strict\\";
 
+var __createBinding = this && this.__createBinding || (Object.create ? function (o, m, k, k2) {
```
balazsbajorics, line 1657: where do these extra autogenerated functions come from?

balazsbajorics: are these because the code now says `import * as React from 'react'` and for that it needs the importStar function?

balazsbajorics (approved):

seanparsons (approved):

PR #1168, opened 2021-04-28
Fix VSCode not loading if the editor tab is in the background

### LINKED ISSUE #1077: [Bug] VS Code doesn't correctly load when the tab is moved to the background whilst loading the editor
STR:
1) Start loading the browser in a new tab
2) Immediately switch to a different tab before the editor has loaded
3) Wait whilst checking /r/blursedimages for a bit
4) Eventually (having waited long enough for the editor to completely load) switch back to the editor tab

If you have waited long enough the code editor pane will be completely empty, and you'll see the following errors in the console (taken from pizza):

```
Uncaught (in promise) Error: Unable to figure out browser width and height
    at Object.W [as getClientArea] (workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:66)
    at O.getClientArea (workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:2269)
    at O.createGridDescriptor (workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:2269)
    at O.createWorkbenchLayout (workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:2269)
    at workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:2269
    at C.invokeFunction (workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:612)
    at O.startup (workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:2269)
    at pe.open (workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:2269)
```

```
workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:2269 Uncaught TypeError: Cannot read property 'layout' of undefined
    at O.layout (workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:2269)
    at n.onWindowResize (workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:1540)
    at workbench.web.api.js?hash=3e167302a580412916685998772938f79d1de493:1540
```

balazsbajorics:
Fixes #1077 

**Problem:**
It takes 3 years to load Utopia, which encourages users and testers to switch to another tab. By the time they switch back, Utopia is fully loaded – except for VSCode which silently crashes in the background. The problem is a particular line of code that throws an error if it cannot get a window size other than 0 by 0.

**Fix:**
I removed the error and simply return a Dimension(0, 0). It seems to be not causing any problems, I suspect switching to the front fires event listeners that re-adjust the size to a now-measurable one.

**Commit Details:**
- Removed error throwing, added a comment explaining my reasons
- Return new Dimension(0, 0)

**Notes:**
The bug does not reproduce on localhost, so I had to do horrible things with path mappings to trick chrome into belieiving the local dev env is an actual website.

### DIFF
--- a/vscode-build/vscode.patch
+++ b/vscode-build/vscode.patch
@@ -1,5 +1,5 @@
 diff --git a/build/gulpfile.vscode.js b/build/gulpfile.vscode.js
-index 6d3a369082..7ad3222305 100644
+index 6d3a369082a..7ad32223052 100644
 --- a/build/gulpfile.vscode.js
 +++ b/build/gulpfile.vscode.js
 @@ -35,11 +35,12 @@ const { compileExtensionsBuildTask } = require('./gulpfile.extensions');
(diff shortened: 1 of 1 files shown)


vscode-build/vscode.patch
```diff
@@ -982,7 +999,7 @@ index 1fe5aa621c..18d257fab3 100644
  
 diff --git a/src/vs/workbench/browser/parts/dummies/activitybarPart.ts b/src/vs/workbench/browser/parts/dummies/activitybarPart.ts
 new file mode 100644
-index 0000000000..455a50bb45
```
seanparsons, line 985: Why did all these other files change?

balazsbajorics: it seems like they did not... only their hash became one charger longer...

enidemi (approved):

Rheeseyb (approved):

PR #1170, opened 2021-04-28
Use setInterval instead of setTimeout.

### LINKED ISSUE #1070: FS Utils `polledWatch` is a recursive function that constantly pushes to the stack
Offending line here: https://github.com/concrete-utopia/utopia/blob/master/utopia-vscode-common/src/fs/fs-utils.ts#L407

The `polledWatch` function should not be pushing itself to onto the stack after every invocation. Whoops! This one was definitely me...

seanparsons:
Fixes #1070

**Problem:**
The way `polledWatch` is implemented appears to be creating a near infinite chain of async calls by the use of `setTimeout`.

**Fix:**
Switch the starting call to using `setInterval`.

**Commit Details:**
- Use `setInterval` in the `watch` function, negating the
  need for repeatedly calling `setTimeout`.

### DIFF
--- a/utopia-vscode-common/src/fs/fs-utils.ts
+++ b/utopia-vscode-common/src/fs/fs-utils.ts
@@ -404,7 +404,6 @@ async function polledWatch(): Promise<void> {
   })
 
   await Promise.all(promises)
-  watchTimeout = setTimeout(polledWatch, POLLING_TIMEOUT) as any
 }
 
 export async function watch(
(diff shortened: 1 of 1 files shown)


enidemi (approved):

balazsbajorics (approved):

balazsbajorics:
I think setTimeout "resets" the stack trace as it adds an event listener to a timer event. but nevertheless this is cleaner!

Rheeseyb (approved):

PR #1176, opened 2021-04-29
Feature/utopia api layout prop

enidemi:
**Problem:**
Layout prop is not used anymore and it's still part of the utopia-api code. When we moved layout to style props old tests were not updated because Views still supported the layout prop, so now lots of tests are changed.

**Commit Details:**
- removed layoutsystem computing from utopia-api
- found some jsx imports in some places, removed those
- changed layout props to style props in tests
- updated lots of snapshots


editor/src/components/canvas/canvas-utils.spec.browser.ts
```diff
@@ -617,82 +553,16 @@ describe('updateFramesOfScenesAndComponents - pinMoveChange -', () => {
...
-    } as CanvasVector)
-
-    await renderResult.dispatch([setCanvasFrames([pinChange], false)], true)
-
-    expect(getPrintedUiJsCode(renderResult.getEditorState())).toEqual(
-      makeTestProjectCodeWithSnippet(
-        `<View style={{ ...props.style }} layout={{ layoutSystem: 'pinSystem' }} data-uid='aaa'>
-        <View
-          style={{ backgroundColor: '#0091FFAA', left: 20, top: 20 }}
-          layout={{ layoutSystem: 'pinSystem', centerX: 68, centerY: 59 }}
```
balazsbajorics, line 662: WOW, I thought that this was killed a _long_ time ago! Apparently this is still kicking on master? @enidemi while you are spring cleaning, could you murder the code responsible for updating centerX and centerY values? :D

Rheeseyb (approved):
I see merge conflict hell in my very near future :grimacing:

editor/src/components/canvas/canvas-utils.spec.browser.ts
enidemi: Because Views supported layout props, centerX and centerY worked when you just typed it to the code editor. There are some TODOs left in the code to remove layoutsystem prop, it should be the next step.

balazsbajorics: yeah but apparently resizing properly updates centerX and centerY! that's a good chunk of code making the resize logic very complicated, I think it'd be worth to make a follow-up PR and delete half of that

enidemi: 👍

utopia-api/src/primitives/common.tsx
```diff
@@ -1,254 +1,11 @@
...
-        return {
-          ...flexElementPropsToStyle(layout, {}, 0, 1), // TODO instead of flexElementPropsToStyle, make a function that moves only the non-magic props to style
-          ...flexParentPropsToStyle(layout),
-          ...convertPinsToStyleProps(layout as any),
-        }
-    }
-  }
-}
-
 export function addEventHandlersToDivProps(
```
balazsbajorics, line 9: wow View does _this_? I think at this point this is more likely to just catch out someone rather than help them?

enidemi:
There is a real regression in the opacity performance test. It was still using the layout prop and changing it to style suddenly doubled the rendercount which happens even on master without the utopia-api changes. This is an unfortunate consequence of not cleaning up the tests when the layout->style shift happened.

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -50,33 +49,30 @@ describe('React Render Count Tests - ', () => {
...
           data-uid='bbb'
         />
       </View>`,
       ),
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(310) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(320)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(670) // if this breaks, GREAT NEWS but update the test please :)
```
seanparsons, line 62: It more than doubles?

seanparsons (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
Rheeseyb: Eni commented above that this is a real regression already on master - https://github.com/concrete-utopia/utopia/pull/1176#issuecomment-829939128

seanparsons: Aha!

PR #1177, opened 2021-04-29
Natural Scenes Tidying Part 3

Rheeseyb:
This is yet more removal of `ScenePath` and `InstancePath`, replacing them with `TemplatePath` everywhere.

The core changes are around the introduction of a `templatePath` factory function, which is now being used in place of the previous `scenePath`, `instancePath`, `staticScenePath` and `staticInstancePath` everywhere outside of `template-path.ts` (with the exception of `template-path.test-utils.ts`. Along with that, the serialising and deserialising functions have been updated to reflect the upcoming change, so you'll annoyingly see that the colon prefixes on printed single length paths (that were introduced in my previous PR) are gone again (they'll never ever ever return!). Also, `appendToPath` and `parentPath` in `template-path.ts` have been updated to reflect the upcoming changes and remove special casing around `ScenePath`s vs `InstancePath`s.

The vast vast majority of other changes were swapping `InstancePath` with `TemplatePath`. The next step after this will be the complete removal of the `ScenePath` and `InstancePath` types.


editor/src/components/canvas/canvas-utils.spec.browser.ts
```diff
@@ -45,7 +45,7 @@ describe('updateFramesOfScenesAndComponents - pinFrameChange -', () => {
     )
 
     const pinChange = pinFrameChange(
-      TP.instancePath(TestScenePath, ['aaa', 'bbb']),
+      TP.appendNewElementPath(TestScenePath, ['aaa', 'bbb']),
```
balazsbajorics, line 48: hah, this is a good trick that `TP.appendNewElementPath` has the same signature as instancePath :)

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1940,26 +1922,26 @@ export function moveTemplate(
...
 
               const newHighlightedViews =
                 newParentPath == null
-                  ? highlightedViews.map((t) => (TP.pathsEqual(t, target) ? newInstancePath : t))
+                  ? highlightedViews.map((t) => (TP.pathsEqual(t, target) ? newPath : t))
                   : [newParentPath]
 
               const updatedEditorState: EditorState = {
                 ...workingEditorState,
-                selectedViews: filterMultiSelectScenes(Utils.stripNulls(newSelectedViews)),
```
balazsbajorics, line 1956: is this a change in behavior? I don't remember why we filtered multiselected scenes: was it because there was a technical problem with them, or to prevent the user from doing something potentially weird? I _think_ this was just a technicality but I don't remember

Rheeseyb: It was an old world technicality based on the fact that interactions made whilst a scene was selected would either change the root view of the component rendered by the scene, or would change the json object that used to be at the top of the old `ui.jsx` file :D

editor/src/components/inspector/common/css-utils.spec.browser.ts
```diff
@@ -26,7 +27,7 @@ describe('toggle style prop', () => {
     await renderResult.dispatch(
       [
         toggleProperty(
-          TP.instancePath(TestScenePath, ['aaa', 'bbb']),
+          TP.appendNewElementPath(TestScenePath, ['aaa', 'bbb']) as InstancePath,
```
balazsbajorics, line 30: TODO for next PR?

Rheeseyb: Was done about 10 mins ago (for next PR) :D

editor/src/components/custom-code/code-file.ts
```diff
@@ -386,7 +387,8 @@ export function normalisePathToUnderlyingTarget(
       // which now doesn't exist.
       return normalisePathUnableToProceed(currentFilePath)
     } else {
-      const staticPath = elementPath == null ? null : TP.dynamicPathToStaticPath(elementPath)
+      const staticPath =
+        elementPath == null ? null : (TP.dynamicPathToStaticPath(elementPath) as StaticInstancePath)
```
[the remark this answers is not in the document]
Rheeseyb: With this PR, all template paths are actually instance paths.

balazsbajorics (approved):

seanparsons (approved):

Issue #1171, opened 2021-04-28
Missing Emotion error / Instead get Utopia Error: ScenePath is not provided for Card


maltenuhn:
Here's another (broken) project. It should fail because the styledComponent is using a component that's defined after it is declared in the file. 

Codesandbox [here](https://codesandbox.io/s/suspicious-tharp-i209p?file=/src/App.js) shows:
```
You are trying to create a styled element with an undefined component.
You may have forgotten to import it.
```

Utopia [here](https://utopia.pizza/p/5df4e6cd-successful-plain/) instead shows:
```
Error: Utopia Error: ScenePath is not provided for Card
```

balazsbajorics:
the linked utopia project contains another nice new bug: it fails to load because of a duplicate UID issue. the original behavior was to auto-fix uid issues but I think the recent changes moving the validation to multi-file broke that behavior . I'm gonna make this a separate ticket because it leads to perceived data loss by the user!!!

maltenuhn:
The sample project still won't load for me. I'm getting a "No File Open" message
![image](https://user-images.githubusercontent.com/2945037/118108344-cb97dd00-b3d7-11eb-8322-e38972a8fdb3.png)

When I click on the left menu, I get
![image](https://user-images.githubusercontent.com/2945037/118108398-dbafbc80-b3d7-11eb-92cb-5b4ae616550d.png)

Should I log a new ticket for this @balazsbajorics  ?

balazsbajorics:
Here's a reproduction project:

https://utopia.pizza/p/4d33c60f-successful-plain/

and the same in codesandbox with the "proper" error message: https://codesandbox.io/s/gracious-lovelace-nr0in?file=/src/card.js

balazsbajorics:
@maltenuhn this bug hits a Utopia shortcoming: we are not evaluating the real js file in a single pass, instead we slice it up to chunks of "arbitrary js" (which we eventually evaluate) and "parsed" parts (which we **never** evaluate, instead we essentially simulate what would React do when evaluating it). The ui-jsx-canvas code spaghetti then takes these arbitrary parts, and the parsed parts, and tries to create a frankenstein's javascript monster which more or less looks like what the original file would produce. the goal of course is to match native javascript 100% both visually and react-behaviorally.

I can see two options to mitigate/solve this problem:
[...]

PR #1178, opened 2021-04-29
Fix higher order components wrapping user-written components

balazsbajorics:
Fixes parts of #1171

**Problem:**
When wrapping Card into `emotion.styled(Card)()`, the ComponentRendererComponent would throw a `Utopia Error: Instance Path is not provided` message.

**Fix:**
The ComponentRendererComponent tries to fall back to using `data-paths` if it does not find the `data-utopia-instance-path` it is looking for.

**Commit Details:**
- Created `tryToGetInstancePath` in ComponentRendererComponent
- the storyboard root now gets an empty instance path instead of an empty template path


enidemi (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -457,7 +458,7 @@ function useGetStoryboardRoot(
   StoryboardRootComponent: ComponentRendererComponent | undefined
   storyboardRootElementPath: InstancePath
   rootValidPaths: Array<InstancePath>
-  rootInstancePath: TemplatePath
+  rootInstancePath: InstancePath
```
Rheeseyb, line 461:
```suggestion
  rootInstancePath: TemplatePath
```

Please do not undo my work :sweat_smile:

Rheeseyb (commented):
Please can you hold off on merging this until #1177 has been merged, as it's introducing more `InstancePath`s that can instead be `TemplatePath`s

seanparsons (approved):

PR #1180, opened 2021-04-30
Unified Template Path

Rheeseyb:
**Problem:**
The separate `ScenePath` and `InstancePath` types no longer make sense with the introduction of natural scenes and the ability to drill into instances.

**Fix:**
Replace them with a unified type that represents a path to an element as an array of arrays of uids

**Commit Details:**
- Updated the types in `project-file-types.ts`
- Updated or removed all functions in `template-path.ts` that relied on one or the other of the above to work on all template paths
  - I had to completely rewrite `replaceIfAncestor` and `closestSharedAncestor` since those functions didn't handle scene paths
- Removed InstacePath specific functions from `element-metadata-utils.ts`, and replaced calls to these with the template path versions created in earlier PRs
- Loads of refactoring to replace the old types with `TemplatePath`

**Note:** the mega diff in `ui-jsx-canvas.spec.tsx.snap` is a result of the metadata object's key order shifting, breaking all of the snapshots in that test. I have eyeballed the diff and it looks to be correct (the metadata is all still there from what I could tell).

I'm going to ~add extra tests for `replaceIfAncestor` and `closestSharedAncestor`~ (now done), and do some more testing on the running app to make sure nothing has broken before merging, but this PR is ready to go.


editor/src/core/shared/template-path.ts
```diff
@@ -234,139 +118,40 @@ export function templatePath(fullElementPath: ElementPath[]): TemplatePath {
...
 export function isStoryboardDescendant(path: TemplatePath): boolean {
-  const asInstancePath = instancePathForElementAtPathDontThrowOnScene(path)
-  return isEmptyScenePath(asInstancePath.scene) && !isStoryboardPath(asInstancePath)
+  return path.parts.length === 1 && path.parts[0].length > 1
 }
 
 export function isStoryboardChild(path: TemplatePath): boolean {
-  const asInstancePath = instancePathForElementAtPathDontThrowOnScene(path)
-  return isEmptyScenePath(asInstancePath.scene) && asInstancePath.element.length === 2
+  return path.parts.length === 1 && path.parts[0].length === 1
```
seanparsons, line 148: This is the same as `isStoryboardPath`, is that right?

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -524,13 +457,13 @@ export const MetadataUtils = {
     let workingFrame: CanvasRectangle = originalFrame
     // If this is held within a group, then we need to add on the frames of the parent groups.
     let ancestorPath = TP.parentPath(target)
-    while (TP.isInstancePath(ancestorPath)) {
-      const ancestorElement = MetadataUtils.getElementByInstancePathMaybe(metadata, ancestorPath)
+    while (!TP.isStoryboardChild(ancestorPath) && !TP.isEmptyPath(ancestorPath)) {
```
seanparsons, line 460: I'm ever so slightly side-eyeing this, I guess `isInstancePath` was performing something like this check by coincidence?

seanparsons (approved):

editor/src/core/shared/template-path.ts
Rheeseyb: Oops!

editor/src/core/model/element-metadata-utils.ts
Rheeseyb: Yeah exactly that (before the natural scenes work)

editor/src/core/shared/template-path.ts
Rheeseyb: Fixed (and added tests covering it).

balazsbajorics (approved):

PR #1181, opened 2021-04-30
Only change the focused element path if it has actually changed

Rheeseyb:
**Problem:**
I discovered we have a bug that means we remount the canvas (triggering dom walking with forced cache blowing etc.) whenever the user clicks the empty space in the canvas, because of a increased mount count.

**Fix:**
~Only trigger the offending dispatched action when the current selection isn't already empty~
When handling the `SET_FOCUSED_ELEMENT` action, we check if the path has actually changed, leaving the editor state unchanged if the path hasn't changed.


seanparsons (approved):

Rheeseyb:
I have updated this PR by moving the logic inside the action handler

enidemi (approved):

balazsbajorics (approved):

PR #1182, opened 2021-04-30
Errors with code should clear their entry from nodeModules.

### LINKED ISSUE #1175: Missing error for this sample project (glistening-apocalypse)
**Describe the bug**
The incorrect code doesn't produce an error on the canvas, or red squiggly underlines in the code editor.

STR:
- open https://utopia.pizza/p/ad0597ba-glistening-apocalypse/
- check the console and line 29 of app.js for the surplus `'`
- behold the canvas, which does not show an error

image

Console:
image

seanparsons:
Fixes #1175.

**Problem:**
If a change was made to the code of a dependency of the storyboard which previously compiled so that it now does not parse, the original transpiled code was held in the `nodeModules` part of the `CodeResultCache`. Which resulted in the canvas contents continuing to appear with no relevant error being shown.

**Fix:**
Modified `incorporateBuiltdResult` to clear out entries instead of just updating them, removing the entry if there's no transpiled code for it.

**Commit Details:**
- Fixes #1175.
- `incorporateBuildResult` on encountering a build result that lacks some
  transpiled code should delete the previous entry held in the
  node modules object.

### DIFF
--- a/editor/src/components/custom-code/code-file.ts
+++ b/editor/src/components/custom-code/code-file.ts
@@ -183,7 +183,9 @@ export function incorporateBuildResult(
   // Mutates nodeModules.
   fastForEach(Object.keys(buildResult), (moduleKey) => {
     const modulesFile = buildResult[moduleKey]
-    if (modulesFile.transpiledCode != null) {
+    if (modulesFile.transpiledCode == null) {
+      delete nodeModules[moduleKey]
+    } else {
       nodeModules[moduleKey] = esCodeFile(modulesFile.transpiledCode, 'NODE_MODULES', moduleKey)
     }
   })
--- a/editor/src/components/custom-code/code-file.spec.ts
+++ b/editor/src/components/custom-code/code-file.spec.ts
@@ -1,5 +1,6 @@
 import {
   generateCodeResultCache,
+  incorporateBuildResult,
   normalisePathEndsAtDependency,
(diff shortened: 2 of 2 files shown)


editor/src/components/custom-code/code-file.ts
```diff
@@ -183,7 +183,9 @@ export function incorporateBuildResult(
   // Mutates nodeModules.
   fastForEach(Object.keys(buildResult), (moduleKey) => {
     const modulesFile = buildResult[moduleKey]
-    if (modulesFile.transpiledCode != null) {
+    if (modulesFile.transpiledCode == null) {
```
Rheeseyb, line 186: Is there ever a case where the build result will only include updated files? I know we originally built the code around supporting incremental builds but don't recall if we ever had that working...

Rheeseyb (approved):

editor/src/components/custom-code/code-file.ts
seanparsons: Only include updated files? There's a `full-build` case in `generateCodeResultCache` if that's what you mean?

Rheeseyb: Ahhh ok sorry I misunderstood where this code was being called

balazsbajorics (approved):

PR #1184, opened 2021-04-30
Add a wait for the mailbox to see if it appears later.

### LINKED ISSUE #1153: Attempting to query mailbox before it exists leads to FS error
See image below. Reproduced while (but probably unrelated to being) unwillingly locked out:

image

seanparsons:
Fixes #1153.

**Problem:**
It seems like sometimes the polling of an inbox manages to start before the inbox is created.

**Fix:**
To give the inbox creation a chance to work the polling now gives it a few seconds leeway to get going before it will fail out. This is somewhat of a speculative fix given that the issue was only observed once in approximately 20 attempts and hasn't been seen previously.

**Commit Details:**
- Fixes #1153.
- Implemented `waitForPathToExist` to handle the spinning on a path
  to see if it appears.
- `pollInbox` now calls `waitForPathToExist` ahead of reading the
  directory to give the other side of the mailbox a chance to create
  it.


utopia-vscode-common/src/mailbox.ts
```diff
@@ -70,27 +70,44 @@ async function initOutbox(outboxToUse: Mailbox): Promise<void> {
...
+async function receiveMessage<T>(
+  messageName: string,
+  parseMessage: (msg: string) => T,
+): Promise<T> {
   const messagePath = pathForInboxMessage(messageName)
   const content = await readFileSavedContentAsUTF8(messagePath)
   return parseMessage(content)
 }
 
+async function waitForPathToExist(path: string, maxWaitTime: number = 5000): Promise<void> {
```
balazsbajorics, line 82: this function kind of makes me uneasy because it has no explicit return at the end, so the happy path (`Promise.resolve()`) is totally implicit.

utopia-vscode-common/src/mailbox.ts
```diff
@@ -70,27 +70,44 @@ async function initOutbox(outboxToUse: Mailbox): Promise<void> {
...
+  if (maxWaitTime >= 0) {
+    const doesItExist: boolean = await exists(path)
+    if (!doesItExist) {
+      await waitForPathToExist(path, maxWaitTime - 100)
+    }
+  } else {
+    return Promise.reject(`Waited too long for ${path} to exist.`)
```
balazsbajorics, line 89: I _think_ we have a convention to use `throw` to return a failed promise in async functions

balazsbajorics (approved):

utopia-vscode-common/src/mailbox.ts
```diff
@@ -70,27 +70,44 @@ async function initOutbox(outboxToUse: Mailbox): Promise<void> {
...
   const messagePath = pathForInboxMessage(messageName)
   const content = await readFileSavedContentAsUTF8(messagePath)
   return parseMessage(content)
 }
 
+async function waitForPathToExist(path: string, maxWaitTime: number = 5000): Promise<void> {
+  if (maxWaitTime >= 0) {
+    const doesItExist: boolean = await exists(path)
+    if (!doesItExist) {
+      await waitForPathToExist(path, maxWaitTime - 100)
```
Rheeseyb, line 86:
```suggestion
      return waitForPathToExist(path, maxWaitTime - 100)
```
wrt Balazs's comment about, this should be a return statement

utopia-vscode-common/src/mailbox.ts
Rheeseyb: We've used `Promise.reject` in quite a few other places in the `utopia-vscode-common` code. We did hit a very bizarre issue when using `throw` inside promises when first experimenting with the VSCode integration - some sequence of events was causing the `throw` to be handled outside of the promise handling code (even though debugging showed it to clearly be thrown from inside the promise correctly), leading to it being treated as an unhandled exception that would break the extension. We never did get to the bottom of that...

Rheeseyb (approved):
This looks fine though I'm very dubious of what would be causing this. Looking at the screenshot in the attached issue ticket, the issue was against a user that isn't signed in, and we seem to have a bunch of similar issues against those.

utopia-vscode-common/src/mailbox.ts
seanparsons: If it's going to potentially breach the Promise contract, I'd definitely much rather go with `Promise.reject` unless there's a strong reason otherwise.

utopia-vscode-common/src/mailbox.ts
seanparsons: *nods in agreement*

seanparsons: Fixed in e7a5df1c.

seanparsons:
> This looks fine though I'm very dubious of what would be causing this. Looking at the screenshot in the attached issue ticket, the issue was against a user that isn't signed in, and we seem to have a bunch of similar issues against those.

The one time I reproduced it was mentioned in the description above was when logged out but the other 10-ish times when logged out it was totally fine which is why I chalked it up to some kind of race condition.

PR #1188, opened 2021-05-03
fix(canvas) fix delete element shortcut

### LINKED ISSUE #1187: 🐛: Multifile editing - delete shortcut is not working
When trying to delete an element nothing happens.

enidemi:
Fixes #1187 

**Problem:**
Canvas delete shortcut sometimes keeps the element in place, trying to delete an element from a different file on the canvas results in a simple deselect. Works only on storyboard elements.

**Fix:**
Update `deleteElements` with multifile support.

**Commit Details:**
- find target filepath using the new `normalisePathToUnderlyingTarget` helper to update the correct parseSuccess in the model
- cleaned up unused code related to groups and autosizing views in `deleteElements`


balazsbajorics (approved):

Rheeseyb (commented):
We have a test for this in `editor-update.spec.ts` (testing `DELETE_SELECTED`) - can you update that to include a multifile case please?

PR #1192, opened 2021-05-04
fix for UID uniqueness across files

### LINKED ISSUE #1174: Duplicate multifile UID kills utopia load
Steps to reproduce: create a project, make sure it has duplicated uids in two separate files. manage to save it, try to reload it.

Result: https://utopia.pizza/p/5df4e6cd-successful-plain/
the LOAD action will die on uid verification, leading to perceived data loss

balazsbajorics:
Fixes #1174

**Problem:**
The parser-printer only guarantees unique UIDs for a single file. If you copy-paste code with an explicit `data-uid`, the project will get into an unloadable state, because the LOAD action will fail the `getAllUniqueUids` test.

**Fix:**
Pass in the `alreadyExistingUIDs` from the outside to the parser.

**Commit Details:**
- add new parameter `alreadyExistingUIDs_MUTABLE` to parseCode and lintAndParse
- Fixed an issue where we would rely on the highlightBounds for uniqueness check instead of the alreadyExistingUIDs
- Fixed a bug where we would not push a uid to alreadyExistingUIDs if it was written by the user and not clashing with an existing UID


balazsbajorics:
I picked this on up to be able to close #1171 

I added a new breaking test case – I believe we "just" need to make that test pass and this PR is ready to go!

enidemi (approved):

seanparsons (approved):

PR #1194, opened 2021-05-04
polling user login state

### LINKED ISSUE #186: Show logged-out users that they are logged out, even if mid-session
**Problem**
1. We currently automatically log you out ~after the initial time expires~ without clearly knowing why. This can even happen mid-session. The console shows 401 errors, but nothing in the editor UI prompts you to log in
2. This also affects tabs left open / computer went to sleep over weekend scenarios.

**Suggested solution**
1. Ideally, we don't log you out mid-session
2. If we do, we use the black bar with a link to re-authenticate, and the text "You were logged out. Click here to log in again".  Make sure the link opens in a new tab, so the work isn't lost.
3. (Make sure that new users _don't_ get that message, at least not in the first session)

balazsbajorics:
Fixes #186

**Problem:**
The editor would not tell users if they are logged out for any reason.

**Fix:**
Poll login status (I set it to every 5 seconds), and if the login status changes, dispatch an action.

**Commit Details:**
- Login status polling function
- I also wrapped the login status fetch to a try-catch to avoid an unhandled promise rejection error

### DIFF
--- a/editor/src/components/editor/action-types.ts
+++ b/editor/src/components/editor/action-types.ts
@@ -49,6 +49,7 @@ import { Notice } from '../common/notice'
 import { BuildType } from '../../core/workers/ts/ts-worker'
 import { ParseResult } from '../../utils/value-parser-utils'
 import { UtopiaVSCodeConfig } from 'utopia-vscode-common'
+import type { LoginState } from '../../common/user'
 export { isLoggedIn, loggedInUser, LoginState, notLoggedIn, UserDetails } from '../../common/user'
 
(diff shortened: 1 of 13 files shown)


balazsbajorics:
@maltenuhn I did not make a new black status bar for logout events, the editor will just start showing the blue "click here to log in" bar at the top.

Rheeseyb (approved):

balazsbajorics:
I'm closing it for now because I started doing modifications so we can show better error messages too

PR #1195, opened 2021-05-04
Fix/canvas shortcut copy elements

### LINKED ISSUE #1189: Cut (of "cut and paste") is no longer working
STR
- Insert View
- Right-click on it
- Cut

Observed: parent element gets selected (correct), but cut element persists

Expected: it, well, cuts.

enidemi:
Fixes #1189

**Problem:**
Cut/copy/paste doesn't work on non-storyboard elements. The element copied to the clipboard is a `null`.

**Fix:**
Updating `createClipboardDataFromSelectionNewWorld` to handle different files.

**Commit Details:**
- using `normalisePathToUnderlyingTarget` to find the filepath for templatepath
- added a test

### DIFF
--- a/editor/src/utils/clipboard.ts
+++ b/editor/src/utils/clipboard.ts
@@ -5,7 +5,7 @@ import { EditorModes } from '../components/editor/editor-modes'
 import {
   DerivedState,
   EditorState,
-  getOpenUIJSFile,
+  getOpenUIJSFileKey,
   withUnderlyingTarget,
 } from '../components/editor/store/editor-state'
 import { getFrameAndMultiplier } from '../components/images'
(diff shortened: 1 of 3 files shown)


editor/src/utils/clipboard.ts
```diff
@@ -150,28 +160,39 @@ export function createDirectInsertImageActions(
...
   }
-  const parseSuccess = openUIJSFile.fileContents.parsed
   const filteredSelectedViews = editor.selectedViews.filter((view) => {
     return R.none((otherView) => EP.isDescendantOf(view, otherView), editor.selectedViews)
   })
-  const utopiaComponents = getUtopiaJSXComponentsFromSuccess(parseSuccess)
-  const jsxElements = filteredSelectedViews.map((view) => {
-    return findElementAtPath(view, utopiaComponents)
-  })
+  const jsxElements = stripNulls(
```
seanparsons, line 175: Can use `mapDropNulls` for this.

seanparsons (approved):

balazsbajorics (approved):

editor/src/utils/clipboard.ts
```diff
@@ -150,28 +160,39 @@ export function createDirectInsertImageActions(
 
 export function createClipboardDataFromSelectionNewWorld(
```
Rheeseyb, line 161: Might as well rename this to remove the `NewWorld` whilst you're here (assuming the old world version no longer exists)

Rheeseyb (approved):

editor/src/utils/clipboard.ts
enidemi: fixed

PR #1198, opened 2021-05-04
Poll logged in state

### LINKED ISSUE #186: Show logged-out users that they are logged out, even if mid-session

balazsbajorics:
Fixes #186

image

**Problem:**
The editor would not tell users if they lost their login for any reason.

**Fix:**
Poll login status (I set it to every 5 seconds), and if the login status changes, dispatch an action. Show a toast. 
I also made a new kind of black status bar for users that were loggin in but are not logged in anymore. This new status bar opens the login screen in a new tab to avoid data loss.

**Commit Details:**
- Login status polling function
- I also wrapped the login status fetch to a try-catch to avoid an unhandled promise rejection error
- new `LOGIN_LOST` UserState to indicate a user that was logged in but not anymore
- a new `getLoginStateFromResponse` which sometimes returns the real login state coming from the server, or LoginLost (because LoginLost is not coming from the server, that is a UserState internal to the editor)
- I also noticed that if the server/internet is down, the black background png of the status bar can never load, so I changed it to have a backgroundColor and backgroundImage as well

### DIFF
--- a/editor/src/components/editor/notification-bar.tsx
+++ b/editor/src/components/editor/notification-bar.tsx
@@ -26,8 +26,12 @@ export const EditorOfflineBar = betterReactMemo('EditorOfflineBar', () => {
 export const LoginStatusBar = betterReactMemo('LoginStatusBar', () => {
   const loginState = useEditorState((store) => store.userState.loginState, 'LoginStatusBar')
 
-  const onClickCallback = React.useCallback(() => {
+  const onClickLoginSameTab = React.useCallback(() => {
     setRedirectUrl(window.top.location.pathname).then(() => window.top.location.replace(auth0Url))
+  }, [])
+
+  const onClickLoginNewTab = React.useCallback(() => {
+    window.open(auth0Url, '_blank')
   }, [])
 
   switch (loginState.type) {
@@ -38,7 +42,7 @@ export const LoginStatusBar = betterReactMemo('LoginStatusBar', () => {
         <NotificationBar
           level='PRIMARY'
           message={'Welcome to Utopia. Click here to sign in and save your projects.'}
-          onClick={onClickCallback}
+          onClick={onClickLoginSameTab}
           style={{ cursor: 'pointer' }}
         />
       )
@@ -47,7 +51,7 @@ export const LoginStatusBar = betterReactMemo('LoginStatusBar', () => {
         <NotificationBar
           level='ERROR'
           message={'You were logged out. Click here to log in again.'}
-          onClick={onClickCallback}
+          onClick={onClickLoginNewTab}
           style={{ cursor: 'pointer' }}
         />
       )
--- a/editor/src/components/editor/notification-bar.tsx
+++ b/editor/src/components/editor/notification-bar.tsx
@@ -50,7 +50,7 @@ export const LoginStatusBar = betterReactMemo('LoginStatusBar', () => {
       return (
         <NotificationBar
           level='ERROR'
-          message={'You were logged out. Click here to log in again.'}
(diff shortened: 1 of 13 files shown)


editor/src/components/editor/notification-bar.tsx
```diff
@@ -26,21 +26,38 @@ export const EditorOfflineBar = betterReactMemo('EditorOfflineBar', () => {
...
+          message={'Welcome to Utopia. Click here to sign in and save your projects.'}
+          onClick={onClickLoginSameTab}
+          style={{ cursor: 'pointer' }}
+        />
+      )
+    case 'LOGIN_LOST':
+      return (
+        <NotificationBar
+          level='ERROR'
+          message={'You were logged out. Click here to log in again.'}
```
maltenuhn, line 53:
```suggestion
          message={'You have been logged out. Click here to log in again and save your changes.'}
```

editor/src/components/editor/server.ts
```diff
@@ -386,3 +392,20 @@ export async function downloadGithubRepo(
...
+  dispatch: EditorDispatch,
+  initialLoginState: LoginState,
+): void {
+  let previousLoginState: LoginState = initialLoginState
+  setInterval(async () => {
+    const loginState = await getLoginState('no-cache')
+    if (previousLoginState.type !== loginState.type) {
+      dispatch([setLoginState(loginState)])
+      if (isLoginLost(loginState)) {
+        dispatch([showToast(notice(`You have been logged out.`, 'ERROR', true))])
```
maltenuhn, line 406:
```suggestion
        dispatch([showToast(notice(`You have been logged out. You can continue working, but your work won't be saved until you log in again.`, 'ERROR', true))])
```

editor/src/uuiui/styles/theme.ts
```diff
@@ -522,14 +522,41 @@ const backgroundURLs = {
...
 
-const noticeStyles = {
-  success: { background: backgroundURLs.green, color: 'white' },
-  info: { background: '#f1f1f1', color: colorTheme.darkPrimary.value },
-  primary: { background: backgroundURLs.blue, color: 'white' },
-  notice: { background: backgroundURLs.paleblue, color: 'white' },
-  warning: { background: backgroundURLs.red, color: 'white' },
-  error: { background: backgroundURLs.almostBlack, color: 'white' },
-  disconnected: { background: backgroundURLs.noise, color: 'white' },
+const noticeStyles: { [styleName: string]: React.CSSProperties } = {
```
maltenuhn, line 525: Thanks for spotting this. Classic mistake!

seanparsons (approved):
Is there a sensible way we can unit test this flow?

balazsbajorics:
> Is there a sensible way we can unit test this flow?

hmm I guess I could write unit tests for the logic itself. the polling, not so much, I don't think so

Rheeseyb:
When this happens (either via logging out in another tab or just straight up deleting the session token cookie), do you still see repeated server request failures with a 401 status?

Rheeseyb (approved):

enidemi (approved):

PR #1200, opened 2021-05-05
Fix cyclic import handling when creating execution scopes

### LINKED ISSUE #1131: [BUG] Typing an import in a file other than the storyboard will bring down the editor
Sample project: https://utopia.pizza/p/5585651d-cant-import-this/ - try uncommenting and completing the import statement on line 4 of `app.js`

Whilst trying to write an import statement in any file other than the storyboard file, the editor will crash during module resolution as soon as you start writing the path for the import, with the error:

```
RangeError: Maximum call stack size exceeded
    at indexJsPath (module-resolution.ts:188)
    at failoverResolveResults (module-resolution.ts:32)
    at resolveNonRelativeModule (module-resolution.ts:186)
    at eval (module-resolution.ts:198)
    at failoverResolveResults (module-resolution.ts:32)
    at resolveNonRelativeModule (module-resolution.ts:186)
    at resolveModule (module-resolution.ts:223)
    at Object.resolveModulePath (module-resolution.ts:229)
    at Object.eval [as resolve] (package-manager.ts:21)
    at eval (editor-state.ts:249)
```

I highly suspect this is related to #1082

Rheeseyb:
Fixes #1131 

**Problem:**
Cyclic imports were tripping the editor in two places: when creating execution scopes for the canvas, and when fetching all (transitively) imported `UtopiaJSXComponent`s.

**Fix:**
In both cases filter paths that have already been resolved. 

**Commit Details:**
- In `ui-jsx-canvas.tsx` if we encounter an already resolved file when creating the execution scope, we return a `left` to indicate that the outer call will handle the execution scope creation
- Updated `getImportedUtopiaJSXComponents` to take an array of paths to filter, filling it with the already resolved file paths
- Added a test to ensure the canvas still renders, and that we can still correctly focus on an element coming from a file with a cyclic dependency
- Moved some useful testing functions for extracting a simplified version of the recorded metadata to `utils.test-utils.ts` so that they can be used in other tests

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
@@ -1,13 +1,4 @@
-import { mapArrayToDictionary } from '../../../core/shared/array-utils'
-import { foldEither } from '../../../core/shared/either'
-import {
-  ElementInstanceMetadata,
-  getJSXElementNameAsString,
-  isJSXElement,
-  ElementInstanceMetadataMap,
-} from '../../../core/shared/element-template'
 import { canvasPoint, point } from '../../../core/shared/math-utils'
-import { objectMap } from '../../../core/shared/object-utils'
 import {
   ParsedTextFile,
   isParseFailure,
@@ -18,7 +9,11 @@ import {
 import * as EP from '../../../core/shared/element-path'
 import { lintAndParse } from '../../../core/workers/parser-printer/parser-printer'
 import { defaultProject } from '../../../sample-projects/sample-project-utils'
-import { wait } from '../../../utils/utils.test-utils'
+import {
+  wait,
+  simplifiedMetadataMap,
+  domWalkerMetadataToSimplifiedMetadataMap,
+} from '../../../utils/utils.test-utils'
 import { addFileToProjectContents } from '../../assets'
 import { setFocusedElement } from '../../editor/actions/action-creators'
 import { StoryboardFilePath } from '../../editor/store/editor-state'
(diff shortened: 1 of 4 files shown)


editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -283,38 +283,47 @@ export const UiJsxCanvas = betterReactMemo(
...
-                const exportsDetail = projectFile.fileContents.parsed.exportsDetail
-                let filteredScope: MapLike<any> = {}
-                for (const s of Object.keys(scope)) {
-                  if (s in exportsDetail.namedExports) {
-                    filteredScope[s] = scope[s]
-                  } else if (s === exportsDetail.defaultExport?.name) {
-                    filteredScope[s] = scope[s]
+              if (resolvedFiles.includes(resolvedFilePath)) {
+                // We're inside a cyclic dependency, so bail and the outer call will create the execution scope
+                return left('Ignoring inner cyclic dependency')
```
Rheeseyb, line 294: This bit is a bit tricky. If you have the below set up:
`a.js`:
```
import { B } from './b'

export const C = (props) => {
  return <div/>
}
export const A = (props) => {
  return <B/>
}
```
`b.js`:
```
import { C } from './a'

export const B = (props) => {
  return <C/>
}
```

Then you'll be able to focus `A/B`, but not `A/B/C` (and so won't be able to edit the final underlying `div`).

I tried experimenting with retaining a `resolvedScopes` instead, like so:
```
let resolvedScopes: { [filepath: string]: any } = {}
...
if (resolvedFilePath in resolvedScopes) {
  return right(resolvedScopes[resolvedFilePath])
} else {
  let filteredScope: MapLike<any> = {}
  resolvedScopes[resolvedFilePath] = filteredScope
```
(where `filteredScope` is then filled in below as before), but that results in `C` being `undefined` when rendering.

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -283,38 +283,47 @@ export const UiJsxCanvas = betterReactMemo(
...
+              } else {
+                resolvedFiles.push(resolvedFilePath)
+                const projectFile = getContentsTreeFileFromString(projectContents, resolvedFilePath)
+                if (isTextFile(projectFile) && isParseSuccess(projectFile.fileContents.parsed)) {
+                  const { scope } = createExecutionScope(
+                    resolvedFilePath,
+                    customRequire,
+                    // (innerImportOrigin: string, innerToImport: string) =>
+                    //   requireFn(innerImportOrigin, innerToImport, false),
```
Rheeseyb, line 303: Oops, didn't mean to commit this!

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
```diff
@@ -168,19 +109,43 @@ export const Card = () => {
...
+}
+
 describe('Spy Wrapper Template Path Tests', () => {
   it('a simple component in a regular scene', async () => {
     const { getEditorState } = await createExampleProject()
 
     await wait(20000)
     const spiedMetadata = getEditorState().editor.spyMetadata
-    const sanitizedSpyData = extractElementPathStuffFromElementInstanceMetadata(spiedMetadata)
+    const sanitizedSpyData = simplifiedMetadataMap(spiedMetadata)
```
balazsbajorics, line 142: boo for "fixing" the function name :D

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
```diff
@@ -2007,13 +1972,370 @@ describe('Spy Wrapper Multifile Template Path Tests', () => {
...
+          "name": "HiElement",
+          "rootElements": Array [],
+        },
+      }
+    `)
+  })
+})
+
+describe('Spy Wrapper Multifile With Cyclic Dependencies', () => {
+  it('a generated component instance is focused inside a component instance inside the main App component', async () => {
```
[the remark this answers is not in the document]
Rheeseyb: On master it explodes because of the stack overflow

editor/src/components/canvas/ui-jsx-canvas.tsx
balazsbajorics: this is so weird. but also this is pretty much an edge case so I will resist the temptation to fall into this rabbit hole

Rheeseyb: I'm going to create a follow up ticket on this and add it to the backlog - this solution works because returning the `left` means we fall back to the regular require function, meaning that (as mentioned above) a component that is in a dependency cycle won't be spy wrapped (though one in a file with a cyclic dependency will), so we still are left with the above edge case, but that doesn't seem as high priority as fixing the far more common case (which this does).

balazsbajorics (approved):

PR #1207, opened 2021-05-05
integrate redux devtools, if available

balazsbajorics:
Well this was easier than expected!

to make it work install https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd

image

If a redux devtools extension is detected, the editor will now forward the name of the dispatched actions and the updated state for every dispatch.

This makes it very easy to record (save and SHARE!) a small session of user interactions, alongside the full states for every action.

The best thing is that when you inspect an action, the devtool will intelligently diff the editor state and only show what changed, this is very handy because our full editor state is huge, so being able to quickly focus on the relevant parts is a huge boost.

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -81,6 +81,7 @@ import {
   identifyFilesThatHaveChanged,
 } from '../../../core/shared/project-contents-dependencies'
 import { CodeResultCache, generateCodeResultCache } from '../../custom-code/code-file'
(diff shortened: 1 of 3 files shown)


maltenuhn:
I tested it this evening, works a treat!!

maltenuhn (approved):

enidemi (approved):

Rheeseyb (commented):
:+1:

PR #1211, opened 2021-05-06
Redux Devtool: sanitizing logged state, adding log function 

balazsbajorics:
**Problem:**
The logged state was ginormous and it made the devtools extension to crash after a few minutes of editor use.

**Fix:**
I made a filter function and it is filtering out the history, node_modules, and some other parts of the store that were sticking out. We can refine this any time if it is still too large

**Commit Details:**
- `sanitizeLoggedState` function
- new `reduxDevtoolsLogMessage` function to be able to write custom loglines so they show up among the dispatched actions, like so:
image

### DIFF
--- a/editor/src/core/shared/redux-devtools.ts
+++ b/editor/src/core/shared/redux-devtools.ts
@@ -1,5 +1,6 @@
 import type { EditorAction } from '../../components/editor/action-types'
 import type { EditorStore } from '../../components/editor/store/editor-state'
+import { pluck } from './array-utils'
 
 interface Connection {
   subscribe: (listener: (message: { type: string; state: string }) => void) => () => void // adds a change listener. It will be called any time an action is dispatched from the monitor. Returns a function to unsubscribe the current listener.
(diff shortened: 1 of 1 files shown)


enidemi (approved):

editor/src/core/shared/redux-devtools.ts
```diff
@@ -25,11 +26,45 @@ const maybeDevTools = connectDevToolsExtension()
...
+    history: PlaceholderMessage,
+    workers: PlaceholderMessage,
+    dispatch: PlaceholderMessage,
+  }
+}
+
 export function updateReduxDevtools(actions: Array<EditorAction>, newStore: EditorStore): void {
   if (maybeDevTools != null) {
     // filter out the actions we are not interested in
     if (!actions.some((a) => ActionsToOmit.includes(a.action))) {
```
Rheeseyb, line 63: I'd say instead of this `some` check we should filter the actions themselves

Rheeseyb (approved):

Issue #1137, opened 2021-04-21
🐛 when you rename an imported component as a variable it will not be spy-wrapped


maltenuhn:

seanparsons:
@maltenuhn Is this what is meant by that description?

```javascript
import * as React from 'react'
import { Scene, Storyboard } from 'utopia-api'
import { App } from '/src/app.js'

const AppAsVariable = App

export var storyboard = (
  <Storyboard data-uid='storyboard-entity'>
    <Scene
      data-label='Imported App'
      data-uid='scene-1-entity'
      style={{ position: 'absolute', left: 0, top: 0, width: 375, height: 812 }}
    >
      <AppAsVariable data-uid='app-entity' />
    </Scene>
  </Storyboard>
)
```

maltenuhn:
I think so. @enidemi IIRC you reported this, were there other cases?

enidemi:
This example code is correct from @seanparsons , this is how I found this bug. I haven't seen other cases.

seanparsons:
Closing for now as it appears the current handling is good enough and a good fix would potentially be a lot of work.

PR #1212, opened 2021-05-06
Results of investigating component defined as variable.

seanparsons:
This is not so much a fix for https://github.com/concrete-utopia/utopia/issues/1137 because for now that appears to be a wont fix. But in the process of looking at that I fixed a weird require function that was used in some tests around that and added a comment mentioning how the issue affects one particular part.

**Commit Details:**
- Removed `dumbRequireFn` and replaced it with a delegated call to
  `getRequireFn`.
- Added a FIXME to cover the result of part of the investigation
  around the bug.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas.test-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas.test-utils.tsx
@@ -49,6 +49,8 @@ import { NO_OP } from '../../core/shared/utils'
 import { directory } from '../../core/model/project-file-utils'
 import { contentsToTree } from '../assets'
 import { MapLike } from 'typescript'
+import { generateCodeResultCache } from '../custom-code/code-file'
+import { getRequireFn } from '../../core/es-modules/package-manager/package-manager'
 
 export interface PartialCanvasProps {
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):

editor/src/components/canvas/ui-jsx-canvas.test-utils.tsx
```diff
@@ -49,6 +49,8 @@ import { NO_OP } from '../../core/shared/utils'
 import { directory } from '../../core/model/project-file-utils'
 import { contentsToTree } from '../assets'
 import { MapLike } from 'typescript'
+import { generateCodeResultCache } from '../custom-code/code-file'
```
Rheeseyb, line 52: Looks like this import wasn't needed in the end

Rheeseyb (approved):

editor/src/components/canvas/ui-jsx-canvas.test-utils.tsx
seanparsons:
```suggestion
```

PR #1214, opened 2021-05-06
Trying to fix the auto-forking behavior

balazsbajorics:
**Problem:**
When you open somebody else's project, we immediately trigger a SAVE and as a byproduct, fork the project. This is silly.

**Fix:**
I am trying to prevent the first save from happening if it is not directly a result of a user action. First edit should still save+fork, that is expected behavior. 

**Commit Details:**
- Added `EditorState.alreadySaved: boolean` 
- `shouldSave` logic was triggered on `updateCodeResultCache`, now it is triggered on `updateCodeResultCache && frozenEditorState.alreadySaved`, meaning that a codeResultCache change only triggers a save if the project was already saved for another reason. This prevents `updateCodeResultCache` from triggering a save+fork.


balazsbajorics:
Ping :)

maltenuhn:
@balazsbajorics pong 360,000ms

enidemi (approved):

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -471,16 +475,45 @@ export function editorDispatch(
   }
 
   const isLoaded = frozenEditorState.isLoaded
+  /**
+   * We only want to trigger a save for code result cache change if the project was already saved.
+   * This is to prevent accidentally forking projects before the user even edits anything in them.
+   */
+  const saveBecauseCodeResultCacheChanged = updateCodeResultCache && frozenEditorState.alreadySaved
```
Rheeseyb, line 482: Should we even still trigger a save on a change to the code result cache? I have a feeling this is related to updating the preview rather than actually storing the result of the cache change

balazsbajorics: the original change was a fix so that if you introduce a compile error and reload, the error persists. to be honest I did not verify that removing this would make a regression wrt error messages...

balazsbajorics:
@Rheeseyb I re-did the PR because you were right: we didn't need that flag anymore, and I actually deleted one more flag as well

PR #1225, opened 2021-05-07
Ensure consistent JSX attribute simplifying for the sake of memoisation

Rheeseyb:
**Problem:**
We're blowing memoisation in a lot of places because non-changes to the model will result in a simple `JsxAttributeValue` being replaced with a more complex `JSXAttributeNestedObject`. This was because the simplification was only applied during parsing the initial code, and never applied after making changes to the model.

**Fix:**
Ensure we always simplify the model after making changes. This happens in 2 places:

1. In `element-template.tsx` as part of `setJSXAttributesAttribute` (which updates the list of attributes with a new value)
2. In `jsx-attributes.ts` when unsetting values

### DIFF
--- a/editor/src/core/performance/performance-regression-tests.spec.tsx
+++ b/editor/src/core/performance/performance-regression-tests.spec.tsx
@@ -58,8 +58,8 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(705) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(715)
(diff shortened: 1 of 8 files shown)


editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -58,8 +58,8 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(705) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(715)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(270) // if this breaks, GREAT NEWS but update the test please :)
```
balazsbajorics, line 61: this is too low, I don't trust you

balazsbajorics (approved):
impressive. I believe this fixes the "regression" found by @enidemi

enidemi (approved):

seanparsons (approved):

Issue #1186, opened 2021-05-03
useContextSelector memoization broken


balazsbajorics:
We used to have a custom fork of useContextSelector, I made that because I added an extra third parameter, an optional equality function. 

When we deleted my code and went back to the mainline useContextSelector, we forgot to remove my own type definition, which masked the regression that the real useContextSelector uses Object.is as the equality function. This means that a lot of inspector hooks are now badly memoized!

(We did not realize this for months, because the inspector regression test, by accident was using an example code with lots of "spied" values, where the only value comes from the dom-walker's css computedStyle and not from the parsed props, so the regression tests are not testing if the parsed props are memoized or not)

Eni's removal of the layout prop meant that the regression test suddenly started working again, making us realize that there is a problem somewhere.

Rheeseyb:
This has work done in #1226 , but needs further assessment

maltenuhn:
@Rheeseyb how do we assess it? Or can this now be closed?

balazsbajorics:
@maltenuhn 

we need to make sure that our patch is working and is memoizing right, we can write tests to verify this.

we might as well update the dependency if we touch it anyways. (ideally write the tests first, to validate the API is stable)

PR #1226, opened 2021-05-07
Introduce use-optimized-selector

Rheeseyb:
Part way to a solution to #1186 

**Problem:**
We're still stuck on an old build of [`use-context-selector`](https://github.com/dai-shi/use-context-selector) as we're relying on a patch of that for adding an equality check param.

**Fix:**
Introduce [`use-optimized-selector`](https://github.com/johnrom/use-optimized-selector) for applying a layer of caching around selectors used by the above, as discussed in [this comment thread](https://github.com/dai-shi/use-context-selector/issues/19#issuecomment-798984730). I'm opening this PR primarily to see how the performance tests run with this, since it completely removes the passing of that equality function to the patched build of `use-context-selector`.

I have also tested this out with the latest version of `use-context-selector` (1.3.7), which pushes the render count tests up again, so there is still further work to be done to allow us to move onto the latest version of that.

**Commit Details:**
- Added a new function `useContextSelector` to `react-performance.ts`, which combines the above two libraries
- Point all uses of `use-context-selector` to the above function
- Ensure each selector function is either a constant or is memoised (as [is required](https://github.com/johnrom/use-optimized-selector#getting-started))

Annoyingly, the `react-hooks/exhaustive-deps` eslint rule was tripped in a few cases by generic type symbols (e.g. in `useGetMultiselectedProps` the rule complains that `P` is not included in the list of dependencies, which is nonsense!), so I've had to disable those on a few lines. I'll check to see if there is an update for that rule to fix the issue.

### DIFF
--- a/editor/package-lock.json
+++ b/editor/package-lock.json
@@ -27508,6 +27508,11 @@
       "resolved": "https://registry.npmjs.org/use-memo-one/-/use-memo-one-1.1.1.tgz",
       "integrity": "sha512-oFfsyun+bP7RX8X2AskHNTxu+R3QdE/RC5IefMbqptmACAA/gfol1KDD5KRzPsGMa62sWxGZw+Ui43u6x4ddoQ=="
     },
+    "use-optimized-selector": {
+      "version": "1.0.0-beta3",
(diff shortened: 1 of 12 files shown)


seanparsons (approved):

maltenuhn:
Is this PR still something we want to merge?

Rheeseyb:
Tagged and closed

PR #1227, opened 2021-05-07
Add Auth0 Client ID to PR deploy job

balazsbajorics:
**Problem:**
The sign in button for branch deploys is broken

**Fix:**
pass in the staging auth0 client ID for branch deploys

### DIFF
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -61,6 +61,9 @@ jobs:
     runs-on: ubuntu-latest
     env:
       UTOPIA_SHA: ${{ github.sha }}
+      AUTH0_CLIENT_ID: KB7euFO46rVYeOaWmrEdktdhAFxEO266
+      AUTH0_ENDPOINT: enter.utopia.app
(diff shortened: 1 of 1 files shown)


balazsbajorics:
I verified this works by opening the PR deployed editor link and clicking on sign in

.github/workflows/pull-requests.yml
```diff
@@ -61,6 +61,9 @@ jobs:
     runs-on: ubuntu-latest
     env:
       UTOPIA_SHA: ${{ github.sha }}
+      AUTH0_CLIENT_ID: KB7euFO46rVYeOaWmrEdktdhAFxEO266
+      AUTH0_ENDPOINT: enter.utopia.app
```
seanparsons, line 65: This does expose our staging environment to the public, I thought this was going to go in the settings?

balazsbajorics: oh crap, I misunderstood our conversation, once we settled that these are not private keys, I thought they don't need to be in the Github Secrets either (because the client will have these strings compiled into them anyways)

PR #1229, opened 2021-05-07
Draft - testing updating use-context-selector to the latest alpha version

Rheeseyb:


editor/src/utils/react-performance.ts
```diff
@@ -475,3 +477,12 @@ export function useKeepDeepEqualityCall<T>(newValue: T, keepDeepCall: KeepDeepEq
   oldValue.current = keepDeepCall(oldValue.current, newValue).value
   return oldValue.current
 }
+
```
Rheeseyb, line 480:
```suggestion
 
```

PR #1230, opened 2021-05-10
do not show two black bars when the editor is offline

### LINKED ISSUE #1196: Logging out in one tab shows "Utopia is offline" error bar
STR:

1. Open a project in one tab, log in, and open the Chrome inspector
2. Open another tab and go to `/logout`

The open project will now display the "Utopia is offline" status bar along the top, and will repeatedly make failing (with a 401 status) calls to the server.

MN edit: see also #1196

balazsbajorics:
Fixes #1196

**Problem:**
After my previous PR #1198 we still had an issue: if the editor was offline, it would both say that it was offline and also that it was logged out. 

**Fix:**
I unified the handling of the two error messages, and if the login polling cannot reach the server, it switches to offline state instead of logged out state.

**Commit Details:**
- Moved `EditorOfflineBar` to `LoginStatusBar` so we don't show two of them at the same time
- if the `getLoginState` fetch fails, we assume the editor is offline instead of the user being logged out

### DIFF
--- a/editor/src/components/editor/notification-bar.tsx
+++ b/editor/src/components/editor/notification-bar.tsx
@@ -14,7 +14,7 @@ export const BrowserInfoBar = betterReactMemo('EditorOfflineBar', () => {
   )
 })
 
-export const EditorOfflineBar = betterReactMemo('EditorOfflineBar', () => {
+const EditorOfflineBar = betterReactMemo('EditorOfflineBar', () => {
   return (
     <NotificationBar
       level='ERROR'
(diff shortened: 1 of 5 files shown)


editor/src/components/editor/notification-bar.tsx
```diff
@@ -25,14 +25,24 @@ export const EditorOfflineBar = betterReactMemo('EditorOfflineBar', () => {
...
+    'EditorComponentInner saveError',
+  )
 
   const onClickLoginNewTab = React.useCallback(() => {
     window.open(auth0Url('auto-close'), '_blank')
   }, [])
 
+  if (saveError) {
+    return <EditorOfflineBar />
+  }
```
Rheeseyb, line 39: Is there a situation where being logged out would trigger the save error and so show this rather than the logged out bar?

balazsbajorics: nope, because the save logic uses isLoggedIn to decide to attempt an online save or go for a local cached save:

https://github.com/concrete-utopia/utopia/blob/796e62c4a43b737eac464ea25c13d5093c12a31a/editor/src/components/editor/store/dispatch.tsx#L663-L667

Rheeseyb (approved):

PR #1233, opened 2021-05-10
Fix: do not automatically remount the canvas

### LINKED ISSUE #1089: The editor keeps reloading the images!
This is most visible when you have the chrome dev tools open, as that disables caching. but even when cached, the images are reloaded, just reloaded from cache and so it's not visible.

This causes a problem on the canvas, the dom-walker would sometimes mismeasure an image that was reloaded.

balazsbajorics:
Fixes #1089

**Problem:**
The editor would remount the canvas if: 
* the code changed
* any undo redo happened
* the focused view changed
* unwrapping the selected view

Remounting the canvas means we reset the react state of the components, and also remove + re-add the DOM elements. This was most visible when the Dev Tools was open: especially because of the focus mode changes, all of the images would blink on the canvas.

The original reason why we introduced remounting was this: https://github.com/concrete-utopia/dystopia/issues/1995
Basically the problem it was solving is that if you are working on the style object, if you write a value which React thinks is invalid, it will **not** remove the existing value. So while typing code, you might think your style object looks correct, but after reloading the editor you would see that it is broken. We quickly came up with the remounting solution which also resets the components states, and we kind of thought of it as a side benefit.
The downside, though is that this is not optional and is not aligned with react-fast-refresh: in codesandbox, if you make the same style object mistakes, you will see the same (old utopia) behavior: the rendered style would not update until you typed in a correct value. Plus the component states survive the update, making it easier to debug complicated state, for example if you navigate somewhere with a router or you want to tweak the design of a popup that only comes up after a certain action happened.

**Furthermore,** we started to use the canvas remounting to also mean "invalidate the dom-walker cache". This meant that we started remounting the canvas when the Focus would change and lately, for unwrapping: the intention is a dom-walker reset, which we should be able to separately express.

**Fix:**
Instead of automatically remounting the canvas, if you change the code, the RESET_CANVAS button becomes 'active':
_Look at the Refresh button on the right hand side:_
image
image

Pressing that button would manually trigger a RESET_CANVAS action which remounts the canvas.

I also separated the logic for clearing the dom-walker cache so we can trigger that without remounting the canvas.

**Commit Details:**
- next to the existing `mountCount`, I added `canvasContentInvalidateCount` and `domWalkerInvalidateCount` to the editor state
- new RESET_CANVAS action increases mountCount
- everything else just increases `canvasContentInvalidateCount` which makes the Reset Canvas button turn black
- Plus I use `domWalkerInvalidateCount` where the true intention is to just reset the dom-walker cache.


balazsbajorics:
@maltenuhn this PR is waiting for decision / design from you.

TLDR is that two years ago I eagerly introduced canvas remounting as a tool to make us better than React Fast Refresh – and didn't realize at the time that the price we pay is blinky images and components loosing their state.

This PR brings us inline with React Fast Refresh, but I don't know if that's what we want or not.

Rheese is right that the Reset Canvas button is not very noticable.

maltenuhn:
I've tried the test link but the code editor doesn't load for me there. 

> Basically the problem it was solving is that if you are working on the style object, if you write a value which React thinks is invalid, it will not remove the existing value. 

I don't recall running into this issue much at all with CSB - not sure if that's driven by red squiggly lines, or if they also force a remount once lint errors are gone? In any event, I think the change here is fine as long as we follow it up with adequate linting and autocomplete. CMD+S to refresh the canvas also works for me. 

Re button: we could slap a `primary` on it - right now it's meant more as a backup book of spells / magical mystery / have-you tried-turning-it-off.. button. But if we already map CMD+S to remount, perhaps no need? WDYT?

seanparsons (approved):

maltenuhn:
ping @balazsbajorics - did you want to talk about this further?

balazsbajorics:
> ping @balazsbajorics - did you want to talk about this further?

I think we should merge this as-is, and see how much we want to change the behavior once we used it for a few days!

PR #1236, opened 2021-05-11
Cache the component renderer against the element and props

### LINKED ISSUE #1232: Resizing has drastically regressed
In the last couple of weeks resizing performance has drastically regressed ~with it suspected to be related to snapping (since the frame times when resizing dramatically jump when hitting a half pixel value)~

Rheeseyb:
Fixes #1232 

**Problem:**
With the previous old-style scenes, we were explicitly preventing rendering any scenes that weren't directly updated. However, with the introduction of natural scenes we lost this ability, meaning that updating any element on the canvas would trigger a re-render of anything defined in that same file.

**Fix:**
Introduce heavy use of `React.useMemo` inside the component created by `createComponentRendererComponent`, caching against the top level element itself, along with its props.

This still has an outstanding issue whereby components that are given children will still re-render, as the props object for those will be continually changing, so this still requires a follow up to address that (I'll open a separate ticket for that).


enidemi (approved):

Rheeseyb:
Converted to a draft as there is still a pretty significant issue with this (running on the canvas can lead to a React hooks error)

Rheeseyb:
I'm going to close this as unfortunately it doesn't appear to be a viable solution since it breaks running apps in the Canvas that use hooks. The main issue there being that we would still need to execute any arbitrary JS block when rendering the element, which this caching prevents. I believe this also has other issues around determining when we should render a component, which is a non-trivial problem in the context of our editor.

PR #1238, opened 2021-05-12
Bump ESLint Plugins

### LINKED ISSUE #524: Update ESLint + deps

balazsbajorics:
Fixes #524 

**Problem:**
The eslint react hooks were ancient, and they had a bug: if you would have a type, such as `T` in a useEffect or useCallback hook, eslint would complain that you forgot to list `T` in the dependency list.

**Fix:**
Bump eslint plugins everywhere. It turns out the current version of `eslint-plugin-react-hooks` bombs out if you have 
`React.useEffect()` at the top level of a file. Which broke one of our tests, and also it means that while trying to reproduce it, I realized that the worker just dies if eslint throws an error. I put linting in a try-catch and return a `ESLint runtime error` lint error, which the canvas-error-overlay displays like so:
image

**Commit Details:**
- Bump all the eslints plugins, (this affects our devtools but also the eslint bundled for the users as well)
- Fix all the places I could find that had an eslint-disable rule for hook dependencies
- Add a try-catch around linting inside the editor's own linter worker
- Updated the test called "React.useEffect at the root fails usefully" which was snapshotting garbage (the snapshot claims we are _not_ showing an error message).

**Note:**
We are on the last vesion of `babel-eslint` that can still run in the browser, if we are going to face a situation where we need to bump `babel-eslint`, we are going to have to fork them and create a `babel-eslint4b` (which would be a version of babel-eslint that depends on `eslint4b` instead of `eslint`)


balazsbajorics:
this is ready for review

seanparsons (approved):

Rheeseyb (approved):

balazsbajorics:
ahhh so many merge conflicts!

PR #1239, opened 2021-05-12
Cache dynamic to static path calls

Rheeseyb:
Partial work towards #1232 

**Problem:**
We are performing a *lot* of conversions from dynamic to static paths, which was having a significant impact on performance.

**Fix:**
Cache the results of those conversions.

### DIFF
--- a/editor/src/core/shared/element-path.ts
+++ b/editor/src/core/shared/element-path.ts
@@ -36,12 +36,20 @@ function emptyElementPathCache(): ElementPathCache {
   }
 }
 
+let dynamicToStaticPathCache: Map<ElementPath, StaticElementPath> = new Map()
+let dynamicElementPathToStaticElementPathCache: Map<
+  ElementPathPart,
+  StaticElementPathPart
+> = new Map()
+
 let globalPathStringToPathCache: { [key: string]: ElementPath } = {}
 let globalElementPathCache: ElementPathCache = emptyElementPathCache()
 
 export function clearElementPathCache() {
   globalPathStringToPathCache = {}
   globalElementPathCache = emptyElementPathCache()
+  dynamicToStaticPathCache = new Map()
+  dynamicElementPathToStaticElementPathCache = new Map()
 }
 
 function getElementPathCache(fullElementPath: ElementPathPart[]): ElementPathCache {
@@ -695,11 +703,25 @@ export function isFromSameInstanceAs(a: ElementPath, b: ElementPath): boolean {
 }
 
 function dynamicElementPathToStaticElementPath(element: ElementPathPart): StaticElementPathPart {
-  return element.map(extractOriginalUidFromIndexedUid) as StaticElementPathPart
+  const existing = dynamicElementPathToStaticElementPathCache.get(element)
+  if (existing == null) {
+    const result = element.map(extractOriginalUidFromIndexedUid) as StaticElementPathPart
+    dynamicElementPathToStaticElementPathCache.set(element, result)
+    return result
+  } else {
+    return existing
+  }
 }
 
 export function dynamicPathToStaticPath(path: ElementPath): StaticElementPath {
-  return elementPath(path.parts.map(dynamicElementPathToStaticElementPath))
+  const existing = dynamicToStaticPathCache.get(path)
+  if (existing == null) {
+    const result = elementPath(path.parts.map(dynamicElementPathToStaticElementPath))
+    dynamicToStaticPathCache.set(path, result)
+    return result
+  } else {
+    return existing
+  }
 }
 
 export function makeLastPartOfPathStatic(path: ElementPath): ElementPath {


balazsbajorics:
I'm gonna re-run the test to get one more data point, but it seems like this makes resize ~20-30 ms faster!

balazsbajorics (approved):

editor/src/core/shared/element-path.ts
```diff
@@ -36,12 +36,20 @@ function emptyElementPathCache(): ElementPathCache {
   }
 }
 
+let dynamicToStaticPathCache: Map<ElementPath, StaticElementPath> = new Map()
```
seanparsons, line 39: I think you should be using `ComplexMap` for these, non-primitive keys are slightly dicey with `Map` from what I remember.

seanparsons (approved):

editor/src/core/shared/element-path.ts
balazsbajorics: the ElementPaths are coming from a cache so that `EP.fromString('aaa') === EP.fromString('aaa')`, so I think Map is safe in this instance.

plus the main problem with this function is that it is called _a lot_ (_a lot!_) I think the toString would significantly add to the cost, but of course it's cheap to try.

seanparsons: Does it make more sense to embed the static path directly into the dynamic one? Then there's no need to do any cache invalidation.

Rheeseyb: > Does it make more sense to embed the static path directly into the dynamic one?

Rheeseyb: @seanparsons can you elaborate please? Right now we get the static path from the dynamic one by trimming off all of the dynamic parts, but that still requires string splitting and array iteration, and then a conversion from the result to an `ElementPath` (which includes more array iteration as part of the look up from the cache).

seanparsons: `
export interface ElementPath {
  type: 'elementpath'
  parts: Array<ElementPathPart>
  staticPath: StaticElementPath
}
`
Then fill in the blanks when it comes to things like `appendNewElementPath` and the like. Then theoretically it will be correct by construction and `dynamicPathToStaticPath` becomes `path.staticPath`.

balazsbajorics: I am torn – I think there's lots of merit to storing extra information on the ElementPath object (such as the filepath as I mentioned a few weeks ago), on the other hand, the snapshots would become unreadable :(

seanparsons: I _really_ don't think we should bend the design of the app to make the snapshots less nightmareish.

Rheeseyb: Let me take a quick look at the damage

balazsbajorics: > I really don't think we should bend the design of the app to make the snapshots less nightmareish.

fair point. 

this ties into something I actually wanted to do: we have two or three places with out-of-control snapshots, I think we should write a formatter/pretty-printer function for the ui-jsx-canvas snapshots – and omit 95% of the garbage data, like the source code of the source file of a top level element included for all top level elements

seanparsons: That would be really nice, I've done ad-hoc things like this: https://github.com/concrete-utopia/utopia/pull/1240/files#diff-c698794f5721b9092b87f29c94afd791d509cf5495b0f65a85372f849fb8a2e0R1056

But a common I suppose `MinimalEditorState` or some such, would be pretty handy, much as doing the first version will be a pain in the face.

balazsbajorics: another recent example: https://github.com/concrete-utopia/utopia/blob/3110af410e7c4874a086bf2efdd20019091ed9a6/editor/src/utils/utils.test-utils.ts#L423

balazsbajorics: however, I think for this specific case, the `Map` _is_ correct, and sounds like the simplest solution to this problem

Rheeseyb: Ok I've had a look and taking that approach becomes pretty complex pretty quickly, as it requires either a really nasty hack:
```
export interface StaticElementPath {
  type: 'elementpath'
  parts: Array<StaticElementPathPart>
  staticPath: StaticElementPath | undefined
}

export interface ElementPath {
  type: 'elementpath'
  parts: Array<ElementPathPart>
  staticPath: StaticElementPath | undefined
}
...
function newStaticElementPath(fullElementPath: StaticElementPathPart[]): StaticElementPath {
  return {
    type: 'elementpath',
    parts: fullElementPath,
    staticPath: undefined,
  }
}

function newElementPath(fullElementPath: StaticElementPathPart[]): StaticElementPath
function newElementPath(fullElementPath: ElementPathPart[]): ElementPath
function newElementPath(fullElementPath: ElementPathPart[]): ElementPath {
  const staticPath = newStaticElementPath(
    fullElementPath.map(dynamicElementPathToStaticElementPath),
  )

  return {
    type: 'elementpath',
    parts: fullElementPath,
    staticPath: staticPath,
  }
}
```

or it requires more refactoring of `StaticElementPath` and `StaticElementPathPart` (since the latter is just `export type StaticElementPathPart = StaticModifier & Array<id>`).

I'm going to go with `Map` being the way to go right now since all creations of `ElementPath` will be returning cached objects anyway.

Issue #981, opened 2021-03-19
Fix and improve the hook


maltenuhn:
**Problem**
The hook that returns info / metadata about the currently selected icon doesn't seem to 
- return the correct information (storyboard)
- doesn't match the uniquely identifiable information about which of the extant / insertable components in the project it's talking about

This means that e.g. the dropdown in `render-as.tsx` can't show the correct entry for `Button` if there are multiple `Button` components in the project being exported by different source files or libraries

**Expected behaviour**
- [ ] three `Button` components (same file, different file, NPM project) and one `button` element can correctly be distinguished by `render-as.tsx` and show the correct one for each
- [ ] we provide information about the source file (if applicable) or npm project (if applicable) or element status (if applicable) of the selected thing in the hook, and use that in the popup.

maltenuhn:
@seanparsons Hi there! Is your work in #977 related to this at all?

seanparsons:
@maltenuhn It would likely need the function from that PR to identify where something is imported from, as opposed to the core of the work that PR does.

PR #1242, opened 2021-05-12
Fix the Render As dropdown

balazsbajorics:
**Problem:**
image

The Render As menu would never show a selected value. opening the dropdown and selecting something would work, but then the closed dropdown would show "empty".

I discovered this because I was looking at `useNamesAndIconsAllPaths` and why is it used because that function is terribly slow AND triggers a rerender for any user interaction.

**Fix:**
I removed the usage of `useNamesAndIconsAllPaths` and fixed `currentInsertableComponent`

**Commit Details:**
- Added function `MetadataUtils.getJSXElementNameFromMetadata` that only uses path and metadata to be able to skip having to drill into the project hierarchy
- Use `MetadataUtils.getJSXElementNameFromMetadata` instead of `useNamesAndIconsAllPaths`
- Fix a bug with `currentInsertableComponent` which meant that it could never correctly find the current insertable component

### DIFF
--- a/editor/src/components/canvas/controls/render-as.tsx
+++ b/editor/src/components/canvas/controls/render-as.tsx
@@ -29,9 +29,6 @@ export const RenderAsRow = betterReactMemo('RenderAsRow', () => {
     )
   }, 'RenderAsRow selectedElementName')
 
-  const constrolStatus = 'simple'
-  const controlStyles = getControlStyles(constrolStatus)
-
   const refElementsToTargetForUpdates = usePropControlledRef_DANGEROUS(
     getElementsToTarget(selectedViews),
   )
@@ -112,7 +109,7 @@ export const RenderAsRow = betterReactMemo('RenderAsRow', () => {
       </span>
       {insertableComponents.length > 0 ? (
         <PopupList
-          disabled={!controlStyles.interactive}
+          disabled={false}
           value={currentInsertableComponent}
           onSubmitValue={onSelect}
           options={insertableComponents}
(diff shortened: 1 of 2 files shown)


maltenuhn:
Nice! Does this partially close #981 ?

Rheeseyb (approved):

balazsbajorics:
> Nice! Does this partially close #981 ?

I don't think so. I assume the ticket refers to the `useNameAndIcon` and `useNamesAndIconsAllPaths` hooks. (But I might be wrong.) I did not change those, I just remove one place calling it.

This PR only touches the Render As line and that was not using icons. It was (wrongly) calling `useNamesAndIconsAllPaths` and using the first element from the array to get a `name` but it did not work for multiple reasons: the first element returned by `useNamesAndIconsAllPaths` has nothing to do with the selected element (it is literally just the first element of the first traversed file of the project), plus the matching did not know about `insertableComponents` being grouped, so it tried to match a `name` against a group header, which always failed.

editor/src/components/canvas/controls/render-as.tsx
```diff
@@ -16,16 +15,23 @@ import {
...
   }, 'TopMenuContextProvider')
 
+  const selectedElementName = useEditorState((store) => {
+    return MetadataUtils.getJSXElementNameFromMetadata(
+      store.editor.selectedViews[0],
+      store.editor.jsxMetadata,
+    )
+  }, 'RenderAsRow selectedElementName')
+
+  const constrolStatus = 'simple'
```
seanparsons, line 32:
```suggestion
  const controlStatus = 'simple'
```

seanparsons (approved):

enidemi (approved):

OmarDaSilva (approved):

PR #1243, opened 2021-05-12
Fix/assets toast messages

### LINKED ISSUE #1218: [assets] uploading a new image shows unrelevant toast message
Uploading an asset that didn't exist already shows the "assets replaced" toast, even when nothing was in fact replaced. 
The correct behaviour is that it should only be shown when an existing asset was actually replaced.
This was added because caching(?) makes it show the previous asset until reloading the editor, if this can be fixed the toast can be removed completely.

enidemi:
Fixes #1218

**Problem:**
There were problems with the image drag and drop toast messages. One of the issues was a missing toast message for logged out user about trying to insert an asset.
The other image insertion issue is the red toast message when drag and drop an image to the filebrowser was always visible, even when inserting a completely new image it was showing `Assets replaced. You may need to reload the editor to see changes.`.

To reliably remove the need for the replace toast message hashing would be required, as discussed [here](https://github.com/concrete-utopia/dystopia/pull/3454/files#r421373984).

**Commit Details:**
- fixed  `actions.ts` to add the missing toast to the editor state for logged out users
- fixed `fileResultUploadAction` to make replace toast message show only on image overwrite

### DIFF
--- a/editor/src/components/editor/image-insert.ts
+++ b/editor/src/components/editor/image-insert.ts
@@ -5,15 +5,19 @@ import { notice } from '../common/notice'
 import { EditorAction, EditorDispatch } from './action-types'
 import * as EditorActions from './actions/action-creators'
 
-async function fileUploadAction(file: File, targetPath: string): Promise<EditorAction> {
+async function fileUploadAction(
+  file: File,
+  targetPath: string,
+  replace: boolean,
+): Promise<EditorAction> {
   const fileResult = await extractFile(file)
-  return fileResultUploadAction(fileResult, targetPath)
+  return fileResultUploadAction(fileResult, targetPath, false)
 }
 
 export function fileResultUploadAction(
   fileResult: FileResult,
   targetPath: string,
-  replace: boolean = false,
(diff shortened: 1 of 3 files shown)


editor/src/components/editor/image-insert.ts
```diff
@@ -10,10 +10,16 @@ async function fileUploadAction(file: File, targetPath: string): Promise<EditorA
   return fileResultUploadAction(fileResult, targetPath)
 }
 
-export function fileResultUploadAction(fileResult: FileResult, targetPath: string): EditorAction {
+export function fileResultUploadAction(
+  fileResult: FileResult,
+  targetPath: string,
+  replace: boolean = false,
```
Rheeseyb, line 16: How much work is it to make this param non-optional? If it's not a lot, please can you do that?

enidemi: fixed

seanparsons (approved):

Rheeseyb (approved):

PR #1245, opened 2021-05-12
Close deleted files in VS Code.

### LINKED ISSUE #1221: [vscode] moving an open file in the filebrowser to a different folder, vscode shows it as deleted

seanparsons:
Fixes #1221

**Problem:**
When files are moved around in the editor, VS Code gets confused and for something which has disappeared will show the file as deleted.

**Fix:**
Since there's no way to sensibly track that a file has moved in the main editor _and_ then synchronise that up with the state in VS Code once all the file changes have been applied, it seemed better to close the tabs so that a user wont be typing into the abyss.

**Notes:**
- The only way to close tabs is...dubious...as it involves making a file the active editor and then closing the active editor. By opening the deleted/moved file we even get an error in the console which feels slightly disappointing.

**Commit Details:**
- Fixes #1221.
- Adds `watchForFileDeletions` which performs a slightly unusual
  way of closing the editors for deleted files.

### DIFF
--- a/utopia-vscode-extension/src/extension.ts
+++ b/utopia-vscode-extension/src/extension.ts
@@ -43,6 +43,26 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
   watchForChangesFromVSCode(context, projectID)
 
   sendMessage(sendInitialData())
+
+  watchForFileDeletions()
+}
+
+function watchForFileDeletions() {
+  let fileWatcherChain: Promise<void> = Promise.resolve()
+  const fileWatcher = vscode.workspace.createFileSystemWatcher('**/*')
+  fileWatcher.onDidDelete(async (deletedFile) => {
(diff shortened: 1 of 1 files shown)


enidemi:
did you test what happens on moving/deleting a file with unsaved vscode changes?

seanparsons:
@enidemi Hmmm, testing that now and it doesn't work, that's disappointing. Will investigate to see if there's a bug fix possible.

Rheeseyb:
> did you test what happens on moving/deleting a file with unsaved vscode changes?

We had to deal with a very similar issue with unsaved content, so I've just had a quick chat with Sean about this

seanparsons:
> did you test what happens on moving/deleting a file with unsaved vscode changes?

Fixed in 0690ee63.

Rheeseyb (approved):

enidemi (approved):

PR #1249, opened 2021-05-13
Removing so much `imports` 

balazsbajorics:
**Problem:**
We had some metadata utility functions such as `isAnimatedElement` `isImportedComponent` and `isFocusableComponent` which required the `Imports` object. Once we moved to multi-file editing, getting  the imports object became nontrivial, so functions such as `useNamesAndIconsAllPaths` would burn a lot of CPU by first getting all valid paths and then executing, among other things `normalisePathToUnderlyingTarget` for each path. This meant that for every user interaction we would execute `normalisePathToUnderlyingTarget` thousands of times in a project like the Majestic Broker.

**Fix:**
I added a new `importInfo: ImportInfo | null` field to `ElementInstanceMetadata`. This importInfo contains everything that is needed for the utility functions so they no longer need to depend on Imports. 

**Commit Details:**
- New field `importInfo: ImportInfo | null` in `ElementInstanceMetadata` 
- New function `importInfoFromImportDetails`
- Removed the dependency on `imports: Imports` in 5-6 helper functions
- This led to the imports no longer being needed in _a ton_ of functions
- This led to me being able to kill `getJSXComponentsAndImportsForPathInnerComponentFromState`
- Also remove a lot of uses for `withUnderlyingTarget` and co.


editor/src/core/shared/either.ts
```diff
@@ -119,6 +119,14 @@ export function eitherToMaybe<L, R>(either: Either<L, R>): R | null {
   }
 }
 
+export function maybeEitherToMaybe<L, R>(either: Either<L, R> | null | undefined): R | null {
```
seanparsons, line 122: I feel a little uncomfortable with this, it's a little _too_ custom to be something that should be in here.

seanparsons (approved):

editor/src/core/shared/either.ts
balazsbajorics: nice catch, I actually wanted to ping you about this! 

The problem is that without this it is very hard to write concise code with `?.` operators. Especially if the eitherToMaybe is in the middle of an expression, it forces you to break it out into a ternary.

But I would be happy if you can come up with a nicer solution.

This I think is a good example of code that `maybeEitherToMaybe` makes possible:
`const componentPackageName = maybeEitherToMaybe(componentMetadata?.importInfo)?.path`

The closest alternative I guess would be
`const componentPackageName = optionalMap(eitherToMaybe, componentMetadata?.importInfo)?.path`

but at least for me this is just over the threshold and crosses into hard-to-parse-at-a-glance territory

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1449,3 +1421,9 @@ export function getSimpleAttributeAtPath(
     propsOrAttributes,
   )
 }
+function isGivenUtopiaElementFromMetadata(
```
balazsbajorics, line 1424: This is why the tests break.

I had no recollection of ever writing this function (instead of importing the real function I wrote in another file)

Turns out I clicked on the yellow bulb from muscle memory, and instead of importing the function, it HELPFULLY DECLARED IT FOR ME WTF

image

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1335,23 +1319,20 @@ export const MetadataUtils = {
...
-      elementName != null && isAnimatedElementAgainstImports(elementName, imports)
+    const isAnimatedComponent = isAnimatedElement(element)
     if (isAnimatedComponent) {
       return false
     }
-    const isImported = elementName != null && isImportedComponent(elementName, imports)
+    const isImported = isImportedComponent(element)
+    // BALAZS TODO BEFORE MERGE the original code checked the element being imported in the file it was defined in. the new code checks it against imports of the file it is _used in_.
+    // this is trouble because it means that for multifile components which are _imported_ from other files, this will return isImported === true, and we will conclude that they are not focusable.
+    // I think the solution should be to use isImportedComponentNPM that says if the component is imported from npm and then return false
```
balazsbajorics, line 1335: this is where I'll pick this up tomorrow

editor/src/core/shared/either.ts
seanparsons: For me, it feels like a lens'y thing that you're asking for.

```
maybeEitherToMaybe :: Maybe (Either a b) -> Maybe b
maybeEitherToMaybe = firstOf (_Just . _Right)
```

In that the `_Just . _Right` is the lens.

seanparsons: There's an argument that this `maybeEitherToMaybe(componentMetadata?.importInfo)?.path` is a little too dense to read, because it's optionally getting a field, transforming that and if that result is available getting a field within that value.

Rheeseyb: This to me definitely feels like it's falling into the realm of trying to achieve too much in a single line of code. All of these are because `ElementInstanceMetadata.element` is of type `Either<string, JSXElementChild>`, and I'm now wondering if that even makes sense anymore? If that does still make sense, then probably a clearer way to handle this would be to add functions to `element-metadata-utils` for e.g. `getComponentPackageNameFromMetadata`

enidemi (approved):

editor/src/core/model/project-file-utils.ts
```diff
@@ -175,28 +212,48 @@ function isHTMLComponentFromBaseName(baseName: string, imports: Imports): boolea
...
-export function isImportedComponent(jsxElementName: JSXElementName, imports: Imports): boolean {
-  return Object.keys(imports).some((importKey) => {
-    const fromImports = imports[importKey]
-    return (
-      pluck(fromImports.importedFromWithin, 'name').includes(jsxElementName.baseVariable) ||
-      fromImports.importedWithName === jsxElementName.baseVariable
+export function importInfoFromImportDetails(name: JSXElementName, imports: Imports): ImportInfo {
+  const baseVariable = name.baseVariable
+
+  const err = mapDropNulls((pathOrModuleName) => {
```
Rheeseyb, line 218: `err` - couldn't think of a name? :grinning:

Also this probably should not be a `mapDropNulls` followed by a `[0]` indexing, though I appreciate this will be messy either way.

How about (using `forEachValue` from `object-utils.ts`):
```
let importInfo: ImportInfo | null = null
forEachValue((importDetail, pathOrModuleName) => {
  if (importInfo == null) {
    const importAlias = importDetail.importedFromWithin.find(
      (fromWithin) => fromWithin.alias === baseVariable,
    )

    if (importAlias != null) {
      importInfo = createImportedFrom(importAlias.alias, importAlias.name, pathOrModuleName)
    } else if (importDetail.importedAs === baseVariable) {
      importInfo = createImportedFrom(importDetail.importedAs, null, pathOrModuleName)
    } else if (importDetail.importedWithName === baseVariable) {
      importInfo = createImportedFrom(importDetail.importedWithName, null, pathOrModuleName)
    }
  }
}, imports)

return importInfo ?? createNotImported()
```

Rheeseyb (approved):
This is honestly beautiful :heart:

editor/src/core/model/project-file-utils.ts
balazsbajorics: > err - couldn't think of a name? 😀

haha I just made it a const for debugging and left it by accident :D 

I actually wanted to bring this up in the last couple of weeks!
```typescript
let importInfo: ImportInfo | null = null
forEachValue((importDetail, pathOrModuleName) => {
```
I really do not like that more and more places in the codebase we are "forced" to use lets and mutate them. Especially if it's a function that is longer than a dozen lines, it just invites people to do the wrong thing. When I was refactoring the dom-walker and previously the inspector hooks, I had to fight with this a lot. The problem is that this `let` is just too inviting for people down the line who want to append some functionality to overwrite the value 30-40 lines later, embedded in an if. It makes code hard to refactor, and I think these are the functions that eventually end up being a 200 line spagetthi over time because small changes are easiest to make the worst possible way.

PR #1251, opened 2021-05-13
More granular dom walker caching

Rheeseyb:
More performance improvement work towards #1232 

**Problem:**
We are unnecessarily running expensive measurements on every descendant of a given scene whenever a descendant of that scene is being dragged on the canvas.

**Fix:**
If the user is performing a canvas interaction, fall back to cached values for everything other than the affected elements.

**Commit Details:**
- In `ui-jsx-canvas.tsx` check if `props.transientFilesState != null` to determine if an interaction is occurring
- In `dom-walker.ts`
  - `useResizeObserver` and `useMutationObserver` have been updated to only return the selected views if an interaction is occurring (since we only want to invalidate the cache for those)
  - Moved the cached value retrieval down into `collectMetadata` so that we can support invalidating specific paths rather than just entire scenes (previously this would happen at the scene level and then stop any descending into the scene completely if the scene itself hadn't been invalidated)
  - Invalidating a scene still has special cased logic for the sake of invalidating everything below that scene when a selection changes. This has been updated so that it no longer runs during canvas interactions.
- Updated some helper functions to support getting the stringified and non-stringified paths for an element (to save us from having to make extra unnecessary `EP.toString` calls when checking if a path has been invalidated)

**Note:** Having profiled this whilst running the resize test, it cuts the dom walking time in half during a dragged resize on the project (I tested by selecting the root `div` of the `card` element and running the drag test on that, which was showing ~10ms dom walking time before, ~5ms after). ~I've just however realised that the puppeteer tests select the very topmost navigator row, which will update the scene itself, and thus invalidate everything below it :weary: I think I'll revisit the scene invalidation special cased logic in the morning, since that should only apply on selection changes~ (Done)


seanparsons:
> If the user is performing a canvas interaction, fall back to cached values for everything other than the affected elements.

If a sibling of element A is resized and that means element A moves/resizes as a result, would this change result in element A having some invalid dom walker results?

editor/src/components/canvas/dom-walker.ts
```diff
@@ -179,14 +179,24 @@ function getCachedAttributesComingFromStyleSheets(
...
+  canvasInteractionHappening: React.MutableRefObject<boolean>,
+) {
   const resizeObserver = React.useMemo((): ResizeObserver | null => {
     if (ObserversAvailable) {
       return new ResizeObserver((entries: any) => {
-        for (let entry of entries) {
-          const sceneID = findParentScene(entry.target)
-          if (sceneID != null) {
-            invalidatedSceneIDsRef.current.add(sceneID)
+        if (canvasInteractionHappening.current) {
```
enidemi, line 191: I think this will no longer give images good size, because they might receive their correct size after loading

Rheeseyb:
> > If the user is performing a canvas interaction, fall back to cached values for everything other than the affected elements.
> 
> If a sibling of element A is resized and that means element A moves/resizes as a result, would this change result in element A having some invalid dom walker results?

It would for the duration of the drag, but that would be corrected upon completion of the drag. Be aware that this is specifically only for the case that the user is interacting with something in a way that creates a transient file state.

editor/src/components/canvas/dom-walker.ts
enidemi: oops, this is only while dragging

enidemi (approved):

editor/src/components/canvas/dom-walker.ts
```diff
@@ -324,16 +366,17 @@ export function useDomWalker(props: DomWalkerProps): React.Ref<HTMLDivElement> {
...
         props.canvasRootElementElementPath,
         props.validRootPaths,
         rootMetadataInStateRef,
-        invalidatedSceneIDsRef,
+        invalidatedPathsRef,
+        invalidatedScenesRef,
         invalidatedPathsForStylesheetCacheRef,
         props.selectedViews,
-        initComplete,
+        !initComplete,
```
seanparsons, line 379: Was this wrong originally?

seanparsons: Ah, you've changed what the parameter means.

Rheeseyb: I have changed the param of the function this calls from `initComplete` (which if false would invalidate the cache for all scenes) to `invalidated` (which if true will invalidate the cache for everything from this point onwards)

seanparsons (approved):

Rheeseyb:
The failing tests are new tests coming from master that for updating selected views after an insertion. I'm fixing those now.

Rheeseyb:
> The failing tests are new tests coming from master that for updating selected views after an insertion. I'm fixing those now.

Fixed in bb75358

editor/src/components/canvas/dom-walker.ts
```diff
@@ -280,25 +300,47 @@ interface DomWalkerProps {
...
+  const invalidatedScenesRef = React.useRef<Set<string>>(emptySet()) // For invalidating entire scenes and everything below them
   const invalidatedPathsForStylesheetCacheRef = React.useRef<Set<string>>(emptySet())
   const [initComplete, setInitComplete] = useInvalidateInitCompleteOnMountCount(props.mountCount)
-  const resizeObserver = useResizeObserver(invalidatedSceneIDsRef)
-  const mutationObserver = useMutationObserver(invalidatedSceneIDsRef)
+  const selectedViewsRef = React.useRef(props.selectedViews)
+  const canvasInteractionHappeningRef = React.useRef(props.canvasInteractionHappening)
+
+  if (selectedViewsRef.current !== props.selectedViews) {
+    selectedViewsRef.current = props.selectedViews
```
balazsbajorics, line 324: I think there's no reason not to just always assign the value

Rheeseyb: Good point! I'll fix this.

editor/src/components/canvas/dom-walker.ts
```diff
@@ -374,64 +417,115 @@ function collectMetadataForElement(
...
+        pathsForElement,
+        stringPathsForElement,
+        parentPoint,
+        allUnfilteredChildrenPaths,
+        scale,
+        containerRectLazy,
+        invalidatedPathsRef,
+        invalidatedPathsForStylesheetCacheRef,
+        rootMetadataInStateRef,
+        true,
```
balazsbajorics, line 524: I think it's worth commenting here that this `true` makes it forced

balazsbajorics (commented):
👍

PR #1259, opened 2021-05-14
Update puppeteer-test.ts

maltenuhn:
**Problem:**
Having a varied range makes it hard to compare performance charts at a glance

**Fix:**
Fixes the range from 0 to 200ms, adds a label.

### DIFF
--- a/performance-test/src/puppeteer-test.ts
+++ b/performance-test/src/puppeteer-test.ts
@@ -355,7 +355,7 @@ async function createSummaryPng(
       },
     ],
     xaxis: {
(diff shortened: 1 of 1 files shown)


balazsbajorics:
good idea

balazsbajorics (approved):

maltenuhn:
If your theory is correct (which I believe it is) 200ms is still more than enough to spot incremental regressions. Alternatively, if we find a way to make selection fast, the whole chart would become easy to read :D

enidemi (approved):

seanparsons (approved):

Rheeseyb (approved):

PR #1263, opened 2021-05-14
Fix/name and icon hook

enidemi:
**Problem:**
Name and Icon hooks have memoization problems, one of them has no type check the other has no memoization at all.

**Commit Details:**
- using `combine4EqualityCalls` and `arrayDeepEquality`


balazsbajorics (approved):
so nice

Rheeseyb (approved):

PR #1266, opened 2021-05-14
Testing a pair of baseline values for scaling performance tests

Rheeseyb:
This PR was used to try to find a baseline metric to give us a picture of the overall performance of the machine running the tests, with the idea being that this baseline would strongly correlate to the other measurements enough that we could then use it to scale all other performance tests (thereby offsetting the high variance across runs of the performance tests).

I went with two proposals for this baseline metric: a simple calculation (calculating Pi); and a series of consecutive empty dispatches (since a single empty dispatch is too quick). I have tweaked these two tests a number of times and found that unfortunately they appear to be not very well correlated with the other tests (in many cases the baselines dropped whilst the other values rose, or vice versa).

Still, for the sake of completeness, I've added the tests to the chart, so that we can monitor them over time, tweaking them both so that they'll roughly result in values around the 30ms mark (in the case of the empty dispatch this is required so that it won't run faster than a frame takes to render).

This PR has also updated the other tests to actually gather 100 measurements (in each case they are run 101 times, since the tests will use the deltas between measurements to capture actual frame times, meaning we need 1 extra run).


balazsbajorics:
So what I was thinking about, before we trust any of these values is to add our two proposed scaling measurements to the chart (with the standard 100 runs for each instead of a single run) and just watch the values for a few days.  And not apply any scaling yet

Rheeseyb:
> So what I was thinking about, before we trust any of these values is to add our two proposed scaling measurements to the chart (with the standard 100 runs for each instead of a single run) and just watch the values for a few days. And not apply any scaling yet

Yeah this is pretty much what I was thinking too (I'm just testing the applied scaling in the PR to see how stable it is without having to manually calculate). I think the simple dispatch is triggering a garbage collection so I've made a tweak to how we use that value.

performance-test/src/puppeteer-test.ts
```diff
@@ -102,7 +102,77 @@ export const setupBrowser = async (): Promise<{
...
+
+  return result
+}
+
+function timeBasicCalc(): FrameResult {
+  let times: Array<number> = []
+
+  for (let i = 0; i < 100; i++) {
+    const start = Date.now()
+    calculatePi(7500000)
```
balazsbajorics, line 141: that's accurate

performance-test/src/puppeteer-test.ts
```diff
@@ -102,7 +102,77 @@ export const setupBrowser = async (): Promise<{
...
+  const analytics = {
+    frameMin: sortedTimes[0]!,
+    frameMax: sortedTimes[sortedTimes.length - 1]!,
+    frameAvg: Number((totalTime / sortedTimes.length).toFixed()),
+    percentile25: sortedTimes[Math.floor(sortedTimes.length * 0.25)],
+    percentile50: sortedTimes[Math.floor(sortedTimes.length * 0.5)],
+    percentile75: sortedTimes[Math.floor(sortedTimes.length * 0.75)],
+  }
+  return {
+    title: 'Calc Pi 5m',
```
balazsbajorics, line 158: what does 5m mean ?

balazsbajorics (approved):

performance-test/src/puppeteer-test.ts
Rheeseyb: (5m iterations, not 5m digits)

balazsbajorics: > (5m iterations, not 5m digits)

massively disappointing!

Rheeseyb: Look, I know what you're saying, and that I bring shame on my family, but even [NASA only uses pi correct to 15dp](https://www.jpl.nasa.gov/edu/news/2016/3/16/how-many-decimals-of-pi-do-we-really-need/):
> In other words, by cutting pi off at the 15th decimal point, we would calculate a circumference for that circle that is very slightly off. It turns out that our calculated circumference of the 25 billion mile diameter circle would be wrong by 1.5 inches. Think about that.

Think about that Balazs.

Issue #1261, opened 2021-05-14
Improve preview loading and refresh speed


seanparsons:

seanparsons:
Between #1267 and #1271 this should be improved somewhat, as those drop the size of the bundles by some amount.

The critical path of the preview loading is as follows:
- Download preview code.
- Compile (browser) and run preview code.
- Download tsWorker code.
- Compile (browser) and run tsWorker code.
- tsWorker emits the project code.
- Compile (browser) and run the project code.

Potentially given that there's no concurrency to these it might be worth having the preview code handle the project code emitting without the worker.

PR #1267, opened 2021-05-14
Prevent imports from .ts files to .tsx files.

seanparsons:
**Problem:**
It was surmised that some of our dependency chains was causing too much to be included in some of the various bundles we produce for the editor.

**Fix:**
Using a slightly cheap trick way of only having `.tsx` files import `.ts` files and not vice-versa, this has created some slight separation in the dependencies, along with also alleviating some of the circular dependencies in the editor.

**Commit Details:**
- Partly addresses #1261.
- Adds a dependency cruiser rule to prevent importing from `.ts` files
  to `.tsx` files, with an exception for `built-in-dependencies.ts`.
- Renames some files in either direction trying to get the rule to be
  satisfied.
- Moved some global mouse position values into a shared file to break
  some dependency chains.
- A few functions and types were moved to prevent the need for some core
  logic to depend on a `.tsx` file.


enidemi (approved):

editor/src/components/text-utils.ts
```diff
@@ -247,7 +247,7 @@ export async function measureTextFieldNew(element: ElementInstanceMetadata): Pro
       },
       scale: 1,
     }
-    const textElement = <Text {...textFieldProps} />
+    const textElement = React.createElement(Text, textFieldProps)
```
Rheeseyb, line 250: This one is definitely cheating :eyes:

seanparsons: This whole function put the fear into me, promise based modifying of the DOM that then triggers actions to be dispatched back into the editor.

Rheeseyb: Ha ok I just followed this through and see why you've done this! I was going to ask for this to be removed but given how this is used I'd say it's fine to leave as is.

Rheeseyb (approved):
I really like this. I'd say it's worth disabling the rule for tests though to save us having to deal with renaming .spec files since those don't affect the goal.

PR #1271, opened 2021-05-17
Shrink the bundles by re-using dependencies.

seanparsons:
Partly addresses #1261.

**Problem:**
When looking at the webpack bundle contents it turned out that the `@babel/types` dependency was embedded into nearly 20 projects, bloating the output quite a lot.

**Fix:**
Deleting all the various cases where `@babel/types` was repeated in `package-lcok.json` and then letting a true instance be re-created. This appears to save ~2.5mb of download cost when loading the editor and preview.

**Commit Details:**
- Removed the `@babel/types` dependency from `editor/package.json`,
  as it's not directly in use by the editor itself.
- Removed the various instances of `@babel/types` in `dependencies`
  elements in `package-lock.json` and allowed `npm install` to
  re-establish the transitive dependency.
- Added `webpack-bundle-analyzer` so that people can produce reports
  for the webpack bundles.
- Used `weback-node-externals` in `utopia-vscode-common` project
  so that the contents of the bundle aren't opaque to the analyzer.


Rheeseyb (commented):
Looks like we actually explicitly need that package. Does updating the version in utopia's `package.json` undo the package-lock tidying?

seanparsons:
> Looks like we actually explicitly need that package. Does updating the version in utopia's `package.json` undo the package-lock tidying?

Yup, just in the process of re-adding it.

seanparsons:
> > Looks like we actually explicitly need that package. Does updating the version in utopia's `package.json` undo the package-lock tidying?
> 
> Yup, just in the process of re-adding it.

Oh and no, the tidying is unchanged by re-adding it.

Rheeseyb (approved):

balazsbajorics (approved):

PR #1278, opened 2021-05-18
fix(inspector) layoutsection and inspector props

enidemi:
**Fix:**
Cleaning up some ancient code called InspectorModel.

**Commit Details:**
- still found some layoutwrapper related variables here
- removed inspectorModel, moved the only thing used from it: specialSizeMeasurements to layout-section
- removed underlying path resolving from inspector.tsx, because for the CSSTarget this could be extracted from the metadata
- changed props about the flex direction in the layout-section components because now it uses the specialsizemeasurements metadata

### DIFF
--- a/editor/src/components/inspector/sections/layout-section/layout-section.tsx
+++ b/editor/src/components/inspector/sections/layout-section/layout-section.tsx
@@ -1,42 +1,47 @@
 import * as React from 'react'
-import { CanvasRectangle, LocalRectangle } from '../../../../core/shared/math-utils'
 import { LayoutSystemSubsection } from './layout-system-subsection/layout-system-subsection'
 import { SelfLayoutSubsection } from './self-layout-subsection/self-layout-subsection'
-import { CSSPosition } from '../../common/css-utils'
-import {
-  DetectedLayoutSystem,
-  SpecialSizeMeasurements,
-} from '../../../../core/shared/element-template'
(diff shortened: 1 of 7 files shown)


editor/src/components/inspector/sections/layout-section/layout-section.tsx
```diff
@@ -1,42 +1,47 @@
...
-  parentFlexAxis: 'horizontal' | 'vertical' | null
   aspectRatioLocked: boolean
   toggleAspectRatioLock: () => void
 }
 
 export const LayoutSection = betterReactMemo('LayoutSection', (props: LayoutSectionProps) => {
+  const specialSizeMeasurements = useEditorState(
+    (state) => {
+      let foundSpecialSizeMeasurements = emptySpecialSizeMeasurements
+      fastForEach(state.editor.selectedViews, (path) => {
```
balazsbajorics, line 21: AFAIK the rest of the inspector uses a target path property that is independent of selection – we could use that here too, or at least put a comment here noting the divergence!

seanparsons (approved):

maltenuhn:
I'm merging this to help with debugging (since it probably addresses some of the things I've seen). I've created a follow-up ticket (https://github.com/concrete-utopia/utopia/issues/1297) for your comment @balazsbajorics

PR #1284, opened 2021-05-20
Remove Ramda.

seanparsons:
**Problem:**
Ramda is slow and often has broken types we have to work around.

**Fix:**
Blast Ramda into the sun.

**Commit Details:**
- Remove dependency on `ramda`.
- Backfilled some missing utility functions that we didn't have.
- Replaced all the calls to Ramda with alternatives.


editor/src/components/canvas/canvas.ts
```diff
@@ -122,7 +121,11 @@ const Canvas = {
             if (working == null) {
               return selectedView
             } else {
-              return R.minBy(EP.depth, selectedView, working)
```
balazsbajorics, line 125: Utopia clearly had a ramda phase around 2016-2017 :)

editor/src/components/canvas/controls/svg-control-component.tsx
```diff
@@ -164,22 +163,24 @@ const controlToSvg = (control: SvgFragmentControl, index: number) => {
...
       RU.create(
         control.type,
-        R.merge(control.props, {
+        {
+          ...control.props,
           id: `control-${control.controlid}`,
           fill: `url(#${patternId})`,
           key: `control-${control.controlid}-${index}`,
-        }),
+        } as any,
```
balazsbajorics, line 171: have you tried `as const` ? sometimes it makes React realize that we are using the proper string literals…

editor/src/core/shared/array-utils.ts
```diff
@@ -285,3 +305,45 @@ export function safeIndex<T>(array: Array<T>, index: number): T | undefined {
...
+            break
+          }
+        }
+        if (shouldAdd) {
+          result.push(valueFromFirst)
+        }
+      }
+    }
+    if (foundIntersection) {
+      continue
```
balazsbajorics, line 333: If I read this correctly, this continue will do nothing as this is the last statement of the `for` loop

editor/src/components/canvas/controls/svg-control-component.tsx
seanparsons: This appeared to be because of the `id` field, I just attempted to maintain what was already there.

editor/src/core/shared/array-utils.ts
seanparsons: Hah, that's hilarious, I must've had something after this originally.

seanparsons: Fixed in 82e3a90.

balazsbajorics (approved):

PR #1288, opened 2021-05-20
Accumulate changes to VS Code.

### LINKED ISSUE #1253: VSCode "replays" highlight / selection changes with a delay if performance is bad
In situations when the editor slows down, if you click around to change the selection, you can see VSCode "replaying" your selection changes with a delay, instead of skipping to the last selected/highlighted element.

I think this is worst on Dev mode and slow computers, but it is pretty annoying nonetheless

seanparsons:
Fixes #1253

**Problem:**
Previously we've stacked promises for every message send to VS Code, but that means when the editor gets really busy those promises don't get executed until some amount of idle time. Which results in a bunch of selected element changed events firing one after the other.

**Fix:**
In this PR the regularly dispatched updates get accumulated on the main thread and are sent in as reasonably minimal set of promises as possible, but primarily so that only 0 or 1 updates to decorations or selected element changes are fired. Thereby preventing the "replay" effect described in the bug.

**Notes:**
Potentially in the future file changes could also be compressed so that only the latest write or delete for each filename is sent, but since that was out of the scope of this work it wasn't done here.

**Commit Details:**
- Fixes #1253.
- Moved `ProjectChange`, the associated types and `collateProjectChanges`
  into `dispatch.tsx` to dodge the vscode-bridge mocking.
- Created `AccumulatedVSCodeChanges` to hold updates that are due to
  go over the wire to VS Code.
- Changes accumulated into the `currentVSCodeChanges` variable as they
  are dispatched directly via the main thread without promises being involved.
- Accumulation is done by applying the `combineAccumulatedVSCodeChanges` to
  the current and new instances of `AccumulatedVSCodeChanges`, akin to a
  monoid with `emptyAccumulatedVSCodeChanges`. This adds up all the file changes
  but for decorations and the selected item changed takes the latest update
  if there is one.
- `applyProjectChangesCoordinator` now just resets the `currentVSCodeChanges`
  and sends over the accumulated non-file changes in one message.
- Added the `AccumulatedToVSCodeMessage` interface for holding a bunch of
  messages to process at once, along with the necessary switch case to run
  each message in turn.


editor/src/components/editor/store/dispatch.tsx
```diff
@@ -87,6 +102,149 @@ import {
...
+export function ensureDirectoryExistsChange(fullPath: string): EnsureDirectoryExistsChange {
+  return {
+    type: 'ENSURE_DIRECTORY_EXISTS',
+    fullPath: fullPath,
+  }
+}
+
+export type ProjectChange = WriteProjectFileChange | DeletePathChange | EnsureDirectoryExistsChange
+
+export function collateProjectChanges(
```
balazsbajorics, line 155: I think this 200-300 lines of new code here are enough to warrant a dedicated file!

seanparsons: Not new code. :)

balazsbajorics (approved):

editor/src/components/editor/store/dispatch.tsx
balazsbajorics: _here_ it's new code :P 

but alas, my point was that I think this burdens dispatch.ts and instead could live in a dedicated file

seanparsons: Sure, I'll try to get as much of this into another file.

seanparsons: Fixed in a2eaa373.

balazsbajorics (approved):

PR #1290, opened 2021-05-21
Chore: disallow much truthiness

balazsbajorics:
**Problem:**
We had an informal coding style that we only write code like this:

`if (variable) {`
or 
`variable ? 'hello' : null`

if `variable` is a boolean. If `variable` in these examples is not a boolean, that means we rely on javascript truthiness (javascript will coerce anything into a boolean if we don't watch out).

Sometimes relying on truthiness can cause problems: imagine if we refactor a `isOpen: boolean` into `openState: 'open' | 'closed' | 'half-open'`. code that now says `if (openState)` will pass typescript type check, but it will **always** evaluate to true! 

I realized there's an eslint rule that can help us out https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md

so I enabled it, and fixed the ~200 problems it found.

About 6 or 7 of these were real bugs that I just caught using the lint rule! Not bad!


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -924,7 +924,7 @@ export function getOriginalFrameInCanvasCoords(
   originalFrames: Array<OriginalCanvasAndLocalFrame>,
   target: ElementPath,
 ): CanvasRectangle | null {
-  for (const originalFrame of originalFrames || []) {
+  for (const originalFrame of originalFrames ?? []) {
```
seanparsons, line 927: We probably should just null check this, instead of putting in an empty array that we then iterate over.

editor/src/components/editor/actions/actions.ts
```diff
@@ -1382,7 +1382,7 @@ function getZIndexOrderedViewsWithoutDirectChildren(
   // keep direct children from reparenting
   let filteredTargets: Array<ElementPath> = []
   Utils.fastForEach(orderedTargets, (target) => {
-    if (!orderedTargets.find((tp) => EP.pathsEqual(EP.parentPath(target), tp))) {
+    if (orderedTargets.find((tp) => EP.pathsEqual(EP.parentPath(target), tp)) == null) {
```
seanparsons, line 1385:
```suggestion
    if (!orderedTargets.some((tp) => EP.pathsEqual(EP.parentPath(target), tp))) {
```

editor/src/components/canvas/canvas-utils.ts
balazsbajorics: but that's two more lines of code! :D

editor/src/components/navigator/dependency-list.tsx
```diff
@@ -267,7 +267,7 @@ class DependencyListInner extends React.PureComponent<DependencyListProps, Depen
       )
     }
 
-    if (DefaultPackagesList.find((pkg) => pkg.name === packageName)) {
+    if (DefaultPackagesList.find((pkg) => pkg.name === packageName) != null) {
```
seanparsons, line 270:
```suggestion
    if (DefaultPackagesList.some((pkg) => pkg.name === packageName)) {
```

seanparsons (approved):

enidemi (approved):

PR #1291, opened 2021-05-21
Do not reload vscode when forking a project

### LINKED ISSUE #1289: Make VSCode / canvas not reload after forking
**Problem**
We currently fork projects on first-edit, in line with other online coding environments. However, this causes the project ID to change, which in turn forces a VSCode reload, meaning your second and subsequent keystrokes go into the void. 

What should happen instead is that you can keep typing or making canvas changes, and the fork happens almost-invisibly to you. 

**Solution Notes**
> I haven't fully thought this through, but if we generated some kind of proxy ID for/from the project ID and used that as the identifier in VS Code, then when we fork we could just say that for the new project ID that identifier is the proxy ID. ... We'd have to make it so that if you reloaded the original project it didn't get that ID again, but I think that would work.

https://paradise-engineering.slack.com/archives/C0J0RRN93/p1621518412019400

balazsbajorics:
Fixes #1289

**Problem:**
When forking the project, VSCode would reload. This would mean the user had to wait long seconds AND manually open the edited file again so they can continue editing. Urgh!

**Fix:**
The project ID and vscode's bridge ID were coming from the same editorState.id . The fix is to split `id` and `vscodeBridgeId` into two separate fields. For the sake of not changing too many things, `vscodeBridgeId` is initialized to `editor.id`, but subsequent changes to editor.id (such as when forking a project) do not update `vscodeBridgeId`, because there is no need. This means that the vscode instance can survive an ID change, and keep on working.

**Commit Details:**
- Added `vscodeBridgeId` to editor-state, LOAD initializes it to editor.id
- SET_PROJECT_ID does NOT change `vscodeBridgeId`, does NOT call `initVSCodeBridge`
- `triggerForkProject` does not call `LOAD` anymore, I realized it can get by with only dispatching `SET_PROJECT_ID` and `SET_FORKED_FROM_PROJECT_ID`

### DIFF
--- a/editor/src/components/canvas/controls/insert-mode-control-container.tsx
+++ b/editor/src/components/canvas/controls/insert-mode-control-container.tsx
@@ -73,7 +73,6 @@ function forceSetValueAtPath(
 interface InsertModeControlContainerProps extends ControlProps {
   mode: InsertMode
   keysPressed: KeysPressed
-  projectId: string | null
   dragState: InsertDragState | null
   canvasOffset: CanvasVector
   scale: number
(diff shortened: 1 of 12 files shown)


seanparsons (approved):

balazsbajorics:
I'm gonna merge this now!

PR #1301, opened 2021-05-24
Check the props of scene children.

### LINKED ISSUE #1293: Making a change via the formula bar updates the code, but not the canvas
**Problem**
Making changes to the code via the formula bar (see video) and sometimes the inspector fails, without an obvious problem in the console. See video here, with Redux devtools open. 

(redux devtools export available here https://www.dropbox.com/s/6xcjes9a0pz60jf/canvas%20not%20updating%20json.txt?dl=0
)

https://user-images.githubusercontent.com/2945037/119233738-51c3ca00-bb22-11eb-954d-72e198105c2e.mp4

seanparsons:
Fixes #1293

**Problem:**
Changes to the children of scene elements would not cause changes to the canvas to apply.

**Fix:**
The props of the scene element children need to be included so that changes to their children and changes to things like their `style` property to cause them to re-render.

**Commit Details:**
- Fixes #1293.
- `childUnchanged` now also includes a shallow equality check against
  the props of the scene component children.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/scene-component.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/scene-component.tsx
@@ -5,7 +5,6 @@ import { Scene, SceneProps } from 'utopia-api'
 import { colorTheme, UtopiaStyles } from '../../../uuiui'
 import { betterReactMemo } from '../../../uuiui-deps'
 import { RerenderUtopiaContext } from './ui-jsx-canvas-contexts'
-import { shallowEqual } from '../../../core/shared/equality-utils'
 
 export const SceneComponent = betterReactMemo(
   'Scene',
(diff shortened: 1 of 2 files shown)


enidemi (approved):

balazsbajorics (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/scene-component.tsx
```diff
@@ -54,7 +55,11 @@ function childUnchanged(prevChild: ReactChild, nextChild: ReactChild): boolean {
   if (typeof prevChild === 'string' || typeof prevChild === 'number') {
     return nextChild === prevChild
   } else if (React.isValidElement(prevChild)) {
-    return React.isValidElement(nextChild) && prevChild.type === nextChild.type
+    return (
+      React.isValidElement(nextChild) &&
+      prevChild.type === nextChild.type &&
+      shallowEqual(prevChild.props, nextChild.props)
```
balazsbajorics, line 61: actually this should probably be `deepEqual` because props.style (if present) will blow shallow equality

seanparsons: Fixed in 1b203a4e.

PR #1306, opened 2021-05-26
Improvement: agressively sourcemapped errors

balazsbajorics:
**Problem:**
If a runtime error is thrown when the code is evaluated (like an expression using a variable that is not defined), we show a nice error with a nice source code snippet.
However, if the runtime error happens later when React tries to call the component (if the source code is a function component and the function component is only called by react, not us), the error is almost not actionable: we don't get to know anything other than the error's name:
image

This is hostile to novice users and not how codesandbox / create-react-app works!

It turns out that the problem is that the SafeFunctionCurriedErrorHandler's errorHandler is only called if the code throws an error immediately on evaluation. If the code evaluates correctly, and that code later throws an error in usage, the error would go to the CanvasErrorBoundary and that means we couldn't apply our `processErrorWithSourceMap` function which eats the error, and decorates it with the nice StackFrames based on the source map.

**Fix:**
I realized that the only place where I can sneak information into Chrome's runtime error was the `//# sourceURL=` magic comment, which becomes the "filename" in the error stack:
image
in the screenshot, `canvas-react-utils.ts` and `react-dom.development.js` comes from Chrome parsing the `sourceURL` and trusting its value.
I learned that I can literally put any valid string in here, including a base64 encoded version of the entire encoded source map!
I can use this later when I unpack the error message, to extract the sourcemap from here and use it in `processErrorWithSourceMap`! 
This means that `SafeFunctionCurriedErrorHandler` now converts the soruceMap into base64 and squirrels it away into the evaluated source code's `sourceURL` field. Later, in `CanvasErrorBoundary`'s `componentDidCatch` I can now call `processErrorWithSourceMap` because the Error object itself is enough for processErrorWithSourceMap to extract the source map. This means that when an evaluated component throws a runtime error mid-render, we now have the ability to show an intelligible error message:
image

**Commit Details:**
- Abusing `//# sourceURL=` to include an entire base64 encoded sourcemap to the file
- removing the dependencies on `code` and `rawSourceMap` in `processErrorWithSourceMap`, so we can call it from  `CanvasErrorBoundary`
- calling `processErrorWithSourceMap` from `CanvasErrorBoundary`

**Note:**
The only downside of our abuse of `sourceURL` is that in the real Chrome console, the error message printed by react becomes pretty scary:
image
I consider this to be a cosmetic issue, we can certainly improve on this by only including a key to a source map cache here, and maintain a source map cache instead of serializing the entire map into the fileName field. But I think this PR is good enough as it is, we can improve it later if we want to.

### DIFF
--- a/editor/package-lock.json
+++ b/editor/package-lock.json
@@ -4418,6 +4418,11 @@
         "@react-spring/shared": "9.0.0-rc.3"
       }
     },
+    "@root/encoding": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@root/encoding/-/encoding-1.0.1.tgz",
+      "integrity": "sha512-OaEub02ufoU038gy6bsNHQOjIn8nUjGiLcaRmJ40IUykneJkIW5fxDqKxQx48cszuNflYldsJLPPXCrGfHs8yQ=="
+    },
     "@samverschueren/stream-to-observable": {
       "version": "0.3.0",
       "resolved": "https://registry.npmjs.org/@samverschueren/stream-to-observable/-/stream-to-observable-0.3.0.tgz",
--- a/editor/package.json
+++ b/editor/package.json
@@ -94,6 +94,7 @@
     "@emotion/styled": "11.0.0",
     "@emotion/styled-base": "11.0.0",
     "@popperjs/core": "2.4.4",
+    "@root/encoding": "1.0.1",
     "@seznam/compose-react-refs": "^1.0.4",
     "@svgr/plugin-jsx": "5.5.0",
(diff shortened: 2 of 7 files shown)


enidemi (approved):

balazsbajorics:
performance test is back! yay

seanparsons (approved):
Is it possible to have some kind of test for this as it seems like something we could easily break?

balazsbajorics:
> Is it possible to have some kind of test for this as it seems like something we could easily break?

AFAIK not with Node.js because that has a different stack trace format... but maybe some Karma tests later today! :)

PR #1312, opened 2021-05-27
Test/canvas add karma test runner

enidemi:
Creating a draft PR first to see how it runs on the CI

**Problem:**
Our tests are running in node js environment, while we expect the editor to work in the browser. This issue became visible when writing tests for canvas errors, the node error stacktraces and the browser stacktraces are not the same, making it not possible to write good tests. 

**Fix:**
Trying to add karma to run some of our tests. These tests will be called `browser2` tests in their filenames.
Unfortunately the nice inline snapshots from jest cannot be used here.

**Commit Details:**
- karma packages with webpack and mocha
- karma config file and a small js file to include BrowserFS
- remove ui-js test-utils to depend on `jest` availability


balazsbajorics (approved):

seanparsons:
> Unfortunately the nice inline snapshots from jest cannot be used here.
Can regular snapshots be used?

seanparsons (approved):
Does this mean we can use karma for everything and fling jest into a ditch?

balazsbajorics:
> > Unfortunately the nice inline snapshots from jest cannot be used here.
> 
> Can regular snapshots be used?

regular snapshots are theoretically not possible, because those rely on `fs` to read the snapshot file.

I think the best solution will be to make something similar to the inline snapshots (except the "automagically updating" part, of course) where you can copy-past a stringified value from the terminal into the `expect` line to update a snapshot.

enidemi:
> > Unfortunately the nice inline snapshots from jest cannot be used here.
> 
> Can regular snapshots be used?

There are karma plugins specifically for snapshots, but no support for inline snapshots.

balazsbajorics:
> Does this mean we can use karma for everything and fling jest into a ditch?

I _think_ this means that we have a very good chance of upgrading all `.spec.browser` tests to karma. 

I feel like Jest is still better for regular unit tests, in general Jest has nicer tooling (the watch mode with pattern matching, nice error output, the console log sanitizer we started using etc etc).

Plus, we are yet to give this Karma-based test suit a performance check – I _think_ it's as fast (or faster) than electron, but for regular unit tests, I'm not sure how much parallelization we can expect here

seanparsons:
> I feel like Jest is still better for regular unit tests, in general Jest has nicer tooling (the watch mode with pattern matching, nice error output, the console log sanitizer we started using etc etc).

Unless there's a _really_ strong reason to have 2 different test runners, I'd rather there just be the one so we don't have two things to maintain.

PR #1316, opened 2021-05-28
Don't remove or rewrite selected views when the code breaks.

### LINKED ISSUE #1310: saving the code file sometimes causes selection loss
**Problem**
Occasionally, saving a code file modified and then saved (CMD+S) causes selection loss

**Video**
(I've included the copy/paste job of the sample project in case it has any bearing on anything, but the action starts at about 40 seconds in)

https://user-images.githubusercontent.com/2945037/119820342-463d1e00-bee9-11eb-9250-e347b47137dc.mp4

Redux Devtools json: https://www.dropbox.com/s/4z478wq6d035iro/devtools-selection-loss-after-save.json?dl=1

seanparsons:
Fixes #1310.

**Problem:**
At certain points with a particular timing if the code would produce an error the `navigatorTargets` end up empty, if the mouse position trigger came in at that point the selected views would be set to an empty array.

**Fix:**
The solution was to attempt to keep the selected items as much as possible, by both avoiding updating the selection if the canvas appears to be broken and removing the call to the function that eliminates the "invalid" selections and highlighted views if they don't exist anymore.

**Commit Details:**
- Fixes some unrelated issues in `processErrorWithSourceMap` that logged
  out a lot of console errors.
- Ensures that backspace in VS Code still triggers a cursor position
  changed event as apparently they come without a `kind` property.
- Removed the call to `removeNonExistingViewReferencesFromState` as
  we want those fields to survive a broken canvas.
- Refactored out the logic for updating the selection based on the
  cursor position and added in check to see if the canvas appears to be
  broken current, which doesn't update the selected elements.

### DIFF
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -1447,6 +1447,32 @@ function toastOnGeneratedElementsTargeted(
 let checkpointTimeoutId: number | undefined = undefined
 let canvasScrollAnimationTimer: number | undefined = undefined
 
+function updateSelectedComponentsFromEditorPosition(
+  derived: DerivedState,
+  editor: EditorState,
+  dispatch: EditorDispatch,
+  filePath: string,
+  line: number,
+): EditorState {
+  if (Object.keys(editor.jsxMetadata).length === 0) {
+    // Looks like the canvas has errored out, so leave it alone for now.
+    return editor
+  } else {
+    const allElementPaths = derived.navigatorTargets
+    const highlightBoundsForUids = getHighlightBoundsForFile(editor, filePath)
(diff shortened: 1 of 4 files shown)


maltenuhn:
I've just tried this with the "link to the test editor".

Annoyingly whether or not I get a selection on the canvas for me is still more or less random (see video) - however, the _code editor_ annotation (blue bar on the left) now seems to get this right, so hopefully the canvas issue is orthogonal?

https://user-images.githubusercontent.com/2945037/120016594-b2e51500-bfdc-11eb-9761-6c7ad71ee83c.mp4

balazsbajorics (approved):

enidemi (approved):

PR #1323, opened 2021-06-01
FollowSelectionConfigKey should be Workspace level instead of Global

### LINKED ISSUE #1320: Code editor doesn't follow if the file isn't already open
**Problem**
recent regression: selecting things on the canvas (vanilla project) doesn't open those files anymore (but works if they're already open).

balazsbajorics:
Fixes #1320

**Problem:**
Code editor selection would not follow Canvas. Turns out that it was a user / UX error: we didn't realize that Malte's "Follow selection" icon was toggled off. And we didn't realize that it was a Global setting in VSCode, meaning creating a new project would not fix it.

**Fix:**
I personally wanted to make this a Session-level setting, but as it stands, `utopia.editor.followSelection.enabled` is the only configuration which lives in the VSCode configuration management, so it is kind of an example code. Rheese developed synchronization / retreaval logic for it etc, the system is waiting for more config keys to be added. I don't want to erase this functionality, so I decided to keep it inside VSCode config, but change it from Global to Workspace-level. That means that if you create a new project it will start from enabled, but if you change it, we will save the setting for a project.

**Commit Details:**
- Turned `utopia.editor.followSelection.enabled` from Global to Workspace
- Added tooltip to the follow selection button, copy: `Mirror selection between canvas and code editor`
- Added a forwardRef to `IconToggleButton` to fix Tippy

### DIFF
--- a/editor/src/components/editor/top-menu.tsx
+++ b/editor/src/components/editor/top-menu.tsx
@@ -61,12 +61,14 @@ export const TopMenu = betterReactMemo('TopMenu', () => {
           <MenuIcons.Navigator />
         </SquareButton>
       </Tooltip>
-      <IconToggleButton
-        onToggle={onToggleFollow}
-        value={followSelection.enabled}
-        srcOn={UNSAFE_getIconURL('bracketed-pointer', 'blue', 'semantic', 18, 18)}
-        srcOff={UNSAFE_getIconURL('bracketed-pointer', 'darkgray', 'semantic', 18, 18)}
-      />
+      <Tooltip title='Mirror selection between canvas and code editor' placement='bottom'>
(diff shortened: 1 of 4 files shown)


balazsbajorics:
**Note:**
If you have this setting Globally turned off, now is the time to turn it on, because after this is merged, the code will only be able to flip it on and off per-project, and if you create a new project, it will inherit the global state. This is not an issue for new users or anyone who deletes their cookies.

maltenuhn (approved):

seanparsons:
Performance tests are borked.

balazsbajorics:
> Performance tests are borked.

image

WTF!! hahaha

Issue #1309, opened 2021-05-27
Deleting contents of `card.js` in the test project doesn't give canvas error


maltenuhn:
STR
1. In the test project, delete the contents of card.js (don't save)
2. Behold the canvas

Expected: error
Observed: no error

enidemi:
I see this error when deleting the contents of `card.js`. Is it possible that it was fixed as part of a different bugfix?

<img width="1440" alt="Screenshot 2021-05-28 at 13 54 07" src="https://user-images.githubusercontent.com/4403069/119980193-3d694c80-bfbc-11eb-91ad-a45e981f13cd.png">

maltenuhn:
Yes! Alas, now the error refers to `UtopiaSpiedFunctionComponent` and says
it's coming from Utopia itself...

enidemi:
I actually see it frequently, but not always, that deleting the `card` content then undoing changes doesn't update the canvas and keeps this error message here.

maltenuhn:
Yes; I wonder if it's also related to the issue @seanparsons you're tackling, but I'm going to try and create some additional error cases around 'undo' (in code editor) and commenting things out; I suspect there's still a bug or two hiding there.

PR #1324, opened 2021-06-01
Fix/canvas error missing element

enidemi:
Fixes https://github.com/concrete-utopia/utopia/issues/1309

**Problem:**
Deleting filecontent shows a generic error on the canvas without information where it's coming from, when it's actually an error in the user code.

**Fix:**
Creating a canvas runtime error with stackframes shows the original code.

**Commit Details:**
- throwing a custom error in the ui-jsx-canvas
- making stackframes using the highlightbounds and the source code

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/canvas-render-errors.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/canvas-render-errors.tsx
@@ -1,0 +1,55 @@
+import { FancyError } from '../../../core/shared/code-exec-utils'
+import { JSXElement } from '../../../core/shared/element-template'
+import { HighlightBoundsForUids } from '../../../core/shared/project-file-types'
+import { getUtopiaIDFromJSXElement } from '../../../core/shared/uid-utils'
+import StackFrame, { ScriptLine } from '../../../third-party/react-error-overlay/utils/stack-frame'
+
+export function canvasMissingJSXElementError(
+  factoryFunctionName: string | null,
+  sourceCode: string,
+  jsxElement: JSXElement,
+  filePath: string,
+  highlightBounds: HighlightBoundsForUids | null,
+): FancyError {
+  const error = new Error(`${jsxElement.name.baseVariable} is not defined.`)
+
+  const uid = getUtopiaIDFromJSXElement(jsxElement)
+  const highlightBoundsForElement = highlightBounds != null ? highlightBounds[uid] : null
+  const lineNumber = (highlightBoundsForElement?.startLine ?? 1) + 1
+  ;(error as FancyError).stackFrames = createStackFramesFromPosition(
+    factoryFunctionName,
+    sourceCode,
+    filePath,
+    lineNumber,
+    highlightBoundsForElement?.startCol ?? 1,
+  )
+  return error
+}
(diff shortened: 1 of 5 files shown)


editor/src/components/canvas/ui-jsx-canvas-errors.spec.tsx
```diff
@@ -684,11 +684,117 @@ export var ${BakedInStoryboardVariableName} = (props) => {
...
+              "lineNumber": 3,
+            },
+            ScriptLine {
+              "content": "      const MyCard = undefined",
+              "highlight": false,
+              "lineNumber": 4,
+            },
+            ScriptLine {
+              "content": "      export var App = props => <MyCard data-uid={'bbb'} />",
+              "highlight": false,
```
balazsbajorics, line 714: I don't remember if we need this to be `highlight: true` or not – it might be that the error overlay is not reading this and using the line number from above, but I can also imagine that this is a flag that needs to be set in order to highlight this line

editor/src/components/canvas/ui-jsx-canvas-renderer/canvas-render-errors.tsx
```diff
@@ -0,0 +1,57 @@
...
+  filePath: string,
+  highlightBounds: HighlightBoundsForUids | null,
+): FancyError {
+  const error = new Error(
+    `(${jsxElement.name.baseVariable}) Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.`,
+  )
+
+  const uid = getUtopiaIDFromJSXElement(jsxElement)
+  const highlightBoundsForElement = highlightBounds != null ? highlightBounds[uid] : null
+  const lineNumber = (highlightBoundsForElement?.startLine ?? 1) + 1
```
balazsbajorics, line 20: that 1+1 looks suspicious to me :) 

```suggestion
  const lineNumber = (highlightBoundsForElement?.startLine ?? 0) + 1
```

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx
```diff
@@ -143,3 +157,36 @@ export function createExecutionScope(
     requireResult: requireResult,
   }
 }
+
+export function useGetCodeAndHighlightBounds(
+  filePath: string | null,
+): { code: string; highlightBounds: HighlightBoundsForUids | null } {
+  return useContextSelector(UtopiaProjectContext, (c) => {
```
balazsbajorics, line 164: beware, `useContextSelector`'s default equality function is `Object.is` (strict equality) and not shallowEq. you can change that by providing a shallow eq function as the third parameter

editor/src/components/canvas/ui-jsx-canvas-errors.spec.tsx
enidemi: Screenshot 2021-06-02 at 10 07 03

balazsbajorics: looks great :)

seanparsons (approved):

PR #1325, opened 2021-06-01
Update Component button

OmarDaSilva:
Updated the component button. Removed the placeholder for the name/label of whatever is selected. A component icon is shown for things that are components otherwise a ghost icon will show. Button toggles focus mode also. 

Screenshot 2021-06-01 at 16 32 58

Screenshot 2021-06-01 at 16 32 13

Screenshot 2021-06-01 at 16 31 57

### DIFF
--- a/editor/src/components/editor/component-button.tsx
+++ b/editor/src/components/editor/component-button.tsx
@@ -92,7 +92,7 @@ export const ComponentOrInstanceIndicator = betterReactMemo('ComponentOrInstance
         display: 'flex',
         alignItems: 'stretch',
         height: UtopiaTheme.layout.inputHeight.default,
-        flexBasis: 130,
+        flexBasis: 38,
       }}
     >
       <FlexRow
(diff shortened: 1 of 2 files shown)


OmarDaSilva:
Can I have some feedback?

editor/src/components/editor/component-button.tsx
```diff
@@ -39,39 +46,46 @@ export const ComponentOrInstanceIndicator = betterReactMemo('ComponentOrInstance
...
+  const target = selectedViews[0]
 
-  const [currentEditContext, setCurrentEditContext] = React.useState(editContext.COMPONENT)
+  const isFocused = React.useCallback(() => {
+    if (target != null) {
+      return EP.isFocused(focusedElementPath, target)
+    } else {
+      return false
+    }
+  }, [focusedElementPath, target])
```
balazsbajorics, line 57: Based on how it is used right now, it seems like `ifFocused` does not need to be a callback function, you might as well just save the value into a const and use that:

```suggestion
  const isFocused = target == null ? false : EP.isFocused(focusedElementPath, target)
```

beware that you also need to replace the usage from `isFocused()` to `isFocused` everywhere in the file!

seanparsons (approved):

editor/src/components/editor/component-button.tsx
```diff
@@ -39,39 +46,46 @@ export const ComponentOrInstanceIndicator = betterReactMemo('ComponentOrInstance
...
-  // TODO replace this with a function that enters focus mode
-  const TODOREPLACEMEcycleEditContext = React.useCallback(() => {
-    const nextEditContext = (currentEditContext + 1) % 4
-    setCurrentEditContext(nextEditContext)
-  }, [currentEditContext])
+  const toggleFocusMode = React.useCallback(() => {
+    dispatch([setFocusedElement(isFocused() ? null : target)])
+  }, [dispatch, isFocused, target])
+
+  const isComponent = React.useCallback(() => {
```
balazsbajorics, line 57: same here, it _seems like_ there's no reason why this isn't just a `const isComponent = MetadataUtils.isFocusableComponent(target, metadata)`

enidemi (approved):

balazsbajorics (approved):

PR #1327, opened 2021-06-02
Fix/monkey providers

### LINKED ISSUE #1319: No navigator below App, or editable App component, in this sample project
**Problem**
This project uses a ThemeProvider component (HOC, presumably?) and the useTheme hook. It also doesn't provide a usable navigator, and clicking the 'Edit' entry in the inspector (or double-clicking into `App`) fails, with no obvious error in the console.

https://utopia.pizza/p/2a7b02a0-prototype-theme-%2F-color-variables/

balazsbajorics:
Fixes #1319

**Problem:**
If you are using a ThemeProvider, the spying would stop working and the navigator / selection / etc would give up:
image
Sad!

**Fix:**
Turns out the problem was in the Monkey Function. 
ThemeProvider is an exotic component, but that shouldn't prevent us from printing its UID to the DOM. And as it turns out, nothing technically prevents us from doing so, except some weird (old) special cased handling that was deliberately doing this!
I don't quite know _why_ we were omitting Fragment uids in the first place, but I believe the explanation only makes sense for a single-uid world, and ever since the Monkey started accumulating UIDs, it was doing fragments and exotic components a misservice by deliberately skipping their UIDs from the accumulation.

The fix is to simply delete the 1 line that was snipping out the exotic uid.
image

**Commit Details:**
- Extended the tests to also cover data-paths not just data-uid
- Identified a test case called `class components have a working context, third variant` that demonstrates my problem
- Realized that the culprit was `uidsWithoutExoticUID`
- Delete `uidsWithoutExoticUID` 
- Profit

### DIFF
--- a/editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
+++ b/editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
@@ -29361,9 +29361,9 @@ exports[`UiJsxCanvas render renders correctly with a context 1`] = `
     data-uid=\\"ddd ccc\\"
   >
     <div
-      data-uid=\\"bbb app-entity\\"
+      data-uid=\\"bbb aaa app-entity\\"
       style=\\"width: 100%; height: 100%; background-color: #eb1010;\\"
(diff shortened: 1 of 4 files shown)


maltenuhn (approved):
I'll happily approve anything related to monkey providers

enidemi (approved):

PR #1333, opened 2021-06-02
Theme / Inspector / Controls update

maltenuhn:
**Problem:**
The inspector styles had a lot of outlines and borders, which was fine when the inspector had fewer things to show, but now looks cluttered. I had designs and a prototype (https://codesandbox.io/s/inspector-and-font-testing-38m6m) updating this, but implementing it was surprisingly not straightforward because:

* `control-styles` were mostly using hardcoded values that also weren't always consistent with the theme
* The theme itself file had grown pretty unwieldy and as a result wasn't used in a number of places
* there was a lot of indirection (my doing!) that made it hard to understand "what was what colour when, why"

**Fix:**
* Substantially simplified the theme, including removing entries used not-at-all or infrequently
* Substantially simplified the control-styles, including having them reference the theme
* Removed indirection
* Updated theme values to bring it 80% in line with the prototype

A few caveats:
1. this doesn't implement the "if something is unset, show the whole row in a different colour unless something inside is hovered or focused" (which had been an issue / on various versions of the roadmap for at least 18 months) as at the moment our rows and property labels aren't aware of the `control-status`. 
2. This also doesn't apply the theme to the _icon_ colours (a prerequisite for eg a dark theme). Not hard to do, but: orthogonal.
3. the control-styles for more esoteric cases may not be the last word, partly because some I genuinely didn't know how to produce / reproduce. But... they're also infrequent :) 

After:
image


editor/src/uuiui/inputs/number-input.tsx
```diff
@@ -633,39 +649,30 @@ export const NumberInput = betterReactMemo<NumberInputProps>(
...
                   height: '100%',
                   color: controlStyles.secondaryColor,
                 }}
               >
-                {typeof labelInner === 'object' && 'type' in labelInner ? (
-                  <Icn {...labelInner} />
-                ) : (
-                  labelInner
-                )}
+                {labelInner}
```
enidemi, line 675: I believe this could throw errors when a label is containing iconprops

editor/src/components/inspector/controls/option-chain-control.tsx
```diff
@@ -43,8 +43,8 @@ export const OptionChainControl: React.StatelessComponent<DEPRECATEDControlProps
     <div
       id={props.id}
       key={props.key}
-      style={{
-        width: '100%',
+      css={{
+        // width: '100%',
```
enidemi, line 47: this can be deleted

enidemi (approved):

editor/src/components/inspector/common/control-status.ts
```diff
@@ -39,53 +33,6 @@ export interface ControlStyles {
...
-  DetectedFromCssSecondaryColor: colorTheme.inspectorDetectedFromCssMainColor.value,
-  OffMainColor: colorTheme.inspectorOffMainColor.value,
-  OffSecondaryColor: colorTheme.inspectorOffSecondaryColor.value,
-  OffBackgroundColor: colorTheme.inspectorOffBackgroundColor.value,
-  OffSegmentSelectorColor: colorTheme.inspectorOffSegmentSelectorColor.value,
-  OffSegmentTrackColor: colorTheme.inspectorOffSegmentTrackColor.value,
-  OffBorderColor: colorTheme.inspectorOffBorderColor.value,
-}
-
 export type ControlStatus =
```
balazsbajorics, line 36: two observations:

1. this function tries to reduce repetition by only overwriting a few values, but this actually makes it harder to parse, vs every case just returning a full style object – we have the theme anyways, if you want to mass update a color
2. this function was made years ago, we later introduced the `PropertyStatus` type which is an object with fields like controlled, set, overwritable, detected, fromCssStyleSheet, trivialDefault etc – there's no need to condense these to harder-to-understand statuses like `multiselect-identical-unset`. We could write a styling function that works much more directly (ie `if trivialDefault, set to light grey`)

enidemi: I was thinking similar while reading this, it would be easier to have full style objects where you can see all properties in one place

maltenuhn: > "every function returning a full style object"  / "see all properties in one place"

Admittedly the previous code _was_ closer to this, at least in step 1. One of the downsides was that - more so now than before - most control statuses only change a few values, many just two or three, and returning everything made it hard to see what actually is different.

The other downside - and I did try returning style _objects_ experimentally - is that you basically need to reconstruct your components inside the returned object: the controls often apply the colours in specific places, and almost never do those live on the outer, containing, element. Even on the humble "input"!

editor/src/uuiui/inputs/number-input.tsx
maltenuhn: Oh my. Does anyone recall the no doubt perfectly legitimate reason _why_ labelInner could be a props object?

enidemi: It is used now for different color pickers as an object prop, I think it can be safely changed to icon.

maltenuhn: Fixed by restoring the old code for now.

editor/src/uuiui/inputs/base-input.tsx
```diff
@@ -19,12 +19,22 @@ export type BoxCorners =
...
+  hovered: boolean,
+) {
   const controlStatusEdges = getChainSegmentEdge(controlStyles)
+  const hoveredBoxShadow = `0 0 0 1px hsl(0,0%,83%) inset`
   const focusedBoxShadow = `0 0 0 1px ${UtopiaTheme.color.inspectorFocusedColor.value} inset`
 
   const standardBoxShadow = {
-    boxShadow: focused ? focusedBoxShadow : `0 0 0 1px ${controlStyles.borderColor} inset`,
+    boxShadow: focused
+      ? focusedBoxShadow
```
seanparsons, line 34: Nested ternaries everywhere!

seanparsons (approved):

Issue #1315, opened 2021-05-28
Deleting `card.js` from the filebrowser doesn't update the canvas.


enidemi:
After deleting the file the canvas still renders the `Card` component.
Inserting new components/divs from the insertmenu to the parent `App` still keeps the Card on the canvas. The insertmenu correctly doesn't show the Card, because it was deleted.

How it should work: there is an error on the canvas when using a non-existing component.

After reloading the correct error shows up.

<img width="1094" alt="Screenshot 2021-05-28 at 14 19 49" src="https://user-images.githubusercontent.com/4403069/119983145-10b73400-bfc0-11eb-8365-5a5af9a598d5.png">

maltenuhn:
I've just reproduced this on pizza. (Both removing the `import { Card } from `/src/card.js` and removing the `export const Card = ... ` now are working correctly, but _deleting_ card.js via the file browser still causes this problem)

enidemi:
Fixed by https://github.com/concrete-utopia/utopia/pull/1335

Issue #1330, opened 2021-06-02
changing the name of an import file doesn't update the canvas if there's an error


maltenuhn:
**STR**
1. create a new file
2. export out a component
3. import it into app, but type the wrong filename in the import statement
4. (canvas should error)
5. fix it, save the file

Expected: it's fixed
Observed: error message stays, and refers to the previous file

<img width="1402" alt="Screenshot 2021-06-02 at 20 54 04" src="https://user-images.githubusercontent.com/2945037/120544011-dba64980-c3e4-11eb-9a2e-cf9a1aa81f50.png">

seanparsons:
I have been completely unable to reproduce this, even to the extent of finding a different but related issue: #1334

seanparsons:
@maltenuhn Are you still able to reproduce this? I'm wondering if some other issue happened prior to this which is the actual cause.

maltenuhn:
@seanparsons I just tried it, and now am also unable to reproduce it. I suspect / hope that some of the many fixes around linting / error handling / better error messages fixed it!

seanparsons:
Sounds good to me!

PR #1335, opened 2021-06-02
Cleanup obsolete entries in nodeModules for project files.

### LINKED ISSUE #1334: Using a previously existing filename of a project file in an import surprisingly works.
**Describe the bug**
If the project has a file `a.js` which is renamed `b.js`, it's still possible to import `a.js` from other files and the canvas will render it as it was previously.

**To Reproduce**
- Rename `card.js` to `carp.js`.
- Put the import in `app.js` back to `card.js`.

Result: Canvas renders with frozen in time version of the Card component.

**Expected behaviour**
Error should display indicating that the import failed.

seanparsons:
Fixes #1334

**Problem:**
In the process of trying to reproduce #1330, it was noticed that there's almost a completely inverse issue. If a file is renamed, currently it is possible to import from the old name.

**Fix:**
As a result of falling back from project contents to node modules, as the former are also stored in the latter they were not kept up to date when the project contents had rename or delete type changes which left zombie entries in the node modules which could still be retrieved. The fix was to make `incorporateBuildResult` cull those obsolete entries by comparing them against the project contents.

This also fixes the origin of files being added in there as they were all being incorrectly attributed to `NODE_MODULES` as their origin.

**Commit Details:**
- Fixes #1334.
- `incorporateBuildResult` now includes a fix to ensure that entries
  get tagged with the correct origin.
- `incorporateBuildResult` ensures that entries for project files
  which do not have a matching entry in the project contents are culled.

### DIFF
--- a/editor/src/components/custom-code/code-file.spec.ts
+++ b/editor/src/components/custom-code/code-file.spec.ts
@@ -28,7 +28,14 @@ import {
   getTextFileByPath,
   SampleNodeModules,
 } from './code-file.test-utils'
-import { NodeModules } from '../../core/shared/project-file-types'
+import {
+  NodeModules,
+  RevisionsState,
+  textFile,
+  textFileContents,
+  unparsed,
+} from '../../core/shared/project-file-types'
+import { addFileToProjectContents } from '../assets'
 
 function transpileCode(
   rootFilenames: Array<string>,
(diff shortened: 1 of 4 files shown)


enidemi (approved):

enidemi:
This PR fixes #1315 too.

balazsbajorics (approved):

PR #1336, opened 2021-06-03
Fix: bring back linting

### LINKED ISSUE #1171: Missing Emotion error / Instead get Utopia Error: ScenePath is not provided for Card

balazsbajorics:
Fixes #1171

**Problem:**
This PR tackles two problems:
1. in general we lost Linting when we switched to VSCode. I don't think it was intentional, but none of us was bothered by it enough to bring it back sooner. SAD!
2. In specific, the problem in #1171 is that we are sometimes more permissive on the canvas than javascript Eval would be. To work around this issue, the fastest fix was to enable a lint error that prevents the users from writing illegal JS code. for this I had to resurrect linting!

**Fix:**
Linting is now triggered in UPDATE_FROM_CODE_EDITOR. Lint errors, not just fatal errors now trigger the canvas overlay.

image

**Commit Details:**
- when processing `UPDATE_FROM_CODE_EDITOR`, also trigger a `SEND_LINTER_REQUEST_MESSAGE`
- Include Lint Errors, not just Lint Fatal Errors to turn the canvas overlay on – this might be too agressive behavior
- move the `reduxDevtoolsSendActions` to after the point where the editorState is finalized – this avoids the head scratchy situations where an action seemingly properly updates the editor state, but then the value goes away by the time the next action gets dispatched
- fix `filterEditorForFiles` which broke lint errors and the code cache when we turned `projectContents` into a tree – so about 6 months ago 😅

**Note:**
We have a lot of lint rules which are set to Error severity – some of these should break the canvas, some of these should not.
We have two options here:
1. create a very special Fatal Error category for those lint errors where we want to break the canvas, and otherwise keep Errors and Warnings (red squiggly underline and yellow squiggly underline) in VSCode and the Errors tab
2. review what lint rules are Errors and turn some of them into Warnings, and every Error should break the canvas
This question primarily goes to @maltenuhn 

**PS**
It seems like I accidentally fixed the exportsInfo and now the File navigator shows the exported components again 😅– I thought we deliberately killed that feature, I was not aware it's a regression. We do seem to reliably regress everything that doesn't have explicit tests for it.

### DIFF
--- a/editor/src/components/canvas/canvas-wrapper-component.tsx
+++ b/editor/src/components/canvas/canvas-wrapper-component.tsx
@@ -45,7 +45,7 @@ export const CanvasWrapperComponent = betterReactMemo(
     )
 
     const fatalErrors = React.useMemo(() => {
-      return getAllCodeEditorErrors(editorState, true, true)
+      return getAllCodeEditorErrors(editorState, 'fatal', true)
     }, [editorState])
 
     const safeMode = useEditorState((store) => {
@@ -91,7 +91,7 @@ const ErrorOverlayComponent = betterReactMemo(
       )
     }, 'ErrorOverlayComponent utopiaParserErrors')
     const fatalCodeEditorErrors = useEditorState((store) => {
-      return getAllCodeEditorErrors(store.editor, true, true)
+      return getAllCodeEditorErrors(store.editor, 'error', true)
     }, 'ErrorOverlayComponent fatalCodeEditorErrors')
 
     const runtimeErrors = useReadOnlyRuntimeErrors()
--- a/editor/src/components/editor/actions/action-utils.ts
+++ b/editor/src/components/editor/actions/action-utils.ts
@@ -191,6 +191,7 @@ export function isFromVSCode(action: EditorAction): boolean {
     case 'ATOMIC':
       return action.actions.some(isFromVSCode)
     case 'UPDATE_FROM_CODE_EDITOR':
+    case 'SEND_LINTER_REQUEST_MESSAGE':
       return true
     default:
       return false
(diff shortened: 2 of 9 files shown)


balazsbajorics:
for comparison, this is the runtime error you can see in codesandbox:
image

I think the lint error is actually more helpful in this case, so yay for that!

enidemi (approved):

balazsbajorics:
I'm gona merge this when the tests pass, we can figure out the error /fatal error question later

maltenuhn:
> review what lint rules are Errors and turn some of them into Warnings

I guess we're using the default eslint list for these?

maltenuhn:
> I accidentally fixed the exportsInfo and now the File navigator

I believe we had a ticket to remove them, and closed that when they _were_ removed. 🤦

balazsbajorics:
> > review what lint rules are Errors and turn some of them into Warnings
> 
> I guess we're using the default eslint list for these?

Yes, we mostly are using the default recommended rules.

I just checked with a vanilla `create-react-app` and Error level lint errors do show the error overlay:
image

So I think what I just merged to master is inline with that.

maltenuhn:
> Yes, we mostly are using the default recommended rules.

Ok, in that case I think we just merge it. If there's things where we want to deviate, we can spot & decide case-by-case

Issue #1340, opened 2021-06-03
Support fragments in the navigator / on the canvas in the inspector


maltenuhn:
**STR**
1. Open  https://utopia.pizza/p/37fba76f-todo-app/ 
2. Observe how the first child of App is sometimes showing up in the navigator, and nothing is selectable on the canvas
3. Replace the fragment `<>` in `App`  with a div

Now the navigator populates. But because the original creator of this example chose a coding style where almost _everything_ is a fragment, you still can't select things via the canvas, _and_ the inspector is not letting you modify them if you select them via the navigator.

<img width="1436" alt="Screenshot 2021-06-03 at 21 04 15" src="https://user-images.githubusercontent.com/2945037/120705426-b7fc0580-c4af-11eb-9461-8c78812f3fca.png">

_Originally posted by @maltenuhn in https://github.com/concrete-utopia/utopia/issues/1332#issuecomment-854144981_

enidemi:
We supported fragments in the past, I think it is not covered by tests well enough.

PR #1343, opened 2021-06-07
fix(fragments) apply uid fix on children

enidemi:
**Problem:**
Editing jsx fragments in the code editor throws error. Part of #1340

**Fix:**
The error is coming from uid-fix, it only works on jsx elements and fails on fragments at a point in `transformAtPathOptionally`. The fix is that when the uid fix finds a jsx fragments it skips itself and only applies on child elements, using the original path without appending the special uuid.

**Commit Details:**
- added helper `isJSXElementLikeWithChildren`
- removed duplicated call of `onElement` as the `walkElementChildren` in the next line calls it on the same rootElement too
- extended `walkElementChildren` to handle fragments too
- test for uid fix on fragments

### DIFF
--- a/editor/src/core/workers/parser-printer/uid-fix.ts
+++ b/editor/src/core/workers/parser-printer/uid-fix.ts
@@ -65,7 +65,7 @@ export function fixParseSuccessUIDs(
 
   const newToOldUidMappingArray = Object.values(newToOldUidMapping)
 
-  if (newToOldUidMappingArray.length > 0) {
+  if (newToOldUidMappingArray.length === 1) {
     // we found a single UID mismatch, which means there's a very good chance that it was an update element, let's fix that up
     let workingComponents = getComponentsFromTopLevelElements(newParsed.topLevelElements)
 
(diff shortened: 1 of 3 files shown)


seanparsons (approved):

balazsbajorics (approved):

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -143,10 +144,7 @@ function zipTopLevelElements(
   firstComponents.forEach((firstComponent, index) => {
     if (secondComponents.length > index) {
       const secondComponent = secondComponents[index]
-      const firstUID = getUtopiaID(firstComponent.rootElement)
-      const secondUID = getUtopiaID(secondComponent.rootElement)
 
-      onElement(firstUID, secondUID, EP.emptyElementPathPart, EP.emptyElementPathPart)
```
Rheeseyb, line 149: Is this correct? We're never calling `onElement` on top level elements now?

enidemi: yes, this was using the root element, as the first call of `walkElementChildren` in the next line too, inside the function it's calling `onElement` on the root element and for fragments it skips to its children

Rheeseyb (approved):

PR #1352, opened 2021-06-10
Focus canvas on mouse down.

### LINKED ISSUE #1258: Clicking a drag handle to resize requires a second mouse down before it actually works
When attempting to drag a resize drag handle, the first mouse down will result in the cursor staying as the resize cursor, but won't actually enable resizing. You then have to mouse down again before dragging will have any effect on the element.

seanparsons:
Fixes #1258.

**Problem:**
Should selection be triggered from somewhere other than the canvas that would cause certain events to be either ignored or potentially mishandled because the canvas is not the "focused panel",

**Fix:**
An early exit condition was removed so that the second fix which focuses the canvas panel on a mouse down.

**Commit Details:**
- Fixes #1258.
- Removes an early exit conditional in `handleEvent` which was
  preventing some calls from reaching the canvas when for example
  selection had been triggered from the navigator. As that would
  cause the resize controls to be available.
- A mouse down on the canvas when it is not the focused panel causes
  it to be the focused panel.

### DIFF
--- a/editor/src/templates/editor-canvas.tsx
+++ b/editor/src/templates/editor-canvas.tsx
@@ -565,10 +565,6 @@ export class EditorCanvas extends React.Component<EditorCanvasProps> {
       this.props.dispatch([CanvasActions.setSelectionControlsVisibility(true)], 'canvas')
     }
 
-    if (this.props.model.focusedPanel !== 'canvas' && event.event !== 'WHEEL') {
-      return
-    }
-
     if (
       (event.nativeEvent != null &&
         event.nativeEvent.srcElement != null &&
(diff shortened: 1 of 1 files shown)


enidemi (approved):

editor/src/templates/editor-canvas.tsx
```diff
@@ -565,10 +565,6 @@ export class EditorCanvas extends React.Component<EditorCanvasProps> {
       this.props.dispatch([CanvasActions.setSelectionControlsVisibility(true)], 'canvas')
     }
 
-    if (this.props.model.focusedPanel !== 'canvas' && event.event !== 'WHEEL') {
-      return
-    }
```
Rheeseyb, line 570: Does removing this mean that scrolling the navigator will now also scroll the canvas?

seanparsons: I shall test that case.

seanparsons: I can confirm that navigator and canvas scroll independently.

Rheeseyb: Thanks for checking!

Rheeseyb (approved):

PR #1354, opened 2021-06-10
Default to Selection Only Mode

### LINKED ISSUE #1353: Default to selection only mode
Introduce a mode that supports selection only, and make that the default for the sake of first impressions.

Rheeseyb:
Fixes #1353

**Problem:**
We want to introduce people to Utopia, but with all interactions enabled the editor feels broken.

**Fix:**
Introduce a "Select Only" mode and make that the default. From the user's perspective, the modes available are "edit", "select" and "live", but internally we refer to the new mode as "Select Lite" (purely because it became confusing trying to determine where to draw the lines when renaming existing files and functions).

This also introduces the controls to switch between the 3 options, which float above the canvas in the top right like so:
Screenshot_20210610_181851

(I managed to find a superscript beta to replace the alpha from the prototype :grinning:)

**Commit Details:**
- Added the new mode to `editor-modes.ts`
- Restrict keyboard interactions in `global-shortcuts.ts`
- Restrict dragging interactions in `select-mode-hooks.tsx` and `select-mode-control-container.tsx`
- Prevent the use of the context menu and rendering of resize controls in the new mode in `select-mode-control-container.tsx`
- Made the new mode the default, combating that in the tests (and resizing performance script) by firing an action to switch the mode first
- Added the new mode select buttons to the canvas in `canvas-wrapper-component.tsx`


enidemi (approved):

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -73,11 +81,117 @@ export const CanvasWrapperComponent = betterReactMemo(
...
+  onMouseDown: () => void
+}
+
+const ModeSelectButton = betterReactMemo('ModeSelectButton', (props: ModeSelectButtonProps) => {
+  return props.selected ? (
+    <div
+      // this is the selected variant. No hover effects on this one
+      style={{
+        padding: '2px 4px',
+        background: colorTheme.emphasizedBackground.value,
```
maltenuhn, line 102: I think this wants `secondary`, since the background of the parent is `emphasized` - right now you don't see the selected orb. Same with the hover effects in the one below.

(default editor bg is 98% lightness, `colorTheme.emphasized` is 100% (i.e white), and `colorTheme.secondary` etc are darker at 96% lightness etc. Which, of course, is also a form of emphasis :facepalm:. )

Rheeseyb: Fixed!

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -73,11 +81,117 @@ export const CanvasWrapperComponent = betterReactMemo(
...
   },
 )
 
+interface ModeSelectButtonProps {
+  selected: boolean
+  title: string
+  onMouseDown: () => void
+}
+
+const ModeSelectButton = betterReactMemo('ModeSelectButton', (props: ModeSelectButtonProps) => {
```
balazsbajorics, line 96: could you put this in a new file please? the canvas-wrapper-component already does a lot of the mission critical error boundary and logging related tasks, I'd hate to dilute it with new UI code

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -537,23 +539,24 @@ export function useSelectAndHover(
 } {
   const modeType = useEditorState((store) => store.editor.mode.type, 'useSelectAndHover mode')
   const selectModeCallbacks = useSelectModeSelectAndHover(
```
balazsbajorics, line 541: hmmm if this now covers select, select-lite and live as well, I think we shouldn't be calling this `useSelectModeSelectAndHover` anymore

```suggestion
  const selectAndHoverCallbacks = useSelectAndHover(
```

editor/src/components/canvas/canvas-wrapper-component.tsx
Rheeseyb: Done

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
Rheeseyb: Yep but that name is already taken lower down in the file! Will change it though

Rheeseyb: I've renamed to `useSelectOrLiveModeSelectAndHover` (it was already covering live mode, and select-lite is planned to be binned in the near-ish term)

balazsbajorics: thanks :)

balazsbajorics (approved):

PR #1355, opened 2021-06-10
update website

balazsbajorics:
Sooooo. My initial goal was to set up some sort of routing that would redirect us to Vercel.com if the user wants to open the landing page, but that task turned out to be less than trivial.

After much consulting with Sean, and a bit of trial and error, I decided to just take the plunge, and move every route currently handled by `website` to a new `website-next` project, which is a Next.js app hosting the landing page, projects page and other misc. things. 

The main upside is the following:
create-react-app generated a static html for us that was just an empty white page, and it contained a js bundle that would load and draw the page. This was probably fine for the projects page and preview page, but pretty ugly for the main landing page.
Next.js can generate a static html for every route we have, that will contain either the full content (for the landing page) or a placeholder/empty content (for the projects page, this is not as important). I think I am beyond 80% completing this task, the only remaining main thing is that I need to change the haskell server so if you load something like `utopia.app/projects` it tries to serve `public/projects.html` instead of always resolving to always serving `public/index.html`. Sean told me it should be fine and there's no sharks waiting for me.

TODO:
- [x] Bump create-react-app
- [x] Realize create-react-app can never generate contentful static html, also that it is not too much effort to just kill it and replace it with Next.js
- [x] bring the utopia-landing-page code in to `website-next`, this is the home of our new next.js project
- [x] move all routes over from `website`. as far as I can tell, these all work in next.js too
- [x] verify I didn't accidentally bloat the initial js bundle for the landing page
- [x] make it so that the Haskell server can load projects.html for routes such as `utopia.app/projects` (I think this will be a fallback-type behavior, I'll nag @seanparsons  about how to do it
- [x] for some reason `localhost:8000/login` stopped working, I need to figure this out
- [x] `website` would refuse to build without a UTOPIA_SHA, I probably need to replicate this in Next
- [x] delete the old website project
- [x] change the editor `src/common` symlink to point to `website-next/components/common` 
- [ ] profit
- [x] CDN!?!


website-next/pages/preview/[id].tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react'
+import { useRouter } from 'next/router'
+import { PreviewPage } from '../../components/preview/preview-page-component'
+
+export default function Preview() {
+  const router = useRouter()
+  const { id } = router.query
```
Comment (balazsbajorics) [inline website-next/pages/preview/[id].tsx:7]: routing in Next.js is expressed via the folder/file structure. `pages/preview/[id].tsx` is the same as `/preview/[id]` was in React Router. To get the id parameter, I use `useRouter()` from next. The same `router.query` objects would also contain the query parameters passed to the site.

website-next/pages/login.tsx
```diff
@@ -0,0 +1,11 @@
+import { useRouter } from 'next/router'
+import { auth0Url } from '../components/common/env-vars'
+
+export default function Login() {
```
balazsbajorics, line 4: I put this here because we had the same routing rule in the `website` but I think this can be deleted as it no longer seems to work

website-next/pages/projects.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react'
+import { ProjectsPage } from '../components/projects/projects-page-component'
+
+export default function Projects() {
+  return <ProjectsPage />
```
balazsbajorics, line 5: to keep things readable, I copied the old projects page into `components/` and only reference it here. I think this is a good practice going forward, try to keep the root files clean so we can use them mainly for routing

website-next/pages/_app.js
```diff
@@ -0,0 +1,7 @@
+import '../styles/globals.css'
+
+function MyApp({ Component, pageProps }) {
```
balazsbajorics, line 3: this _app.js is the Next.js entry point. this is where we can intercept what Next is doing and override things if necessary.

This file is vanilla right now, it was generated like this.

balazsbajorics:
there's also a corresponding change to the deploy project here: https://github.com/concrete-utopia/utopia-deploy/pull/20

Rheeseyb:
> move every route currently handled by website to a new website-next project

Since the old website project is being completely deleted, please can you revert the name of the new one to `website`

balazsbajorics:
> > move every route currently handled by website to a new website-next project
> 
> Since the old website project is being completely deleted, please can you revert the name of the new one to `website`

my thinking was that this will make it clear if some script wants to cd into `cd website` and it makes our lives easier fixing those.

plus this PR I _think_ would be much harder to read if the next files, moved files, and the deleted website files would be all intermixed

website-next/components/header.js
```diff
@@ -0,0 +1,45 @@
...
...tic/apple-icon-72x72.png"><link rel="apple-touch-icon" sizes="76x76" href="/static/apple-icon-76x76.png"><link rel="apple-touch-icon" sizes="114x114" href="/static/apple-icon-114x114.png"><link rel="apple-touch-icon" sizes="120x120" href="/static/apple-icon-120x120.png"><link rel="apple-touch-icon" sizes="180x180" href="/static/apple-icon-180x180.png"><link rel="apple-touch-icon" sizes="144x144" href="/static/apple-icon-144x144.png"><link rel="apple-touch-icon" sizes="152x152" href="/static/apple-icon-152x152.png"><link rel="icon" type="image/x-icon" href="/static/apple-icon-152x152.png"><link rel="icon" type="image/png" sizes="16x16" href="/static/favicon-16x16.png"><link rel="icon" type="image/png" sizes="32x32" href="/static/favicon-32x32.png"><link rel="icon" type="image/png" sizes="96x96" href="/static/favicon-96x96.png"><link rel="icon" type="image/png" sizes="192x192" href="/static/android-icon-192x192.png"><meta name="msapplication-TileColor" content="#ffffff"><meta name="msapplication-TileImage" content="/static/ms-icon-144x144.png"><link data-utopia="" rel="stylesheet" href="/editor/css/utopions.css"><title>Utopia</title><style data-emotion="css" data-s=""></style></head>
```
Rheeseyb, line 45: ?

Can be deleted?

balazsbajorics: done

website-next/components/projects/cards.tsx
```diff
@@ -1,6 +1,3 @@
-/** @jsx jsx */
-import React from 'react'
-import { jsx } from '@emotion/react'
```
Rheeseyb, line 3: I'm surprised that removing these didn't cause any issues since this file uses jsx syntax

balazsbajorics: oh I deleted these because at one point I had a very agressive linter enabled that would complain for unused vars. I think I'll put the react import back.

_by the way_ the file does not use JSX syntax! it only uses `styled` functions and style objects

website-next/pages/index.js
```diff
@@ -0,0 +1,210 @@
...
+// [ ] - review responsive design, make tweaks
+// [x] - favicon!
+// [x] - Apple icon!
+// [x] - Social Media OG tags @malte
+// [x] - visually center the rotation, either by wrapping it or by transform-origin pixelfucking
+// [x] - white gradient fade-in of bg, and [ ] - background position @malte
+// [x] - spacing between text (margin-top or gap or something)
+// [x] - fuck-off width for the menu (full screen on MB Air)
+// [x] - font weight 500 for base text in Inter
+
```
Rheeseyb, line 32:
```suggestion
```

Though there appear to be some unchecked items here :wink:

balazsbajorics: some people are hard to please

website-next/components/projects/cards.tsx
Rheeseyb: Ha I saw the `<CardProps>` below and mentally read it as a piece of JSX!

website-next/pages/login.tsx
Rheeseyb: Seems like it should go in the bin, but I'm fine for that to happen in a follow up PR

website-next/next.config.js
```diff
@@ -0,0 +1,8 @@
+const cdnUrl = process.env.UTOPIA_CDN_URL
+
+module.exports = {
+  trailingSlash: true,
+  assetPrefix: cdnUrl ?? '', // Use the CDN in production and localhost for development.
+}
+
+console.log('Running Next with config:', JSON.stringify(module.exports))
```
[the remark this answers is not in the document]
Rheeseyb: :+1:

website-next/public/browserconfig.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<browserconfig>
+    <msapplication>
+        <tile>
+            <square150x150logo src="/mstile-150x150.png"/>
+            <TileColor>#da532c</TileColor>
+        </tile>
+    </msapplication>
+</browserconfig>
```
Rheeseyb, line 9: Oh wow, we're supporting IE now?! :sweat_smile:

balazsbajorics: this is Windows 10 start menu support!!! malte's favorite OS

website-next/tsconfig.json
```diff
@@ -4,24 +4,23 @@
     "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
-    "esModuleInterop": true,
-    "allowSyntheticDefaultImports": true,
+    "strict": false,
```
Rheeseyb, line 7: Does this need to be false? I'd like us to keep this as close to the editor's tsconfig as possible

Rheeseyb (approved):

website-next/public/browserconfig.xml
maltenuhn: Did you really just use my deviant preferences as an excuse for your procrastitask? That's low / genius.

website-next/tsconfig.json
balazsbajorics: fixed

enidemi (approved):

website-next/pages/preview/[id].tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react'
```
Comment (seanparsons) [inline website-next/pages/preview/[id].tsx:1]: Filename of this file makes me scrunch my face up.

seanparsons (commented):
+1

PR #1360, opened 2021-06-12
Hide Landing page placeholders

balazsbajorics:
**Problem:**
The landing page on `master` as of now is full of placeholders. In case we need to push something to production, we would unintentionally deploy this half-baked page. 

**Fix:**
Remove the placeholders, leave a barebones landing page in place.


balazsbajorics:
I'm gonna merge this PR now, to make `master` safe to deploy to production, should we need that

PR #1362, opened 2021-06-14
fix(inspector) row mouseover highlight fix

enidemi:
*Problem:**
On some inspector rows the mouseover highlight effect is missing when only the label is hovered, like from the opacity-row.

**Fix:**
The hover color change was part of the UIGridRow, but nesting them makes the hover effect work only on the inner UIGridRow.
I'm fixing it by moving this extra css to the contextmenu wrapper.

**Commit Details:**
- removed an extra contextmenu wrapper from the property-label
- moved `&:hover` and `&:focus-within` from UIGridRow to InspectorContextMenuWrapper


Rheeseyb (approved):

Rheeseyb:
This only failed because of the render count tests

balazsbajorics (approved):
dont' forget the render count tests :)

PR #1367, opened 2021-06-14
Unreversing the navigator items

balazsbajorics:
**Problem:**
Our navigator used ordering similar to Sketch and Figma: elements that are above another are higher in the list. For applications which deal with mostly absolutely positioned layouts, this is not a problem: if you have a layout where things are visually above each other, you have no problem reordering them in the navigator to reflect this relationship. 
However, if you use a layout system (any layout system, really) this becomes a problem: if something is defined "after" something else in the code, it will go after it on screen, but above it in the navigator!

Example:
image

Here "Beaches" is clearly the topmost element on the page, why is it at the bottom of my navigator?

Figma has the same problem when using Autolayout, and lot of confused users are asking for a solution in their support forums.

We struggled with this question for a long time, and our conclusion is that there is probably no good answer, and this should be a user preference. 

However, in the meantime, I am going to unreverse our navigator just for this week, and create a follow-up ticket to create a user preference entry for this.

**Fix:**
This PR unreverses the navigator, so it will display elements in the same order they are declared in code.

image

### DIFF
--- a/editor/src/components/canvas/canvas-utils.spec.browser.tsx
+++ b/editor/src/components/canvas/canvas-utils.spec.browser.tsx
@@ -1268,13 +1268,13 @@ describe('moveTemplate', () => {
       makeTestProjectCodeWithSnippet(
         `<View style={{ ...props.style }} data-uid='aaa'>
           <View data-uid='eee'>
-            <View data-uid='ddd' style={{ left: 52, top: -141 }} />
             <View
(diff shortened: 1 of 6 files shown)


enidemi (approved):

Rheeseyb (approved):
:heart: :tada: :partying_face:

seanparsons (approved):

PR #1374, opened 2021-06-15
feature(website) add font preload, switch font-display

enidemi:
- font preload tags added to the header
- font-display changed to optional
- update tailwind body fontlist

### DIFF
--- a/website-next/components/header.js
+++ b/website-next/components/header.js
@@ -7,6 +7,11 @@ export function Header() {
       <title>Utopia</title>
       <meta name='viewport' content='width=device-width,initial-scale=1,shrink-to-fit=no' />
 
+      {/* preload fonts */}
+      <link rel="preload" href="/fonts/RecklessNeue-Book.woff2" crossorigin="anonymous" as="font" type="font/woff2" />
+      <link rel="preconnect" href="https://fonts.gstatic.com"/>
+      <link href="https://fonts.googleapis.com/css2?family=Inter&display=swap" rel="stylesheet" />
+
       {/* favicons cropped to fix vertical alignment, but I don't like that they are cropped */}
       <link rel='icon' href='/favicon-32.png' />
       <link rel='icon' href='/favicon-128.png' />
--- a/website-next/styles/globals.css
+++ b/website-next/styles/globals.css
@@ -29,7 +29,7 @@ a {
     Fira Sans, Droid Sans, Helvetica Neue, sans-serif;
   font-style: normal;
   font-weight: 400;
-  font-display: fallback;
+  font-display: optional;
 }
 
 @font-face {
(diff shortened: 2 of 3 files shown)


Rheeseyb (approved):

website-next/components/header.js
```diff
@@ -7,6 +7,11 @@ export function Header() {
       <title>Utopia</title>
       <meta name='viewport' content='width=device-width,initial-scale=1,shrink-to-fit=no' />
 
+      {/* preload fonts */}
+      <link rel="preload" href="/fonts/RecklessNeue-Book.woff2" crossorigin="anonymous" as="font" type="font/woff2" />
```
balazsbajorics, line 11: what about the other versions of Reckless?

enidemi: those are not used yet

website-next/styles/globals.css
```diff
@@ -29,7 +29,7 @@ a {
     Fira Sans, Droid Sans, Helvetica Neue, sans-serif;
   font-style: normal;
   font-weight: 400;
-  font-display: fallback;
+  font-display: optional;
```
balazsbajorics, line 32: what is the difference? it used to be fallback which means it showed an ugly font first, then the loaded nice font, leading to content shift.

but with optional, does it show empty space first, then loaded font? does this decrease content shift? or increase it because now the entire Title will just pop in to place?

what did you see in your testing @enidemi ?

website-next/tailwind-preset.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
     fontFamily: {
       headline: ['RecklessNeue-Book', 'serif'],
       menu: ['RecklessNeue-Book', 'serif'],
-      body: ['Inter', 'serif'],
+      body: ['Inter', 'sans-serif'],
```
balazsbajorics, line 6: ooh, this was a leftover from debugging – I wanted to see when the font fails loading, that's why I set it to serif so it's very glaring difference :)

website-next/styles/globals.css
enidemi: From mdn `optional` means "Gives the font face an extremely small block period and no swap period."
There is no font jumping, it waits a little before showing a font, is uses the loaded font if available, otherwise the fallback.
Locally fonts are loaded very fast, so I always saw the loaded font.

website-next/tailwind-preset.js
maltenuhn: ...and let me tell you, it worked :D

website-next/styles/globals.css
balazsbajorics: oooh I see – so if it fails to load the font in time – tough luck! that means we need to pick and configure a pretty similar fallback font then

balazsbajorics (approved):

PR #1377, opened 2021-06-15
make IconToggleButton tippy compatible

balazsbajorics:
**Problem:**
The "mirror selection" button had no working tooltip

**Fix:**
i wrapped IconToggleButton in a forwardRef

image


Rheeseyb (commented):
:+1:

PR #1380, opened 2021-06-15
Basic analytics on the website

Rheeseyb:
This PR introduces the most basic Google Analytics integration, along with a cookie consent bar. The bar only has the option to accept, and displays an overlay over the content, though that doesn't prevent scrolling so it might be best off disabling that option.

Here's the bar in the current styling (I used colours from the blog post and one of our theme colours):
Screenshot_20210615_123119

With this basic integration we just get data for page hits, so the next step is to add events to each of the navigation buttons and for time spent on page.


website-next/components/header.js
```diff
@@ -8,9 +8,34 @@ export function Header() {
...
+        rel='preload'
+        href='/fonts/RecklessNeue-Book.woff2'
+        crossorigin='anonymous'
+        as='font'
+        type='font/woff2'
+      />
+      <link rel='preconnect' href='https://fonts.gstatic.com' />
+      <link href='https://fonts.googleapis.com/css2?family=Inter&display=swap' rel='stylesheet' />
+
+      {/* Google Analytics */}
```
Rheeseyb, line 21: The below is from the sample next.js integration https://github.com/vercel/next.js/blob/canary/examples/with-google-analytics/pages/_document.js

seanparsons (approved):

enidemi (approved):

PR #1383, opened 2021-06-15
Add a simple even for tracking navigation from the menu

Rheeseyb:
This allows us to track which of the buttons from the top menu were clicked. It will need updating / careful merging with #1382 to ensure neither change is lost

### DIFF
--- a/website-next/components/menu.js
+++ b/website-next/components/menu.js
@@ -28,6 +28,9 @@ export function Menu() {
                   key={item.name}
                   href={item.href}
                   className='hidden md:block font-medium text-gray-900 hover:text-gray-900 text-lg'
+                  onMouseDown={() =>
+                    gtag('event', 'navigate', { category: 'links', label: item.href, value: 1 })
+                  }
                 >
                   {item.name}
                 </a>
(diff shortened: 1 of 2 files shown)


website-next/components/menu.js
```diff
@@ -28,6 +28,9 @@ export function Menu() {
                   key={item.name}
                   href={item.href}
                   className='hidden md:block font-medium text-gray-900 hover:text-gray-900 text-lg'
+                  onMouseDown={() =>
+                    gtag('event', 'navigate', { category: 'links', label: item.href, value: 1 })
```
Rheeseyb, line 32: The docs give the impression that a non-negative integer `value` field is required, hence the `1` here. I've checked and we can see this event in the GA dashboard.

balazsbajorics (approved):
and so it begins: the tracking of absolutely useless information

seanparsons (approved):

enidemi (approved):

Rheeseyb:
> and so it begins: the tracking of absolutely useless information

Conversions vs page views is pretty important.

PR #1393, opened 2021-06-17
quick fix: reduce user state polling

balazsbajorics:
**Problem:**
Every open editor session pings the server every 5 seconds to query login state. This puts load on the infra.

**Fix:**
Reduce the polling rate to 30 seconds. This is at best a temporary solution, because if the user enters a "Login lost" state, and then logs back in again, the "You have been logged out of Utopia." sidebar will stick around for up to 30 seconds.

I am open to merging this or not merging this too – let's discuss!


Rheeseyb:
Tagged and closing

PR #1400, opened 2021-06-18
fix(third-party-controls) fix control names

enidemi:
**Problem:**
Some element types controls are missing for antd components from the inspector that already have their controls added to controls collection.

**Commit Details:**
- It seems like the names were not matching.


balazsbajorics (approved):

balazsbajorics:
maybe it's because they changed the names, and our schema was for an older release? it might be worth updating the version number bounds on the definition!

Rheeseyb:
Heads up, this PR still has breaking tests

seanparsons (approved):

PR #1401, opened 2021-06-18
Fix an undefined error in the dom walker

Rheeseyb:
**Problem:**
When we first render the canvas after removing an error overlay, we return `undefined` for `storyboardRootElementPath` as a result of indexing into an empty array. This causes an error much later during DOM walking when trying to use that `ElementPath`. 

**Fix:**
If the array is empty, return `EP.emptyElementPath`.


balazsbajorics (approved):
`>:D`

seanparsons (approved):
Stupid array indexing type.

PR #1402, opened 2021-06-18
Fix some problems with react-three-fiber

enidemi:
**Fix:**
This PR improves navigator support for react-three-fiber components.

**Commit Details:**
- fixing cullSpyCollector by only deleting elements where the scene is not existing anymore, react-three-fiber elements were missing from the metadata as they are not caught by the dom-walker
- minor fix in component-section for Array controls, they were throwing errors for null values
- hiding component section only for elements that are `isIntrinsicHTMLElement`

**Update (from Balazs)**
- fixing the missing propertyControls, those don't rely on the exportsInfo anymore (I _think_ now we can safely delete exportsInfo, but I didn't want to do it for this branch)
- fix tests broken by my fix


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -2424,7 +2424,7 @@ export function cullSpyCollector(
   })
   // Eliminate the element paths which are invalid.
   fastForEach(Object.keys(spyCollector.current.spyValues.metadata), (elementPath) => {
-    if (!elementPaths.has(elementPath)) {
+    if (!elementPaths.has(elementPath.split(':')[0])) {
```
balazsbajorics, line 2427: if we change this, then we also need to purge the metadata in the scene spy function! otherwise deleted elements will stick around

editor/src/templates/property-controls-info.tsx
```diff
@@ -53,14 +53,18 @@ const initPropertyControlsWorker = () => {
         await createBundle(bundlerWorker, emptyTypeDefinitions, projectContents)
       ).buildResult
 
+      // Mutating the evaluation cache.
+      for (const fileToDelete of model.updatedAndReverseDepFilenames) {
+        delete evaluationCache[fileToDelete]
```
balazsbajorics, line 58: this mirrors the same behavior as in code-file.ts, if the user updates the propertyControls, we need to delete the old file from the evaluationCache, this does that

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -264,6 +264,7 @@ export const UiJsxCanvas = betterReactMemo(
     }
 
     let metadataContext: UiJsxCanvasContextData = React.useContext(UiJsxCanvasContext)
+    metadataContext.current.spyValues.metadata = {}
```
balazsbajorics, line 267: @enidemi I'm very confused! 
I thought we are going to clear the metadata _per scene_, but this seems to clear the entire spy values. isn't this causing problems with scenes that are memoized? if we clear everything here, then we don't even need the cullSpyCollector… am I missing something? or is this code leftover from debugging?

enidemi: I haven't checked the memoization yet, it's true this might cause problems there 😬 
I tried to clear the metadata even if scenes are deleted from the model. With the previous solution those would be kept in the model.

balazsbajorics: but you made this here to get rid of scenes which are deleted! something doesn't add up – I suspect we should also have some scene memoization tests too

https://github.com/concrete-utopia/utopia/pull/1402/files#diff-3ee060e2443faa6e8082bec9c9df5819a3a6d070c17446342ef65e737c9f02ebR2427

editor/src/components/canvas/ui-jsx-canvas-renderer/scene-component.tsx
```diff
@@ -5,13 +5,28 @@ import { Scene, SceneProps } from 'utopia-api'
...
+import { UiJsxCanvasContext } from '../ui-jsx-canvas'
+import { fastForEach } from '../../../core/shared/utils'
+
+function useClearMetadataForScene(props: any) {
+  const metadataContext = React.useContext(UiJsxCanvasContext)
+  const scenePath = props['data-utopia-scene-id']
+  if (scenePath != null) {
+    fastForEach(Object.keys(metadataContext.current.spyValues.metadata), (path) => {
+      if (path.includes(scenePath)) {
+        // delete metadataContext.current.spyValues.metadata[path]
```
Rheeseyb, line 17: Does this need uncommenting?

editor/src/core/property-controls/property-controls-utils.ts
```diff
@@ -397,14 +400,19 @@ export function getPropertyControlsForTarget(
...
+          if (isIntrinsicHTMLElement(element.name)) {
+            /**
+             * We detected an intrinsic HTML Element (such as div, a, span, etc...)
+             * for the sake of simplicity, we assume here that they all support the style prop. if we need more detailed
+             * information for them, feel free to turn this into a real data structure that contains specific props for specific elements,
+             * but for now, I just return a one-size-fits-all PropertyControls result here
+             */
+            return HtmlElementStyleObjectProps
+          } else {
+            // you can add more intrinsic (ie not imported) element types here
```
Rheeseyb, line 413: I believe R3F elements (e.g. `mesh`) will fall into this category as they're not explicitly imported (e.g. in https://codesandbox.io/s/rrppl0y8l4?file=/src/App.js)

editor/src/templates/property-controls-info.tsx
Rheeseyb: Is it worth adding that as a comment into the code?

enidemi:
Closing this PR as the issue found with how the metadata is constructed will be solved in a different PR.

Issue #1408, opened 2021-06-21
Colour picker scrubbing makes changes arrive twenty seconds later in the code editor


maltenuhn:
With the VSCode changes, we have a pretty knarly performance regression that seems to affect some scrubbable controls:

STR:
- new project
- select innermost div
- use the background fill selector from the inspector
- scrub any of the controls inside the colour picker. Stop, slide again.

Observed: queues up updates to the code editor that can arrive up to twenty seconds later.
Expected: shouldn't : ) 

NB: this does work "as expected" with opacity - canvas updates real-time, code updates at end of scrub. List of things to check:

- [x] `number-control` with scrubbable label
- [x] `slider control` (seems fine, maybe optimisations only exist for opacity?)
- [x] all scrubbable controls in the colour picker
- [x] all scrubbable controls in the gradient selector, including the gradient stops

Rheeseyb:
Closed by #1409 and #1410

PR #1409, opened 2021-06-21
Only send the last file write to VS Code for a given file

Rheeseyb:
First part of the fix to #1408

**Problem:**
When accumulating the changes to send to VS Code, we were still sending multiple successive writes to a given file, each of which contains the full file contents.

**Fix:**
Filter all but the last file write update for each file, leaving other file changes as they were.

### DIFF
--- a/editor/src/components/editor/store/vscode-changes.ts
+++ b/editor/src/components/editor/store/vscode-changes.ts
@@ -10,7 +10,7 @@ import {
   getUnsavedCodeFromTextFile,
 } from '../../../core/model/project-file-utils'
 import { ProjectFile, isTextFile } from '../../../core/shared/project-file-types'
-import { isBrowserEnvironment } from '../../../core/shared/utils'
+import { fastForEach, isBrowserEnvironment } from '../../../core/shared/utils'
 import {
   applyProjectChanges,
   getCodeEditorDecorations,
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):

editor/src/components/editor/store/vscode-changes.ts
```diff
@@ -209,12 +209,36 @@ export interface AccumulatedVSCodeChanges {
...
+  first: Array<ProjectChange>,
+  second: Array<ProjectChange>,
+): Array<ProjectChange> {
+  let writeFilePathsSeen: Set<string> = new Set()
+  let reversedResult: Array<ProjectChange> = []
+  fastForEach([...first, ...second].reverse(), (change) => {
+    if (change.type === 'WRITE_PROJECT_FILE') {
+      if (writeFilePathsSeen.has(change.fullPath)) {
+        // We only want the last version of each file
+        return
```
seanparsons, line 222: Doesn't this just exit the whole thing regardless of there being other file changes?

seanparsons (approved):

editor/src/components/editor/store/vscode-changes.ts
Rheeseyb: This just exits this iteration of the for loop (because this is inside the `fastForEach`'s anonymous function). If it were in a regular `for( ... of ...)` then it would exit the `combineFileChanges`

seanparsons: But that's what I'm saying, if there's a change for file `A` and then two changes for file `B`, this will result in the change for file `A` being lost surely?

seanparsons: Oh no, ignore me, this is being run per-file so it's not an issue.

PR #1411, opened 2021-06-22
Support spread operators at the root of an element.

### LINKED ISSUE #1365: spreading in a mapped jsx element removes it from `elementsWithin`
image

if you have a mapping like

```javascript
<div data-uid='mapper-parent'>
{props.data.map((entry) => (
        <Card
          stlye={{ flexBasis: 50 }}
          name={entry.name}
          country={entry.country}
          image={entry.image}
        />
      ))}
</div>
```
this works. but if you have something like this:
```javacript
<div data-uid='mapper-parent'>
{props.data.map((entry) => (
        <Card
          {...entry}
          stlye={{ flexBasis: 50 }}
        />
      ))}
</div>
```
Then the navigator collapses the mapped elements. The JSXElement for `mapper-parent` will have an empty `elementsWithin`, which means we are not spying on the mapped Card instances.

seanparsons:
Fixes #1365

**Problem:**
Currently if a user includes a root level spread operator in an element the element becomes an arbitrary block to the parser, which means it becomes effectively opaque.

**Fix:**
Added support for parsing and printing root level spread operators, so that elements can continue to be parsed validly when those are present. Currently these are handled correctly as far as rendering and reproduction, but are not manipulable when it comes to modifying their values as it seems like this would be a rare case compared to their use for spreading an object passed in as part of the props.

**Commit Details:**
- Fixes #1365.
- Created the `JSXAttributesSpread` type to represent the segment
  that contains a spread.
- `JSXAttributes` is now an array of `JSXAttributesEntry` and/or
  `JSXAttributesSpread`.
- Updated various functions like `getJSXAttribute` to cater for the
  changed type of `JSXAttributes`.
- Updated the parser logic to pull out the root level spread.
- Code printer also prints out the root level spreads.
- Added the new hierarchy of types in the deep equality calls.

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer-bugs.spec.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer-bugs.spec.ts
@@ -232,7 +232,7 @@ export var ${BakedInStoryboardVariableName} = (
 `
     testParseThenPrint(code, code)
   })
-  it('does something odd for spread mapping - #1365', () => {
+  it('parses elements that use props spreading - #1365', () => {
     const spreadCode = `import * as React from 'react'
 const Test = (props) => {
   return (
--- a/editor/src/core/workers/parser-printer/parser-printer-bugs.spec.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer-bugs.spec.ts
@@ -86,7 +86,7 @@ export var App = props => {
       fail(parsedPlainCode)
     }
   })
-  it('does something odd for spread mapping - #1365', () => {
+  it('parses elements that use props spreading - #1365', () => {
     const nonSpreadCode = `
     import * as React from "react"
     import { Card } from './card'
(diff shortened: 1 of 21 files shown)


editor/src/core/workers/parser-printer/parser-printer-bugs.spec.ts
```diff
@@ -85,6 +86,63 @@ export var App = props => {
       fail(parsedPlainCode)
     }
   })
+  it('does something odd for spread mapping - #1365', () => {
```
Rheeseyb, line 89:
```suggestion
  it('parses elements that use props spreading - #1365', () => {
```
? (also below)

seanparsons: Ah, since this was in a file with `-bugs` in the filename, I wrote that from that viewpoint. But I'm not at all attached to it.

Rheeseyb (approved):

balazsbajorics (approved):

PR #1412, opened 2021-06-22
Base the cyclic dependency resolution in the canvas on the import origin

### LINKED ISSUE #1361: Importing App before Card means Card is not getting spied on the Canvas
test project https://utopia.app/p/0c33412b-freckle-afternoon/

if you import App (which imports Card) before importing Card, it means Card is not going to be auto-spied by the scene hosting Card. 

However you can still drill-to-spy on Card inside the Scene inside App

Rheeseyb:
Fixes #1361
Also fixes the issue outlined in [this comment](https://github.com/concrete-utopia/utopia/pull/1200/files#r626512793), meaning that (in that comment) it is now possible to focus `A/B/C`

**Problem:**
When resolving files in the canvas, we were defending against cyclic imports by checking if the resolved path had already been previously resolved, falling back to the default handling (meaning no spying) if so. The issue here being that we were ignoring the context of where that file was being resolved from.

**Fix:**
Instead of the above, we track the previously resolved paths (and prevent cyclic dependencies from killing the canvas) by keeping a map from the `importOrigin` to all of the `toImport` paths that it imports.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas.tsx
@@ -281,45 +281,49 @@ export const UiJsxCanvas = betterReactMemo(
 
     // TODO after merge requireFn can never be null
     if (requireFn != null) {
-      let resolvedFiles: string[] = []
+      let resolvedFiles: MapLike<Array<string>> = {} // Mapping from importOrigin to an array of toImport
       const customRequire = React.useCallback(
         (importOrigin: string, toImport: string) => {
-          const filePathResolveResult = resolve(importOrigin, toImport)
+          if (resolvedFiles[importOrigin] == null) {
+            resolvedFiles[importOrigin] = []
+          }
+          let resolvedFromThisOrigin = resolvedFiles[importOrigin]
+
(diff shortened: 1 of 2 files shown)


seanparsons (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -281,45 +281,49 @@ export const UiJsxCanvas = betterReactMemo(
...
         (importOrigin: string, toImport: string) => {
-          const filePathResolveResult = resolve(importOrigin, toImport)
+          if (resolvedFiles[importOrigin] == null) {
+            resolvedFiles[importOrigin] = []
+          }
+          let resolvedFromThisOrigin = resolvedFiles[importOrigin]
+
+          const alreadyResolved = resolvedFromThisOrigin.includes(toImport) // We're inside a cyclic dependency, so trigger the below fallback
+          const filePathResolveResult = alreadyResolved
+            ? left<string, string>('Already resolved')
```
balazsbajorics, line 294: so what I _still_ don't understand is why can't we just return the already resolved (and cached) result here?

balazsbajorics (approved):
I hope that one day we can significantly simplify this, but for now 👍

enidemi (approved):

PR #1418, opened 2021-06-22
Correct the parsing of hsl colours.

### LINKED ISSUE #1216: Inspector does not show correct background colour
STR: create a div, give it a backgroundColour key in style of `hsl(0,0%,95%)` or `hsl(0,0%,95%, 1)`

Observed: inspector shows `#FFFFFF`
Expected: shows the correct colour

Sample project here https://utopia.pizza/p/ecebe791-foregoing-trunk/
Screenshot 2021-05-07 at 10 50 16

seanparsons:
Fixes #1216

**Problem:**
Previously when parsing `hsl(...)` colours, the wrong range was used when invoking the conversion function in Chroma.

**Fix:**
Reduced the range of saturation and lightness values when passing them to Chroma.

**Commit Details:**
- Fixes #1216.
- When invoking the `Chroma.hsl` function, reduce the saturation
  and lightness values down to the range `[0..1]`.

### DIFF
--- a/editor/src/components/inspector/common/css-utils.ts
+++ b/editor/src/components/inspector/common/css-utils.ts
@@ -2055,7 +2055,7 @@ export function cssColorToChromaColor(c: CSSColor): Either<string, Chroma.Color>
   } else if (isHex(c)) {
     return right(Chroma.hex(c.hex))
   } else if (isHSL(c)) {
-    return right((Chroma.hsl as any)(c.h, c.s, c.l, c.a))
+    return right((Chroma.hsl as any)(c.h, c.s / 100.0, c.l / 100.0, c.a))
   } else {
     return right(Chroma(c.r, c.g, c.b, c.a))
   }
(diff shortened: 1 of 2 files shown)


Rheeseyb (approved):

enidemi (approved):

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -2055,7 +2055,7 @@ export function cssColorToChromaColor(c: CSSColor): Either<string, Chroma.Color>
   } else if (isHex(c)) {
     return right(Chroma.hex(c.hex))
   } else if (isHSL(c)) {
-    return right((Chroma.hsl as any)(c.h, c.s, c.l, c.a))
+    return right((Chroma.hsl as any)(c.h, c.s / 100.0, c.l / 100.0, c.a))
```
balazsbajorics, line 2058: is this `c.s / 100.0` different from `c.s / 100` in javascript?

balazsbajorics (approved):

editor/src/components/inspector/common/css-utils.ts
seanparsons: I don't believe so, they're apparently equal, it just seemed like the right thing to do given that it's going to be double-like.

PR #1423, opened 2021-06-23
ux/slider behaviour

maltenuhn:
**Problem:**
For years now, we've had scrubbing/sliding behaviours that could trigger other hover effects in the UI. It's a detail, but it _really_ felt wrong / brittle. 

**Fix:**
This trivially fixes the problem by temporarily rendering a hover-effect-blocking div for the scrubbable label, and the slider. It doesn't include the sliding behaviours of the colour picker. 

NB we discussed in the past to do this via on-click-outside; can confirm this won't work (you can put a wrapper div, but that then will swallow the first click even while _not_  using the slider. I guess we could do this via editor state etc, but not only does that feel like a sledge hammer, but that would make it v difficult to put the insulating div in the right place, z-index-wise.

### DIFF
--- a/editor/src/components/inspector/controls/slider-control.tsx
+++ b/editor/src/components/inspector/controls/slider-control.tsx
@@ -20,7 +20,101 @@ export type SliderControlProps = DEPRECATEDControlProps<number> & {
   DEPRECATED_controlOptions: DEPRECATEDSliderControlOptions
 }
 
-export class SliderControl extends React.Component<SliderControlProps> {
+export const SliderControl: React.FunctionComponent<SliderControlProps> = (props) => {
+  const [isSliding, setIsSliding] = React.useState(false)
+
+  const handleBeforeChange = React.useCallback(() => {
+    setIsSliding(true)
+  }, [])
+
+  const onChangeFn = props.onForcedSubmitValue ?? props.onSubmitValue
+
+  const handleOnAfterChange = React.useCallback(
+    (n: number) => {
+      onChangeFn(n)
+      setIsSliding(false)
+    },
+    [onChangeFn],
+  )
+
+  const controlOptions = {
+    ...defaultSliderControlOptions,
+    ...props.DEPRECATED_controlOptions,
+  }
+
+  let marks: Marks = {}
+  if (typeof controlOptions.origin === 'number') {
+    marks[controlOptions.origin] = {
+      style: {
+        display: 'none',
+      },
+      label: String(controlOptions.origin),
+    }
+  }
+  return (
+    <FlexRow
+      style={{
+        width: '100%',
+        height: UtopiaTheme.layout.inputHeight.default,
+        paddingLeft: 4,
+        paddingRight: 4,
+        alignItems: 'center',
+        ...props.style,
+      }}
+      className={props.controlClassName}
+      onContextMenu={props.onContextMenu}
+    >
+      <Slider
+        disabled={!props.controlStyles.interactive}
+        value={props.value}
+        onBeforeChange={handleBeforeChange}
+        onChange={props.onTransientSubmitValue}
+        // onAfterChange={props.onForcedSubmitValue ?? props.onSubmitValue}
+        onAfterChange={handleOnAfterChange}
+        min={controlOptions.minimum}
+        max={controlOptions.maximum}
+        step={controlOptions.stepSize}
+        marks={marks}
+        handleStyle={{
+          backgroundColor: controlOptions.filled ? props.controlStyles.trackColor : undefined,
+        }}
+        trackStyle={{
+          backgroundColor: controlOptions.filled ? props.controlStyles.trackColor : undefined,
+        }}
+        activeDotStyle={{
+          backgroundColor: controlOptions.filled
+            ? props.controlStyles.trackColor
+            : props.controlStyles.railColor,
+        }}
+        railStyle={{
+          backgroundColor: props.controlStyles.railColor,
+        }}
+        dotStyle={{
+          backgroundColor: props.controlStyles.railColor,
+        }}
+      />
+      {/* This div blocks other controls from receiving hover events 
+          while you're sliding the slider. NB this needs to come here, 
+          i.e. *after* the Slider component - so it doesn't obscure it. 
+      */}
+      {isSliding ? (
+        <div
+          style={{
+            position: 'fixed',
+            background: 'transparent',
+            left: 0,
+            top: 0,
+            right: 0,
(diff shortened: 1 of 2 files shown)


editor/src/components/inspector/controls/slider-control.tsx
```diff
@@ -20,7 +20,101 @@ export type SliderControlProps = DEPRECATEDControlProps<number> & {
...
+            bottom: 0,
+            zIndex: 1,
+          }}
+        />
+      ) : null}
+    </FlexRow>
+  )
+}
+
+export class SliderControl2 extends React.Component<SliderControlProps> {
```
Rheeseyb, line 117: Delete me!

editor/src/components/inspector/controls/slider-control.tsx
```diff
@@ -20,7 +20,101 @@ export type SliderControlProps = DEPRECATEDControlProps<number> & {
...
+      }}
+      className={props.controlClassName}
+      onContextMenu={props.onContextMenu}
+    >
+      <Slider
+        disabled={!props.controlStyles.interactive}
+        value={props.value}
+        onBeforeChange={handleBeforeChange}
+        onChange={props.onTransientSubmitValue}
+        // onAfterChange={props.onForcedSubmitValue ?? props.onSubmitValue}
```
Rheeseyb, line 72:
```suggestion
```

editor/src/components/inspector/controls/slider-control.tsx
```diff
@@ -20,7 +20,101 @@ export type SliderControlProps = DEPRECATEDControlProps<number> & {
...
+      {isSliding ? (
+        <div
+          style={{
+            position: 'fixed',
+            background: 'transparent',
+            left: 0,
+            top: 0,
+            right: 0,
+            bottom: 0,
+            zIndex: 1,
```
Rheeseyb, line 109: Is the `zIndex` required?

editor/src/uuiui/inputs/number-input.tsx
```diff
@@ -760,19 +760,35 @@ export const NumberInput = betterReactMemo<NumberInputProps>(
...
-              textAlign: 'center',
-              display: 'block',
-              color: controlStyles.secondaryColor,
-              paddingTop: 2,
-            }}
-          >
-            {DEPRECATED_labelBelow != null ? DEPRECATED_labelBelow : null}
-          </div>
+          <React.Fragment>
+            {isFauxcused ? (
```
Rheeseyb, line 764: `isFauxcused` I feel sick

editor/src/components/inspector/controls/slider-control.tsx
maltenuhn: Sadly, yes. Without this, it ends up behind some of the inspector inputs.

editor/src/uuiui/inputs/number-input.tsx
```diff
@@ -760,19 +760,35 @@ export const NumberInput = betterReactMemo<NumberInputProps>(
...
+              <div
+                css={{
```
Rheeseyb, line 766: Any reason why you used `css` here rather than `style`?

maltenuhn: No, left over from debugging (via label / hover). Fixed.

editor/src/uuiui/inputs/number-input.tsx
maltenuhn: I went through _all_ of the emotions on reading this. Think I settled on something like angryupvote :D

Rheeseyb (approved):
Delete `SliderControl2` and the commented out prop and then this looks good to me

maltenuhn:
> Delete `SliderControl2` and the commented out prop and then this looks good to me

awkward. Done.

editor/src/components/inspector/controls/slider-control.tsx
```diff
@@ -20,7 +20,101 @@ export type SliderControlProps = DEPRECATEDControlProps<number> & {
...
+
+  const handleOnAfterChange = React.useCallback(
+    (n: number) => {
+      onChangeFn(n)
+      setIsSliding(false)
+    },
+    [onChangeFn],
+  )
+
+  const controlOptions = {
```
seanparsons, line 40: I believe these variables will need wrapping in `useMemo` invocations, otherwise they'll be regenerated every time.

editor/src/components/inspector/controls/slider-control.tsx
```diff
@@ -20,7 +20,101 @@ export type SliderControlProps = DEPRECATEDControlProps<number> & {
...
+        value={props.value}
+        onBeforeChange={handleBeforeChange}
+        onChange={props.onTransientSubmitValue}
+        // onAfterChange={props.onForcedSubmitValue ?? props.onSubmitValue}
+        onAfterChange={handleOnAfterChange}
+        min={controlOptions.minimum}
+        max={controlOptions.maximum}
+        step={controlOptions.stepSize}
+        marks={marks}
+        handleStyle={{
```
seanparsons, line 78: This will also be a newly created object, which the Slider may re-render each time.

seanparsons (approved):
It looks like there's a bunch of value re-generation going on in this.

editor/src/components/inspector/controls/slider-control.tsx
maltenuhn: Thanks! I've moved all the conditional styles in to useMemo fns - do I need to do that for the unconditional styles as well? Eg
```
dotStyle={{
          backgroundColor: props.controlStyles.railColor,
        }}
```

editor/src/components/inspector/controls/slider-control.tsx
maltenuhn: fixed

editor/src/components/inspector/controls/slider-control.tsx
seanparsons: I believe so, that's right isn't it @balazsbajorics ?

balazsbajorics: for totally static styled, the easiest is to just put them in the file root. then they will only be generated once, when loading the editor.

btw I don't know if the `<Slider />` component is memoized. If it's not memoized, this is all moot

maltenuhn: These still depend on the props. I've made the changes, no idea about the internals of rc-slider but it seems our own Slider is not memoized. 

Btw is it possible that making these changes (useMemo) create a significant performance hit?

PR #1424, opened 2021-06-23
Fix: improve our ability to spy on elements which do not show up in the DOM

balazsbajorics:
**Problem:**
We had two problems:
1. react-three-fiber and other libraries render to an HTML Canvas, which means the rendered elements will not show up in the DOM. We used the dom-walker's result to cull / clean the spy metadata, which meant that we would remove valid metadata entries for these components, and only show them as Element in the navigator.
2. The dom-walker was only triggered by editor interactions. If something updated on the canvas but it was not the effect of an EditorAction, we would only re-run the DOM-walker on canvas scroll for example, leading to the navigator suddenly become "fixed" on first scroll. A good example is that if you go to Live mode on this project https://utopia.app/project/ad8f6416/ and use the + / - buttons to dynamically add / remove cards to the screen, the Navigator would not update. But if you go back to Edit mode, or press and hold cmd to scroll the canvas, the navigator would suddenly be updated with the correct number of Cards.

**Fix:**
1. Do not cull the spy metadata based on the DOM-walker. (only purge scenes if they were deleted)
2. For elements that render on the canvas without EditorAction, we now provide a new way to let the DOM-walker know that it should re-run: the `DomWalkerInvalidatePathsContext` contains a function that can be called at any time. We made sure it is properly throttled, so it's not a problem that every rendered element calls it individually
3. Only clean the spyMetadata accumulator on canvas remount / dom walker invalidation
4. as a bonus, we also made a new hook to manage our invalidated Sets instead of just mutating them inline. This change is not necessary to the working of the end result, but IMO it makes the dom-walker code cleaner and so we left it in.


editor/.eslintrc.js
```diff
@@ -161,19 +161,19 @@ module.exports = {
...
+    //     allowString: false,
+    //     allowNumber: false,
+    //     allowNullableObject: false,
+    //     allowNullableBoolean: true,
+    //     allowNullableString: false,
+    //     allowNullableNumber: false,
+    //     allowAny: false,
+    //     allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing: false,
+    //   },
+    // ],
```
Rheeseyb, line 176: Bin it! This damn rule is such a pain when committing :( 

```suggestion
```

balazsbajorics: I'm gonna chuck it on master

editor/src/components/canvas/dom-walker.ts
```diff
@@ -307,16 +311,72 @@ function useInvalidateInitCompleteOnMountCount(
 
   return [initCompleteRef.current, setInitComplete]
 }
+type ValueOrUpdater<S> = S | ((prevState: S) => S)
+// todo move to file
```
Rheeseyb, line 315: But to which file?

editor/src/components/canvas/dom-walker.ts
```diff
@@ -307,16 +311,72 @@ function useInvalidateInitCompleteOnMountCount(
...
+      let resolvedNewValue: S
+      if (isSimpleValue(valueOrUpdater)) {
+        // TODO make this type nicer using a type guard
+        resolvedNewValue = valueOrUpdater
+      } else {
+        resolvedNewValue = valueOrUpdater(stateRef.current)
+      }
+      stateRef.current = resolvedNewValue
+
+      // TODO invalidate
```
Rheeseyb, line 340: I don't understand this TODO - isn't that exactly what is happening below?

editor/src/components/canvas/dom-walker.ts
```diff
@@ -365,47 +425,14 @@ function mergeFragmentMetadata(
...
-  )
-  useInvalidateScenesWhenSelectedViewChanges(
-    invalidatedScenesRef,
-    invalidatedPathsForStylesheetCacheRef,
-  )
 
-  React.useLayoutEffect(() => {
+  const fireThrottledCallback = useThrottledCallback(() => {
     if (containerRef.current != null) {
+      // console.log('DOM_WALKER_START')
```
Rheeseyb, line 435:
```suggestion
```

editor/src/components/canvas/dom-walker.ts
```diff
@@ -449,11 +476,59 @@ export function useDomWalker(props: DomWalkerProps): React.Ref<HTMLDivElement> {
       // Fragments will appear as multiple separate entries with duplicate UIDs, so we need to handle those
       const fixedMetadata = mergeFragmentMetadata(metadata)
 
+      // console.log('DOM_WALKER_FINISH')
```
Rheeseyb, line 479:
```suggestion
```

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -58,8 +58,8 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(270) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(280)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThan(395) // if this breaks, GREAT NEWS but update the test please :)
+    expect(renderCountAfter - renderCountBefore).toBeLessThan(405)
```
Rheeseyb, line 62: YIKES!

editor/src/components/canvas/dom-walker.ts
```diff
@@ -307,16 +311,72 @@ function useInvalidateInitCompleteOnMountCount(
...
+function isSimpleValue<S>(valueOrUpdater: ValueOrUpdater<S>): valueOrUpdater is S {
+  return typeof valueOrUpdater !== 'function'
+}
+function useStateAsyncInvalidate<S>(
+  onInvalidate: () => void,
+  initialState: S,
+): [S, SetValueCallback<S>] {
+  const stateRef = React.useRef(initialState)
+  const setAndMarkInvalidated = React.useCallback(
+    (valueOrUpdater: ValueOrUpdater<S>, doNotInvalidate?: 'do-not-invalidate') => {
```
seanparsons, line 330: Wait, the text `'do-not-invalidate' | undefined` is the type? I find this a bit confusing.

balazsbajorics: you either omit this parameter, or you write `'no-not-invalidate'`. I typed it like this to avoid a magic `..., true)` on the call site

editor/src/components/canvas/dom-walker.ts
```diff
@@ -307,16 +311,72 @@ function useInvalidateInitCompleteOnMountCount(
...
+        onInvalidate()
+      }
+    },
+    [stateRef, onInvalidate],
+  )
+
+  return [stateRef.current, setAndMarkInvalidated]
+}
+
+function useThrottledCallback(callback: () => void): (immediate?: 'immediate') => void {
```
seanparsons, line 351: I know booleans are bad, but I find `immediate: boolean` much clearer than `immediate: 'immediate' | undefined`.

balazsbajorics: I think 
```typescript
 React.useLayoutEffect(() => {
     fireThrottledCallback('immediate')
   })
```
reads so much better than
```typescript
 React.useLayoutEffect(() => {
     fireThrottledCallback(true)
   })
```

We had / have this big regression in the require function because the createRequireFunction took a boolean parameter and we just missed updating it during a copy-paste, because no one went in to read the type, and the parameter name is not visible on the callsite.

editor/src/components/canvas/dom-walker.ts
seanparsons: I think if we want to avoid booleans it's much clearer to have a type of `'leave-alone' | 'invalidate'` than make it optional.

editor/src/components/canvas/dom-walker.ts
```diff
@@ -523,7 +592,11 @@ function collectMetadata(
     )
 
     const collectedMetadata = pathsForElement.map((path) => {
-      invalidatedPathsRef.current.delete(EP.toString(path))
+      updateInvalidatedPaths((current) => {
+        // sneaky mutation to improve performance
+        ;(current as Set<string>).delete(EP.toString(path))
```
seanparsons, line 597: Does this have the potential for sneaky bugs?

editor/src/components/canvas/dom-walker.ts
Rheeseyb: +1 to that

seanparsons (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
balazsbajorics: after testing this for a day, I believe the test is simply more honest now and this is not a true regression

balazsbajorics:
the tests pass, we fixed most of the complaints. the performance number is scary but I think it is because of the synthetic nature of the test. I don't expect this pr to cause a major performance regression. 

there's a minor performance regression it does cause: in some circumstances we will run the dom-walker for a second time. the reason we are doing this is to actually true up the editor state with the real measurements of the canvas, and not be one-frame lockstep out of date.
to fix the performance regression, the most straightforward way would be to shift around the render logic: on every dispatch, first render the canvas, then the rest of the editor. we had a system like this about 2.5 years ago, we left it during the big refactor, but I think there's no technical reason to not have it again.

however, I think this pr is fine to merge as-is.

Rheeseyb (approved):

enidemi (approved):

PR #1427, opened 2021-06-23
First stab at contributing guide

maltenuhn:
A start at some docs, mainly pulled from my chat log with Omar and focused on how our editor-state and redux architecture works (since almost all UI-level changes will probably need to touch that). It's _very_ basic but probably better than typing this up (and losing it) in Discord

### DIFF
--- a/contributing-docs.md
+++ b/contributing-docs.md
@@ -1,0 +1,58 @@
+## Brief Overview
+TL;DR: Code = React + TS, heavy use of useCallback / memo for performance. 
+
+NB: all filenames given without path unless there are two identical filenames. 
+
+### Architecture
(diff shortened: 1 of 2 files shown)


Rheeseyb:
We already have https://github.com/concrete-utopia/utopia/blob/master/contributing.md - is it worth appending this to the end of that? Or alternatively starting an effort to document everything under `docs/` and move the above file there?

Rheeseyb (approved):

maltenuhn:
@Rheeseyb I didn't realise! On reading it, it seems to be mainly about rules to comply with. `/docs` seems like the right place (or github pages?)

seanparsons (approved):

maltenuhn:
All right, I've moved this to `docs`. I hope it's juicy enough to attract a roving Docusaurus or two...

PR #1429, opened 2021-06-24
Support error mapping in multi-file stack traces

### LINKED ISSUE #1415: Missing source link in the error overlay when a RecoilRoot is missing
When trying to use Recoil, if the user is missing a `<RecoilRoot>` component wrapping a use of `useRecoilState`, the error message overlay displays an error `Error: This component must be used inside a <RecoilRoot> component.` with no link back to the source.

Sample project here (comment out the `RecoilRoot` element in storyboard.js): https://utopia.pizza/p/37fba76f-todo-app/

Rheeseyb:
Fixes #1415 

**Problem:**
The mapping from stack trace lines in errors to source files (for the sake of the error overlay) was using the assumption that we were only interested in the first line. The recoil error attached was tripping that because the first actually mappable line comes a lot later in the stack.

**Fix:**
Rather than only extracting the sourcemap for the first applicable stack frame we find and attempting to use that to enhance all frames, apply the process for each individual frame. Here's how it looks now (previously only the topmost mapping would be shown):

Screenshot_20210624_132317

**Commit Details:**
- Removed `parseUtopiaError.ts` and moved the relevant contents into `code-exec-utils.ts`
- Changed the stack frame mapping functions `react-error-overlay/utils/mapper.ts` to work on individual frames rather than the entire stack
- Updated `processErrorWithSourceMap` in `code-exec-utils.ts` to attempt to enhance each individual stack frame by checking if the frame has a source map attached to it
- Updated `testCanvasErrorInline` to handle each stack frame and include the file name of each stack frame, and updated the affected snapshots
- Added a test to cover the case where an error is thrown with multiple files in the stack trace, to ensure that each line in each file is correctly mapped

I recommend enabling "Hide whitespace changes" for this diff when checking the tests because of the amount of noise caused by updating the existing snapshots


editor/src/components/canvas/ui-jsx-canvas-errors.spec.tsx
```diff
@@ -828,110 +873,276 @@ export var ${BakedInStoryboardVariableName} = (props) => {
...
-            ScriptLine {
-              "content": "      export var storyboard = (props) => {",
-              "highlight": false,
+          ],
+        },
+      ]
+    `)
+  })
+
+  it('Handles errors with multifile stack traces', () => {
```
Rheeseyb, line 992: The new test

editor/src/components/canvas/ui-jsx-canvas-errors.spec.tsx
```diff
@@ -828,110 +873,276 @@ export var ${BakedInStoryboardVariableName} = (props) => {
...
+    }
+    `,
+      },
+    )
+    expect(canvasErrors).toMatchInlineSnapshot(`
+      Array [
+        Object {
+          "message": "I am an error!",
+          "name": "Error",
+          "stackFrames": Array [
```
Rheeseyb, line 1040: Note that there is now a stack frame for each of the 3 files in the stack

balazsbajorics (approved):

enidemi (approved):

editor/src/components/canvas/ui-jsx-canvas-errors.spec.tsx
```diff
@@ -32,45 +32,50 @@ describe('UiJsxCanvas errors', () => {
...
-              "highlight": false,
-              "lineNumber": 4,
-            },
-            ScriptLine {
-              "content": "    export var App = props => <MyCard data-uid={'bbb'} />",
-              "highlight": true,
+          "stackFrames": Array [
+            Object {
+              "columnNumber": 32,
+              "fileName": "code.tsx",
```
seanparsons, line 40: Is this filename relative to the file the exception originates from?

seanparsons (approved):

editor/src/components/canvas/ui-jsx-canvas-errors.spec.tsx
Rheeseyb: It's the actual file name. That's just what we're using in the test

seanparsons: I was only asking because some of the filenames in the snapshots were prefixed by a `/` and most are not, like this one has a mix in the same set of stack frames: https://github.com/concrete-utopia/utopia/pull/1429/files/7d5aebd3b147fb1eaaa75a9098a026d3f3afc1d7#diff-16cbaeb060f62d1bad49e833b2e2eb5a955edba01a78dc60e7f189a2bf63082dR1035

Rheeseyb: I'm pretty sure this is just an artefact of the test. The actual result on a real project shows the file paths from the root of the project (see the image above)

PR #1432, opened 2021-06-25
Fix the basic github import from the left pane

Rheeseyb:
**Problem:**
The Github Import functionality in the left pane wasn't hooked up

**Fix:**
Provide it with the most basic functionality, meaning that clicking "Start" will actually trigger the import in a new tab, and upon completion will add a storyboard file to the imported project.

### DIFF
--- a/editor/src/components/navigator/left-pane.tsx
+++ b/editor/src/components/navigator/left-pane.tsx
@@ -13,7 +13,7 @@ import { getAllUniqueUids } from '../../core/model/element-template-utils'
 import { getUtopiaJSXComponentsFromSuccess } from '../../core/model/project-file-utils'
 import { isParseSuccess, isTextFile, ProjectFile } from '../../core/shared/project-file-types'
 import { NO_OP } from '../../core/shared/utils'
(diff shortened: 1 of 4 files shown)


editor/src/components/navigator/left-pane.tsx
```diff
@@ -657,6 +660,28 @@ const SharingPane = betterReactMemo('SharingPane', () => {
 })
 
 const GithubPane = betterReactMemo('GithubPane', () => {
+  const [githubRepoStr, setGithubRepoStr] = React.useState('')
+  const parsedRepo = parseGithubProjectString(githubRepoStr)
```
seanparsons, line 664: Does this need memoisation?

seanparsons (approved):

editor/src/components/navigator/left-pane.tsx
Rheeseyb: Nah, because it will only be called whenever the `githubRepoStr` changes

enidemi (approved):

PR #1434, opened 2021-06-27
feat: added theme toggle to project settings

cscheffauer:
Fixes #[1417]

**Problem:**
Toggling the theme of the editor

**Fix:**
See commit details below

**Commit Details:**
- added theme toggle component to project settings
- added set_current_theme action
- added theme to editor state
- exported useColorTheme hook from theme
- consumed useColorTheme in functional components


Rheeseyb:
Thank you for this! Amazing work!

Heads up, I have just made the following changes so that we can merge this today:

- I've reverted the changes to the styled components in `documentation-components.tsx` - these changes actually broke the components since it changed them from styled components to regular functional components without props
- `repositionable-control.tsx` and `yoga-control.tsx` were updated to become functional components so that they could use the new `useColorTheme` hook, as they called a function that now required a `colorTheme` param
- I extracted the `themeOptions` from the `ProjectPane` component in `left-pane.tsx` so that we weren't creating a new object each render there
- Also in `left-pane.tsx` I reverted a change to the logic checking the user's login state (which looked to have been committed in error?)
- I added some memoisation of 3 components that you had converted into functional components (thank you!)

editor/src/uuiui/styles/theme.ts
```diff
@@ -190,8 +158,154 @@ const light = {
...
+  fg6: createUtopiColor('#6F778B', 'grey', 'grey'),
+  fg7: createUtopiColor('#525B72', 'grey', 'grey'),
+  fg8: createUtopiColor('#2F374A', 'lightgrey', 'lightgrey'),
+  fg9: createUtopiColor('#151A27', 'lightgrey', 'lightgrey'),
+  border1: createUtopiColor('#181C20', '', ''),
+  border2: createUtopiColor('#181C20', '', ''),
+  border3: createUtopiColor('#181C20', '', ''),
+}
+
+const darkPrimitives = {
```
Rheeseyb, line 196: Follow up for @maltenuhn - I guess you'll want to update the values used here? I'll create a ticket

Rheeseyb (approved):

editor/src/uuiui/styles/theme.ts
Rheeseyb: #1436

cscheffauer:
> Thank you for this! Amazing work!
> 
> Heads up, I have just made the following changes so that we can merge this today:
> 
> * I've reverted the changes to the styled components in `documentation-components.tsx` - these changes actually broke the components since it changed them from styled components to regular functional components without props
> * `repositionable-control.tsx` and `yoga-control.tsx` were updated to become functional components so that they could use the new `useColorTheme` hook, as they called a function that now required a `colorTheme` param
> * I extracted the `themeOptions` from the `ProjectPane` component in `left-pane.tsx` so that we weren't creating a new object each render there
> * Also in `left-pane.tsx` I reverted a change to the logic checking the user's login state (which looked to have been committed in error?)
> * I added some memoisation of 3 components that you had converted into functional components (thank you!)

Thanks a lot, @Rheeseyb.
Regarding the styled-components - we should create functional components for all of them and use styled-components within them. This will give us more flexibility when we want to pass in props, add state or hooks (like i was trying to).

Oh thank u, the logic for the user login state was a mistake :)

PR #1438, opened 2021-06-28
fixing npm dependencies that require react/jsx-runtime

### LINKED ISSUE #1437: Support @material-ui/core@next and other similar types of package
**Original problem**: trying to add  @material-ui/core@next fails with this error:

image

**Root cause**
This type of dependency isn't supported. Additionally, we should bump our version of React to latest 17 while that's not a user-editable dependency.

balazsbajorics:
Fixes #1437

**Problem:**
If you try to import a package from npm which uses `react/jsx-runtime` it would fail with an error message.

**Fix:**
The issue was that we have `react` as a hardcoded dependency, but the built-in dependency resolver is dead simple, so I had to manually add `react/jsx-runtime` as a new dependency. All is well now!

**Commit Details:**
- Add `react/jsx-runtime` to the list of built-in dependencies.
- Add `react/jsx-runtime` to the list of fixed aliases in the webpack config so we can import from 'react/jsx-runtime'

### DIFF
--- a/editor/src/core/es-modules/package-manager/built-in-dependencies-list.ts
+++ b/editor/src/core/es-modules/package-manager/built-in-dependencies-list.ts
@@ -1,6 +1,7 @@
 import * as UtopiaAPI from 'utopia-api'
 import * as UUIUI from '../../../uuiui'
 import * as UUIUIDeps from '../../../uuiui-deps'
+import * as ReactJsxRuntime from 'react/jsx-runtime'
 import * as React from 'react'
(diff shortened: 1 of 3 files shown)


maltenuhn (approved):

Rheeseyb (approved):
This is fine as a quick fix but are there other cases of `react/*` or `react-dom/*` that we need to support?

balazsbajorics:
> This is fine as a quick fix but are there other cases of `react/*` or `react-dom/*` that we need to support?

not to my knowledge.

PR #1439, opened 2021-06-29
Fix/assorted property control related fixes

balazsbajorics:
**Problem:**
While trying to use property controls for a first time in _a long while_ I encountered some issues.

**Commit Details:**
- jsxAttributeNestedArrays can now be updated with paths such as `['props', 'someArray', '3']`, we now properly treat the string '3' as an array index. This is necessary because `Object.keys([1, 2, 3])` returns `["0", "1", "2"]`...
- While I was fixing that, I realized that there's a bug in the behavior that converts the jsxAttributeNestedArray into a jsxAttributeNestedObject. I fixed that and wrote a test for it
- `isComponentInstance` only returns false if the selected element is a HTML DOM Host
- A fix to the array supercontrol in the inspector (via @enidemi )

### DIFF
--- a/editor/src/components/inspector/sections/component-section/component-section.tsx
+++ b/editor/src/components/inspector/sections/component-section/component-section.tsx
@@ -286,7 +286,7 @@ const RowForArrayControl = betterReactMemo(
     )
 
     const rowHeight = UtopiaTheme.layout.rowHeight.max
-    const transformedValue = Array.isArray(value) ? value : []
+    const transformedValue = Array.isArray(value) ? value : [value]
     const { springs, bind } = useArraySuperControl(
       transformedValue,
       onSubmitValue,
--- a/editor/src/components/inspector/sections/component-section/component-section.tsx
+++ b/editor/src/components/inspector/sections/component-section/component-section.tsx
@@ -286,7 +286,13 @@ const RowForArrayControl = betterReactMemo(
     )
 
     const rowHeight = UtopiaTheme.layout.rowHeight.max
-    const { springs, bind } = useArraySuperControl(value, onSubmitValue, rowHeight, false)
+    const transformedValue = Array.isArray(value) ? value : []
+    const { springs, bind } = useArraySuperControl(
+      transformedValue,
+      onSubmitValue,
+      rowHeight,
+      false,
+    )
     const [insertingRow, setInsertingRow] = React.useState(false)
 
     let warningTooltip: string | undefined = undefined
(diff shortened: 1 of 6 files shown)


editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -286,7 +286,13 @@ const RowForArrayControl = betterReactMemo(
     )
 
     const rowHeight = UtopiaTheme.layout.rowHeight.max
-    const { springs, bind } = useArraySuperControl(value, onSubmitValue, rowHeight, false)
+    const transformedValue = Array.isArray(value) ? value : []
```
Rheeseyb, line 289: I think this should instead be:
```const transformedValue = Array.isArray(value) ? value : [value]```
As otherwise it runs the risk of all-too-easily accidentally binning off the original value if the control was incorrect.

File: editor/src/core/model/element-metadata-utils.ts [hunk 2]
```diff
@@ -1278,7 +1278,7 @@ export const MetadataUtils = {
   },
   isComponentInstance(path: ElementPath, rootElements: Array<UtopiaJSXComponent>): boolean {
     const elementName = MetadataUtils.getStaticElementName(path, rootElements)
-    return elementName != null && !isIntrinsicElement(elementName)
+    return elementName != null && !isIntrinsicHTMLElement(elementName)
```
Rheeseyb, line 1281: I think this is fine but I wonder if we're missing anything else specific to non-HTML intrinsic elements :thinking:

editor/src/components/inspector/sections/component-section/component-section.tsx
balazsbajorics: pushing a fix

enidemi (approved):

editor/src/core/model/element-metadata-utils.ts
balazsbajorics: I combed over the code and pushing a commit where I fix a few more places that were confusing the two

editor/src/core/shared/jsx-attributes.ts
```diff
@@ -414,6 +414,9 @@ export function deeplyCreatedValue(path: PropertyPath, value: JSXAttribute): JSX
   }, value)
 }
 
+function isArrayIndex(maybeIndex: string | number): maybeIndex is number {
+  return !isNaN(maybeIndex as any)
```
Rheeseyb, line 418: We have a function `parseNumber` in `math-utils.ts` that is similar, but calls `parseFloat` first. It seems like that would be a safer option here.

Rheeseyb (approved):

PR #1440, opened 2021-06-29
Add an entry point for a react-three-fiber schema

balazsbajorics:
**Problem:**
react-three-fiber uses intrinsic elements which means we don't have an imported component to attach our `.propertyControls` to. We do have a way to describe third party and intrinsic elements, and this PR adds a super barebones schema for a few react-three-fiber elements.

**Commit Details:**
- Extended `property-controls-utils` so we can add multiple schemas describing intrinsic elements
- Sketched up a barebones first version of a react-three-fiber schema

**Notes:**
We don't have a nice control for Vector3 so I hacked up an object control that uses 3 number controls.

image

### DIFF
--- a/editor/src/core/property-controls/property-controls-utils.ts
+++ b/editor/src/core/property-controls/property-controls-utils.ts
@@ -27,6 +27,7 @@ import {
   UtopiaJSXComponent,
   JSXElement,
   getJSXElementNameLastPart,
+  isIntrinsicElement,
 } from '../shared/element-template'
 import {
   esCodeFile,
@@ -42,6 +43,7 @@ import {
   getOpenUIJSFileKey,
   EditorState,
   withUnderlyingTarget,
+  packageJsonFileFromProjectContents,
 } from '../../components/editor/store/editor-state'
 import { MetadataUtils } from '../model/element-metadata-utils'
 import { HtmlElementStyleObjectProps } from '../third-party/html-intrinsic-elements'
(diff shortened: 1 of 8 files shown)


editor/src/core/property-controls/property-controls-utils.ts
```diff
@@ -397,20 +401,35 @@ export function getPropertyControlsForTarget(
...
+             * for the sake of simplicity, we assume here that they all support the style prop. if we need more detailed
+             * information for them, feel free to turn this into a real data structure that contains specific props for specific elements,
+             * but for now, I just return a one-size-fits-all PropertyControls result here
+             */
+            return HtmlElementStyleObjectProps
+          } else {
+            // you can add more intrinsic (ie not imported) element types here
+            const packageJsonFile = packageJsonFileFromProjectContents(projectContents)
+            const dependencies = dependenciesFromPackageJson(packageJsonFile)
+            if (dependencies.some((dependency) => dependency.name === '@react-three/fiber')) {
```
seanparsons, line 417: Could this logic live in `third-party-components.ts`?

seanparsons (approved):

enidemi (approved):

editor/src/core/property-controls/property-controls-utils.ts
seanparsons: Sad ignored feedback...

PR #1441, opened 2021-06-29
Spike of classroom support.

seanparsons:
This is a _really_ rough and ready spike of classroom support, which triggers it with a `classroom` query parameter.

https://user-images.githubusercontent.com/217400/123823658-4e77e700-d8f5-11eb-8ff8-47cfad608ef8.mp4

### To experiment with this do the following:
- Open a project as the controller with a URL like this: `http://localhost:8000/p/<projectid>/?classroom=controller`
- Shortly afterwards open the same project as the attendant with a URL like this: `http://localhost:8000/p/<projectid>/?classroom=attendant`


Rheeseyb:
Tagged and closing

PR #1443, opened 2021-06-30
Fix: React elements in props

balazsbajorics:
**Problem:**
When writing code like `<App left={<Card />} right={<Card />} />` which passes in react elements as props, if those elements referenced a user-written component, the canvas would throw an error:
image

The reason is that when parsing the jsx attributes, we did not feed the parser with the elementsWithin stuff that replaces the component references with the utopiaCanvasJsxLookup function.

**Fix:**
With @seanparsons 's help the fix was pretty straightforward, I just had to weave thorugh `elementsWithin` from the parser all the way through `JSXAttributeOtherJavaScript` to `renderCoreElement`. 

**Commit Details:**
- Observe `parsedElementsWithin` in `parseArbitraryNodes`
- Add `elementsWithin` to `JSXAttributeOtherJavaScript`
- Make new function `getAccumulatedElementsWithin` to be able to collect all the elementsWithin before feeding it to `createLookupRender `
- Use `createLookupRender` in `renderCoreElement`/`JSX_ELEMENT` and put it in the scope before calling `jsxAttributesToProps`


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -131,6 +135,10 @@ function monkeyUidProp(uid: string | undefined, propsToUpdate: MapLike<any>): Ma
   return monkeyedProps
 }
 
+function NoOpLookupRender(element: JSXElement, scope: MapLike<any>): React.ReactChild {
+  throw new Error('Utopia Error: createLookupRender was not used properly')
```
Rheeseyb, line 139: It's worth including the element's name or type in the error here to help us debug if this does happen

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -158,7 +166,41 @@ export function renderCoreElement(
...
+            jsxFactoryFunctionName,
+            shouldIncludeCanvasRootInTheSpy,
+            filePath,
+            imports,
+            code,
+            highlightBounds,
+          )
+        : NoOpLookupRender
+
+      const blockScope = {
```
seanparsons, line 194: I guess we could also skip creating this as well right?

seanparsons (approved):

editor/src/core/shared/jsx-attributes.ts
```diff
@@ -859,6 +861,17 @@ function walkAttribute(
   }
 }
 
+export function getAccumulatedElementsWithin(attributes: JSXAttributes): ElementsWithin {
+  let elementsWithinAccumulator: ElementsWithin = {}
+  walkAttributes(attributes, (attribute, path) => {
+    if (isJSXAttributeOtherJavaScript(attribute)) {
+      Object.assign(elementsWithinAccumulator, attribute.elementsWithin)
```
Rheeseyb, line 868: Is there a reason for the `Object.assign` here, rather than just `elementsWithinAccumulator = attribute.elementsWithin`?

balazsbajorics: this code is copying the values from attribute.elementsWithin to elementsWithinAccumulator

Usually we would write `elementsWithinAccumulator = {...elementsWithinAccumulator, ... attribute.elementsWithin}` but that desugars to `Object.assign({}, elementsWithinAccumulator, attribute.elementsWithin)` which copies elementsWithinAccumulator _and_  attribute.elementsWithin  as well into a new empty object.

While writing it I got nervous about not blowing the performance on a hot path such as the renderCoreElement

Rheeseyb (approved):

PR #1446, opened 2021-06-30
Spike/component state in inspector

enidemi:
This spike is exploring what options there are to show and give control over React useState to users. 
In this video it's visible that the App component uses the number of cards shown from the state. It is showing the current number of cards on the right side in the inspector Component Section. 

With this it's also possible to manipulate the state, allowing setting different states to check the application with different number of list items, optionally rendered content, toggling buttons etc. without manually triggering the state by clicking in the App. 

https://user-images.githubusercontent.com/4403069/123987139-3cff0f80-d9c7-11eb-9643-3676cd91752b.mov


Rheeseyb:
Tagged and closing

PR #1451, opened 2021-07-01
Chore/typescript esmodule interop

balazsbajorics:
**Problem:**
We never enabled `esModuleInterop` but I think it's time we do

(edit: I'll write a less snarky PR description once the tests pass)


Rheeseyb:
Should this be closed?

PR #1457, opened 2021-07-05
esModuleInterop

balazsbajorics:
This is a tech-debt PR. Even though I fix some code in preparation for Jest 27, this PR does not switch our jest version or how we run jest tests. So there's no immediate performance benefits here, just working down the required changes so I can hopefully turn on Jest 27 soon.
There's some changes in this PR (that I will mark) that are unrelated and about making the code future proof for Jest 27:
- `it('test', async (done) =>` is not supported anymore, a test case can either be async or it can take a `done`
- there were some timer changes which I solved
- I add configuration for babel (for babel-jest), but babel-jest is NOT enabled in this PR
- however, I did fix a few complaints generated by babel-jest: it throws errors if jest.mock calls are using things out of scope, unless their name starts with `module*` (case insensitive)

---------

I was looking into using babel-typescript to transpile our code. This is relevant to us both because of Jest tests, but also because this would enable us to try things like Vite.js for development.
There's a list of deprecated typescript features that needed removing for babel-typescript. Most notably you can't use `export const enum` but only legacy code used it anyways, and we can't use the old style import (`import Something = require('something')`) which we only ever accidentally used. 
But then as I was trying to run the editor and tests, I realized that the babel transpile did something differently wrt how the imports are transpiled, and this triggered me to finally turn on `esModuleInterop` and `allowSyntheticDefaultImports`. (This is something that was on the bottom of my TODO list for more than a year now, because we had this issue where for some files the .d.ts was suggesting that the module is exported as a default, but in reality it had to be imported with the `import * as` style.)

This PR introduces the following changes to our `tsconfig.json`: 
```json
"esModuleInterop": true,
"allowSyntheticDefaultImports": true,
"isolatedModules": true,
```

Most of the rest of the PR is dealing with the fallout, for most errors I had to fix whether something is * imported or default imported.

## Changes about the monkey patching:
The single biggest change caused by this PR:
The `_interopRequireWildcard` helper created for `esModuleInterop` clones the required object. Which means that if this _interopRequireWildcard runs _before_ monkey patching, the cloned object will be unaffected.
Babel also hoists all imports to the top of the module, meaning we CAN NOT call `applyUIDMonkeyPatch()` before React is imported in modules. (This will be important once we switch to babel-jest, not in this PR)
This only affects places which import react like so: `import * as React from 'react'` 
if you import it like this: `import React from 'react'` then the monkey patching works.

This DOES NOT AFFECT the user's evaluated code, only our own code. So I used this as an accidental feature to be able to select where I import React as * or as default. This way, our own UI code will not be monkey-wrapped anymore!

There's some caveats though: we have a bunch of _test code_ that directly evaluates in the test environment (meaning it's not a full end-to-end test). These tests were relying on the accidental fact that the react monkey patching was so pervasive that it affected our own React code as well. I fixed most of these tests.

The one area most affected is the tests which rely on `builtinDependencies` which are injected for the tests using jest.mock. I had to disable a recent react-three-fiber test, and I had to delete an antd test. I think these not as important to block this PR, I will look into fixing the r3f test in the future.

------------

While making this PR I switched to Jest 27 and babel jest, because current jest is so broken on my machine it is not capable of printing error messages. I am happy to report that Jest 27 (with jest-circus instead of jasmine2) plus babel-jest makes the tests run five times faster on my machine! However, there's some kinks with our Electron tests which I haven't yet solved, so I decided to make this PR only about the tsconfig and code fixed.


editor/src/components/canvas/canvas-types.ts
```diff
@@ -22,7 +22,7 @@ import { xor } from '../../core/shared/utils'
 
 export const CanvasContainerID = 'canvas-container'
 
-export const enum CSSCursor {
+export enum CSSCursor {
```
balazsbajorics, line 25: `const enum`s are a special thing in typescript and are not supported by babel-typescript. We don't really care about the difference so I changed them all.

editor/src/components/canvas/console-proxy.ts
```diff
@@ -17,6 +17,12 @@ const ConsoleMethodsToProxy: Array<string> = [
   'assert',
 ]
 
+const SuppressedReactMessages = [
```
balazsbajorics, line 20: we were printing so much of these React warnings, it made it impossible for me to debug jest tests locally. I wrote a simple filtering for them in console-proxy.

editor/src/components/canvas/controls/multiselect-resize-control.tsx
```diff
@@ -173,7 +173,7 @@ export class MultiselectResizeControl extends React.Component<
               testID={'component-resize-control-0'}
               maybeClearHighlightsOnHoverEnd={this.props.maybeClearHighlightsOnHoverEnd}
             />
-            {...guidelineElements}
+            {guidelineElements}
```
balazsbajorics, line 176: apparently babel is doing transpile-time checks (almost like a soft compiler) and apparently spreading here is not legal JSX ¯\_(ツ)_/¯

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper-third-party.spec.browser.tsx
```diff
@@ -9,10 +9,10 @@ jest.mock('../../../core/es-modules/package-manager/built-in-dependencies-list',
     {
       moduleName: '@react-three/fiber',
       nodeModule: {
-        ...ReactThreeFiber,
-        default: ReactThreeFiber,
+        ...MockReactThreeFiber,
```
balazsbajorics, line 12: babel only lets you use things from out of scope which are named mock* (case insensitive)

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser.tsx
```diff
@@ -144,7 +144,6 @@ describe('Spy Wrapper Template Path Tests', () => {
   it('a simple component in a regular scene', async () => {
     const { getEditorState } = await createExampleProject()
 
-    await wait(20000)
```
balazsbajorics, line 147: oops!

editor/src/components/canvas/ui-jsx-canvas.test-utils.tsx
```diff
@@ -155,8 +154,7 @@ export function renderCanvasReturnResultAndError(
   const baseRequireFn = getRequireFn(NO_OP, updatedContents, {}, {}, 'canvas')
   const requireFn: UiJsxCanvasProps['requireFn'] = (importOrigin: string, toImport: string) => {
     switch (toImport) {
-      case 'antd':
-        return ANTD
+      // here we can manually insert extra dependencies, such as antd
```
balazsbajorics, line 157: I left this as a placeholder here, but I removed antd because I deleted the one test that used it

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -4,12 +4,15 @@ try {
     loadStoredState: () => Promise.resolve(null),
     saveStoredState: () => Promise.resolve(),
   }))
+  jest.mock('../code-editor/console-and-errors-pane', () => ({
+    ConsoleAndErrorsPane: () => null,
```
balazsbajorics, line 8: I am mocking out the `ConsoleAndErrorsPane` from all end-to-end tests because it was throwing _100,000 error messages_

editor/src/core/es-modules/package-manager/package-manager.spec.ts
```diff
@@ -196,7 +202,7 @@ describe('ES Dependency Manager — Real-life packages', () => {
     expect(onRemoteModuleDownload).toBeCalledTimes(0)
   })
 
-  it('antd@4.2.5', async (done) => {
+  it('antd@4.2.5', (done) => {
```
balazsbajorics, line 205: in Jest 27, you can't write a test that is `async (done) =>`  you need to pick async or done()

editor/babel.config.js
```diff
@@ -0,0 +1,13 @@
+// We only use this for Jest for now
+
+module.exports = (api) => {
```
balazsbajorics, line 3: I add a babel.config but the tests are not yet using babel-jest. that will come in a later PR

enidemi (approved):

seanparsons (approved):

editor/package.json
```diff
@@ -194,6 +194,10 @@
   },
   "devDependencies": {
     "@babel/cli": "7.8.4",
+    "@babel/core": "7.14.6",
```
Rheeseyb, line 197: Is it worth updating all of the babel dependencies in this PR?

editor/src/bundled-dependencies/babelHelpers.ts
```diff
@@ -2,6 +2,7 @@
 // @ts-nocheck
 /* eslint-disable */
 // Generated by running npx babel-external-helpers > babelHelpers.js, renaming, and then adding the above
+// @ts-ignore
```
Rheeseyb, line 5: Omg 5 lines of "Please just ignore this damn code!"

editor/src/components/canvas/canvas-types.ts
Rheeseyb: There's an explanation of them here: https://www.typescriptlang.org/docs/handbook/enums.html#const-enums
TLDR - they are completely removed during compilation, so their only benefit is to slightly reduce the amount of output code.

editor/src/components/canvas/controls/multiselect-resize-control.tsx
Rheeseyb: Huh interesting!

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
```diff
@@ -1241,110 +1241,6 @@ export var ${BakedInStoryboardVariableName} = (props) => {
     `)
   })
 
-  it('antd DatePicker works', () => {
```
Rheeseyb, line 1244: Does the datepicker still work in app though?

Rheeseyb: Oh this is because you removed the built in dependency from the test. Got it.

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -93,11 +96,7 @@ const FailJestOnCanvasError = () => {
   const stableCallback = React.useCallback((newRuntimeErrors: Array<RuntimeErrorInfo>) => {
     // we have new runtime errors, let's take the tests down
     if (newRuntimeErrors.length > 0) {
-      const errorMessage = `The test run threw the following canvas error(s): ${newRuntimeErrors
-        .map((e) => e.error)
-        .join(', ')}`
-      fail(errorMessage)
-      throw new Error(errorMessage) // just in case (?) the fail() wouldn't be effective
+      expect(newRuntimeErrors[0]?.error ?? null).toEqual(null)
```
Rheeseyb, line 99: I don't understand why this has been changed?

editor/src/core/workers/ts/ts-worker.spec.ts
```diff
@@ -9,16 +9,18 @@ import {
...
 
 describe('Typescript worker builds the project', () => {
   it('initializing a new project', (done) => {
     handleMessage(SampleInitTSWorkerMessage, (msg) => {
-      expect(msg).toMatchSnapshot()
-      done()
+      if (msg.type === 'build') {
+        expect(msg).toMatchSnapshot()
+        done()
+      }
```
Rheeseyb, line 23: Is there now a risk of this never completing?

editor/src/bundled-dependencies/babelHelpers.ts
balazsbajorics: :D one for humans!

editor/package.json
balazsbajorics: maybe in a follow up PR?

editor/src/missing-types/index.d.ts
```diff
@@ -97,3 +97,5 @@ declare module '@svgr/plugin-jsx'
 declare module '@root/encoding/base64'
 
 declare module 'react/jsx-runtime'
+
+declare module 'react-contexify'
```
Rheeseyb, line 101: Have we somehow lost the types for this?

editor/src/components/canvas/ui-jsx.test-utils.tsx
balazsbajorics: riiight. this is jest 27 related, not tsconfig :/

editor/src/core/workers/ts/ts-worker.spec.ts
balazsbajorics: nope, Jest has a global timeout on tests (currently set to 10000 ms)

editor/src/missing-types/index.d.ts
balazsbajorics: let me double check.

Rheeseyb (approved):

editor/src/components/canvas/ui-jsx.test-utils.tsx
Rheeseyb: What I mean is that previously this would explicitly fail the test with the error message, whereas now you're checking if the error is null

balazsbajorics: I don't remember quite well, but I _think_ this printed a way nicer and more concise output, something about failing with `fail` was less nice, but my memory is fuzzy about it

PR #1461, opened 2021-07-05
Some module import fixes

Rheeseyb:
**Problem:**
We had a few issues in the module importing that were exposed when trying to import [`twind`](https://github.com/tw-in-js/twind), namely:

1. Execution can fail on a `.cjs` file as it may include reference to variables provided by the node.js runtime (e.g. `__filename`) - I will create a follow up ticket to tackle injecting all of the node.js variables
2. The module resolution logic when using the `module` field was incorrect as it was doubling up the folder path
3. Importing whilst resolving these issues was causing the error parsing to blow out as the attached source map was *huge*

**Fix:**

1. Prevent resolving of `.cjs` files (but support resolving of `.mjs` files)
2. Fixed module resolution when a `module` or `name` value is present in the `package.json`
3. Constrain error stack parsing if the size of the stack entry is too huge

### DIFF
--- a/editor/src/core/es-modules/package-manager/module-resolution.ts
+++ b/editor/src/core/es-modules/package-manager/module-resolution.ts
@@ -221,7 +221,7 @@ function processPackageJson(
   return foldEither(
     (_) => resolveNotPresent,
     (packageJson) => {
-      const moduleName: string | null = packageJson.name ?? last(containerFolder) ?? null
+      const moduleName: string | null = packageJson.name ?? null
       const mainEntry: string | null = packageJson.main ?? null
       const moduleEntry: string | null = packageJson.module ?? null
       if (mainEntry != null && !mainEntry.endsWith('.cjs')) {
@@ -232,6 +232,8 @@ function processPackageJson(
       }
       if (moduleName != null) {
         return resolveSuccess(normalizePath([...containerFolder, ...getPartsFromPath(moduleName)]))
+      } else if (containerFolder.length > 0) {
+        return resolveSuccess(normalizePath(containerFolder))
       }
       return resolveNotPresent
     },
--- a/editor/src/core/es-modules/package-manager/module-resolution.ts
+++ b/editor/src/core/es-modules/package-manager/module-resolution.ts
@@ -221,14 +221,17 @@ function processPackageJson(
   return foldEither(
     (_) => resolveNotPresent,
     (packageJson) => {
-      const moduleName: string = packageJson.name ?? makePathFromParts(containerFolder)
+      const moduleName: string | null = packageJson.name ?? last(containerFolder) ?? null
       const mainEntry: string | null = packageJson.main ?? null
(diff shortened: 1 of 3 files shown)


editor/src/core/es-modules/package-manager/module-resolution.ts
```diff
@@ -221,15 +221,18 @@ function processPackageJson(
...
-        return resolveSuccess(normalizePath([...containerFolder, ...getPartsFromPath(moduleName)]))
-      }
-      if (mainEntry != null) {
+      if (mainEntry != null && !mainEntry.endsWith('.cjs')) {
         return resolveSuccess(normalizePath([...containerFolder, ...getPartsFromPath(mainEntry)]))
       }
+      if (moduleEntry != null && !moduleEntry.endsWith('.cjs')) {
+        return resolveSuccess(normalizePath([...containerFolder, ...getPartsFromPath(moduleEntry)]))
+      }
+      if (moduleName != null) {
```
balazsbajorics, line 233: this is surprising here, are we saying we resolve modules that only have a name field? is that according to the spec?

Rheeseyb: Module name here is falling back to the directory, but this is actually missing a follow up commit that makes that clearer - will push that when I'm back at my keyboard

seanparsons (approved):

balazsbajorics (changes requested):
I think there's a mistake in the assumption that `.cjs` files are special in NodeJS and they need a special module context. I believe .js _and_ .cjs files are both treated as RequireJS files in Node, and they need the same context.

With the case of `twind`, we just hit our first module that actually uses __filepath for anything.

I think even if we merge this PR, we need to make sure we will revert this and put a proper fix in place. It would make me easier if we'd special case it for twind, actually. Why not move it in as a built-in dependency until we fix the missing __filepath?

second best option is to _temporarily_ prioritize packageJson.module over packageJson.main until we place filePath and co in the module scope.

editor/src/core/es-modules/package-manager/module-resolution.ts
Rheeseyb: Fixed in the latest commit

Rheeseyb:
> I think there's a mistake in the assumption that `.cjs` files are special in NodeJS and they need a special module context. I believe .js _and_ .cjs files are both treated as RequireJS files in Node, and they need the same context.
> 
> With the case of `twind`, we just hit our first module that actually uses __filepath for anything.
> 
> I think even if we merge this PR, we need to make sure we will revert this and put a proper fix in place. It would make me easier if we'd special case it for twind, actually. Why not move it in as a built-in dependency until we fix the missing __filepath?
> 
> second best option is to _temporarily_ prioritize packageJson.module over packageJson.main until we place filePath and co in the module scope.

I think I'm going to sit on this PR for the moment and investigate further this evening if I get a chance.

Rheeseyb:
Tagged and closing

PR #1465, opened 2021-07-06
Chore/enter escape blur formula bar

### LINKED ISSUE #1464: Make the formula bar input behave like a normal inspector input w/r/t updating and focus surrendering
**Problem**
The formula bar today uses a standard html `input` element with some styles. As a result, it "feels" different from the rest of the inspector inputs.

**Solution**
Either we can use a standard inspector input, or make it participate in the following interaction patterns:
- [x] Enter confirms and `blur`s 
- [ ] Escape abandons* and `blur`s

The main goal here is that hitting CMD+Enter (or enter) on an element containing text or an expression can focus (the children 'tab' of) the formula bar, and "finishing" (Enter or Escape) puts the perceived focus back on the canvas.

* Separately, we can revisit whether we like the real-time update when typing; in the code editor it's the pattern we use, in the inspector we require an `enter` to confirm, but until we nail the focus mgmt here I'd not make changes either way.

balazsbajorics:
Fixes #1464

**Problem:**
The FormulaBar used a simple input which meant that escape / enter would not blur focus.

**Fix:**
I took the non-styled parts of `StringInput` and created `HeadlessStringInput` which only contains the common blur logic we want as standard.

**Commit Details:**
- New `HeadlessStringInput` that are also used by `StringInput`
- Using `HeadlessStringInput` in FormulaBar
- Listening to `onBlur` in FormulaBar

**Note:**
In the PR @maltenuhn mentions escape abandons: currently the none of the input fields in the inspector abandon, they behave like Enter. I decided to not to fix that in this PR, instead I am bringing Current Inspector Behavior to the FormulaBar. We can fix them together later (but I suspect the fix is going to be a bit involved because the inputs don't store a "value when editing started" to which they could easily abandon to)


editor/src/uuiui/inputs/base-input.tsx
```diff
@@ -112,35 +113,45 @@ interface InspectorInputProps extends React.InputHTMLAttributes<HTMLInputElement
...
-    },
-    '&:focus': {},
-  }),
-)
+export const InspectorInputEmotionStyle = ({
+  chained = 'not-chained',
+  controlStyles,
+  hasLabel,
+  roundCorners = 'all',
+}: {
```
seanparsons, line 127: Could we extract this type out so that it makes the top of this function a little less busy please?

editor/src/uuiui/inputs/string-input.tsx
```diff
@@ -187,3 +144,54 @@ const LabelBelow = styled.label({
...
+
+    const handleOnKeyDown = React.useCallback(
+      (e: React.KeyboardEvent<HTMLInputElement>) => {
+        if (onKeyDown != null) {
+          onKeyDown(e)
+        }
+        if (e.key === 'Escape' || e.key === 'Enter') {
+          e.preventDefault()
+          e.stopPropagation()
+          if (typeof ref !== 'function' && ref.current != null) {
```
seanparsons, line 164: Can `ref` ever be a function? The type of `useRef` implies not.

seanparsons (approved):

editor/src/uuiui/inputs/string-input.tsx
balazsbajorics: ahh, I just moved this code but that indeed looks wrong

balazsbajorics: fixed!

editor/src/uuiui/inputs/base-input.tsx
balazsbajorics: fixed!

Rheeseyb (approved):

PR #1466, opened 2021-07-06
Feature/tailwind support

### LINKED ISSUE #1458: Basic Tailwind support via Twind
In order to support Tailwindcss (including supporting basic customisation), we can use Twind. To do this we need to do the following:

- [ ] Detect if user has a `tailwind.config.js`
- [ ] Use twind to create a custom sheet and observer to generate base classes (without the above)
- [ ] Use the detected config in the twind instance
- [ ] Prefix generated rules with a selector restricting them to the canvas
- [ ] As above for the preview

seanparsons:
Fixes #1458

**Problem:**
Users would like to be able to use Tailwind in projects, which currently they cannot do because Tailwind can often involve a command line process being run against a project.

**Fix:**
Primarily we use Twind to incorporate the configured style parts into the canvas, as unlike Tailwind itself it can be run in the browser. Once appropriate Tailwind and Postcss configuration is detected then it becomes automatically incorporated into the canvas and preview.

**Commit Details:**
- `useTwind` hook implemented to either remove or add in the processed style element containing the configured Tailwind content. Uses the regular Tailwind configuration file format and filename.
- Updated TypeScript to 4.1.3.
- Added in support for `module.exports` in project files, similar to what is done for dependencies by including the same variables into the scope. Also extracts the contents of `module.exports` when producing the output of a module.
- Trigger Twind when tailwind is configured in a postcss config.
- Ensure that the required dependencies for tailwind are included in the dependencies of the project.
- Support disabling the preflight of Tailwind.
- Invokes `injectTwind` for the preview.
- Better handle the selectors produced by Twind to ensure that they are encapsulated inside the canvas, having `html` and `body` target the root of the canvas.
- Move `all: initial` from the canvas container to an enclosing div.
- Memoized rule modifications to keep things smooth in the canvas.


editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -431,7 +436,11 @@ export const UiJsxCanvas = betterReactMemo(
       })
 
       return (
-        <>
+        <div
+          style={{
+            all: 'initial',
+          }}
+        >
```
Rheeseyb, line 443: This change was deliberate, and is the cause of the snapshot changes. Tailwind provides a ["preflight"](https://tailwindcss.com/docs/preflight) which adds CSS rules for the root of the doc. We obviously don't want that to affect the editor, so we scope all of the rules so that they are applied to descendants of the `canvas-container` only. That means we have to update those root rules to apply directly to the `canvas-container` element, which this `all: 'initial'` was interfering with. By moving the `all: 'initial'` up a level, we are still able to reset the styling on the canvas in a way that doesn't interfere with the tailwind preflight.

editor/package.json
```diff
@@ -183,7 +183,8 @@
     "string-hash": "1.1.3",
     "strip-ansi": "6.0.0",
     "tippy.js": "6.2.6",
-    "typescript": "4.0.5",
+    "twind": "0.16.16",
+    "typescript": "4.1.3",
```
Rheeseyb, line 187: Twind required a slightly updated version of TS, which is the cause of some unrelated changes elsewhere. There is a much newer version, but we didn't want to introduce too much noise into this PR, so that is something we should tackle separately.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx
```diff
@@ -84,7 +84,24 @@ export function createExecutionScope(
...
+    },
+  }
+
+  let executionScope: MapLike<any> = {
+    require: userRequireFn,
+    module: module,
+    exports: module.exports,
+    process: process,
+    ...requireResult,
+  }
```
Rheeseyb, line 104: As mentioned in the comment, this is to mirror the behaviour we use when evaluating npm dependencies in [`evaluator.ts`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/es-modules/evaluator/evaluator.ts#L72), as without this we weren't able to use the [sample config](https://tailwindcss.com/docs/configuration) from tailwind's docs

editor/src/components/custom-code/code-file.spec.ts
```diff
@@ -333,11 +333,13 @@ describe('transpileCode', () => {
 
       var icon_css_1 = __importDefault(require(\\"./icon.css\\"));
 
-      exports.App = function (props) {
+      var App = function App(props) {
```
Rheeseyb, line 336: This is fallout from the TS update

editor/src/components/editor/npm-dependency/npm-dependency.ts
```diff
@@ -324,6 +324,26 @@ export function dependenciesFromPackageJson(
...
+  dependencyToCheck: string,
+): boolean {
+  if (packageJsonFile != null) {
+    if (isTextFile(packageJsonFile)) {
+      const parsedJSON = json5.parse(packageJsonFile.fileContents.code)
+      const fromDependencies = Utils.path<unknown>(['dependencies', dependencyToCheck], parsedJSON)
+      const fromDevDependencies = Utils.path<unknown>(
+        ['devDependencies', dependencyToCheck],
+        parsedJSON,
+      )
```
Rheeseyb, line 338: I've created the follow up ticket #1468 for this, because we need further support for `devDependencies`

editor/src/core/tailwind/tailwind.ts
```diff
@@ -0,0 +1,243 @@
...
+        const lowerCaseSelector = s.toLowerCase().trim()
+
+        if (
+          lowerCaseSelector === ':root' ||
+          lowerCaseSelector === 'html' ||
+          lowerCaseSelector === 'head'
+        ) {
+          return prefixSelector
+        } else if (lowerCaseSelector === 'body') {
+          return `${prefixSelector} > *`
```
Rheeseyb, line 165: This is required because of the subtle difference between using the `html` vs the `body` selector. Tailwind uses both in the preflight, in one case attaching a specific `font-family` rule to the `html` selector, and then attaching `font-family: inherit` to the `body` selector.

editor/src/core/tailwind/tailwind.ts
```diff
@@ -0,0 +1,243 @@
...
+  React.useMemo(() => {
+    if (shouldUseTwind) {
+      updateTwind(tailwindConfig, prefixSelector)
+    } else {
+      clearTwind()
+    }
+  }, [prefixSelector, shouldUseTwind, tailwindConfig])
+}
+
+export function injectTwind(
```
Rheeseyb, line 223: This is just a non-hooks version of the above for use in the preview because the preview render function (which needs to call this) isn't a React component. It's a bit rubbish having to effectively duplicate the logic like this :/

editor/webpack.config.js
```diff
@@ -281,7 +281,7 @@ const config = {
         contentBase: path.join(__dirname, 'resources'),
         watchContentBase: true, // Watch the above folder for changes too
         overlay: {
-          warnings: true,
+          warnings: false,
```
Rheeseyb, line 284: This is because of the TS update. For some reason (a few hours were lost trying to figure this one out), TS 4.1+ results in the following warning, which we weren't able rectify: 
```
WARNING in ./node_modules/typescript/lib/typescript.js
Module not found: Error: Can't resolve 'perf_hooks' in '/home/rheese/Workspace/Utopia/utopia/editor/node_modules/typescript/lib'
 @ ./node_modules/typescript/lib/typescript.js
 @ ./src/core/workers/ts/ts-worker.ts
 @ ./src/core/workers/ts/ts.worker.ts
 ```
 
This appears to be related to [this change](https://github.com/microsoft/TypeScript/pull/40593), and maybe needs further investigation if it can't be fixed by replacing the ts-loader with babel

File: editor/src/core/tailwind/tailwind.ts [hunk 9]
```diff
@@ -0,0 +1,243 @@
...
+
+  return true
+}
+
+function convertTailwindToTwindConfig(tailwindConfig: any): Configuration {
+  const preflightEnabled = enablesPreflight(tailwindConfig)
+
+  return {
+    ...tailwindConfig,
+    preflight: preflightEnabled,
```
enidemi, line 88: what is the difference between preflight and non-preflight for the canvas?

enidemi (approved):

editor/src/core/tailwind/tailwind.ts
Rheeseyb: This is just determining whether or not the [preflight](https://tailwindcss.com/docs/preflight) should be included. It doesn't mess up the styling of the canvas.

balazsbajorics (approved):

PR #1470, opened 2021-07-07
Include the devDependencies from the package.json when building projects

### LINKED ISSUE #1468: Treat devDependencies as regular dependencies
Utopia only supports regular `dependencies` from a user's package.json. It should also include imported `devDependencies`, as e.g. `tailwindcss` will typically be included in the latter.

Rheeseyb:
Fixes #1468 

**Problem:**
Currently Utopia will only include regular `dependencies` when running a project. Since this is a development environment, it should also include the `devDependencies`.

**Fix:**
Update the internal package parsing to include both fields, keeping them separate for the sake of the caching logic. The functions that expose the results to the rest of the editor then mostly return the combined list of packages, with the exception being around `dependency-list.tsx` since that only focuses on the regular (i.e. non-dev) dependencies.

### DIFF
--- a/editor/src/components/editor/npm-dependency/npm-dependency.ts
+++ b/editor/src/components/editor/npm-dependency/npm-dependency.ts
@@ -372,11 +372,11 @@ function maybeCachedDependenciesFromPackageJson(
 
 export function dependenciesFromPackageJson(
   packageJsonFile: ProjectFile | null,
-  includeDevDependencies: boolean,
+  combinedOrRegularOnly: 'combined' | 'regular-only',
 ): Array<RequestedNpmDependency> {
   const npmDependencies = maybeCachedDependenciesFromPackageJson(packageJsonFile)
 
-  if (includeDevDependencies) {
+  if (combinedOrRegularOnly === 'combined') {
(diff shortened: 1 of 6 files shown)


editor/src/components/editor/npm-dependency/npm-dependency.ts
```diff
@@ -324,19 +370,32 @@ export function dependenciesFromPackageJson(
   }
 }
 
+export function dependenciesFromPackageJson(
+  packageJsonFile: ProjectFile | null,
+  includeDevDependencies: boolean,
```
balazsbajorics, line 375: in my fight against boolean flags, can we make this something like:

```suggestion
  includeDevDependencies: 'dependencies' | 'combined',
```

or

```suggestion
  includeDevDependencies: 'productionDependencies' | 'combined',
```

Rheeseyb: Yep good shout.

Rheeseyb: Done

PR #1471, opened 2021-07-07
Fixed parsing of postcss and tailwind config when using module.exports

Rheeseyb:
**Problem:**
When using `module.exports =` vs `export default`, the imported object will differ between having everything at the top level of the object (in the former case), or having a `default` key containing everything (in the latter case).

**Fix:**
Try both to see which contains the part we're interested in

### DIFF
--- a/editor/src/core/tailwind/tailwind.ts
+++ b/editor/src/core/tailwind/tailwind.ts
@@ -35,9 +35,8 @@ function postCSSIncludesTailwindPlugin(postCSSFile: ProjectFile, requireFn: Requ
   if (isTextFile(postCSSFile)) {
     try {
       const requireResult = requireFn('/', PostCSSPath)
-      if (requireResult?.default != null) {
-        return requireResult?.default?.plugins?.tailwindcss != null
-      }
+      const plugins = requireResult?.plugins ?? requireResult?.default?.plugins
+      return plugins?.tailwindcss != null
     } catch (e) {
       /* Do nothing */
     }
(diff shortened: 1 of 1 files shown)


editor/src/core/tailwind/tailwind.ts
```diff
@@ -96,8 +95,9 @@ function getTailwindConfig(
   if (tailwindFile != null && isTextFile(tailwindFile)) {
     try {
       const requireResult = requireFn('/', TailwindConfigPath)
-      if (requireResult?.default != null) {
-        const twindConfig = convertTailwindToTwindConfig(requireResult.default)
+      const rawConfig = requireResult?.default ?? requireResult
```
balazsbajorics, line 98: here you prioritize default over "wildcard", in line 38 it's the other way around.

for reference, this is the babel-emitted function:
```
function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { 'default': obj }; }
```

I think we should actually put this function in a Util file, and use it!

Rheeseyb: Oh wait, I totally forgot, we actually already have this! Will update and use that

balazsbajorics: oh we do? ha!

Rheeseyb: Yeah, we had to use something like that as part of the module loading. This has been fixed and is now using this https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/es-modules/commonjs-interop.ts#L28

PR #1473, opened 2021-07-08
Feature/topmenu select menu css

enidemi:
**Problem:**
The new topmenu select looks different from the rest of the editor.

**Fix:**
Using colourStyles and custom components in topmenu.

**Commit Details:**
- fixing a bug where selecting doesn't update the classname 
- added new components and styles

Screenshot when typing w-2:
Screenshot 2021-07-08 at 11 06 21


editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -2,26 +2,116 @@
...
+const DropdownIndicator = (props: IndicatorProps<TailWindOption, true>) => (
+  <components.DropdownIndicator {...props}>
+    <span> ↓ </span>
+  </components.DropdownIndicator>
+)
+
+const ClearIndicator = () => null
+const IndicatorSeparator = () => null
+
+const NoOptionsMessage = (props: any) => <span {...props}>No other classes available</span>
```
Rheeseyb, line 48: Is this what we display when there are no results? If so, I think this should read "No results found"

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -2,26 +2,116 @@
...
+})
+
+const getColorForCategory = (category: string) => {
+  if (category === 'layout') {
+    return '#5FACFF'
+  } else if (category === 'typography') {
+    return '#F7B500'
+  } else if (category === 'interaction') {
+    return '#B620E0'
+  } else return 'pink'
```
Rheeseyb, line 64: Are these standard values coming from somewhere else? Should these be part of our theme? (Should the styling of this as a whole be part of our theme)?

I'm not suggesting to change this now either way - more for a follow up ticket

editor/src/components/canvas/controls/classname-select.tsx
enidemi: fixed

editor/src/components/canvas/controls/classname-select.tsx
enidemi: yes this should be using the theme, there is another css todo to make the omnibox use inverted theme, it should be fixed with that

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -2,26 +2,116 @@
...
+  value: string
+  categories?: string[]
+}
+
 const TailWindOptions = TailWindList.classNames.map((className, index) => ({
   label: className,
   value: className,
 }))
 
+const DropdownIndicator = (props: IndicatorProps<TailWindOption, true>) => (
```
Rheeseyb, line 39: Please can you use `betterReactMemo` for all of the function components in here

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: :+1:

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -66,12 +157,160 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
           value: name,
         }))
 
+  const colourStyles: StylesConfig = {
```
Rheeseyb, line 160: Is this something that should be wrapped in a `React.useMemo`?

balazsbajorics: better yet, could this be moved to the file root?

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -66,12 +157,160 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
...
+        borderRadius: 3,
+        position: 'relative',
+        flexGrow: 1,
+        display: 'flex',
+        alignItems: 'center',
+        '&:focus-within': { boxShadow: `0px 0px 0px 1px ${theme.primary.value}` },
+      }}
+    >
+      <WindowedSelect
+        filterOption={createFilter({ ignoreAccents: false })}
```
Rheeseyb, line 298: The `createFilter` call should probably be wrapped in a `React.useMemo` (or `React.useCallback`?)

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: It uses `theme` so it would have to become a function that is wrapped in a `useMemo` still.

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -2,26 +2,116 @@
...
+    return '#5FACFF'
+  } else if (category === 'typography') {
+    return '#F7B500'
+  } else if (category === 'interaction') {
+    return '#B620E0'
+  } else return 'pink'
+}
+
+const MultiValueContainer = (props: MultiValueProps<TailWindOption>) => {
+  const getStriped = (data: TailWindOption) => {
```
balazsbajorics, line 68: should this be `getStripe`? or is my hunglish the problem here :)

enidemi: I don't know 😅 I used this part directly from the prototype

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -66,12 +157,160 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
...
+      ...styles,
+      height: 20,
+    }),
+    option: (styles: React.CSSProperties, { data, isDisabled, isFocused, isSelected }) => {
+      // a single entry in the options list
+      const categories = data?.categories ?? []
+
+      let optionColor = chroma('black')
+      if (categories.length === 1) {
+        optionColor = chroma(theme.primary.value)
```
balazsbajorics, line 240: how fast is this chroma business? sounds like we could at least put this behind a useMemo!

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: Yeah `getStripes` makes more sense. Also, this too should be wrapped in a `React.useMemo`

editor/src/components/canvas/controls/classname-select.tsx
balazsbajorics: all we need from out of this are the `optionColor` and the value of `chroma.contrast(optionColor, 'white') > 2`

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -66,12 +172,168 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
...
+          },
+        }
+      },
+    }),
+    [theme, ChromaThemePrimary],
+  )
+
+  const filterOption = React.useCallback((option: Option, rawInput: string) => {
+    return createFilter({ ignoreAccents: false })(option, rawInput)
+  }, [])
```
Rheeseyb, line 305: Ahhh I misunderstood what was happening here. In this case, the best option is to have a `const filterOption = createFilter({ ignoreAccents: false })` at the root level of the file

Rheeseyb (approved):

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -2,26 +2,135 @@
...
+        return categories.map((category, index) => (
+          <AngledStripe
+            key={data?.label ?? index}
+            style={{ backgroundColor: getColorForCategory(category) }}
+          />
+        ))
+      } else {
+        return []
+      }
+    }, [data])
```
balazsbajorics, line 91: is `data` here referentially stable? or would it be better to feed it `label` and `categories`? (genuine question, I can't tell from just reading the code)

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -2,26 +2,135 @@
...
+        <components.ValueContainer {...props} />
+      </div>
+    )
+  },
+)
+
+const filterOption = (option: Option, rawInput: string) => {
+  return createFilter({ ignoreAccents: false })(option, rawInput)
+}
+
```
balazsbajorics, line 131: I think @Rheeseyb was asking for something like this:

```suggestion
const filterIgnoreAccents = createFilter({ ignoreAccents: false })
const filterOption = (option: Option, rawInput: string) => {
  return filterIgnoreAccents(option, rawInput)
}
```

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -66,12 +176,164 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
...
+        } else if (isDisabled) {
+          backgroundColor = undefined
+        }
+
+        let activeBackgroundColor =
+          !isDisabled && (isSelected ? data.color : optionColor.alpha(0.3).css())
+
+        let color: string = optionColor.css()
+        if (isSelected) {
+          if (chroma.contrast(optionColor, 'white') > 2) {
```
balazsbajorics, line 278: we could also memoize this, i'm not sure how expensive chroma.contrast() is, but I remember that chroma was not _too_ fast when we first tried it...

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -66,12 +176,164 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
           value: name,
         }))
 
+  const ChromaThemePrimary = React.useMemo(() => chroma(theme.primary.value), [theme])
```
balazsbajorics, line 179: a couple of months ago we bumped eslint, and now it understands property access like this:

 
```suggestion
  const ChromaThemePrimary = React.useMemo(() => chroma(theme.primary.value), [theme.primary.value])
```

since we only use that one value, why rely on the referential stability of `theme` itself

enidemi: I can use this for the data too

balazsbajorics: right!

editor/src/components/canvas/controls/classname-select.tsx
enidemi: fixed

balazsbajorics (approved):

Issue #1474, opened 2021-07-08
Include default tailwind classes' attributes in search


Rheeseyb:
For the sake of the class selector autocompletion, we want to also use the actual attributes set (but not their values) whilst filtering search results

PR #1475, opened 2021-07-08
Parsed Tailwind Classes

Rheeseyb:
First step in #1474 

**Problem:**
We want to be able to search tailwind classes based on what attributes they affect.

**Fix:**
I have created a script that will parse the full CSS to pull out the list of class names and the attributes that they affect. I have used the output of that to replace the previous list of class names. Ideally we should set up an automated job to run this periodically, but I think that should be left until the overall work is complete.

**Edit:** This has been updated using `css-tree` as the regex attempt was just hitting constant edge cases.

**Commit Details: (updated)**
- Created `fetch-and-parse-tailwind-classes.ts`, a script for pulling down the minified tailwind CSS file uploaded to the CDN, extract the data we are interested in (a list of classes, a list of all attributes affected, and a mapping from attribute to the classes that affect it), and write the result to `tailwind-default.ts` ~extract an object containing the data we are interested in (a mapping from class name to the list of attributes it affects), and write the resultant JSON object to `tailwind-all-classnames.json`~
- Added an `npm` script for running the above
- Ran the script updating `tailwind-default.ts` ~`tailwind-all-classnames.json`~

Since the updated JSON file is so huge, github just straight up refuses to display it, so here is an updated sample of the result when running against the sample data in the script:
```
import type { MapLike } from 'typescript'

export const AllTailwindClasses: Array<string> = [
  'absolute',
  'container',
  'fixed',
  'gap-y-0.5',
  'gap-y-1.5',
  'gap-y-2.5',
  'gap-y-3.5',
  'invisible',
  'not-sr-only',
  'pointer-events-auto',
  'pointer-events-none',
  'relative',
  'space-x-0',
  'space-x-1',
  'space-x-2',
  'sr-only',
  'static',
  'sticky',
  'visible',
  '32xl:ease-in',
  '32xl:ease-in-out',
  '32xl:ease-linear',
  '32xl:ease-out',
  'focus-within:not-sr-only',
  'focus-within:sr-only',
  'focus:not-sr-only',
  'focus:sr-only',
  'sm:diagonal-fractions',
  'sm:lining-nums',
  'sm:normal-nums',
  'sm:oldstyle-nums',
  'sm:ordinal',
  'sm:proportional-nums',
  'sm:slashed-zero',
  'sm:stacked-fractions',
  'sm:tabular-nums',
]

export const AllAttributes: Array<string> = [
  'border-width',
  'clip',
  'font-variant-numeric',
  'height',
  'margin',
  'margin-left',
  'margin-right',
  'max-width',
  'overflow',
  'padding',
  'pointer-events',
  'position',
  'row-gap',
  'transition-timing-function',
  'visibility',
  'white-space',
  'width',
]

export const AttributeToClassNames: MapLike<Array<string>> = {
  width: ['container', 'not-sr-only', 'sr-only'],
  'max-width': ['container'],
  position: ['absolute', 'fixed', 'not-sr-only', 'relative', 'sr-only', 'static', 'sticky'],
  height: ['not-sr-only', 'sr-only'],
  padding: ['not-sr-only', 'sr-only'],
  margin: ['not-sr-only', 'sr-only'],
  overflow: ['not-sr-only', 'sr-only'],
  clip: ['not-sr-only', 'sr-only'],
  'white-space': ['not-sr-only', 'sr-only'],
  'border-width': ['sr-only'],
  'pointer-events': ['pointer-events-auto', 'pointer-events-none'],
  visibility: ['invisible', 'visible'],
  'row-gap': ['gap-y-0.5', 'gap-y-1.5', 'gap-y-2.5', 'gap-y-3.5'],
  'margin-right': ['space-x-0', 'space-x-1', 'space-x-2'],
  'margin-left': ['space-x-0', 'space-x-1', 'space-x-2'],
  'font-variant-numeric': [],
  'transition-timing-function': [],
}
```


editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -32,7 +32,7 @@ interface TailWindOption {
   categories?: string[]
 }
 
-const TailWindOptions = TailWindList.classNames.map((className, index) => ({
+const TailWindOptions = Object.keys(TailwindClasses).map((className, index) => ({
```
balazsbajorics, line 35: do we know how much does it cost to evaluate this?

editor/src/scripts/fetch-and-parse-tailwind-classes.ts
```diff
@@ -0,0 +1,110 @@
...
+    return response.text()
+  } else {
+    throw new Error(
+      `Failed to fetch latest tailwind css (${response.status}): ${response.statusText}`,
+    )
+  }
+}
+
+const SmolSample = [
+  String.raw`/*! tailwindcss v2.2.4 | MIT License | https://tailwindcss.com *//*! modern-normalize v1.1.0 | MIT License | https://github.com/sindresorhus/modern-normalize */`,
```
balazsbajorics, line 85: oh I didn't know about `String.raw` but this could have been useful when making the projectBundle scripts!

balazsbajorics: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw

If I understand correctly: String.raw parses the string with the escape character behavior (for example `\n` makes a newline) switched off. So
```
String.raw`hi\nthere` 
```
will parse `\` and `n` as two regular characters: `['h', 'i', '\', 'n', 't'...]`

Rheeseyb: Correct. We're using it elsewhere with Regexs, because otherwise there is a load of confusion when trying to use regex escaping, especially when using an external tool (e.g. https://regexr.com/) to craft a regex and then trying to correctly escape that when pasting it into your code.

balazsbajorics (approved):
I'm very curious how many seconds will this add to our load+boot time 😅

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: I shall attempt to profile this.

enidemi (approved):

balazsbajorics (approved):

enidemi (approved):

Issue #1455, opened 2021-07-05
Speedrun "Insert ... " feature


maltenuhn:
**Problem**
Insertion today requires you to draw-to-insert. That's tricky for a bunch of reasons tied to our canvas controls. 

**Solution**
Similar to #1453, expose a menu with the same content, and insert the component. We _might_ want a toggle to add content (so you don't end up inserting 0x0 things)

balazsbajorics:
from Malte: "images should apply a default src IMO", we could pick some small image to be the placeholder

PR #1479, opened 2021-07-09
Insert with defaults action.

seanparsons:
Partly address #1455.

**Problem:**
To support a new menu for directly inserting without select-dragging the dimensions, a new action is needed to be triggered which handles the instant insertion.

**Fix:**
An action `INSERT_WITH_DEFAULTS` implements the desired insertion behaviour, taking a definition of what should be inserted from the editor-wide insertion descriptors. To support the ability to optionally add style sizing, the property controls info has also been extended to provide an indicator as to whether or not that can be added.

**Commit Details:**
- Added `INSERT_WITH_DEFAULTS` action along with action creator, which inserts directly into the hierarchy without any partly inserted state.
- `stylePropOptions` field added to `InsertableComponent`, which flags if the size can be added.
- `getComponentGroups` now ensures the default properties of an element are also included.
- `getComponentGroups` determines that value that should be supplied for `stylePropOptions`.
- Broke apart `defaultPropertiesForComponentInFile` and slightly refactored it to include the parsed property controls in the result so that it can be used in the caller. Also fixed a slight performance issue where an entire object was mapped but a single value was plucked from it, by flipping that around. This appears to have also fixed a bug where the properties weren't coming through for at least the antd components, which has been highlighted by the change in a snapshot test.
- Added a `getDefaultPropsAsAttributes` shared function for the third party components as it's used in both we currently have.
- For the antd and utopia-api components, the properties with default values are added to the props.

### DIFF
--- a/editor/src/components/editor/import-utils.ts
+++ b/editor/src/components/editor/import-utils.ts
@@ -43,36 +43,3 @@ export function importedFromWhere(
   }
   return null
 }
-
-/*
-interface OriginatingNameResult {
-  filePath: string
-  variableName: string
-}
-
-export function findOriginatingName(
-  projectContents: ProjectContentTreeRoot,
-  originFilePath: string,
-  originVariableName: string,
-): OriginatingNameResult | null {
-  const file = getContentsTreeFileFromString(projectContents, originFilePath)
-  if (file == null) {
-    return null
-  } else if (isTextFile(file)) {
-    const fileParseResult = file.fileContents.parsed
-    if (isParseSuccess(fileParseResult)) {
-      const importedFromResult = importedFromWhere(
-        originFilePath,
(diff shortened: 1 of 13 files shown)


editor/src/components/editor/import-utils.ts
```diff
@@ -42,3 +43,36 @@ export function importedFromWhere(
   }
   return null
 }
+
+/*
```
balazsbajorics, line 47: should this be deleted?

balazsbajorics (approved):

enidemi (approved):

editor/src/components/editor/import-utils.ts
seanparsons: It should and now has been. :)

PR #1480, opened 2021-07-10
Chore: Emotion ThemeProvider

balazsbajorics:
**Problem:**
Thanks to @cscheffauer we had a hook to get the colorTheme, but that hook cannot work in our old Class components. 

_While some of the class components could be converted to function components, the truth is that we did convert most of our classes to function components two years ago, and what we have left as class components were the ones where the conversion's risk / benefit ratio would simply be too bad. Some of these would need a complete rewrite to keep their current functionality. Whenever we make a new component, we make it a Function component, but for these old components (which work fine today) I simply see no reason to rewrite them just for the sake of rewriting them, and we would risk introducing subtle regressions all over the oldest parts of our codebase, which is of course also the least tested parts of the codebase._

**Fix:**
Introduce Emotion ThemeProvider, which comes with a HOC named `withTheme()` which can provide the theme for class-based components.

**Commit Details:**
- Put the entire UtopiaTheme in the ThemeProvider
- Use the ThemeProvider at the EditorComponent root
- `useColorTheme` now returns `useTheme().color`
- Function components can still use `useColorTheme()` and they can also use `useTheme()`
- `Tooltip`, `EditorCanvas` now use `withTheme`. They are still class components, but are now themeable!

We have three big type of TODOs to work down:
1. classes that import `colorTheme` directly
2. Function components and class components that import `UtopiaTheme` directly (`UtopiaTheme.color` is a pointer to `colorTheme`)
3. At the bottom of `theme.ts` we have `UtopiaStyles` (which I completely forgot about, and it shouldn't even live in this file) which not only uses `colorTheme` directly, but it even uses the unexported base colors which are not themed... so for these, we'd need to move them to a separate file, and make them use the theme. I updated `UtopiaStyles.canvas` and `UtopiaStyles.popup` as examples of how can we give them a dynamic theme.

### DIFF
--- a/editor/src/components/editor/editor-component.tsx
+++ b/editor/src/components/editor/editor-component.tsx
@@ -1,5 +1,5 @@
 /** @jsx jsx */
-import { jsx } from '@emotion/react'
+import { jsx, ThemeProvider } from '@emotion/react'
 import { ResizeDirection } from 're-resizable'
 import * as React from 'react'
 import { DndProvider } from 'react-dnd'
@@ -42,14 +42,13 @@ import {
 import {
   SimpleFlexRow,
   SimpleFlexColumn,
-  UtopiaTheme,
   FlexRow,
   ResizableFlexColumn,
-  useColorTheme,
   TabComponent,
   LargerIcons,
   Subdued,
   FlexColumn,
+  useGetThemeForEditor,
 } from '../../uuiui'
 import { betterReactMemo } from '../../uuiui-deps'
 import { createIframeUrl } from '../../core/shared/utils'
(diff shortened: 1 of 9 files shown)


enidemi (approved):

seanparsons (approved):

cscheffauer (approved):
looks good to me!

maltenuhn:
@balazsbajorics can this be merged, or is the merge conflict something to decide (eg how to deal w/ "inverted" theme being available)?

balazsbajorics:
> @balazsbajorics can this be merged, or is the merge conflict something to decide (eg how to deal w/ "inverted" theme being available)?

the merge conflict is fine, but the breaking tests revealed an issue (not being able to set default theme, so all tests need a theme provider or we need to fork the theme provider and fix this somehow smartly) that will take a few hours to fix

Rheeseyb:
Tagged and closing

PR #1481, opened 2021-07-10
Pre-filtering of class names in the top menu

Rheeseyb:
**Problem:**
10 bajillion options cause havoc with react-select's filtering.

**Fix:**
Perform the filtering upfront instead, and feed the results to react-select. This was based on [this comment](https://github.com/JedWatson/react-select/issues/3128#issuecomment-812619125) suggesting pre-filtering, but uses a sparse array to allow us to keep the best matches at the front.

### DIFF
--- a/editor/src/components/canvas/controls/classname-select.tsx
+++ b/editor/src/components/canvas/controls/classname-select.tsx
@@ -32,7 +32,7 @@ interface TailWindOption {
   categories?: string[]
 }
 
-const TailWindOptions = AllTailwindClasses.map((className, index) => ({
+const TailWindOptions: Array<TailWindOption> = AllTailwindClasses.map((className, index) => ({
   label: className,
   value: className,
 }))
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -171,11 +171,48 @@ const ValueContainer = betterReactMemo(
...
 
-const filterOption = createFilter({ ignoreAccents: false })
+const filterOption = () => true
+const MaxResults = 500
 
 export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNameSelect', () => {
   const theme = useColorTheme()
   const dispatch = useEditorState((store) => store.dispatch, 'ClassNameSelect dispatch')
+  const [input, setInput] = React.useState('')
+  const filteredOptions = React.useMemo(() => {
```
maltenuhn, line 181: Does this imply that continued typing would lose / reset temporary selection (eg type three letters, use arrow keys to navigate, continue typing)? I'm guessing it would - but if so, does it 'feel wrong'?

balazsbajorics (approved):

enidemi (approved):

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: I've just tested and it doesn't. I tested by typing `text-`, using the arrow keys to highlight `text-blue-300`, then typed the letter 'b', and though the filtered results had updated (and the highlighted result wouldn't have been first), the highlighted result was still retained

Rheeseyb:
This has been updated to trim the input and cast it to a lower case.

enidemi (approved):

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -171,11 +171,49 @@ const ValueContainer = betterReactMemo(
...
+    const trimmedLowerCaseInput = input.trim().toLowerCase()
+    if (trimmedLowerCaseInput === '') {
+      return TailWindOptions.slice(0, MaxResults)
+    } else {
+      // First find all matches, and use a sparse array to keep the best matches at the front
+      let orderedMatchedResults: Array<Array<TailWindOption>> = []
+      let perfectMatchCount = 0
+      for (var i = 0; i < TailWindOptions.length && perfectMatchCount < MaxResults; i++) {
```
seanparsons, line 189: Should this be checking the length of `existingMatched` against `MaxResults`?

seanparsons (approved):

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: Nah, this was deliberate, so that we don't cut the search too early. In this case we only want to end the search early if we've already found the maximum number of perfect matches

PR #1482, opened 2021-07-12
"Wrap in…" menu

### LINKED ISSUE #1453: Speedrun Wrap in ... Feature
**Problem** 
It's hard to quickly compose a UI with Utopia without strong canvas controls. Also, layout components force a different workflow than "draw-to-insert" + inspector edits.

**Solution**
- [ ] Keyboard shortcut
- [ ] Context menu entry
- [ ] Popup with auto-focused input, autocomplete and permanently visible list
- [ ] Escape to exit

Prototype layout here https://utopia.app/p/52915edf-on-canvas-insert%2Fwrap%2Fchange/

shortcuts proposed by Malte: https://discord.com/channels/647102474519576589/861627672513216522/861633514545348618

for rapidfire classnames, maybe cmd + .
(although in design tools that just hides all of the UI, so scratch that) - maybe cmd+j (don't think it's taken in design tools)
for the three speed actions, how about
- insert = f or r (muscle memory). change. change = j. wrap g (analogous to cmd+g to group)

balazsbajorics:
Fixes #1453

**Problem:**
We want a quick way to be able to wrap the selected element (or elements) in a user-specified element. (a good example is layout elements, such as FlexRow and FlexColumn)

**Fix:**
The PR uses the existing "Wrap in View" feature we had. `CMD + G` triggers the existing wrap in view feature. `G` without cmd brings up a new floating popup menu, where you can select what to use for wrapping:
image

**Commit Details:**
- Changed `wrapInView` so it now takes a `whatToWrapWith` property
- Added new FloatingInsertMenu
- moved some things around in the layer above the canvas so that they are not covered by the navigator overlay anymore
- put the canvas overlay layer into a `transform: translateZ(0)` hack, so it doesn't render on the same css layer as the canvas contents
- new shortcut `G` to Wrap in...
- new context menu item for Wrap in...
- removed "Group" and "Ungroup" from context menu


editor/src/components/canvas/ui/floating-insert-menu.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+export const Subdued = styled.div({
+  color: 'hsl(0,0%,70%)',
+})
+
+const showInsertionOptions = false
+
+export var FloatingMenu = () => {
+  const colorTheme = useColorTheme()
+  const [highlightedKey, setHighlightedKey] = React.useState('')
```
seanparsons, line 150:
```suggestion
  const [highlightedKey, setHighlightedKey] = React.useState<string | null>(null)
```

editor/src/components/editor/actions/actions.ts
```diff
@@ -2088,36 +2097,71 @@ export const UPDATE_FNS = {
...
                 null,
               )
 
               viewPath = EP.appendToPath(parentPath, newUID)
 
+              const importsToAdd: Imports =
+                action.whatToWrapWith === 'default-empty-View'
+                  ? {
+                      // the default View import
+                      ['utopia-api']: {
```
seanparsons, line 2152: Could use `importDetails` here if you wanted to.

editor/src/components/canvas/ui/floating-insert-menu.tsx
```diff
@@ -0,0 +1,335 @@
...
+            : colorTheme.contextMenuForeground.value,
+        }}
+        {...props}
+      />
+    )
+  },
+)
+
+export const Subdued = styled.div({
+  color: 'hsl(0,0%,70%)',
```
enidemi, line 143: please use the theme for colors, backgroundcolors and borders

seanparsons (approved):

enidemi (approved):

editor/src/components/canvas/ui/floating-insert-menu.tsx
balazsbajorics: fixed

editor/src/components/editor/actions/actions.ts
balazsbajorics: fixed

balazsbajorics:
I'm merging this so I can open a new PR that uses react-select

PR #1487, opened 2021-07-12
Support searching class names via the attributes they affect

Rheeseyb:
Continued work towards #1474 

**Problem:**
Users won't necessarily know the tailwind classes they are after, so we want to support searching their affected attributes too.

**Fix:**
After searching class names, if we still haven't reached our max results we also search the list of attributes. Then, if any matches are found, we perform the reverse lookup to find the classes that are matched.

Note that the UX for this is very raw. Continued work is in progress on that front. This PR is just for the core work to provide that search and include the results.

### DIFF
--- a/editor/src/components/canvas/controls/classname-select.tsx
+++ b/editor/src/components/canvas/controls/classname-select.tsx
@@ -262,7 +262,7 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
 
       // Next if we haven't hit our max result count, we find matches based on attributes
       const remainingAllowedMatches = MaxResults - matchedResults.length
-      if (remainingAllowedMatches) {
+      if (remainingAllowedMatches > 0) {
         const orderedAttributeMatchedResults = findMatchingOptions(
           trimmedLowerCaseInput,
           AllAttributes,
--- a/editor/src/components/canvas/controls/classname-select.tsx
+++ b/editor/src/components/canvas/controls/classname-select.tsx
@@ -4,16 +4,18 @@ import React from 'react'
 import { jsx } from '@emotion/react'
 import styled from '@emotion/styled'
 
-import { AllTailwindClasses } from '../../../core/third-party/tailwind-defaults'
+import {
+  AllTailwindClasses,
+  AllAttributes,
+  AttributeToClassNames,
+} from '../../../core/third-party/tailwind-defaults'
 import WindowedSelect, {
   components,
-  createFilter,
   IndicatorProps,
   MultiValueProps,
   ValueContainerProps,
 } from 'react-windowed-select'
 import type { StylesConfig } from 'react-select'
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -196,31 +250,37 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
...
-        const nextMatches = orderedMatchedResults[j]
-        if (nextMatches != null) {
-          matchedResults.push(...nextMatches.slice(0, MaxResults - matchCount))
-          matchCount += nextMatches.length
-        }
+      let matchedResults = takeBestOptions(orderedMatchedResults, MaxResults)
+
+      // Next if we haven't hit our max result count, we find matches based on attributes
+      const remainingAllowedMatches = MaxResults - matchedResults.length
+      if (remainingAllowedMatches) {
```
balazsbajorics, line 265:
```suggestion
      if (remainingAllowedMatches > 0) {
```

balazsbajorics (approved):

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: Oh my word

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -196,31 +250,37 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
...
+        const bestMatchedAttributes = takeBestOptions(
+          orderedAttributeMatchedResults,
+          remainingAllowedMatches,
+        )
+        const classNamesFromBestMatchedAttributes = mapToClassNames(
+          bestMatchedAttributes,
+          remainingAllowedMatches,
+        )
+        const optionsForClassNames = classNamesFromBestMatchedAttributes.map((className) =>
+          TailWindOptions.find((option) => option.label === className),
```
seanparsons, line 281: For each entry in `classNamesFromBestMatchedAttributes` this will (potentially) look through a ~40,000 entry array, that doesn't sound like it will be fast to me.

seanparsons (approved):

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: Having tested this out, it doesn't noticeably impact the performance there, though I agree it does feel like there might be some cases that would.

I'm actually wondering how changes to the model will affect this though since we also want to show all attributes affected by a class in the menu, which means an upfront change to the options, which would in turn probably alter the searching.

I think I'm going to merge this as is so that it doesn't become stale, and then probably end up re-writing a lot of it in the very near term future anyway.

enidemi (approved):

PR #1492, opened 2021-07-13
Capture the focused option in the classname-select

Rheeseyb:
**Problem:**
We want to be able to act based on the focused option in the classname-select without the user having to select it (e.g. to display data about it or preview the result of selecting it)

**Fix:**
Use [this cheeky hack](https://github.com/JedWatson/react-select/issues/4496) to expose it via the aria live messages, storing the result in a `useState` and clearing it when there are no filter results or the menu is closed. Then add a context to support providing that information to upcoming custom components.

This PR is literally just for capturing that data and adding it to the context, not for doing anything with it. This is to unblock other work requiring it.

### DIFF
--- a/editor/src/components/canvas/controls/classname-select.tsx
+++ b/editor/src/components/canvas/controls/classname-select.tsx
@@ -7,13 +7,11 @@ import styled from '@emotion/styled'
 import { AllTailwindClasses } from '../../../core/third-party/tailwind-defaults'
 import WindowedSelect, {
   components,
-  createFilter,
   IndicatorProps,
   MultiValueProps,
   ValueContainerProps,
 } from 'react-windowed-select'
 import type { StylesConfig } from 'react-select'
-import type { Option } from 'react-select/src/filters'
 import chroma from 'chroma-js'
 
 import * as EditorActions from '../../editor/actions/action-creators'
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -428,25 +460,33 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
...
-        components={{
-          DropdownIndicator,
-          ClearIndicator,
-          IndicatorSeparator,
-          NoOptionsMessage,
-          MultiValueContainer,
-          ValueContainer,
+      <ClassNameMenuContext.Provider
+        value={{
+          focused: focusedOption,
```
balazsbajorics, line 465: I'm not sure how much the context menu items rerender on hover change to begin with. but anyone listening to this context will rerender whenever ClassNameSelect rerenders: in the new-ish React APIs, like useStatea and useContext the value is compared using `===` strict reference equality for contexts. 
The simplest workaround for now is that the type of value should not be `{focused: TailWindOption}`, instead just `TailWindOption`

balazsbajorics (approved):

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: Testing now with a component that consumes the context

Rheeseyb: Just verified and it doesn't cause unnecessary re-renders as is. I believe because of our patching of the `useContextSelector`.

Rheeseyb:
I've discovered that this causes some very unexpected scrolling behaviour when the mouse is over one of the rows. Trying to debug now.

Rheeseyb:
I'm going to close this PR as this is now being rolled into other work

PR #1497, opened 2021-07-13
Show the affected attributes in the classname select

Rheeseyb:
**Problem:**
We want to show the attributes affected by the currently hovered option in the classname select

**Fix:**
We use [this cheeky hack](https://github.com/JedWatson/react-select/issues/4496) to expose the currently focused option via the aria live messages, and then use the [atom-pub-sub](https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/atom-with-pub-sub.ts) to store that data in an atom that is accessible elsewhere.

To then show that data, we use a custom `Menu` which subscribes to that atom, and uses it to pull out the attributes associated with a given option.

Since we weren't previously storing the attributes in the options, I have made a change to the tailwind parsing script to create two separate mappings (one from classname to attributes, and one from attribute to classnames). Since loading that in sync when evaluating the `classname-select` file might be a bit expensive, I have also used an `async` function to load that data and create the options, as well as another object for quickly finding the options associated with a given attribute.

The result is this:
Screenshot_20210713_163238

**Note:** There is a caveat to this. Some of the classes only exist to set the values of internal variables, which are effectively meaningless without knowing the context of where they are used. So for example, the class `backdrop-blur-none` sets the value of `--tw-backdrop-blur` and nothing else. For those, I have opted to not show anything in the footer, like so:
Screenshot_20210713_163924


editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -491,10 +572,13 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
       }}
     >
       <WindowedSelect
+        menuIsOpen={true}
```
enidemi, line 575: is this added for debugging?

enidemi (approved):

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: Ha oops! Yeah I was using it to fix the styles :grinning: Will remove

Rheeseyb:
To save you looking at the new `tailwind-defaults.ts` file, it looks like this:
```
export const ClassNameToAttributes: MapLike<Array<string>> = {
  absolute: ['position'],
  'align-baseline': ['vertical-align'],
  'align-bottom': ['vertical-align'],
  'align-middle': ['vertical-align'],
  'align-text-bottom': ['vertical-align'],
  'align-text-top': ['vertical-align'],
  ...
}
```

Rather than the previous 
```
export const AllTailwindClasses: Array<string> = [
  'absolute',
  'align-baseline',
  'align-bottom',
  'align-middle',
  'align-text-bottom',
  ...
]
```

The rest is sorted alphabetically for the sake of searches, but otherwise unchanged.

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -37,18 +38,43 @@ import {
...
+  TailWindOptions = mapToArray(
+    (attributes, className) => ({
+      label: className,
+      value: className,
+      attributes: attributes,
+    }),
+    ClassNameToAttributes,
+  )
+
+  AttributeOptionLookup = mapValues((classNames: Array<string>) => {
```
seanparsons, line 69: This looks to be O(n * m * o) which could be pretty slow in the worst case.

seanparsons (approved):

editor/src/components/canvas/controls/classname-select.tsx
Rheeseyb: This only happens once on start up and was supposed to be wrapped in a Promise so that it didn't block first load. I'll fix that part but otherwise I think this is our best option rather than having to perform the lookup on each search

Rheeseyb: Ok I've fixed this now. I also timed this and it is consistently taking ~65ms for me, which I'd say is acceptable since it's a one time thing and happens before the editor is responsive.

PR #1502, opened 2021-07-14
feature(classname-select) focused items are added to the canvas

enidemi:
**Problem:**
It's difficult to figure out what the select menu items are gonna do before selecting them.

**Fix:**
On menu item focus the hovered item is added to the classname list using transient canvas state.
I had to change how the selected value is created for the dropdown because of 2 issues:
- the fallback metadata props (that is used for controlled) are changing while in this transient state, triggering extra renders
- the classname attribute was also read from the jsx element found in instance metadata, which is now the transient jsx element, so I'm digging the projectContents to find the non-transient attributes. The attribute also had problems with memoization, now it's converted to a className string

**Commit Details:**
- actions to set/clear transient properties
- create transient canvas state
- changes about how the select menu values are made


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -2617,3 +2622,61 @@ export function getValidElementPathsFromElement(
...
+      editor,
+    )
+
+    let transientFilesState: TransientFilesState = {}
+    fastForEach(Object.values(editor.canvas.transientProperties) ?? [], (prop) => {
+      forUnderlyingTargetFromEditorState(
+        prop.elementPath,
+        updatedEditor,
+        (success, underlyingElement, underlyingTarget, underlyingFilePath) => {
+          transientFilesState[underlyingFilePath] = {
```
balazsbajorics, line 2668: if I have two `editor.canvas.transientProperties` entries pointing to the same file, will they overwrite each other and the last one win out?

enidemi: nope, the transientproperties are applied to the jsx elements in a reduce in a few lines above this based on elementpath

enidemi: I should add some tests

balazsbajorics: or at least make yourself a TODO for tests :) but maybe not for today, as we are still figuring this feature out

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1603,6 +1604,10 @@ export function produceCanvasTransientState(
     }
   }
 
+  if (editorState.canvas.transientProperties != null && editorState.canvas.dragState == null) {
+    transientState = createCanvasTransientStateFromProperties(editorState)
```
Rheeseyb, line 1608: I think this should be moved to an `else if` clause inside the above block

editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -413,22 +458,37 @@ export const ClassNameSelect: React.FunctionComponent = betterReactMemo('ClassNa
...
         }))
-  }, [classNameAttribute, classNameFromProps])
-
-  const isMenuEnabled = React.useMemo(
-    () =>
-      classNameAttribute == null ||
-      isJSXAttributeValue(classNameAttribute) ||
-      isJSXAttributeNotFound(classNameAttribute),
-    [classNameAttribute],
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Rheeseyb, line 467: This shouldn't be needed?

seanparsons (approved):

Rheeseyb (approved):

PR #1512, opened 2021-07-15
Made the add file and folder buttons consistent.

### LINKED ISSUE #1425: Creating a new file leaves the file name edit field focused
**Describe the bug**
When creating a new file via the file browser, after creating the file, giving it a name, and then hitting enter, a new row appears in the file browser with the file name in a focused and editing state.

**To Reproduce**
1. Open the file browser
2. Click the "Add Code File" button that appears when highlighting the "src" folder
3. Give the new file a name and hit enter

Result: The newly created row is focused and in an editing state

**Expected behaviour**
The newly created row should be selected but not in an editing state

file_insert_name_bug

### LINKED ISSUE #1478: Creating a new file at the root level doesn't open it upon creation
When inserting a new file via the file browser, the expected behaviour is for that new file to be opened in the code editor. However, when creating one at the root level of the project, that behaviour breaks.

seanparsons:
Fixes #1478 
Fixes #1425 

**Problem:**
The add file and folder buttons for the root don't open a file after creating it. In the process of fixing this it was also discovered that the add file and folder buttons also work inconsistently for non-root items.

**Fix:**
Most of the changes were around making the buttons for root items and the non-root items use the same logic. But also making them consistent with each other, for example having the `ADD_FOLDER` action also supply a filename the way the `ADD_TEXT_FILE` action.

**Commit Details:**
- Fixes #1478.
- Refactored the logic for displaying the new name input field
  into a function `addingChildElement`.
- The `FileBrowser` component now includes some callbacks for
  manipulating root element insertion.
- Created an interface `AddingFile` for maintaining the state of
  what is being added.
- `applyAddingFile` is the common function for actioning an instance
  of `AddingFile`.
- Removed `isAddingChildNameValid` as it was completely unused.
- The add folder button for items not at the root level is consistent
  with the add file button, whereby it now doesn't create the folder
  eagerly.
- Stopped setting `renamingTarget` in `ADD_TEXT_FILE` and `ADD_FOLDER`
  as that was causing them to be editable after they had been created.
- Added `when` and `unless` functions which can be used in-place of ternaries
  in JSX.

### DIFF
--- a/editor/src/components/filebrowser/filebrowser.tsx
+++ b/editor/src/components/filebrowser/filebrowser.tsx
@@ -35,7 +35,7 @@ import {
   getOpenUIJSFileKey,
 } from '../editor/store/editor-state'
 import { useEditorState } from '../editor/store/store-hook'
-import { FileBrowserItem } from './fileitem'
+import { addingChildElement, FileBrowserItem } from './fileitem'
 import { dropFileExtension } from '../../core/shared/file-utils'
 import { objectMap } from '../../core/shared/object-utils'
 import { defaultPropertiesForComponentInFile } from '../../core/property-controls/property-controls-utils'
@@ -53,6 +53,8 @@ import {
   Icons,
 } from '../../uuiui'
 import { isLocal } from '../editor/persistence'
+import { unless, when } from '../../utils/react-conditionals'
+import { AddingFile, applyAddingFile } from './filepath-utils'
 
 export type FileBrowserItemType = 'file' | 'export'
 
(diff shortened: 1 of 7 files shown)


editor/src/components/filebrowser/filebrowser.tsx
```diff
@@ -152,17 +154,72 @@ export const FileBrowser = betterReactMemo('FileBrowser', () => {
...
-            <FileBrowserActionSheet visible={!minimised} />
+            <FileBrowserActionSheet
+              visible={!minimised}
+              setAddingFileOrFolder={setAddingFileOrFolder}
+            />
           </FlexRow>
         </SectionTitleRow>
         <SectionBodyArea minimised={minimised}>
-          {minimised ? null : <FileBrowserItems />}
+          {unless(minimised, () => (
```
balazsbajorics, line 209: unless minimised! nice

editor/src/utils/react-conditionals.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (React.isValidElement(element)) {
+    return element
+  } else if (typeof element === 'function') {
+    return element()
+  } else {
+    return element
+  }
+}
+
+export function when(condition: boolean, element: ReactNode | (() => ReactNode)): ReactNode {
```
enidemi, line 14: ✨

balazsbajorics (approved):

enidemi (approved):

Rheeseyb:
> Stopped setting renamingTarget in ADD_TEXT_FILE and ADD_FOLDER
as that was causing them to be editable after they had been created.

Beautiful! That means this also closes another ticket (will update the description to include that)

PR #1513, opened 2021-07-15
Feature/dependencies button to add tailwind

enidemi:
**Fix:**
Added a button to dependencies menu that creates the default tailwind config files and adds the 2 dependencies to the package.json

**Commit Details:**
- new button in dependencies
- action that handles fetching the dependencies, adding the 2 new files and updating the package.json

### DIFF
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -386,6 +386,7 @@ import {
   updateDependenciesInPackageJson,
   createLoadedPackageStatusMapFromDependencies,
   dependenciesFromPackageJson,
+  findLatestVersion,
 } from '../npm-dependency/npm-dependency'
 import { updateRemoteThumbnail } from '../persistence'
 import {
@@ -516,9 +517,7 @@ import {
   DefaultPostCSSConfig,
   DefaultTailwindConfig,
   PostCSSPath,
-  PostCSSVersion,
   TailwindConfigPath,
-  TailwindCSSVersion,
 } from '../../../core/tailwind/tailwind-config'
 
 export function updateSelectedLeftMenuTab(editorState: EditorState, tab: LeftMenuTab): EditorState {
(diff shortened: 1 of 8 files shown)


editor/src/components/editor/actions/actions.ts
```diff
@@ -4531,6 +4544,61 @@ export const UPDATE_FNS = {
...
+    action: AddTailwindConfig,
+    editor: EditorModel,
+    dispatch: EditorDispatch,
+  ): EditorModel => {
+    const packageJsonFile = getContentsTreeFileFromString(editor.projectContents, '/package.json')
+    const currentNpmDeps = dependenciesFromPackageJson(packageJsonFile, 'regular-only')
+    const updatedNpmDeps = [
+      ...currentNpmDeps,
+      requestedNpmDependency('postcss', PostCSSVersion),
+      requestedNpmDependency('tailwindcss', TailwindCSSVersion),
```
Rheeseyb, line 4557: I think we can add these without specifying a version (so that it takes the latest version available)

editor/src/components/editor/actions/actions.ts
```diff
@@ -4531,6 +4544,61 @@ export const UPDATE_FNS = {
...
+      const packageErrorActions = Object.keys(loadedPackagesStatus).map((dependencyName) =>
+        setPackageStatus(dependencyName, loadedPackagesStatus[dependencyName].status),
+      )
+      dispatch([
+        ...packageErrorActions,
+        updateNodeModulesContents(fetchNodeModulesResult.nodeModules, 'full-build'),
+      ])
+    })
+
+    const projectContentsWithTailwind = addFileToProjectContents(
```
Rheeseyb, line 4575: I think for both of these we should maybe check if those files already exist first?

editor/src/components/editor/actions/actions.ts
Rheeseyb: Also, thinking about it, we should probably check that these aren't already in the list of dependencies

editor/src/core/tailwind/tailwind-config.ts
```diff
@@ -0,0 +1,49 @@
...
+  },
+  variants: {},
+  plugins: [],
+}`
+
+const PostCSSConfigJs = `// postcss.config.js
+module.exports = {
+  plugins: {
+    tailwindcss: {},
+    autoprefixer: {},
```
Rheeseyb, line 25: We should either remove `autoprefixer` from this or add it to the list of dependencies too (I think we might have to go with the latter, I haven't checked if the default tailwind set up works without autoprefixer)

enidemi: fixed

editor/src/components/editor/actions/actions.ts
enidemi: fixed

editor/src/components/editor/actions/actions.ts
enidemi: I changed this to load the latest version

editor/src/core/tailwind/tailwind-config.ts
```diff
@@ -0,0 +1,45 @@
+import { RevisionsState, TextFile, textFile, textFileContents } from '../shared/project-file-types'
+import { lintAndParse } from '../workers/parser-printer/parser-printer'
+
+export const PostCSSPath = '/postcss.config.js'
+export const TailwindConfigPath = '/tailwind.config.js'
+
+const TailwindConfigJs = `// tailwind.config.js
```
Rheeseyb, line 7:
```suggestion
const TailwindConfigJs = `
```

editor/src/core/tailwind/tailwind-config.ts
```diff
@@ -0,0 +1,45 @@
...
+  purge: [],
+  darkMode: false, // or 'media' or 'class'
+  theme: {
+    extend: {},
+  },
+  variants: {},
+  plugins: [],
+}`
+
+const PostCSSConfigJs = `// postcss.config.js
```
Rheeseyb, line 18:
```suggestion
const PostCSSConfigJs = `
```

Rheeseyb (approved):
:dash:

seanparsons (approved):

balazsbajorics:
merge merge merge!

PR #1516, opened 2021-07-15
Chore/remove utopions

maltenuhn:
**Problem:**
We were still using utopions in the editor, and also in the loading screen. We also had some other generically named css classes that could clash with user css.

**Fix:**
* Removes utopions.css and converts its usage mostly-blind into `style` objects
* Removes our beautiful load screen and replaces it with a streamlined version
* Removes unused css files, and un-uses a few more
* Improves perceived load performance ever so slightly by putting key editor css and font reqs into the initial-load.css file

**Discussion:**
I realised that the way we're loading our css actually caused some layout thrashing, mainly because fonts arrive v late in the game. We probably should inline into HTML **all** core editor css (layout, positioning etc), better prioritise the external css files, and make sure they are at least pre-fetched while our JS is already running. The changes here seem to address the most egregious points though.


enidemi (approved):

editor/src/components/canvas/controls/margin-controls.tsx
```diff
@@ -21,9 +21,11 @@ export const MarginControls = betterReactMemo('MarginControls', (props: MarginCo
       props.margin.left != null && props.margin.left !== 0 ? (
         <div
           key='marginLeft'
-          className='f10 fw5 tc overflow-visible'
           style={{
             background: colorTheme.canvasLayoutFillTranslucent.value,
+            fontWeight: 500,
```
Rheeseyb, line 26: I believe this is changing the font size since `f10` was setting it to `10px`? Was that change deliberate?

editor/src/templates/index.html
```diff
@@ -53,85 +76,48 @@
...
+            right: 0px;
+            display: flex;
+            flex-direction: column;
+            align-items: center;
+            justify-content: center;
+            border: 1px solid beige;
+          "
+        >
+          <img src='/editor/pyramid_fullsize@2x.jpg?hash=%UTOPIA_SHA%'' height='128px' alt='Utopia
+          Logo '/>
```
Rheeseyb, line 95: Does the `alt` being split over two lines cause any issues? I'm guessing this was caused by prettier doing its thing

Rheeseyb (approved):

editor/src/templates/index.html
maltenuhn: Huh, that does look like prettier. It was rendering ok FWIW

editor/src/components/canvas/controls/margin-controls.tsx
maltenuhn: No, will add it back in. That said, I don't even know if we still _render_ these margin controls...

PR #1523, opened 2021-07-16
Fix Focus Component Issues

balazsbajorics:
**Commit Details:**
- Double click to focus on component works in the non-label parts of the navigator row
- New feature switch called `Click on empty canvas unfocuses`. It's set to true by default to keep current behavior

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item.tsx
@@ -313,6 +313,10 @@ export const NavigatorItem: React.FunctionComponent<NavigatorItemInnerProps> = b
       () => highlightItem(dispatch, elementPath, selected, isHighlighted),
       [dispatch, elementPath, selected, isHighlighted],
     )
+    const focusComponent = React.useCallback(() => {
+      dispatch([EditorActions.setFocusedElement(elementPath)])
+    }, [dispatch, elementPath])
+
     const containerStyle: React.CSSProperties = React.useMemo(() => {
       return {
         opacity: isElementVisible ? undefined : 0.5,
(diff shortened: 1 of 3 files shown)


enidemi (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -329,7 +333,13 @@ export const NavigatorItem: React.FunctionComponent<NavigatorItemInnerProps> = b
...
 
     return (
-      <FlexRow ref={domElementRef} style={rowStyle} onMouseDown={select} onMouseMove={highlight}>
+      <FlexRow
+        ref={domElementRef}
+        style={rowStyle}
+        onMouseDown={select}
+        onMouseMove={highlight}
+        onDoubleClick={focusComponent}
+      >
```
Rheeseyb, line 342: I think this is fine but it leaves a problem where double clicking different parts of the row have very different actions. IMO if we're going to support double clicking to focus (which definitely feels right to me), it should apply to the label as well, and we should change renaming to be an explicit action triggered by the context menu

Rheeseyb (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
balazsbajorics: I recommend you to try it first, to me it was surprisingly uncontentious that double clicking can lead to two actions – but of course I knew where to click

Rheeseyb: I think that's the thing, you know where to click. I've seen this pattern in a few different apps and I've always found it irritating tbh.

PR #1526, opened 2021-07-16
Prevent preview on focus in certain cases

### LINKED ISSUE #1514: Prevent first item in classname select from autofocusing when the react-select menu opens
With the new "preview on focus" feature of the omnibox CSS menu, due to react-select's default behaviour, when opening the classname select menu we will automatically apply the very first classname.

I had a look into how to prevent this in react-select, and there are a number of open PRs for it (e.g. [this one](https://github.com/JedWatson/react-select/pull/4080)). I haven't found if there is one that has been closed, or if any of the workarounds still work in v4 of react-select.

Rheeseyb:
Fixes #1514 

**Problem:**
When focusing the CSS bar, we open the react-select menu, which triggers a preview of the first option in the menu

**Fix:**
Use a pair of refs to track whether the menu is open, and whether we want to allow previewing on focus. Then we only preview on focus if both are true.

I initially went with the assumption that we never want to trigger it when the menu is opened, but after having played with it a bit I realised the situation was a bit more complex, since:

1. If the topbar is closed, but the input field is focused, pressing an arrow key to open the menu should enable previewing, but clicking the menu to open it shouldn't
2. Typing to filter should immediately trigger previewing
3. Clearing the typed filter shouldn't leave the last focused option previewed
4. Selecting an option or deleting a set value shouldn't trigger a preview of whatever row is focused now

These settings feel very close to right to me, though there are still 2 edge cases that feel wrong that I'm struggling to code for:

1. Opening the menu with a click shouldn't trigger previewing of the first row, but hovering that row with the mouse should - I think this can be fixed but would tackle that in an immediate follow up PR
2. Pressing the left arrow key when nothing is typed will move focus to a previously set element and so clear the current row's focus (which definitely feels correct). That's by default in react-select. But pressing the right arrow key doesn't do that and feels wrong.

I did capture a video but it's very misleading since it doesn't show when I'm opening via the arrow keys or the shortcut key, so really the best option is to try it on pizza

### DIFF
--- a/editor/src/components/canvas/controls/classname-select.tsx
+++ b/editor/src/components/canvas/controls/classname-select.tsx
@@ -337,20 +337,20 @@ export const ClassNameSelect = betterReactMemo(
       dispatch([EditorActions.clearTransientProps()], 'canvas')
     }, [updateFocusedOption, dispatch])
 
-    const isMenuOpen = React.useRef(false)
-    const shouldPreviewOnFocus = React.useRef(false)
+    const isMenuOpenRef = React.useRef(false)
+    const shouldPreviewOnFocusRef = React.useRef(false)
     const onMenuClose = React.useCallback(() => {
-      shouldPreviewOnFocus.current = false
-      isMenuOpen.current = false
+      shouldPreviewOnFocusRef.current = false
+      isMenuOpenRef.current = false
       clearFocusedOption()
     }, [clearFocusedOption])
     const onMenuOpen = React.useCallback(() => {
-      isMenuOpen.current = true
+      isMenuOpenRef.current = true
     }, [])
     const onInputChange = React.useCallback(
       (newInput: string) => {
         if (newInput === '') {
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -336,6 +337,27 @@ export const ClassNameSelect = betterReactMemo(
       dispatch([EditorActions.clearTransientProps()], 'canvas')
     }, [updateFocusedOption, dispatch])
 
+    const isMenuOpen = React.useRef(false)
```
balazsbajorics, line 340: I like to call out refs in their name, like `isMenuOpenRef`

Rheeseyb: Ah yes, good shout

balazsbajorics (approved):

enidemi (approved):

PR #1541, opened 2021-07-19
Assorted performance tweaks.

### LINKED ISSUE #1519: "focus on component" action feels like it takes ~1 second and it horrible
I haven't done actual perf. measurements.

steps to reproduce: keep double clicking on the canvas to focus Card in the beaches project https://utopia.pizza/p/be249c7a-beaches-talwind-v01/

when the Card becomes focused, there's what feels like a half second - second penalty

seanparsons:
Fixes #1519 

**Problem:**
Changing the focus can feel quite sluggish.

**Fix:**
Mostly this was a series of tweaks on this quite heavy code path. The `InspectorEntryPoint` structural tweak does do a lot of the heavy lifting in making this seem to run a bit quicker, along with removing Tippy, which was causing a mid-render recalculate style. Now should only be running the `updateTwind` function when we really need to by checking if the config is deeply equal as that eats about 4ms each time it is called.

**Notes:**
- Our new tooltip doesn't seem to be sitting on top of everything, oddly even tweaking the z-index value didn't seem to have an effect on that which I didn't investigate further.

**Commit Details:**
- Removed Tippy dependency.
- Reworked `Tooltip` implementation that does not rely on Tippy.
- `useSelectOrLiveModeSelectAndHover` hook cancels queued selections
  which prevents a bunch of them replaying in order.
- `OptionChainControl` now wrapped in `betterReactMemo`, along with
  memoisation of the CSS objects passed to Emotion.
- `InspectorEntryPoint` now always shows the `SingleInspectorEntryPoint`
  instance and conditionally shows the others, so that the first is always
  in the same structural position. This maximises the structural re-use,
  otherwise a change to `showSceneInspector` dumps the entire contents.
- `NavigatorRowLabel` now wrapped in `betterReactMemo`.
- Slight improvement to `failoverResolveResults` to make it do less work.
- `combinedFileLookup` now exits as early as possible rather than
  working through the entire array via a `reduce`.
- `failoverResolveResults` now evades the need to check the index of the
  result types as there are only 2 anyway.
- `useGetTailwindConfigFile` checks for deep equality of the config to
  avoid re-running the Twind logic.
- `css-loader` avoids checking a single item array.
- Small tweaks to `NumberInput` to dodge using Emotion for simple styles
  and double manipulating Eithers.


editor/src/uuiui/tooltip.tsx
```diff
@@ -33,41 +22,24 @@ export class Tooltip extends React.Component<React.PropsWithChildren<TooltipProp
...
-    )
+    if (this.props.disabled) {
+      return this.props.children
+    } else {
+      return (
+        <div
+          css={{
+            position: 'relative',
+            '&:hover:after': {
+              background: 'orange',
```
balazsbajorics, line 33: something tells me this is a placeholder design

editor/src/components/menubar/menubar.tsx
```diff
@@ -218,7 +218,7 @@ export const Menubar = betterReactMemo('Menubar', () => {
           </a>
         </Tile>
 
-        <Tooltip title={'Project Info'} placement={'right'}>
```
balazsbajorics, line 221: do we need to lose the placement prop? I guess it wouldn't be hard for someone to use the placement prop in the css of our in-house Tooltip, and restoring these later on would be a one+ hour chore of cherry picking and merge conflicts

balazsbajorics:
we should re-run the tests now that America is sleeping – the selection measurement is _literally_ off the charts

Rheeseyb:
> we should re-run the tests now that America is sleeping – the selection measurement is _literally_ off the charts

And not in the good way :smile:

editor/src/uuiui/tooltip.tsx
seanparsons: It's the one I was supplied with!

editor/src/components/menubar/menubar.tsx
seanparsons: The removal is in this one commit, so it would be more like 10 minutes of copy-paste times: https://github.com/concrete-utopia/utopia/pull/1541/commits/e83d0f857d6899c24d71d78dda28c4496ac08358

But it was removed because the new tooltip doesn't support it and even the type itself was coming from Tippy.

seanparsons:
> we should re-run the tests now that America is sleeping – the selection measurement is _literally_ off the charts

Strictly selection per-se wasn't targeted by this PR but focusing, which is kinda selection++.

editor/src/components/menubar/menubar.tsx
balazsbajorics: it's easier to reverse cherry-pick it while it's fresh and the code did not drift – even if we are not using the information in this PR

seanparsons: That's with the assumption the support will be restored, but if it's left in it's a lie.

balazsbajorics: my assumption is that if we merge orange tooltips to master, someone will need to fix it ASAP

maltenuhn: I wasn't aware we even _have_ a new tooltip - thought you were still investigating whether replacing it with the dummy one I sent you actually makes a performance improvement, or still suffers from the same re-render problems.

seanparsons: OK then, it appears it was purely for diagnostic purposes, so I'll revert the changes.

editor/src/uuiui/tooltip.tsx
maltenuhn: Yeah, this was the sample code to even test if it solves the re-render problem and still even _displays_, what with z-index issues lurking.

Rheeseyb (approved):

PR #1542, opened 2021-07-20
Adding the tailwind classname select into the Inspector

Rheeseyb:
**Problem:**
We want a similar experience to the omnibar's tailwind classname selector in the Inspector

**Fix:**
This PR extracts the core functionality from the existing tailwind classname select, and re-uses it in the inspector's existing classname subsection. It also does the following:

- In the Omnibar:
  - Only opens the menu after a deliberate user action
  - Closes the menu on selection
  - Reduces the menu size to 5 rows + the footer
- In the Inspector section:
  - Always shows the menu, with 9 rows shown (maybe that's too much tbh). The choice to always have this open is deliberate.
  - Moves the classname subsection to the very top
  - Uses similar auto-previewing of the focused element, albeit hooked in slightly differently since this is stuck using the older version of `react-select` as it's using the createable version to allow users to add their own classes)
  - Restricts auto-previewing so that it only starts acting when the user performs a keyboard interaction (to ensure that it is a deliberate thing)
    - This to me feels right, though it also feels like it should also happen on a mouse over of the menu itself when the component is focused, which will require a little extra work

Note that this is still using the original styling of that section. I wanted to get this working first so we could evaluate it before tweaking the styling to match the omnibar. It also still has the dot prefix, though that's easy to remove.

**Screenshots:**
Screenshot_20210720_161929
Screenshot_20210720_161943

### DIFF
--- a/editor/src/components/canvas/controls/classname-select.tsx
+++ b/editor/src/components/canvas/controls/classname-select.tsx
@@ -1,14 +1,9 @@
 /** @jsx jsx */
 
-import React from 'react'
 import { jsx } from '@emotion/react'
 import styled from '@emotion/styled'
-
-import {
-  AllAttributes,
-  AttributeToClassNames,
-  ClassNameToAttributes,
-} from '../../../core/third-party/tailwind-defaults'
+import React from 'react'
+import type { StylesConfig } from 'react-select'
 import WindowedSelect, {
   components,
   FormatOptionLabelMeta,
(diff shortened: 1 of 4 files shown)


editor/src/components/canvas/controls/classname-select.tsx
```diff
@@ -730,7 +568,7 @@ export const ClassNameSelect = betterReactMemo(
             NoOptionsMessage,
             Menu,
             MultiValueContainer,
-            ValueContainer,
+            // ValueContainer,
```
enidemi, line 571: I think this is still neeeded, the ValueContainer had a wrapper div for the css, it was added when we were fixing the sizing with lots of classnames selected the editor became wider than the screen

Rheeseyb: Oops! Good catch, this wasn't supposed to be commented out

Rheeseyb:
Closing this PR for now as it is broken and needs further tweaking. Will reopen a revised version later today

PR #1551, opened 2021-07-21
Fix tailwind media queries bug

Rheeseyb:
**Problem:**
The CSS rule selector scoping was being incorrectly applied to `@media` rules

**Fix:**
Corrected the logic to offset the index when looking for the selectors in `@media` rules, and completely skip other `@` rules (the only other one used by tailwindcss is `@keyframes`, as mentioned in the comment). Also added tests.


seanparsons (commented):
:+1:

PR #1555, opened 2021-07-21
Reduce navigator renders during reparenting.

### LINKED ISSUE #1545: navigator reparent performance is abysmal
**Problem**
Navigator performance in the speedrun was so poor it was running for seconds after the interaction was completed

seanparsons:
Fixes #1545

**Problem:**
Currently re-parenting is surprisingly slow, to the point of making it slowly replay the previous movements once you stop moving it.

**Fix:**
The primary fix in this is to make the `dropTargetHint` value only be present when the target matches the current item in the navigator. Previously as the value changed (3 times approximately if an item is dragged over fully), that would cause every single item in the navigator to be re-rendered.

**Commit Details:**
- Fixes #1545.
- Separated out the type of the `navigator` field in `EditorState`.
- Moved `DropTargetType` and `DropTargetHint` to live next to the
  newly created `NavigatorState` type.
- Made `NavigatorItemDragAndDropWrapperProps.dropTargetHint` be a
  nullable value and renamed it slightly to give some indication
  that it would only be there if relevant.
- `NavigatorItemWrapper` only includes the (previously named)
  `dropTargetHint` value if it points to the same path as the current
  element. Previously when this changed as the user dragged around
  the entire navigator was re-rendered.
- Created a `KeepDeepEqualityCall` instance for the `NavigatorState`
  and used it in the actions in place of `immutability-helper`.

### DIFF
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -518,6 +518,7 @@ import {
   PostCSSPath,
   TailwindConfigPath,
 } from '../../../core/tailwind/tailwind-config'
+import { NavigatorStateKeepDeepEquality } from '../../../utils/deep-equality-instances'
 
 export function updateSelectedLeftMenuTab(editorState: EditorState, tab: LeftMenuTab): EditorState {
   return {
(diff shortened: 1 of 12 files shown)


balazsbajorics (approved):

seanparsons:
@balazsbajorics Would you fancy giving this a spin when it's available to see how much of a difference it makes?

balazsbajorics:
> @balazsbajorics Would you fancy giving this a spin when it's available to see how much of a difference it makes?

yes! :)

balazsbajorics:
it's **much** better now! 🦜

enidemi (approved):

PR #1557, opened 2021-07-21
Chore/darkmode icon

maltenuhn:
**Problem:**
In our slow journey towards a dark theme, the icons had even less contextual awareness of whether it's light outside than our class components do.

**Fix:**
Given that 99% of our icons run through either `<Icn />` or the UNSAFE_helper function, and the rest should, I changed their type signatures so they now takes a semantic colour value ('main', 'subdued', 'primary') instead of a colour, and added a very primitive hook that returns the themed value for it. I then changed the literal colours for the semantic ones everywhere.

**Explanation**
Did it that way because getting the value from the theme in each place would have forced my hand to also make the theme available to all other components -  which then would have taken this 15-minute job far beyond the mindless displacement / replacement activity that could be done while muted on a phone call this turned out to be.

**Caveat**
The dark mode icons themselves are not nice, will definitely need fine tuning like the rest of the theme, but at least they are now visible. I also didn't move any files around yet, to keep this in sync with the Figma icon export.


seanparsons (approved):

editor/src/uuiui/icn.tsx
```diff
@@ -3,8 +3,22 @@ import * as React from 'react'
 import { betterReactMemo } from '../utils/react-performance'
 import { getPossiblyHashedURL } from '../utils/hashed-assets'
 import { Tooltip } from './tooltip'
+import { useColorTheme } from '.'
```
Rheeseyb, line 6:
```suggestion
```

Very minor, but it looks like this import wasn't used in the end

Rheeseyb (approved):

editor/src/uuiui/icn.tsx
balazsbajorics: also, what the hell is `import from '.'`

Rheeseyb: Let's just get this line removed from the code and then we can pretend it never happened and never speak of it again

balazsbajorics (approved):

PR #1566, opened 2021-07-22
Feature/post style action toast

balazsbajorics:
**Problem:**
The editor sometimes converts layout systems totally changing the style props, adding and removing stuff.

**Fix:**
image

This is **not** a fix, only a UI to show when this happens to raise awareness. There's no "undo" or "revert" actions that the user can take, this just shows what happened. This is already helpful because
1. the user can fix the code if they don't like it
2. it helps me **a lot** while I'm going to fix and debug some of the unnecessary style conversions
3. there's a post-action UI in the works that this system can feed into

**Commit Details:**
- I moved the toast timeout to the Toast component, so we can have the toasts inserted in any of the update functions
- Added `createStylePostActionToast` that is called from the two places where layout conversion might happen
- Collecting the toasts to show when we apply a transient drag state


editor/src/components/common/notices.tsx
```diff
@@ -26,6 +26,8 @@ export const getStylesForLevel = (level: NoticeLevel): React.CSSProperties => {
   return resultingStyle
 }
 
+const ToastTimeout = 5500
```
Rheeseyb, line 29: I thought we already had a timeout for these?

Rheeseyb (approved):

editor/src/components/common/notices.tsx
balazsbajorics: I removed the setTimeout in the ADD_TOAST update function, because this PR is making toasts inside the transient canvas state, and pushing the toasts to editorState.toasts only if the transient state is applied at the end of the interaction (we are not showing the toasts during the drag or if the user abandons). 

I actually like the timeout better in the Toast component as well, because it starts counting when the toast visually shows up, not when the ADD_TOAST is called for it.

maltenuhn: does this still preserve the distinction between toasts you need to close manually, and those that close themselves automatically?

balazsbajorics: yep, the `persistent` property controls that

balazsbajorics: https://github.com/concrete-utopia/utopia/blob/e998d9d9dab8ba6e865343b29862152d18de1d2a/editor/src/components/common/notices.tsx#L45

enidemi (approved):

PR #1569, opened 2021-07-22
adding FlexRow and FlexCol to utopia-api

balazsbajorics:
This PR adds FlexRow and FlexCol layout helper components. 

I decided to use Emotion inside them so that the flexDirection is overwriteable from CSS classnames or style prop.

so this: `<FlexRow className="flex-col" />` in a tailwind context turns it into a flexCol as expected.

(using props.style internally would mean that the style prop "wins" over the className)


seanparsons:
I thought we were moving away from the idea of providing our own components?

seanparsons (approved):

enidemi (approved):

Rheeseyb (approved):

Rheeseyb:
> I thought we were moving away from the idea of providing our own components?

Agreed, though I think this is more a stop gap so that we have _something_?

PR #1576, opened 2021-07-22
Fix/wrap in element 1

### LINKED ISSUE #1546: Editor shouldn't add unnecessary style prop / settings when wrapping
**Problem**
We use the same logic for wrapping as for reparenting, which causes unnecessary style prop and values to be inserted

balazsbajorics:
Fixes #1546

**Fix:**
WRAP_IN_ELEMENT action only wraps things and reparents things without touching their style object. This will work for most flex cases, and will not produce nice results for Pinned cases

**Commit Details:**
- made new `editorReparentNoStyleChange` which I copy-pasted from `moveTemplate`
- new WRAP_IN_ELEMENT action uses this instead of moveTemplate
- floating insert menu has a checkbox to switch between new WRAP_IN_ELEMENT and old WRAP_IN_VIEW behavior

image

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -1777,6 +1777,80 @@ export function getFrameChange(
   }
 }
 
+function editorReparentNoStyleChange(
+  target: ElementPath,
+  indexPosition: IndexPosition,
+  newParentPath: ElementPath,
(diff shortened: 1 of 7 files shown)


balazsbajorics:
I'm going rogue and merging this WITHOUT REVIEWS! please still check it

PR #1577, opened 2021-07-23
Flex insertion handles cope with the flex direction correctly.

seanparsons:
**Problem:**
The flex insertion controls only worked for flex rows.

**Fix:**
Check the appropriate field in the size measurements to determine which axis the controls should align against.

**Commit Details:**
- `parentFlexDirection` is consulted when determining which direction
  the flex insertion buttons should be arrayed along.

### DIFF
--- a/editor/src/components/canvas/controls/insertion-plus-button.tsx
+++ b/editor/src/components/canvas/controls/insertion-plus-button.tsx
@@ -52,7 +52,7 @@ export const InsertionControls: React.FunctionComponent<ControlProps> = betterRe
         props.componentMetadata,
       )
       if (child.specialSizeMeasurements.position !== 'absolute' && childFrame != null) {
-        let direction: 'row' | 'column' = 'column'
+        let direction: 'row' | 'column' | null = null
         if (child.specialSizeMeasurements.parentLayoutSystem === 'flex') {
           switch (child.specialSizeMeasurements.parentFlexDirection) {
             case 'row':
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/controls/insertion-plus-button.tsx
```diff
@@ -52,10 +52,17 @@ export const InsertionControls: React.FunctionComponent<ControlProps> = betterRe
...
-            : 'column'
+        let direction: 'row' | 'column' = 'column'
+        if (child.specialSizeMeasurements.parentLayoutSystem === 'flex') {
+          switch (child.specialSizeMeasurements.parentFlexDirection) {
+            case 'row':
+            case 'column':
+              direction = child.specialSizeMeasurements.parentFlexDirection
+              break
+            default:
+            // Ignore any other values.
```
balazsbajorics, line 63: hmm, how could we safeguard against forgetting to make this work for row-reverse etc? maybe we should bail out from shoing the control, and early return with a `null`

seanparsons: Yeah, I was on the fence about that, we could make the `direction` field nullable and skip adding to `controlProps` if it's null.

balazsbajorics: sounds good! also, `controlProps` is probably a _very bad_ variable name because of the loaded meaning coming from the Inspector codebase

Rheeseyb (approved):

editor/src/components/canvas/controls/insertion-plus-button.tsx
seanparsons: Fixed in https://github.com/concrete-utopia/utopia/pull/1577/commits/00bb0fb34cfbc91b93ccf9a1d4a4e5485876c460

balazsbajorics (approved):

PR #1579, opened 2021-07-23
WRAP_IN_VIEW fixes

balazsbajorics:
**Problem:**
WRAP_IN_VIEW is the old wrapping function, (can be enabled if you pick 'Try to preserve visual position' in the floating wrapper menu)
image
the old wrapping function was built in an absolute positioned world. The flex functionality was sort of slapped on it from an absolute positioning POV. Back then we didn't respect the idea that a lot of flex elements **should not have any explicit size properties at all.**

**Fix:**
For specific cases of flex-to-flex reparenting, try not to insert size properties if they weren't there before. If there is a flexBasis or width or height, then we run the old code. If none of the three is present, we don't forcefully insert them.

**Commit Details:**
- Fixed the wrapping index position I already fixed for WRAP_IN_ELEMENT
- Added new conversion function for flex-to-flex layout change, if the flexDimension is different
- Wired through the "new parent flex direction" property so we don't try to fish it out of the metadata, which would fail for obvious reasons during insertion

**Note:**
This is one of the messiest part of the code base. I think that my fix is not harmful, and in the context of working with flex layouts it brings a significant improvements. But the code quality around here deteriorates quickly with every change. I think as we work on more fixes in this area, it's worth considering to not use `moveTemplate` for any new feature ever again because of the incredible baggage.

### DIFF
--- a/editor/src/components/canvas/ui/floating-insert-menu.tsx
+++ b/editor/src/components/canvas/ui/floating-insert-menu.tsx
@@ -38,6 +38,7 @@ import { generateUidWithExistingComponents } from '../../../core/model/element-t
 import {
   jsxAttributeValue,
   jsxElement,
+  JSXElementName,
   jsxTextBlock,
   setJSXAttributesAttribute,
 } from '../../../core/shared/element-template'
@@ -51,6 +52,7 @@ import { InspectorInputEmotionStyle } from '../../../uuiui/inputs/base-input'
 import { when } from '../../../utils/react-conditionals'
 import { ElementPath } from '../../../core/shared/project-file-types'
 import { safeIndex } from '../../../core/shared/array-utils'
+import { LayoutSystem } from 'utopia-api'
 
 type InsertMenuItemValue = InsertableComponent & {
   source: InsertableComponentGroupType | null
@@ -124,6 +126,26 @@ function useGetInsertableComponents(): InsertableComponentFlatList {
   }, [packageStatus, propertyControlsInfo, projectContents, dependencies, fullPath])
 
   return insertableComponents
+}
+
+function getIsFlexBasedOnName_KILLME_EXPERIMENTAL(name: JSXElementName): boolean {
+  return (
+    name.propertyPath.propertyElements.length === 0 &&
+    (name.baseVariable === 'FlexRow' || name.baseVariable === 'FlexCol')
+  )
+}
+
+function getIsFlexDirectionBasedOnName_KILLME_SERIOUSLY_EXPERIMENTAL(
+  name: JSXElementName,
+): 'horizontal' | 'vertical' | null {
+  if (name.propertyPath.propertyElements.length === 0) {
+    if (name.baseVariable === 'FlexRow') {
+      return 'horizontal'
+    } else if (name.baseVariable === 'FlexCol') {
+      return 'vertical'
+    }
+  }
+  return null
 }
 
 function useComponentSelectorStyles(): StylesConfig<InsertMenuItem, false> {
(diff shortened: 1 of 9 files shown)


editor/src/core/layout/layout-helpers.ts
```diff
@@ -220,9 +228,10 @@ export const FlexLayoutHelpers = {
     height: number,
     props: JSXAttributes,
     parentProps: PropsOrJSXAttributes,
+    possibleMainAxis: 'horizontal' | 'vertical' | null,
     edgePosition: EdgePosition | null,
   ): Either<string, FixedAndBases> => {
-    const possibleMainAxis = FlexLayoutHelpers.getMainAxis(parentProps)
+    // const possibleMainAxis = FlexLayoutHelpers.getMainAxis(parentProps)
```
Rheeseyb, line 234:
```suggestion
```

editor/src/components/canvas/ui/floating-insert-menu.tsx
```diff
@@ -126,6 +128,26 @@ function useGetInsertableComponents(): InsertableComponentFlatList {
   return insertableComponents
 }
 
+function getIsFlexBasedOnName_KILLME_EXPERIMENTAL(name: JSXElementName): boolean {
```
Rheeseyb, line 131: Can you add a ticket for removing these please

Rheeseyb (approved):

editor/src/components/canvas/ui/floating-insert-menu.tsx
balazsbajorics: big time

editor/src/components/editor/actions/actions.ts
```diff
@@ -2093,6 +2100,15 @@ export const UPDATE_FNS = {
           derived,
         )
         const parentPath = EP.getCommonParent(orderedActionTargets)
+        const indexInParent = optionalMap(
+          (firstPsthMatchingCommonParent) =>
```
seanparsons, line 2104: `Psth`?

balazsbajorics: and this is now 4 times in the codebase! :D will fix it in a moment

editor/src/core/layout/layout-utils.ts
```diff
@@ -122,13 +129,39 @@ export function maybeSwitchLayoutProps(
...
 
   let wasFlexContainer = MetadataUtils.isFlexLayoutedContainer(originalParent)
+  const oldParentMainAxis: 'horizontal' | 'vertical' | null = eitherToMaybe(
+    flatMapEither<string, JSXElementChild, 'horizontal' | 'vertical'>((parentElement) => {
+      if (isJSXElement(parentElement)) {
+        return FlexLayoutHelpers.getMainAxis(right(parentElement.props))
+      } else {
+        return leftEither('parent is not JSXElement')
+      }
+    }, newParent?.element ?? leftEither('no parent provided')),
```
seanparsons, line 139: This parameter could do with pulling out into a variable to make this expression a little simpler.

editor/src/core/layout/layout-utils.ts
```diff
@@ -122,13 +129,39 @@ export function maybeSwitchLayoutProps(
...
+  )
+
   let isFlexContainer =
     parentLayoutSystem === 'flex' || MetadataUtils.isFlexLayoutedContainer(newParent)
+
+  const parentMainAxis: 'horizontal' | 'vertical' | null =
+    newParentMainAxis ??
+    // if no newParentMainAxis is provided, let's try to find one
+    // TODO maybe overkill
+    eitherToMaybe(
```
seanparsons, line 149: This whole blocks _appears_ to be the same as the one above.

seanparsons (approved):

editor/src/core/layout/layout-utils.ts
balazsbajorics: fixed

editor/src/core/layout/layout-utils.ts
balazsbajorics: fixed

editor/src/components/editor/actions/actions.ts
balazsbajorics: fixed

editor/src/components/canvas/ui/floating-insert-menu.tsx
balazsbajorics: added ticket https://github.com/concrete-utopia/utopia/issues/1580

PR #1583, opened 2021-07-23
Fix/inspector delete layout config

enidemi:
Fixes the inspector, allows deleting layout properties.

**Problem:**
An element style set to `position: absolute` can be deleted only using the code editor.

**Fix:**
Updating the current layout system list, because it still had the old layoutSystem in it.
Adding the ability to unset from the new experimental layout section for self layout properties.
The list is not complete yet, it contains style properties that we were already showing in the inspector.


balazsbajorics (approved):

Rheeseyb (approved):

balazsbajorics:
Eni I'm merging this if you don't mind, to squeeze it in time!!

PR #1587, opened 2021-07-23
Blur classname select when pressing esc

### LINKED ISSUE #1586: Esc should exit the classname input field
Hitting "esc" with the classname input field in the Inspector focused fails to remove the focus, meaning we can't use shortcuts without clicking to focus another area of the editor

Rheeseyb:
Fixes #1586 

**Problem:**
react-select doesn't blur when hitting `esc`

**Fix:**
Manually handle that key in the same place we have added other custom key logic


Rheeseyb:
I'm going to just merge this one straight away since it's super simple

PR #1606, opened 2021-07-26
Handle components being created from within external components.

### LINKED ISSUE #1540: Custom components in react-select break the canvas
`react-select` allows users to provide overrides to its internal components in a somewhat unique way, that for some reason breaks our canvas.

In the attached project, the canvas shows an error indicating that an Instance Path is missing for the custom component: https://utopia.app/p/3163e90b-react-select-bug/

We see a somewhat related bug when trying to use R3F, which is caused by the context being reset (see #1000), so this bug may be tangentially related to that.

seanparsons:
Fixes #1540

**Problem:**
With `react-select` it's possible to pass our components into the `Select` component for it to use them in construction of the UI. However that means that those elements are created potentially without some special fields like `data-paths`.

**Fix:**
Made it so that the canvas would not kill the rendering of any components if it was unable to retrieve the `data-paths` attribute.

**Commit Details:**
- Fixes #1540.
- Make `tryToGetInstancePath` return null if it can't obtain one,
  instead of throwing an exception.
- Handle the fallout of the above change by adding `null` to parameter
  types of a few functions and use `optionalMap` to cater for these
  parameters.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
@@ -58,14 +58,14 @@ function tryToGetInstancePath(
   topLevelElementName: string,
   maybePath: ElementPath | null,
   pathsString: string | null,
-): ElementPath {
+): ElementPath | null {
   const paths = getPathsFromString(pathsString)
   if (EP.isElementPath(maybePath)) {
     return maybePath
   } else if (paths.length > 0) {
     return paths[0]
   } else {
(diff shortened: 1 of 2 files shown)


Rheeseyb (approved):

Rheeseyb:
So this means we won't be able to select anything coming from those passed in components?

enidemi (approved):

seanparsons:
> So this means we won't be able to select anything coming from those passed in components?

Ah, sorry, didn't notice this reply in here, the components will likely be invisible to the editor. The difficulty here is that there's no element, only a reference to the component. With `react-select` creating the element and supplying it with properties.

PR #1610, opened 2021-07-27
feature(canvas) Add parent highlighting.

### LINKED ISSUE #1604: Resurrect the "show additional info on hover" experiment
**Problem**
Doing layout tasks via Utopia is tricky at the moment because it's not as easy as it should be to see the results while make configuration changes via code or the inspector.

We had #514 which experimented with ideas for how to solve this, so we could resurrect it https://github.com/concrete-utopia/utopia/pull/514

The purpose here is to _then_ turn this into a featureset that shows additional controls for parent and siblings while (e.g.) hovering over the inspector layout section, or while making changes to an element via code. 

What exactly we show, and when (triggers and conditions) we show it, will probably change as we use it for the speedruns (for instance, we showing the parent markers is valuable, showing the numbered number of siblings less so).  If that means this PR here is no longer a good base, and we'd be faster doing this zero-based, please flag _before_ spending time resurrecting it

seanparsons:
Fixes #1604

**Problem:**
When adjusting the layout of an element, users want to be able to see the parent dimensions they're adjusting the layout in relation to the child's dimensions.

**Fix:**
When either `cmd` is pressed or if the layout section is hovered we display an indicator showing the 4 corners of the parent.

**Commit Details:**
- Fixes #1604.
- Partly resurrects #514.
- Added `layoutSectionHovered` property to `EditorState.inspector`.
- Added action for updating `layoutSectionHovered`.
- Added a div around the layout section to capture mouse over and out.
- Added parent highlights to the `OutlineControl` component, optionally
  depending on if `cmd` is pressed or if the layout section is hovered.

### DIFF
--- a/editor/src/components/canvas/controls/outline-control.tsx
+++ b/editor/src/components/canvas/controls/outline-control.tsx
@@ -18,6 +18,7 @@ import { fastForEach } from '../../../core/shared/utils'
 import { isFeatureEnabled } from '../../../utils/feature-switches'
 import { useColorTheme } from '../../../uuiui'
 import { useEditorState } from '../../editor/store/store-hook'
+import { KeysPressed } from '../../../utils/keyboard'
 
 export function getSelectionColor(
   path: ElementPath,
@@ -42,6 +43,7 @@ export function getSelectionColor(
 
 export interface OutlineControlsProps extends ControlProps {
   dragState: MoveDragState | ResizeDragState | null
+  keysPressed: KeysPressed
 }
 
 function isDraggingToMove(
@@ -56,6 +58,10 @@ function isDraggingToMove(
 export const OutlineControls = (props: OutlineControlsProps) => {
   const colorTheme = useColorTheme()
   const { dragState } = props
+  const layoutInspectorSectionHovered = useEditorState(
+    (store) => store.editor.inspector.layoutSectionHovered,
+    'OutlineControls layoutInspectorSectionHovered',
+  )
   const getDragStateFrame = React.useCallback(
     (target: ElementPath): CanvasRectangle | null => {
       if (isDraggingToMove(dragState, target)) {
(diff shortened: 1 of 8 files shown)


enidemi (approved):

editor/src/components/canvas/controls/outline-control.tsx
```diff
@@ -124,10 +130,72 @@ export const OutlineControls = (props: OutlineControlsProps) => {
     [props.canvasOffset.x, props.canvasOffset.y, props.componentMetadata, props.dragState],
   )
 
-  const anySelectedElementIsYogaLayouted = anyInstanceYogaLayouted(
-    props.componentMetadata,
-    props.selectedViews,
-  )
+  let parentHighlights: React.ReactNode = null
+  if (props.keysPressed['cmd'] || layoutInspectorSectionHovered) {
+    parentHighlights = props.selectedViews.map((view) => {
```
Rheeseyb, line 135: This should probably be wrapped in a `React.useMemo`

editor/src/components/inspector/sections/layout-section/layout-section.tsx
```diff
@@ -53,9 +56,16 @@ export const LayoutSection = betterReactMemo('LayoutSection', (props: LayoutSect
   )
 
   return (
-    <>
+    <div
+      onMouseOver={() => {
+        dispatch([setInspectorLayoutSectionHovered(true)], 'everyone')
+      }}
+      onMouseOut={() => {
+        dispatch([setInspectorLayoutSectionHovered(false)], 'everyone')
```
Rheeseyb, line 64: Is the hover state still cleared if I mouse over and then `cmd` tab to a different application, move the mouse, then switch back to Utopia?

seanparsons: I'll test that and report back.

seanparsons: That appears to work fine testing it here.

editor/src/components/canvas/controls/outline-control.tsx
seanparsons: Fixed in 8b4f17d8.

Issue #1602, opened 2021-07-26
Resurrect "Parent layout element info" experiment


maltenuhn:
https://github.com/concrete-utopia/utopia/pull/504

PR #576, opened 2020-09-29
experiment/parent-layout-controls-1

enidemi:
Canvas Experiment

Based on #514

How to try it?
Select a flex element to see the controls above the flex container. It offers changing flex container properties: align-items, justify-content and flex-direction.

https://utopia.pizza/p/81e13404-bow-wound/?branch_name=experiment/parent-layout-controls-1

PR #1611, opened 2021-07-27
feature(canvas-controls) parent control experiment

enidemi:
Fixes https://github.com/concrete-utopia/utopia/issues/1602

**Problem:**
It would be nice to show more information about how an element is layouted.

**Fix:**
Resurrecting experiment https://github.com/concrete-utopia/utopia/pull/576.

**Commit Details:**
- it uses the same feature switch as the inspector layout section
- for flex elements it's possible to switch parent flex direction
- it shows the parent layout type


editor/src/components/canvas/controls/layout-parent-control.tsx
```diff
@@ -0,0 +1,193 @@
...
+    return null
+  }
+  return (
+    flexDirectionOptions(flexWrap).find((option) => option.value === flexDirection)?.icon ?? null
+  )
+}
+
+export const LayoutParentControl = betterReactMemo(
+  'LayoutParentControl',
+  (): JSX.Element | null => {
```
Rheeseyb, line 34: We don't need the return type here

Rheeseyb (approved):

PR #1613, opened 2021-07-27
Fix/inspector flex container unset labels

enidemi:
**Problem:**
Some inspector control labels are shown as 'set' while their values are actually empty.

**Fix:**
Fixing align items, flex wrap rows.

**Commit Details:**
- wrapped align items in contextmenu wrapper that applies the color change on hover
- pass controlstatus to flex wrap menu


editor/src/components/inspector/controls/option-chain-control.tsx
```diff
@@ -45,6 +45,7 @@ export const OptionChainControl: React.FunctionComponent<DEPRECATEDControlProps<
       display: 'flex',
       flexDirection: 'column',
       marginBottom: 0,
+      width: '100%',
```
Rheeseyb, line 48: Were these changes intended as part of this?

Rheeseyb (approved):

editor/src/components/inspector/controls/option-chain-control.tsx
enidemi: yes, the align-items control is now in a contextmenu wrapper and it shrinked in size.

PR #1616, opened 2021-07-27
Update global-shortcuts.ts

maltenuhn:
**Problem:**
The canvas got stuck in zoom in / out mode triggered by `z` key. Diagnosing it revealed it's because we weren't clearing the `z` key on key up.

**Fix:**
Clear the key. I don't know why we weren't doing this before. Perhaps it conflicted with draftJS?

### DIFF
--- a/editor/src/components/editor/global-shortcuts.ts
+++ b/editor/src/components/editor/global-shortcuts.ts
@@ -810,22 +810,13 @@ export function handleKeyUp(
   // Ensure that any key presses are appropriately recorded.
   const key = Keyboard.keyCharacterForCode(event.keyCode)
   const editorTargeted = editorIsTarget(event, editor)
-  let updatedKeysPressed: KeysPressed
-  if (editorTargeted) {
-    updatedKeysPressed = updateKeysPressed(
-      editor.keysPressed,
-      key,
-      false,
-      Modifier.modifiersForKeyboardEvent(event),
-    )
-  } else {
-    updatedKeysPressed = updateKeysPressed(
-      editor.keysPressed,
-      key,
-      false,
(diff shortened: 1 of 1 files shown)


editor/src/components/editor/global-shortcuts.ts
```diff
@@ -812,8 +812,10 @@ export function handleKeyUp(
   const editorTargeted = editorIsTarget(event, editor)
   let updatedKeysPressed: KeysPressed
   if (editorTargeted) {
```
maltenuhn, line 814: If the code here makes sense, there's no longer a need for this condition here, since both branches of the `if` statement do the same thing...

Rheeseyb: This looks correct, but I'm not sure of the reasoning why it was done like this in the first place? @seanparsons do you recall?

seanparsons: I think this was done this way to mirror what is in `handleKeyDown`, I don't immediately see any reason why we shouldn't collapse it for key up, as it looks wrong now in the cold light of day.

Rheeseyb (approved):
Approved on the assumption that you do indeed bin off the `if` statement

seanparsons (approved):

editor/src/components/editor/global-shortcuts.ts
maltenuhn: I guess we'll discover if there truly _was_ a reason for this, but the reference to editorTargeted makes me think it's from draft days.

Rheeseyb: Quite possibly yeah. Worst case scenario we break the editor, but that's really _worst case_ so it's probs fine

PR #1630, opened 2021-07-28
Make menus scrollable to stop them offsetting the editor

Rheeseyb:
**Problem:**
When resizing the code pane on certain screen sizes it is possible (and often the case) that the editor will become permanently offset.

**Fix:**
I managed to track this issue down to the left and right menu panes. When resizing the code pane, those two can become slightly squished, causing them to stretch slightly, and potentially offset the editor. To fix this, I have given them both fixed widths and `overflowX: 'scroll'`.


maltenuhn:
Amazing find. How did you isolate this to those two menus?! Intuition? I want to know the secret!

PR #1631, opened 2021-07-28
Feature/layout system section reduced

maltenuhn:
**Problem:**
The Layout System section was not clear enough, and behaves differently from *cough* other famous design tools. Specifically, clicking on it didn't add a system, the "systems" it could add included flow which caused the thing itself to be laid out via flow, some of the buttons didn't do anything etc.

**Fix:**
1. I made the section header interactive. Clicking on it now _adds a flex system_ as the default. It's what you want all the time, 70% of the time.
2. I removed 'flow' and 'pin system' as options.
3. Clicking the layout system "-" button now actually removes it. 

NB There's almost certainly a cleaner implementation, and I didn't clean up the code around the "layout systems" - just removed how it's exposed here.

### DIFF
--- a/editor/src/components/inspector/sections/layout-section/layout-system-subsection/layout-system-subsection.tsx
+++ b/editor/src/components/inspector/sections/layout-section/layout-system-subsection/layout-system-subsection.tsx
@@ -4,8 +4,6 @@ import { jsx, css } from '@emotion/react'
 import { UIGridRow } from '../../../widgets/ui-grid-row'
 import {
   LayoutSystemControl,
-  FlexPaddingControl,
-  paddingPropsToUnset,
   DeleteAllLayoutSystemConfigButton,
   useLayoutSystemData,
 } from './layout-system-controls'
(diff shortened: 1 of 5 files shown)


Rheeseyb (approved):

editor/src/components/inspector/sections/layout-section/layout-system-subsection/layout-system-subsection.tsx
```diff
@@ -37,27 +45,49 @@ export const LayoutSystemSubsection = betterReactMemo<LayoutSystemSubsectionProp
...
-          <SquareButton highlight onClick={toggleSection}>
-            <ExpandableIndicator
-              testId='layout-system-expand'
-              visible
-              collapsed={!layoutSectionOpen}
-              selected={false}
-            />
-          </SquareButton>
-        </InspectorSubsectionHeader>
+          {!layoutSectionOpen && (
```
seanparsons, line 83: Better to use `unless` for this, and `when` for the line above it would appear.

seanparsons (approved):

PR #1634, opened 2021-07-29
Stop removing functions from recorded props

Rheeseyb:
**Problem:**
Adding an element `<div style={{ top }} />` will lock up the editor when running locally

**Fix:**
When recording the props of an element into our metadata objects, we were blanking out functions and various other specific cases, since previously those would trigger infinite re-renders of the canvas. It seems that is no longer the case (the canvas was previously taking the metadata objects as props directly, whereas we no longer provide them like that).

The reason this is a problem is that during regular use you can end up with a gigantic object in the props, e.g. when typing out the code `style={{ top: 10 }}` you will for a split second have the code `style={{ top }}`, which compiles and is runs since `top` is a global variable referring to the same object as `window`. That specific case will result in cross origin errors when running in production, but when running locally will result the function `makeCanvasElementPropsSafe` taking an age to run (repeatedly) over that object, locking up the editor.

By replacing that function with a very simple one that adds the `skipDeepFreeze` key to the props, the issue is no longer present, but it creates another issue in that the Event Handler section of the Inspector renders far more often than it needs to if your code has an inlined event handler (e.g. `<div onMouseDown={() => console.log('You should be using React.useCallback instead') />`). The reason for that is simply that this section uses the standardised Inspector hooks to gather the data it needs, so by replacing that hook use with a far simpler one the problem there is solved.

I've also added some extra render count tests that include an inlined event handler like this, ~and updated the snapshots broken by this change (hence the huge diff :/ ). I'm not sure if it's worth replacing those affected snapshots with something more sensible in light of this or as part of this.~ and updated the snapshot tests to ensure children are removed from metadata during those.


editor/src/components/canvas/ui-jsx-canvas.test-utils.tsx
```diff
@@ -89,7 +89,7 @@ function resolveTestFiles(
   }
 }
 
-export function stripUidsFromMetadata(metadata: ElementInstanceMetadata): ElementInstanceMetadata {
+function stripUidsFromMetadata(metadata: ElementInstanceMetadata): ElementInstanceMetadata {
```
seanparsons, line 92: Any reason for unexporting this?

Rheeseyb: It wasn't used anywhere

editor/src/components/inspector/sections/event-handlers-section/event-handlers-section.tsx
```diff
@@ -42,15 +43,58 @@ export const EventHandlerControl = betterReactMemo(
...
+
+function useGetEventHandlerInfo(): EventHandlerValues {
+  const multiselectedProps = useGetMultiselectedProps(ppCreate, regularArrayDOMEventHandlerNames)
+  let result: EventHandlerValues = {}
 
-  const valueKeys = Object.keys(value)
+  forEachValue((values, eventHandlerName) => {
+    if (values.length === 1) {
+      const eitherValue = values[0]
+      if (isRight(eitherValue) && isJSXAttributeOtherJavaScript(eitherValue.value)) {
```
seanparsons, line 56: `forEachRight` is made for this exact kind of case.

editor/src/components/inspector/sections/event-handlers-section/event-handlers-section.tsx
```diff
@@ -42,15 +43,58 @@ export const EventHandlerControl = betterReactMemo(
...
+  (props: { eventHandlerName: string; value: string }) => {
+    const { eventHandlerName, value } = props
+
+    const { onContextUnsetValue } = useInspectorContext()
+    const onUnsetValue = React.useCallback(
+      () => onContextUnsetValue([ppCreate(eventHandlerName)], false),
+      [eventHandlerName, onContextUnsetValue],
+    )
+
+    const eventHandlersContextMenuItems = utils.stripNulls([
```
seanparsons, line 76: Looks like this should be wrapped in `useMemo`.

seanparsons (approved):

editor/src/components/inspector/sections/event-handlers-section/event-handlers-section.tsx
Rheeseyb: Done

editor/src/components/inspector/sections/event-handlers-section/event-handlers-section.tsx
Rheeseyb: Done

PR #1636, opened 2021-07-30
Safer focusing handling.

### LINKED ISSUE #1632: **Regression: Every element has become focusable via double-click**
We previously had logic that determined what components could be focused via double-click (on canvas, and via navigator). It seems that's no longer the case, and as a result even a div can be focused.

seanparsons:
Fixes #1632

**Problem:**
It's possible to focus elements that shouldn't be focusable.

**Fix:**
Prevent the navigator items from attempting to focus ineligible items but also make the focusing action defend against rogue invocations as a final insurance measure.

**Commit Details:**
- Fixes #1632.
- Make the `focusComponent` handler in `NavigatorItem` only attempt to
  focus a component that is focusable.
- Made the `SET_FOCUSED_ELEMENT` more defensive by having it check if
  an element is focusable.
- Flip an if condition in `normalisePathToUnderlyingTarget` to avoid
  a double negative.

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item.tsx
@@ -314,8 +314,10 @@ export const NavigatorItem: React.FunctionComponent<NavigatorItemInnerProps> = b
       [dispatch, elementPath, selected, isHighlighted],
     )
     const focusComponent = React.useCallback(() => {
-      dispatch([EditorActions.setFocusedElement(elementPath)])
-    }, [dispatch, elementPath])
+      if (isFocusedComponent) {
+        dispatch([EditorActions.setFocusedElement(elementPath)])
+      }
+    }, [dispatch, elementPath, isFocusedComponent])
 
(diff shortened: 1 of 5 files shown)


editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -314,8 +314,10 @@ export const NavigatorItem: React.FunctionComponent<NavigatorItemInnerProps> = b
       [dispatch, elementPath, selected, isHighlighted],
     )
     const focusComponent = React.useCallback(() => {
-      dispatch([EditorActions.setFocusedElement(elementPath)])
-    }, [dispatch, elementPath])
+      if (isFocusedComponent) {
```
Rheeseyb, line 317: Shouldn't this be `if (isFocusableComponent) {`?

seanparsons: Why yes, yes it should, fixed in 650a8f4e.

Rheeseyb (approved):

enidemi (approved):

PR #1640, opened 2021-08-02
Quick fixes 31 July

### LINKED ISSUE #1641: Multiselect triggers infinite navigator scrolling loop
STR
1. Open https://utopia.pizza/p/e45f6da0-peat-brush/
2. Select last item in the navigator
3. Holding shift, select first item

Possibly interact with the canvas ( ?)
Result: infinite scrolling loop

maltenuhn:
Fixes #1641 

**Fixes various problems in the speedruns**
* Strip unnecessary whitespace around text content causes issues in the content bar
* Strip unnecessary whitespace around text content causes issues in the toast
* Adds translucency to padding controls
* Navigator needs to be resizable


editor/src/components/canvas/controls/formula-bar.tsx
```diff
@@ -166,7 +166,7 @@ export const FormulaBar = betterReactMemo('FormulaBar', () => {
           }}
           onChange={onInputChange}
           onBlur={onBlur}
-          value={simpleText}
+          value={simpleText.trimStart()}
```
Rheeseyb, line 169: This should probably be using `.trim()` rather than `.trimStart()`

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -65,6 +65,7 @@ export const CanvasWrapperComponent = betterReactMemo(
           // ^ prevents Monaco from pushing the inspector out
         }}
       >
+        {/* <FloatingInsertMenu /> */}
```
Rheeseyb, line 68: Don't forget to uncomment at least one of these!

editor/src/components/canvas/controls/formula-bar.tsx
maltenuhn: Haha, that was my thinking too. Until I tried typing `Hello World` and couldn't get the space in :D

editor/src/components/canvas/design-panel-root.tsx
```diff
@@ -301,21 +309,50 @@ export const DesignPanelRoot = betterReactMemo('DesignPanelRoot', (props: Design
...
+                }}
+              >
+                <ResizableFlexColumn
+                  style={{
+                    overscrollBehavior: 'contain',
+                    backgroundColor: UtopiaTheme.color.bg0.o(90).value,
+                    backdropFilter: 'blur(7px)',
+                  }}
+                  defaultSize={{
+                    width: 280,
```
Rheeseyb, line 331: We have `LeftPaneDefaultWidth` for this, which is being used elsewhere in the editor for some offsetting. I'm not sure if you want to update that to 280 and use it here

editor/src/components/canvas/canvas-wrapper-component.tsx
maltenuhn: actually, removed both of these. The whole menu needed to move file, otherwise nothing doing with the z-index

editor/src/components/canvas/controls/formula-bar.tsx
Rheeseyb: Ha of course!

editor/src/components/canvas/design-panel-root.tsx
maltenuhn: Good question. I think I'll keep them separate for now, not sure if the ideal navigator width is the same thing as the ideal leftMenu width.

enidemi (approved):

editor/src/components/canvas/controls/formula-bar.tsx
Rheeseyb: I've tested this out, and it gives a rather unsavoury result when the text is on a separate line to the element, e.g.:
```
    <div
      style={{
        width: '100%',
        height: '100%',
        backgroundColor: '#FFFFFF',
        position: 'relative',
      }}
    >
      Hello World
    </div>
```
when making a non-change via the omnibar will result in:
```
    <div
      style={{
        width: '100%',
        height: '100%',
        backgroundColor: '#FFFFFF',
        position: 'relative',
      }}
    >
      Hello World{' '}
    </div>
```

I've checked and you can fix that by using `trim()` when setting the value on blur, but I've also noticed another issue in the way we're setting the initial value of the omnibar which potentially will result in data loss when typing fast. I'm going to create an issue ticket for it.

Rheeseyb: I've created https://github.com/concrete-utopia/utopia/issues/1646 for this

Rheeseyb (approved):

seanparsons (approved):

PR #1643, opened 2021-08-02
Moving moveTemplate tests to a separate workflow

Rheeseyb:
**Problem:**
The `.spec.browser` tests are taking forever to run, with one specific suite taking the bulk of that time.

**Fix:**
Move the `moveTemplate` tests into a separate file, and create a separate workflow for running only those tests, allowing us to cut 10-15 mins off the PR checks time


enidemi (commented):
👍

PR #1650, opened 2021-08-02
Only display resize options on mouse down.

### LINKED ISSUE #1649: Resizing canvas controls (the menu that shows what is being updated) should be triggered on mouse down rather than on hover.

seanparsons:
Fixes #1649.

**Problem:**
Displaying the resize options was too aggressive as they would show on mouse over of an edge.

**Fix:**
Mouse over no longer triggers the resize options and also a bug with not updating the target property mid-drag has also been fixed.

**Commit Details:**
- Fixes #1649.
- Removed `ResizeEdgeState.showLabel` completely, along with the
  mouse event handlers that changed the value.
- Trigger for displaying `PropertyTargetSelector` is only triggered
  when the edge is being dragged.
- `ResizeDragState.targetProperty` is updated each time
  `updateResizeDragState` is called so that it can be changed while
  dragging the edge.

### DIFF
--- a/editor/src/components/canvas/controls/size-box.tsx
+++ b/editor/src/components/canvas/controls/size-box.tsx
@@ -128,9 +128,7 @@ interface ResizeEdgeProps {
   dragState: DragState | null
 }
 
-interface ResizeEdgeState {
-  showLabel: boolean
-}
+interface ResizeEdgeState {}
 
 function allowsInteractiveResize(layoutSystem: DetectedLayoutSystem): boolean {
   switch (layoutSystem) {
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/controls/size-box.tsx
```diff
@@ -128,9 +128,7 @@ interface ResizeEdgeProps {
   dragState: DragState | null
 }
 
-interface ResizeEdgeState {
-  showLabel: boolean
-}
+interface ResizeEdgeState {}
```
Rheeseyb, line 131: This should just be deleted now then

Rheeseyb (commented):
This looks good but it seems the flex controls also need the same change to switch from mouse over to mouse down

editor/src/components/canvas/controls/size-box.tsx
seanparsons: I just kept it because it didn't even seem worth removing.

seanparsons:
> This looks good but it seems the flex controls also need the same change to switch from mouse over to mouse down

Good catch.

enidemi (approved):

seanparsons:
> This looks good but it seems the flex controls also need the same change to switch from mouse over to mouse down

Fixed in e3e9f04a.

editor/src/components/canvas/controls/size-box.tsx
Rheeseyb: Can you delete it please since there's no reason to keep it

Rheeseyb (approved):

PR #1652, opened 2021-08-02
Quickfix - outline draws the 'x' in the right place(s)

maltenuhn:
**Problem:**
The prototype parent marks used the letter 'x', which therefore wasn't perfectly center-aligned.

**Fix:**
This adds a new SVG-based control, which is (self)aligned perfectly
NB: like all of the experimental canvas controls, it's not yet zoom-capable, but is zoom-friendly because of the below v v v.

**Details**
The SVG here has a fixed width, and is self-centering because of the `translate: '-50% -50%'` at any zoom/scale factor.


editor/src/components/canvas/controls/outline-control.tsx
```diff
@@ -55,6 +55,45 @@ function isDraggingToMove(
   return dragState != null && dragState?.type === 'MOVE_DRAG_STATE' && targetIsDragged
 }
 
+interface CenteredCrossSVGProps {
+  id: string
+  style?: React.CSSProperties
+}
+
+const CenteredCrossSVG = ({ id, style }: CenteredCrossSVGProps) => {
```
seanparsons, line 63: Should be wrapped in `betterReactMemo`.

seanparsons (approved):

Rheeseyb (approved):

enidemi (approved):

editor/src/components/canvas/controls/outline-control.tsx
```diff
@@ -55,6 +56,48 @@ function isDraggingToMove(
   return dragState != null && dragState?.type === 'MOVE_DRAG_STATE' && targetIsDragged
 }
 
+interface CenteredCrossSVGProps {
+  id: string
+  style?: React.CSSProperties
```
Rheeseyb, line 61: Unless we're intending to pass in custom styling here, this should probably just be replaced with `left` and `top` props

editor/src/components/canvas/controls/outline-control.tsx
maltenuhn: done

editor/src/components/canvas/controls/outline-control.tsx
maltenuhn: done

PR #1657, opened 2021-08-03
Fix the cached require and resolve functions

Rheeseyb:
**Problem:**
We're using incorrectly cached require and resolve functions, since they are based off the `projectContents` but are only updated on a build response from the `ts-worker`, meaning that the canvas will often be rendered using code the last version of the code. Though this problem fixes itself after that build response comes back, it still means that for a period of time after making a code change the editor will be using old code and therefore old metadata.

**Fix:**
Rather than including the project contents in the above mentioned cached functions, instead cache a curried version of those functions, allowing us to provide the latest project contents when calling them.

**Commit Details:**
- Replaced the `requireFn` and `resolve` in `CodeResultCache` with curried versions that take a `ProjectContentTreeRoot`
- Updated all places using these to call them with the `projectContents` (all places already had `projectContents` in scope)
- Removed an now defunct `if` statement in `ui-jsx-canvas.tsx` that should have been removed a long time ago
- Replaced the `resolvedFiles` variable with a ref, since that was making the `useCallback` immediately below pointless (as it was invalidating the dependencies on every render) - this hadn't been previously flagged because of the above mentioned `if` statement
- Added a missing key in `CompileErrorContainer.tsx` since that was generating noise in the console that was getting in the way of working on this
- Added `test-all-except-move-template` so that we can retain the split workflows for CI from #1643 but still be able to use `npm test <filename>`, since that was preventing me running individual tests whilst working on this


enidemi (approved):

editor/src/components/canvas/ui-jsx-canvas.test-utils.tsx
```diff
@@ -156,14 +155,9 @@ export function renderCanvasReturnResultAndError(
...
 
-  const baseRequireFn = getRequireFn(NO_OP, updatedContents, {}, {}, 'canvas')
-  const requireFn: UiJsxCanvasProps['requireFn'] = (importOrigin: string, toImport: string) => {
-    switch (toImport) {
-      // here we can manually insert extra dependencies, such as antd
-      default:
-        return baseRequireFn(importOrigin, toImport)
-    }
-  }
+  const curriedRequireFn = (innerProjectContents: ProjectContentTreeRoot) =>
```
seanparsons, line 158: I don't think I'd refer to this as curried because the result of calling this isn't itself curried. Maybe `requireFnFromProjectContents`?

seanparsons (approved):

editor/src/components/canvas/ui-jsx-canvas.test-utils.tsx
Rheeseyb: Yep that's fair, I'll come back and rename this

PR #1659, opened 2021-08-03
Fix uid retention after parse failure

### LINKED ISSUE #1645: Adding style via code didn’t update layout section of the inspector, but then saving and updating them did

Rheeseyb:
Fixes #1645 

**Problem:**
The logic that allows us to retain UIDs for parsed elements after a code change requires the previous parsed model to supply the lookup, but since we don't store the last parse success we were unable to retain UIDs after a parse failure. This results in selection being lost, and UIDs changing constantly when typing any code that would temporarily result in an error.

**Fix:**
Store the last successful parse of each file, and provide that (or `null` if the file has never been successfully parsed) when parsing the file.


editor/src/core/shared/project-file-types.ts
```diff
@@ -426,18 +426,21 @@ export interface TextFile {
   type: 'TEXT_FILE'
   fileContents: TextFileContents
   lastSavedContents: TextFileContents | null // it is null when the file is saved
+  lastParseSuccess: ParseSuccess | null
```
seanparsons, line 429: Is the only time that this is null right at the very start? Once it has become not null, it should never be null again?

seanparsons (approved):

editor/src/core/shared/project-file-types.ts
Rheeseyb: Yeah, once it has successfully been parsed it shouldn't ever be `null` again

PR #1660, opened 2021-08-04
feature(canvas-controls) flex/grid parent outline

enidemi:
**Problem:**
It is difficult to see how the selected element is layouted.

**Fix:**
Show parent outline for flex/grid elements

**Commit Details:**
- added outline

### DIFF
--- a/editor/src/components/canvas/controls/outline-control.tsx
+++ b/editor/src/components/canvas/controls/outline-control.tsx
@@ -229,7 +229,9 @@ export const OutlineControls = (props: OutlineControlsProps) => {
                 top: parentFrame.y + props.canvasOffset.y,
                 width: parentFrame.width,
                 height: parentFrame.height,
-                outline: `1px dotted ${colorTheme.primary.value}`,
+                outlineStyle: 'dotted',
+                outlineColor: colorTheme.primary.value,
+                outlineWidth: 1 / props.scale,
               }}
             ></div>
           )
@@ -241,6 +243,7 @@ export const OutlineControls = (props: OutlineControlsProps) => {
     props.canvasOffset.x,
     props.canvasOffset.y,
     props.componentMetadata,
+    props.scale,
     props.selectedViews,
   ])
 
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/controls/outline-control.tsx
```diff
@@ -208,6 +208,42 @@ export const OutlineControls = (props: OutlineControlsProps) => {
...
+        if (parentFrame != null) {
+          return (
+            <div
+              style={{
+                position: 'absolute',
+                left: parentFrame.x + props.canvasOffset.x,
+                top: parentFrame.y + props.canvasOffset.y,
+                width: parentFrame.width,
+                height: parentFrame.height,
+                outline: `1px dotted ${colorTheme.primary.value}`,
```
maltenuhn, line 232: I know we're not overly concerned with zoomable canvas here yet, but we do need to find a way to make this work with zoom / scale, otherwise we're restricted to `boxShadow` here (which means no dotted outline), or need differential styling based on canvas zoom values.

I did some digging here. Previously I thought values of < 1px get rounded up to the nearest integer. That is still the case. But apparently if you set `zoom` on the parent, you _can_ use a value of <1px on the outline, and it'll render correctly!

image


For zoom < 1 we can obviously scale this up

enidemi: fixed! 🔍

Rheeseyb (approved):

editor/src/components/canvas/controls/outline-control.tsx
```diff
@@ -208,6 +208,49 @@ export const OutlineControls = (props: OutlineControlsProps) => {
...
+        parentPath,
+      )
+      const parentFrame = MetadataUtils.getFrameInCanvasCoords(parentPath, props.componentMetadata)
+      if (
+        MetadataUtils.isFlexLayoutedContainer(parentElement) ||
+        MetadataUtils.isGridLayoutedContainer(parentElement)
+      ) {
+        if (parentFrame != null) {
+          return (
+            <div
```
Rheeseyb, line 225: This needs a `key`

enidemi: fixed!

PR #1669, opened 2021-08-05
ux(layout section) - legibility and interaction improvements

maltenuhn:
**Problem:**

- the subsection toggles weren't clear enough
- the indicator labels sometimes were incorrect (eg showing presence of children or content when there wasn't any)
- there was a gap in the `flow` width / height row 
- the × button in the layout section also removed width and height. Annoying especially if empty.
- there was a missing icon for the infobox for flex layouts

**Fix:**
- adds new  `InlineToggleButton` and `InlineIndicator` components, uses them in the inspector
- always shows the buttons for the absolute/relative part of height
- makes the × button in the layout section not unset width and height
- uses an available icon for the flex layouts

image

### DIFF
--- a/editor/src/uuiui/inline-button.tsx
+++ b/editor/src/uuiui/inline-button.tsx
@@ -60,9 +60,9 @@ export const InlineIndicator = styled.div<{
 }))
 
 export const InlineToggleButton = styled(InlineButton)<{
-  value: boolean
+  toggleValue: boolean
 }>((props) => ({
-  color: props.value ? colorTheme.primary.value : colorTheme.primary.shade(30).value,
+  color: props.toggleValue ? colorTheme.primary.value : colorTheme.primary.shade(30).value,
   '&:hover': {
     background: colorTheme.primary.shade(5).value,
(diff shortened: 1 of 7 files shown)


Rheeseyb:
Can you add a screenshot to this PR please?

maltenuhn:
Done

editor/src/uuiui/inline-button.tsx
```diff
@@ -47,3 +45,29 @@ export const InlineButton = styled.button({
     outline: 'none',
   },
 })
+
+export const InlineIndicator = styled.div<{
+  value: boolean
```
Rheeseyb, line 50: This `value` should be renamed because it's unclear from the prop name what the value represents / what it does. From my understanding, the `value` is used to determine whether the text inside this styled div should be highlighted to indicate something special about it. I think maybe `shouldIndicate` could work here, or something along those lines?

maltenuhn: fixed

Rheeseyb (approved):

enidemi (approved):

PR #1672, opened 2021-08-05
Multi-option resize.

### LINKED ISSUE #1654: Property target selector should handle multiple types of change.
Shift should cycle through contents in real time, so that one dragging session can update multiple values. I.e. click, shift, drag, shift, drag should update two values.

Deltas applied to sizing should apply to the following values in order of preference:
- Observed value.
- Value from the attributes.
- Zero.

Also, for flex, it should offer `flexBasis`, `flexGrow`, and `flexShrink` on the edge(s) aligned with the flex direction of the parent, and first in the menu

seanparsons:
Fixes #1654

**Problem:**
The resize options are a bit awkward if only one setting can be changed at a time.

Since classes can also be used to set a width of an element, it doesn't make sense to treat an element as having a zero or otherwise width for example and start modifying it using zero as the base value. It would be better to start with the observable dimensions instead if those exist.

**Fix:**
Resizing using the resize options now keeps multiple properties modified in-flight simultaneously.

The observable values for the resize options (where they exist or make sense) are used as the starting value when applying the delta from the mouse movement.

There are also some small fixes which have been done to the mouse handling, like extracting out the resize options into a function which can be invoked at will instead of those values only being attached to components. This was necessary to fix an issue with changing vertical axis properties as the resize options were being set _after_ the mouse down was triggered, that resulted in the wrong property being handled for a brief moment.

**Notes:**
One potential oddity might exist around handling subsequent changes to properties that have already been changed and switched away from. For example modifying the width, then margin-left, then width again. As the original starting point is used each time, changing the margin-left will have put the mouse potentially further along the axis so that when the width is changed again it will "jump" in value.

**Commit Details:**
- Fixes #1654.
- Extracted out `DragStatePositions` as a common interface.
- Created `ResizeDragStatePropertyChange` to hold the specific
  parts of the resize for each property being changed. An array
  of these now lives in `ResizeDragState`.
- Implemented `updatedResizeDragStatePropertyChange` to support
  the changes necessary in `updateResizeDragState`.
- Added `getDragStatePositions`, `getDragStateDrag`,
  `getDragStateStart`, `anyDragStarted`, `anyDragMovement`,
  `getResizeOptions`, `dragExceededThreshold`,
  `getObservableValueForLayoutProp`, `getTargetableProp` and
  `findResizePropertyChange` utility functions.
- Functions like `clearDragState` and `updateFramesOfScenesAndComponents`
  have been reworked using the above functions to handle multiple
  simultaneous changes in a resize.
- Removed the `labels` props that have been used in a bunch of
  React components to use `getResizeOptions` in their place.
- `isResizing`, `isDragging` and `isInserting` rewritten to use
  `getDragStateDrag`.
- `ResizeControl` now handles the differing types of resize.
- Added `flexGrow` and `flexShrink` to `LayoutTargetableProp`.
- Mouse handling logic in the `EditorCanvas` component had a fairly
  major reworking to correctly handle multi-option resize and also
  to fix a first mouse down issue when changing vertical coordinates.

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -20,6 +20,7 @@ import {
   LayoutProp,
   pinnedPropForFramePoint,
   LayoutPinnedProp,
+  LayoutTargetableProp,
 } from '../../core/layout/layout-helpers-new'
 import {
   maybeSwitchLayoutProps,
@@ -143,6 +144,7 @@ import {
   TransientFileState,
   withUnderlyingTarget,
   transformElementAtPath,
+  ResizeOptions,
 } from '../editor/store/editor-state'
 import * as Frame from '../frame'
 import { getImageSizeFromMetadata, MultipliersForImages, scaleImageDimensions } from '../images'
@@ -152,15 +154,19 @@ import Canvas, { TargetSearchType } from './canvas'
 import {
   CanvasFrameAndTarget,
   CSSCursor,
+  DragState,
   DuplicateNewUID,
   EdgePosition,
   flexResizeChange,
   MoveDragState,
   oppositeEdgePositionPart,
+  DragStatePositions,
   pinFrameChange,
   PinOrFlexFrameChange,
   ResizeDragState,
   singleResizeChange,
+  ResizeDragStatePropertyChange,
+  CanvasPositions,
 } from './canvas-types'
 import {
   collectParentAndSiblingGuidelines,
@@ -204,6 +210,7 @@ import { importedFromWhere } from '../editor/import-utils'
 import { Notice } from '../common/notice'
 import { createStylePostActionToast } from '../../core/layout/layout-notice'
 import { uniqToasts } from '../editor/actions/toast-helpers'
+import { LayoutTargetablePropArrayKeepDeepEquality } from '../../utils/deep-equality-instances'
 
 export function getOriginalFrames(
   selectedViews: Array<ElementPath>,
@@ -328,7 +335,11 @@ export function clearDragState(
   applyChanges: boolean,
 ): EditorState {
   let result: EditorState = model
-  if (applyChanges && result.canvas.dragState != null && result.canvas.dragState.drag != null) {
+  if (
+    applyChanges &&
+    result.canvas.dragState != null &&
+    getDragStateDrag(result.canvas.dragState, result.canvas.resizeOptions) != null
+  ) {
     const producedTransientCanvasState = produceCanvasTransientState(
       derived.canvas.transientState.selectedViews,
       result,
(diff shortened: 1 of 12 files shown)


Rheeseyb (commented):
`flexGrow` and `flexShrink` definitely need to be adjusted here so that their unit change is maybe 1/10th the drag delta

seanparsons:
> `flexGrow` and `flexShrink` definitely need to be adjusted here so that their unit change is maybe 1/10th the drag delta

Yeah, that crossed my mind as soon as I tried them, but I think that should go in separately so that this doesn't become a million line PR.

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -2813,3 +2875,163 @@ function createCanvasTransientStateFromProperties(
...
+        default:
+          const _exhaustiveCheck: never = controlDirection
+          throw new Error(`Unhandled control direction ${JSON.stringify(controlDirection)}`)
+      }
+    case null:
+      switch (controlDirection) {
+        case 'horizontal':
+          return ['Height', 'marginTop', 'marginBottom', 'minHeight', 'maxHeight']
+        case 'vertical':
+          return ['Width', 'marginLeft', 'marginRight', 'minWidth', 'maxWidth']
```
Rheeseyb, line 2980: Looks like these should also include `left`, `right`, `top` and `bottom`

Rheeseyb: Also why are the `controlDirection`s inverted?

seanparsons: > Looks like these should also include `left`, `right`, `top` and `bottom`

Why? Those weren't in either the experiment or in #1654.

Rheeseyb:
> > `flexGrow` and `flexShrink` definitely need to be adjusted here so that their unit change is maybe 1/10th the drag delta
> 
> Yeah, that crossed my mind as soon as I tried them, but I think that should go in separately so that this doesn't become a million line PR.

I wonder if we should exclude (or comment out) those props from this PR then, and include them in a follow up PR that also scales the value when updating those two only

editor/src/components/canvas/canvas-utils.ts
seanparsons: > Also why are the `controlDirection`s inverted?

That's how they are seemingly: https://github.com/concrete-utopia/utopia/blob/feature/multi-option-resize/editor/src/components/canvas/controls/size-box.tsx#L617-L618

Rheeseyb: > > Looks like these should also include `left`, `right`, `top` and `bottom`
> 
> Why? Those weren't in either the experiment or in #1654.

Because I can no longer drag to set those values if they've been set and didn't realise that was the case with this menu. Now if I have `left` and `right` set, dragging this control will add `width`, whereas dragging the corner controls will update the correct props.

Rheeseyb (approved):

Rheeseyb:
Discussed elsewhere, and I'm fine with the remaining work coming in a follow up PR, since this work is still an improvement over the current controls in its own right

enidemi (approved):

PR #1673, opened 2021-08-05
Wrap Content Option

### LINKED ISSUE #1670: Inserting a span into an element that only has text content should wrap that text in the inserted span
If you currently have `<div>Hello World</div>` and choose to add a `span` to that via the Navigator, the result right now is:
```
<div>
  Hello World
  <span />
</div>
```
But it should instead be:
```
<div>
  <span>Hello World</span>
</div>
```

Rheeseyb:
Fixes #1670

**Problem:**
When inserting a new `span` into an element that only has text (e.g. `<div>Hello</div>`), typically you'll want to wrap that existing text in the newly inserted element.

**Fix:**
Provide an extra option to "Wrap content" when inserting a new element from the floating menu, and default to that being enabled if the target parent only has text as its content.

**Commit Details:**
- Corrected the behaviour of `getJSXElementFromMetadata` (previously it was returning the element name, rather than the element), and switched existing uses of that function to the other (correctly named) version `getJSXElementNameFromMetadata`
- Added `WrapContentOption` to the `InsertWithDefaults` action, for specifying if the inserted element should wrap the target parent's content
- Added a new check box "Wrap content" to the bottom of the "Add Element" floating menu
- Default the value of that check box to `true` if the target parent only has a single text block child, using the new function `elementOnlyHasSingleTextChild`
- Updated the handler for the `INSERT_WITH_DEFAULTS` to move children from the target parent onto the newly inserted element if the `WrapContentOption` is 'wrap-content'
- Added a test to cover that updated action handling

**Screenshot:**
Screenshot_20210805_173306

### DIFF
--- a/editor/src/components/canvas/controls/render-as.tsx
+++ b/editor/src/components/canvas/controls/render-as.tsx
@@ -24,8 +24,8 @@ export const RenderAsRow = betterReactMemo('RenderAsRow', () => {
 
   const selectedElementName = useEditorState((store) => {
     return MetadataUtils.getJSXElementNameFromMetadata(
-      store.editor.selectedViews[0],
       store.editor.jsxMetadata,
+      store.editor.selectedViews[0],
     )
   }, 'RenderAsRow selectedElementName')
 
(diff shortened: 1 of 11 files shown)


maltenuhn:
Nice! Does this work only with actual text, or also with things like 

- `<div>{props.username}</div>` 
- `<span>{' '}</span>`

enidemi (approved):

PR #1675, opened 2021-08-06
Feature/canvas controls zero sized elements

### LINKED ISSUE #1662: canvas controls for zero sized elements
- show them on parent-selected and cmd down
- default, hover and selected state
- double click to add fixed dimensions

https://utopia.pizza/p/af6126f0-glamorous-jar/

enidemi:
Fixes #1662 

**Problem:**
Canvas controls for zero sized elements are selectable through the navigator only. Clicking on them loses selection. In it's curent form it's an unfinished part of an old experiment.
It's also behind a feature-switch and without even that the zero sized elements are just shown as 1 bottom-right resize point.

**Fix:**
The new version introduces making them highlightable and selectable when the parent is selected and double clicking on them adds fixed dimensions. For this it needed a default and highlighted state and a new resize control to add the fixed dimensions.

**Commit Details:**
- removed the old feature-switch
- added a component that is shown with an unselected state which triggers the highlight on mouseover, this is different from the rest of the canvas controls, that uses the selection and highlight hooks, but it's the quickest way to try these controls first.
- added components for highlighted, selected
- added component to handle the resize double click
- resize double click adds the following props: it adds the missing dimension (flexbasis, width or height) and when both dimensions are missing for non-flex it's adding the position absolute. The absolute positioning is added because for inline elements the size properties are not having an effect on their own. I tried these in an example project and with the added props the element had a non-zero frame this way, it's possible that it needs adjusments.


maltenuhn:
Oh, I'm so curious about trying this!!

editor/src/components/canvas/controls/outline-utils.tsx
```diff
@@ -1,50 +1,10 @@
...
-    return {
-      extraWidth,
-      extraHeight,
-      showingInvisibleElement,
-      borderRadius,
-      dimension,
-    }
+export const ZeroControlSize = 5
+export function isZeroSizedElement(frame: CanvasRectangle | null | undefined): boolean {
+  if (frame == null) {
```
Rheeseyb, line 5: Is this case even possible? Everywhere that we're calling this is also directly reading values from within the `frame`, so if `frame` can be null we should guard against that (I'm bringing this up because IMO it doesn't make sense to return `false` here)

Rheeseyb (approved):

editor/src/components/canvas/controls/outline-utils.tsx
enidemi: fixed

PR #1676, opened 2021-08-06
Scale flex grow / shrink drags

Rheeseyb:
**Problem:**
Dragging to increment the `flexGrow` or `flexShrink` props by 1 per 1px drag feels very wrong

**Fix:**
Scale the applied delta so that it only increments by units of 1 per 100px drag.

I also tried with 10px, 50px, and also incrementing in units of 0.5 rather than 1. Incrementing in whole units rather than 0.5 certainly feels right. The only think I'm not 100% certain on is the 1/100 scaling - it feels right to me, but I'm not sure if to others that will feel like it's too far (especially on your first attempt at updating the `flexGrow` via this). 1/10 is way too fine grained, and 1/50 feels OK but still a bit too fine grained.

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -473,9 +473,15 @@ export function updateFramesOfScenesAndComponents(
             )
             // Defer through these in order: observable value >>> value from attribute >>> 0.
             const currentAttributeToChange = valueFromDOM ?? valueFromAttributes ?? 0
+            const shouldScaleDelta =
+              frameAndTarget.targetProperty === 'flexGrow' ||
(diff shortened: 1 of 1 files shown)


maltenuhn:
> I also tried with 10px, 50px, and also incrementing in units of 0.5 rather than 1. Incrementing in whole units rather than 0.5 certainly feels right. The only think I'm not 100% certain on is the 1/100 scaling - it feels right to me, but I'm not sure if to others that will feel like it's too far (especially on your first attempt at updating the flexGrow via this). 1/10 is way too fine grained, and 1/50 feels OK but still a bit too fine grained.

I guess it'll just come down to discoverability here, since at < 100px there's no way of knowing anything will happen.

enidemi (approved):

PR #1677, opened 2021-08-09
Correctly handle wrapping the root element

### LINKED ISSUE #1638: Wrapping root element of a component doesn’t work
Selecting the root element of `App`, right clicking, selecting "Wrap in" followed by "div" will fail to wrap the selected element in a `div`, and will instead add a new `div` as a child of the `App` instance

Rheeseyb:
Fixes #1638 

**Problem:**
When attempting to use the "Wrap in" options from the floating menu whilst targeting the root element of a component, the result is a new element added as a child of an instance of the component.

**Fix:**
The problem here is that we are blindly using the parent parent, and not checking if that parent path is on the instance. I think we might have some other cases of this problem elsewhere in the code, but for now I've fixed the handling of that parent path here by explicitly checking if the target was the root element of the insertion parent, and if so we use `transformJSXComponentAtPath` to insert a new root element above that.

### DIFF
--- a/editor/src/components/editor/actions/actions.ts
+++ b/editor/src/components/editor/actions/actions.ts
@@ -2094,6 +2094,8 @@ export const UPDATE_FNS = {
     derived: DerivedState,
     dispatch: EditorDispatch,
   ): EditorModel => {
+    // FIXME This and WRAP_IN_ELEMENT are very similar, the only difference being that this attempts to maintain
+    // the positioning. The two handlers should probably be combined, or perhaps have core shared logic extracted
     return toastOnGeneratedElementsSelected(
       `Generated elements can't be wrapped into other elements.`,
       editorForAction,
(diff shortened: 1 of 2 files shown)


editor/src/components/editor/actions/actions.ts
```diff
@@ -2143,7 +2156,9 @@ export const UPDATE_FNS = {
             editor.projectContents,
             editor.nodeModules.files,
             uiFileKey,
-            parentPath,
+            targetThatIsRootElementOfCommonParent != null
+              ? targetThatIsRootElementOfCommonParent
+              : parentPath,
```
balazsbajorics, line 2161: why not 
```suggestion
            targetThatIsRootElementOfCommonParent ?? parentPath,
```

balazsbajorics (approved):

editor/src/components/editor/actions/actions.ts
Rheeseyb: Ha! No reason at all, good catch

PR #1683, opened 2021-08-09
Fix/font size focus

### LINKED ISSUE #1678: Inspector Controls: if there's no font size defined, adding a font size will defocus the inspector control
I think this is the same issue with the background color picker which closes on first click. hopefully I can fix both

balazsbajorics:
Fixes #1678

**Problem:**
If a control, such as the font size control or font color control was "unset" and then we use it to set a value, it would lose focus. Similarly, if it was "set" and we use it to unset content using empty value + enter, it would lose focus.

**Fix:**
Turns out the culprit was `InspectorContextMenuWrapper` which, depending on the context menu length, it would render the props.children directly, or wrapped in a context menu wrapper div. This caused any children to remount (because even if they have a key provided, moving react nodes from one parent to another in the tree causes a remount). This caused the NumberInput (and ColorPicker etc) to remount, which caused it to lose focus.

**Commit Details:**
- Changed `InspectorContextMenuWrapper` so that it always wraps the props.children in a wrapper div, even if the contextMenu is empty
- For the empty context menu, to avoid a buggy behavior (right click triggers an invisible menu, editor state change means the menu now has items, the menu unexpectedly appears on screen), we don't call `show` menu if it's empty

### DIFF
--- a/editor/src/components/context-menu-wrapper.tsx
+++ b/editor/src/components/context-menu-wrapper.tsx
@@ -171,6 +171,7 @@ export class ContextMenuWrapper<T> extends ReactComponent<
           key={`${this.props.id}-provider`}
           id={this.props.id}
           style={this.props.providerStyle}
+          itemsLength={this.props.items.length}
         >
           {this.props.children}
         </MenuProvider>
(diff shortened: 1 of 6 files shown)


editor/src/components/context-menu-wrapper.tsx
```diff
@@ -217,45 +218,45 @@ export class InspectorContextMenuWrapper<T> extends ReactComponent<ContextMenuWr
...
   }
 }
 
 interface MenuProviderProps {
   id: string
+  itemsLength: number
   style?: React.CSSProperties
 }
 
 export const MenuProvider: React.FunctionComponent<MenuProviderProps> = (props) => {
```
Rheeseyb, line 251: I think we could change this to use `PropsWithChildren` here to save having to explicitly pass in the `itemsLength`

Rheeseyb: Oh no, I misunderstood, ignore that

Rheeseyb (approved):

PR #1684, opened 2021-08-09
Fix/feature switch redux measure

### LINKED ISSUE #1550: Put the redux logging behind a feature switch
By default, the feature should be on for development, off for production.

Reason: redux devtool logging takes a significant amount of performance (5-25ms per frame!!!)

### LINKED ISSUE #1556: Put the monkey function performance marks behind a feature switch
https://github.com/concrete-utopia/utopia/blob/340d44d72501a3c40a284a83728415a0455a75ab/editor/src/utils/canvas-react-utils.ts#L167

MeasureRenderTimes should be put behind a feature switch so we can enable it in production (off by default)

balazsbajorics:
Fixes #1550 and Fixes #1556

**Problem:**
Redux devtools logging is very slow. Performance.marks and Performance.measures (in development mode only) are also pretty slow. Bonus problem: sometimes we would like performance.marks in production for debugging!

**Fix:**
Put redux devtools logging and performance.measures behind feature switches.

**Note:**
Because the feature switches are extracted a few moments after the editor loads, I had to move some code around and get the feature switch right before the code would redux log or performance.mark. This shouldn't be a major performance issue because checking the value of a feature switch just returns an object's property.

### DIFF
--- a/editor/src/core/shared/redux-devtools.ts
+++ b/editor/src/core/shared/redux-devtools.ts
@@ -1,5 +1,6 @@
 import type { EditorAction } from '../../components/editor/action-types'
 import type { EditorStore } from '../../components/editor/store/editor-state'
+import { isFeatureEnabled } from '../../utils/feature-switches'
 import { pluck } from './array-utils'
 
 interface Connection {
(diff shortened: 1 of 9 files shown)


editor/src/core/shared/redux-devtools.ts
```diff
@@ -107,25 +112,37 @@ export function reduxDevtoolsSendActions(
 }
 
 export function reduxDevtoolsSendInitialState(newStore: EditorStore): void {
-  if (maybeDevTools != null && sendActionUpdates) {
+  if (maybeDevTools != null) {
```
Rheeseyb, line 115: Oooh is this expensive? If it is, we should have this behind the feature flag and have some way of calling it if the feature flag is later enabled

editor/src/templates/editor.tsx
```diff
@@ -1,3 +1,6 @@
+// import feature switches so they are loaded before anything else can read them
+import '../utils/feature-switches'
```
Rheeseyb, line 2: Why do we need this in both files?

balazsbajorics: I got nervous that for the tests for example editor.tsx is more of an entry point, so putting it here "just in case".

editor/src/core/shared/redux-devtools.ts
balazsbajorics: no, this is actually fine to happen once. it's a couple of ms but only happens once!

PR #1685, opened 2021-08-09
Fix/scaling controls why

### LINKED ISSUE #1598: Flex insert controls shouldn't grow when zooming
**Problem**
The on-canvas insert controls (from the experiment) grow and shrink with zoom.

**Solution Notes**
We have a pattern to address this (via the zoom factor we pass into props / editor state?).

Additionally, I suggest we solve this by passing in the desired _anchor point_ for the controls, and let them handle their own positioning without complex computations. We can accomplish that via `translate (-50% -50%)`  for centering.

balazsbajorics:
Fixes #1598

This PR fixes the scaling issues for the Plus Button Insert control and the floating parent layout control

### DIFF
--- a/editor/src/components/canvas/controls/insertion-plus-button.tsx
+++ b/editor/src/components/canvas/controls/insertion-plus-button.tsx
@@ -14,6 +14,7 @@ const InsertionButtonOffset = 10
 
 interface ButtonControlProps {
   key: string
+  scale: number
   positionX: number
   positionY: number
   lineEndX: number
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/controls/insertion-plus-button.tsx
```diff
@@ -275,6 +278,9 @@ const PlusButton = betterReactMemo('PlusButton', (props: ButtonControlProps) =>
     },
     [dispatch, parentPath, indexPosition],
   )
+
+  const ButtonSize = 12
```
Rheeseyb, line 282: Was this supposed to be scaled too?

balazsbajorics: nah, I'm using zoom and transorm: scale in this one because of the text

Rheeseyb (approved):

PR #1687, opened 2021-08-10
Ux/inspector sections

maltenuhn:
**Problem:**
The inspector had become hard to (visually) navigate, especially with the varying length and visibility of sections / subsections.

**Fix:**
- Remove clutter by moving the right menu to the top menu
- Add icons for sections / subsections
- Create visually consistent sections / subsection headers

image


Rheeseyb:
Wrong icon for the border section?
Screenshot_20210810_114327

balazsbajorics (approved):

editor/src/uuiui/icons.tsx
```diff
@@ -70,6 +70,79 @@ export const LargerIcons = {
...
+  Background: makeIcon({
+    category: 'inspector',
+    type: 'background',
+    color: 'main',
+    width: 16,
+    height: 16,
+  }),
+  Border: makeIcon({
+    category: 'inspector',
+    type: 'layout',
```
Rheeseyb, line 104: This would be the reason for the incorrect border icon

balazsbajorics: 🧐

maltenuhn: how peculiar!

Rheeseyb:
This appears to have pushed the render count tests up quite a lot :thinking:

maltenuhn:
> This appears to have pushed the render count tests up quite a lot 🤔

Yeah, but why? If anything, I _removed_ some things that were in the right menu before...

Rheeseyb:
> > This appears to have pushed the render count tests up quite a lot thinking
> 
> Yeah, but why? If anything, I _removed_ some things that were in the right menu before...

At the least I can see that we weren't previously re-rendering the right menu when selection changed, whereas we are now re-rendering those buttons since moving them to the top bar, so fixing that should drop the render count. I guess the addition of the icons will push the count up a little too, but that part is fine.

Rheeseyb (approved):

PR #1688, opened 2021-08-10
Use tab / shift+tab to change properties in the resize controls

### LINKED ISSUE #1674: Shift property target selector with 'tab' instead of 'shift'.

Rheeseyb:
Fixes #1674

**Problem:**
The resize controls that all you to choose a target property use the shift key to change the property, whereas they should use tab instead

**Fix:**
Use a window event listener to support tab to increment the selected property index, or shift+tab to decrement it. In making this change I discovered that for `window.removeEventListener` to actually remove an event listener, it needs to be provided the same options as the corresponding `window.addEventListener` was, so I fixed the other places in the editor that weren't previously doing that.


balazsbajorics (approved):
hah I was surprised why we didn't have a DECREMENT before, but then I realized the shortcut would have been Shift + Shift 😂

PR #1695, opened 2021-08-11
Remove ElementInstanceMetadata.rootElements

balazsbajorics:
**Problem:**
ElementInstanceMetadata.rootElements can only be collected during the dom-walker phase. For renderers which only support the Spy (such as react-three-fiber), this means the user can't use the 'focus component' feature to drill into components.

But it turns out that we don't even _need_ the `rootElements` field! The metadata map's keys are giving us enough information.

**Fix:**
Use the metadata map to find root elements. Delete the unused and unnecessary rootElements field


editor/src/components/canvas/dom-walker.ts
```diff
@@ -610,11 +609,10 @@ function collectMetadata(
       const unfilteredChildrenPaths = allUnfilteredChildrenPaths.concat(rootsOrChildrenToAdd)
 
       let filteredChildPaths: ElementPath[] = []
-      let filteredRootElements: ElementPath[] = []
       fastForEach(unfilteredChildrenPaths, (childPath) => {
         if (EP.isParentOf(path, childPath)) {
           if (EP.isRootElementOfInstance(childPath)) {
-            filteredRootElements.push(childPath)
+            // Delete this when killing child paths
```
seanparsons, line 615: Is this something that needs fixing imminently or for functionality we don't currently have?

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -503,8 +503,11 @@ export const MetadataUtils = {
     })
   },
   getRootViewPaths(elements: ElementInstanceMetadataMap, target: ElementPath): Array<ElementPath> {
-    const element = MetadataUtils.findElementByElementPath(elements, target)
-    return element?.rootElements ?? []
+    const allPahtsUnordered = Object.keys(elements).map(EP.fromString)
```
seanparsons, line 506:
```suggestion
    const allPathsUnordered = Object.keys(elements).map(EP.fromString)
```

editor/src/components/canvas/dom-walker.ts
balazsbajorics: so this is just an empty IF branch, which I should have flipped. However, (spoilers) I plan to do the same thing to ElementInstanceMetadata.children, which means I'll delete this entire block

seanparsons (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -503,8 +503,11 @@ export const MetadataUtils = {
     })
   },
   getRootViewPaths(elements: ElementInstanceMetadataMap, target: ElementPath): Array<ElementPath> {
-    const element = MetadataUtils.findElementByElementPath(elements, target)
-    return element?.rootElements ?? []
+    const allPathsUnordered = Object.keys(elements).map(EP.fromString)
+    const possibleRootElementsOfTarget = allPathsUnordered.filter((path) =>
+      EP.isRootElementOf(path, target),
+    )
```
Rheeseyb, line 509: I fear that this is particularly expensive, especially on larger projects, and is now called very often. I'm not sure if this is something we should try to address via some additional caching, or by relying on the string representation (which honestly makes me feel very uneasy)

Rheeseyb (approved):

PR #1700, opened 2021-08-12
Don't attach data-* props to non-html intrinsic elements wrapped in fragments

### LINKED ISSUE #1698: Wrapping a `mesh` in a fragment results in it receiving a `data-uid`
We have logic in place for preventing utopia from adding internal `data-` props to non-HTML intrinsic elements, which is missed when wrapping those elements in a fragment.

Rheeseyb:
Fixes #1698 

**Problem:**
We explicitly exclude our internal `data-*` props when working with intrinsic elements that aren't from a specific set of HTML elements, as we found that clashes with some logic in R3F (there is logic around providing a way to set a nested prop by using dashes), which was originally fixed [here](https://github.com/concrete-utopia/utopia/pull/346). However, there was an edge case that we were missing when dealing with fragments.

**Fix:**
Patch that edge case and add a test.

### DIFF
--- a/editor/src/utils/canvas-react-utils.spec.tsx
+++ b/editor/src/utils/canvas-react-utils.spec.tsx
@@ -638,4 +638,30 @@ describe('Monkey Function', () => {
       "
     `)
   })
+
+  it('Does not add a uid to a non-html intrinsic element', () => {
+    const Component = () => {
+      return <mesh />
+    }
+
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):

seanparsons (commented):
:+1:

PR #1705, opened 2021-08-13
Turn canvas renderer contexts into context-like atoms

balazsbajorics:
**Problem:**
React contexts are lost if the code is spanning multiple renderers. Most notably they don't work in react-three-fiber unless you are manually bridging the context value. 

**Fix:**
Do not use react contexts to pass down data. Instead use custom Atoms that do not rely on react contexts to work. For our use case (single editor instance, single canvas instance) they are equivalent. 

**Commit Details:**
- Added a Provider component to atom-with-pub-sub. This makes it easy to upgrade the code without messy PR
- Changing all of the contexts used by the canvas renderer into these context-like atoms
- I did not simplify any of the contexts, to make the diff easier to comprehend. There's a 1-1 mapping from Context -> CtxAtom 
- I deleted two contexts that were not used anymore: MutableUtopiaContext, ParentLevelUtopiaContext


editor/src/core/shared/atom-with-pub-sub.tsx
```diff
@@ -17,6 +26,16 @@ export function atomWithPubSub<T>(options: { key: string; defaultValue: T }): At
   const newAtom: AtomWithPubSub<T> = {
     key: key,
     currentValue: defaultValue,
+    Provider: ({ children, value }) => {
+      const updateAtomSynchronously = usePubSubAtomWriteOnlyInner(newAtom, 'sync')
+      newAtom.currentValue = value // TODO this is sneaky and we should use API instead
+      useIsomorphicLayoutEffect(() => {
+        unstable_batchedUpdates(() => {
```
Rheeseyb, line 33: We might be able to bin this off with React 18 :crossed_fingers: https://github.com/reactwg/react-18/discussions/21

editor/src/core/shared/atom-with-pub-sub.tsx
```diff
@@ -46,12 +65,23 @@ export function useSubscribeToPubSubAtom<T>(
 
 export function usePubSubAtomReadOnly<T>(atom: AtomWithPubSub<T>): T {
   const forceUpdate = useForceUpdate()
-  useSubscribeToPubSubAtom(atom, React.useCallback(forceUpdate, [forceUpdate]))
+  const previousValueRef = React.useRef(atom.currentValue)
+  useSubscribeToPubSubAtom(
+    atom,
+    React.useCallback(() => {
+      // TODO only forceUpdate if the last returned atomValue does not equal the current one
```
Rheeseyb, line 72: Is that not what this is already doing?

Rheeseyb (approved):

editor/src/core/shared/atom-with-pub-sub.tsx
balazsbajorics: oops

balazsbajorics, line 72:
```suggestion
```

editor/src/core/shared/atom-with-pub-sub.tsx
```diff
@@ -1,10 +1,19 @@
...
 import * as React from 'react'
+import { unstable_batchedUpdates } from 'react-dom'
 import { useForceUpdate } from '../../components/editor/hook-utils'
 
+// From https://github.com/dai-shi/use-context-selector/blob/2dd334d727fc3b4cbadf7876b6ce64e0c633fd25/src/index.ts#L25
+const isSSR =
+  typeof window === 'undefined' ||
+  /ServerSideRendering/.test(window.navigator && window.navigator.userAgent)
+
+export const useIsomorphicLayoutEffect = isSSR ? React.useEffect : React.useLayoutEffect
```
seanparsons, line 11: `Isomorphic`? I aint letting that one get past me!

seanparsons (approved):

editor/src/core/shared/atom-with-pub-sub.tsx
balazsbajorics: sorry Sean. you gotta let this go, it is _the_ common way to call react code that renders both on the backend and frontend isomorphic.

PR #1710, opened 2021-08-13
Pass through refs in spy wrapped elements

### LINKED ISSUE #1707: Focused components don't correctly forward refs
If a component uses a ref, when it is focused on the canvas the ref will be used. e.g. in the below code:
```

const Clicky = () => {
  const divRef = React.useRef()

  const onMouseDown = React.useCallback(() => {
    console.log(`ref: `, divRef)
  }, [divRef])

  return (
    <div ref={divRef} onMouseDown={onMouseDown}>
      Clicky
    </div>
  )
}

export var App = () => {
  return (
    <div
      style={{
        width: '100%',
        height: '100%',
        backgroundColor: '#FFFFFF',
        position: 'relative',
      }}
    >
      <Clicky />
    </div>
  )
}
```
Clicking the text will log out a ref for the `div`, but focusing the `Clicky` component and trying again will show that the ref is now `null`

Rheeseyb:
Fixes #1707 

**Problem:**
When spy wrapping an element, the `SpyWrapper` wasn't forwarding refs

**Fix:**
Use a `React.forwardRef` to capture and pass on the refs

### DIFF
--- a/editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
+++ b/editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
@@ -22617,6 +22617,7 @@ Object {
     "props": Object {
       "data-paths": "utopia-storyboard-uid/scene-aaa/app-entity:aaa utopia-storyboard-uid/scene-aaa/app-entity",
       "data-uid": "aaa app-entity",
+      "ref": [Function],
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):
propsThroughProps

Rheeseyb:
> propsThroughProps

passThroughPropsFromPropsPassedThroughThisFar

PR #1712, opened 2021-08-16
project contents json import field

balazsbajorics:
**Problem:**
It's very hard for me to move a project between pizza to localhost to production. It's also hard to pass projects to each other.

**Fix:**
The dead simple solution for v1: just an input field where you can paste the project contents json. The JSON can be copied from
`https://utopia.app/v1/project/<id>/contents.json`

image

It's an unstyled input field right above the Feature Toggles UI

you paste in the copied json, and press enter – voila

**Note:**
Does not support assets.

### DIFF
--- a/editor/src/components/inspector/sections/settings-panel/inspector-settingspanel.tsx
+++ b/editor/src/components/inspector/sections/settings-panel/inspector-settingspanel.tsx
@@ -113,16 +113,21 @@ export const SettingsPanel = betterReactMemo('SettingsPanel', () => {
 
   const loadProjectContentJson = React.useCallback(
     (value: string) => {
-      const persistentModel = json5.parse(value)
-      console.info('attempting to load new Project Contents JSON', persistentModel)
-      load(
-        dispatch,
-        persistentModel,
-        entireStateRef.current.editor.projectName,
-        entireStateRef.current.editor.id!,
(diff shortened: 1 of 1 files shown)


enidemi (approved):

Rheeseyb (changes requested):
I'm very wary of the fact that this will blow away the existing project. Please can you hide this behind a feature flag please. It might also be worth showing a warning dialog that requires clicking "Ok" to continue.

balazsbajorics:
> I'm very wary of the fact that this will blow away the existing project. Please can you hide this behind a feature flag please. It might also be worth showing a warning dialog that requires clicking "Ok" to continue.

I added a dumb confirm dialog.

I am not putting it behind a feature switch for now, because it is already almost impossible to discover this feature without knowing about this.

Rheeseyb:
> > I'm very wary of the fact that this will blow away the existing project. Please can you hide this behind a feature flag please. It might also be worth showing a warning dialog that requires clicking "Ok" to continue.
> 
> I added a dumb confirm dialog.
> 
> I am not putting it behind a feature switch for now, because it is already almost impossible to discover this feature without knowing about this.

That's good enough, since it fails if you just give it an empty JSON object anyway!

Rheeseyb (approved):

Issue #1715, opened 2021-08-16
Top level JSX elements are being incorrectly parsed as components


Rheeseyb:
e.g. `const A = <div/>` is being parsed as a component, meaning we cannot later use it like `<div>{A}</div>`

PR #1716, opened 2021-08-16
Switch to new react jsx transform

Rheeseyb:
**Problem:**
As it stands, the new React JSX transform (introduced [here](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)) isn't supported in Utopia when running projects in the preview.

**Fix:**
Update the `target` and `jsx` options of the TS compiler options that use in the `ts-worker.ts` for bundling projects to `esnext` and `react-jsx` accordingly. For the sake of the canvas, we still use the older transform, but ensure that `React` is always in scope. The reason for this is that to using the new transform [here](https://github.com/concrete-utopia/utopia/blob/eadacef7c25fa3b7247319badf112da8ede8bc78/editor/src/core/workers/parser-printer/parser-printer-transpiling.ts#L280) causes issues when transpiling blocks of code (because of the insertion a new `require` statement). This still needs addressing.

Another point to note - I've not been able to test this with top level JSX elements, since those are currently being incorrectly parsed as components. I will revisit that as part of #1715


enidemi (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx
```diff
@@ -97,6 +97,7 @@ export function createExecutionScope(
   }
 
   let executionScope: MapLike<any> = {
+    React: React,
```
seanparsons, line 100: Does this mean it's effectively imported even if it's not?

seanparsons (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx
Rheeseyb: For the sake of the Canvas, yeah, because the Canvas is still being forced to use the old style transform. I'm going to add a follow up ticket for this actually, because really this PR is doing part of the work and then using a workaround for the rest.

Issue #1718, opened 2021-08-16
Internally a lot of code relies on the storyboard element being treated as a component


Rheeseyb:
The interface [`UtopiaJSXComponent`](https://github.com/concrete-utopia/utopia/blob/43aaae60fa2542e642f8b5183434aaa60631d875/editor/src/core/shared/element-template.ts#L1264) has a field `isFunction`, which when false is supposed to indicate that the parsed thing is a top level element (e.g. `const a = <div/>`) rather than a component (despite the name of the interface). However, when creating the canvas's execution context, [it is converted into a component](https://github.com/concrete-utopia/utopia/blob/43aaae60fa2542e642f8b5183434aaa60631d875/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx#L117). Solving this isn't as simple as changing how we insert that into the execution scope if it has `isFunction` set to false, because there may be dependencies between it and surrounding arbitrary JS (which will be executed [a few lines down via the `combinedTopLevelArbitraryBlock`](https://github.com/concrete-utopia/utopia/blob/43aaae60fa2542e642f8b5183434aaa60631d875/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx#L159)).

To add further complications here, it seems that whilst the `storyboard` is defined as a top level element in the default project, internally there is a lot of code assuming that it is a component, thereby implicitly relying on the above bug.

maltenuhn:
@Rheeseyb is this still an issue?

balazsbajorics:
@maltenuhn I think it is still relevant, yes

Rheeseyb:
Yeah, this is still relevant

PR #1719, opened 2021-08-16
Workaround to support top level JSX variables

### LINKED ISSUE #1715: Top level JSX elements are being incorrectly parsed as components

Rheeseyb:
Fixes #1715 

**Problem:**
Top level JSX variables (e.g. `const a = <div/>`) are being incorrectly converted into components during execution in the Canvas

**Fix:**
The actual underlying issue is covered in more depth in #1718. As a work around here, I have put in a check to stop us parsing a top level JSX variable as a `UtopiaJSXComponent` unless it is the `storyboard`, since we rely on that being a `UtopiaJSXComponent`. Any other top level variables will be treated as an arbitrary JS block.

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer.spec.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer.spec.ts
@@ -2632,7 +2632,7 @@ import {
   Scene
 } from "utopia-api";
 import { cake } from 'cake'
-export var whatever = <View data-uid='aaa'>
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):
😬

seanparsons (approved):

PR #1723, opened 2021-08-17
Fix/remove metadata children

### LINKED ISSUE #1711: Remove ElementInstanceMetadata.children so that we don't need to rely on the dom-walker for generated children
do the same as https://github.com/concrete-utopia/utopia/pull/1695
but for children

balazsbajorics:
Fixes #1711 

**Problem:**
In the spy-wrapper, ElementInstanceMetadata.children is generated _before_ rendering happens, therefore it doesn't contain generated elements. The Dom-walker runs _after_ the render, and we use the dom walker to patch up the data in ElementInstanceMetadata.children. However, the dom-walker cannot run for certain renderers, such as react-three-fiber.

Futhermore, the original shape of ElementInstanceMetadata was a tree where children were `Array<ElementInstanceMetadata>`. We since moved to a flat map structure, where the key is the full ElementPath, and `children` now is `Array<ElementPath>`. But the keys of the metadata map also hold sufficient information to find the children of a given ElementPath, rendering `ElementInstanceMetadata.children` redundant.

So this information is both unreliable (for things that cannot be dom-walked) and redundant. Let's kill it!

**Fix:**
Remove `ElementInstanceMetadata.children`. To get the children element paths, use `MetadataUtils.getChildrenPaths()`.

**Commit Details:**
- MetadataUtils.getChildrenPaths now does the same as `getRootViewPaths`, uses the keys to find children.
- I added an extra check `!EP.isRootElementOfInstance(elementPath)` so that we don't list root views as children.
- Updating snapshots

Before:
image

After:
image
(notice the generated Boxes showing up in the navigator)


enidemi (approved):

balazsbajorics:
the tests now pass, and I think I reduced the brittleness of one test too

editor/src/components/canvas/dom-walker.ts
```diff
@@ -603,29 +602,13 @@ function collectMetadata(
...
-      const rootsOrChildrenToAdd = pathsForElement.filter((otherPath) =>
-        EP.isParentOf(path, otherPath),
-      )
-      const unfilteredChildrenPaths = allUnfilteredChildrenPaths.concat(rootsOrChildrenToAdd)
-
-      let filteredChildPaths: ElementPath[] = []
-      fastForEach(unfilteredChildrenPaths, (childPath) => {
-        if (EP.isParentOf(path, childPath)) {
-          if (EP.isRootElementOfInstance(childPath)) {
-            // Delete this when killing child paths
```
Rheeseyb, line 615: Woo hoo!

editor/src/components/canvas/move-template.spec.browser.tsx
```diff
@@ -525,11 +525,11 @@ describe('moveTemplate', () => {
       makeTestProjectCodeWithSnippet(`
         <View style={{ ...props.style }} data-uid='aaa'>
           <View
-            style={{ backgroundColor: '#0091FFAA', position: 'relative', left: 52, top: 61, width: 256, height: 202 }}
+            style={{ backgroundColor: '#0091FFAA', position: 'absolute', left: 52, top: 61, width: 256, height: 202 }}
             data-uid='bbb'
           >
             <View data-uid='ccc'>
-              <View data-uid='ddd' />
+              <View data-uid='ddd'  style={{ backgroundColor: '#0091FFAA', position: 'absolute', left: 10, top: 10, width: 100, height: 100 }} />
```
Rheeseyb, line 532: This looks suspicious. Why have these tests changed?

Rheeseyb: Oh was this the general problematic tests that were occasionally failing on master?

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1042,26 +1055,14 @@ export const MetadataUtils = {
 
       // Checking if our elements support children should prevent us from ending up with the
       // internals of draft-js showing up underneath Text elements.
+      // TODO is data-utopia-do-not-traverse still relevant?
```
Rheeseyb, line 1058: This prop mainly existed to prevent us digging into DraftJS, but I'm pretty sure that's no longer an issue.

Rheeseyb (approved):

PR #1725, opened 2021-08-17
Re-enable the use of ReactDOM.render in the Canvas

### LINKED ISSUE #1722: The `Html` elements are missing from this R3F project
In [this R3F project](https://utopia.pizza/p/5db4ee87-wandering-sorrel/), copied from the "HTML annotations" sample project, the `Html` elements aren't being rendered on the Canvas, but are being rendered in the embedded Preview

Rheeseyb:
Fixes #1722 

**Problem:**
We replace `ReactDOM.render` in the canvas with a no-op. The reason for this was so that nobody would unintentionally break the editor by replacing the root element, but by doing this it means legitimate use of `ReactDOM.render` is not possible.

**Fix:**
Remove the restriction, and make the root element's ID more explicitly to guard against it being accidentally replaced.

**Commit Details:**
- Updated the root element's ID from `'editor'` to `'utopia-editor-root'` in `utils.ts`
- Removed the `SafeReactDOM` function from `built-in-dependencies-list.ts`
- Removed all threading through of the mode (preview or canvas) that was previously required to support the above

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas.test-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas.test-utils.tsx
@@ -159,7 +159,7 @@ export function renderCanvasReturnResultAndError(
   const updatedContents = contentsToTree(projectContents)
 
   const curriedRequireFn = (innerProjectContents: ProjectContentTreeRoot) =>
-    getRequireFn(NO_OP, innerProjectContents, {}, {}, 'canvas')
+    getRequireFn(NO_OP, innerProjectContents, {}, {})
 
   storeHookForTest.updateStore((store) => {
     const updatedEditor = {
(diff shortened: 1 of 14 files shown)


balazsbajorics (changes requested):
I think you missed the actual entry point div's id in index.html

https://github.com/concrete-utopia/utopia/blob/b54495cd1acce119f0274bb4dba7df482b000bd3/editor/src/templates/index.html#L66

balazsbajorics (approved):
much better

Rheeseyb:
> I think you missed the actual entry point div's id in index.html
> 
> https://github.com/concrete-utopia/utopia/blob/b54495cd1acce119f0274bb4dba7df482b000bd3/editor/src/templates/index.html#L66

Hahaha yeah I did indeed!

enidemi (approved):

PR #1726, opened 2021-08-17
feature(inspector) vector controls for component section

### LINKED ISSUE #1720: Inspector custom controls: vector3, vector2

enidemi:
Fixes #1720

**Problem:**
Component controls for vectors take too much space in the inspector. 

**Fix:**
Making the inspector more compact with special controls for vectors.

**Commit Details:**
- added new row with vector controls
- new type of property control for vector2 and vector3
- parser for vector types

**Screenshot**
Screenshot 2021-08-18 at 15 12 07

### DIFF
--- a/editor/src/components/inspector/sections/component-section/component-section.tsx
+++ b/editor/src/components/inspector/sections/component-section/component-section.tsx
@@ -49,6 +49,9 @@ import {
   Subdued,
   VerySubdued,
   FlexRow,
+  useWrappedEmptyOrUnknownOnSubmitValue,
+  ChainedNumberInput,
+  NumberInputProps,
 } from '../../../../uuiui'
 import { getControlStyles } from '../../../../uuiui-deps'
 import { InfoBox } from '../../../common/notices'
@@ -101,6 +104,7 @@ import {
   isAnimatedElement,
   isImportedComponentNPM,
 } from '../../../../core/model/project-file-utils'
+import { CSSNumber, cssNumber, printCSSNumber } from '../../common/css-utils'
 
 function useComponentPropsInspectorInfo(
   partialPath: PropertyPath,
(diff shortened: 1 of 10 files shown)


editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -496,6 +497,81 @@ const RowForUnionControl = betterReactMemo(
...
+              }, controlDescription.object)}
+            </UIGridRow>
+          </UIGridRow>
+        </InspectorContextMenuWrapper>
+      </>
+    )
+  },
+)
+
+function isVectorControlDescription(controlDescription: ObjectControlDescription): boolean {
```
balazsbajorics, line 548: have you considered adding dedicated Vector3 (and vector2) control types? (adding it here https://github.com/concrete-utopia/utopia/blob/b54495cd1acce119f0274bb4dba7df482b000bd3/utopia-api/src/property-controls/property-controls.ts#L67 )

I think it's less confusing if it's explicit, but I am curious what you and others think

editor/src/core/property-controls/third-party-property-controls/react-three-fiber-controls.ts
```diff
@@ -1,9 +1,15 @@
...
+  Vector3ControlDescription,
+  PropertyControls,
+} from 'utopia-api'
 
-const Vector3: ObjectControlDescription = {
-  type: 'object',
-  object: {
+const Vector3: Vector3ControlDescription = {
+  type: 'vector3',
+  controls: {
```
balazsbajorics, line 12: hmm, this controls field is very verbose for something which is user facing. I think it's fine if vector3 is more opinionated than this.
can the controls even be not numbers? 

Should the labels be user-settable? Or is it always x, y, z?

If it should be user configurable, then I this is how it should look like:
```javascript
const Vector3: Vector3ControlDescription = {
   type: 'vector3',
   labels: ['x', 'y', 'z'],
}
```

This should not be a "higher level" control, because it can _only_ be three numbers. Now it looks like the user can change this vector to something which has a different dimension or type – which I believe the inspector does not support because it uses the NumberControls

utopia-api/src/property-controls/property-controls.ts
```diff
@@ -107,6 +109,13 @@ export interface StyleObjectControlDescription
   defaultValue?: CSSProperties
   placeholder?: CSSProperties
 }
+export interface Vector2ControlDescription extends AbstractBaseControlDescription<'vector2'> {
+  defaultValue?: unknown
```
Rheeseyb, line 113: These default values should probably be `[number, number]` / `[number, number, number]` or `Array<number>`

balazsbajorics: avoid Array<number> if possible!

editor/src/components/inspector/sections/component-section/property-control-controls.tsx
```diff
@@ -381,3 +394,50 @@ export const ControlForStringProp = betterReactMemo(
...
+
+export const ControlForVectorProp = betterReactMemo(
+  'ControlForVectorProp',
+  (props: ControlForPropProps<Vector2ControlDescription | Vector3ControlDescription>) => {
+    const { propPath, propMetadata, controlDescription } = props
+
+    const vectorValue =
+      (propMetadata.propertyStatus.set ? propMetadata.value : controlDescription.defaultValue) ?? []
+
+    const keys = controlDescription.type === 'vector3' ? ['x', 'y', 'z'] : ['x', 'y']
```
balazsbajorics, line 406: @enidemi how confident are you that vector3 (in context of react-three-fiber) can only be "x, y, z"?

enidemi: This is what I found in the three js docs https://threejs.org/docs/#api/en/math/Vector3

utopia-api/src/property-controls/property-controls.ts
enidemi: fixed!

editor/src/components/inspector/sections/component-section/property-control-controls.tsx
balazsbajorics: nice! can you add this as a comment to the Vector3 interface in property-controls?

maltenuhn: For vector3 I've seen some samples that use their third definition ("Any arbitrary ordered triplet of numbers"), but the vast majority are xyz. vector2 seems to be the more open-ended one but even there it's prob 80/20.

in conclusion: as long as we don't _force_ these for labels, I'd be ok _defaulting_ to them for labels. Does that work with our property controls?

Rheeseyb: That would require changing the definition to include labels and provide defaults to them. Looking at that definition (and the one of `Vector2`), they even use x,y,z as the constructor param names.

Rheeseyb (approved):

balazsbajorics (approved):

PR #1727, opened 2021-08-17
Handle user components called Scene.

### LINKED ISSUE #1717: A user component named `Scene` is being swapped out for our internal `Scene` component.

seanparsons:
Fixes #1717

**Problem:**
If a user creates a component called `Scene` our canvas logic replaces it with our internal scene component which results in the canvas throwing an exception.

**Fix:**
Replaced the logic in a few places which relied on (in a lot of cases) just checking if the element name was `Scene`. For the canvas this means checking the import information related to an element and double checking the value imported and in scope.

**Commit Details:**
- Fixes #1717.
- Modified `importedFromWhere` to include the exported name of the
  value being imported if there is one.
- `renderJSXElement` uses the new information provided by `importedFromWhere`
  to identify the `Scene` component from `utopia-api`.
- `lookupElementImport` now also uses the new information from `importedFromWhere`.
- `INSERT_SCENE` action now uses `getNewSceneName` instead of the
  retired `getNumberOfScenes` function to get the new name.
- Deleted `getNumberOfScenes`, `getSceneElements` and
  `getSceneElementsFromParseSuccess`.
- Removed a couple of unused bits of selector code from `name-and-icon-hook.ts`.
- `isProbablySceneFromMetadata` now relies on the `importInfo` field
  to identify if it is a `Scene`.

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -132,7 +132,6 @@ import {
   transientFileState,
   getStoryboardElementPathFromEditorState,
   addSceneToJSXComponents,
-  getNumberOfScenes,
   StoryboardFilePath,
   modifyUnderlyingTarget,
   modifyParseSuccessAtPath,
@@ -2700,6 +2699,7 @@ export function getValidElementPaths(
     transientFilesState,
   )
   const importSource = importedFromWhere(filePath, topLevelElementName, topLevelElements, imports)
+    ?.filePath
   if (importSource != null) {
     const resolvedImportSource = resolve(filePath, importSource)
     if (isRight(resolvedImportSource)) {
--- a/editor/src/components/canvas/ui-jsx-canvas-bugs.spec.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-bugs.spec.tsx
@@ -228,7 +228,7 @@ export default function App(props) {
         <div
           id=\\"canvas-container\\"
           style=\\"position: absolute;\\"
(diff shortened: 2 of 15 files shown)


balazsbajorics (approved):

enidemi (approved):

Rheeseyb (approved):
There is still a residual issue around the auto-focusing behaviour with scenes, e.g. in [this project](https://utopia.pizza/p/cb178a13-glossy-utahceratops/?branch_name=fix-handle-user-components-called-scene). I'm not sure if that will cause issues for user defined `Scene` components, but worth checking to see if that needs resolving

seanparsons:
> There is still a residual issue around the auto-focusing behaviour with scenes, e.g. in [this project](https://utopia.pizza/p/cb178a13-glossy-utahceratops/?branch_name=fix-handle-user-components-called-scene). I'm not sure if that will cause issues for user defined `Scene` components, but worth checking to see if that needs resolving

I'm confused, as in that's an already existing issue we know about?

Rheeseyb:
> > There is still a residual issue around the auto-focusing behaviour with scenes, e.g. in [this project](https://utopia.pizza/p/cb178a13-glossy-utahceratops/?branch_name=fix-handle-user-components-called-scene). I'm not sure if that will cause issues for user defined `Scene` components, but worth checking to see if that needs resolving
> 
> I'm confused, as in that's an already existing issue we know about?

As in that behaviour appears to still be assuming anything named `Scene` is a Utopia Scene, and therefore is treating it as being focused by default.

seanparsons:
> As in that behaviour appears to still be assuming anything named `Scene` is a Utopia Scene, and therefore is treating it as being focused by default.

Fixed in c049faf0.

Rheeseyb (approved):

PR #1738, opened 2021-08-18
Handle class exports.

### LINKED ISSUE #1713: Parser is broken for`export class Cica` and `export default class Cica`
`export class Cica` and `export default class Cica` do not show up on `exportsDetail` . 
This means the customRequireFn inside ui-jsx-canvas will return a broken import

seanparsons:
Fixes #1713

**Problem:**
If a class component is exported like `export class App` or `export default class App` it doesn't get exposed appropriately so that the canvas will supply the values when imported.

**Fix:**
Include class declarations in the pass which collates the exports, as they're flagged against class declarations with a modifier rather than export declarations.

**Commit Details:**
- Fixes #1713.
- `parseCode` now handles class declarations in the export details
  by checking if the declaration has the export keyword attached to it.

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer.ts
@@ -1177,8 +1177,8 @@ export function parseCode(
         pushUnparsedCode(prefixedCode)
       }
 
-      // Handle export assignments: `export default App`
       if (TS.isExportAssignment(topLevelElement)) {
+        // Handle export assignments: `export default App`
         const fromAssignment = detailsFromExportAssignment(sourceFile, topLevelElement)
         // Parsed it fully, so it can be incorporated.
         forEachRight(fromAssignment, (toMerge) => {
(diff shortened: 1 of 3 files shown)


Rheeseyb (approved):

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -1201,8 +1201,34 @@ export function parseCode(
...
           pushArbitraryNode(exportDeclaration)
         })
-        // Handle imports.
+      } else if (
+        TS.isClassDeclaration(topLevelElement) &&
+        topLevelElement.modifiers != null &&
+        topLevelElement.modifiers.some((modifier) => modifier.kind === TS.SyntaxKind.ExportKeyword)
+      ) {
+        // Handle classes.
+        // Check if this is the 'default' export, which is to say the entire export for the file.
```
balazsbajorics, line 1210: warning, helpful comment is not completely right: in the ESModules (this is how our ts code is compiled too), a module can have named exports -and- default exports simultaneously. The default export is attached under the name `default`...

balazsbajorics (approved):

PR #1746, opened 2021-08-19
r3f controls should work if react-three-fiber is imported

balazsbajorics:
**Problem:**
We only apply the propertyControls for react-three-fiber host elements if @react-three/fiber is in the package.json. But some older projects use the `react-three-fiber` package instead of the `@react-three/fiber` package. We should support both.

**Fix:**
Support both


Rheeseyb (approved):
This makes me uneasy because it assumes backwards compatibility of all property controls, but it's better than the alternative for the time being.

enidemi (approved):

PR #1750, opened 2021-08-19
navigator resize fixes

balazsbajorics:
**Problem:**
The navigator hides part of the error overlay.

**Fix:**
Turns out the navigator size has recently been changed, but a few places used the wrong constant (`LeftPaneDefaultWidth`) and so the navigator was 20 pixels wider than the canvas was expecting it. I also realized that the navigator can resize, so while there I fixed the following things:

**Commit Details:**
- Navigator default width codified into new constant and Atom
- Update the `NavigatorWidthAtom` when the navigator is resized
- Update the canvas offset when navigator is resized
- Changed the default canvas offset so that the new "parent floating UI" thingy fits on the screen
- Use the new `NavigatorWidthAtom` wherever needed

### DIFF
--- a/editor/src/components/canvas/design-panel-root.tsx
+++ b/editor/src/components/canvas/design-panel-root.tsx
@@ -1,9 +1,14 @@
-import { Resizable, ResizeDirection } from 're-resizable'
+import { Resizable, ResizeCallback, ResizeDirection } from 're-resizable'
 import * as React from 'react'
 import { FancyError, RuntimeErrorInfo } from '../../core/shared/code-exec-utils'
 import * as EditorActions from '../editor/actions/action-creators'
 
-import { ConsoleLog, LeftPaneDefaultWidth, RightMenuTab } from '../editor/store/editor-state'
+import {
+  ConsoleLog,
+  LeftPaneDefaultWidth,
+  RightMenuTab,
+  NavigatorWidthAtom,
+} from '../editor/store/editor-state'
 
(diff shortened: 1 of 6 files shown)


seanparsons (approved):

editor/src/components/canvas/design-panel-root.tsx
```diff
@@ -183,26 +191,15 @@ export const DesignPanelRoot = betterReactMemo('DesignPanelRoot', (props: Design
...
-    codeEditorResizingWidth,
-    interfaceDesigner.codePaneVisible,
-    interfaceDesigner.codePaneWidth,
-    leftMenuExpanded,
-    navigatorVisible,
-  ])
+  const onNavigatorResizeStop = React.useCallback<ResizeCallback>(
+    (_event, _direction, _ref, delta) => {
+      setNavigatorWidth((currentWidth) => currentWidth + delta.width)
+      dispatch([CanvasActions.scrollCanvas(canvasPoint({ x: -delta.width, y: 0 }))])
```
Rheeseyb, line 199: This line feels like it would be a pretty jarring experience tbh :grimacing:

Rheeseyb (approved):

editor/src/components/canvas/design-panel-root.tsx
balazsbajorics: it's quite nice actually

enidemi (approved):

PR #1755, opened 2021-08-20
Fix/props.children

### LINKED ISSUE #1754: `<div children={children}/>` doesn’t work
welp

balazsbajorics:
Fixes #1754

**Problem:**
The children div with Sigh will not render on the Canvas:
```javascript
<div
  children={
    <div>Sigh</div>
  }
/>
```

**Issue:**
The root of the problem is twofold:
1. Most importantly, we were kind of confusing the following cases:
```javascript
// JSX:
<div />
// transpiled javascript:
React.createElement('div', {})
```
```javascript
// JSX:
<div>{null}</div>
// transpiled javascript:
React.createElement('div', {}, null)
```
```javascript
// JSX:
<div><ChildOne /></div>
// transpiled javascript:
React.createElement('div', {}, React.crateElement(ChildOne, {}))
```
```javascript
// JSX:
<div>{[<ChildOne />]}</div>
// transpiled javascript:
React.createElement('div', {}, [ React.crateElement(ChildOne, {}) ])
```
```javascript
// JSX:
<div><ChildOne /><ChildTwo /></div>
// transpiled javascript:
React.createElement('div', {}, React.crateElement(ChildOne, {}), React.crateElement(ChildTwo, {}))
```
```javascript
// JSX:
<div>{[<ChildOne />, <ChildTwo />]}</div>
// transpiled javascript:
React.createElement('div', {}, [ React.crateElement(ChildOne, {}), React.crateElement(ChildTwo, {}) ])
```
2. For this code snippet: `<div data-uid='no-children' />` our render helper code would end up calling a createElement with null passed in as children prop, instead of nothing passed in as children prop:
```javascript
// for this JSX
<div data-uid='no-children' />
// we'd call this:
React.createElement('div', {'data-uid': 'no-children'}, null)
// instead of this:
React.createElement('div', {'data-uid': 'no-children'})
```

Now that's an issue because the _props_ might have a children key!
This is valid and happy react that renders a child:
```javascript
// JSX:
<div children={'child-one'} />
// transpiled javascript:
React.createElement('div', {children: 'child-one'})
```
This react is sad because the children prop gets overridden with a null:
```javascript
React.createElement('div', {children: 'child-one'}, null)
```

**Fix:**
I cleaned up the code that passes through the jsx children elements, removed the helper code which would pass down `null` children, and removed the helper code which was trying to unwrap `['one-child']` to `'one-child'`

**Commit Details:**
- Remove childrenOrNull and spread childrenElements in renderJSXElement
- Remove childrenElementsOrNull and spread childrenElements in the spy wrapper
- Now that we are spreading the children elements, I could remove the code that was unwrapping arrays in `renderComponentUsingJsxFactoryFunction`
- Added a test

**Note:**
This PR fixes the rendering of props.children on the canvas. But we are not properly spying these elements! They are not going to appear in the navigator, and they are not editable by the Canvas yet. I think it's not urgent to fix that, but I'm going to create a follow up ticket.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
@@ -425,13 +425,12 @@ function renderJSXElement(
       imports,
     )
   } else {
-    const childrenOrNull = childrenElements.length !== 0 ? childrenElements : null
     return renderComponentUsingJsxFactoryFunction(
       inScope,
       jsxFactoryFunctionName,
       FinalElement,
       finalPropsIcludingElementPath,
-      childrenOrNull,
+      ...childrenElements,
     )
   }
 }
@@ -518,14 +517,7 @@ export function renderComponentUsingJsxFactoryFunction(
       throw new Error(`Unable to find factory function ${factoryFunctionName} in scope.`)
     }
   }
-  // This is disgusting, but we want to make sure that if there is only one child it isn't wrapped in an array,
-  // since that code that uses `React.Children.only`
-  const childrenToRender = children.map((innerChildren) =>
-    innerChildren != null && Array.isArray(innerChildren) && innerChildren.length === 1
-      ? innerChildren[0]
-      : innerChildren,
-  )
-  return factoryFunction.call(null, type, fixedProps, ...childrenToRender)
+  return factoryFunction.call(null, type, fixedProps, ...children)
 }
 
 function fixStyleObjectRemoveCommentOnlyValues(props: Readonly<unknown>): any {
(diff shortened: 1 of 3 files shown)


Rheeseyb:
> I think it's not urgent to fix that, but I'm going to create a follow up ticket.

This is part of a more general problem - we capture elements used in arbitrary JS / JSX blocks, but not in attributes. I thought there was already a ticket for that issue, but I'm struggling to find it now...

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -518,14 +517,7 @@ export function renderComponentUsingJsxFactoryFunction(
...
   }
-  // This is disgusting, but we want to make sure that if there is only one child it isn't wrapped in an array,
-  // since that code that uses `React.Children.only`
-  const childrenToRender = children.map((innerChildren) =>
-    innerChildren != null && Array.isArray(innerChildren) && innerChildren.length === 1
-      ? innerChildren[0]
-      : innerChildren,
-  )
-  return factoryFunction.call(null, type, fixedProps, ...childrenToRender)
+  return factoryFunction.call(null, type, fixedProps, ...children)
```
Rheeseyb, line 520: Have you double checked that this fix is compatible with the use of `React.Children.only`?

Rheeseyb (approved):

enidemi (approved):

PR #1757, opened 2021-08-20
Fork project assets when forking a project

### LINKED ISSUE #1742: Forked projects lose their assets
Bug: If I fork a project, all the S3-uploaded assets are lost.

Rheeseyb:
Fixes #1742 

**Problem:**
When forking a project, Utopia doesn't also fork the assets of that project.

**Fix:**
First this PR creates a single path for dealing with project forking (previously there were a few). When forking a project we will now download the base64 content for any assets or images that are missing that data, and then uploading that content against the new project ID if the user is signed in, or store it in the local project if the user isn't signed in (this required a change to the `AssetFile` file type since that previously had no way of storing the base64 locally). We also make sure that when uploading that asset data we also scrub the base64 from the project model.

As part of this work, I realised that we were over-eagerly stripping the mime-type data that prefixes the base64. The reason for this is that we don't want that prefix when uploading. However, we _do_ want that prefix when inlining base64 via the [file-loader](https://github.com/concrete-utopia/utopia/blob/baea4c4aa28439bcdfde7a0a0036ffde6815b76c/editor/src/core/webpack-loaders/file-loader.ts#L8). I've removed that code, so that now we only strip the mime-type prefix at the point of uploading the asset.

**Commit Details:**
- Updated the definition of `AssetFile` to include the base64 content
- Added a function `getAllProjectAssetFiles` which pulls out all asset files (along with their file paths) from a project (replacing a previously dead function)
- Added a function `downloadAssetFromProject` for downloading an asset's base64
- Moved the function `getMimeStrippedBase64` (which was previously stripping the mime-type) to `server.ts` so that it is only applied at the point of uploading an asset
- Updated `triggerForkProject` so that it now downloads any assets that require downloading, prepares them to be uploaded (if signed in), updates the project contents with the updated asset files, saves the forked project, and then uploads the assets (the upload has to happen after the initial server save as otherwise the user doesn't "own" that project yet)
- Made a corresponding change in the local project uploading that happens when signing in
- Added tests to cover the various project forking cases

### DIFF
--- a/editor/src/components/assets.ts
+++ b/editor/src/components/assets.ts
@@ -9,6 +9,7 @@ import {
   isParsedTextFile,
   isTextFile,
   isParseSuccess,
+  isAssetFile,
 } from '../core/shared/project-file-types'
 import { isDirectory, directory, isImageFile } from '../core/model/project-file-utils'
 import Utils from '../utils/utils'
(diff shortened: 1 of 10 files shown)


seanparsons (approved):
Probably not something to worry about just yet but I can imagine that a project with a lot of asset files might end up being a memory hog. As this is currently downloading everything first and holding all the assets as Base64 strings for some brief moment in time.

balazsbajorics (approved):

PR #1761, opened 2021-08-23
Explicitly communicate when a remote dependency is being downloaded

Rheeseyb:
**Problem:**
The Canvas and Preview will both blindly evaluate code even if it depends on a remote dependency that hasn't been downloaded yet. This is because the package manager will return an empty object for the remote module placeholder.

**Fix:**
Throw an error specifically created for this case at the point of attempting to evaluate the placeholder file, then catch and handle that both in the Preview (in this case we add a div to the dom and include the message in that) and in the Canvas (using another error boundary specifically for catching and handling this error). In both cases the message is cleared when the dependency is downloaded (in the Preview this happens when `rerenderPreview` is called, which completely rewrites the document).

The styling of the message is far from ideal, but I wanted to open this PR up so that the approach could be commented on first.

This PR also adds the loading screen to the preview, and adds a preload tag for the logo to both the preview and the main editor.

**Screenshots:**

Canvas:
Screenshot_20210824_101334

Preview:
Screenshot_20210824_101354

### DIFF
--- a/editor/src/components/canvas/canvas-component-entry.tsx
+++ b/editor/src/components/canvas/canvas-component-entry.tsx
@@ -210,10 +210,7 @@ export class RemoteDependencyBoundary extends React.Component<
     }
   }
 
-  componentDidUpdate(
-    prevProps: RemoteDependencyBoundaryProps,
-    _prevState: RemoteDependencyBoundaryState,
-  ): void {
+  componentDidUpdate(prevProps: RemoteDependencyBoundaryProps): void {
     if (
       prevProps.projectContents !== this.props.projectContents ||
       prevProps.requireFn !== this.props.requireFn
@@ -223,7 +220,7 @@ export class RemoteDependencyBoundary extends React.Component<
     }
   }
 
-  componentDidCatch(error: Error, errorInfo: React.ErrorInfo): void {
+  componentDidCatch(error: Error): void {
     if (error?.name !== ResolvingRemoteDependencyErrorName) {
       throw error
     }
--- a/editor/src/components/canvas/canvas-component-entry.tsx
+++ b/editor/src/components/canvas/canvas-component-entry.tsx
@@ -23,6 +23,7 @@ import {
 import { ProjectContentTreeRoot } from '../assets'
 import { processErrorWithSourceMap } from '../../core/shared/code-exec-utils'
 import { DomWalkerProps, useDomWalker } from './dom-walker'
+import { ResolvingRemoteDependencyErrorName } from '../../core/es-modules/package-manager/package-manager'
 
 interface CanvasComponentEntryProps {}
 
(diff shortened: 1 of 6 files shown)


balazsbajorics:
have you considered using the same remote dependency boundary for both the preview and the canvas? I guess they'd need slightly different designs, but it'd make it easier to keep them in sync, behaviorally

Rheeseyb:
> have you considered using the same remote dependency boundary for both the preview and the canvas? I guess they'd need slightly different designs, but it'd make it easier to keep them in sync, behaviorally

The preview runs the project as is, so we have no way of wrapping it in an error boundary. That's why I've had to use `document.createElement` to surface the message

editor/src/templates/preview.tsx
```diff
@@ -276,12 +279,28 @@ const initPreview = () => {
...
+        try {
+          /**
+           * require the js entry point file which will evaluate the module.
+           * the React entry point js file traditionally has a top level side effect,
+           * calling ReactDOM.render() which starts the Preview app.
+           */
+          requireFn('/', previewJSFilePath)
+        } catch (e) {
+          if (e?.name === ResolvingRemoteDependencyErrorName && e?.message != null) {
+            const loadingMessageDiv = document.createElement('div')
```
maltenuhn, line 291: Seems legit!

maltenuhn (approved):
This reviewer will blindly evaluate design even if it depends on a screenshot that hasn't been uploaded yet. This is because the code contains a styled div for the remote reviewer.

Rheeseyb:
> This reviewer will blindly evaluate design even if it depends on a screenshot that hasn't been uploaded yet. This is because the code contains a styled div for the remote reviewer.

Hahaha I have now updated the description to include screenshots of the message displaying in both the Canvas and the Preview just for you :heart:

balazsbajorics (approved):

editor/src/components/canvas/canvas-component-entry.tsx
```diff
@@ -172,3 +178,62 @@ export class CanvasErrorBoundary extends React.Component<
...
+
+  static getDerivedStateFromError(e: any): RemoteDependencyBoundaryState {
+    return {
+      resolvingMessage: maybeGetResolvingMessage(e),
+    }
+  }
+
+  componentDidUpdate(
+    prevProps: RemoteDependencyBoundaryProps,
+    _prevState: RemoteDependencyBoundaryState,
```
seanparsons, line 215: You should be able to just omit the parameter right?

editor/src/templates/preview.tsx
```diff
@@ -276,12 +279,28 @@ const initPreview = () => {
...
+            loadingMessageDiv.innerHTML = e.message
+            const loadingMessageStyle = [
+              `font-family: -apple-system, BlinkMacSystemFont, Helvetica, 'Segoe UI', Roboto, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'`,
+              'font-size: 11px',
+              'cursor: default',
+            ].join('; ')
+            loadingMessageDiv.setAttribute('style', loadingMessageStyle)
+            document.body.append(loadingMessageDiv)
```
seanparsons, line 299: Does this mean this is created for each remote dependency error triggered? Does anything clean them up?

seanparsons (approved):

editor/src/templates/preview.tsx
Rheeseyb: It gets cleaned up when the preview is re-rendered, as that completely re-writes the document [here](https://github.com/concrete-utopia/utopia/blob/9ba418253746781432d49cca76c54e4397c1d755/editor/src/templates/preview.tsx#L257)

editor/src/components/canvas/canvas-component-entry.tsx
Rheeseyb: ...yes :upside_down_face:

Rheeseyb: Done

PR #1762, opened 2021-08-24
redux devtools: sanitize spy metadata which was too large

balazsbajorics:
Fixing Redux Devtools logging so it is usable again, should someone need it


Rheeseyb (approved):

editor/src/core/shared/redux-devtools.ts
```diff
@@ -102,6 +112,9 @@ export function reduxDevtoolsSendActions(
     const filteredActions = actions
       .flat()
       .filter((action) => !ActionsToOmit.includes(action.action))
+      .map((action) =>
+        ActionsWithoutPayload.includes(action.action) ? { action: action.action } : action,
```
seanparsons, line 116: Action action action action action action. :)

seanparsons (approved):

PR #1764, opened 2021-08-24
Expanded export support.

### LINKED ISSUE #1737: Parser is broken for `export const thing = 'hello'`
If a file contains `export const thing = 'hello'`, that should show up in the parsed file's `exportDetails`, but doesn't, meaning it cannot be imported by another file running in the canvas

seanparsons:
Mostly fixes #1737.

**Problem:**
Utopia does not support a few of the many different styles of export that are possible in JavaScript modules.

**Fix:**
Primarily this work has been to test that the various exports work, with the bulk of the changes being to handle `export default function...` as a lot of the canvas especially is designed around the assumption that a component would have a name.

**Notes:**
This does not currently address or test re-exports, which should be done in follow up work.

**Commit Details:**
- Mostly addresses #1737.
- `UtopiaJSXComponent.name` is now nullable, which indicates a component exported
  as the default exported component for a module with no fixed name.
- Added `ExportDefaultExpression` and `ExportDefaultFunction` as additional
  cases for the default export.
- Made `ComponentRendererComponent.topLevelElementName` nullable and modified
  the type guard to use a different property to identify the component.
- Removed `topLevelElementName` parameter from `tryToGetInstancePath` as it
  wasn't being used.
- Return value from `createExecutionScope` allows for top level components that
  are keyed against the value null as well as a string for the name.
- In `createExecutionScope` use `default` as the name for inserting into the scope
  for a component with no name.
- `UiJsxCanvas` handles the new cases for a `defaultExport`.
- `useGetStoryboardRoot` allows for null keyed top level components.
- `turnCodeSnippetIntoSourceMapNodes` includes a fix to get around the issue
  of attempting to transpile a function without a name or export modifiers.
- Small fix to `addStoryboardFileToProject` around handling the new export types.
- Added `markedAsExported` and `markedAsDefault` functions and standardised around
  those in a couple of places.
- Modified `printUtopiaJSXComponent` to handle components with a null name.
- `detailsFromExportDeclaration` handles `default` specifier names correctly, as well
  as null names.
- `parseCode` has some additional cases to pick up on the export types that were
  missing.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
@@ -41,8 +41,9 @@ export type ComponentRendererComponent = React.ComponentType<{
   [UTOPIA_INSTANCE_PATH]: ElementPath
   [UTOPIA_PATHS_KEY]?: string
 }> & {
-  topLevelElementName: string
+  topLevelElementName: string | null
   propertyControls?: PropertyControls
+  utopiaType: 'UTOPIA_COMPONENT_RENDERER_COMPONENT'
 }
 
 export function isComponentRendererComponent(
@@ -51,12 +52,11 @@ export function isComponentRendererComponent(
   return (
     component != null &&
     typeof component === 'function' &&
-    (component as ComponentRendererComponent).topLevelElementName != null
+    (component as ComponentRendererComponent).utopiaType === 'UTOPIA_COMPONENT_RENDERER_COMPONENT'
   )
 }
 
 function tryToGetInstancePath(
-  topLevelElementName: string,
   maybePath: ElementPath | null,
   pathsString: string | null,
 ): ElementPath | null {
@@ -71,7 +71,7 @@ function tryToGetInstancePath(
 }
 
 export function createComponentRendererComponent(params: {
-  topLevelElementName: string
+  topLevelElementName: string | null
   filePath: string
   mutableContextRef: React.MutableRefObject<MutableUtopiaCtxRefData>
 }): ComponentRendererComponent {
@@ -126,11 +126,7 @@ export function createComponentRendererComponent(params: {
 
     let codeError: Error | null = null
 
-    const instancePath: ElementPath | null = tryToGetInstancePath(
-      params.topLevelElementName,
-      instancePathAny,
-      pathsString,
-    )
+    const instancePath: ElementPath | null = tryToGetInstancePath(instancePathAny, pathsString)
 
     const rootElementPath = optionalMap(
       (path) => EP.appendNewElementPath(path, getUtopiaID(utopiaJsxComponent.rootElement)),
@@ -214,5 +210,6 @@ export function createComponentRendererComponent(params: {
   }
   Component.displayName = `ComponentRenderer(${params.topLevelElementName})`
   Component.topLevelElementName = params.topLevelElementName
+  Component.utopiaType = 'UTOPIA_COMPONENT_RENDERER_COMPONENT' as const
   return Component
 }
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx
@@ -63,7 +63,7 @@ export function createExecutionScope(
   shouldIncludeCanvasRootInTheSpy: boolean,
 ): {
   scope: MapLike<any>
-  topLevelJsxComponents: Map<string, UtopiaJSXComponent>
+  topLevelJsxComponents: Map<string | null, UtopiaJSXComponent>
   requireResult: MapLike<any>
 } {
   if (!(filePath in topLevelComponentRendererComponents.current)) {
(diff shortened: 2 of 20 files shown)


editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -417,6 +425,16 @@ function turnCodeSnippetIntoSourceMapNodes(
...
     } else {
+      // When there's an `export default function()` we strip off the `export default`, but without
+      // that it becomes an invalid chunk of JavaScript without a name so add in a default one.
+      if (
+        exportFound &&
+        defaultFound &&
+        currentStringBuffer === 'function' &&
+        FunctionStart.test(sourceCode.substr(i))
+      ) {
+        currentStringBuffer = 'function defaultFunctionName'
```
Rheeseyb, line 436: Should this default name be a constant? Is it worth making this function name more obscure / specific to utopia to prevent any potential name clash (e.g. `utopia_defaultFunctionName`)?

Rheeseyb: Genuine question, because I'm not seeing this function name referenced anywhere, so I'm not sure of the scope of any potential issues here.

seanparsons: Having this name gets this to just slide under the door of Babel and hasn't needed to be used so far, but there's certainly no downside in making the name less collision prone.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -41,8 +41,9 @@ export type ComponentRendererComponent = React.ComponentType<{
   [UTOPIA_INSTANCE_PATH]: ElementPath
   [UTOPIA_PATHS_KEY]?: string
 }> & {
-  topLevelElementName: string
+  topLevelElementName: string | null
```
Rheeseyb, line 44: It looks like `topLevelElementName` can be removed from here now (and from `Component.topLevelElementName`, since it doesn't appear to be used for anything after this work

Rheeseyb (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
balazsbajorics: hm, I liked your comment at first, but I think it's "free" to have this here, and might make debugging easier

Rheeseyb: Ok that's fair

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
seanparsons: Fixed in c9bb0678.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx
```diff
@@ -106,16 +106,15 @@ export function createExecutionScope(
...
   fastForEach(topLevelElements, (topLevelElement) => {
     if (isUtopiaJSXComponent(topLevelElement)) {
       topLevelJsxComponents.set(topLevelElement.name, topLevelElement)
-      if (!(topLevelElement.name in topLevelComponentRendererComponentsForFile)) {
-        topLevelComponentRendererComponentsForFile[
-          topLevelElement.name
-        ] = createComponentRendererComponent({
+      const elementName = topLevelElement.name ?? 'default'
+      if (!(elementName in topLevelComponentRendererComponentsForFile)) {
+        topLevelComponentRendererComponentsForFile[elementName] = createComponentRendererComponent({
```
balazsbajorics, line 117: can the elementName `'default'` clash with a user element name? I _think_ not because components must start with a capital letter

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
```diff
@@ -9,6 +9,88 @@ import {
...
+        '/defaultexpression.js': `import * as React from 'react'
+export default 2 + 2`,
+        '/defaultfunction.js': `import * as React from 'react'
+export default function() { return <div>Default Function</div> }`,
+        '/defaultnamedfunction.js': `import * as React from 'react'
+export default function DefaultNamedFunction() { return <div>Default Named Function</div> }`,
+        '/namedasdefault.js': `import * as React from 'react'
+const ToBeDefaultExported = () => <div>Named As Default</div>
+export { ToBeDefaultExported as default }`,
+      },
```
balazsbajorics, line 90: `export default class` is not listed here – is that on purpose? or is it already covered by another test case?

seanparsons: Huh, that's an omission from here, there is a separate test case which covers it here: https://github.com/concrete-utopia/utopia/blob/feature/expanded-export-support/editor/src/components/canvas/ui-jsx-canvas.spec.tsx#L1906

But I'll add an entry to this as well.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx
seanparsons: I don't _believe_ so no.

balazsbajorics (approved):

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
seanparsons: Fixed in 57fe3fe3].

PR #1771, opened 2021-08-25
Eagerly create a new project ID when creating a new project

Rheeseyb:
**Problem:**
Creating a new project (deliberately) results in a save, which in turn generates a new project ID. However, there is a race condition here as a separate action can be triggered before that save completes, resulting in another new project ID being created, and another save using that project ID.

**Fix:**
Eagerly create the project ID as part of the new project creation flow, to ensure that it is set on the model before any saving can take place.


balazsbajorics (approved):

balazsbajorics:
we forgot to merge this!

Rheeseyb:
Superseded by #1798

PR #1779, opened 2021-09-01
Re-export support.

### LINKED ISSUE #1737: Parser is broken for `export const thing = 'hello'`

seanparsons:
Fixes #1737

**Problem:**
Currently Utopia does not recognise or support module re-exports at all.

**Fix:**
This change reworks the entire set of export types so as to get a more granular and accurate representation of the various cases, as well as adding extra cases for re-exports which didn't particularly mesh well with the original types. One possible future for these new types is that they could be handled similarly to arbitrary blocks, being executed and then having their results accumulated at the end. As it's possible that the current handling does not handle particularly esoteric cases like a chain of destructured exports.

Re-exports in the canvas just defer onwards to the module that is being re-exported and then includes the values into the current scope.

**Limitations:**
Currently the `EXPORT_DESTRUCTURED_ASSIGNMENT` case will cause the functionality to degrade slightly in using the fallback require on the canvas. Handling those more correctly in the way exports have been handled so far would be difficult because of the dependency on an expression on the right hand side of the export statement.

**Commit Details:**
- Completes the fix for #1737.
- Adds `@babel/plugin-proposal-export-namespace-from` to support
  the ECMAScript 2021 re-export into a namespace.
- Changed `resolveTextFiles` to make it more closely approximate
  what real world projects would do and look like, along with
  fixing some of the test data that was affected by this change.
- Completely rewrote the various types that represent exports to be
  more specific to the various cases that arise, along with adding
  types for re-exports.
- Canvas export handling updated for new types.
- `getExportedComponentImports` fixed to handle new export types.
- `transformToCommonJS` utilises the above mentioned Babel plugin
  for namespaced imports.
- `addStoryboardFileToProject` updated for new export types.
- Fixed some of the `Arbitrary` instances to generate the updated types.
- Updated `getModifiersForComponent`, `detailsFromExportAssignment`,
  `detailsFromExportDeclaration` and `parseCode` to handle the new
  export types.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-bugs.spec.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-bugs.spec.tsx
@@ -378,7 +378,7 @@ export default function () {
         <div
           id=\\"canvas-container\\"
           style=\\"position: absolute;\\"
-          data-utopia-valid-paths=\\"storyboard-entity storyboard-entity/scene-1-entity storyboard-entity/scene-1-entity/app-entity\\"
+          data-utopia-valid-paths=\\"storyboard-entity storyboard-entity/scene-1-entity storyboard-entity/scene-1-entity/app-entity storyboard-entity/scene-1-entity/app-entity:app-outer-div storyboard-entity/scene-1-entity/app-entity:app-outer-div/card-instance storyboard-entity/scene-1-entity/app-entity:app-outer-div/d7f\\"
           data-utopia-root-element-path=\\"storyboard-entity\\"
         >
           <div
@@ -404,7 +404,7 @@ export default function () {
                 height: 100%;
                 background-color: #ffffff;
               \\"
-              data-paths=\\"storyboard-entity/scene-1-entity/app-entity\\"
+              data-paths=\\"storyboard-entity/scene-1-entity/app-entity:app-outer-div storyboard-entity/scene-1-entity/app-entity\\"
             >
               <div
                 data-uid=\\"card-outer-div card-instance\\"
@@ -415,6 +415,7 @@ export default function () {
                   width: 133px;
                   height: 300px;
                 \\"
+                data-paths=\\"storyboard-entity/scene-1-entity/app-entity:app-outer-div/card-instance:card-outer-div storyboard-entity/scene-1-entity/app-entity:app-outer-div/card-instance\\"
               >
                 <div
                   data-uid=\\"card-inner-div\\"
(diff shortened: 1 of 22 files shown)


editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -361,29 +370,123 @@ export const UiJsxCanvas = betterReactMemo(
...
+                  filteredScopeKey: string,
+                  scopeKey: string,
+                  scopeToWorkWith: MapLike<any>,
+                ): void {
+                  if (scopeKey in scopeToWorkWith) {
+                    filteredScope[filteredScopeKey] = scopeToWorkWith[scopeKey]
                   }
                 }
+
+                function addToFilteredScope(filteredScopeKey: string, scopeKey: string): void {
```
balazsbajorics, line 388: can this be moved to be a top level function?

balazsbajorics (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
seanparsons: It could but the two scopes would have to be pulled up into parameters.

balazsbajorics: I'm just a bit sad that the ui-jsx-canvas render function is becoming this unholy amalgam of module resolution + entry point for the canvas rendering

seanparsons: A better solution would be pull the whole block of export handling into a top level function, I _think_ that would work.

editor/src/components/canvas/ui-jsx-canvas-bugs.spec.tsx
```diff
@@ -378,7 +378,7 @@ export default function () {
         <div
           id=\\"canvas-container\\"
           style=\\"position: absolute;\\"
-          data-utopia-valid-paths=\\"storyboard-entity storyboard-entity/scene-1-entity storyboard-entity/scene-1-entity/app-entity\\"
+          data-utopia-valid-paths=\\"storyboard-entity storyboard-entity/scene-1-entity storyboard-entity/scene-1-entity/app-entity storyboard-entity/scene-1-entity/app-entity:app-outer-div storyboard-entity/scene-1-entity/app-entity:app-outer-div/card-instance storyboard-entity/scene-1-entity/app-entity:app-outer-div/d7f\\"
```
Rheeseyb, line 381: Nice!

Rheeseyb (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
seanparsons: Fixed in 5591ee12.

PR #1784, opened 2021-09-02
Draft/show editor before canvas load finished

balazsbajorics:


balazsbajorics:
I forgot to close this, as per conversation with Malte we decided to tackle this later down the line once we fixed the more fundamental editor load issues

PR #1788, opened 2021-09-03
fix(editor) Handle the same import with multiple different paths.

### LINKED ISSUE #1773: Import merging fails in the beaches project
Open https://utopia.app/p/36ae27be-welcome-to-utopia and make a change via the canvas or inspector to something in the `app.js` file. This will result in a new import `import { Ratings } from './card.js'` being added, without replacing the existing `import { Ratings } from '/src/card.js'`

seanparsons:
Fixes #1773

**Problem:**
When reproducing the code from the parsed model, the printing logic wasn't correctly handling there case where multiple import statements refer to the same file but with differing paths.

**Fix:**
Consolidate paths for imports by converting them into their absolute path.

**Commit Details:**
- Fixes #1773.
- Changed `mergeImports` to cater for multiple imports for the same
  absolute path with differing paths. Imports are now merged against
  the existing value if there is one for the import keyed against
  `existingKeyToUse`.
- `printCodeImpl` now checks the absolute path when finding import
  statements that match the given import origin.
- Added the file path as a parameter to functions that call
  `printCodeImpl` as it now requires the path of the code being printed.

### DIFF
--- a/editor/src/core/workers/common/project-file-utils.spec.ts
+++ b/editor/src/core/workers/common/project-file-utils.spec.ts
@@ -128,4 +128,20 @@ describe('mergeImports', () => {
       'component-library': importDetails(null, [importAlias('Card'), importAlias('FlexRow')], null),
     })
   })
+
+  it('handles multiple imports for the same file with different paths', () => {
+    const result = mergeImports(
+      '/src/code.js',
+      {
+        '/src/fileA.js': importDetails(null, [importAlias('Card')], null),
+        './fileA.js': importDetails(null, [importAlias('OtherCard')], null),
+      },
+      { '/src/fileB.js': importDetails(null, [importAlias('FlexRow')], null) },
+    )
+
+    expect(result).toEqual({
+      '/src/fileA.js': importDetails(null, [importAlias('Card'), importAlias('OtherCard')], null),
+      '/src/fileB.js': importDetails(null, [importAlias('FlexRow')], null),
+    })
+  })
 })
--- a/editor/src/core/workers/common/project-file-utils.spec.ts
+++ b/editor/src/core/workers/common/project-file-utils.spec.ts
@@ -133,15 +133,15 @@ describe('mergeImports', () => {
     const result = mergeImports(
       '/src/code.js',
       {
-        '/src/fileA.js': importDetails(null, [importAlias('Card')], null),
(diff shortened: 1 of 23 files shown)


editor/src/core/workers/parser-printer/parser-printer-bugs.spec.ts
```diff
@@ -285,7 +285,23 @@ export default function () {
 }
 `
 
-    testParseThenPrintWithoutUids(spreadCode, spreadCode)
+    testParseThenPrintWithoutUids('/index.js', spreadCode, spreadCode)
+  })
+
+  it('#1773 - Handles imports which relate to the same file but have differing paths.', () => {
+    const spreadCode = `import * as React from 'react'
+import { FirstComponent } from './components.js'
```
balazsbajorics, line 293: do I read it correctly that we are reserving the user's multiple imports? that's great!

balazsbajorics (approved):

editor/src/core/workers/parser-printer/parser-printer-bugs.spec.ts
seanparsons: Attempts to, but doesn't guarantee it is probably the better way to describe it.

editor/src/core/workers/common/project-file-utils.spec.ts
```diff
@@ -128,4 +128,20 @@ describe('mergeImports', () => {
       'component-library': importDetails(null, [importAlias('Card'), importAlias('FlexRow')], null),
     })
   })
+
+  it('handles multiple imports for the same file with different paths', () => {
+    const result = mergeImports(
+      '/src/code.js',
+      {
+        '/src/fileA.js': importDetails(null, [importAlias('Card')], null),
```
Rheeseyb, line 136: Very minor, but please can you remove the `.js` suffixes from the import statements, since you wouldn't typically include those, and I just want to make sure we're not accidentally (now or in the future) depending on them

seanparsons: I only did it like this because the entire rest of the file is like it.

Rheeseyb (approved):

editor/src/core/workers/common/project-file-utils.spec.ts
seanparsons: Fixed in 8c71e97c.

PR #1789, opened 2021-09-03
webpack-5, part 1: Bump webpack-dev-server

balazsbajorics:
**Problem:**
The new webpack-dev-server has slightly different configuration

**Fix:**
Update the configuration

**Commit Details:**
- move things around in webpack.config.js
- the `--define` no longer works, so I use the `EnvironmentPlugin` instead. 

**Note:**
The `EnvironmentPlugin` means that in the package.json scripts, I moved a whole bunch of `--define process.env.VARIABLE=VALUE` stuff to be `VARIABLE=VALUE webpack...` instead. the diff is very hard to read and I ask you to check my work because I _think_ I managed to copy paste things in the correct order but maybe I messed it up.

https://v4.webpack.js.org/plugins/environment-plugin/#usage

The good part of the EnvironmentPlugin is that it prints an error if an env var is expected but not present, whereas omiting a `--define` would have lead to silent errors. yay! Unfortunately I couldn't find a way to make it an error error, that would be even better.


balazsbajorics:
oops I forgot to request reviews on this one!

maltenuhn:
Performance test sus?

balazsbajorics:
> Performance test sus?

I'm re-running the tests but I believe the previous run was a fluke

editor/package.json
```diff
@@ -27,11 +27,11 @@
...
+    "move-fast-and-break-things-hot": "REACT_APP_ENVIRONMENT_CONFIG=development REACT_APP_AUTH0_CLIENT_ID=$AUTH0_CLIENT_ID REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=`git rev-parse HEAD` NODE_OPTIONS=--max_old_space_size=8192 HOT='true' webpack-dev-server --progress",
+    "production": "REACT_APP_ENVIRONMENT_CONFIG=production REACT_APP_AUTH0_CLIENT_ID=$AUTH0_CLIENT_ID REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=$UTOPIA_SHA WEBPACK_MODE='production' REACT_APP_COMMIT_HASH=`git rev-parse HEAD` NODE_OPTIONS=--max_old_space_size=6656 webpack",
+    "staging": "REACT_APP_ENVIRONMENT_CONFIG=staging REACT_APP_AUTH0_CLIENT_ID=$AUTH0_CLIENT_ID REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=$UTOPIA_SHA WEBPACK_MODE='staging' REACT_APP_COMMIT_HASH=`git rev-parse HEAD` NODE_OPTIONS=--max_old_space_size=6656 webpack",
```
enidemi, line 34: The GOOGLE_WEB_FONTS_KEY env is missing from these, is it not used anymore?

enidemi (approved):

editor/package.json
balazsbajorics: good catch. it is used, however, that one did not need renaming.

seanparsons (approved):

editor/package.json
Rheeseyb: I don't understand this comment. So was accidentally removed and needs to be put back before this can be merged?

editor/package.json
```diff
@@ -27,11 +27,11 @@
...
-    "staging": "WEBPACK_MODE='staging' NODE_OPTIONS=--max_old_space_size=6656 webpack --define process.env.REACT_APP_ENVIRONMENT_CONFIG=\"'staging'\" --define process.env.REACT_APP_AUTH0_CLIENT_ID=\"'$AUTH0_CLIENT_ID'\" --define process.env.REACT_APP_AUTH0_ENDPOINT=\"'$AUTH0_ENDPOINT'\" --define process.env.REACT_APP_AUTH0_REDIRECT_URI=\"'$AUTH0_REDIRECT_URI'\" --define process.env.REACT_APP_COMMIT_HASH=\"'$UTOPIA_SHA'\" --define process.env.GOOGLE_WEB_FONTS_KEY=\"'$GOOGLE_WEB_FONTS_KEY'\"",
+    "move-fast-and-break-things": "REACT_APP_ENVIRONMENT_CONFIG=development REACT_APP_AUTH0_CLIENT_ID=$AUTH0_CLIENT_ID REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=`git rev-parse HEAD` NODE_OPTIONS=--max_old_space_size=8192 webpack-dev-server --progress",
+    "performance-test": "NODE_ENV=production REACT_APP_ENVIRONMENT_CONFIG=development REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=`git rev-parse HEAD` NODE_OPTIONS=--max_old_space_size=8192 PERFORMANCE='true' webpack-dev-server --progress",
```
Rheeseyb, line 31: This one is missing `REACT_APP_AUTH0_CLIENT_ID`

editor/package.json
balazsbajorics: so the env vars listed here are the ones that need renaming / aliasing, because the "outside" env var is called, for example AUTH0_ENDPOINT but the "internal" env var is named REACT_APP_AUTH0_ENDPOINT

Rheeseyb: Riiight ok I'm with you now :+1:

editor/package.json

PR #1792, opened 2021-09-06
Chore/bump webpack 5

balazsbajorics:
This is the second PR following #1789 

The PR updates webpack to version 5. 

**Changes:**
* Webpack no longer shims Node.js builtins like `process` and `Buffer`. Since we don't really use these I decided to follow the webpack documentation's advice and not include browser shims. This was mostly painless, but I had to add some specific entries to the config file to fix runtime type errors, such as code testing if `process.os` is defined (throws because process is undefined, so process.os us accessing a property on undefined...)
* I'm using `webpack.DefinePlugin` which is a string replacer. The config entry `'process.platform': 'undefined'` means that wherever the code says `process.platform`, it will literally be replaced with an `undefined` literal. so code like `process.platform != null` will be transpiled to `undefined != null` 
* I think I'm not causing issues by my usage of the DefinePlugin, but we can fine tune the config file.
* `node_modules/istextorbinary` was killing webpack 5 with spectacularly undecipherable errors. I "fixed it" by removing the `exports` field from its package.json and creating a patch-package for that. It's a very ugly solution to a WTF problem, but here's hope by the time my hotfix breaks, either istextorbinary or webpack will come up with a solution to this problem
* I updated Karma, and the karma.conf.js file is altering the webpack config so that had to be updated as well


enidemi (approved):

editor/package.json
```diff
@@ -28,7 +28,7 @@
     "test-browser": "karma start --single-run",
     "test-browser-watch": "karma start",
     "move-fast-and-break-things": "REACT_APP_ENVIRONMENT_CONFIG=development REACT_APP_AUTH0_CLIENT_ID=$AUTH0_CLIENT_ID REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=`git rev-parse HEAD` NODE_OPTIONS=--max_old_space_size=8192 webpack-dev-server --progress",
-    "performance-test": "NODE_ENV=production REACT_APP_ENVIRONMENT_CONFIG=development REACT_APP_AUTH0_CLIENT_ID=$AUTH0_CLIENT_ID REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=`git rev-parse HEAD` NODE_OPTIONS=--max_old_space_size=8192 PERFORMANCE='true' webpack-dev-server --progress",
+    "performance-test": "NODE_ENV=production REACT_APP_ENVIRONMENT_CONFIG=development REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=`git rev-parse HEAD` NODE_OPTIONS=--max_old_space_size=8192 PERFORMANCE='true' webpack-dev-server --progress",
```
Rheeseyb, line 31: You appear to have removed `REACT_APP_AUTH0_CLIENT_ID=$AUTH0_CLIENT_ID`?

editor/patches/istextorbinary+5.11.0.patch
```diff
@@ -0,0 +1,21 @@
...
+-  "exports": {
+-    "node": {
+-      "import": "./edition-es2019-esm/index.js",
+-      "default": "./index.cjs",
+-      "require": "./edition-esnext/index.js"
+-    },
+-    "browser": {
+-      "import": "./edition-browsers/index.js"
+-    }
+-  },
```
Rheeseyb, line 18: Can this issue be fixed by updating `istextorbinary` to `6.0.0`?

balazsbajorics: I tried it last week on a branch and did not help

Rheeseyb (approved):

editor/package.json
balazsbajorics: oh no

PR #1794, opened 2021-09-06
Load VS Code earlier.

### LINKED ISSUE #1793: Make VS Code load earlier.
Currently the VS Code iframe loads comparatively late once the project has loaded. If it were to start loading earlier the end-to-end loading time would be reduced.

seanparsons:
Fixes #1793

**Problem:**
Currently the VS Code iframe loads quite late, but makes about 1/3 of the requests during the startup of the editor including a few sizeable downloads.

**Fix:**
The iframe is now loaded effectively immediately as a value for `vscodeBridgeId` will now be available immediately and cannot be null.

An additional fix was made which potentially only helps locally as it delays the property controls component from starting up until the VS Code bridge is ready. As it was blocking execution on the main thread which delayed the VS Code iframe from opening.

**Commit Details:**
- Fixes #1793.
- `EditorState.vscodeBridgeId` is now not nullable and a value is
  always supplied to it during startup.
- The `CodeEditorWrapper` now does not wait for the bridge to be
  marked as ready.
- The `PropertyControlsInfoComponent` holds off from starting up
  until the VS Code bridge is ready.

### DIFF
--- a/editor/src/components/editor/editor-component.tsx
+++ b/editor/src/components/editor/editor-component.tsx
@@ -61,6 +61,7 @@ interface NumberSize {
 
 export interface EditorProps {
   propertyControlsInfoSupported: boolean
+  vscodeBridgeReady: boolean
 }
 
 function useDelayedValueHook(inputValue: boolean, delayMs: number): boolean {
@@ -331,7 +332,9 @@ export const EditorComponentInner = betterReactMemo(
         <ModalComponent />
         <ToastRenderer />
         <CanvasCursorComponent />
-        {props.propertyControlsInfoSupported ? <PropertyControlsInfoComponent /> : null}
+        {props.propertyControlsInfoSupported && props.vscodeBridgeReady ? (
+          <PropertyControlsInfoComponent />
+        ) : null}
       </SimpleFlexRow>
     )
   },
--- a/editor/src/components/canvas/dom-walker.spec.browser.tsx
+++ b/editor/src/components/canvas/dom-walker.spec.browser.tsx
@@ -83,6 +83,7 @@ async function renderTestEditorWithCode(appUiJsFileCode: string) {
       useStore={storeHook}
       spyCollector={spyCollector}
       propertyControlsInfoSupported={false}
+      vscodeBridgeReady={false}
     />,
   )
   const noFileOpenText = result.getByText('No file open')
--- a/editor/src/components/canvas/ui-jsx.test-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx.test-utils.tsx
@@ -218,6 +218,7 @@ export async function renderTestEditorWithModel(
         useStore={storeHook}
         spyCollector={spyCollector}
         propertyControlsInfoSupported={false}
+        vscodeBridgeReady={false}
       />
     </React.Profiler>,
   )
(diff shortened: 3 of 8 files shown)


balazsbajorics (approved):
looks legit. however, the VSCode iframe still only starts loading once the editor first renders. 

in Discord last week we talked about trying to append a dummy vscode iframe to the editor template html, to see if it helps with first load time (basically acting as a prefetch for Chrome)? or appending the iframe, and then instead of destroying it and creating the 'real' iframe, try to somehow keep the one iframe alive and move it to a new position in the HTML document?

if you haven't looked into those yet, can you make a follow-up ticket for that?

seanparsons:
> in Discord last week we talked about trying to append a dummy vscode iframe to the editor template html, to see if it helps with first load time (basically acting as a prefetch for Chrome)? or appending the iframe, and then instead of destroying it and creating the 'real' iframe, try to somehow keep the one iframe alive and move it to a new position in the HTML document?

I did look into this, the issues I could see:
- Having a dummy iframe means that the real iframe loads at some indeterminate point later, which means there _may_ be some benefit in loading if the timing is just right. But if the timing is slightly off we end up making two requests to a lot of the VS Code components thereby slowing down.
- Shifting the iframe makes me a little nervous, as we'd have to make it not display via zero-sizing or some such and potentially while VS Code is initialising shunt it to a different part of the DOM. With that being done by some weird side effect'y component.
- Having VS Code queue up a bunch of downloads during the editor download is entirely likely to make the main editor load slower because it will be sharing that same downstream.

editor/src/components/editor/editor-component.tsx
```diff
@@ -331,7 +332,9 @@ export const EditorComponentInner = betterReactMemo(
         <ModalComponent />
         <ToastRenderer />
         <CanvasCursorComponent />
-        {props.propertyControlsInfoSupported ? <PropertyControlsInfoComponent /> : null}
+        {props.propertyControlsInfoSupported && props.vscodeBridgeReady ? (
+          <PropertyControlsInfoComponent />
```
Rheeseyb, line 336: Can you add a comment to this line please, because I can definitely see us forgetting why we're reliant on `props.vscodeBridgeReady` here

Rheeseyb (changes requested):
In `actions.ts` we update the `vscodeBridgeId` when setting the project ID, [only if it is null](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/editor/actions/actions.ts#L3433), so that will never happen any more. Because of this, it means VS Code won't connect to the correct IndexedDB store when the project ID is first set (i.e. when creating an empty project). I've just tested and it seems to work when reloading the editor after creating an empty project, probably because at that point a project ID exists from the beginning

seanparsons:
> In `actions.ts` we update the `vscodeBridgeId` when setting the project ID, [only if it is null](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/editor/actions/actions.ts#L3433), so that will never happen any more. Because of this, it means VS Code won't connect to the correct IndexedDB store when the project ID is first set (i.e. when creating an empty project). I've just tested and it seems to work when reloading the editor after creating an empty project, probably because at that point a project ID exists from the beginning

Fixed in e14d0689.

seanparsons:
> In `actions.ts` we update the `vscodeBridgeId` when setting the project ID, [only if it is null](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/editor/actions/actions.ts#L3433), so that will never happen any more. Because of this, it means VS Code won't connect to the correct IndexedDB store when the project ID is first set (i.e. when creating an empty project). I've just tested and it seems to work when reloading the editor after creating an empty project, probably because at that point a project ID exists from the beginning

Fixed in 3615bbd6.

Rheeseyb (approved):

PR #1795, opened 2021-09-07
Fix/npm packages editor load

enidemi:
**Problem:**
When loading a project with npm dependencies the editor checks the version for each package before loading the packages.
When changing the project package.json file all dependencies are redownloaded.

**Fix:**
Removed version check from `fetchNodeModules`, it is using the version defined in package.json.

**Commit Details:**
- filter existing dependencies when changing the package.json
- removing the version checking when fetching


balazsbajorics (approved):
wow, so it turns out that looking up the package version before loading was entirely unnecessary?

seanparsons (approved):

enidemi:
> wow, so it turns out that looking up the package version before loading was entirely unnecessary?

I think it is. When using the dependency list UI there is a version check before fetching to find the 'latest' version, and after that it's always using a specific version.

balazsbajorics:
> > wow, so it turns out that looking up the package version before loading was entirely unnecessary?
> 
> I think it is. When using the dependency list UI there is a version check before fetching to find the 'latest' version, and after that it's always using a specific version.

what if I write something like "latest" to the package.json? or a "*"? does the packager server work correctly with those as versions?

enidemi:
> > > wow, so it turns out that looking up the package version before loading was entirely unnecessary?
> > 
> > 
> > I think it is. When using the dependency list UI there is a version check before fetching to find the 'latest' version, and after that it's always using a specific version.
> 
> what if I write something like "latest" to the package.json? or a "*"? does the packager server work correctly with those as versions?

I checked this scenario and the packager loads dependencies with both * and latest.

PR #1797, opened 2021-09-07
worker chunking

balazsbajorics:
**Problem:**
We have 3 large workers – tsWorker, parserPrinterWorker, and linterWorker. They are very heavy because they have dependencies like standalone Babel and Typescript. The three workers actually share a majority of dependencies, but in our current setup each worker is compiled into an isolated js bundle without chunking, they cannot reuse the same dependencies.  This means that when fully loading the utopia editor, we download the same typescript and babel about 4 times.

**Fix:**
With webpack v5, if you load a worker using a specific syntax that webpack can statically analyze (`return new Worker(new URL('./parser-printer/parser-printer.worker.ts', import.meta.url))`) Webpack automatically creates proper reusable chunks!

**Commit Details:**
- Use webpack-friendly worker loading syntax
- There was an issue with the webpack publicPath pointing to the CDN (workers are not allowed to load from cdn because of same-origin policies). The workaround for the issue is to dynamically change `__webpack_public_path__` when loading the workers and then reset it to the CDN URL. It's not nice, but it works
- had to fix Jest tests because they don't support `import.meta.url`

Here's master from a few days ago:
old
Notice that there's multiple typescript.js'es, among other things.

Here's how the new bundles look like:
new
Note that @enidemi 's #1801 is not merged yet, so these two PRs combined should make for an even nicer reduction!
You can see that there's only a single typescript.js (instead of three!) There's only one remaining mystery, there's still two babel.js's, but I don't want to investigate until Eni's PR is on master.

### DIFF
--- a/editor/src/core/es-modules/package-manager/built-in-dependencies-list.ts
+++ b/editor/src/core/es-modules/package-manager/built-in-dependencies-list.ts
@@ -7,8 +7,8 @@ import * as ReactDOM from 'react-dom'
 import * as EmotionReact from '@emotion/react'
 import * as EmotionStyled from '@emotion/styled'
 
-import * as editorPackageJSON from '../../../../package.json'
-import * as utopiaAPIPackageJSON from '../../../../../utopia-api/package.json'
+import editorPackageJSON from '../../../../package.json'
+import utopiaAPIPackageJSON from '../../../../../utopia-api/package.json'
 import { applyUIDMonkeyPatch } from '../../../utils/canvas-react-utils'
 
 applyUIDMonkeyPatch()
--- a/editor/src/core/workers/utils.ts
+++ b/editor/src/core/workers/utils.ts
@@ -22,7 +22,6 @@ export function createTsWorker(): Worker {
 export function createParserPrinterWorker(): Worker {
   const oldPublicPath = __webpack_public_path__
   __webpack_public_path__ = BASE_URL
-  console.info('createParserPrinterWorker', import.meta.url, oldPublicPath, __webpack_public_path__)
   const worker = new Worker(new URL('./parser-printer/parser-printer.worker.ts', import.meta.url))
   __webpack_public_path__ = oldPublicPath
   return worker
--- a/editor/src/core/workers/utils.ts
+++ b/editor/src/core/workers/utils.ts
@@ -13,7 +13,6 @@ import { BASE_URL } from '../../common/env-vars'
 export function createTsWorker(): Worker {
   const oldPublicPath = __webpack_public_path__
   __webpack_public_path__ = BASE_URL
-  console.info('create ts worker', import.meta.url, oldPublicPath, __webpack_public_path__)
   const worker = new Worker(new URL('./ts/ts.worker.ts', import.meta.url))
   __webpack_public_path__ = oldPublicPath
   return worker
--- a/editor/src/core/workers/utils.ts
+++ b/editor/src/core/workers/utils.ts
@@ -1,10 +1,45 @@
-import { BASE_URL, URL_HASH } from '../../common/env-vars'
-import { setBranchNameFromURL } from '../../utils/branches'
+import { BASE_URL } from '../../common/env-vars'
 
-export function workerForFile(fileName: string): Worker {
-  let workerURL = new URL(`${BASE_URL}${fileName}`)
-  let workerQueryParams = workerURL.searchParams
-  workerQueryParams.set('hash', URL_HASH)
-  setBranchNameFromURL(workerQueryParams)
-  return new Worker(workerURL.href)
+/**
+ * Webpack 5 only supports a single public path for assets and feeding the Worker url via `import.meta.url`
+ * This causes problems when using a CDN, because of same origin restrictions, you are not allowed to load workers from cdns.
+ *
+ * The solution is to use the webpack-supported global __webpack_public_path__ variable, which updates the value returned by import.meta.url
+ *
+ * in all of these cases, oldPublicPath is the asset url (the CDN url in deployed editors). we temporarily change it to BASE_URL (which is the url the editor is loaded from)
+ * and then after loading the worker, restore it to the CDN url.
+ */
+
+export function createTsWorker(): Worker {
+  const oldPublicPath = __webpack_public_path__
+  __webpack_public_path__ = BASE_URL
+  console.info('create ts worker', import.meta.url, oldPublicPath, __webpack_public_path__)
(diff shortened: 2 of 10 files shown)


balazsbajorics:
wow, all checks pass!!!!!!!

seanparsons (approved):

editor/src/core/es-modules/package-manager/built-in-dependencies-list.ts
```diff
@@ -7,8 +7,8 @@ import * as ReactDOM from 'react-dom'
 import * as EmotionReact from '@emotion/react'
 import * as EmotionStyled from '@emotion/styled'
 
-import * as editorPackageJSON from '../../../../package.json'
-import * as utopiaAPIPackageJSON from '../../../../../utopia-api/package.json'
+import editorPackageJSON from '../../../../package.json'
+import utopiaAPIPackageJSON from '../../../../../utopia-api/package.json'
```
Rheeseyb, line 11: Was this related?

balazsbajorics: yep, changing the module resolution in tsconfig made webpack realize that this should be a default import

editor/src/core/workers/utils.ts
```diff
@@ -1,10 +1,45 @@
...
+import { BASE_URL } from '../../common/env-vars'
+
+/**
+ * Webpack 5 only supports a single public path for assets and feeding the Worker url via `import.meta.url`
+ * This causes problems when using a CDN, because of same origin restrictions, you are not allowed to load workers from cdns.
+ *
+ * The solution is to use the webpack-supported global __webpack_public_path__ variable, which updates the value returned by import.meta.url
+ *
+ * in all of these cases, oldPublicPath is the asset url (the CDN url in deployed editors). we temporarily change it to BASE_URL (which is the url the editor is loaded from)
+ * and then after loading the worker, restore it to the CDN url.
```
Rheeseyb, line 10: > and then after loading the worker

So does this mean that all of the workers will be downloading their js bundles directly from the server rather than the CDN?

balazsbajorics: yes, that's happening now too, because you can't eval js coming from the cdn

editor/src/core/workers/utils.ts
```diff
@@ -1,10 +1,45 @@
...
+ */
+
+export function createTsWorker(): Worker {
+  const oldPublicPath = __webpack_public_path__
+  __webpack_public_path__ = BASE_URL
+  console.info('create ts worker', import.meta.url, oldPublicPath, __webpack_public_path__)
```
Rheeseyb, line 16:
```suggestion
```

editor/src/core/workers/utils.ts
```diff
@@ -1,10 +1,45 @@
...
+  const worker = new Worker(new URL('./ts/ts.worker.ts', import.meta.url))
+  __webpack_public_path__ = oldPublicPath
+  return worker
+}
+
+export function createParserPrinterWorker(): Worker {
+  const oldPublicPath = __webpack_public_path__
+  __webpack_public_path__ = BASE_URL
+  console.info('createParserPrinterWorker', import.meta.url, oldPublicPath, __webpack_public_path__)
```
Rheeseyb, line 25:
```suggestion
```

editor/src/core/workers/utils.ts
Rheeseyb: Well that's thoroughly disappointing :disappointed:

editor/src/core/workers/utils.ts
```diff
@@ -1,10 +1,43 @@
...
+ *
+ * The solution is to use the webpack-supported global __webpack_public_path__ variable, which updates the value returned by import.meta.url
+ *
+ * in all of these cases, oldPublicPath is the asset url (the CDN url in deployed editors). we temporarily change it to BASE_URL (which is the url the editor is loaded from)
+ * and then after loading the worker, restore it to the CDN url.
+ */
+
+export function createTsWorker(): Worker {
+  const oldPublicPath = __webpack_public_path__
+  __webpack_public_path__ = BASE_URL
```
Rheeseyb, line 15: Did you try using the [`publicPath`](https://webpack.js.org/loaders/worker-loader/#publicpath) config option?

editor/src/core/workers/utils.ts
seanparsons: Oh that's bad times, I thought it was fine because it's a subdomain.

Rheeseyb (approved):

editor/src/core/workers/utils.ts
Rheeseyb: It's still a vast improvement because it means the workers themselves will now be sharing chunks. It does however mean that chunks shared between the editor and the workers will still be downloaded twice, with the worker downloads being slower

editor/src/core/workers/utils.ts
balazsbajorics: that is set to our CDN

editor/src/core/workers/utils.ts
balazsbajorics: are we serving the editor js from the cdn?! I'm a bit unclear on these things. what actually triggered me doing this __webpack_public_path__ business was seeing the same origin error message when trying to load the workers on staging

seanparsons: I checked just now on and we are loading it from the CDN.

editor/src/core/workers/utils.ts
Rheeseyb: You can provide a separate value for the worker loader via the rules (that link shows a couple of examples)

editor/src/core/workers/utils.ts
Rheeseyb: The editor loads all of its js bundles from CDN

editor/src/core/workers/utils.ts
balazsbajorics: the worker loader is for webpack 4, it has been superseded by the `new Worker(...` syntax in webpack 5, which has no configuration options

Rheeseyb: :facepalm:

balazsbajorics: yep.

editor/src/core/workers/utils.ts
balazsbajorics: image

@Rheeseyb I just opened utopia.pizza to double check and it seems like we are loading the workers from not-the-cdn.

PR #1798, opened 2021-09-07
Persistence state machine

Rheeseyb:
This PR replaces our (terrible) persistence core logic with a state machine. The final state machine can be viewed in the xstate visualiser [here](https://stately.ai/viz/a65e6117-5dfd-49af-98ea-d22f1e3ebc2c)

The state machine itself is comprised of 3 parallel inner machines:

1. The "core" machine itself, which is the part responding to events from the interpreter and determining how to handle those
2. The "backend" machine, for communicating with the server etc.
3. The "user" machine, for handling login and logout events

The machine has a context, for keeping track of the project and its ID, as well as whether the project is owner, whether the user is logged in, and for holding a queued save if one is in progress right now (not to be confused with throttling - this is purely for the case where a save event was received by the machine whilst a save is in progress). We could remove the current project from the context, but doing so requires passing it through in more events. Also, the `loggedIn` boolean exists in the context because there doesn't appear to be a clean way for one part of the machine to check the current state of another.

For error handling, the machine just logs out the error and returns to the "Ready" state - this is pretty much exactly the same situation we had previously, and means that further saves won't be blocked if one fails.

`persistence-backend.ts` provides the logic for interacting with the server and locally stored projects, including handling asset uploading and downloading, giving us a concrete version of the `PersistenceBackendAPI`.

In `persistence.ts` we use an interpreter, which allows us to send events to the machine, as well as listen out for state changes and events coming from the machine. We listen for transitions through the machine to build up an array of editor actions to dispatch, which are then all dispatched together once the machine reaches the "Ready" state. The interpreter is also where we handle throttling of saves, as there didn't seem to be a nice way to fit that into the state machine itself.

The whole thing is exposed via the class `PersistenceMachine`, which takes callbacks for handling a 404, as well as handling a project being loaded or created (as it that part is then handled via existing functions `actions.ts`). Because new and loaded projects will now go through the same flow, it means we can remove the `NEW` editor action (and the `new` function in `actions.ts`), which will happen in a follow up PR.

**Notes:**

- In a lot of places, the `assign` action is needlessly taking params that aren't used or required - this is a quirk with the xstate types (see [here](https://xstate.js.org/docs/guides/typescript.html#assign-action-behaving-strangely))
- We no longer have a separate flow for just updating the name of the project. We can bring that back if needed, but that was adding unnecessary complexity to the system for no real gain
- This change loses the toast that showed after uploading an existing local project, as it no longer tracks whether the project is stored locally. I have an idea of how to trigger that from the interpreter though.


editor/src/components/editor/persistence/persistence-machine.ts
```diff
@@ -0,0 +1,1216 @@
...
+            id: 'server-save-project',
+            src: (context, event) => {
+              if (
+                event.type === 'BACKEND_SERVER_SAVE' &&
+                event.projectModel != null &&
+                context.projectId != null
+              ) {
+                return saveProjectToServer(context.projectId, event.projectModel)
+              } else {
+                throw new Error(
```
seanparsons, line 738: What happens to exceptions in the state machine?

Rheeseyb: This will trigger the `onError` below, which then sends an event for the core machine to handle, which right now just squashes them and returns to the "ready" state, so they still need handling before this is complete.

editor/src/templates/editor.tsx
```diff
@@ -227,120 +255,62 @@ export class Editor {
...
-                () => {
-                  renderProjectNotFound()
-                },
-              )
-            }
-          })
+          this.storedState.persistence.load(projectId)
         }
+
+        // FIXME Restore github project importing
```
Rheeseyb, line 268: I'm fixing this up now, so this will either be fixed in time for the PR to be merged, or in an immediate follow up

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -484,14 +486,6 @@ export function editorDispatch(
     })
   })
 
-  if ((isLoadAction || nameUpdated) && frozenEditorState.id != null) {
```
Rheeseyb, line 487: This is now handled by the interpreter

editor/src/components/editor/server.ts
```diff
@@ -396,14 +395,6 @@ export function startPollingLoginState(
           // Login was lost and subsequently regained so remove the persistent toast.
           dispatch([removeToast(loginLostNoticeID)])
         }
-        if (isNotLoggedIn(previousLoginState)) {
```
Rheeseyb, line 399: This doesn't appear to be necessary anymore

editor/src/components/editor/persistence/persistence.spec.ts
```diff
@@ -0,0 +1,528 @@
...
+    expect(localProjects[localProjectKey(capturedData.newProjectId!)]).toBeUndefined()
+    expect(capturedData.newProjectId).toBeDefined()
+    expect(mockSaveLog[capturedData.newProjectId!]).toEqual([BaseModel, FirstRevision])
+  })
+
+  it('Throttles saves', async () => {
+    const { capturedData, testMachine } = setupTest(10000)
+
+    testMachine.login()
+    await delay(20)
```
balazsbajorics, line 244: we should use Jest's mock timers because this file adds at least 500ms of extra delays to the test run!

balazsbajorics: wait, I'm realizing you don't have the delays becuase of throttling – you have the delays to make sure that the state machine completed running? :O that's kinda horrifying!

editor/src/components/editor/persistence/persistence.spec.ts
```diff
@@ -0,0 +1,528 @@
...
+    addLocalProject(projectId, BaseModel)
+
+    testMachine.load(projectId)
+    await delay(20)
+
+    expect(capturedData.projectNotFound).toBeFalsy()
+    expect(capturedData.createdOrLoadedProject).toEqual(BaseModel)
+  })
+
+  it('Favours a local project over a server project', async () => {
```
balazsbajorics, line 391: does it mean if a project exists both locally and on the server, we favor the local version? is this right? is this because we want to handle the editor going temporarily offline?

editor/src/components/editor/persistence/persistence.ts
```diff
@@ -0,0 +1,231 @@
...
+  private getRemainingSaveDelay(): number {
+    return Math.max(0, this.lastSavedTS + this.saveThrottle - Date.now())
+  }
+
+  private shouldThrottle(forceOrThrottle: 'force' | 'throttle'): boolean {
+    return forceOrThrottle === 'throttle' && this.getRemainingSaveDelay() > 0
+  }
+
+  // Exposed for testing
+  sendThrottledSave(): void {
```
balazsbajorics, line 183: I think all of these functions should be replaced to arrow functions to avoid accidentally messing up where the `this` points to. I actually thought we had an eslint rule preventing this style!

balazsbajorics (approved):

editor/src/components/editor/persistence/persistence.spec.ts
Rheeseyb: Yeah. This is because you can't really fire an event and track when _that_ event has been completely processed (or even handled). I thought about trying to give all events an ID and then tracking which events were handled in a given run, which I think we could do in the interpreter, but that doesn't necessarily correspond to e.g. a specific save actually completing (since the event could result in it being handled on the next cycle)

editor/src/components/editor/persistence/persistence.spec.ts
Rheeseyb: Yeah exactly that. Every server save of a project ensures that the local version of it is deleted.

Rheeseyb: (This mimics the current persistence behaviour btw)

editor/src/components/editor/persistence/persistence.ts
Rheeseyb: Done :+1:

seanparsons (approved):

PR #1800, opened 2021-09-08
Fix/webpack terser plugin

enidemi:
**Problem:**
With the upgrade to webpack 5 when running the production build Terser throws errors.

**Fix:**
Updating the terser plugin to the latest version.

**Commit Details:**
- added new version of terser-webpack-plugin
- updated the config, cache and sourceMap are no longer in the options, as I see it generates the sourcemaps by default.
- fixing `REACT_APP_COMMIT_HASH` in the production and staging build scripts


Rheeseyb (approved):

balazsbajorics (commented):
👍

PR #1801, opened 2021-09-08
fix(packages) unpinning babel packages and generating new package-lock

enidemi:
Result of inspecting [this PR](https://github.com/concrete-utopia/utopia/pull/1457) from July: The problem is the node_modules chunk size increased from 12.9 MiB to 15.3 MiB (parsed size) when using the production build. This affected the tsWorker.js chunk similarly.
Looking at it with inspectpack and the bundle analyzer this PR introduced lots of duplicated packages inside babel.
I tried the idea from yesterday about 2 solutions:
1. unpinning babel packages
2. generating a new package-lock

Based on the recommendations from [here](https://github.com/FormidableLabs/inspectpack#duplicates).
Resetting the package-lock was needed to allow npm to find a nicely matching version for the duplicated transitive dependencies.

With this the new size of the node_modules bundle is back to 12.23 MiB.

I'm including the analyzer screenshots because the duplications are very visible.

Old:
Screenshot 2021-09-08 at 18 30 13

New:
Screenshot 2021-09-08 at 18 30 30


maltenuhn:
Wow, that's quite the before/after screenshot!!

balazsbajorics (approved):
carrots for the carets!

Rheeseyb (approved):

PR #1802, opened 2021-09-08
Separate editor logic from workers logic.

### LINKED ISSUE #1796: Refactoring to prevent the editor from pulling in worker logic.

seanparsons:
Fixes #1796 

**Problem:**
The editor pulls in a large amount of logic and dependencies from the workers, which bloats the initial load and download size of the editor itself.

**Fix:**
Guided primarily by the compiler and dependency-cruiser, types and functions imported (or transitively so) from the ts-worker and the parser-printer-worker have been moved to common files or pulled fully into the editor code. A core part of this was to ensure that `lintAndParse` wasn't accessible to the main editor code, so a few places that would pre-parse previously now do not do so. This has resulted in the `typescript` dependency now not ending up in the primary editor bundle, which in development is a saving of about 3MB.

**Commit Details:**
- Fixes #1796.
- Adds several dependency-cruiser rules to prevent things from
  leaking back into the editor.
- Types and some functions from a couple of the workers have moved
  into `worker-types.ts` as that does not pull in the workers itself.
- `simpleDefaultProject` and `complexDefaultProject` now have
  pre-parsed versions available in `simpleDefaultProjectPreParsed`
  and `complexDefaultProjectPreParsed` respectively which are only
  available to tests.
- Moved a couple of types and functions from `parser-printer-comments.ts`
  into `element-template.ts` so they can be shared from there.
- `filterOldPasses` moved into `canvas-wrapper-component.tsx` as
  that is the only place it was used.
- `createTestProjectWithCode` moved to `sample-project-utils.test-utils.ts`.
- Fixed a few tests that had an `unparsed` being passed into
  `textFileContents` paired with `RevisionsState.BothMatch`.
- `JSX_CANVAS_LOOKUP_FUNCTION_NAME` moved into `dom-utils.ts`.
- `LOAD` action now no longer pre-parses the project code before
  completing.
- Removed some old UI builder sample project logic.
- Moved some `isXFile` functions into `file-utils.ts`.
- `appJSFile` and `getDefaultUIJsFile` now no longer pre-parse.
- Moved `parseSuccess` and `parseFailure` functions to live next to
  their types in `project-file-types.ts`.
- Changed the type of `ParseFailure.diagnostics` to hold an array of
  `ErrorMessage` instead of `TS.Diagnostic`.
- Removed `ParsedJSONFailure.errorNode` as it was unused and had a
  type of `TS.Node`.
- Tailwind files injected into a project are no longer pre-parsed.
- Removed `isParsedTextFile` check from `OpenFileEditor`.


Rheeseyb (approved):
Nice work!

balazsbajorics (approved):

enidemi (approved):

PR #1811, opened 2021-09-13
Fix/utopia api dependencies

enidemi:
**Problem:**
The utopia-api has some extra dependencies that could be cleared. React-dom can be a peer dependency and draft-js can be removed as it's no longer fits the editor needs.

**Fix:**
Move react-dom to peer dependencies and update Text component in utopia-api to be a span element. This also means clearing code that was used to handle the autosizing size updates and properties.

**Commit Details:**
- updating package.json in editor and utopia-api
- new generated utopia-api d.ts file
- change utopia-api/Text to render a span
- change insertion logic to add a default text as children to Text elements, this used to be the `text` prop
- removed autosizing helpers and `textSizing` prop
- kept the inspector toggle buttons but they are always in disabled state


seanparsons (approved):

balazsbajorics (approved):

Rheeseyb (commented):
Looks good, but the version number in the utopia-api's `package.json` needs updating please

PR #1823, opened 2021-09-16
fix(editor) Carry over the branch_name parameter into iframe URLs.

seanparsons:
**Problem:**
Currently the iframes used in our editor do not honour the `branch_name` query parameter.

**Fix:**
Carry the `branch_name` parameter over wherever we can.

**Commit Details:**
- Makes sure that our iframes also carry over the `branch_name`
  parameter into their URLs.

### DIFF
--- a/editor/src/components/code-editor/code-editor-container.tsx
+++ b/editor/src/components/code-editor/code-editor-container.tsx
@@ -16,6 +16,10 @@ const VSCodeIframeContainer = betterReactMemo(
     const url = new URL(baseIframeSrc)
     url.searchParams.append('project_id', projectID)
 
+    const branchName = url.searchParams.get('branch_name')
+    if (branchName != null) {
+      url.searchParams.append('branch_name', branchName)
+    }
     return (
       <iframe
         key={'vscode-editor'}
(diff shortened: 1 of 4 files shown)


editor/src/components/code-editor/code-editor-container.tsx
```diff
@@ -16,6 +16,10 @@ const VSCodeIframeContainer = betterReactMemo(
     const url = new URL(baseIframeSrc)
     url.searchParams.append('project_id', projectID)
 
+    const branchName = url.searchParams.get('branch_name')
```
balazsbajorics, line 19: there's a `getEditorBranchNameFromURL` helper function, currently used by no one. I think it's better to use a helper function so we don't have to hunt for 'branch_name' in the codebase

Rheeseyb: Agreed, or at least replace it was a const please

Rheeseyb (approved):

editor/src/components/code-editor/code-editor-container.tsx
seanparsons: Fixed in c96f554a.

balazsbajorics (commented):
👍

PR #1829, opened 2021-09-16
snip the BundlerWorker / ts-worker from the main editor

balazsbajorics:
During my digging through wtf the editor is doing when loading a project, I noticed that the BundlerWorker (also called `ts-worker` to make things confusing) is not producing useful output for the main editor thread. I am simply removing it and no one will notice. It doesn't bring significant loading time benefit, but at least we are using significantly less resources this way!

The property-controls-iframe and the preview.html are still using the ts-worker on their own. I'm not touching that, it's good and fine.

### DIFF
--- a/editor/src/components/canvas/ui-jsx.test-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx.test-utils.tsx
@@ -102,6 +102,7 @@ const FailJestOnCanvasError = () => {
   const stableCallback = React.useCallback((newRuntimeErrors: Array<RuntimeErrorInfo>) => {
     // we have new runtime errors, let's take the tests down
     if (newRuntimeErrors.length > 0) {
+      console.error('Canvas Error!!!!!', newRuntimeErrors[0]?.error)
       fail(newRuntimeErrors[0]?.error)
       expect(newRuntimeErrors[0]?.error ?? null).toEqual(null)
     }
--- a/editor/src/components/canvas/ui-jsx.test-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx.test-utils.tsx
@@ -102,6 +102,7 @@ const FailJestOnCanvasError = () => {
   const stableCallback = React.useCallback((newRuntimeErrors: Array<RuntimeErrorInfo>) => {
     // we have new runtime errors, let's take the tests down
     if (newRuntimeErrors.length > 0) {
+      fail(newRuntimeErrors[0]?.error)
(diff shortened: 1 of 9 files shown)


maltenuhn:
Nice! Does this give us any kind of performance advantage, or mainly delay the heat death of the universe by a bit?

balazsbajorics:
> Nice! Does this give us any kind of performance advantage, or mainly delay the heat death of the universe by a bit?

It certainly reduces the resource usage by a bit, and removes some unnecessary CPU spinning, but I don't think it will have a direct effect on canvas first load speed.

enidemi (approved):

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -103,6 +102,8 @@ const FailJestOnCanvasError = () => {
   const stableCallback = React.useCallback((newRuntimeErrors: Array<RuntimeErrorInfo>) => {
     // we have new runtime errors, let's take the tests down
     if (newRuntimeErrors.length > 0) {
+      console.error('Canvas Error!!!!!', newRuntimeErrors[0]?.error)
```
Rheeseyb, line 105: Hahahaha! Is this needed or is the `fail` enough?

Rheeseyb (approved):

editor/src/components/canvas/ui-jsx.test-utils.tsx
balazsbajorics: it's needed 🤦 
this way it's 100% certain that the CI output contains a readable stack trace and error info

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -103,6 +102,8 @@ const FailJestOnCanvasError = () => {
...
+      fail(newRuntimeErrors[0]?.error)
       expect(newRuntimeErrors[0]?.error ?? null).toEqual(null)
```
seanparsons, line 107: Isn't this unreachable?

seanparsons (approved):

editor/src/components/canvas/ui-jsx.test-utils.tsx
balazsbajorics: this is basically failover, because last time I removed the fail (to clean up unreachable code) and only left the expect, and then a few weeks later when bumping jest it silently regressed to not even being able to fail the test properly... so I'm not comfortable with leaving just the fail there anymore.

PR #1830, opened 2021-09-16
Add a ready indicator for VS Code to the editor state.

seanparsons:
Fixes #1821

**Problem:**
Updates to the project contents were being sent to VS Code before it had been fully initialised, resulting in errors on creating a new project.

**Fix:**
Added a flag to indicate VS Code has fully initialised so that the editor knows it can send those changes.

**Commit Details:**
- Fixes #1821.
- Added `vscodeReady` to `EditorState`.
- In `editorDispatch`, don't send any changes to VS Code unless
  `vscodeReady` is set to `true`.
- Have the `SEND_CODE_EDITOR_INITIALISATION` action mark VS Code
  as ready.
- Wire in the `SEND_CODE_EDITOR_INITIALISATION` action, as it
  wasn't ever previously being fired...

PR #1825, opened 2021-09-16
Feature/vscode editor loading screen

enidemi:
**Fix:**
Updated version of the loading screen. It becomes hidden when the storyboard file is already opened.

**Commit Details:**
- moved loading screen from the iframe to the iframe container inside the editor
- added shimmer to the loading screen
- new type of message for file opened
- first file open in `initVSCodeBridge` fires an action to hide the loading screen
- when the `initVSCodeBridge` is called without a file to open it hides the loading screen without waiting for the opened file message

PR #1831, opened 2021-09-17
fix(vscode) fix when there is no selectedview

enidemi:
**Problem:**
Fixing an error coming from #1825 and #1830

**Fix:**
Check if there are any selectedviews when trying to send it in the action handler `SEND_CODE_EDITOR_INITIALISATION`

**Commit Details:**
- update `getSelectedElementChangedMessage`


Rheeseyb (commented):
:+1:

PR #1840, opened 2021-09-20
Protect against infinite depth content.

### LINKED ISSUE #1590: right now App is filtered out from the insert menu based on name. we should be smarter than that
The insert menu should try to not contain elements that can lead to circular render loops.

I think a good heuristic would be to filter out every element that matches the edited component hierarchy. 

So if we are editing Card focused inside App, filter out App and Card.

seanparsons:
Fixes #1590

**Problem:**
Previously we prevented insertion of an element called `App` as a heuristic to avoid having an infinite render loop in the canvas if that component ever ended up inside itself. But that only works if it matches that name, which is highly unlikely and a complex project could easily have many components which could induce infinite recursion.

**Fix:**
The original intent of the bug was to improve the filtering, but that becomes an increasingly complicated nest of checks especially once an element in the hierarchy has any kind of conditional rendering or if it gets passed in a prop, which may hide it from introspection.

Instead this thresholds the depth of rendering in the canvas, so that an exception will be thrown to prevent it from running forever or crashing the whole editor, with a message indicating that it was because of an excessive render depth.

**Commit Details:**
- Fixes #1590.
- Add a condition to `createComponentRendererComponent` to protect against
  an excessive depth of calls.
- Removed the filtering of `App` from the insertable components.

### DIFF
--- a/editor/src/components/shared/project-components.ts
+++ b/editor/src/components/shared/project-components.ts
@@ -294,10 +294,7 @@ export function getComponentGroups(
         file.fileContents.parsed,
       )
       if (possibleExportedComponents != null) {
-        const componentsWithoutApp_KILLME = possibleExportedComponents.filter(
-          (c) => c.listingName !== 'App', // TODO make this filtering smarter. Filter out the component(s) that we are currently inserting into instead of this
-        )
-        const insertableComponents = componentsWithoutApp_KILLME.map((exportedComponent) => {
+        const insertableComponents = possibleExportedComponents.map((exportedComponent) => {
           const pathWithoutExtension = dropFileExtension(fullPath)
           const { defaultProps, parsedControls } = defaultPropertiesForComponentInFile(
             exportedComponent.listingName,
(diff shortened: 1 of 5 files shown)


editor/src/components/shared/project-components.ts
```diff
@@ -294,10 +294,7 @@ export function getComponentGroups(
         file.fileContents.parsed,
       )
       if (possibleExportedComponents != null) {
-        const componentsWithoutApp_KILLME = possibleExportedComponents.filter(
```
balazsbajorics, line 297: I think the depth detection is a good idea, but I'd keep the filtering out of App, as a fix for an obvious default project footgun.

enidemi (approved):

editor/src/components/shared/project-components.ts
Rheeseyb: I don't think that is a call we should make tbh. The default standard is to use `App` as the main component, but that's not something which is set in stone or required at all. As long as this PR catches that case I'd say that's good enough.

balazsbajorics: well, we will let you insert App and immediately show an exception screen on the canvas, making you undo what you just did. I'd say for the specific case of App, that's a regression in user experience

seanparsons: Well you could argue the inverse, as it was preventing anything from being inserted if it was called `App`.

PR #1843, opened 2021-09-21
fix(canvas) resize snapping for absolute elements

enidemi:
**Problem:**
Resizing an element with position absolute on the edges no longer has snapping.

**Commit Details:**
- use snapped drag rectangle to calculate size delta for absolutely positioned elements.


balazsbajorics (approved):

enidemi:
Closing this as it introduces other issues

PR #1846, opened 2021-09-22
Fix/prettier should run in every folder, not just in editor

### LINKED ISSUE #988: Make prettier work everywhere (performance-test too)

enidemi:
Fixes #988 

**Problem:**
The pre-commit hook only watches for changes editor/src. The existing hook is including utopia-api, but it doesn't do anything now (maybe this was working in the past?).

**Fix:**
Adding the pre-commit hook to the root folder allows it to check in all places.

**Commit Details:**
- updated husky to the latest version
- extended `lint-staged` to run prettier on editor, utopia-api, utopia-vscode-common, performance-test directories
- change `prettier-check` script to include these new directories
- ran prettier once to make the existing code pretty, this effected utopia-vscode-common


Rheeseyb (commented):
This needs an update to `shell.nix` to ensure `npm install` is being called at the root level as part of all developer workflows. It's probably easiest to add this as part of either `start-full` or `install-editor`

balazsbajorics (approved):

shell.nix
```diff
@@ -23,6 +23,7 @@ let
     (pkgs.writeScriptBin "install-editor" ''
       #!/usr/bin/env bash
       set -e
+      ${node}/bin/npm --scripts-prepend-node-path=true install
```
Rheeseyb, line 26:
```suggestion
      cd $(${pkgs.git}/bin/git rev-parse --show-toplevel)
      ${node}/bin/npm --scripts-prepend-node-path=true install
```
I'm pretty sure we'd need to `cd` here to ensure we're in the right directory first

Rheeseyb (approved):
With the proposed `cd` this looks good to me

shell.nix
enidemi: thanks! fixed :)

PR #1856, opened 2021-09-23
Introducing pnpm

balazsbajorics:
**Problem:**
I couldn't bump us to Jest 27 because there were symlinks in the monorepo between packages and Jest were trying to resolve internal dependencies at the wrong place. 
This was a last straw kind of situation, we've been eyeing with pnpm before, as it promises to solve a lot of monorepo-related pains. 

The main promise of pnpm is that it gives us much more control about versions of packages in node_modules. It gives a much better understanding of transitive dependencies and conflicting transitive dependencies too.

Here's an overview of how it works https://pnpm.io/symlinked-node-modules-structure

## PNPM TLDR
the main difference between `npm` and `pnpm` is that npm tries to flatten the node_modules. This means that npm will try to put transitive dependencies in the node_modules/ root as well. In case of conflicting transitive dependencies, it will put them into sub-modules inside node_modules, such as `node_modules/package-a/node_modules/transitive-package`. This can then lead to npm doppelgangers https://rushjs.io/pages/advanced/npm_doppelgangers/

Instead of this, what PNPM does is that it creates a pnpm repo at `~/.pnpm` and installs every package there. in the specific `node_modules` folders, it only uses symlinks to point to the main repo. The benefit is that the `node_modules` folder doesn't need to be flattened in order to save space / install size. The structure can be a deep tree with every specific transitive dependency placed where it is linked in a package.json. So if a package has _any_ transitive dependency, it will be installed as `node_modules/package-a/node_modules/transitive-package` but `transitive-package` will just be a symlink to `~/.pnpm/transitive-package@<version-number>`. 

In my opinion this is a best of both worlds kind of solution. It saves space, and makes the node_modules structure easier to debug. It also means that the node_modules structure is "strict": if you try to import/require a package that is not listed in your dependencies or devDependencies, it will fail because the node module resolution will not find it. If you are using npm, by accident in the flattened npm-type node_modules will contain a thousand transitive dependencies that can be required and will resolve, and you will never know that you are using something that is not listed as a direct dependency.

**Fix:**
Throw out npm and use pnpm

**Commit Details:**
- Use pnpm in shell.nix instead of npm
- Add `"preinstall": "npx only-allow pnpm"` which throws a helpful error message if someone tries to use npm because of muscle memory
- Fix a conflicting type issue with emotion/serialize 
- Delete package-lock.json's and add pnpm-lock.yaml instead


.github/workflows/pull-requests.yml
```diff
@@ -155,14 +155,26 @@ jobs:
           webpackStatsFile: editor/lib/staging-stats.json
           key: ${{ secrets.RELATIVE_CI_KEY }}
           debug: false
+      - name: Delete node_modules
```
seanparsons, line 158: Does this not mean we lose the cached content of the `node_modules` folders?

balazsbajorics: I just pushed 40658c809 and 2a9ba9909 which caches `~/.pnpm` instead of the node_modules folders. Since the node_modules folders are just empty symlinks, there's no need caching them anymore.

seanparsons (approved):

seanparsons (approved):

enidemi (approved):

PR #1860, opened 2021-09-24
Fixed up the husky install

Rheeseyb:
**Problem:**
#1846 Ensured that linting (and prettier) are applied in all projects as part of the commit hook, but was missing a required `prepare` script to actually install the commit hook.

**Fix:**
Add the `prepare` script to the root `package.json` and remove the old (incorrectly placed and now referring to a file that no longer exists) one from `editor/package.json`


enidemi (approved):

balazsbajorics (commented):
👍

PR #1863, opened 2021-09-24
Make browser spec tests await the first DOM report

### LINKED ISSUE #1766: Fix the race condition in the periodically failing test
One of our tests around updating a frame periodically fails on CI due to a race condition. I'm creating this ticket now so that it is tracked, and the next time it fails I'll update it with the name and file of the test, since annoyingly there doesn't appear to be a way to view the logs of old workflow runs after clicking "re-run" on them.

Rheeseyb:
Fixes #1766 (I think)

**Problem:**
I believe the issue at play here is that we don't await the first DOM report before attempting to update the model in our `.spec.browser` tests, implying that we're not awaiting the correct metadata or (I believe) a complete rendered canvas.

This is just a theory because I'm not entirely sure how to verify it, but I have verified that in tests that select a component and then call e.g. `const areaControl = renderResult.renderedDOM.getByTestId('ccc')`, those tests will still pass even if the initial selection is removed, leading me to believe that the presence of that test ID does not necessarily mean the component has been rendered or is selected.

**Fix:**
Update the function `renderTestEditorWithModel` to take a second param `awaitFirstDomReport: 'await-first-dom-report' | 'dont-await-first-dom-report'`, which determines whether or not the function blocks until the first DOM report is received. Then, for all `spec.browser` tests that call this we block on that first report, and for all non-browser tests we don't.

### DIFF
--- a/editor/src/components/canvas/canvas-actions.spec.tsx
+++ b/editor/src/components/canvas/canvas-actions.spec.tsx
@@ -30,6 +30,7 @@ describe('createDragState', () => {
         </div>
       </div>
       `),
+      'dont-await-first-dom-report',
     )
 
     const selectedViews = [EP.appendNewElementPath(TestScenePath, ['aaa', 'bbb'])]
--- a/editor/src/components/canvas/ui-jsx.test-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx.test-utils.tsx
@@ -114,15 +114,25 @@ const FailJestOnCanvasError = () => {
   return null
 }
 
-export async function renderTestEditorWithCode(appUiJsFileCode: string) {
-  return renderTestEditorWithModel(createTestProjectWithCode(appUiJsFileCode))
+export async function renderTestEditorWithCode(
+  appUiJsFileCode: string,
+  awaitFirstDomReport: 'await-first-dom-report' | 'dont-await-first-dom-report',
+) {
+  return renderTestEditorWithModel(createTestProjectWithCode(appUiJsFileCode), awaitFirstDomReport)
 }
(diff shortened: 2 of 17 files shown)


seanparsons (approved):

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -260,6 +270,10 @@ export async function renderTestEditorWithModel(
     )
   })
 
+  if (awaitFirstDomReport === 'await-first-dom-report') {
```
balazsbajorics, line 273: have you tried just always awaiting for this, and running all tests? I can't come up with any test that should use `renderTestEditorWithModel` and not await the first dom report

Rheeseyb: Yep. The non-browser tests failed (or at least some of them did).

enidemi (approved):

editor/src/components/canvas/canvas-actions.spec.tsx
```diff
@@ -30,6 +30,7 @@ describe('createDragState', () => {
         </div>
       </div>
       `),
+      'dont-await-first-dom-report',
```
Rheeseyb, line 33:
```suggestion
      'dont-await-first-dom-report'
```
Minor change to trigger workflows

editor/src/components/canvas/canvas-actions.spec.tsx
```diff
@@ -30,6 +30,7 @@ describe('createDragState', () => {
         </div>
       </div>
       `),
+      'dont-await-first-dom-report'
```
Rheeseyb, line 33:
```suggestion
      'dont-await-first-dom-report',
```

PR #1864, opened 2021-09-28
Chore/jest 27

balazsbajorics:
This PR bumps Jest from 26 to 27. Thanks to pnpm and the karma runner, there were barely any problems remaining here to be solved.

**Commit Details:**
- Jest 26 -> 27
- Removing traces of ts-jest, even though I believe we were not using it anymore.
- annoyingly, changing the values of the performance regression render count test.


editor/package.json
```diff
@@ -41,28 +41,6 @@
     "fetch-tailwind-classes": "npx -s sh ts-node --pretty --require ignore-styles --transpile-only ./src/scripts/fetch-and-parse-tailwind-classes.ts && npx prettier --write ./src/core/third-party/tailwind-defaults.ts",
     "lint-editor": "lint-staged"
   },
-  "jest": {
```
balazsbajorics, line 44: this was not used since we moved to jest.config.js, I removed it to avoid confusing people

editor/src/components/editor/persistence/persistence.spec.ts
```diff
@@ -34,7 +34,7 @@ let serverProjects: { [key: string]: PersistentModel } = {}
 let localProjects: { [key: string]: LocalProject<PersistentModel> } = {}
 
 const base64Contents = 'data:asset/xyz;base64,SomeBase64'
-const AssetFileWithBase64 = assetFile(base64Contents)
+const mockAssetFileWithBase64 = assetFile(base64Contents)
```
balazsbajorics, line 37: babel-jest is more nitpicky about variable names that are in the scope of the mock.

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -237,7 +237,7 @@ describe('React Render Count Tests - ', () => {
     )
 
     const renderCountAfter = renderResult.getNumberOfRenders()
-    expect(renderCountAfter - renderCountBefore).toBeGreaterThanOrEqual(2725) // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toBeLessThan(2735)
+    expect(renderCountAfter - renderCountBefore).toBeGreaterThanOrEqual(1900) // if this breaks, GREAT NEWS but update the test please :)
```
balazsbajorics, line 240: I -think- this happens because something changed around module resolution, and some React components doesn't see the monkey patched react that counts renders

seanparsons:
Does this make any difference to the performance of the tests?

editor/src/core/performance/performance-regression-tests.spec.tsx
seanparsons: When this shot up previously I captured the components that were being rendered by modifying the patched React function, might be worth doing that to confirm your theory?

balazsbajorics:
> Does this make any difference to the performance of the tests?

unfortunately, no.

we got a speed gain when leaving ts-jest behind (but I forgot I already did that in the beginning of the summer)

there's some suspiciously slow things that I want to look into tomorrow, and hopefully with everything bumped to latest it will be easier to figure out if something is wrong / misconfigured

enidemi (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
balazsbajorics: #1872 fixed this a lot, it's still a slightly smaller number, but I think it's acceptable now

seanparsons (approved):

PR #1865, opened 2021-09-28
Feature/karma snapshot tests

enidemi:
**Problem:**
The electron browser tests are not supporting the latest jest, it's time to use an alternative solution.

**Fix:**
Switching the browser tests to use karma.

**Commit Details:**
- added pretty format from jest
- added karma-viewport to set size
- introduced internal inline snapshot test
- switched prettier import to use prettier/standalone in browser tests
- removing `setElectronWindow`
- changed dom-walker tests to use inline snapshots
- test case called: `dragging a scene child’s root view sets the root view position` is temporarly removed as it takes longer than 10 seconds to complete


editor/src/components/canvas/canvas-utils-scene.spec.browser2.tsx
```diff
@@ -106,7 +105,7 @@ describe('moving a scene/rootview on the canvas', () => {
       await domFinished
       await dispatchDone
     })
-
+    await wait(1000)
```
balazsbajorics, line 108: accidental await wait
```suggestion
```

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -15,7 +14,10 @@ import CanvasActions from '../../canvas-actions'
 import { CanvasControlsContainerID } from '../new-canvas-controls'
 
 describe('Select Mode Selection', () => {
-  beforeAll(setElectronWindow)
+  // TODO find types for Mocha `before`
```
balazsbajorics, line 17: @enidemi I guess we should fix this before merging!

enidemi: fixed

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper-third-party.spec.browser2.tsx
```diff
@@ -37,6 +21,17 @@ import {
 import { addFileToProjectContents } from '../../assets'
 import { StoryboardFilePath } from '../../editor/store/editor-state'
 import { applyUIDMonkeyPatch } from '../../../utils/canvas-react-utils'
+import { BuiltInDependencies } from '../../../core/es-modules/package-manager/built-in-dependencies-list'
+import { matchInlineSnapshotBrowser } from '../../../../test/karma-snapshots'
+
+BuiltInDependencies.push({
```
balazsbajorics, line 27: this replaces the old jest.mock . while we are looking for a working module mocking solution, I realized that in this particular case we did not need one.

seanparsons (approved):

PR #1881, opened 2021-10-05
Re-enable the R3F test.

seanparsons:
**Problem:**
We want to ensure that React Three Fiber continues to render properly in the editor.

**Fix:**
The primary change is to introduce a `wait` call after the render call is made. This was done because R3F appears to do some internal render pass after the React render, so the `wait` means it gets a chance to trigger and the spy metadata to get its update.

**Commit Details:**
- Add a `wait` to the test, so that R3F gets a chance to render internally.
- Fixed up the snapshot values in the test.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper-third-party.spec.browser2.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper-third-party.spec.browser2.tsx
@@ -54,7 +54,7 @@ const exampleFiles = {
       </div>
     )
   };
-  
+
   export var storyboard = (
     <Storyboard data-uid="storyboard">
       <Scene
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper-third-party.spec.browser2.tsx
```diff
@@ -146,99 +145,57 @@ xdescribe('Spy Wrapper Tests For React Three Fiber', () => {
...
           "name": "ambientLight",
-          "rootElements": Array [],
         },
         "storyboard/scene-1/canvas-app:canvas-app-div/test-directionalLight": Object {
-          "children": Array [],
           "name": "directionalLight",
-          "rootElements": Array [],
         },
         "storyboard/scene-1/canvas-app:canvas-app-div/test-mesh": Object {
-          "children": Array [
```
balazsbajorics, line 197: what's with these changes? did we remove the children and rootElements fields since this test was disabled?

seanparsons: `simplifiedMetadataMap` strips them from the output.

balazsbajorics (approved):

enidemi (approved):

Rheeseyb (commented):
:+1:

PR #1882, opened 2021-10-05
Prevent npm from including peer dependencies on the server

Rheeseyb:
**Problem:**
`@react-three/drei` was causing issues in Utopia, because the peer dependencies for `react` and `react-dom` were resulting in multiple versions of react, which in turn was causing an R3F hooks error. This was previously not a problem, but an npm update changed the handling of peer dependencies such that they are now installed by default.

**Fix:**
Include the argument `--omit peer` when calling `npm install` on the server.


enidemi (approved):

seanparsons (approved):

balazsbajorics (commented):
beautiful

PR #1960, opened 2021-10-29
`registerControls` function in utopia-api

enidemi:
**Problem:**
We want to let the users register `propertyControls` descriptors for components imported from npm and other third party sources. Today we only support hard-coded third-party controls, which require forking utopia and making a pull request to extend.

**Fix:**
A new function called `registerControls` which can be imported from utopia-api.

**Note:**
The API is still in flux.

Usage: `registerControls(componentName: string, packageName: string, propertyControls: PropertyControls): void`
```javascript
import { registerControls } from 'utopia-api'


// Register a control for an npm package named `antd` for the `Row` component
registerControls('Row', 'antd', {
  align: {
    control: 'popuplist',
    label: 'align',
    options: ['top', 'middle', 'bottom'],
    defaultValue: 'top',
  },
  justify: {
    control: 'popuplist',
    label: 'justify',
    options: ['start', 'end', 'center', 'space-around', 'space-between'],
    defaultValue: 'center',
  },
})

// Register a control for the App component from the `/src/app.js` file inside the current utopia project
registerControls('App', '/src/app', {
  firstName: { control: 'string-input' },
  lastName: { control: 'string-input' },
  age: { control: 'number-input' },
})
```

you can register multiple components per file. The file that has the `registerControls` calls needs to be imported to be evaluated. It can be imported from `storyboard.js`, or any other file imported by `storyboard.js`.

**Details:**
- add a NO_OP `registerControls` function to `utopia-api`
- Move `builtInDependencies` to the EditorState – this is a massive change touching on a lot of test cases as well. Most of the diff is coming from this change
- `createBuiltInDependenciesList` replaces the native `registerControls` function of `utopia-api` with a custom function created by `createRegisterControlsFunction`
- `createRegisterControlsFunction` returns a function which has access to the editor state and dispatch.
- Now, when userland code imports `import { registerControls } from 'utopia-api'` and calls it, the function will dispatch an `UPDATE_PROPERTY_CONTROLS_INFO` action, updating the EditorState.propertyControlsInfo.
- We are updating the running editor's EditorState from userland code! This might come handy for other things in the future too.

**TODO:**
- [x] a basic regression test to make sure things work
- [ ] `registerControls` should also make the registered components show up in the Insert Menu (next PR)


editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -0,0 +1,63 @@
...
+  projectContents: ProjectContentTreeRoot,
+): PropertyControls | null {
+  const packageJsonFile = packageJsonFileFromProjectContents(projectContents)
+  const dependencies = dependenciesFromPackageJson(packageJsonFile, 'combined')
+  const foundPackageWithElement = Object.keys(propertyControlsInfo).find((key) => {
+    return propertyControlsInfo[key][elementName] != null
+  })
+  if (
+    foundPackageWithElement != null &&
+    dependencies.some((dependency) => dependency.name === foundPackageWithElement)
```
Rheeseyb, line 58: Is this dependency check necessary? If the intention here is to ensure we don't serve up controls for an incorrect package, then it should probably be rolled into the above `find`.

enidemi: fixed

PR #1964, opened 2021-11-01
Feature/property control slight refactor 2

Rheeseyb:
**Problem:**
We had multiple property controls for selection a value from some kind of list of options, which all had similar but differing interfaces (specifically referring to the `popuplist`, `radio`, and `expression-popuplist` controls).

**Fix:**
The most significant changes in this PR take place in `utopia-api/src/property-controls/property-controls.ts`
- Introduced the type `BasicControlOption<T>` for representing a value and its label, and then used `AllowedEnumType[] | BasicControlOption<T>[]` for the type of options in a `popuplist` and a `radio` control, to align those two control types.
- Tweaked and renamed `ExpressionEnum` to `ExpressionControlOption<T>`, to also make `expression-popuplist`'s interface consistent with `popuplist`'s and `radio`'s.
- Updated all current uses of these controls
- Updated some tests


utopia-api/src/property-controls/property-controls.ts
```diff
@@ -42,12 +42,16 @@ export interface ColorControlDescription extends AbstractBaseControlDescription<
...
+}
+
+export type BasicControlOptions<T> = AllowedEnumType[] | BasicControlOption<T>[]
 
 export interface PopUpListControlDescription extends AbstractBaseControlDescription<'popuplist'> {
-  defaultValue?: AllowedEnumType
-  options: AllowedEnumType[]
-  optionTitles?: string[] | ((props: unknown | null) => string[])
-  displaySegmentedControl?: boolean
+  defaultValue?: unknown
```
balazsbajorics, line 53: why not a `T` here?

balazsbajorics (approved):

utopia-api/src/property-controls/property-controls.ts
Rheeseyb: tbh I didn't really think it would be worth it. I think using the `T` above is possibly a bit over-engineered since I don't think there would be any way for it to provide any actual use in practice

balazsbajorics: this is the user-facing type definition, and for that it makes sense to make it a T.

when parsing it doesn't matter at all

PR #1966, opened 2021-11-01
fix(inspector) Inferring a hex requires the hash prefix

### LINKED ISSUE #1965: Strings should not be inferred as a hex color if there is no hash prefix
When inferring the property controls for a selected element, if a property has a string value that could represent a valid hex color, **but without the leading hash**, the Inspector will incorrectly show a color control for it, rather than a string control.

Rheeseyb:
Fixes #1965 

**Problem:**
The Inspector was incorrectly inferring strings without a leading hash as a hex color if the string could be parsed as a hex in the absence of that leading hash.

**Fix:**
This was caused by us relying on color parsing logic that deliberately allows an absent hash when typing in a hex (for e.g. use when typing a hex code into the color control). To support both use cases, I have updated the parsing function to take a second param which explicitly states whether the hash is required of optional, and then used the stricter form when inferring the property controls to render.


enidemi (approved):

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -2178,8 +2178,37 @@ function printBackgroundColor(value: CSSDefault<CSSSolidColor>): JSXAttributeVal
 
 const matchColorKeyword = combineRegExp(['^', '(', RegExpLibrary.colorKeyword, ')', '$'])
 const matchColorHex = combineRegExp(['^', RegExpLibrary.colorHexOptionalOctothorp, '$'])
+const matchColorHexStrict = combineRegExp(['^', RegExpLibrary.colorHex, '$'])
+
+function parseHexColor(
+  color: string,
+  strictHash: 'hex-hash-required' | 'hex-hash-optional',
```
balazsbajorics, line 2185: thanks for using a verbose flag :)

balazsbajorics (approved):

PR #1968, opened 2021-11-02
Fix Color Control HSL Printing

### LINKED ISSUE #1967: Color Control HSL printing is broken
If a color is specified using HSL, the Inspector's Color Control will correctly display the color, but any attempts to update the color will lead to an incorrect HSL string. If the color is a shade of grey, it will also replace the hue with `NaN`.

Rheeseyb:
Fixes #1967 

**Problem:**
Chroma's `color.hsl()` function returns the saturation and lightness as values between 0 and 1, rather than percentages, and returns `NaN` for the hue when the color is a shade of grey.

**Fix:**
Safeguard against `NaN` hues by using the already stored state hue value, and convert the saturation and lightness values to percentages.


balazsbajorics (approved):

maltenuhn:
I'm almost afraid to ask, but... did we have a different way of guarding against that in the background/font color sections of the inspector? Or did this also affect us there, we just hadn't run into it yet?

Rheeseyb:
> I'm almost afraid to ask, but... did we have a different way of guarding against that in the background/font color sections of the inspector? Or did this also affect us there, we just hadn't run into it yet?

It was there all along.

enidemi (approved):

PR #1972, opened 2021-11-02
Ensure the evaluationCache is cleared on transitive dependencies

### LINKED ISSUE #1969: The `evaluationCache` isn't correctly updated on changes to a transitive dependency
The `evaluationCache` that we use to prevent re-evaluating files repeatedly is keyed on the file's content. This works well to ensure we always blow the cache when making a change to a file directly imported by the storyboard, but it means that we aren't blowing the cache for any files along the dependency chain for changes to a file indirectly imported by the storyboard.

e.g. in the case where `/utopia/storyboard.js` imports from `/src/app.js`, which imports from `/src/thing.js`, a change to `/src/thing.js` would not be reflected on the canvas without also making a change to `/src/app.js`

Rheeseyb:
Fixes #1969 

**Problem:**
When making changes to a file, the bundler worker was previously relied upon to clear all cached evaluations of transitive dependencies of that file. When we removed the bundler worker from the main editor, we regressed and lost that functionality.

**Fix:**
Restore the cache clearing, this time incorporating it into the `editorDispatch` via a newly added `applyProjectChanges` function. This function uses the `AccumulatedVSCodeChanges` (which includes an array of all changes made to files as part of this action dispatch handling), allowing us to call any side-effecting (yuck, evil!) code that should result from those project content changes.

**TODO**

- [ ] Add tests

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -510,9 +510,31 @@ export function editorDispatch(
   }
 
   const newVSCodeChanges = getVSCodeChanges(storedState.editor, frozenEditorState)
+  applyProjectChanges(frozenEditorState, newVSCodeChanges, updatedFromVSCode)
+
+  const shouldUpdatePreview =
+    anySendPreviewModel || frozenEditorState.projectContents !== storedState.editor.projectContents
+  if (shouldUpdatePreview) {
+    updateEmbeddedPreview(frozenEditorState.id, frozenEditorState.projectContents)
+  }
+
+  if (frozenEditorState.id != null && frozenEditorState.id != storedState.editor.id) {
+    storedState.workers.initWatchdogWorker(frozenEditorState.id)
+  }
+
+  return finalStore
+}
+
+function applyProjectChanges(
+  frozenEditorState: EditorState,
+  accumulatedChanges: AccumulatedVSCodeChanges,
+  updatedFromVSCode: boolean,
+) {
+  triggerPropertyControlsIframeIfNeeded(frozenEditorState, accumulatedChanges.fileChanges)
+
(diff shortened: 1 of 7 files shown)


editor/src/components/editor/store/dispatch.tsx
```diff
@@ -510,9 +510,31 @@ export function editorDispatch(
...
+  if (frozenEditorState.id != null && frozenEditorState.id != storedState.editor.id) {
+    storedState.workers.initWatchdogWorker(frozenEditorState.id)
+  }
+
+  return finalStore
+}
+
+function applyProjectChanges(
+  frozenEditorState: EditorState,
+  accumulatedChanges: AccumulatedVSCodeChanges,
```
balazsbajorics, line 530: is `accumulatedChanges` a good name? this are the `getVSCodeChanges` for the current dispatch. does it mean accumulated changes of the current dispatch? 
I think the naming is backwards with `currentVSCodeChanges = combineAccumulatedVSCodeChanges(...` because `currentVSCodeChanges` is the _real_ accumulated changes waiting to be processed

balazsbajorics (approved):

editor/src/components/editor/store/dispatch.tsx
Rheeseyb: Yeah, I'd agree. I went with this name because of the name of the type, but I'm tempted to change that now...

Rheeseyb: I believe the reasoning for the type's name btw was since it is an "accumulation" (in this case meaning "combination") of the file changes, highlight changes, and any selection change. But I agree it has become confusing.

balazsbajorics (approved):

enidemi (approved):

PR #1973, opened 2021-11-03
feature(components) component with registerControls is added to the insert menu list

enidemi:
**Problem:**
Components using registerControls should appear as insertable elements.

**Fix:**
Using the propertyControlsInfo from the editorState to create insert items. Previously it was using built-in component descriptions with import details but this was missing from the propertyControlsInfo. It is changed to contain an optional info object with import details.

**Commit Details:**
- updated `getComponentGroups` used by the insert menus (sidebar, popup) to use property controls info to create insertion groups.
- changed `registerControls` with optional `requiredImports` parameter
- `ComponentDescriptor` is changed to have utopia-api ImportType instead of ImportDetails to avoid double conversion.

### DIFF
--- a/editor/src/components/shared/project-components.ts
+++ b/editor/src/components/shared/project-components.ts
@@ -1,6 +1,7 @@
 import { PropertyControls } from 'utopia-api'
 import { URL_HASH } from '../../common/env-vars'
 import { defaultPropertiesForComponentInFile } from '../../core/property-controls/property-controls-utils'
+import { mapArrayToDictionary } from '../../core/shared/array-utils'
 import { flatMapEither, foldEither, right } from '../../core/shared/either'
 import {
   emptyComments,
@@ -19,6 +20,7 @@ import {
   PossiblyUnversionedNpmDependency,
 } from '../../core/shared/npm-dependency-types'
 import {
+  importDetailsFromImportOption,
   Imports,
   isParsedTextFile,
   isParseSuccess,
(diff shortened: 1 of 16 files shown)


balazsbajorics (approved):

editor/src/components/shared/project-components.ts
```diff
@@ -391,19 +399,33 @@ export function getComponentGroups(
...
-        )
-        const components =
-          dependencyStatus === 'loaded' ? possibleComponentDescriptor.components : []
+      const propertyControlsForDependency = propertyControlsInfo[dependency.name]
+      if (propertyControlsForDependency != null) {
+        const components = Object.keys(propertyControlsForDependency).map((name) => {
+          const elementNameAsArray = name.split('.')
+          const elementName =
+            elementNameAsArray.length > 0
+              ? jsxElementName(elementNameAsArray.shift()!, elementNameAsArray)
```
Rheeseyb, line 414: Rather than doing this, please can we do:
```
const [baseVariable, ...propertyPathParts] = elementNameAsArray
...
jsxElementName(baseVariable, propertyPathParts)
```

editor/src/core/third-party/antd-components.ts
```diff
@@ -23,10 +23,14 @@ function createBasicComponent(
...
+    [
+      {
+        source: 'antd',
+        name: baseVariable,
+        type: null,
+      },
+      { source: 'antd/dist/antd.css', name: null, type: null },
+    ],
     jsxElementWithoutUID(jsxElementName(baseVariable, propertyPathParts), defaultAttributes, []),
     name,
```
Rheeseyb, line 35: Looks like this `name` is no longer used anywhere after this change (looking at the `project-components.spec` snapshot change)

Rheeseyb (approved):

editor/src/core/third-party/antd-components.ts
balazsbajorics: I plan to pretty much delete this system of `DependencyBoundDescriptors`, because insertion will use `EditorState.propertyControlsInfo` and my PR will feed propertyControlsInfo from another (simpler) source.

editor/src/components/shared/project-components.ts
balazsbajorics: pushed a fix

enidemi: thanks, this is much nicer

PR #1990, opened 2021-11-09
Simplify the type declaration of ControlTypes

balazsbajorics:
**Problem:**
The control-types.ts file is using a lot of indirection and abstract base interfaces. This makes it hard to "parse" if someone just opens this file looking for the definition of control types.

**Fix:**
Flatten the file, list every field for every control type, even the common ones.

**Commit Details:**
- Instead of the abstract base control type, I made a new `ControlBaseFields` that lists the 4 shared fields of every control
- `getControlSharedFields` function takes a `RegularControlDescription` and returns a `ControlBaseFields`. This makes sure that if someone comes up with a new control type that has one of the ControlBaseFields missing, we break the build. So we get the same level of type safety as today.


utopia-api/src/property-controls/property-controls.ts
```diff
@@ -21,71 +21,84 @@ export type BaseControlType =
...
+  visibleByDefault?: boolean
   defaultValue?: unknown
   options: BasicControlOptions<unknown>
 }
 
-export interface ImportType {
+export type ImportType = {
   source: string // importSource
   name: string | null
   type: 'star' | 'default' | null
```
maltenuhn, line 59: If this is for version bounds, maybe something more evocative? like `versioning`, with descriptors like  'any' | 'latest' | 'default'? Also what's the difference between `null` and `default` ?

utopia-api/src/property-controls/property-controls.ts
```diff
@@ -21,71 +21,84 @@ export type BaseControlType =
...
+export type ColorControlDescription = {
+  control: 'color'
+  label?: string
+  visibleByDefault?: boolean
   defaultValue?: string
 }
 
 export type AllowedEnumType = string | boolean | number | undefined | null
-export interface BasicControlOption<T> {
+export type BasicControlOption<T> = {
```
seanparsons, line 41: Why are all these changing from `interface` to `type`?

utopia-api/src/property-controls/property-controls.ts
balazsbajorics: this is for:

star: import * as Something from
default: import Something from
null: import { Something } from

utopia-api/src/property-controls/property-controls.ts
Rheeseyb: Yeah, please can we stick with `interface`

seanparsons (approved):

utopia-api/src/property-controls/property-controls.ts
balazsbajorics: sure thing

utopia-api/src/property-controls/property-controls.ts
```diff
@@ -21,71 +21,84 @@ export type BaseControlType =
...
+  label?: string
+  visibleByDefault?: boolean
   defaultValue?: never
 }
 
-export interface Matrix3ControlDescription extends AbstractBaseControlDescription<'matrix3'> {
+export type Matrix3ControlDescription = {
+  control: 'matrix3'
+  label?: string
+  visibleByDefault?: boolean
```
Rheeseyb, line 94: we can use `// prettier-ignore` to further improve the matrix controls like so:
```ts
// prettier-ignore
defaultValue?: [
  number, number, number,
  number, number, number,
  number, number, number,
]
```

utopia-api/src/property-controls/property-controls.ts
balazsbajorics: @Rheeseyb I'm actually not sure about this one: `null: import { Something } from`. did i guess it right? :) maybe we could rename it to "named"

utopia-api/src/property-controls/property-controls.ts
```diff
@@ -216,6 +258,21 @@ export type RegularControlDescription = BaseControlDescription | HigherLevelCont
...
+}
+function getControlSharedFields(control: RegularControlDescription): ControlBaseFields {
+  return {
+    control: control.control,
+    label: control.label,
+    defaultValue: control.defaultValue,
+    visibleByDefault: control.visibleByDefault,
+  }
+}
+
```
Rheeseyb, line 275: I think this should either live at the very top or very bottom of the file, with a comment explaining its purpose.

utopia-api/src/property-controls/property-controls.ts
Rheeseyb: It is indeed "named". This was changed fairly recently to align with the type used by the parser printer since this needs to be injected into the parsed model, so changing the `type` would also require a change a little later down the line when it comes to trying to use this import.

It's definitely worth doing, but maybe as a second PR immediately after this, because I'm also thinking we should clarify when the `name` field would be `null`, and it would also be better off making separate `ImportType`s to clarify that a bit further. e.g.:
```ts
// import * as name from source
interface ImportStar {
  source: string,
  name: string,
  type: 'star',
}

// import source
interface ImportScript { // Not sure of the best name for this one
  source: string,
  type: 'source',
}
...

type ImportType = ImportStar | ImportNamed | ImportDefault | ImportScript
```

enidemi (approved):

utopia-api/src/property-controls/property-controls.ts
balazsbajorics: I moved it to the top.

utopia-api/src/property-controls/property-controls.ts
balazsbajorics: done!

PR #1995, opened 2021-11-09
Update VS Code build to 1.62.0

### LINKED ISSUE #1945: Update our build of VS Code
Utopia uses a custom build script to build VS Code for the web, as well as a patch that we apply. This specifies a version of VS Code to build, which we need to update.

Right now this is blocking the build on the latest Apple hardware, as that wasn't stable until a later version than we're using.

**Note:** This ticket needs to be repeated each month to coincide with the VS Code release cadence.

Rheeseyb:
Fixes #1945 

**Problem:**
Our version of VS Code had fallen far behind

**Fix:**
Update it to use the latest version (1.62.0), updating the patch along the way. Note that viewing the patch as a diff is going to be hard work, so you're best off comparing the before and after separately, though there isn't much of interest to see there (with the exception of me removing the unused dummy services I had created previously, because they no longer compile).

**Note:**
Right now we are still patching the main `gulp.vscode.js` file to allow us to build for the web, but there has been a lot of activity recently towards opening the browser build back up (meaning that [`gulp.vscode.web.js`](https://github.com/microsoft/vscode/blob/main/build/gulpfile.vscode.web.js) is no longer empty), which hopefully means we will soon be able to remove a lot of this patch. See https://github.com/microsoft/vscode/issues/135856 for the work around that.

### DIFF
--- a/vscode-build/build.js
+++ b/vscode-build/build.js
@@ -5,7 +5,7 @@ const fse = require('fs-extra')
 const glob = require('glob')
 const rmdir = require('rimraf')
 
-const vscodeVersion = '1.53.0'
+const vscodeVersion = '1.62.0'
 
 if (fs.existsSync('vscode')) {
   process.chdir('vscode')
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):
kind of a blind approve

maltenuhn:
> there has been a lot of activity recently towards opening the browser build back up

That's reassuring. Also, thanks for including that in the PR description, helpful to keep tabs on things!

seanparsons (approved):

PR #1996, opened 2021-11-09
Maintain selection while editing arbitrary elements.

### LINKED ISSUE #1986: Editor loses selection sometimes
Problem

The editor sometimes loses selection when editing CSS properties

https://utopia.app/p/a9781ba2-abstracted-sloth/

https://user-images.githubusercontent.com/2945037/140741842-78c08f77-bc9d-486c-ae70-e02c6aef7305.mp4

seanparsons:
Fixes #1986

**Problem:**
The UID fixing logic that attempts to keep `data-uid` values consistent between parse passes on a file did not peer inside arbitrary blocks, which meant that any edit would result in slight edits changing the path to address them.

**Fix:**
The logic within `fixParseSuccessUIDs` now includes the case of drilling into `JSXArbitraryBlock` instances. Also `getHighlightBoundsForFile` includes a fallback case so that if the parse fails while mid-typing the old highlights are used, which should prevent the selection flickering off and back on again.

A previous fix to try and ensure that backspaces in VS Code were handled correctly has been resurrected as well, with an additional catch around the handler for that logic to protect against shenanigans.

**Commit Details:**
- Fixes #1986.
- Remove the check for `event.kind` in `initMessaging` so that backspaces
  definitely get included.
- Add a catch inside `cursorPositionChanged` as a safety measure against
  failures in there.
- Added `walkElementsWithin` to handle `ElementsWithin` for an arbitrary
  block similarly to `walkElementChildren`.
- Factored out `compareAndWalkElements` from `walkElementChildren` so that
  it could be used in `walkElementsWithin`, adding an additional clause for
  `JSXArbitraryBlock` instances.
- `getHighlightBoundsForFile` now includes a fallback case to the previous parse result.
- Removed `getHighlightBoundsForUids` as it's unused.
- Removed `removeNonExistingViewReferencesFromState` and `elementPathStillExists`
  as they are unused.

### DIFF
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -1880,22 +1880,18 @@ export function getStoryboardElementPathFromEditorState(
   )
 }
 
-export function getHighlightBoundsForUids(editorState: EditorState): HighlightBoundsForUids | null {
-  const selectedFile = getOpenFile(editorState)
-  if (isTextFile(selectedFile)) {
-    return getHighlightBoundsFromParseResult(selectedFile.fileContents.parsed)
-  }
-
-  return null
-}
-
 export function getHighlightBoundsForFile(
   editorState: EditorState,
   fullPath: string,
 ): HighlightBoundsForUids | null {
   const file = getContentsTreeFileFromString(editorState.projectContents, fullPath)
-  if (isTextFile(file) && isParseSuccess(file.fileContents.parsed)) {
-    return getHighlightBoundsFromParseResult(file.fileContents.parsed)
+  if (isTextFile(file)) {
+    if (isParseSuccess(file.fileContents.parsed)) {
+      return getHighlightBoundsFromParseResult(file.fileContents.parsed)
+    }
+    if (file.lastParseSuccess != null) {
+      return getHighlightBoundsFromParseResult(file.lastParseSuccess)
+    }
   }
   return null
 }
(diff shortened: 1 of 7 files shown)


editor/src/components/editor/store/editor-state.ts
```diff
@@ -1880,22 +1880,18 @@ export function getStoryboardElementPathFromEditorState(
...
 ): HighlightBoundsForUids | null {
   const file = getContentsTreeFileFromString(editorState.projectContents, fullPath)
-  if (isTextFile(file) && isParseSuccess(file.fileContents.parsed)) {
-    return getHighlightBoundsFromParseResult(file.fileContents.parsed)
+  if (isTextFile(file)) {
+    if (isParseSuccess(file.fileContents.parsed)) {
+      return getHighlightBoundsFromParseResult(file.fileContents.parsed)
+    }
+    if (file.lastParseSuccess != null) {
+      return getHighlightBoundsFromParseResult(file.lastParseSuccess)
```
balazsbajorics, line 1893: can this lead to mispositioned highlights while you insert newlines to a file?

seanparsons: Only if those newlines are causing the parse to fail as this is a fallback case.

seanparsons: (adding detail to the PR description as I missed out this bit)

Rheeseyb (approved):

PR #2003, opened 2021-11-11
fancy registerComponent

balazsbajorics:
**Problem:**
We'd like the `registerComponent` function to be used like this:

```javascript
registerComponent({
  name: "Fancy Button", // name is what appears in the insertion list
  moduleName: "my-module", // name of module, or relative file path
  controls: propertyControls, // the existing property controls
  insert: `<FancyButton label="Click Me!" />`, 
  requiredImports: `import { FancyButton } from 'my-app'; import 'my-app/dist/styles.css';`
});
```

**Details:**
- createRegisterComponentFunction has been changed to use getCachedParseResultForUserStrings
- new `getCachedParseResultForUserStrings` that talks to the parser-printer worker to give us a parse result for the user's insert + requiredImports strings
- wiring the worker to createRegisterComponentFunction
- changing the shape of `ComponentDescriptor.componentInfo` to match what the parser-printer is giving us.
- the new `ComponentDescriptor` happens to be exactly inline with the shape of `project-components@InsertableComponent` , which means I could remove some data transformation
- I also had to re-add some things I deleted from antd-components and the other component declarations last week

**Notes:**
- We have some very early-stage error messages (printed to the console, probably not discoverable enough for the users) that need lots of follow-up love
- VSCode doesn't have any autocomplete today, which makes it pretty hard to guess this function right
- input shape validation is missing (ie we don't give good error messages if the user makes a typo)


editor/src/core/property-controls/property-controls-local-parser-bridge.ts
```diff
@@ -0,0 +1,100 @@
...
+import { Imports, isParseFailure, isParseSuccess } from '../shared/project-file-types'
+import { createParseFile, getParseResult, UtopiaTsWorkers } from '../workers/common/worker-types'
+
+type ProcessedParseResult = Either<
+  string,
+  { importsToAdd: Imports; elementToInsert: JSXElementWithoutUID }
+>
+const resultCache: Map<string, ProcessedParseResult> = new Map()
+
+export async function getCachedParseResultForUserStrings(
```
Rheeseyb, line 20: Please can you move this file out of `property-controls`, since this code is purely related to communication with the parser-printer and so should live closer to that

editor/src/core/third-party/shared.ts
```diff
@@ -0,0 +1,20 @@
...
+): JSXAttributes {
+  const defaultProps = getDefaultProps(propertyControls ?? {})
+  let result: JSXAttributes = []
+  for (const key of Object.keys(defaultProps)) {
+    result.push(
+      jsxAttributesEntry(key, jsxAttributeValue(defaultProps[key], emptyComments), emptyComments),
+    )
+  }
+  return result
+}
```
Rheeseyb, line 20:
```suggestion
```

Bin this all please. This function was removed in #1981 because it assumes that the property controls are correctly formed, meaning it will bring down the editor if they aren't. We have another function `getDefaultPropsFromParsedControls` that we should be using instead.

editor/src/core/shared/jsx-attributes.ts
```diff
@@ -244,6 +245,35 @@ export function jsxAttributesToProps(
...
+          throw new Error(`Unhandled entry ${JSON.stringify(entry)}`)
+      }
+    },
+    {},
+    attributes,
+  )
+
+  return result
+}
+
```
Rheeseyb, line 276: Looks like this function is unused?

Rheeseyb (changes requested):
Please remove the unsafe `getDefaultPropsAsAttributes` and replace it with the existing version for working with parsed property controls, and then this is good to go.

editor/src/core/shared/jsx-attributes.ts
balazsbajorics: ah yeah it is!

seanparsons (approved):

editor/src/core/third-party/shared.ts
balazsbajorics: fixed. I did not bin because it is still doing something specific to built-in tempaltes (turning an object into JSXAttributes) but it is using the safe function now

editor/src/core/shared/jsx-attributes.ts
balazsbajorics: removed

editor/src/core/third-party/shared.ts
Rheeseyb: Thanks! We're actually doing exactly this elsewhere too in [`project-components.ts`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/shared/project-components.ts#L371-L376) so it's worth referencing the new function there then

Rheeseyb (approved):
Nice work!

editor/src/core/third-party/shared.ts
balazsbajorics: actually that line has been superseded in this PR, as `ComponentDescriptor` now has `componentInfo: { importsToAdd: Imports; elementToInsert: JSXElementWithoutUID }` which means no further data transformation is needed

Rheeseyb: Even better!

PR #2004, opened 2021-11-11
Restore "Update VS Code build to 1.62.0"

Rheeseyb:
Revert the revert, and thus un-revert #1995 

Before this can be merged:

- [ ] #2001 needs to be merged to ensure local dev is using the correct version of node when using yarn and pnpm
- [ ] Ensure #2001 is enough to build VS Code on macos
- [ ] https://github.com/concrete-utopia/utopia-deploy/pull/34 needs to be merged to ensure a valid version of python is available when building on CI

### DIFF
--- a/vscode-build/build.js
+++ b/vscode-build/build.js
@@ -5,7 +5,7 @@ const fse = require('fs-extra')
 const glob = require('glob')
 const rmdir = require('rimraf')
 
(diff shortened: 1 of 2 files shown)


Rheeseyb:
Closing as this isn't going to be possible yet (see #2005 for more info)

PR #2010, opened 2021-11-22
feature(components) add code snippet for easier insertion

enidemi:
**Problem:**
The new register component helper function is difficult to use.

**Fix:**
Add a custom code snippet that inserts the function when starting to type the 'registerComponent'

**Commit Details:**
- created a simple snippet
- added snippet to extension package.json

Should be merged after https://github.com/concrete-utopia/utopia/pull/2003

### DIFF
--- a/utopia-vscode-extension/snippets.json
+++ b/utopia-vscode-extension/snippets.json
@@ -5,9 +5,14 @@
       "registerComponent({",
       "  name: ${1}, // name is what appears in the insertion list",
       "  moduleName: ${2}, // name of module, or relative file path",
-      "  controls: ${3}, // property controls",
-      "  insert: ${4}, ",
-      "  requiredImports: ${5}",
+      "  controls: {",
+      "    replaceMe: {",
+      "      control: 'string-input',",
+      "      label: 'placeholder control'",
+      "    },",
+      "  }, // property controls",
+      "  insert: ${4}, // `<Button label='Click Me!' />`",
+      "  requiredImports: ${5} // `import { Button } from 'my-app';` ",
       "})"
     ],
     "description": "Register Component"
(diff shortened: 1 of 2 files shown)


utopia-vscode-extension/snippets.json
```diff
@@ -0,0 +1,15 @@
+{
+  "registerComponent": {
+    "prefix": "registerComponent",
+    "body": [
+      "registerComponent({",
+      "  name: ${1}, // name is what appears in the insertion list",
+      "  moduleName: ${2}, // name of module, or relative file path",
+      "  controls: ${3}, // property controls",
```
Rheeseyb, line 8: I think it's worth also including a snippet for a very basic control here. Perhaps a dummy string-input control would be enough. I say this because I don't think there's any way we could otherwise provide a code snippet for the property controls (unless you can think of one?), and without that it's still a bit too unclear how to proceed.

utopia-vscode-extension/snippets.json
```diff
@@ -0,0 +1,15 @@
...
+      "  insert: ${4}, ",
+      "  requiredImports: ${5}",
```
Rheeseyb, line 10: I think these two also need a comment or example

Rheeseyb (commented):
This is a really nice solution to an otherwise tricky problem! Nice one!

utopia-vscode-extension/snippets.json
enidemi: fixed

utopia-vscode-extension/snippets.json
enidemi: I added a dummy control here. 
In general the property controls need more snippets and examples about how to write them.

seanparsons (approved):

Rheeseyb (approved):

balazsbajorics (approved):

PR #2012, opened 2021-11-22
Use explicit selection change ignoring in utopia-vscode-extension

### LINKED ISSUE #2011: Selection following has regressed
In the default project running locally, when selecting via the navigator, if I have app.js open and I click "Imported App", it will first select the scene correctly, and then select the "App" component. If I have the storyboard.js open and click the "div" that is a child of "App", a similar thing will happen (it first selects the "div" and then selects the "Card")

Then, if I repeatedly click "App" and "Card" over and over, the selection continues cycling through them for a while

broken-selection-following

Rheeseyb:
Fixes #2011 

**Problem:**
Selection following in the direction from VS Code to the main Utopia editor was being filtered based on the `TextEditorSelectionChangeEvent.kind`, which is designed to flag what the cause of the selection change event was (`Keyboard`, `Mouse` or `Command`). However, we cannot rely on that value, since it is `undefined` when the change is triggered by a backspace, enter, or a file opening.

As well as this, we were preventing certain legitimate selection changes from reaching VS Code, based on previous logic (when we weren't filtering selection change messages that had originated from VS Code itself).

**Fix:**
Rather than rely on the `event.kind`, we now use a flag for explicitly ignoring the next selection change event when we know we're about to trigger one via a file opening or by explicitly setting the new selection.

### DIFF
--- a/utopia-vscode-extension/src/extension.ts
+++ b/utopia-vscode-extension/src/extension.ts
@@ -350,6 +350,7 @@ function getFullConfig(): UtopiaVSCodeConfig {
 
 let currentDecorations: Array<DecorationRange> = []
 let currentSelection: BoundsInFile | null = null
+let squashNextSelectionChange: boolean = false
 
 function initMessaging(context: vscode.ExtensionContext, workspaceRootUri: vscode.Uri): void {
   function handleMessage(message: ToVSCodeMessage): void {
@@ -394,7 +395,10 @@ function initMessaging(context: vscode.ExtensionContext, workspaceRootUri: vscod
       updateDecorations(currentDecorations)
     }),
     vscode.window.onDidChangeTextEditorSelection((event) => {
-      if (event.kind !== vscode.TextEditorSelectionChangeKind.Command) {
+      if (squashNextSelectionChange) {
+        // Ignore selection changes coming from Utopia
+        squashNextSelectionChange = false
+      } else {
         cursorPositionChanged(event)
       }
     }),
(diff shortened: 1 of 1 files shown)


utopia-vscode-extension/src/extension.ts
```diff
@@ -505,17 +505,18 @@ async function revealRangeIfPossibleInVisibleEditor(boundsInFile: BoundsInFile):
...
-    const alreadySelected = rangesIntersectLinesOnly(visibleEditor.selection, rangeToReveal)
     const alreadyVisible = visibleEditor.visibleRanges.some((r) =>
       r.contains(visibleEditor.selection),
     )
 
-    if (!alreadySelected) {
-      const selectionRange = getVSCodeRange(boundsInFile)
-      visibleEditor.selection = new vscode.Selection(selectionRange.start, selectionRange.start)
+    const selectionRange = getVSCodeRange(boundsInFile)
+    const newSelection = new vscode.Selection(selectionRange.start, selectionRange.start) // selectionRange.end?
```
Rheeseyb, line 513: Right now we're setting the code selection to be the very starting point of the element, and so we're using `selectionRange.start` for both the `anchor` and the `active` values here, where the `anchor` is the end of the selection without a cursor, and the `active` is the end of the selection _with_ the cursor. This is the current setup on production.

If we change the `active` value to use `selectionRange.end` the result is that the entire block of code is selected (and the highlighting reflects that), meaning it would look like this when selecting an element in the navigator:
Screenshot_20211122_162915

enidemi (approved):

balazsbajorics (approved):

PR #2017, opened 2021-11-25
Use Yarn instead of NPM for the packager.

### LINKED ISSUE #2009: Investigate (and probably use) pnpm rather than npm in the server
Right now we use `npm` in the server for installing packages ([see here](https://github.com/concrete-utopia/utopia/blob/3da7f891c0bfa800f011d2dab6de94a90d020ec3/server/src/Utopia/Web/Packager/NPM.hs#L121)). We made the switch to using `pnpm` for our development environment, and likely should be using that in the server too, as in our experience we have found it to be both more performant and more robust.

However, we can't just make the swap without first ensuring that the structure or contents of the installed modules differ in a way that would break the editor, so we first need to investigate and verify that the switch would be safe to do.

seanparsons:
Fixes #2009

**Problem:**
Some packages don't want to be installed by npm for varying reasons.

**Fix:**
Switch to using yarn when pulling down a particular dependency in the package server.

**Commit Details:**
- Fixes #2009.
- Switched to using `yarn add` instead of `npm install` in `withInstalledProject`.

### DIFF
--- a/server/src/Utopia/Web/Packager/NPM.hs
+++ b/server/src/Utopia/Web/Packager/NPM.hs
@@ -118,14 +118,14 @@ withInstalledProject logger NPMMetrics{..} semaphore versionedPackageName withIn
   let deleteDir = ignoringIOErrors . removeDirectoryRecursive
   bracketP createDir deleteDir $ \tempDir -> do
     -- Run `npm install "packageName@packageVersion"`.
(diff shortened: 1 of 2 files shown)


Rheeseyb (approved):
Just tested with an R3F project and this appears to have been far smoother than I had anticipated! Should I be suspicious?

enidemi (approved):

balazsbajorics (approved):

PR #2019, opened 2021-11-26
Use more nix for the server environment.

seanparsons:
**Problem:**
When run inside a Docker container or on a fresh setup of whatever kind the server build takes a little while as it builds so many dependencies.

**Fix:**
The builds used in development now get all their Haskell dependencies via nix, which results in it obtaining pre-built and cached versions of those libraries instead of building them from scratch. There is also some support for a fully nix built server, but in dev the main application is still built in the regular way so that small changes result in smaller incremental builds.

**Commit Details:**
- Moved where the database runs to the root of the project so that
  files created by it do not interfere with the server nix builds.
- Deleted `cabal.project.freeze` as it's not hugely useful anymore
  and can cause a bunch of server dependencies to need rebuilding.
- Created `server/default.nix` which defines a series of targets for
  the server:
  - `server-dev`: Derivation that produces a build of the server
    as fast as it reasonably can.
  - `server-shell`: Derivation that includes the JS tools which the
    server depends on, so they can be used alongside the server.
  - `server-tested`: Derivation which tests the server alongside
    producing the server itself.
- Bumped some versions in `package.yaml` and added a flag which allows
  us to disable the tests which attempt to reach out from the sandbox.
- Created `server/shell.nix` to provide a server specific shell.
- Server tests now support the ability to disable the tests which do
  things like talk to PostgreSQL or invoke npm to get access to the
  npmjs registry.
- `cabal-update` script now uses a much later revision so that some
  of the more recent servant libraries are available to the dependency
  solver as they're used in nixpkgs.
- `test-server-inner` now runs cabal within the shell provided by
  `server/shell.nix`.
- `run-server-inner` now runs cabal within the shell provided by
  `server/shell.nix`.


.gitignore
```diff
@@ -127,6 +127,9 @@ editor/src/components/editor/npm-dependency/utopia-api-typings.ts
 .github-test-projects
 
 # Database.
-server/pglog.txt
-server/.pglock
-server/utopia-db
+pglog.txt
+.pglock
+utopia-db
```
Rheeseyb, line 132: Looking forward to the upcoming PRs with the previous files now suddenly included :grinning:

seanparsons: Muahahaha.

balazsbajorics: are you trying to suggest we should keep the old one sticking around?
```suggestion
pglog.txt
.pglock
server/utopia-db
utopia-db
```

Rheeseyb (approved):

balazsbajorics (approved):

PR #2020, opened 2021-11-26
support for pkg.browser field

### LINKED ISSUE #2015: support package.json Browser field
currently utopia supports pkg.main and pkg.module as the two entry points in package.json
Apparently there's pkg.browser too, and I found some mention of pkg.exports which terrifies me.

about pkg.exports: https://webpack.js.org/guides/package-exports/#general-syntax
about pkg.browser: https://github.com/defunctzombie/package-browser-field-spec

a scary github issue thread from 2017 that shows that I'm not the only one who is absolutely confused: https://github.com/webpack/webpack/issues/4674

balazsbajorics:
Fixes #2015

**Problem:**
We encountered a module [public-ip](https://unpkg.com/public-ip@4.0.4/package.json) that uses the `package.json@browser` field with no fallback for the `package.json@module` field.

**Fix:**
Turns out that we missed the package.json browser field: https://github.com/defunctzombie/package-browser-field-spec

This PR implements support for the browser field, the "advanced mode" module replacement and the module ignore features. I've also added new tests for it too.

**Commit Details:**
- New module resolution type `ResolveSuccessIgnoreModule` which means "we found the module, it is explicitly set to ignore by the browser field. the expected require result should be empty object (`{}`)"
- Module resolution handles "simple" browser field case: uses it just like main and module. browser has precedence.
- Module resolution is capable of resolving replaced modules now.
- new tests


editor/src/core/es-modules/package-manager/module-resolution.ts
```diff
@@ -183,22 +209,31 @@ interface PartialPackageJsonDefinition {
...
       setOptionalProp(result, 'main', main)
       setOptionalProp(result, 'module', module)
+      setOptionalProp(result, 'browser', browser)
       return result
     },
     optionalObjectKeyParser(parseString, 'name')(value),
     optionalObjectKeyParser(parseString, 'main')(value),
     optionalObjectKeyParser(parseString, 'module')(value),
+    optionalObjectKeyParser(
+      parseAlternative<string | MapLike<string | null>>(
```
Rheeseyb, line 231: According to the official spec this should be `parseAlternative<string | MapLike<string | boolean>>`

balazsbajorics: fixed!

editor/src/utils/value-parser-utils.ts
```diff
@@ -267,6 +267,14 @@ export function parseBoolean(value: unknown): ParseResult<boolean> {
   }
 }
 
+export function parseFalse(value: unknown): ParseResult<false> {
+  if (typeof value === 'boolean' && value === false) {
```
Rheeseyb, line 271: `&& value === false`
This feels so very wrong general, and yet in this specific case it feels like the correct thing to do. I am so conflicted right now

Rheeseyb (approved):

enidemi (approved):

PR #2023, opened 2021-11-29
fix(insertion) for parsed components use registerComponent element and imports when available

enidemi:
**Problem:**
When using the `registerComponent` for internal components the `requiredImports` and `insert` fields are ignored.

**Fix:**
This was missing from the insert menu support, originally it was prepared to work on registering third party components.

**Commit Details:**
- when creating the component insert menu entries it now checks for existing component descriptor

### DIFF
--- a/editor/src/components/shared/project-components.ts
+++ b/editor/src/components/shared/project-components.ts
@@ -333,16 +333,6 @@ export function getComponentGroups(
             },
             parsedControls,
           )
-          let attributes: JSXAttributes = []
-          for (const key of Object.keys(defaultProps)) {
-            attributes.push(
-              jsxAttributesEntry(
-                key,
-                jsxAttributeValue(defaultProps[key], emptyComments),
-                emptyComments,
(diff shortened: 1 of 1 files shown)


editor/src/components/shared/project-components.ts
```diff
@@ -343,12 +343,29 @@ export function getComponentGroups(
...
+              stylePropOptions,
+            )
+          } else {
+            return insertableComponent(
+              exportedComponent.importsToAdd,
+              jsxElementWithoutUID(exportedComponent.listingName, attributes, []),
+              exportedComponent.listingName,
+              stylePropOptions,
+            )
+          }
```
Rheeseyb, line 368: Since the `attributes` array is only used in this block, we should probably move the creation of it inside this block to

Rheeseyb (approved):

balazsbajorics (approved):

seanparsons (approved):

PR #2025, opened 2021-11-30
fix(components) show property controls for elements imported with import as

enidemi:
**Problem:**
Components using named imports, like `import {Card as MyCard} from "..."`, are not showing the property controls in the inspector.

**Fix:**
The issue is that component section is relying on the jsx element name to find the matching property controls. The fix is to use the `importedFrom` result to find the original exported name for elements imported from other files.

**Commit Details:**
- imported elements use `importedFrom.exportedName` as a name, fallback is the jsx element name


Rheeseyb (approved):
Nice!

PR #2028, opened 2021-11-30
Replace registerComponent with registerModule

Rheeseyb:
**Problem:**
We wanted to replace the `registerComponent` call required for each component with a single `registerModule` call that could be called once per module or file

**Fix:**
Define a new `registerModule` function with the following API:
``` ts
function registerModule(
  moduleName: string,
  components: {
    [componentName: string]: {
      controls: PropertyControls,
      insertOptions: [
        {
          codeToInsert: string
          additionalRequiredImports?: string
          menuLabel?: string
        }
      ]
    }
  },
)
```

This in turn means that `ComponentDescriptor` had to be updated to support multiple insert options for a given component, and that the snippets needed to be updated too. As I was updating the snippets, I've taken advantage of the available [variables](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_variables), along with placeholders, to include the current file as a placeholder, meaning the snippet will now show as follows:
Screenshot_20211130_180615
(in each case, updating the highlighted text will update all cases of that highlighted text whilst the snippet is being inserted)

**Note:**

There are still some open questions or issues following on from this, which will be handled in separate PRs:

1. There exists a bug around the use of `getParseResult` which currently prevents the multiple insert options from working unless they're carefully added one at a time
2. Deleting a `registerModule` call won't un-register that module
3. Adding an import for a file that doesn't exist will crash the canvas on insertion in a non-recoverable way 
4. Adding a CSS import (like in the sample snippet) won't actually add that import on insertion, even if the file exists.


utopia-vscode-extension/snippets.json
```diff
@@ -1,25 +1,33 @@
...
     "body": [
-      "registerComponent({",
-      "  name: '${1}', // name is what appears in the insertion list",
-      "  moduleName: '${2}', // name of module, or relative file path",
-      "  controls: {",
-      "    replaceMe: {",
-      "      control: 'string-input',",
-      "      label: 'placeholder control'",
+      "registerModule(",
+      "  '${1:${TM_FILEPATH/~(.*)\\..+$/$1/}}', // module name or absolute file path",
```
enidemi, line 11: this is cool

enidemi (approved):

utopia-vscode-extension/snippets.json
balazsbajorics: or scary

PR #2030, opened 2021-12-01
Fix/import side effect printing

Rheeseyb:
**Problem:**
Imports of the form `import '/src/file'` (i.e. imports that are for executing the file for the sake of any side effects) aren't printed to the code when inserting them via the parsed model.

**Fix:**
The issue was that when printing the code we print each of the types of imports separately, and we hadn't added the code for printing a side effect only import statement. I've added that, and also a test for import printing.


enidemi (approved):

seanparsons (approved):

editor/src/core/workers/parser-printer/parser-printer-imports.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+        '/index.js',
+        printCodeOptions(false, true, true),
+        expectedImports,
+        parsedCode.topLevelElements,
+        null,
+        [],
+      )
+      expect(printedCode).toEqual(expectedCodeAfter)
+    } else {
+      fail('Parse result is not a success.')
```
balazsbajorics, line 88: I just learned today that `fail` is not coming from Jest, it is coming from Jasmine and we should be throwing an error instead

balazsbajorics (approved):

PR #2031, opened 2021-12-01
add jest/eslint rules, fix problems that came up

balazsbajorics:
**Problem:**
We want to avoid accidentally committing and merging `describe.only` and `it.only` tests to the repository, as those "focus" the test runner and disable other tests.

**Fix:**
ESLint has a [Jest Plugin](https://github.com/jest-community/eslint-plugin-jest) which will throw us an Error when encountering a `describe.only`

I enabled most of the recommended rules, and fixed the issues that came up.

Most importantly, I deleted all uses of `fail()` and instead throw an error.

We also had a few tests with matching names so I appended ` 2` to some of those test names

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer-arbitrary-elements.spec.tsx
+++ b/editor/src/core/workers/parser-printer/parser-printer-arbitrary-elements.spec.tsx
@@ -74,7 +74,9 @@ export var ${BakedInStoryboardVariableName} = (props) => {
     ).formatted
     const parseResult = testParseCode(code)
     foldParsedTextFile(
-      (failure) => fail(failure),
+      (failure) => {
+        throw new Error(JSON.stringify(failure))
+      },
       (success) => {
         const firstComponent = success.topLevelElements.find(isUtopiaJSXComponent)
         if (firstComponent != null) {
@@ -87,19 +89,22 @@ export var ${BakedInStoryboardVariableName} = (props) => {
                 firstChild.elementsWithin[Object.keys(firstChild.elementsWithin)[0]]
               expect(getJSXAttribute(elementWithin.props, 'data-uid')).not.toBeNull()
             } else {
-              fail('First child is not an arbitrary block of code.')
+              throw new Error('First child is not an arbitrary block of code.')
             }
           } else {
-            fail('Root element not a JSX element.')
+            throw new Error('Root element not a JSX element.')
           }
         } else {
-          fail('Not a component at the root.')
+          throw new Error('Not a component at the root.')
         }
       },
-      (unparsed) => fail(unparsed),
+      (unparsed) => {
+        throw new Error(JSON.stringify(unparsed))
+      },
       parseResult,
     )
   })
+  // eslint-disable-next-line jest/expect-expect
   it('should write updated arbitrary elements back into code', () => {
     const code = applyPrettier(
       `import * as React from "react";
@@ -772,7 +777,7 @@ return { arr: arr };`
     )
     expect(actualResult).toEqual(expectedResult)
   })
-  it('supports passing down the scope to children of components', () => {
+  it('supports passing down the scope to children of components 2', () => {
     const code = `import React from "react";
 import { View } from "utopia-api";
 export var whatever = (props) => {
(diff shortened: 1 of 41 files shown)


editor/src/core/workers/parser-printer/parser-printer-arbitrary-elements.spec.tsx
```diff
@@ -87,19 +89,22 @@ export var ${BakedInStoryboardVariableName} = (props) => {
...
         }
       },
-      (unparsed) => fail(unparsed),
+      (unparsed) => {
+        throw new Error(JSON.stringify(unparsed))
+      },
       parseResult,
     )
   })
+  // eslint-disable-next-line jest/expect-expect
```
enidemi, line 107: what does this do? why is it disabled?

balazsbajorics: it expects an expect, but there's no expect here because this test calls a helper function which in turn calls expect :)

editor/src/core/workers/ts/ts-worker.spec.ts
```diff
@@ -43,7 +44,7 @@ describe('Typescript worker applies the loaders', () => {
         // Ensure no errors
         for (const builtFile in msg.buildResult) {
           if (msg.buildResult[builtFile].errors.length > 0) {
-            fail(`Build errors found in built file ${builtFile}`)
+            done.fail(`Build errors found in built file ${builtFile}`)
```
enidemi, line 47: should this throw an error too?

enidemi (approved):

Rheeseyb:
> We also had a few tests with matching names so I appended 2 to some of those test names

Hahahaha good enough :smile:

editor/src/core/workers/parser-printer/parser-printer-arbitrary-elements.spec.tsx
Rheeseyb: Well that's not at all what I expected

editor/src/core/workers/parser-printer/parser-printer-exports.spec.tsx
```diff
@@ -286,28 +286,7 @@ export function App() {
         ]
       `)
     } else {
-      fail('Did not parse successfully.')
-    }
-  })
-
-  it(`parses 'export function' marked values`, () => {
```
Rheeseyb, line 293: Why was this test removed?

editor/src/core/workers/ts/ts-worker.spec.ts
balazsbajorics: I'm not sure. this test uses `done` so I figured the safest bet is to use `done.fail` so we are not mixing two different ways of using async jest

editor/src/core/workers/parser-printer/parser-printer-exports.spec.tsx
balazsbajorics: it seemed like not only the name was redundant, but it was the same test as the `it` before it

editor/src/utils/react-performance.spec.ts
```diff
@@ -46,12 +46,12 @@ describe('keepDeepReferenceEqualityIfPossible', () => {
...
   })
 
-  it('keeps reference equality for matching parts of arrays even if they are different in length', () => {
+  it('keeps reference equality for matching parts of arrays even if they are different in length 2', () => {
     const a = deepFreeze([{ a: 5 }, { b: { c: { d: 6 } } }]) as any[]
     const b = deepFreeze([{ a: 5 }]) as any[]
     const r = keepDeepReferenceEqualityIfPossible(a, b)
     expect(fastDeepEquals(b, r)).toBeTruthy()
-    expect(r.length === 1).toBeTruthy
+    expect(r.length === 1).toBeTruthy()
```
Rheeseyb, line 54: omg

balazsbajorics: to be honest, I should have rewritten it to be 
```suggestion
    expect(r).toHaveLength(1)
```

editor/src/core/workers/parser-printer/parser-printer-exports.spec.tsx
Rheeseyb: Oh wow! It certainly looks like it was copy pasta'd

Rheeseyb (approved):

PR #2034, opened 2021-12-01
Handle rapidly fired requests to getParseResult.

seanparsons:
**Problem:**
When multiple insert options are added via `registerModule`, often the first one is used for all of them when inserting.

The underlying issue was that `getParseResult` didn't discriminate between the results coming back from the worker, so would accept any that arrived.

**Fix:**
Each message going to the parser printer worker now gets a unique ID attached to it which is returned as part of the result, so that the handler function can ensure it is only handling results that match the request sent.

**Commit Details:**
- Created a base interface for the various parse-print message types
  called `ParsePrintBase` which holds the `messageID` value used
  to coordinate the requests and responses.
- `getParseResult` now uses a global of `PARSE_PRINT_MESSAGE_COUNTER`
  to create unique ids for each request so that the handler function
  only handles the messages with that id.
- Parser printer worker now sends back the id given in the request
  as part of the response.
- A few minor tweaks to various interfaces in the workers around using
  `ParsePrintFilesRequest` instead of an array of `ParseOrPrint`.

### DIFF
--- a/editor/src/core/es-modules/package-manager/utopia-api-typings.ts
+++ b/editor/src/core/es-modules/package-manager/utopia-api-typings.ts
@@ -1,12 +1,5 @@
 export const utopiaApiTypings = `declare module 'utopia-api/helpers/helper-functions' {
   import { PropertyControls } from 'utopia-api/property-controls/property-controls';
-  export function registerComponent(paramsObj: {
-      name: string;
-      moduleName: string;
-      controls: PropertyControls;
-      insert: string;
-      requiredImports: string;
-  }): void;
   export interface ComponentInsertOption {
       codeToInsert: string;
       additionalRequiredImports?: string;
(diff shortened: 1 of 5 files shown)


editor/src/core/es-modules/package-manager/utopia-api-typings.ts
```diff
@@ -1,12 +1,5 @@
 export const utopiaApiTypings = `declare module 'utopia-api/helpers/helper-functions' {
   import { PropertyControls } from 'utopia-api/property-controls/property-controls';
-  export function registerComponent(paramsObj: {
```
Rheeseyb, line 3: Merging with master should remove this now - I missed this change previously but have since committed it

Rheeseyb (approved):

PR #2035, opened 2021-12-02
run the performance step in a separate job, on a self-hosted runner

balazsbajorics:
**Problem**
The puppeteer performance test were executed on GithubActions along with all the rest of the jobs. A few months ago something changed, and the test results became incredibly unreliable, seemingly the test runner loosing CPU allocation _during_ the test run.

**Solution**
Move the puppeteer test (and only the puppeteer test) to a Linode dedicated CPU instance.

This PR is already showing performance tests that are executed on the Linode instance.

I picked the cheapest dedicated CPU instance, 2 cores, $30 / month + $5 for backup

**Details**
- created a new linode instance `github-actions-1` https://cloud.linode.com/linodes/32420665
- I have no idea who can see this, I think Malte and Balazs are the ones who are in the Utopia Linode origanization
- this instance runs the github action runner as a systemd process

**Notes**
- The node has manually installed nix and the github actions runner. There's no setup script or Dockerfile to instantly recreate it
- I did manage to make a "Backup Snapshot" of the moment it all started working. that should be a stopgap for now
- We should run this for a couple of days and observe reliabilty
- If it is unreliable, consider other options and possibly a solution which lets us make easy to reproduce setups (Docker or nixos or otherwise)


maltenuhn:
> I have no idea who can see this
I'll invite everyone once we're happy we'll stick with this

Rheeseyb (commented):
The Discord integration is failing here because it seems the Linode instance is missing `jq`:

```jq: command not found```

That log line is at the end of the `Build Discord Message` step of the workflow, which is used for building the JSON embeds before sending the message.

balazsbajorics:
> The Discord integration is failing here because it seems the Linode instance is missing `jq`:
> 
> `jq: command not found`
> 
> That log line is at the end of the `Build Discord Message` step of the workflow, which is used for building the JSON embeds before sending the message.

oh I missed this comment. I'll fix this, thanks!

balazsbajorics:
@Rheeseyb I fixed the jq command a few weeks ago! I think this is ready to merge, the server was running for weeks now

Rheeseyb (approved):

PR #2044, opened 2021-12-07
Remove the defaultValue field from property controls.

seanparsons:
**Problem:**
The `defaultValue` field in the property controls has become a lot less useful in the presence of the insert options.

**Fix:**
Removed the `defaultValue` and where possible moved existing values into the insert options.

**Commit Details:**
- Removed the `defaultValue` field from all the types in `property-controls.ts`.
- Deleted the `getDefaultProps` function from `property-controls.ts`.
- Added the utility function `simpleAttribute` as a way of making an instance
  of `JSXAttributesEntry` for the most simple case.
- Removed all the `defaultValue` fields in the antd and r3f control definitions,
  shifting those properties into the insert option for those elements.
- Retired the `RESET_PROP_TO_DEFAULT` editor action as we now have no defaults
  to base it upon.
- Changed `INSERT_WITH_DEFAULTS` action to be the `INSERT_INSERTABLE` action.
- Deleted `instance-context-menu.tsx` as it's not used anymore.
- Updated `makeHTMLDescriptor` to support default attributes to replace the
  default values for properties.
- All the parsers in `property-controls-parser.ts` updated for the lack of
  `defaultValue`.
- Removed an interface and some functions in `property-controls-utils.ts` as
  they relate to default prop values.
- `getComponentGroups` now uses `parsedPropertyControlsForComponentInFile`
  instead of `defaultPropertiesForComponentInFile`.
- Deleted `shared.ts` as it is not used anymore.


editor/src/components/editor/actions/actions.spec.ts
```diff
@@ -1423,7 +1423,7 @@ describe('INSERT_WITH_DEFAULTS', () => {
                 />
                 <img
                   style={{ width: 100, height: 100 }}
-                  src='/editor/icons/favicons/favicon-128.png?hash=nocommit\\"'
+                  src='/editor/icons/favicons/favicon-128.png?hash=nocommit'
```
Rheeseyb, line 1426: Huh?

Rheeseyb (approved):

utopia-api/src/property-controls/property-controls.ts
```diff
@@ -5,14 +5,12 @@ import { fastForEach } from '../utils'
 interface ControlBaseFields {
   control: RegularControlType
   label?: string
-  defaultValue?: unknown
   visibleByDefault?: boolean
```
balazsbajorics, line 9: hmm I thought we have a "placeholder" field too, which makes the defaultValue field redundant, but it doesn't seem to be the case anymore!

is there even a need in the inspector to display a greyed out value for a field which has no value? or we are not showing those controls anyways because we are not rendering those controls because they are collapsed into the "available properties to insert" list below the controls

editor/src/core/shared/element-template.ts
```diff
@@ -648,6 +648,10 @@ export function jsxAttributesEntry(
   }
 }
 
+export function simpleAttribute(key: string, value: unknown): JSXAttributesEntry {
```
balazsbajorics, line 651: good idea!

balazsbajorics (approved):

editor/src/components/editor/actions/actions.spec.ts
seanparsons: There appeared to be an errant `"` on the end of that.

utopia-api/src/property-controls/property-controls.ts
seanparsons: If it has no value, then it should be in that that unused props bit below the controls like you say.

PR #2046, opened 2021-12-08
Fix handling of node module add / update / delete logic

Rheeseyb:
**Problem:**
The node modules stored in the editor were being wiped under certain conditions, breaking the running project with reference errors.

**Fix:**
The actual issue here was that the logic for updating the node modules was still based on old logic that relied on a bundler worker reporting whether an update was a full build or an incremental one, combined with a change that prevented repeated requests for packages that hadn't been added or removed when updating the `package.json`. The result being that when saving a change to the `package.json`, the entire `nodeModules` would be emptied, and only newly added dependencies would be inserted back into it.

The fix here was to remove the now defunct `'full-build'` vs `'incremental'` logic, and instead manually handle package addition, updating, and removal.


seanparsons (approved):
With a simplification of the existing logic as a bonus, nice!

editor/src/components/editor/actions/actions.ts
```diff
@@ -1517,6 +1519,18 @@ function updateSelectedComponentsFromEditorPosition(
   }
 }
 
+function removeModulesFromNodeModules(
+  modulesToRemove: Array<string>,
+  nodeModules: NodeModules,
+): NodeModules {
+  const filePathsToRemove = modulesToRemove.map((m) => `/node_modules/${m}/`)
+  return Object.fromEntries(
```
balazsbajorics, line 1527: I've been using the badly named objectMapDropNulls for this https://github.com/concrete-utopia/utopia/blob/6fba04e8ab53daefdba16ad4aa7e0e392d89b804/editor/src/core/shared/object-utils.ts#L126

balazsbajorics (approved):

editor/src/components/editor/actions/actions.ts
Rheeseyb: Ah good shout, that would make this a lot clearer

PR #2048, opened 2021-12-09
Merge calls to registerModule for the same module.

seanparsons:
**Problem:**
If a user unknowingly uses `registerModule` more than once for the same module, the latest invocation "wins" and the earlier ones are all ignored.

**Fix:**
Combine the results of the different calls together when closing off the render.

**Commit Details:**
- Modified `validateControlsToCheck` to merge the result of multiple
  calls made in one pass.

### DIFF
--- a/editor/src/components/canvas/canvas-globals.spec.ts
+++ b/editor/src/components/canvas/canvas-globals.spec.ts
@@ -138,9 +138,22 @@ const selectorPropertyControlsInfo: PropertyControlsInfo = {
   },
 }
 
+const otherCardComponentDescriptorWithName: ComponentDescriptorWithName = {
+  ...cardComponentDescriptor,
+  componentName: 'Other Card',
+}
+
+const otherCardControlsToCheck: ControlsToCheck = Promise.resolve(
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-globals.spec.ts
```diff
@@ -138,12 +138,25 @@ const selectorPropertyControlsInfo: PropertyControlsInfo = {
...
 describe('validateControlsToCheck', () => {
   beforeEach(() => {
     // Twice because the first will leave some data in `previousModuleNamesOrPaths`.
     resetControlsToCheck()
     resetControlsToCheck()
   })
+  afterAll(() => {
+    resetControlsToCheck()
+    resetControlsToCheck()
+  })
```
Rheeseyb, line 159: Why the `afterAll` as well now?

seanparsons: Just means any subsequent test in the future wont be caught out by this having some data kicking around.

balazsbajorics (approved):

PR #2050, opened 2021-12-10
Add factory functions for the property controls

Rheeseyb:
**Problem:**
The creation of property controls is a lot of effort

**Fix:**
Make it less effort by providing factory functions. I've also updated the code snippets to use those, and updated the default project to now include the following import statement in the storyboard file:
```js
import Utopia, {
  Scene,
  Storyboard,
  registerModule,
} from 'utopia-api'
```
That change meant having to update a whole load of test files.

Also I had to make a slight tweak to the `optionalObjectKeyParser` to explicitly handle `undefined` values (this was causing an issue with the parsing of `registerModule` when using a factory function with an optional param).


editor/src/utils/value-parser-utils.ts
```diff
@@ -181,7 +181,7 @@ export function optionalObjectKeyParser<V>(
     if (typeof value === 'object' && !Array.isArray(value) && value != null) {
       const withErrorParser = objectValueParserWithError(parser)
       const valueAsObject: any = value
-      if (key in valueAsObject) {
+      if (key in valueAsObject && valueAsObject[key] !== undefined) {
```
seanparsons, line 184: I'm not sure how I feel about this, as this implies that it should be passed to the `withErrorParser`.

utopia-vscode-extension/snippets.json
```diff
@@ -30,282 +22,140 @@
...
-    "description": "Register Module Tuple Control"
+      "Utopia.tupleControl([",
+      "  ${1:Utopia.stringControl()},",
+      "  ${2:Utopia.numberControl()},",
+      "])"
+    ]
   },
-  "registerModule:folder property control": {
-    "prefix": "registerModule:Folder Control",
+  "Union Prop Control": {
```
seanparsons, line 148: I think this naming will mean users wont know how to find it, especially with the shorter `Prop` in the name.

editor/src/utils/value-parser-utils.ts
Rheeseyb: Yeah I feel a bit uneasy about this tbh too. I'm going to tweak the factories so that I can revert this

utopia-vscode-extension/snippets.json
Rheeseyb: I'll update `Prop` to `Property`

editor/src/utils/value-parser-utils.ts
Rheeseyb: Done

utopia-vscode-extension/snippets.json
Rheeseyb: Done

seanparsons (approved):

enidemi (approved):

PR #2052, opened 2021-12-14
Fix/react memo

enidemi:
**Problem:**
This PR is related to enabling fast refresh.

**Fix:**
Replacing betterReactMemo with React.memo. Originally betterReactMemo was created because the component names were missing from the react devtools.

**Commit Details:**
- remove betterReactMemo
- replace it in all places
- this surfaced minor issues: for example FormulaBar received a style prop but the component FormulaBar was not using it, and NumberInputPropertyControl used a hook inside an if-else.


balazsbajorics (approved):
important note that this PR removes all `.displayName = 'ComponentName'` assignments, we will need to add those back once we know if/how the displayNames work again with an updated React

Rheeseyb (approved):

seanparsons (approved):

PR #2053, opened 2021-12-14
feature(vscode-common) update build

enidemi:
**Problem:**
The vscode-common bundle result is not working with vite.

**Fix:**
Changing the tsconfig module to ES2020 from commonJS and removing webpack. Instead of one bundle this creates individual files, that works both with vscode-extension, our current and future editor development.

**Commit Details:**
- removing webpack from vscode-common
- using build to use `tsc`


seanparsons (approved):

utopia-vscode-common/package.json
```diff
@@ -14,10 +14,10 @@
     "url": "https://github.com/concrete-utopia/utopia"
   },
   "scripts": {
-    "build": "webpack-cli",
-    "production": "webpack-cli --mode production",
-    "watch": "webpack-cli --mode production --watch --info-verbosity verbose",
-    "watch-dev": "webpack-cli --watch --info-verbosity verbose",
+    "build": "tsc",
+    "production": "tsc",
+    "watch": "tsc --watch --verbose",
```
balazsbajorics, line 19: do we need thise --verbose here? what does it do? I can't find much info by googling for it

balazsbajorics (approved):

PR #2056, opened 2021-12-15
Fix/change require to import

enidemi:
**Problem:**
Part of changing our development using vite and fast refresh. Require is not available, only the es import statements are accepted.

**Fix:**
Changing all the requires to import.

**Commit Details:**
- also updated 2 dependencies


editor/src/missing-types/index.d.ts
```diff
@@ -16,7 +16,7 @@ declare module '@babel/plugin-transform-modules-commonjs'
 declare module '@babel/plugin-proposal-export-namespace-from'
 
 declare module 'lodash.clamp' {
-  export const clamp = (number: number, lower: number, upper: number) => number
+  export default clamp = (number: number, lower: number, upper: number) => number
   export const clamp = (number: number, upper: number) => number
```
seanparsons, line 20: Should this definition change as well?

seanparsons (approved):

editor/src/missing-types/index.d.ts
enidemi: I've checked the lodash library and updated this

editor/src/third-party/react-error-overlay/utils/generateAnsiHTML.ts
```diff
@@ -7,7 +7,7 @@
 
 /* @flow */
 
-const Anser = require('anser') // TODO does this work?
+import Anser from 'anser'
```
Rheeseyb, line 10: LOL at the previous `TODO` :grin:

Rheeseyb (approved):

balazsbajorics (approved):

PR #2058, opened 2021-12-17
Remove the LayoutProp type and mostly support the target selector for layout.

seanparsons:
**Problem:**
In trying to support the target selector for the layout section, it was discovered that a lot of the old pinning logic tied to `LayoutProp` would cause a lot of pain.

**Fix:**
Starting with deleting `LayoutProp` and reshaping the connected types, which then necessitated the very base level support for the target selector in layout sections. A lot of old pinning and group related code was threaded through this and was deleted/fixed as necessary.

**Notes:**
There are a couple of follow on changes:
- Moving and renaming `stylePropPathMappingFn` as it lives in some of the inspector code.
- Decide how to handle `css` versus `style` properties, as when the target selector is set to `style`, the inspector shows the dimensions as coming from some CSS, which seems a little odd as they're not _strictly_ coming from some regular CSS.

**Commit Details:**
- Removed `LayoutProp` from `layout-helpers-new.ts`.
- Reworked a bunch of the sister types to `LayoutProp`, lowercasing
  `'Width'` to `'width'` as well as removing the center pinning fields.
- Used `stylePropPathMappingFn` when constructing property paths relating
  to the `style` attribute so they can be used similarly with `css`
  and other attributes.
- Previously static collections of property paths made into functions
  as a result of them needing to use `stylePropPathMappingFn`.
- Removed some group handling functions and logic because they're not
  really a thing anymore.
- Fixed a heinous bug in `getLayoutProperty` where the result from
  `getSimpleAttributeAtPath` would be a raw value typed to `any`,
  but the result would be typed to something like `CSSNumber` wrapped
  in an `Either`. There were also some tests that needed fixing as a
  result of relying on the value and not matching the type of the result.
- Deleted `size-box-label.tsx` as it wasn't used.
- For a lot of the canvas controls used `['style']` as the property
  target for now.


editor/src/components/canvas/controls/property-target-selector.tsx
```diff
@@ -26,17 +27,17 @@ interface PropertyTargetSelectorProps {
 
 function labelForOption(option: LayoutTargetableProp): string {
```
Rheeseyb, line 28: Looks like this function is now redundant

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -175,11 +175,10 @@ export function getSelectableViews(
 
     const selectableViewsFiltered = uniqueSelectableViews.filter((view) => {
       // I kept the group-like behavior here that the user can't single-click select the parent group, even though it is a view now
-      const isGroup = MetadataUtils.isAutoSizingViewFromComponents(componentMetadata, view)
       const isAncestorOfSelected = selectedViews.some((selectedView) =>
         EP.isDescendantOf(selectedView, view),
       )
-      if (isGroup && isAncestorOfSelected) {
```
Rheeseyb, line 182: This looks wrong. It looks like this `if` statement should go in the bin now as it would have previously always been `false` (so the `else` clause would execute, returning `true`)

seanparsons: That works for me, I _think_ at the point when I looked at it I was concerned that I would be removing prior (to the group condition) behaviour.

editor/src/components/canvas/controls/property-target-selector.tsx
seanparsons: Ha, so it is.

Rheeseyb (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
seanparsons: Fixed in 7585ad850669034f876221620ecc02d4879ec51c.

editor/src/components/canvas/controls/property-target-selector.tsx
seanparsons: Fixed in 7585ad850669034f876221620ecc02d4879ec51c.

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -4047,6 +4048,10 @@ export interface ParsedCSSProperties {
   flexGrow: number
   flexShrink: number
   display: string
+  width: CSSNumber | undefined
+  height: CSSNumber | undefined
+  flexBasis: CSSNumber | undefined
+  flexGap: number
```
balazsbajorics, line 4054: I think there's no `flex-gap` css property. there's a new-ish `gap` property which wors for flex. https://developer.mozilla.org/en-US/docs/Web/CSS/gap

it was introduced in Chrome 70

PR #2063, opened 2022-01-03
Spike: canvas controls clean 

balazsbajorics:
### How to add a strategy
In `components/canvas/canvas-strategies/canvas-strategies.ts` there's a const `RegisteredCanvasStrategies` where you can add a new strategy.

A strategy has the following type:
```typescript
export interface CanvasStrategy {
  name: string
  updateFn: CanvasStrategyUpdateFn
  fitnessFn: CanvasStrategyFitnessFn
}
```

The `fitnessFn` returns a number, the "fitness" of the strategy for the given editorState and session state. You can determine fitness for example based on "there is a single selected view", "the parent of the selected view is Flex", "the selected view has absolute positioning", etc.

the `updateFn` is the main part of the strategy.:
```typescript
type CanvasStrategyUpdateFn = (
  lifecycle: 'transient' | 'final',
  editorState: EditorState,
  currentSession: SelectModeCanvasSession,

) => {
  newSessionState: SelectModeCanvasSession
  transientFilesState: TransientFilesState
  editorStatePatch: EditorStatePatch
}
```

It takes the editorState, the currentSession, and returns a new session state, a transientFilesState, and an editorStatePatch. The filesState is the same as the DragState's old filesState: you can specify a file by path and a new array of topLevelElements and imports. The transientFilesState is respected by the navigator/canvas/inspector.

The editorStatePatch is applied as the last step of the editor `dispatch` function. You can patch any part of the editor state. For the syntax, I picked [ `Spec` ](https://github.com/kolodny/immutability-helper#available-commands) from `immutability-helper`. For example if we want to clear the highlighted views during an interaction, you can write this in your strategy update function:
```typescript
const editorStatePatch = {
  highlightedViews: { $set: [] }
}
```

*Aside: How the state patching is implemented: 
The editor `dispatch` function updates the new `unpatchedEditorState`. All the action update functions see the old `unpatchedEditorState` and operate on it and return a new one. As the last step of the `dispatch`, we call immutability-helper's `update` on the `unpatchedEditorState`, generating the `editorState` all of the application code will see. All this is to ensure that the patched editor state is a pure function of the unpatched editor state and the active Strategy.*

### Add visual controls
Once the strategy is added, we might also want to show new controls for it!
I made a simple example control called `FlexAlignControls` in `components/canvas/controls/select-mode/flex-align-controls.tsx`. I added a new object to the EditorState, at `EditorState.canvas.controls`: I intend this to be a place to put the state of controls and a point of communication between the Strategy update function and the visual controls.
How my example control works: the `FlexAlignStrategy` patches `EditorState.canvas.controls.flexAlignDropTargets`, and fills it with a bunch of rectangle coordinates. `flexAlignDropTargets` is then read and rendered by FlexAlignControls to the canvas. 
In my example, I tried to keep all logic code in the FlexAlignStrategy update function, and keep FlexAlignControls as simple as possible. My thinking was that the controls should _only_ focus on the visual display of visual controls, logic and the complicated code of a strategy should live in the strategy update function. If we can stick to this concept, I think that would provide us a nice place to put the "messy" code, and keep the canvas controls clean, and prevent leaking too much implementation detail into the react components.
We might as well experiment with putting the Strategy and the canvas control react component into the same file: that way, a single strategy plus its drawing function would be in the same place, which might make it easier in the future to read the code and understand the original intentions.


editor/src/components/canvas/canvas-strategies/canvas-strategies.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const result =
+    strategy?.updateFn?.(
+      lifecycle,
+      editorState,
+      canvasSession.sessionProps,
+      sessionStateWithStrategy,
+    ) ?? null
+
+  if (result != null) {
```
seanparsons, line 62: Could flip the conditional clauses here.

editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
```diff
@@ -0,0 +1,126 @@
...
+import type { CanvasPoint, CanvasVector } from '../../../core/shared/math-utils'
+import type {
+  EditorState,
+  EditorStatePatch,
+  TransientCanvasState,
+  TransientFilesState,
+} from '../../editor/store/editor-state'
+import type { EdgePosition } from '../canvas-types'
+
+type BoundingArea = {
```
seanparsons, line 11:
```suggestion
interface BoundingArea {
```

editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
```diff
@@ -0,0 +1,126 @@
...
+  type: 'BOUNDING_AREA'
+}
+
+type ResizeHandle = {
```
seanparsons, line 15:
```suggestion
interface ResizeHandle {
```

editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
```diff
@@ -0,0 +1,126 @@
...
+  type: 'RESIZE_HANDLE'
+  edgePosition: EdgePosition
+}
+
+type FlexGapHandle = {
```
seanparsons, line 20:
```suggestion
interface FlexGapHandle {
```

editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
```diff
@@ -0,0 +1,126 @@
...
+  start: CanvasPoint
+  mousePosition: CanvasPoint
+  drag: CanvasVector | null
+  activeControl: CanvasControlType
+  globalTime: number
+  lastTimeMouseMoved: number
+}
+
+export interface SelectModeCanvasSessionState {
+  activeStrategy: CanvasStrategy | null // should this live here? should it exist at all?
```
seanparsons, line 86: Looks like we're using it for the indicator at least.

```suggestion
  activeStrategy: CanvasStrategy | null
```

editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
```diff
@@ -0,0 +1,126 @@
...
+
+export function startNewSelectModeCanvasSession(
+  start: CanvasPoint,
+  activeControl: CanvasControlType,
+): SelectModeCanvasSession {
+  return {
+    type: 'SELECT_MODE_CANVAS_SESSION',
+    sessionProps: {
+      start: start,
+      mousePosition: start, // TODO maybe this should be independent of start?
```
seanparsons, line 103: Isn't it already?

editor/src/components/canvas/canvas-strategies/flex-align-parent-strategy.ts
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    const targetParent = MetadataUtils.getParent(editorState.jsxMetadata, draggedElement)
+    const indicatorBoxes = calcualteFlexAlignIndicatorBoxes(
+      targetParent,
+      sessionProps.mousePosition,
+    )
+
+    // if any indicator box is highlighted, we also want to change the parent's style too
+    const higlightedIndicator = indicatorBoxes.filter((b) => b.highlighted === true)[0]
```
seanparsons, line 87: `find` just gives us the first one anyway right?
```suggestion
    const higlightedIndicator = indicatorBoxes.find((b) => b.highlighted === true)
```

editor/src/components/canvas/canvas-strategies/flex-align-parent-strategy.ts
```diff
@@ -0,0 +1,260 @@
...
+              },
+            },
+          },
+        },
+      }
+    }
+  },
+}
+
+type AssociatedFlexProp = {
```
seanparsons, line 169:
```suggestion
interface AssociatedFlexProp {
```

editor/src/components/canvas/canvas-types.ts
```diff
@@ -628,11 +632,16 @@ interface ClearDragState {
   applyChanges: boolean
 }
 
-type CreateDragState = {
+export type CreateDragState = {
```
seanparsons, line 635:
```suggestion
export interface CreateDragState {
```

editor/src/components/canvas/canvas-types.ts
```diff
@@ -628,11 +632,16 @@ interface ClearDragState {
...
   action: 'CREATE_DRAG_STATE'
   dragState: DragState
 }
 
+export type UpdateCanvasSessionProps = {
```
seanparsons, line 640:
```suggestion
export interface UpdateCanvasSessionProps {
```

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1466,6 +1502,9 @@ export function getCursorFromDragState(editorState: EditorState): CSSCursor | nu
         } else {
           return null
         }
+      case 'SELECT_MODE_CANVAS_SESSION':
+        // TODO Cursor
```
seanparsons, line 1506: Subsequent work?

balazsbajorics: yep!

editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
balazsbajorics:
```suggestion
      mousePosition: start,
```

editor/src/components/canvas/canvas-strategies/canvas-strategies.ts
balazsbajorics: fixed!

editor/src/components/canvas/canvas-strategies/flex-align-parent-strategy.ts
```diff
@@ -0,0 +1,260 @@
...
+    }
+    return 0 // not fit
+  },
+  updateFn: (
+    lifecycle: 'transient' | 'final',
+    editorState: EditorState,
+    sessionProps: SelectModeCanvasSessionProps,
+    sessionState: SelectModeCanvasSessionState,
+  ): CanvasStrategyUpdateFnResult => {
+    // only apply after a certain treshold IF we hadn't already passed that treshold once
```
Rheeseyb, line 58: This check should arguably be part of the fitness function

editor/src/components/canvas/canvas-strategies/flex-gap-strategy.ts
```diff
@@ -0,0 +1,184 @@
...
+  SelectModeCanvasSessionProps,
+  SelectModeCanvasSessionState,
+} from './canvas-strategy-types'
+import { aperture, mapDropNulls } from '../../../core/shared/array-utils'
+import { stylePropPathMappingFn } from '../../inspector/common/property-path-hooks'
+import { optionalMap } from '../../../core/shared/optional-utils'
+
+export const flexGapStrategy: CanvasStrategy = {
+  name: 'Change Flex Gap',
+  fitnessFn: (editor, currentSession) => {
```
Rheeseyb, line 48: Looks like the fitness functions need to check which control is actually being interacted with

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -315,11 +320,39 @@ function applyTransientFilesState(
...
-export function clearDragState(
+let dragStateTimerHandle: any = null
+
+export function createOrUpdateDragState(
+  dispatch: EditorDispatch,
+  model: EditorState,
+  action: CreateDragState,
+): EditorState {
+  if (model.canvas.dragState == null && action.dragState.type === 'SELECT_MODE_CANVAS_SESSION') {
+    // create canavs session session, start setInterval to keep globalTime updated
```
Rheeseyb, line 331:
```suggestion
    // create canvas session, start setInterval to keep globalTime updated
```

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -397,9 +407,17 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
       onMouseDown={onMouseDown}
       onMouseMove={onMouseMove}
     >
+      <AnimatedPlaceholderBoxes />
```
Rheeseyb, line 410: Should this be behind a `when(isFeatureEnabled('Canvas Strategies')` ?

balazsbajorics: fixed!

editor/src/components/canvas/canvas-strategies/flex-align-parent-strategy.ts
balazsbajorics: thinking out loud from a user POV: if I press mouse down, do I want to immediately see the name of the active strategy, even if it has an internal rule to wait for a dragging treshold? or should we only enter the strategy once we are beyond that treshold? 
@seanparsons had an idea that the "wait for crossing treshold" could be either 1. it's own strategy, or 2. a "super-strategy" that wraps other strategies

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/flex-gap-strategy.ts
balazsbajorics: that's part of `currentSession`, so the signature is fine... I'm gonna put the check in the fitnessFn body. 
BUT in general, the two example strategies we have are more for demonstration and they are not intended to be production quality code

editor/src/components/canvas/canvas-strategies/flex-align-parent-strategy.ts
Rheeseyb: I'd argue that it would be best to wait until the threshold has been passed before communicating the active strategy, otherwise it gives the false (and potentially alarming) impression that the model is already being updated.
I quite like the idea of the "wait for crossing treshold" being its own strategy

balazsbajorics: yeah I think I err on this side too, but I have an argument for immediately selecting a strategy on mouse down: once we have a strategy picker, we probably want to show the strategy picker ASAP, and not only after a threshold.

editor/src/components/canvas/canvas-strategies/flex-gap-strategy.ts
balazsbajorics: fixed

editor/src/components/canvas/canvas-strategies/flex-align-parent-strategy.ts
balazsbajorics: I'm gonna leave the code as-is for now, will revisit in further PRs

PR #2064, opened 2022-01-05
Feature/fast refresh canvas

enidemi:
This PR introduces Vite HMR for development while keeping webpack bundling.

- Splitting the utopia-api to 2 parts. The utopia-api contains the components, and utopia-api/core has all the utility functions and non-components. The reason for this that importing module containing a React Component inside a worker triggers the react fast refresh and error layer, which is not working in a worker environment.
- The workers are using different imports and constructors for vite and webpack, so alias is added to use 2 different files for the different modes. (and jest mock)
- There are changes to the html variables and the html file structures to load them with vite and webpack too without maintaining duplicates of the files. The `HtmlWebpackPlugin` and Vite html plugin is using the same syntax for template variables.

The existing `start-` scripts are changed to use Vite, development with webpack can be run with `start-minimal-webpack` or `start-full-webpack`


editor/src/core/workers/linter/linter.ts
```diff
@@ -9,8 +9,17 @@ BrowserFS.configure({ fs: 'InMemory', options: {} }, (e) => {
...
+  ;(globalThis as any).global = globalThis
+}
 ;(global as any).BrowserFS = BrowserFS
 
+// worker polyfills
+self.process = process ?? {}
+self.process.cwd = process.cwd ?? new Function()
+import { Buffer } from 'buffer'
+self.Buffer = Buffer
+
```
Rheeseyb, line 22: Some (or all) of these shims need to happen elsewhere, but I'm not sure where :thinking: Perhaps in the entry points for each of the individual workers

Rheeseyb: I have no extracted these to `vite-shims.ts`. The embedded preview still doesn't work when running with Vite though (Buffer is still not quite being shimmed correctly for the preview)

editor/src/templates/vscode-editor-outer-iframe/index.html
```diff
@@ -9,5 +9,13 @@
       id="vscode-outer-root"
       style="width: 100%; height: 100%; display: flex; flex-direction: column;"
     ></div>
+
+    <% if (VITE) { %>
```
seanparsons, line 13: Only while hosted from Vite?

Rheeseyb: Yeah. Vite works by starting from the various `index.html` files it finds in the specified entry point folder, and resolves the source files from there, meaning we need to explicitly include the scripts in the HTML. Webpack works in the opposite direction, starting with the source files, and then injecting the output modules into the HTML via a plugin.

editor/patches/babel-eslint+10.0.1.patch
```diff
@@ -0,0 +1,24 @@
...
+#
+diff --git a/node_modules/babel-eslint/lib/analyze-scope.js b/node_modules/babel-eslint/lib/analyze-scope.js
+index 9b2a40d..05b470b 100644
+--- a/node_modules/babel-eslint/lib/analyze-scope.js
++++ b/node_modules/babel-eslint/lib/analyze-scope.js
+@@ -327,11 +327,6 @@ module.exports = function(ast, parserOptions) {
+     fallback,
+   };
+ 
+-  if (OriginalReferencer._babelEslintPatched) {
```
balazsbajorics, line 17: what does this fix?

enidemi: it tries to require a non-existing file, before this PR it was 'patched' in our webpack config

seanparsons (approved):
Can't see any reason why not.

editor/src/core/shared/code-exec-utils.ts
```diff
@@ -223,7 +223,7 @@ export const SafeFunctionCurriedErrorHandler = {
   },
 }[UTOPIA_FUNCTION_ROOT_NAME]
 
-export function SafeFunction(
+export function safeFunction(
```
balazsbajorics, line 226: why the rename?

enidemi: https://github.com/vitejs/vite/issues/6148

editor/src/templates/editor-entry-point.tsx
```diff
@@ -7,6 +7,8 @@
 // import feature switches so they are loaded before anything else can read them
 import '../utils/feature-switches'
 
+import '../vite-shims'
```
balazsbajorics, line 10: is there a reason why this is not the top import?

editor/src/templates/index.html
```diff
@@ -25,6 +25,18 @@
       crossorigin="anonymous"
     ></script>
 
+    <% if (VITE) { %>
+    <script>
+      window.global = window
+      window.require = () => {}
+      window.exports = {}
```
balazsbajorics, line 32: I'm somewhat terrified that this is needed. clearly this prevents code that uses exports from _throwing_ an error. but it will still not actually export anything!

editor/src/templates/editor-entry-point.tsx
Rheeseyb: Actually no, it should be

editor/src/templates/index.html
Rheeseyb: Hmmm I wonder if actually some of these can go in the bin now. Will verify and comment

editor/src/templates/preview/index.html
```diff
@@ -6,11 +6,22 @@
...
       type="image/png"
-      href="%UTOPIA_DOMAIN%/editor/pyramid_fullsize@2x.jpg?hash=%UTOPIA_SHA%"
+      href="<%- UTOPIA_DOMAIN %>/editor/pyramid_fullsize@2x.jpg?hash=<%- UTOPIA_SHA %>"
       as="image"
     />
 
     <!--editorScript-->
+    <% if (VITE) { %>
+    <script>
+      window.global = window
```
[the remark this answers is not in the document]
Rheeseyb: We might be able to inject this via the vite config instead

server/src/Utopia/Web/Server.hs
```diff
@@ -63,6 +63,17 @@ redirector redirections applicationToWrap request sendResponse =
       redirectTo target = sendResponse $ responseLBS temporaryRedirect307 [("Location", target)] mempty
   in  maybe passthrough redirectTo possibleRedirection
 
+viteFudgeMiddleware :: Middleware
```
balazsbajorics, line 66: ah, the Vite _fudge_ middleware 😂

balazsbajorics (approved):

server/src/Utopia/Web/Server.hs
Rheeseyb: Ah yes, this could do with a comment actually. This was to return json imports as js, right @seanparsons ?

seanparsons: Vite was returning something that ended in `.json?import` as JavaScript and with the content type for it, but our server was automagically returning it as the JSON content type because of the extension.

editor/src/templates/index.html
Rheeseyb: These have now been removed (along with the Buffer script)

server/src/Utopia/Web/Server.hs
Rheeseyb: Thanks. I've added a comment now reflecting this

editor/src/templates/editor-entry-point.tsx
Rheeseyb: Done

PR #2067, opened 2022-01-05
Move target selector section.

seanparsons:
**Problem:**
We want the target selector to be in a different position.

**Fix:**
Moved the target selector so that is is in a different position.

**Commit Details:**
- Shift the `TargetSelectorSection` up in the `Inspector` component.


enidemi (approved):

balazsbajorics (approved):
> Shift the TargetSelectorSection up in the Inspector component.

fun fact: git thinks you moved everything else _down_ instead of moving the target selector _up_ :D

PR #2069, opened 2022-01-05
Add some caching around the registerModule function

Rheeseyb:
**Problem:**
The `registerModule` function includes parsing of the params and then updating the editor model, both of which we want to minimise for performance.

**Fix:**
This work adds a layer of caching around the parsing of the second param of `registerModule` (the object containing the property controls and insertion variants), and the creation of the component descriptors. This happens via a memoized function stored against the module name (the first param of `registerModule`), so that calling `registerModule` twice (with different module names) will use different memoized functions (to ensure that we're only memoizing the relevant parts).

There is a glaring memory leak problem him, which I'm open to suggestions about - since the `cachedParseAndPrepareComponentsMap` cache is unbounded, it will keep growing as more `registerModule` calls are added (and as the first input is being typed since that will be considered to be a different module each time). The simplest choice here seems to be to trim it when it passes a given size, but I'm not really sure what arbitrary size to give it.

### DIFF
--- a/editor/src/core/property-controls/property-controls-local.ts
+++ b/editor/src/core/property-controls/property-controls-local.ts
@@ -4,26 +4,15 @@ import {
   ComponentInsertOption,
   PropertyControls,
 } from 'utopia-api'
-import deepEqual from 'fast-deep-equal'
 import { ProjectContentTreeRoot } from '../../components/assets'
 import {
-  ComponentDescriptorsForFile,
   ComponentDescriptorWithName,
   ComponentInfo,
   PropertyControlsInfo,
 } from '../../components/custom-code/code-file'
-import type { EditorDispatch } from '../../components/editor/action-types'
 import { dependenciesFromPackageJson } from '../../components/editor/npm-dependency/npm-dependency'
-import {
-  EditorState,
-  packageJsonFileFromProjectContents,
-} from '../../components/editor/store/editor-state'
-import { updatePropertyControlsInfo } from '../../components/editor/actions/action-creators'
-import {
-  parseControlDescription,
-  ParsedPropertyControls,
-  parsePropertyControls,
-} from './property-controls-parser'
+import { packageJsonFileFromProjectContents } from '../../components/editor/store/editor-state'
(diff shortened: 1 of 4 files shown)


editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -174,49 +174,91 @@ export function parseComponentToRegister(value: unknown): ParseResult<ComponentT
...
+            partiallyParseAndPrepareComponents(workers, moduleName),
+            {
+              equals: fastDeepEqual,
+              maxSize: 5,
+            },
+          )
+          cachedParseAndPrepareComponentsMap.set(moduleName, parseAndPrepareComponentsFn)
+        }
+
+        const parsedPreparedDescriptors = parseAndPrepareComponentsFn(unparsedComponents)
```
seanparsons, line 248: Looks like a candidate for a `foldEither`, but that's very minor.

seanparsons (approved):

editor/src/core/property-controls/property-controls-local.ts
Rheeseyb: Ah you're right, both this and the surrounding `if` statement would work better as a `foldEither`. Will update them :+1:

editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -174,49 +175,97 @@ export function parseComponentToRegister(value: unknown): ParseResult<ComponentT
...
+      return `registerModule second param (components): ${errorDetails.description} [${errorDetails.path}]`
+    },
+    (components: { [componentName: string]: ComponentToRegister }) => {
+      return prepareComponentDescriptorsForRegistering(workers, moduleNameOrPath, components)
+    },
+    parsedComponents,
+  )
+}
+
+type PartiallyAppliedParseAndPrepareComponents = (
```
balazsbajorics, line 203: could you put Fn in the name of the type? I think that'd make the type of cachedParseAndPrepareComponentsMap clearer

balazsbajorics (approved):

PR #2070, opened 2022-01-07
TimedCacheMap

Rheeseyb:
**Problem:**
The caching introduced in #2069 is unconstrained, meaning that it will introduce a memory leak whenever `registerModule` is introduced.

**Fix:**
This PR introduces a variant of `Map` which cleans stale keys on all access operations after a given amount of time. It takes two (optional) times (in ms) as input - the time until an entry is considered stale, and the time to wait between removing stale entries.

~I have a number of questions in `TODO`s in the code, to which I'm pretty sure the answer is always "yes", so I'm going to make those changes now (but wanted to create the PR first so others could see the code)~ This has now been resolved

### DIFF
--- a/editor/src/core/shared/timed-cache-map.ts
+++ b/editor/src/core/shared/timed-cache-map.ts
@@ -29,32 +29,42 @@ export class TimedCacheMap<K, V> {
     return this.innerMap.has(k)
   }
 
+  private updateLastAccessTimeForAllKeys() {
+    const now = Date.now()
+    for (const key of this.lastAccessTimes.keys()) {
+      this.lastAccessTimes.set(key, now)
+    }
+  }
+
   keys(): IterableIterator<K> {
-    // TODO Should this update lastAccessTimes?
     this.optionallyClean()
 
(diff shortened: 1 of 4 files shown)


editor/src/core/shared/timed-cache-map.ts
```diff
@@ -0,0 +1,111 @@
...
+  private lastAccessTimes: Map<K, number>
+  private timeSinceLastClean: number
+
+  constructor(
+    private timeToStaleInMs: number = 60000,
+    private timeToCleanInMs: number = timeToStaleInMs,
+  ) {
+    this.innerMap = new Map<K, V>()
+    this.lastAccessTimes = new Map<K, number>()
+    this.timeSinceLastClean = Date.now()
```
seanparsons, line 14: Is this actually `timeOfLastClean`?

Rheeseyb: Bah, yes!

Rheeseyb: Fixed

balazsbajorics (approved):

PR #2071, opened 2022-01-14
create or update performance test comment 

balazsbajorics:
**Problem:**
The performance test would keep adding new comments on our long-lived pull requests. For a while we thought it was the good tradeoff (we can keep an eye on the performance as it changes), but I'd like to try out how it feels if we have a single ever-updating comment at the top of the PR. (similar to Relative-CI)

**Fix:**
Use Find Comment to try to find a previous performance result comment and override it with the new one.

### DIFF
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -140,13 +140,13 @@ jobs:
           PERFORMANCE_GRAPHS_PLOTLY_API_KEY: ${{secrets.PERFORMANCE_GRAPHS_PLOTLY_API_KEY}}
         run: |
           nix-shell --arg includeServerBuildSupport false --arg includeRunLocallySupport false --run "run-puppeteer-test"
-      - name: Find Comment
-        uses: peter-evans/find-comment@v1
-        id: fc
-        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-requests.yml
```diff
@@ -140,11 +140,20 @@ jobs:
...
         run: |
           nix-shell --arg includeServerBuildSupport false --arg includeRunLocallySupport false --run "run-puppeteer-test"
-      - name: Create comment
+      - name: Find Comment
+        uses: peter-evans/find-comment@v1
+        id: fc
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: Compare with last deployed Master
```
Rheeseyb, line 149: Rather than this, can we add a more explicit prefix to the actual comment as part of the workflow step below (e.g. `Performance test results:`), otherwise this will break if we change the output of the puppeteer script.

balazsbajorics: fixed!

Rheeseyb (approved):

enidemi (approved):

PR #2072, opened 2022-01-14
removing animationsEnabled and highlightsEnabled as they were mostly …

balazsbajorics:
**Problem:**
@seanparsons and I were digging through some old code and we realized that `animationsEnabled` and `highlightsEnabled` are not respected anymore, and the Canvas Strategies' state patching system supersedes them.

**Fix:**
KILL EM'

**Note**
Highlights enabled was connected to a Keyboard Shortcut: pressing Z would set highlightsEnabled to false – and releasing Z would set highlightsEnabled to true. I deleted these shortcut alongside the editorState. Oddly enough pressing Z also hides all canvas controls - including highlights, so I think we are not losing any functionality with this deletion. I just wanted to call out to @maltenuhn if he remembers the reason we had a configurable shortcut to hide highlights – pre-set to 'Z'

### DIFF
--- a/editor/src/components/canvas/canvas.ts
+++ b/editor/src/components/canvas/canvas.ts
@@ -304,8 +304,6 @@ const Canvas = {
   },
   handleKeyUp(key: KeyCharacter, editor: EditorState, derived: DerivedState): Array<EditorAction> {
     switch (key) {
-      case 'z':
-        return [EditorActions.setHighlightsEnabled(true)]
       default:
         return []
     }
--- a/editor/src/components/canvas/controls/new-canvas-controls.tsx
+++ b/editor/src/components/canvas/controls/new-canvas-controls.tsx
@@ -101,7 +101,6 @@ export interface ControlProps {
   openFile: string | null
   hiddenInstances: Array<ElementPath>
   focusedElementPath: ElementPath | null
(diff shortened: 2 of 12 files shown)


Rheeseyb (approved):

seanparsons (approved):

maltenuhn:
Holding down `z` enters zoom mode (mouse pointer changes, click to zoom in). Obviously while in zoom mode, no highlights should trigger or other canvas controls show up, since they can make it harder to see what you want to zoom in on. Especially if you're trying to zoom in _in order to see something you know is obscured by canvas controls_.

balazsbajorics:
> Holding down `z` enters zoom mode (mouse pointer changes, click to zoom in). Obviously while in zoom mode, no highlights should trigger or other canvas controls show up, since they can make it harder to see what you want to zoom in on. Especially if you're trying to zoom in _in order to see something you know is obscured by canvas controls_.
Okay then I think the new behavior is correct: we still hide all controls while in zoom mode. (while Z is pressed)

> (This should IMO be orthogonal to the strategies, unless we consider `zoom mode` to be a strategy itself?)
Yep it's orthogonal, and for now I think we shouldn't make zoom a strategy, so all is good

PR #2073, opened 2022-01-17
Reworked absolute resize into a strategy.

seanparsons:
**Problem:**
We want to handle more canvas behaviour with strategies.

**Fix:**
Move absolute resizing behaviour into a strategy.

**Commit Details:**
- Created `resize-absolute-strategy.ts` which ports over the existing
  resize logic and makes a canvas strategy out of it.
- `pickDefaultCanvasStrategy` filters out strategies with a null
  fitness and executes the fitness function only once per strategy.
- Added some fields to `ResizeHandle` as some context is needed
  which comes from the control.
- Created `AbsoluteResizeStrategyState` and attached it to
  `SelectModeCanvasSessionState`.
- A few functions in `canvas-utils.ts` have been exported and/or
  refactored lightly to take values _from_ `ResizeDragState` rather
  than `ResizeDragState` itself.
- `ResizeControl` now includes the feature switch handling to flip
  between the two styles of resizing.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategies.ts
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategies.ts
@@ -15,28 +15,44 @@ import {
 } from './canvas-strategy-types'
 import { flexAlignParentStrategy } from './flex-align-parent-strategy'
 import { flexGapStrategy } from './flex-gap-strategy'
+import { resizeAbsoluteStrategy } from './resize-absolute-strategy'
 
-const RegisteredCanvasStrategies: Array<CanvasStrategy> = [flexGapStrategy, flexAlignParentStrategy]
+const RegisteredCanvasStrategies: Array<CanvasStrategy> = [
+  resizeAbsoluteStrategy,
+  flexGapStrategy,
+  flexAlignParentStrategy,
+]
 
 export function pickDefaultCanvasStrategy(
   editorState: EditorState,
   sessionProps: SelectModeCanvasSessionProps,
   sessionState: SelectModeCanvasSessionState,
 ): CanvasStrategy | null {
-  const applicableStrategies = RegisteredCanvasStrategies.map((s) => ({
-    fitness: s.fitnessFn(editorState, sessionProps, sessionState),
-    strategy: s,
-  })).filter((s) => {
-    // discard strategies with 0 fitness
-    return s.fitness > 0
+  // Compute the fitness results upfront.
+  const strategiesWithFitness = RegisteredCanvasStrategies.map((strategy) => {
+    return {
+      fitness: strategy.fitnessFn(editorState, sessionProps, sessionState),
+      strategy: strategy,
+    }
   })
 
-  return (
-    sortBy(applicableStrategies, (l, r) => {
-      // sort by fitness, descending
-      return r.fitness - l.fitness
-    })[0]?.strategy ?? null
-  )
+  // Filter out those which have declared (by virtue of a null result) that they
+  // cannot be candidates.
+  let filteredStrategies: Array<{ fitness: number; strategy: CanvasStrategy }> = []
+  for (const strategyWithFitness of strategiesWithFitness) {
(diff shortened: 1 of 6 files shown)


editor/src/components/canvas/canvas-strategies/canvas-strategies.ts
```diff
@@ -15,28 +15,44 @@ import {
...
   editorState: EditorState,
   sessionProps: SelectModeCanvasSessionProps,
   sessionState: SelectModeCanvasSessionState,
 ): CanvasStrategy | null {
-  const applicableStrategies = RegisteredCanvasStrategies.map((s) => ({
-    fitness: s.fitnessFn(editorState, sessionProps, sessionState),
-    strategy: s,
-  })).filter((s) => {
-    // discard strategies with 0 fitness
-    return s.fitness > 0
```
balazsbajorics, line 31: I think master was already doing the same (only calculating fitness once), the only difference is that it discards strategies with 0 fitness

editor/src/components/canvas/canvas-strategies/resize-absolute-strategy.ts
```diff
@@ -0,0 +1,473 @@
...
+    // More than one selected view which is absolute positioned, so this should work.
+    return 10
+  },
+  updateFn: (
+    lifecycle: 'transient' | 'final',
+    editorState: EditorState,
+    sessionProps: SelectModeCanvasSessionProps,
+    sessionState: SelectModeCanvasSessionState,
+  ): CanvasStrategyUpdateFnResult => {
+    if (sessionProps.activeControl.type === 'RESIZE_HANDLE') {
```
balazsbajorics, line 301: this check should (also?) live in the fitness function

editor/src/components/canvas/canvas-strategies/resize-absolute-strategy.ts
```diff
@@ -0,0 +1,473 @@
...
+      const resizeHandle = sessionProps.activeControl
+
+      const elementsToTarget = determineElementsToOperateOnForDragging(
+        editorState.selectedViews,
+        editorState.jsxMetadata,
+        false,
+        true,
+      )
+
+      const newSize = calculateNewBounds(editorState, resizeHandle, sessionProps, sessionState)
```
balazsbajorics, line 311: TODO: the snapping happens inside `calculateDraggedRectangle` inside `calculateNewBounds`, this function should return the snap guidelines and the "winning" snap, so we can feed that to the editor state patch, so that a new snapping control can draw the snaplines based on what we calculated here. 
The benefit of this approach is that we don't need to maintain two parallel places with snapping code, the calculations happen here, and the snapping controls just draw what the editor state patch instructs them to draw.

editor/src/components/canvas/canvas-strategies/resize-absolute-strategy.ts
```diff
@@ -0,0 +1,473 @@
...
+
+function applyResize(
+  jsxMetadata: ElementInstanceMetadataMap,
+  frameAndTarget: PinFrameChange,
+  element: JSXElement,
+): Array<ValueAtPath> {
+  let propsToSet: Array<ValueAtPath> = []
+  let propsToSkip: Array<PropertyPath> = []
+  const nonGroupParent = MetadataUtils.findParent(jsxMetadata, frameAndTarget.target)
+  const parentFrame =
```
balazsbajorics, line 169: TODO: we have `specialSizeMeasurements.immediateParentBounds` that is the "bounds to use" as determined by the dom-walker, and handles some edge cases this code doesn't. (like if the parent is a component with hierarchy itself)

editor/src/components/canvas/canvas-strategies/canvas-strategies.ts
seanparsons: I implemented what we discussed, which was to have `null` indicate that it cannot be handled by this strategy at all.

PR #2077, opened 2022-01-21
using css vars for canvas scroll

balazsbajorics:
**Problem:**
The canvas offset is a prop of the UiJsxCanvas root, which means every single scroll triggers a re-render of the entire canvas.

**Fix:**
I added a new wrapper around the DesignPanelRoot, which is a place where we can put css vars for our canvas and controls to use. The first two cssvars is canvas offset X and Y, which I use in the UiJsxCanvas. 

**Commit Details:**
- New wrapper around DesignPanelRoot which provides css vars
- I use a style tag to skip emotion's Dev Mode overhead, so we can have smooth scroll
- Removing canvasOffset prop from the canvas props
- Look at the Scroll Canvas test go from 63ms to 23ms

**Future Work TODO**
This approach should continue to replace all the places that are reading canvas offset from the editor state. 
The new canvas controls we are making should not even know that there are canvas offset and zoom properties in the editor state, they should be written for a cssvar world from the ground up.

### DIFF
--- a/editor/src/components/canvas/canvas-component-entry.tsx
+++ b/editor/src/components/canvas/canvas-component-entry.tsx
@@ -86,7 +86,7 @@ export const CanvasComponentEntry = React.memo((props: CanvasComponentEntryProps
           zoom: canvasProps.scale >= 1 ? `${canvasProps.scale * 100}%` : 1,
           transform:
             (canvasProps.scale < 1 ? `scale(${canvasProps.scale})` : '') +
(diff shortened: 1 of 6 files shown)


enidemi (approved):

Rheeseyb (approved):

maltenuhn:
That performance gain is quite something!

seanparsons (approved):

PR #2078, opened 2022-01-25
Integrate commands as the output of strategies.

seanparsons:
**Problem:**
Commands weren't being used to handle the output from strategies which were creating a bunch of arbitrary changes.

**Fix:**
Now all the strategies return commands in their output which get re-processed into editor state patches.

**Commit Details:**
- `applyCanvasStrategy` now uses `foldCommands` to produce the
  editor state patches needed to perform that update.
- Simplified down the `CanvasStrategyUpdateFnResult` type.
- Updated some existing strategies to have them return commands.
- `clearDragStateAndInteractionSession` now applies the state patches
  from `editorStatePatch`.
- Created new `WildcardPatch` command that supports any arbitrary
  editor state patch.
- `editorDispatch` applies the editor state patch but also attempts to
  maintain referential equality with the previously patched state.


editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -64,7 +64,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`493`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`396`)
```
Rheeseyb, line 67: These render count changes are highly suspicious. Do you know what has caused them?

seanparsons: I am also suspicious of them and have no idea why they have changed. Some went up and some went down by the same kind of amounts, which makes me _think_ it's not accounting the numbers correctly.

seanparsons: It appears to relate to the use of `keepDeepReferenceIfPossible` in `applyStatePatches`, I'm changing that to only be called if there are patches to apply. As it's needed to avoid infinite React re-renders when there's changes.

Rheeseyb: My concern is that something in this change has resulted in an actual interaction relying on the experimental code that is behind a feature flag

seanparsons: Fixed in b0543d9ad8abde6cbf924fa86e7453e777928a7d.

editor/src/components/canvas/controls/select-mode/flex-gap-controls.tsx
```diff
@@ -53,6 +54,13 @@ export const FlexGapControls = React.memo(() => {
     [dispatch, canvasScale, roundedCanvasOffset],
   )
 
+  const completeDrag = React.useCallback(
+    (event: React.MouseEvent) => {
+      dispatch([CanvasActions.clearDragState(true)], 'canvas')
```
enidemi, line 59: why is this mouseup not caught by the canvas handleCanvasEvent?

enidemi (approved):

editor/src/components/canvas/controls/select-mode/flex-gap-controls.tsx
seanparsons: Felt like the entire lifecycle should live in the strategy.

PR #2082, opened 2022-02-04
Remove reliance on Inspector context in the property target selector

Rheeseyb:
**Problem:**
The property target selector (which shows when resizing on the canvas) regressed, as it was accidentally relying on context data that is specific to the Inspector

**Fix:**
Remove the reliance on the Inspector's context, instead using the hardcoded value `[ "style" ]` as the target path used for determining where to make style changes. I've also removed some redundant functions whilst searching for any other possible problematic uses of the `stylePropPathMappingFn`.


seanparsons (approved):

balazsbajorics (approved):
note: we should probably change useContextSelector to throw an error if you try to use an unprovided Context

PR #2083, opened 2022-02-07
tell Vite that editor.tsx is not hot-reloadable

balazsbajorics:
**Problem:**
Changing something for example in `dispatch.ts` would make Vite think it has to emit a refresh to editor.tsx. However, something is wrong with how we set up the editor and/or we hit a limitation of HMR, and whenever Vite prints `[vite] hmr update /editor.tsx` it is effectively not doing anything, forcing us to manually reload the editor. This makes us distrust Vite and hate software development.

**Fix:**
I first tried to use `import.meta.hot.decline()` to tell Vite that editor.tsx is not hot-reloadable, but it wasn't working. 
So instead I'm using `import.meta.hot.on` to register to updates from Vite, and check if the updated path is `/editor.tsx` and then use `import.meta.hot?.invalidate()` to trigger a full reload.

### DIFF
--- a/editor/src/templates/editor.tsx
+++ b/editor/src/templates/editor.tsx
@@ -83,6 +83,20 @@ import { PersistenceBackend } from '../components/editor/persistence/persistence
 import { defaultProject } from '../sample-projects/sample-project-utils'
 import { createBuiltInDependenciesList } from '../core/es-modules/package-manager/built-in-dependencies-list'
 
+// TODO we should bump TS to latest (we are 0.4 versions behind!)
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+if (import.meta.hot) {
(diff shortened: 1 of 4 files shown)


editor/src/templates/editor.tsx
```diff
@@ -83,6 +83,19 @@ import { PersistenceBackend } from '../components/editor/persistence/persistence
 import { defaultProject } from '../sample-projects/sample-project-utils'
 import { createBuiltInDependenciesList } from '../core/es-modules/package-manager/built-in-dependencies-list'
 
+// TODO we should bump TS to latest (we are 0.4 versions behind!)
```
balazsbajorics, line 86: btw, we _really_ should update our typescript dependency

gbalint (approved):

Rheeseyb (approved):

enidemi (approved):

PR #2090, opened 2022-02-18
New absolute resize canvas control and selection outline

enidemi:
**Fix:**
The goal of this PR to have faster canvas controls. Updating the control position and size using the element ref and useSelectorWithCallback we can skip react rerenders.
We tried an approach using MutationObserver but it was still slow and it's easy to cause [layout thrashing](https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing#avoid_forced_synchronous_layouts).

*How to create controls in the future?*
You can use `useBoundingBox` passing in the control's target element paths and receive an element ref and update the control html element style directly in the callback. 
```typescript
const leftRef = useBoundingBox(absoluteElements, (ref, boundingBox) => {
  ref.current.style.height = boundingBox.height + 'px'
})

...

<div ref={leftRef} .../>
```

This PR contains 2 controls, behind a new feature flag `Canvas Absolute Resize Controls`. The resize control does work with or without the canvas strategies feature switch.
Absolute Resize Control: this control is used when selecting absolute positioned elements. It draws 4 visible corner controls, and 4 invisible edge drag handles.
Outline Control: this control shows a simple selection outline using the bounding box. 
The outline control has less functionality and quite experimental but it's behind the feature flag `Canvas Absolute Resize Controls`, so it can be merged. We made a follow up TODO to create the full version outline controls.

**Commit Details:**
- the new feature flag
- two control components
- new hooks


editor/src/components/canvas/controls/observer-hooks.ts
```diff
@@ -0,0 +1,82 @@
...
+import { ElementPath } from '../../../core/shared/project-file-types'
+import { fastForEach } from '../../../core/shared/utils'
+import * as EP from '../../../core/shared/element-path'
+import { CanvasRectangle, windowPoint } from '../../../core/shared/math-utils'
+import { windowToCanvasCoordinatesGlobal } from '../dom-lookup'
+
+function attachObserverToScrollingElement(
+  observerRef: React.MutableRefObject<MutationObserver | null | undefined>,
+) {
+  // TODO FIND SOMETHING NICER
```
Rheeseyb, line 11: Something nicer than what?

Rheeseyb: Oh, than the old `node-connectors` div. Yes. Please create an element specifically for this so that we're not attaching to something unrelated that may well change behaviour later.

editor/src/components/canvas/controls/observer-hooks.ts
```diff
@@ -0,0 +1,82 @@
...
+    })
+  }
+}
+
+function findTargetHtmlElement(path: ElementPath): HTMLElement | null {
+  return document.querySelector(`*[data-paths~="${EP.toString(path)}"]`)
+}
+
+export function findFramesFromDOM(targets: Array<ElementPath>): Array<CanvasRectangle> {
+  const frames: Array<CanvasRectangle> = []
```
Rheeseyb, line 26:
```suggestion
  let frames: Array<CanvasRectangle> = []
```
To highlight this variable is being mutated

editor/src/components/canvas/controls/observer-hooks.ts
```diff
@@ -0,0 +1,82 @@
...
+    observerRef.current = observer
+    return function cleanup() {
+      observer.disconnect()
+      observerRef.current = null
+    }
+  }, [observerCallback])
+
+  React.useEffect(() => {
+    if (selectedElements.length > 0) {
+      attachObserverToScrollingElement(observerRef)
```
Rheeseyb, line 62: Looks like this line should go in the previous `useEffect()`, otherwise changing the `selectedElements` without changing the `observerCallback` will result in multiple calls to `observerRef.current.observe(thisElementRerendersOnScroll`

editor/src/components/canvas/controls/observer-hooks.ts
```diff
@@ -0,0 +1,82 @@
...
+import * as EP from '../../../core/shared/element-path'
+import { CanvasRectangle, windowPoint } from '../../../core/shared/math-utils'
+import { windowToCanvasCoordinatesGlobal } from '../dom-lookup'
+
+function attachObserverToScrollingElement(
+  observerRef: React.MutableRefObject<MutationObserver | null | undefined>,
+) {
+  // TODO FIND SOMETHING NICER
+  // TODO use HTMLElement that is changing attributes on scroll (instead of cssvar) specifically for these controls
+  const thisElementRerendersOnScroll = document.getElementById('node-connectors')
```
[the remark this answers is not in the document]
balazsbajorics: the original problem it was trying to solve is that we wanted to re-run the observer callback on canvas scroll. When we scroll the canvas, the MutationObserver for elements on the canvas does not fire, because the scroll is not mutating those elements, only changing their parent offset. 

but @enidemi we should make sure this is still necessary!

it might not be necessary anymore, because the first prototype implementation relied on this so that the controls would follow the canvas scroll, but we stopped doing that and the controls use the canvas offset css vars now.

editor/src/components/canvas/controls/select-mode/absolute-resize-control.tsx
```diff
@@ -0,0 +1,349 @@
...
+  const bottomRef = React.useRef<HTMLDivElement>(null)
+
+  const selectedElements = props.selectedElements
+  const selectedElementsRef = React.useRef(selectedElements)
+  selectedElementsRef.current = selectedElements // TODO new-canvas-controls@localSelectedViews should be atom-like so we can get a live ref to it
+
+  const allSelectedElementsAbsolute = useEditorState((store) => {
+    return (
+      selectedElementsRef.current.length > 0 &&
+      selectedElementsRef.current.every((path) => {
```
Rheeseyb, line 53: Why is this using the ref rather than `selectedElements`?

balazsbajorics: we should rename selectedElements to localSelectedElements and selectedElementsRef to localSelectedElementsRef.

TLDR is that the canvas controls have a local (inner) selected elements array that is updated one frame before dispatching a selection change. this was implemented so that the visual selection change is as fast as possible, and then the slow inspector rerender happens _after_ we already drew the new controls.

I added a TODO to the paper doc about this, it has to do with the `TODO new-canvas-controls@localSelectedViews` todo.

editor/src/components/canvas/utopia-canvas-vars.tsx
```diff
@@ -1,12 +1,20 @@
...
     'DesignPanelRoot roundedCanvasOffset',
   )
+  const canvasScale = useEditorState(
+    (store) => store.editor.canvas.scale,
+    'DesignPanelRoot roundedCanvasOffset',
+  )
 
+  CanvasScrollOffset.x = roundedCanvasOffset.x
+  CanvasScrollOffset.y = roundedCanvasOffset.y
+  CanvasScale.current = canvasScale
```
Rheeseyb, line 17: Why is this happening inside a react component? This should be happening as part of the dispatch function or action handling

Rheeseyb (commented):
Please can you a whole load of comments to this, because the code clarity has suffered as a result of trying to optimise performance here.

editor/src/components/canvas/controls/observer-hooks.ts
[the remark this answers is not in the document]
enidemi: that's true, we no longer need this function

editor/src/components/canvas/utopia-canvas-vars.tsx
balazsbajorics: this is deliberate. we are mutating the CanvasScrollOffset variable exactly at the same time we are chaning the value of the css vars in the DOM. this means that code can rely on the two being in sync.

editor/src/components/canvas/controls/observer-hooks.ts
balazsbajorics: the problem is that you can't selectively unObserve an element. so when the selected views change, we call `disconnect()` which unsubscribes from all targets, then we need to re-call observe for all elements, including the scrolling observer.

editor/src/components/canvas/controls/bounding-box-hooks.ts
```diff
@@ -0,0 +1,58 @@
...
+import { ElementPath } from '../../../core/shared/project-file-types'
+import { fastForEach } from '../../../core/shared/utils'
+import { boundingRectangleArray, CanvasRectangle } from '../../../core/shared/math-utils'
+import { useRefEditorState, useSelectorWithCallback } from '../../editor/store/store-hook'
+import { MetadataUtils } from '../../../core/model/element-metadata-utils'
+import { getMetadata } from '../../editor/store/editor-state'
+
+export function useBoundingBox(
+  selectedElements: Array<ElementPath>,
+  onChangeCallback: (ref: any, boundingBox: any) => void,
```
Rheeseyb, line 11:
```suggestion
  onChangeCallback: (ref: any, boundingBox: CanvasRectangle) => void,
```

editor/src/components/canvas/controls/select-mode/absolute-resize-control.tsx
```diff
@@ -0,0 +1,334 @@
...
+  const topRightRef = useBoundingBox(absoluteElements, (ref, boundingBox) => {
+    ref.current.style.left = boundingBox.width + 'px'
+  })
+  const bottomLeftRef = useBoundingBox(absoluteElements, (ref, boundingBox) => {
+    ref.current.style.top = boundingBox.height + 'px'
+  })
+  const bottomRightRef = useBoundingBox(absoluteElements, (ref, boundingBox) => {
+    ref.current.style.left = boundingBox.width + 'px'
+    ref.current.style.top = boundingBox.height + 'px'
+  })
```
Rheeseyb, line 83: All of these refs look like they're setting the wrong values on the style object, e.g. the `leftRef` (which I'm assuming is for the left edge control) is setting the height?

Rheeseyb: Oh, sorry, ignore this! :facepalm:

editor/src/components/canvas/utopia-canvas-vars.tsx
Rheeseyb: Can you add this as a comment please

editor/src/components/canvas/controls/select-mode/absolute-resize-control.tsx
```diff
@@ -0,0 +1,334 @@
...
+    )
+    let possibleTargetProperty: LayoutTargetableProp | undefined = undefined
+    if (position.x === 0.5) {
+      possibleTargetProperty = 'height'
+    } else if (position.y === 0.5) {
+      possibleTargetProperty = 'width'
+    }
+
+    if (originalSize != null) {
+      const newDragState = updateResizeDragState(
```
seanparsons, line 303: I guess we need a follow on task from this to bring this into the canvas strategy world?

editor/src/core/model/performance-scripts.ts
```diff
@@ -30,6 +31,7 @@ export function useTriggerScrollPerformanceTest(): () => void {
   )
   const allPaths = useRefEditorState((store) => store.derived.navigatorTargets)
   const trigger = React.useCallback(async () => {
+    setFeatureEnabled('Canvas Absolute Resize Controls', true)
```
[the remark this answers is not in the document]
balazsbajorics: ah, very good catch! I enabled it for performance measurements on the branch only

editor/src/components/canvas/controls/select-mode/absolute-resize-control.tsx
balazsbajorics: yep, I updated the paper doc!

editor/src/components/canvas/controls/bounding-box-hooks.ts
balazsbajorics: fixed the types!

editor/src/components/canvas/controls/bounding-box-hooks.ts
```diff
@@ -0,0 +1,58 @@
...
+    const boundingBox = boundingRectangleArray(frames)
+
+    boundingBoxCallbackRef.current(boundingBox)
+  }, [selectedElements, metadataRef])
+
+  useSelectorWithCallback(
+    (store) => getMetadata(store.editor),
+    (newMetadata) => {
+      innerCallback()
+    },
```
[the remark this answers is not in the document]
balazsbajorics: we also call innerCallback in a useEffect a few lines below

Rheeseyb (approved):

seanparsons (approved):

PR #2091, opened 2022-02-18
Force explicit choice of unpatched vs patched state in dispatch

Rheeseyb:
**Problem:**
With the introduction of `unpatchedEditor` as part of the canvas interactions investigation, we were incorrectly reading from the patched state during the dispatch function, whilst making changes to the unpatched state. This is suspected to be related to a number of recent regressions in the editor.

**Fix:**
Force the choice of unpatched vs patched to be explicit when working inside the dispatch function.

I'm actually not sure if this is enough, so I'm tempted to completely remove the `editor` field from `EditorStore`, as I fear we could also be updating that directly or indirectly elsewhere via object spreads, which this change won't protect us against.

### DIFF
--- a/editor/src/components/canvas/dom-walker.spec.browser2.tsx
+++ b/editor/src/components/canvas/dom-walker.spec.browser2.tsx
@@ -59,6 +59,7 @@ async function renderTestEditorWithCode(appUiJsFileCode: string) {
 
   const initialEditorStore: EditorStore = {
     unpatchedEditor: emptyEditorState,
+    patchedEditor: emptyEditorState,
     editor: emptyEditorState,
     derived: derivedState,
     history: history,
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -44,6 +44,7 @@ import {
   EditorState,
   EditorStatePatch,
   EditorStore,
+  EditorStoreExplicit,
   getAllBuildErrors,
   getAllErrorsFromFiles,
   getAllLintErrors,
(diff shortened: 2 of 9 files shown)


editor/src/components/canvas/dom-walker.spec.browser2.tsx
```diff
@@ -59,6 +59,7 @@ async function renderTestEditorWithCode(appUiJsFileCode: string) {
 
   const initialEditorStore: EditorStore = {
     unpatchedEditor: emptyEditorState,
+    patchedEditor: emptyEditorState,
     editor: emptyEditorState,
```
seanparsons, line 63: Are we not removing this reference?

Rheeseyb: This is what my note at the end is about - I've kept this because of the `useEditorState` hooks relying on it, but I'm currently looking into how painful it would be to remove it completely from the `EditorStore`

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -94,11 +95,14 @@ import { isFeatureEnabled } from '../../../utils/feature-switches'
 import { isJsOrTsFile, isCssFile } from '../../../core/shared/file-utils'
 import { applyStatePatches } from '../../canvas/commands/commands'
 
-export interface DispatchResult extends EditorStore {
+interface DispatchResultFields {
   nothingChanged: boolean
   entireUpdateFinished: Promise<any>
 }
 
+export interface InnerDispatchResult extends EditorStoreExplicit, DispatchResultFields {}
```
balazsbajorics, line 103: as we talked about this previously, I think this `InnerDispatchResult` could Omit<'patchedEditor'> to make sure that the code around processing actions is not even aware of the state patching

Rheeseyb:
Closing this, as it is now superseded by #2092

PR #2092, opened 2022-02-18
Remove ambiguous "editor" field from EditorStore

Rheeseyb:
**Problem:**
With the introduction of `unpatchedEditor` as part of the canvas interactions investigation, we were incorrectly reading from the patched state during the dispatch function, whilst making changes to the unpatched state. This is suspected to be related to a number of recent regressions in the editor.

**Fix:**
This PR goes further than #2091 by removing the type `EditorStore`, and replacing it with two separate types, `EditorStoreFull` (for use only as part of the dispatch function), and `EditorStorePatched` (for use everywhere else in the code):
```ts
type EditorStoreShared = {
  derived: DerivedState
  history: StateHistory
  userState: UserState
  workers: UtopiaTsWorkers
  persistence: PersistenceMachine
  dispatch: EditorDispatch
  builtInDependencies: BuiltInDependencies
  alreadySaved: boolean
}

export type EditorStoreFull = EditorStoreShared & {
  unpatchedEditor: EditorState
  patchedEditor: EditorState
}

export type EditorStorePatched = EditorStoreShared & {
  editor: EditorState
}
```

By making this change, we can then use `EditorStorePatched` in the various editor store hooks, meaning minimal changes elsewhere in the code.

I have also updated the `InnerDispatchResult` type to explicitly omit the `patchedEditor` field.

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -453,7 +453,7 @@ export function editorDispatch(
 
   const patchedEditorState = applyStatePatches(
     frozenEditorState,
-    storedState.unpatchedEditor,
+    storedState.patchedEditor,
     frozenDerivedState.canvas.transientState.editorStatePatch,
   )
 
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -43,8 +43,7 @@ import {
   deriveState,
   EditorState,
   EditorStatePatch,
-  EditorStore,
-  EditorStoreExplicit,
+  EditorStoreFull,
   getAllBuildErrors,
   getAllErrorsFromFiles,
   getAllLintErrors,
@@ -100,8 +99,10 @@ interface DispatchResultFields {
   entireUpdateFinished: Promise<any>
 }
 
-export interface InnerDispatchResult extends EditorStoreExplicit, DispatchResultFields {}
-export interface DispatchResult extends EditorStore, DispatchResultFields {}
+type EditorStoreUnpatched = Omit<EditorStoreFull, 'patchedEditor'>
+
+export interface InnerDispatchResult extends EditorStoreUnpatched, DispatchResultFields {}
+export interface DispatchResult extends EditorStoreFull, DispatchResultFields {}
 
 function simpleStringifyAction(action: EditorAction): string {
   switch (action.action) {
@@ -120,10 +121,10 @@ export function simpleStringifyActions(actions: ReadonlyArray<EditorAction>): st
 
 function processAction(
   dispatchEvent: EditorDispatch,
-  working: EditorStoreExplicit,
+  working: EditorStoreUnpatched,
   action: EditorAction,
   spyCollector: UiJsxCanvasContextData,
-): EditorStoreExplicit {
+): EditorStoreUnpatched {
   const workingHistory = working.history
   // Sidestep around the local actions so that we definitely run them locally.
   if (action.action === 'TRANSIENT_ACTIONS') {
@@ -193,7 +194,6 @@ function processAction(
 
     return {
       unpatchedEditor: editorAfterNavigator,
-      patchedEditor: editorAfterNavigator,
       derived: working.derived,
       history: newStateHistory,
       userState: working.userState,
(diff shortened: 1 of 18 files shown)


editor/src/components/editor/store/dispatch.tsx
```diff
@@ -94,11 +94,16 @@ import { isFeatureEnabled } from '../../../utils/feature-switches'
 import { isJsOrTsFile, isCssFile } from '../../../core/shared/file-utils'
 import { applyStatePatches } from '../../canvas/commands/commands'
 
-export interface DispatchResult extends EditorStore {
+interface DispatchResultFields {
   nothingChanged: boolean
   entireUpdateFinished: Promise<any>
 }
 
+type EditorStoreUnpatched = Omit<EditorStoreFull, 'patchedEditor'>
```
balazsbajorics, line 102: thanks!

Rheeseyb: This line is my gift to you! :grinning:

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -94,11 +94,16 @@ import { isFeatureEnabled } from '../../../utils/feature-switches'
...
+
+export interface InnerDispatchResult extends EditorStoreUnpatched, DispatchResultFields {}
```
balazsbajorics, line 104: this is probably my own personal preference but I find this form to be vastly more readable:

```suggestion
export type InnerDispatchResult = EditorStoreUnpatched & DispatchResultFields 
```

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -590,44 +588,50 @@ function editorDispatchInner(
...
+      result.unpatchedEditor.canvas.dragState == null
     const metadataChanged = domMetadataChanged || spyMetadataChanged || dragStateLost
     // TODO: Should this condition actually be `&&`?
     // Tested quickly and it broke selection, but I'm mostly certain
     // it should only merge when both have changed.
     if (metadataChanged) {
-      if (result.editor.canvas.dragState != null && 'metadata' in result.editor.canvas.dragState) {
+      if (
+        result.unpatchedEditor.canvas.dragState != null &&
+        'metadata' in result.unpatchedEditor.canvas.dragState
```
seanparsons, line 608: Is that `result.unpatchedEditor.canvas.dragState.metadata != null` written in an interesting way?

seanparsons (approved):

editor/src/components/editor/store/dispatch.tsx
balazsbajorics: > Is that `result.unpatchedEditor.canvas.dragState.metadata != null` written in an interesting way?

yes! the difference is that if metadata is set to `null` or `undefined`, `'metadata' in canvas.dragState` will still return true.

Rheeseyb: I'm not sure. This wasn't part of the change (just the switch from `editor` to `unpatchedEditor`. Will try to find the source of the original code because I'm intrigued now

Rheeseyb: Ah, Balazs has saved me the time there!

seanparsons: Exactly, which doesn't seem like we want that.

Rheeseyb: I'm inclined to agree, this seems wrong to me. It was introduced in #2063. Either way, I definitely don't want to address it as part of this PR.

editor/src/components/editor/store/dispatch.tsx
Rheeseyb: Done

editor/src/components/editor/store/dispatch.tsx
balazsbajorics: In that PR, introduced a new dragState type which has no metadata field. the code here picks which metadata to update: the during-drag-special one (that makes sure the inspector does not re-render) or the regular one. Because in this new dragState there's no metadata field in the dragState, the intention of the code is to pick the regular editorState.jsxMetadata field.

probably a type guard on the dragState type would be more readable/maintainable, but ! we should be removing this code now anyways, because the interactions proposal PR supersedes it

balazsbajorics (approved):

PR #2093, opened 2022-02-18
removed inner context provider, not needed anymore

Seanmclem:
No Ticket Number..?

**Problem:**
When dynamically generating components and their prop-values via something like iterating over an array/map, the values were not populating in the inspector when selected

**Fix:**
A redundant `InspectorPropsContext.Provider` was wrapping the UIGridRow and interfering with fetching the prop values. So it was determined that it could be removed. -Which fixes the issue

**Commit Details:**
- Removed `InspectorPropsContext.Provider` from `editor/src/components/inspector/sections/component-section/property-controls-section.tsx`
- Removed supporting code for the useContext/provider in that file


Rheeseyb (approved):
Nice work!

enidemi (approved):

seanparsons (approved):

PR #2094, opened 2022-02-21
Pass existing uids when requesting parsed model

Rheeseyb:
**Problem:**
When parsing the code, we weren't preventing duplicate UIDs across multiple files if those files were parsed separately, since we were starting the parse with an [empty set for the existing UIDs](https://github.com/concrete-utopia/utopia/blob/a3cd28d96f20d7ba986cf89b977990834c2b3163/editor/src/core/workers/parser-printer/parser-printer-worker.ts#L39)

**Fix:**
Capture the existing UIDs (from the files that aren't about to be parsed), and use those to seed the set of all existing UIDs used during the parse. I've also added a test to explicitly cover this case.

### DIFF
--- a/editor/src/core/workers/common/worker-types.ts
+++ b/editor/src/core/workers/common/worker-types.ts
@@ -8,7 +8,6 @@ import {
   ParsedTextFile,
   ProjectFile,
 } from '../../shared/project-file-types'
-import { emptySet } from '../../shared/set-utils'
 import { RawSourceMap } from '../ts/ts-typings/RawSourceMap'
 
 export type FileContent = string | TextFile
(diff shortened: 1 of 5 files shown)


balazsbajorics (approved):

editor/src/core/workers/common/worker-types.ts
```diff
@@ -8,6 +8,7 @@ import {
   ParsedTextFile,
   ProjectFile,
 } from '../../shared/project-file-types'
+import { emptySet } from '../../shared/set-utils'
```
gbalint, line 11: New import but it is not used anywhere as far as I see

Rheeseyb: Fixed

gbalint (approved):

enidemi (approved):

PR #2098, opened 2022-02-22
Clean up the canvas session DragState

balazsbajorics:
**Problem:**
We have a new implementation of the strategy proposal on the branch `refactor/interactions-proposal`. We soon want to merge that branch to master.

This PR has two goals:
1. remove the now obsolete dragState-based strategy implementation, to de-clutter master and make the way for the new strategies
2. As much as possible, make the new Strategy PR zero-based. Meaning I removed some files that the new PR will have to re-introduce, instead of slightly modifying (prototype quality) code that was currently on master.

**Commit Details:**
- Removing SelectModeCanvasSession
- Remove all changes made to dragState
- remove Commands
- Remove the example controls and strategies
- Remove the strategy indicator


gbalint (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -312,9 +311,7 @@ function useStartCanvasSession(): (
       }
 
       dispatch([
-        CanvasActions.createDragState(
-          startNewSelectModeCanvasSession(start, { type: 'BOUNDING_AREA' }),
-        ),
+        // TODO actually dispatch a canvas interaction session
```
Rheeseyb, line 314: If we're making the strategies PR zero based then the whole of `useStartCanvasSession` should be deleted

balazsbajorics: fair point

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -64,7 +64,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`493`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`483`)
```
Rheeseyb, line 67: Why did the render counts go down? The only new rendered elements were behind a flag that was disabled by default, so this is ... interesting

Rheeseyb (approved):

enidemi (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
balazsbajorics: done! I actually removed the feature flag entirely

editor/src/core/performance/performance-regression-tests.spec.tsx
balazsbajorics: the CanvasStrategyIndicator was not behind a flag – might be that, or maybe some other component that I removed now?

Rheeseyb: Ah it was probably that then

PR #2102, opened 2022-02-23
perf(editor) Wrapped some path calculations in memoize calls

### LINKED ISSUE #2101: Too much state derivation when scrolling
When scrolling on the canvas we are triggering state derivation (both in the function `editor-state.tsx#deriveState` and in the `SelectModeControlContainer` class component), which is resulting in expensive calls to determine all element paths based on the metadata.

Rheeseyb:
Fixes #2101 

**Problem:**
We are performing expensive element path calculations based on unchanging metadata when scrolling on the canvas.

**Fix:**
Wrap the offending calls in `memoize` calls, as well as a function for determining element warnings (when deriving state).

**Performance Results:**
Both profiles were taken whilst running a local version of our own new performance testing project. The runs were performed in regular dev mode, but I had already confirmed that relevant function calls were roughly as expensive (and are the cause of slow scrolling) via a profile in production.

Before:
Screenshot_20220223_123153

After:
Screenshot_20220223_123100

### DIFF
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -157,6 +157,7 @@ import { PersistenceMachine } from '../persistence/persistence'
 import type { BuiltInDependencies } from '../../../core/es-modules/package-manager/built-in-dependencies-list'
 import { DefaultThirdPartyControlDefinitions } from '../../../core/third-party/third-party-controls'
 import { Spec } from 'immutability-helper'
+import { memoize } from '../../../core/shared/memoize'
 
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):
nice! can't wait to see speedy scrolling 😎 

can you please leave TODO comments and a follow up github issue to fix/review the performance getAllPaths and createOrderedElementPathsFromElements? because now when the memoization fails, we will still have to pay all those milliseconds and I have a strong suspicion that we can improve their performance.

enidemi (approved):

PR #2103, opened 2022-02-23
when multiselecting, show individual outlines of every selected element

balazsbajorics:
**Problem:**
The new ref-using outline only showed the multiselect bounds, not revealing which of the elements are actually selected.
<p>
    image<br />
<em>is the middle Rectangle selected?</em>
</p>

**Fix:**
Resurrect the individual selection outlines:
image

_look! the middle rectangle is not selected_

**Commit Details:**
- OutlineControl now takes a color prop and a targets array
- Created new MultiSelectOutlineControl that renders an OutlineControl for the bounds, and an individual OutlineControl for each target

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/simple-outline-control.tsx
+++ b/editor/src/components/canvas/controls/select-mode/simple-outline-control.tsx
@@ -1,20 +1,45 @@
 import React from 'react'
+import * as EP from '../../../../core/shared/element-path'
 import { ElementPath } from '../../../../core/shared/project-file-types'
+import { when } from '../../../../utils/react-conditionals'
 import { useColorTheme } from '../../../../uuiui'
 import { useEditorState } from '../../../editor/store/store-hook'
 import { useBoundingBox } from '../bounding-box-hooks'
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/controls/select-mode/simple-outline-control.tsx
```diff
@@ -1,20 +1,45 @@
...
+        ...localSelectedElements.map((path) => (
+          <OutlineControl key={EP.toString(path)} targets={[path]} color='primary' />
+        )),
+      ]}
+    </>
+  )
+})
+
+interface OutlineControlProps {
+  targets: Array<ElementPath>
```
Rheeseyb, line 33: With the `MultiSelectOutlineControl` does it still make sense to use an array here?

enidemi (approved):

editor/src/components/canvas/controls/select-mode/simple-outline-control.tsx
balazsbajorics: I think so! for the outline bounds case

PR #2104, opened 2022-02-23
Feature/canvas flex resize control

enidemi:
Part of the canvas control improvements.
**Fix:**
After the absolute resize controls the next is the flex resize controls to update with the new performance optimized bounding box hook + scaling with css vars.
The old control is unchanged.

Similarly to the absolute resize control the control mousedown callback needs to be updated in the future with the interaction state for the strategies.

**Commit Details:**
- created flex resize controls
- updated the render count tests, this changed because of the extra `when` in new-canvas-controls


editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -64,7 +64,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`483`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`485`)
```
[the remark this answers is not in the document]
enidemi: I was also surprised but the conditional rendering is behind it

balazsbajorics (approved):

editor/src/components/canvas/controls/select-mode/flex-resize-control.tsx
```diff
@@ -0,0 +1,258 @@
...
+})
+
+interface ResizeEdgeProps {
+  cursor: CSSCursor
+  position: EdgePosition
+  direction: 'horizontal' | 'vertical'
+  enabledDirection: EdgePosition
+}
+
+const ResizeEdgeMouseAreaSize = 12
```
gbalint, line 101: Don't we use snake case for global consts? I think that is more idiomatic

editor/src/components/canvas/controls/select-mode/flex-resize-control.tsx
```diff
@@ -0,0 +1,258 @@
...
+    // TODO update this to call createInteractionState to use canvas strategies
+    const centerBasedResize = event.altKey
+    const keepAspectRatio = event.shiftKey // || props.elementAspectRatioLocked ???
+    const enableSnapping = !event.metaKey
+    const canvasPositions = windowToCanvasCoordinatesGlobal(
+      windowPoint({ x: event.nativeEvent.x, y: event.nativeEvent.y }),
+    )
+    const start: CanvasPoint = canvasPositions.canvasPositionRaw
+    const originalFrames = getOriginalFrames(selectedViews, metadata)
+    const isMultiSelect = selectedViews.length !== 1
```
gbalint, line 163: I would write `selectedViews.length > 1`
I know it can not be `0` here, but still that is more expressive in my opinion

editor/src/components/canvas/controls/select-mode/flex-resize-control.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+    const originalSize = boundingRectangleArray(
+      selectedViews.map((path) => MetadataUtils.getFrameInCanvasCoords(path, metadata)),
+    )
+    const flexDirection = MetadataUtils.getFlexDirection(
+      MetadataUtils.getParent(metadata, selectedViews[0]),
+    )
+    let possibleTargetProperty: LayoutTargetableProp | undefined = undefined
```
gbalint, line 171: Why `undefined` and not `null`?

gbalint (approved):

editor/src/components/canvas/controls/select-mode/flex-resize-control.tsx
enidemi: this part will be changed soon, it's waiting for the strategies

editor/src/components/canvas/controls/select-mode/flex-resize-control.tsx
enidemi: it looks better! fixed

PR #2106, opened 2022-02-23
Canvas Highlight Performance Improvements

Rheeseyb:
A couple of improvements to canvas highlighting:

1. `findFirstParentWithValidElementPath` was in serious need of some optimisation, as it was performing excessive comparisons of element paths, as well as excessive conversions of those paths (from dynamic to static). This change replaces those comparisons of the path objects with comparisons of the stringified paths, and uses a `Set` for the intersection check rather than an array.
2. Functions in `element-metadata-utils.ts` for fetching child or root view metadata were performing extra work to first find the paths of those elements before fetching the metadata, which is not necessary. 

There is also a minor change in the `inspector.tsx` - an inner function has been extracted from a loop.


enidemi (approved):

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -417,7 +417,7 @@ export const Button = styled.button({
 });
 
 export const CardList = (props) => {
-  cards = [1, 2, 3, 4, 5];
+  const cards = [1, 2, 3, 4, 5];
```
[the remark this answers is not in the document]
Rheeseyb: Ok sure, good shout :+1:

seanparsons (approved):

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -31,35 +31,51 @@ export function findParentSceneValidPaths(target: Element): Array<ElementPath> |
...
+      return staticAndDynamicTargetElementPaths.find(
+        (staticAndDynamic) => staticAndDynamic.static === validPath,
+      )?.dynamic
+    },
+    [...validStaticElementPaths],
+  )
 
   if (filteredValidPathsMappedToDynamic.length > 0) {
-    return last(filteredValidPathsMappedToDynamic) ?? null
+    const asPaths = filteredValidPathsMappedToDynamic.sort((l, r) => EP.depth(l) - EP.depth(r))
```
balazsbajorics, line 69: maybe sortedSomethingSomething is a better name

Rheeseyb: Done :+1:

balazsbajorics (commented):
👍

PR #2112, opened 2022-02-25
Feature/canvas strategies

balazsbajorics:
This is the Canvas Strategies Mega PR to close the long-lived-branch `refactor/interactions-proposal`. https://github.com/concrete-utopia/utopia/pull/2080

Paper docs that were related to the work: 

[Canvas Track Roadmap](https://paper.dropbox.com/doc/Canvas-Track-Roadmap--BcxlUb6o3_FimG3DUSJBO_ryAg-GOjRfmwKNi5WKy7MbmDUw)

older paper docs: [Canvas Interactions Roadmap](https://paper.dropbox.com/doc/Canvas-Interactions-Roadmap--BcxhM_cVeAh0yuxuLRZ_9xQzAg-9AKASINajdVgpzor1rZdr), [Canvas Modernisation](https://paper.dropbox.com/doc/Track-Canvas-Modernisation--BcwLnvxbjxL4~7uqgnMBpGA_Ag-FlkNmvrA9pyvFxUQVZbvn), 

**Commit Details:**
- This PR takes the commits from  `refactor/interactions-proposal`, and removes every non-essential code we could find:
- No actual strategies, only the strategy lifecycle methods and management code
- It contains only two example commands
- No new canvas controls


editor/src/components/canvas/canvas-types.ts
```diff
@@ -649,13 +669,22 @@ type ZoomUI = {
   zoomIn: boolean
 }
 
+type SetUsersPreferredStrategy = {
+  action: 'SET_USERS_PREFERRED_STRATEGY'
+  strategyName: string // TODO limit it to string literal union of registered strategy names?
```
gbalint, line 674: I think this TODO should be done. In a way, that a strategy has a type, which is a string literal union, and a user digestible name, which can be shown in a ui, but not used as an identifier.

balazsbajorics: so you propose to add a new field to the strategy, which is the non-user-facing strategy name? and restrict this type here to only valid string literals, right?

editor/src/components/canvas/canvas-strategies/canvas-strategy-inspector.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const activeStrategy = useEditorState(
+    (store) => isStrategyActive(store.strategyState),
+    'CanvasStrategyInspector activeStrategy',
+  )
+  const commandsToList = useEditorState(
+    (store) => store.strategyState.commandDescriptions,
+    'CanvasStrategyInspector accumulatedCommands',
+  )
+
+  if (!activeStrategy) {
```
seanparsons, line 17: Could flip this for clarity.

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -41,22 +41,26 @@ import { useWindowToCanvasCoordinates } from '../../dom-lookup-hooks'
...
 const DRAG_START_TRESHOLD = 2
 
 export function isResizing(editorState: EditorState): boolean {
   const dragState = editorState.canvas.dragState
   return (
-    dragState?.type === 'RESIZE_DRAG_STATE' &&
-    getDragStateDrag(dragState, editorState.canvas.resizeOptions) != null
+    (dragState?.type === 'RESIZE_DRAG_STATE' &&
+      getDragStateDrag(dragState, editorState.canvas.resizeOptions) != null) ||
+    editorState.canvas.interactionSession != null
```
seanparsons, line 54: This doesn't _feel_ right as not every interaction is resizing, maybe needs a `TODO` to indicate it needs removing/replacing later?

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -41,22 +41,26 @@ import { useWindowToCanvasCoordinates } from '../../dom-lookup-hooks'
...
 }
 
 export function isDragging(editorState: EditorState): boolean {
   const dragState = editorState.canvas.dragState
   return (
-    dragState?.type === 'MOVE_DRAG_STATE' &&
-    getDragStateDrag(dragState, editorState.canvas.resizeOptions) != null
+    (dragState?.type === 'MOVE_DRAG_STATE' &&
+      getDragStateDrag(dragState, editorState.canvas.resizeOptions) != null) ||
+    editorState.canvas.interactionSession != null
```
seanparsons, line 63: This doesn't _feel_ right as not every interaction is dragging, maybe needs a `TODO` to indicate it needs removing/replacing later?

editor/src/core/shared/jsx-attributes.ts
```diff
@@ -2,7 +2,16 @@ import * as ObjectPath from 'object-path'
 import { MapLike } from 'typescript'
 import { UtopiaUtils } from 'utopia-api/core'
 import { findLastIndex, uniqBy } from './array-utils'
-import { Either, isLeft, left, mapEither, reduceWithEither, right, sequenceEither } from './either'
+import {
```
seanparsons, line 5: Looks like this change could be scrapped.

editor/src/components/editor/global-shortcuts.ts
```diff
@@ -372,6 +373,11 @@ export function handleKeyDown(
     directionModifier: -1 | 1,
     adjustment: 1 | 10,
   ): Array<EditorAction> {
+    if (isFeatureEnabled('Canvas Strategies')) {
+      // Disable these keyboard shortcuts so they don't interfere with strategies
```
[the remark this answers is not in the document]
balazsbajorics: I think so

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
balazsbajorics: added a comment in 6b90b5b8

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
balazsbajorics: added a comment in 6b90b5b8

editor/src/components/canvas/canvas-strategies/canvas-strategy-inspector.tsx
balazsbajorics: fixed in c4dead52

editor/src/components/canvas/canvas-types.ts
balazsbajorics: update: we decided to only fix this after this is merged to master

editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -0,0 +1,321 @@
...
+  }
+}
+
+export function hasModifiersChanged(
+  prevInteractionData: InputData | null,
+  interactionData: InputData | null,
+): boolean {
+  return (
+    interactionData?.type === 'DRAG' &&
+    prevInteractionData?.type === 'DRAG' &&
```
gbalint, line 295: If this is only for dragging, then the function name should reflect that - or maybe only allow DragInteractionData parameters?

editor/src/components/canvas/commands/commands.ts
```diff
@@ -0,0 +1,197 @@
...
+    transient: transient,
+    patch: patch,
+  }
+}
+
+export interface StrategySwitched extends BaseCommand {
+  type: 'STRATEGY_SWITCHED'
+  reason: 'automatic' | 'user-input'
+  newStrategy: string
+  accumulatedCommands: boolean
```
gbalint, line 49: Is this really a boolean? Then the naming is pretty bad

editor/src/components/canvas/canvas-types.ts
gbalint: Note: we postponed this because there are no actual strategies in this PR

editor/src/components/editor/store/dispatch-strategies.tsx
```diff
@@ -0,0 +1,600 @@
...
+        if (result.unpatchedEditor.canvas.interactionSession?.userPreferredStrategy != null) {
+          const userChangedStrategy =
+            result.unpatchedEditor.canvas.interactionSession?.userPreferredStrategy !=
+            storedState.unpatchedEditor.canvas.interactionSession?.userPreferredStrategy
+          if (userChangedStrategy) {
+            return interactionUserChangedStrategy(storedState, result)
+          }
+        }
+
+        const strategy = findCanvasStrategyFromDispatchResult(result)
```
gbalint, line 591: This is probably too big task to include in this PR, but I don't like that we have to find the canvas strategy to decide whether there is a strategy change, and then inside `interactionStrategyChangeStacked` we find the canvas strategy again. It seems this big `if` here to handle the different interaction strategy cases should be _after_ finding the canvas strategy, and not before. That would also decrease the code duplication between the different `interaction*` functions.
Note that this will be even worse after we actually apply the accumulated commands before finding the strategy, even more code will has to be duplicated.
I think it would make sense to refactor this after the PR is merged

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -0,0 +1,231 @@
...
+  return RegisteredCanvasStrategies.filter((strategy) => {
+    return strategy.isApplicable(canvasState, interactionSession, metadata)
+  })
+}
+
+const getApplicableStrategiesSelector = createSelector(
+  (store: EditorStorePatched): InteractionCanvasState => {
+    return {
+      selectedElements: store.editor.selectedViews,
+      // metadata: store.editor.jsxMetadata, // We can add metadata back if live metadata is necessary
```
Rheeseyb, line 29: This line might as well go since we're pulling the live metadata below anyway
```suggestion
```

editor/src/components/canvas/commands/commands.ts
gbalint: Furthermore, it seems this is unused, let's just remove it

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -0,0 +1,230 @@
...
+    return { strategy: currentBestStrategy, previousStrategy: previousStrategy }
+  }
+}
+
+function pickStrategy(
+  sortedApplicableStrategies: Array<StrategyWithFitness>,
+  interactionSession: InteractionSession,
+  previousStrategyName: string | null,
+): { strategy: StrategyWithFitness | null; previousStrategy: StrategyWithFitness | null } {
+  // FIXME Explicitly picking a strategy will prevent natural handovers that otherwise should occur
```
Rheeseyb, line 142: Is this captured in the roadmap doc?

editor/src/components/editor/store/dispatch-strategies.tsx
gbalint: done in 6ae9ccf2139367612c61e0c74a73915ecd254138

editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -0,0 +1,321 @@
...
+}
+
+export function updateInteractionViaKeyboard(
+  currentState: InteractionSession,
+  addedKeysPressed: Array<KeyCharacter>,
+  keysReleased: Array<KeyCharacter>,
+  modifiers: Modifiers,
+  sourceOfUpdate: CanvasControlType,
+): InteractionSessionWithoutMetadata {
+  if (currentState.interactionData.type === 'KEYBOARD') {
```
Rheeseyb, line 179: Can we replace this with a `switch` statement please so that we can ensure completeness with the usual:
```ts
    default:
      const _exhaustiveCheck: never = currentState.interactionData
      throw new Error(`Unhandled interactionData type ${JSON.stringify(currentState.interactionData)}`)
```

editor/src/components/canvas/canvas-strategies/interaction-state.ts
balazsbajorics: renamed

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -122,6 +123,7 @@ export const CanvasWrapperComponent = React.memo(() => {
         >
           {safeMode ? <SafeModeErrorOverlay /> : <ErrorOverlayComponent />}
           <ModeSelectButtons />
+          <CanvasStrategyIndicator />
```
Rheeseyb, line 126: Shouldn't this be behind a feature flag check?

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
balazsbajorics: I added an entry to paper now

editor/src/components/editor/actions/action-utils.ts
```diff
@@ -213,3 +218,31 @@ export function isFromVSCode(action: EditorAction): boolean {
       return false
   }
 }
+
+export function isClearInteractionSession(action: EditorAction): action is ClearInteractionSession {
```
Rheeseyb, line 222: This function should absolutely not be a type guard

```suggestion
export function isClearInteractionSession(action: EditorAction): boolean {
```

editor/src/components/editor/actions/action-utils.ts
```diff
@@ -213,3 +218,31 @@ export function isFromVSCode(action: EditorAction): boolean {
...
+    case 'CLEAR_INTERACTION_SESSION':
+      return true
+    default:
+      return false
+  }
+}
+
+export function shouldApplyClearInteractionSessionResult(
+  action: EditorAction,
+): action is ClearInteractionSession {
```
Rheeseyb, line 237: Likewise here

```suggestion
export function shouldApplyClearInteractionSessionResult(
  action: EditorAction,
): boolean {
```

editor/src/components/editor/editor-component.tsx
```diff
@@ -129,6 +134,23 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
 
   const onWindowKeyDown = React.useCallback(
     (event: KeyboardEvent) => {
+      if (isFeatureEnabled('Canvas Strategies')) {
+        const existingInteractionSession = editorStoreRef.current.editor.canvas.interactionSession
+        // TODO: we don't handle new interaction started with keyboard now
```
Rheeseyb, line 139: Is this / should this be captured in the road map?

editor/src/components/editor/actions/action-utils.ts
balazsbajorics: why not? it's called `isClearInteractionSession` and returns if the action is CLEAR_INTERACTION_SESSION

editor/src/components/editor/actions/action-utils.ts
balazsbajorics: I agree this should be a boolean

editor/src/components/editor/actions/action-utils.ts
Rheeseyb: Because it isn't a type guard! Otherwise we're telling TS that a `TRANSIENT_ACTIONS` action or `ATOMIC` action is a `ClearInteractionSession`, when clearly neither of those are.

balazsbajorics: ah yeah you are right about that :D

Rheeseyb (approved):

editor/src/components/editor/actions/action-utils.ts
balazsbajorics: fixed both in 68f29127

editor/src/components/editor/editor-component.tsx
balazsbajorics: yep I created an entry for this in the paper doc

editor/src/components/editor/global-shortcuts.ts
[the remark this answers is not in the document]
balazsbajorics: I added a todo to follow up on keyboard handling for the strategies

seanparsons (approved):

gbalint (approved):

PR #2116, opened 2022-02-28
Limit the amount of tracing done by the performance tests.

seanparsons:
**Problem:**
If a performance test does a lot of work or takes a little longer than usual the tracing file can become multiple hundreds of megabytes in size. Which then means this code:

```
```

Fails with a bizarre error about not being able to create a string of longer than 0x1fffffe8 characters.

**Fix:**
The tracing is now supplied with a category of `blink.user_timing`, which covers the performance mark events we're interested in and brings the trace file down to at most a few hundred lines.

**Commit Details:**
- Constrain the tracing to the `blink.user_timing` category so that
  the tracing file doesn't balloon to a gigantic size.


balazsbajorics (approved):
neat!

enidemi (approved):

Rheeseyb (approved):

PR #2120, opened 2022-03-01
Added absolute move and reparent strategies

gbalint:
We need to have strategies to validate how the canvas strategy based interaction handling work.
I migrated two example strategies (absolute move and absolute reparent) from the experimental branch.
The strategies are not perfect (e.g. reparent doesn't support reparent to different files), but they seem to be good enough as starting points.


Rheeseyb (approved):

balazsbajorics (approved):

seanparsons (approved):

editor/src/components/canvas/commands/command-runners.ts
```diff
@@ -0,0 +1,361 @@
...
+import {
+  AdjustNumberProperty,
+  CommandFunction,
+  PathMappings,
+  ReparentElement,
+  SetProperty,
+  UpdateSelectedViews,
+} from './commands'
+
+export const runAdjustNumberProperty: CommandFunction<AdjustNumberProperty> = (
```
balazsbajorics, line 48: I approved the strategy but now I wonder if splitting comands and command-runners is the sensible thing, or if we should split them into files per command. 
So one file per command that holds the strategy type, creator function, and the runner.
I think in hindsight the gigantic actions.ts file was a mistake that we shouldn't repeat.

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -3,14 +3,19 @@ import { createSelector } from 'reselect'
...
 import { arrayEquals } from '../../../core/shared/utils'
-import { InteractionSession, StrategyState } from './interaction-state'
 import { InnerDispatchResult } from '../../editor/store/dispatch'
 import { EditorStorePatched } from '../../editor/store/editor-state'
 import { useEditorState } from '../../editor/store/store-hook'
 import { CanvasCommand } from '../commands/commands'
+import { absoluteMoveStrategy } from './absolute-move-strategy'
+import { absoluteReparentStrategy } from './absolute-reparent-strategy'
 import { CanvasStrategy, ControlWithKey, InteractionCanvasState } from './canvas-strategy-types'
+import { InteractionSession, StrategyState } from './interaction-state'
```
balazsbajorics, line 13: I took the liberty to run Organize Imports on some of these files, sorry for the noise, but I thought it's the right opportunity to sneak this in!

PR #2122, opened 2022-03-01
Tests for some of the core strategies code.

seanparsons:
**Problem:**
There's not enough tests covering the internals of the strategy code.

**Fix:**
Implemented some tests for the internals of the strategy code.

**Commit Details:**
- Strategies can now be supplied directly to a bunch of the functions rather
  than relying on the embedded `RegisteredCanvasStrategies`.
- Added unit test code for the internal strategy functions.

### DIFF
--- a/editor/src/components/editor/store/dispatch-strategies.spec.tsx
+++ b/editor/src/components/editor/store/dispatch-strategies.spec.tsx
@@ -12,16 +12,50 @@ import { EditorDispatch, notLoggedIn } from '../action-types'
 import * as History from '../history'
 import { DummyPersistenceMachine } from '../persistence/persistence.test-utils'
 import { DispatchResult, editorDispatch } from './dispatch'
-import { interactionCancel } from './dispatch-strategies'
+import {
+  interactionCancel,
+  interactionHardReset,
+  interactionStart,
+  interactionStrategyChangeStacked,
+  interactionUpdate,
+  interactionUserChangedStrategy,
+} from './dispatch-strategies'
 import { createEditorState, deriveState, EditorStoreFull } from './editor-state'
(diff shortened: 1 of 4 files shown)


editor/src/components/editor/store/dispatch-strategies.spec.tsx
```diff
@@ -83,3 +123,589 @@ describe('interactionCancel', () => {
...
+        { type: 'BOUNDING_AREA', target: EP.elementPath([['aaa']]) },
+      ),
+    )
+    const actualResult = interactionStart(
+      [testStrategy],
+      editorStore,
+      dispatchResultFromEditorStore(editorStore),
+    )
+    expect(actualResult.newStrategyState).toMatchInlineSnapshot(`
+      Object {
```
balazsbajorics, line 168: I fear that if we write object snapshots like this for `newStrategyState` and `interactionData`, it will be an absolute pain to change / refactor these once we have dozens of tests. 
could this test get away by only checking to see if the strategy was picked and the initial state was created? somthing more similar to how you are checking the patched and unpatched canvas scale

seanparsons: I'd surmise these functions aren't likely to change a lot, if we check less of the data then we're making the tests less likely to catch actual breakages. Plus this is being tested with a test strategy that is local to the tests themselves so other strategies changing wont affect these tests at all.

balazsbajorics: yeah I get it that this is a test strategy, I'm just concerned whether adding a new key to StrategyState or interactionData will be breaking dozens of tests in the new future

Rheeseyb (approved):

PR #2124, opened 2022-03-01
Adding performance scripts for canvas highlighting

Rheeseyb:
This is the first part of a 2 part PR. This part introduces the changes on the editor side for supporting 2 types of canvas highlight performance tests.

This work extracts the core logic used when determining which element to highlight based on the mouse's position on the canvas, and the provides 2 scripts for profiling the performance of that code (when `cmd` is held down vs when it isn't). Since these scripts obviously aren't already available on staging, to ensure we can run the existing performance tests (to ensure this hasn't introduced regressions there) I have separated the changes to the puppeteer code into the next PR.


seanparsons (approved):

editor/src/components/menubar/menubar.tsx
```diff
@@ -355,6 +359,12 @@ export const Menubar = React.memo(() => {
           <Tile style={{ marginTop: 12, marginBottom: 12 }} size='large'>
             <a onClick={onTriggerResizeTest}>P R</a>
           </Tile>
+          <Tile style={{ marginTop: 12, marginBottom: 12 }} size='large'>
+            <a onClick={onTriggerRegularHighlightTest}>PRH</a>
```
balazsbajorics, line 363: love the new mistery buttons! :D

balazsbajorics (approved):

editor/src/components/menubar/menubar.tsx
Rheeseyb: I was cursing you whilst adding them :grinning:

PR #2126, opened 2022-03-02
Typesafe CanvasStrategy ids

gbalint:
We identify canvas strategies by names, which are simple strings. This is not typesafe, typos can cause problems, and uniqueness is not guaranteed.

In this PR I introduced an id field in CanvasStrategy, which has a string literal union type. This id is typesafe, and we can still use the name field to contain a user digestible name.

Questions:
- uniqueness is still not guaranteed, but maybe it is not that important, we don't have that quarantee with actions neither.
- do we need this at all? We could just store the strategy objects everywhere, they are constants and never recreated, so they always keep their references.

### DIFF
--- a/editor/src/components/canvas/canvas-actions.ts
+++ b/editor/src/components/canvas/canvas-actions.ts
@@ -1,4 +1,5 @@
 import type { CanvasPoint, CanvasVector } from '../../core/shared/math-utils'
+import { CanvasStrategyId } from './canvas-strategies/canvas-strategy-types'
 import {
   InteractionSession,
   InteractionSessionWithoutMetadata,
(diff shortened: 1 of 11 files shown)


Rheeseyb (approved):

Rheeseyb:
> uniqueness is still not guaranteed, but maybe it is not that important, we don't have that quarantee with actions neither.

I think this is good enough, because it forces you to refer to the full list of IDs when adding a new strategy (like we have to with actions).

> do we need this at all? We could just store the strategy objects everywhere, they are constants and never recreated, so they always keep their references.

I agree, we could just store the strategy objects themselves wherever we currently use these IDs, with the caveat that we'd want to ensure we only ever compare them with reference equality any time they are featured in a `keedDeepEquality` call

balazsbajorics:
> > do we need this at all? We could just store the strategy objects everywhere, they are constants and never recreated, so they always keep their references.
> 
> I agree, we could just store the strategy objects themselves wherever we currently use these IDs, with the caveat that we'd want to ensure we only ever compare them with reference equality any time they are featured in a `keedDeepEquality` call

I think I prefer this ID, for example it makes the redux devtool a little bit better

PR #2131, opened 2022-03-03
Refactor dispatch-strategies to reduce code duplication

gbalint:
The motivation for this PR is this comment: https://github.com/concrete-utopia/utopia/pull/2112#discussion_r816051646

The decision whether there is a strategy change or not can only be made after the strategy has been chosen. This resulted in the hard to understand, inefficient, and not safe code in which we searched for the canvas strategy twice: First to decide which `interaction*` function to run, and then start everything from scratch again inside those functions. 

As a fix for this duplication I merged `interactionUpdate`, `interactionUserChangedStrategy` and `interactionStrategyChangeStacked` into `interactionUpdate`, and created 2 different handler functions on how to apply the new strategies.

### DIFF
--- a/editor/src/components/editor/store/dispatch-strategies.spec.tsx
+++ b/editor/src/components/editor/store/dispatch-strategies.spec.tsx
@@ -16,9 +16,7 @@ import {
   interactionCancel,
   interactionHardReset,
   interactionStart,
-  handleStrategyChangeStacked,
   interactionUpdate,
-  handleUserChangedStrategy,
 } from './dispatch-strategies'
 import { createEditorState, deriveState, EditorStoreFull } from './editor-state'
 import * as EP from '../../../core/shared/element-path'
--- a/editor/src/components/editor/store/dispatch-strategies.spec.tsx
+++ b/editor/src/components/editor/store/dispatch-strategies.spec.tsx
@@ -491,7 +491,7 @@ describe('interactionHardReset', () => {
   })
 })
 
-describe('interactionUpdate with strategy change', () => {
+describe('interactionUpdate with stacked strategy change', () => {
   it('steps an interaction session correctly', () => {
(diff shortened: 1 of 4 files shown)


editor/src/components/editor/store/dispatch-strategies.spec.tsx
```diff
@@ -502,7 +504,8 @@ describe('interactionStrategyChangeStacked', () => {
       interactionSession.interactionData.prevDrag = canvasPoint({ x: 30, y: 120 })
     }
     const editorStore = createEditorStore(interactionSession)
-    const actualResult = interactionStrategyChangeStacked(
+    editorStore.strategyState.currentStrategy = 'EMPTY_TEST_STRATEGY' as CanvasStrategyId
+    const actualResult = interactionUpdate(
```
seanparsons, line 508: This looks to still be in a test block marked as `interactionStrategyChangeStacked`.

seanparsons (approved):

editor/src/components/editor/store/dispatch-strategies.spec.tsx
```diff
@@ -16,9 +16,9 @@ import {
   interactionCancel,
   interactionHardReset,
   interactionStart,
-  interactionStrategyChangeStacked,
+  handleStrategyChangeStacked,
```
balazsbajorics, line 19: these new "inner" functions seem to be unused here. I think it'd make the code nicer if they wouldn't even be exported from `dispatch-strategies.tsx`!

balazsbajorics (approved):

editor/src/components/editor/store/dispatch-strategies.spec.tsx
gbalint: true, fixed in 830607b4fd239aa890d919a32b8ef6d3eb7887da

PR #2132, opened 2022-03-03
Use a tree to hold parent-child relationships.

seanparsons:
**Problem:**
In a couple of places the hierarchy of elements was being traversed, but this was being done by starting at the roots and then for each element the children were identified by checking the path against each and every element in the metadata. That approximates O(n^2) where n is the number of elements, for a project with 161 elements it was making two calls to `isChildOf` for each element which equates to about 26,000 times.

**Fix:**
This introduces a simple tree of parents and children for element paths, which can be built once and is then used for all the lookups instead of iterating through the metadata.

Currently it is being created twice in two different places, but probably should be shared if it's used in more places than this in the future as it does (in a large project) have a cost of 1ms per invocation. However even with that cost it has dropped the runtime of the mouse move event by 6ms.

**Screenshots:**
Before:
before
After:
after

**Commit Details:**
- Added `buildTree`, which takes a bunch of element paths and produces
  a n-way tree from the paths so that children of a target can be identified
  quickly.
- Implemented `printTree` which takes the tree produced by `buildTree` and
  returns a string formatted for testing purposes.
- Added `forEachChildOfTarget` to call a callback for each child path
  of a given path.
- Simplified some functions like `getRootViews`, `getChildrenPaths`,
  `getImmediateChildrenPaths` and `getStorboardMetadata`.
- `getAllPathsIncludingUnfurledFocusedComponents` and `getFramesInCanvasContext`
  now use the new tree to identify the children of targets.

### DIFF
--- a/editor/src/core/shared/element-path-tree.ts
+++ b/editor/src/core/shared/element-path-tree.ts
@@ -1,0 +1,113 @@
+import { ElementPath, ElementPathPart } from './project-file-types'
+import * as EP from './element-path'
+import { fastForEach } from './utils'
+
+export interface ElementPathTree {
+  path: ElementPath
+  children: ElementPathTreeRoot
+}
+
+function emptyElementPathTree(path: ElementPath): ElementPathTree {
+  return {
+    path: path,
+    children: [],
+  }
+}
+
+export type ElementPathTreeRoot = Array<ElementPathTree>
+
+export function buildTree(elementPaths: Array<ElementPath>): ElementPathTreeRoot {
+  let result: ElementPathTreeRoot = []
+  let workingRoot: ElementPathTreeRoot = result
+  fastForEach(elementPaths, (elementPath) => {
+    workingRoot = result
+    const totalParts = elementPath.parts.reduce((workingCount, elem) => {
+      return workingCount + elem.length
+    }, 0)
+    for (let pathSize = 1; pathSize <= totalParts; pathSize++) {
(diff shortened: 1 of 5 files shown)


editor/src/core/shared/element-path-tree.ts
```diff
@@ -0,0 +1,113 @@
...
+function emptyElementPathTree(path: ElementPath): ElementPathTree {
+  return {
+    path: path,
+    children: [],
+  }
+}
+
+export type ElementPathTreeRoot = Array<ElementPathTree>
+
+export function buildTree(elementPaths: Array<ElementPath>): ElementPathTreeRoot {
```
Rheeseyb, line 19: This function has replicated the behaviour of the code it has optimised, but there's a huge caveat here that we need to be careful of. This function (and the code it replaces) doesn't guarantee the correct ordering of the tree, so e.g. for the jsx:
```
<div data-uid='aaa'>
  <div data-uid='bbb'/>
  <div data-uid='ccc'/>
</div>
```
both this function and the previous code can return the children paths in the order `[ 'aaa/ccc', 'aaa/bbb' ]` (depending on what interactions in the editor lead to this point). So with that in mind, please can you either add a comment to that effect, or name the function something that clarifies the order isn't guaranteed.

seanparsons: Added caveat note in 2ad32980.

gbalint (approved):

PR #2137, opened 2022-03-04
Absolute move strategy: pins, % support

balazsbajorics:
**Problem:**
The absolute move strategy was an example implementation that could only update top left pins, did not know how to handle numbers with units and percentages.

**Fix:**
- A new `ADJUST_CSS_LENGTH_PROPERTY` command which can handle the following values: `5`, `'5px'`, `'5%'`. Units other than px are not (yet?) supported.
- `absoluteMoveStrategy` is now aware of `['top', 'bottom', 'left', 'right']` pins, not just top and left.
- Test coverage

**TODO for next PR**
- decide what to do with missing pins (automatically insert? ask the user?)
- decide what to do if the pin value is an expression (automatically override with number? ask the user? leave it intact?)


editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
```diff
@@ -38,27 +45,58 @@ export const absoluteMoveStrategy: CanvasStrategy = {
...
+            selectedElement,
+            canvasState.projectContents,
+            {},
+            canvasState.openFile,
+            null,
+            (_, e) => e,
+          )
+          const elementParentBounds =
+            MetadataUtils.findElementByElementPath(
+              sessionState.startingMetadata, // TODO should this be using the current metadata?
```
Rheeseyb, line 60: I think that would only make sense if this strategy could affect the parent bounds, _but_ this code is being called by the reparenting strategy :grimacing:

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+  return {
+    ...getEditorState(makeTestProjectCodeWithSnippet(codeSnippet)),
+    selectedViews: selectedViews,
+  }
+}
+
+function dragBy15Pixels(editorState: EditorState): EditorState {
+  const interactionSession: InteractionSession = {
+    ...createInteractionViaMouse(
+      null as any,
```
seanparsons, line 35: *scrunches face up*

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
```diff
@@ -38,27 +45,58 @@ export const absoluteMoveStrategy: CanvasStrategy = {
...
+          return mapDropNulls(
+            (pin) => {
+              const horizontal = isHorizontalPoint(
+                // TODO avoid using the loaded FramePoint enum
+                framePointForPinnedProp(pin),
+              )
+              const negative = pin === 'right' || pin === 'bottom'
+              const value = getLayoutProperty(pin, right(element.props), ['style'])
+              if (isRight(value) && value.value != null) {
+                // TODO what to do about missing properties?
```
Rheeseyb, line 77: In theory we should insert them based on the computed values in the metadata, but I question whether that's something that should happen in this strategy, or in a separate strategy that exists purely for the purpose of inserting a missing layout prop (I'm thinking along the lines of the "convert to absolute" strategy we have spoken about)

editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -91,12 +91,13 @@ export function createInteractionViaMouse(
   mouseDownPoint: CanvasPoint,
   modifiers: Modifiers,
   activeControl: CanvasControlType,
+  drag?: CanvasVector,
```
Rheeseyb, line 94: This shouldn't live in here. We have a function for updating an existing interaction, which we can use for the sake of the tests

editor/src/components/canvas/commands/adjust-css-length-command.ts
```diff
@@ -0,0 +1,241 @@
...
+      editorStatePatch: {},
+      pathMappings: pathMappings,
+      commandDescription: `Adjust Css Length Prop: ${EP.toUid(command.target)}/${PP.toString(
+        command.property,
+      )} not applied as the property is an expression we did not want to override.`,
+    }
+  }
+
+  const parsePxResult = parseCSSPx(simpleValueResult.value) // TODO make type contain px
+  const parsePercentResult = parseCSSPercent(simpleValueResult.value) // TODO make type contain %
```
seanparsons, line 107: You could put this one inside the else clause that checks `parsePxResult`, otherwise both always run.

editor/src/utils/modifiers.ts
```diff
@@ -5,6 +5,8 @@ export type Modifiers = {
   shift: boolean
 }
 
+export const emptyModifiers: Modifiers = { alt: false, cmd: false, ctrl: false, shift: false }
```
Rheeseyb, line 8: It's probably worth wrapping this in an `optionalDeepFreeze`

editor/src/core/shared/array-utils.ts
```diff
@@ -36,7 +36,7 @@ export function flatMapArray<T, U>(
 
 export function mapDropNulls<T, U>(
```
Rheeseyb, line 37: I really think we need to create a `filterMap` (or `mapFilter`), because that's one of the main uses of this function

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
balazsbajorics: I should clarify that this is a Design TODO

editor/src/core/shared/array-utils.ts
balazsbajorics: I actually like this API because the callback function can just return null to filter something out.

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
balazsbajorics: good shout, I'm gonna make a todo to figure out what do we want to do if the move affects the parent bounds

editor/src/core/shared/array-utils.ts
Rheeseyb: Right, which was the point of this function in the first place. I mean that since we use it heavily for mapping an array and filtering it at the same time, we should make a function (or pair of functions) more clearly designed for that.

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
balazsbajorics: todo made

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
balazsbajorics: added todo to doc

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
```diff
@@ -38,27 +45,58 @@ export const absoluteMoveStrategy: CanvasStrategy = {
...
-        adjustNumberProperty(
-          'permanent',
-          selectedElement,
-          stylePropPathMappingFn('top', ['style']),
-          drag.y,
-          true,
-        ),
+      return [
+        ...canvasState.selectedElements.flatMap((selectedElement) => {
+          const element: JSXElement | null = withUnderlyingTarget(
```
gbalint, line 50: This is just a code style remark, i think this is too much inlining here with the `return` and the spread out flatmap in it with a long inline function inside.
I would at least give the result of the flatmap a name, maybe even the mapping function could be named too.

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
```diff
@@ -38,27 +45,58 @@ export const absoluteMoveStrategy: CanvasStrategy = {
...
+                  true,
+                )
+              } else {
+                return null
+              }
+            },
+            ['top', 'bottom', 'left', 'right'] as const,
+          )
+        }),
         wildcardPatch('transient', {
```
gbalint, line 94: I know this is not new code, but this should have its own `setHighlightedViews` command

editor/src/components/canvas/commands/adjust-css-length-command.ts
balazsbajorics: fixed!

editor/src/utils/modifiers.ts
balazsbajorics: done

editor/src/components/canvas/commands/adjust-css-length-command.ts
```diff
@@ -0,0 +1,244 @@
...
+    property: property,
+    valuePx: valuePx,
+    parentDimensionPx: parentDimensionPx,
+    createIfNonExistant: createIfNonExistant,
+  }
+}
+
+export const runAdjustCssLengthProperty: CommandFunction<AdjustCssLengthProperty> = (
+  editorState: EditorState,
+  pathMappings: PathMappings,
```
gbalint, line 59: You already removed `pathMappings` in https://github.com/concrete-utopia/utopia/pull/2136
Maybe `master` is not merged into this branch yet?

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
balazsbajorics: I made a todo in paper

editor/src/components/canvas/commands/adjust-css-length-command.ts
balazsbajorics: fixed, thanks

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
balazsbajorics: I improved it with 7c4e5bc6

editor/src/components/canvas/canvas-strategies/interaction-state.ts
balazsbajorics: fixed in 32a18870

gbalint (approved):

Rheeseyb (approved):

PR #2141, opened 2022-03-07
Feature/canvas controls performance cssvars

enidemi:
**Problem:**
Solving performance problems with the new canvas controls. With the usage of css variables the new controls became very slow because the canvas offset css variable has to be outside of the canvas controls to be visible for them, but updating it causes Chrome to invalidate all child element styles that are using css variables (including css variables with a different name).

**Fix:**
Removing css variables for offset and scale. Offset is updated similarly how the new controls position themselves using the element ref directly. Scale is from editor state.

**Commit Details:**
- removed 3 css variables
- created a component for controls that automatically sets the offset
- set the offset on the container
- updated the 3 new canvas controls

### DIFF
--- a/editor/src/components/canvas/canvas-component-entry.tsx
+++ b/editor/src/components/canvas/canvas-component-entry.tsx
@@ -88,6 +88,7 @@ export const CanvasComponentEntry = React.memo((props: CanvasComponentEntryProps
         style={{
           position: 'absolute',
           transition: canvasProps.scrollAnimation ? 'transform 0.3s ease-in-out' : 'initial',
+          transform: 'translate3d(0px, 0px, 0px)',
         }}
       >
         <CanvasErrorBoundary
--- a/editor/src/components/canvas/canvas-component-entry.tsx
+++ b/editor/src/components/canvas/canvas-component-entry.tsx
@@ -25,7 +25,7 @@ import { DomWalkerProps, useDomWalker } from './dom-walker'
 import { ResolvingRemoteDependencyErrorName } from '../../core/es-modules/package-manager/package-manager'
 import { CanvasLoadingScreen } from './canvas-loading-screen'
 import { isHooksErrorMessage } from '../../utils/canvas-react-utils'
-import { useApplyCanvasOffset } from './controls/canvas-offset-wrapper'
+import { useApplyCanvasOffsetToStyle } from './controls/canvas-offset-wrapper'
 
 interface CanvasComponentEntryProps {}
 
@@ -76,7 +76,7 @@ export const CanvasComponentEntry = React.memo((props: CanvasComponentEntryProps
     clearRuntimeErrors()
   }, [clearRuntimeErrors])
 
-  const containerRef = useApplyCanvasOffset(canvasProps?.scale ?? null)
+  const containerRef = useApplyCanvasOffsetToStyle(canvasProps?.scale ?? null)
 
   if (canvasProps == null) {
     return <CanvasLoadingScreen />
@@ -87,7 +87,6 @@ export const CanvasComponentEntry = React.memo((props: CanvasComponentEntryProps
         ref={containerRef}
         style={{
           position: 'absolute',
-          zoom: canvasProps.scale >= 1 ? `${canvasProps.scale * 100}%` : 1,
           transition: canvasProps.scrollAnimation ? 'transform 0.3s ease-in-out' : 'initial',
         }}
       >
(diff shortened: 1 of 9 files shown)


editor/src/components/canvas/canvas-component-entry.tsx
```diff
@@ -75,18 +76,18 @@ export const CanvasComponentEntry = React.memo((props: CanvasComponentEntryProps
...
   if (canvasProps == null) {
     return <CanvasLoadingScreen />
   } else {
     return (
       <div
         id='canvas-container-outer'
+        ref={containerRef}
         style={{
           position: 'absolute',
           zoom: canvasProps.scale >= 1 ? `${canvasProps.scale * 100}%` : 1,
```
balazsbajorics, line 90: I think zoom and scale should be set at the same place, now one file is setting scale if scale < 1, and another file is setting zoom if scale > 1

editor/src/components/canvas/controls/canvas-offset-wrapper.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const elementRef = useApplyCanvasOffset(null)
+
+  return (
+    <div ref={elementRef} style={{ position: 'absolute' }}>
+      {props.children}
+    </div>
+  )
+})
+
+export function useApplyCanvasOffset(scale: number | null): React.RefObject<HTMLDivElement> {
```
balazsbajorics, line 15: I think for future debugging, the name of this hook should tell us that it overrides the style.transform prop of the ref

enidemi: fixed

editor/src/components/canvas/canvas-component-entry.tsx
enidemi: fixed

editor/src/components/canvas/controls/canvas-offset-wrapper.tsx
```diff
@@ -0,0 +1,43 @@
...
+        elementRef.current.style.setProperty(
+          'zoom',
+          scale != null && scale >= 1 ? `${scale * 100}%` : '1',
+        )
+      }
+    },
+    [elementRef, scale],
+  )
+
+  useSelectorWithCallback((store) => store.editor.canvas.roundedCanvasOffset, applyCanvasOffset)
```
seanparsons, line 35: Is this needed as well as the `useLayoutEffect` lower down? As it appears to trigger internally using `useEffect`.

balazsbajorics (approved):

editor/src/components/canvas/utopia-canvas-vars.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { CanvasScale, CanvasScrollOffset } from '../../utils/global-positions'
 import { useEditorState } from '../editor/store/store-hook'
 
-export const UtopiaCanvasVarStyleTag = React.memo(() => {
+export const UtopiaCanvasVars = React.memo(() => {
```
seanparsons, line 5: Almost feels like this should be a hook rather than a component.

editor/src/components/canvas/canvas-component-entry.tsx
```diff
@@ -75,19 +76,19 @@ export const CanvasComponentEntry = React.memo((props: CanvasComponentEntryProps
...
         id='canvas-container-outer'
+        ref={containerRef}
         style={{
           position: 'absolute',
-          zoom: canvasProps.scale >= 1 ? `${canvasProps.scale * 100}%` : 1,
-          transform:
-            (canvasProps.scale < 1 ? `scale(${canvasProps.scale})` : '') +
-            ` translate3d(var(--utopia-canvas-offset-x), var(--utopia-canvas-offset-y), 0)`,
           transition: canvasProps.scrollAnimation ? 'transform 0.3s ease-in-out' : 'initial',
+          transform: 'translate3d(0px, 0px, 0px)',
```
Rheeseyb, line 91: Does this `translate3d` still serve a purpose?

editor/src/components/canvas/utopia-canvas-vars.tsx
enidemi: this will be removed in the next PR

Rheeseyb (approved):

editor/src/components/canvas/controls/canvas-offset-wrapper.tsx
enidemi: `useSelectorWithCallback` will be triggered only with scroll changes, `useLayoutEffect` makes sure it's positioned at first render

PR #2143, opened 2022-03-08
Fix to skip extra dispatches when hovering empty canvas/none-selected

Seanmclem:
**Problem:**
maybeClearHighlightsOnHoverEnd was still dispatching many extra times when it already knew nothing was selected to highlight

**Fix:**
Check selectedViews length before dispatch

**Commit Details:**
- Did [other things]

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
@@ -97,6 +97,7 @@ export function useMaybeHighlightElement(): {
       selectionEnabled: pickSelectionEnabled(store.editor.canvas, store.editor.keysPressed),
       inserting: isInserting(store.editor),
       highlightedViews: store.editor.highlightedViews,
+      selectedViews: store.editor.selectedViews,
     }
   })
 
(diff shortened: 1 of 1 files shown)


balazsbajorics (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -122,8 +123,9 @@ export function useMaybeHighlightElement(): {
   )
 
   const maybeClearHighlightsOnHoverEnd = React.useCallback((): void => {
-    const { dispatch, dragging, resizing, selectionEnabled } = stateRef.current
-    if (selectionEnabled && !dragging && !resizing) {
+    const { dispatch, dragging, resizing, selectionEnabled, selectedViews } = stateRef.current
+
+    if (selectionEnabled && !dragging && !resizing && selectedViews.length > 0) {
```
enidemi, line 128: should this also check the length of the highlightedViews?

Rheeseyb: Yep, it looks like this should be checking the length of `highlightedViews`, and not `selectedViews`

Seanmclem: yeah, I think both would be better

Rheeseyb: To be clear here, this shouldn't be using both. It should only be using `highlightedViews`

Rheeseyb (approved):

balazsbajorics (approved):

enidemi (approved):

seanparsons (approved):

balazsbajorics (approved):

PR #2144, opened 2022-03-08
Absolute move snapping

balazsbajorics:
**Problem:**
The absolute move strategy did not support snapping.

**Fix:**
This PR adds the existing snapping logic to the absolute move strategy. I extracted the old move snapping code and created a single function that runs only the snapping and returns the snap result. I called this function `runLegacySnapping`, if you find it too dumb I can rename it :D 
This PR also introduces snap guideline controls made by @enidemi. They were designed to avoid react rerenders as much as possible, it always renders a fixed number of guidelines and the guidelines are set to `display: none` instead of being removed from the DOM and then re-added to the DOM as needed. The idea is that they should be able to just snap into existence when they are needed, with minimal overhead, then disappear maybe a single frame later, if the user is moving their mouse fast.

**Commit Details:**
- Added `snapDrag` to `absoluteMoveStrategy`, which uses `runLegacySnapping` for now. If we need more snap rules, we can add them to `snapDrag`
- `absoluteMoveStrategy` patches `EditorState.canvas.controls.snappingGuidelines` to store the guidelines of the active snap
- `GuidelineControls` reads from `EditorState.canvas.controls.snappingGuidelines` and renders the visual guidelines on the screen


editor/src/components/canvas/controls/guideline-controls.tsx
```diff
@@ -0,0 +1,136 @@
...
+        onChangeCallback(result)
+      }
+    },
+    [onChangeCallback],
+  )
+
+  const guidelineCallbackRef = React.useRef(guidelineCallback)
+  guidelineCallbackRef.current = guidelineCallback
+
+  const guidelineRef = useRefEditorState((store) => store.editor.canvas.controls.snappingGuidelines)
```
seanparsons, line 86: Should this selector be a constant too?

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
```diff
@@ -78,6 +94,13 @@ export const absoluteMoveStrategy: CanvasStrategy = {
           highlightedViews: {
             $set: [],
           },
+          canvas: {
+            controls: {
+              snappingGuidelines: {
+                $set: guidelinesWithSnappingVector, // these will be used by the guideline controls
```
Rheeseyb, line 100: Can you create a command for this rather than resorting to the wildcard patch please

editor/src/components/canvas/controls/select-mode/move-utils.ts
```diff
@@ -132,7 +132,7 @@ export function dragComponent(
               roundedDragDelta,
             ),
             scale,
-          )
+          ).delta
```
Rheeseyb, line 135: Is this a bug on production right now? If not, it looks like this is going to affect the existing behaviour

balazsbajorics: I changed the function's signature

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
```diff
@@ -117,3 +140,40 @@ function createMoveCommandsForElement(
...
+  canvasScale: number,
+): {
+  snappedDragVector: CanvasPoint
+  guidelinesWithSnappingVector: Array<GuidelineWithSnappingVector>
+} {
+  const multiselectBounds = getMultiselectBounds(jsxMetadata, selectedElements)
+
+  // This is the entry point to extend the list of snapping strategies, if we want to add more
+
+  const { snappedDragVector, guidelinesWithSnappingVector } = runLegacySnapping(
```
gbalint, line 158: There is no reason to destructure if you just recreate and return the same object again in line 167

editor/src/components/canvas/controls/select-mode/move-utils.ts
Rheeseyb: Ah, I missed that :+1:

Rheeseyb (approved):

PR #2147, opened 2022-03-09
Apply accumulated commands when there is no strategy

gbalint:
When there is no applicable strategy, but we are in an active interaction with accumulated commands, we need to apply the accumulated commands to the unpatched editor state (even though there are no new commands to apply, because without an active strategy we don't generate new commands).

### DIFF
--- a/editor/src/components/editor/store/dispatch-strategies.tsx
+++ b/editor/src/components/editor/store/dispatch-strategies.tsx
@@ -64,32 +64,26 @@ export function interactionFinished(
       sortedApplicableStrategies: sortedApplicableStrategies,
     }
 
-    // If there is a current strategy, produce the commands from it.
-    if (strategy == null) {
-      return {
-        unpatchedEditorState: newEditorState,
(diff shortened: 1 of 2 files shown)


editor/src/components/editor/store/dispatch-strategies.tsx
```diff
@@ -391,22 +385,25 @@ function handleUpdate(
...
     const commandResult = foldAndApplyCommands(
       newEditorState,
       storedEditorState,
       [...strategyState.accumulatedCommands.flatMap((c) => c.commands), ...commands],
       'transient',
     )
     const newStrategyState: StrategyState = {
-      currentStrategy: strategy.strategy.id,
-      currentStrategyFitness: strategy.fitness,
+      currentStrategy: strategy != null ? strategy.strategy.id : null,
```
Rheeseyb, line 405: This could be replaced with `strategy?.strategy.id ?? null`

Rheeseyb (approved):

balazsbajorics (approved):

PR #2150, opened 2022-03-09
Keyboard absolute move strategy

gbalint:
This PR introduces the first keyboard strategy.

Keyboard strategies are really different from drag strategies, mostly because they manage multiple keypresses as single interactions.
In this PR I handle a series of keypresses in the following way:
- the first keydown creates the interaction and starts a timer to clear the interaction later
- the subsequent keydowns update the interaction and reset the timer
- when the timer is off the interaction is cleared
- in dispatch we always accumulate the commands coming from the strategy so their result is not lost

Optionally you can enable the 'Keyboard up clears interaction' and then there is a much simple workflow: every keyboard up clears the interaction, so every separate keypress is a separate undo step. Multiple steps achieved by a long keypress stay as one undo step though (because there is no key up event between the steps).

I introduced a basic (prototype quality) keyboard absolute move strategy to demonstrate how this can work.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/interaction-state.ts
+++ b/editor/src/components/canvas/canvas-strategies/interaction-state.ts
@@ -280,7 +280,7 @@ export function interactionSessionHardReset(
   }
 }
 
-export const KEYBOARD_INTERACTION_TIMEOUT = 3000
+export const KeyboardInteractionTimeout = 3000
 
 export function hasDragModifiersChanged(
   prevInteractionData: InputData | null,
--- a/editor/src/components/canvas/canvas-strategies/interaction-state.ts
+++ b/editor/src/components/canvas/canvas-strategies/interaction-state.ts
@@ -26,7 +26,7 @@ export interface DragInteractionData {
 }
 
 interface KeyboardInteractionData {
-  type: 'KEYBOARD'
+  type: 'KEYBOARD_ARROW'
   keysPressed: Array<KeyCharacter>
   // keysPressed also includes modifiers, but we want the separate modifiers array since they are captured and mapped to a specific
   // set via modifiersForEvent in keyboard.ts
@@ -152,7 +152,7 @@ export function createInteractionViaKeyboard(
 ): InteractionSessionWithoutMetadata {
   return {
     interactionData: {
-      type: 'KEYBOARD',
+      type: 'KEYBOARD_ARROW',
       keysPressed: keysPressed,
       modifiers: modifiers,
     },
@@ -172,7 +172,7 @@ export function updateInteractionViaKeyboard(
   sourceOfUpdate: CanvasControlType,
 ): InteractionSessionWithoutMetadata {
   switch (currentState.interactionData.type) {
-    case 'KEYBOARD': {
+    case 'KEYBOARD_ARROW': {
       const withRemovedKeys = currentState.interactionData.keysPressed.filter(
         (k) => !keysReleased.includes(k),
       )
@@ -180,7 +180,7 @@ export function updateInteractionViaKeyboard(
 
       return {
         interactionData: {
-          type: 'KEYBOARD',
+          type: 'KEYBOARD_ARROW',
           keysPressed: newKeysPressed,
           modifiers: modifiers,
         },
@@ -228,7 +228,7 @@ export function strategySwitchInteractionDataReset(interactionData: InputData): 
           prevDrag: null,
         }
       }
-    case 'KEYBOARD':
+    case 'KEYBOARD_ARROW':
       return interactionData
     default:
       const _exhaustiveCheck: never = interactionData
@@ -253,7 +253,7 @@ export function interactionDataHardReset(interactionData: InputData): InputData 
           ),
         }
       }
-    case 'KEYBOARD':
+    case 'KEYBOARD_ARROW':
       return interactionData
     default:
       const _exhaustiveCheck: never = interactionData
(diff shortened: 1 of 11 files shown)


editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -26,7 +26,7 @@ export interface DragInteractionData {
 }
 
 interface KeyboardInteractionData {
-  type: 'KEYBOARD'
+  type: 'KEYBOARD_ARROW'
```
seanparsons, line 29: Why's this becoming `KEYBOARD_ARROW`?

editor/src/utils/keyboard.ts
```diff
@@ -254,6 +254,13 @@ export const Keyboard = {
       return char === keyChar
     })
   },
+  keyIsArrow: function (keyChar: KeyCharacter): boolean {
+    return ['left', 'right', 'up', 'down'].some((char) => char === keyChar)
```
seanparsons, line 258: Surely a switch statement makes more sense here?

gbalint: I just borrowed the pattern from the function above (`keyIsModifier`)
But surely I can make it a switch case.

editor/src/components/canvas/canvas-strategies/interaction-state.ts
gbalint: The idea was that because we only create this interaction for arrow keys, this is not a general keyboard interaction data, but only for moving with the arrows.
This is more relevant in my other PR, where we store the necessary state for moving elements there.
https://github.com/concrete-utopia/utopia/pull/2151

editor/src/utils/keyboard.ts
gbalint: Fixed in 5ee2ed1f5201d1d7179d95afa51a3a95ecf4dea1

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/keyboard-absolute-move-strategy.ts
```diff
@@ -0,0 +1,101 @@
...
+import { stylePropPathMappingFn } from '../../inspector/common/property-path-hooks'
+import { adjustNumberProperty } from '../commands/adjust-number-command'
+import { wildcardPatch } from '../commands/wildcard-patch-command'
+import { Keyboard } from '../../../utils/keyboard'
+import { CanvasStrategy } from './canvas-strategy-types'
+
+export const keyboardAbsoluteMoveStrategy: CanvasStrategy = {
+  id: 'KEYBOARD_ABSOLUTE_MOVE',
+  name: 'Keyboard absolute Move',
+  isApplicable: (canvasState, _interactionState, metadata) => {
```
seanparsons, line 11: Are `isApplicable` and `fitness` the same as the mouse based ones for the current absolute move strategy, with the exception of checking the key that is pressed?

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/keyboard-absolute-move-strategy.ts
gbalint: yes, they are basically the same

enidemi (approved):

editor/src/components/editor/store/dispatch-strategies.spec.tsx
```diff
@@ -655,6 +687,97 @@ describe('interactionUpdate with stacked strategy change', () => {
...
 })
 
+describe('interactionUpdate with accumulating keypresses', () => {
+  it('steps an interaction session correctly', () => {
+    let interactionSession = createInteractionViaMouse(
+      canvasPoint({ x: 100, y: 200 }),
+      { alt: false, shift: false, ctrl: false, cmd: false },
+      { type: 'BOUNDING_AREA', target: EP.elementPath([['aaa']]) },
+    )
+    if (interactionSession.interactionData.type === 'KEYBOARD') {
```
balazsbajorics, line 697: you create the interactionSession with `createInteractionViaMouse`, so it looks like this IF will never be true

editor/src/components/editor/store/dispatch-strategies.spec.tsx
```diff
@@ -655,6 +687,97 @@ describe('interactionUpdate with stacked strategy change', () => {
...
+        "currentStrategy": null,
+        "currentStrategyCommands": Array [],
+        "currentStrategyFitness": 0,
+        "sortedApplicableStrategies": Array [],
+        "startingMetadata": Object {},
+      }
+    `)
+    expect(actualResult.patchedEditorState.canvas.scale).toEqual(100)
+    expect(actualResult.unpatchedEditorState.canvas.scale).toEqual(1)
+    expect(actualResult.patchedEditorState.canvas.interactionSession?.interactionData)
```
balazsbajorics, line 755: is this a relevant thing for us to check here?

editor/src/components/editor/store/dispatch-strategies.tsx
```diff
@@ -164,6 +165,7 @@ export function interactionUpdate(
   strategies: Array<CanvasStrategy>,
   storedState: EditorStoreFull,
   result: InnerDispatchResult,
+  isInteractionAction: boolean,
```
balazsbajorics, line 168: this boolean flag is pretty hard to read in the tests. one thing we'be been doing in some places is to use a string literal union instead of a boolean, so reading the callsite explains what the flag does.

editor/src/components/editor/store/dispatch-strategies.spec.tsx
```diff
@@ -260,7 +261,7 @@ describe('interactionStart', () => {
   })
 })
 
-describe('interactionUpdate', () => {
+describe('interactionUpdatex', () => {
```
balazsbajorics, line 264: is this an accidental rename?

balazsbajorics (approved):

PR #2151, opened 2022-03-10
Keyboard absolute move strategy Option 2

gbalint:
This PR is an alternative solution for https://github.com/concrete-utopia/utopia/pull/2150

Instead of accumulating the commands generated by the key presses, I accumulate the keypresses inside the interaction state, and always recreate the actions in the strategy for all of them.


gbalint:
We chose the solution implemented in https://github.com/concrete-utopia/utopia/pull/2151, closing this

PR #2152, opened 2022-03-10
Measure Absolute Move Strategy Performance.

seanparsons:
**Problem:**
Currently there are no performance tests checking any of the canvas strategies work.

**Fix:**
Added a performance test which enables canvas strategies and then triggers the absolute move strategy by dragging an absolutely positioned element.

**Notes:**
- This does make one destructive change to the project, which is to delete the siblings of the target element in the project. This was necessary as they ended up on top of the target and ended up receiving the mouse events.

**Commit Details:**
- Added additional menubar item to trigger the test.
- Implemented `useTriggerAbsoluteMovePerformanceTest` to slightly
  manipulate the test project and then run the tests against it.
- Added `testAbsoluteMovePerformance` for triggering the performance
  test via the menubar.

### DIFF
--- a/editor/src/components/menubar/menubar.tsx
+++ b/editor/src/components/menubar/menubar.tsx
@@ -11,6 +11,7 @@ import {
   useTriggerBaselinePerformanceTest,
   useTriggerRegularHighlightPerformanceTest,
   useTriggerAllElementsHighlightPerformanceTest,
+  useTriggerAbsoluteMovePerformanceTest,
 } from '../../core/model/performance-scripts'
 import { useReParseOpenProjectFile } from '../../core/model/project-file-helper-hooks'
 import { shareURLForProject } from '../../core/shared/utils'
@@ -192,6 +193,7 @@ export const Menubar = React.memo(() => {
   const onTriggerRegularHighlightTest = useTriggerRegularHighlightPerformanceTest()
   const onTriggerAllElementsHighlightTest = useTriggerAllElementsHighlightPerformanceTest()
   const onTriggerSelectionTest = useTriggerSelectionPerformanceTest()
+  const onTriggerAbsoluteMoveTest = useTriggerAbsoluteMovePerformanceTest()
   const onTriggerBaselineTest = useTriggerBaselinePerformanceTest()
 
   const previewURL =
@@ -367,6 +369,9 @@ export const Menubar = React.memo(() => {
           </Tile>
           <Tile style={{ marginTop: 12, marginBottom: 12 }} size='large'>
             <a onClick={onTriggerSelectionTest}>P E</a>
+          </Tile>
+          <Tile style={{ marginTop: 12, marginBottom: 12 }} size='large'>
+            <a onClick={onTriggerAbsoluteMoveTest}>PAM</a>
           </Tile>
           <Tile style={{ marginTop: 12, marginBottom: 12 }} size='large'>
             <a onClick={onTriggerBaselineTest}>B L</a>
--- a/editor/src/core/model/performance-scripts.tsx
+++ b/editor/src/core/model/performance-scripts.tsx
@@ -47,7 +47,7 @@ export function wait(timeout: number): Promise<void> {
   })
 }
 
-const NumberOfIterations = 100
+const NumberOfIterations = 5
 
 export function useTriggerScrollPerformanceTest(): () => void {
   const dispatch = useEditorState(
@@ -457,17 +457,19 @@ export function useTriggerAbsoluteMovePerformanceTest(): () => void {
           buttons: 1,
         }),
       )
-      controlsContainerElement.dispatchEvent(
-        new MouseEvent('mousemove', {
-          detail: 1,
-          bubbles: true,
-          cancelable: true,
-          metaKey: false,
-          clientX: targetBounds.left + 35,
-          clientY: targetBounds.top + 45,
-          buttons: 1,
-        }),
-      )
+      for (let moveCount = 1; moveCount <= 10; moveCount++) {
+        controlsContainerElement.dispatchEvent(
+          new MouseEvent('mousemove', {
+            detail: 1,
+            bubbles: true,
+            cancelable: true,
+            metaKey: false,
+            clientX: targetBounds.left + (5 + moveCount * 3),
+            clientY: targetBounds.top + (5 + moveCount * 4),
+            buttons: 1,
+          }),
+        )
+      }
       controlsContainerElement.dispatchEvent(
         new MouseEvent('mouseup', {
           detail: 1,
--- a/editor/src/core/model/performance-scripts.tsx
+++ b/editor/src/core/model/performance-scripts.tsx
@@ -443,7 +443,7 @@ export function useTriggerAbsoluteMovePerformanceTest(): () => void {
         ],
         'everyone',
       ).entireUpdateFinished
-      performance.mark(`absolute_move_step_${framesPassed}`)
+      performance.mark(`absolute_move_interaction_step_${framesPassed}`)
 
       // Move it down and to the right.
       controlsContainerElement.dispatchEvent(
@@ -457,6 +457,9 @@ export function useTriggerAbsoluteMovePerformanceTest(): () => void {
           buttons: 1,
         }),
       )
+
+      // Mouse move and performance marks for that.
+      performance.mark(`absolute_move_move_step_${framesPassed}`)
       for (let moveCount = 1; moveCount <= 10; moveCount++) {
         controlsContainerElement.dispatchEvent(
           new MouseEvent('mousemove', {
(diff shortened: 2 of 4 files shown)


editor/src/core/model/performance-scripts.tsx
```diff
@@ -331,6 +338,174 @@ export function useTriggerSelectionPerformanceTest(): () => void {
...
+          detail: 1,
+          bubbles: true,
+          cancelable: true,
+          metaKey: false,
+          clientX: targetBounds.left + 35,
+          clientY: targetBounds.top + 45,
+          buttons: 1,
+        }),
+      )
+      controlsContainerElement.dispatchEvent(
```
balazsbajorics, line 471: This current test will skew extremely towards mouse up performance (which right now is _very_ expensive because of the inspector remount), and thus it will actually measure the strategy lifecycle more than the move speed itself. I'd be tempted to rename this to Something Something Strategy Lifecycle.

And for the Absolute Move Strategy Performance itself I'd add a second test in which only the mouse move is inside the loop, so we have a number for continuous scrolling performance.

seanparsons: Stepped mouse move added in 3bbedfa25.

seanparsons: 9060f4687 captures both the entire interaction time and just the mouse move segments of the absolute move interaction.

editor/src/components/menubar/menubar.tsx
```diff
@@ -368,6 +370,9 @@ export const Menubar = React.memo(() => {
           <Tile style={{ marginTop: 12, marginBottom: 12 }} size='large'>
             <a onClick={onTriggerSelectionTest}>P E</a>
           </Tile>
+          <Tile style={{ marginTop: 12, marginBottom: 12 }} size='large'>
+            <a onClick={onTriggerAbsoluteMoveTest}>PAM</a>
```
Rheeseyb, line 374: pan_or_pam

seanparsons: It's PAM.

editor/src/core/model/performance-scripts.tsx
```diff
@@ -331,6 +339,197 @@ export function useTriggerSelectionPerformanceTest(): () => void {
...
+  const selectedViews = useRefEditorState(
+    React.useCallback((store) => store.editor.selectedViews, []),
+  )
+  const trigger = React.useCallback(async () => {
+    const initialTargetPath = [...allPaths.current].sort(
+      (a, b) => EP.toString(b).length - EP.toString(a).length,
+    )[0]
+    // This is very particularly tied to the test project, we _really_ need to pick the
+    // right element because our changes can cause other elements to end up on top of the
+    // target we want.
```
Rheeseyb, line 360: Is this comment still accurate (specifically the part about it being tied to the test project itself) now that the test is no longer picking a very specific element?

seanparsons: Yes, the steps of taking the longest path, going up two levels and then picking the last (in order) grandchild of that grandparent are just a programmatic way of getting the same static path that was originally in here.

puppeteer-tests/src/performance-test.ts
```diff
@@ -370,7 +375,43 @@ export const testSelectionPerformance = async function (
   await page.tracing.stop()
   let traceData = fs.readFileSync('trace.json').toString()
   const traceJson = JSON.parse(traceData)
-  return getFrameData(traceJson, 'select_step_', 'Selection')
+  return getFrameData(traceJson, 'select_frame_', 'Selection')
```
Rheeseyb, line 378: > This change looks incorrect and presumably unintended?

Likewise here

puppeteer-tests/src/utils.ts
```diff
@@ -18,8 +18,8 @@ export const setupBrowser = async (
     executablePath: process.env.BROWSER,
   })
   const page = await browser.newPage()
-  await page.setDefaultNavigationTimeout(120000)
-  await page.setDefaultTimeout(defaultTimeout)
+  page.setDefaultNavigationTimeout(120000)
+  page.setDefaultTimeout(defaultTimeout)
```
Rheeseyb, line 22: Why are we no longer `await`ing these?

puppeteer-tests/src/performance-test.ts
```diff
@@ -352,7 +357,7 @@ export const testHighlightAllElementsPerformance = async function (
   await page.tracing.stop()
   let traceData = fs.readFileSync('trace.json').toString()
   const traceJson = JSON.parse(traceData)
-  return getFrameData(traceJson, 'highlight_all-elements_step_', 'Highlight All Elements')
+  return getFrameData(traceJson, 'highlight_all-elements_frame_', 'Highlight All Elements')
```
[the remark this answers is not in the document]
seanparsons: When I was looking at the calls to mark and measure, I noticed an inconsistency between the various tests and corrected them to make them look for what _appears_ to be the right entry.

puppeteer-tests/src/utils.ts
seanparsons: They don't return a promise, I had a warning saying it wasn't doing anything.

puppeteer-tests/src/performance-test.ts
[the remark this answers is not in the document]
Rheeseyb: Hmmm can you pull this into a separate PR then please? I'm looking at the other tests and they're all using `_step_` in the `getFrameData` calls (e.g. `highlight_regular_step_`)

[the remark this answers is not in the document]
seanparsons: Fixed in 4b4d5ce006d373f0e3f0b3eb6cee4c0b7a8bc87f.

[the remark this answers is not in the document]
Rheeseyb: Thank you

editor/src/core/model/performance-scripts.tsx
```diff
@@ -331,6 +339,197 @@ export function useTriggerSelectionPerformanceTest(): () => void {
...
+    // Delete the other children that just get in the way.
+    const parentPath = EP.parentPath(targetPath)
+    const siblingPaths = allPaths.current.filter(
+      (path) => EP.isChildOf(path, parentPath) && !EP.pathsEqual(path, targetPath),
+    )
+    await dispatch(
+      siblingPaths.map((path) => deleteView(path)),
+      'everyone',
+    ).entireUpdateFinished
+    // Focus the target so that we can edit the child div inside it.
```
balazsbajorics, line 383: what if I don't have a focusable component in my test project? will this cause a problem?

seanparsons: Yes.

Rheeseyb: Unfortunately right now this test is very specific to the current performance test project, which does seem to imply that either there's an issue with applying the strategy directly on any element (even after giving it the required style props), or that there's an issue with firing the required mouse events to invoke the strategy directly on a specific element. I suggest we look to address that separately in some follow up work, perhaps after we have had time to evaluate this testing strategy (pun partially intended) on a number of strategies.

seanparsons: I think the ideal case would be to pair a testing strategy with a project definition. Especially if it was to be done in such a way that didn't involve the saving and the shenanigans it brings in as well.

seanparsons: All of the existing tests are brittle like this to some extent, but since this test is for absolutely positioned elements and our current test project of choice is entirely done with flex this one is just that bit more troublesome.

Rheeseyb: > I think the ideal case would be to pair a performance test with a project definition. Especially if it was to be done in such a way that didn't involve the saving and the shenanigans it brings in as well.

Hmmm I really like this idea, but that definitely makes testing with puppeteer a bit awkward (because we'd either have to run multiple separate tests like that, or trigger page navigations after each test). I don't think that's entirely implausible though.

Alternatively, we could investigate a way to designate the target elements of each test in a given project (maybe via a `data-label` prop on the element for a specific test, or by enabling staging to retain UIDs via a feature flag)?

seanparsons: I say that as well given that there's already interest in testing with different projects, which as-is would be very clunky as we're having to make sure that projects exist in each environment its being run in.

balazsbajorics: we could store the test projects in the editor code, and clicking the test button could start by loading the project contents and runnin the test. that would ensure we really are testing the same thing everywhere

balazsbajorics: okay I don't think this is a blocker issue, but we should have a follow up TODO about making the tests more portable 👍

balazsbajorics (approved):

PR #2154, opened 2022-03-11
Feature/absolute resize strategy

enidemi:
Introducing a new strategy for absolute resize, it is a simple resize strategy where you can grab the edge or corner of an absolute element and the correct top/left/width/height/bottom/right properties will be updated.
Later in the next PR it will be extended to have more complex features (multiselect, guidelines and snap points).

Also updated the new Resize Controls to dispatch mouse interaction session on mousedown instead of dragstate and added the Resize Controls to the strategy `controlsToRender`. For that localSelectedViews is changed to use the editor state hook and as `isApplicable` already checks if the elements are absolutely positioned that is also removed. Without strategies feature flag you see the old resize control.
As a result in `new-canvas-controls` I switched the outline to be below the strategy controls. I believe it is correct for all strategy controls in general to be placed over the selection outline.

**Commit Details:**
- new absolute resize strategy
- changed the resize controls to be included in the strategy
- tests


editor/src/components/canvas/canvas-strategies/absolute-resize-strategy.tsx
```diff
@@ -0,0 +1,148 @@
...
+      framePointForPinnedProp(pin),
+    )
+    const negative =
+      pin === 'right' ||
+      pin === 'bottom' ||
+      (pin === 'width' && edgePosition.x === 0) ||
+      (pin === 'height' && edgePosition.y === 0)
+    const value = getLayoutProperty(pin, right(element.props), ['style'])
+    if (isRight(value) && value.value != null) {
+      // TODO what to do about missing properties?
```
Rheeseyb, line 135: We have a similar problem right now in the absolute move strategy. I have a suspicion that maybe this falls into the same problem category as the escape hatch strategy that we've discussed, and so should be solved as part of that.

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/absolute-resize-strategy.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { AbsoluteResizeControl } from '../controls/select-mode/absolute-resize-control'
+import { CanvasStrategy } from './canvas-strategy-types'
+
+type AbsolutePin = 'left' | 'top' | 'right' | 'bottom' | 'width' | 'height'
+
+export const absoluteResizeStrategy: CanvasStrategy = {
+  id: 'ABSOLUTE_RESIZE',
+  name: 'Absolute Resize',
+  isApplicable: (canvasState, _interactionState, metadata) => {
+    if (canvasState.selectedElements.length > 0) {
```
seanparsons, line 27: I wonder if this logic should be shared between all the absolutely positioned strategies.

editor/src/components/canvas/canvas-strategies/absolute-resize-strategy.tsx
```diff
@@ -0,0 +1,148 @@
...
+      interactionState.interactionData.type === 'DRAG' &&
+      interactionState.interactionData.drag != null &&
+      interactionState.activeControl.type === 'RESIZE_HANDLE'
+    ) {
+      const drag = interactionState.interactionData.drag
+      const edgePosition = interactionState.activeControl.edgePosition
+
+      const commandsForSelectedElements = canvasState.selectedElements.flatMap(
+        (selectedElement) => {
+          const element: JSXElement | null = withUnderlyingTarget(
```
seanparsons, line 62: I _think_ @balazsbajorics made a utility method that is exactly this...

seanparsons (approved):

PR #2155, opened 2022-03-11
Correct the calls to getFrameData.

seanparsons:
**Problem:**
The performance marks being used by the performance tests to analyse the time taken for the tests are the wrong ones.

**Fix:**
Instead of taking a measurement of the timestamp of step n + 1 and subtracting the timestamp of step n. The timestamp of the "end" performance mark has the timestamp of the "start" performance mark subtracted from it.

**Commit Details:**
- Added in a bonus deselection performance result which was being measured
  but ignored.
- Added some utility functions to wrap `performance.mark` and
  `performance.measure` to get consistent calls to them.
- Added `--files` to the `ts-node` invocation in the puppeter project
  so that the missing types for JSONStream are handled correctly.
- Added JSONStream to the puppeter-tests dependencies, along with a
  missing types module declaration for it.
- Replaced the original trace file loading with a streaming approach
  so that the entire file doesn't need to be loaded in full.
- Removed the categories filter so that any entries can be loaded in
  subsequently.
- Rewrote `getFrameData` to just use the start and end marks to bound
  the time a step took.


Rheeseyb (approved):
Thanks!

PR #2156, opened 2022-03-11
Improved keyboard absolute move strategy

gbalint:
I improved the keyboard absolute move strategy:

- extracted the much better code from absolute move strategy into a shared file, and use that for the keyboard strategy. This makes it work with all different kinds off css props.
- Added shift key modifier (bigger step)
- Added unit tests (TODO: need to add more keys and shift)

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
+++ b/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
@@ -25,7 +25,7 @@ import { wildcardPatch } from '../commands/wildcard-patch-command'
 import { runLegacySnapping } from '../controls/guideline-helpers'
 import { ConstrainedDragAxis, Guideline, GuidelineWithSnappingVector } from '../guideline'
 import { CanvasStrategy } from './canvas-strategy-types'
-import { getMoveCommandsForSelectedElement } from './shared-move-strategy-helpers'
+import { getAbsoluteMoveCommandsForSelectedElement } from './shared-absolute-move-strategy-helpers'
 
 export const absoluteMoveStrategy: CanvasStrategy = {
   id: 'ABSOLUTE_MOVE',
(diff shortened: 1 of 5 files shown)


seanparsons (approved):

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
```diff
@@ -25,6 +25,7 @@ import { wildcardPatch } from '../commands/wildcard-patch-command'
 import { runLegacySnapping } from '../controls/guideline-helpers'
 import { ConstrainedDragAxis, Guideline, GuidelineWithSnappingVector } from '../guideline'
 import { CanvasStrategy } from './canvas-strategy-types'
+import { getMoveCommandsForSelectedElement } from './shared-move-strategy-helpers'
```
Rheeseyb, line 28: Either the file name or the imported function should be updated to reflect that this is for absolute positioning

editor/src/components/canvas/canvas-strategies/keyboard-absolute-move-strategy.ts
```diff
@@ -46,56 +46,49 @@ export const keyboardAbsoluteMoveStrategy: CanvasStrategy = {
     return 0
   },
   apply: (canvasState, interactionState, sessionState) => {
-    // TODO: absolutely minimal implementation
     if (interactionState.interactionData.type === 'KEYBOARD') {
       const key = interactionState.interactionData.keysPressed[0]
```
Rheeseyb, line 50: I believe this should be mapping over the `keysPressed`, so that it can handle e.g. up + left being held down

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
balazsbajorics: I renamed the file

balazsbajorics (approved):

Rheeseyb (approved):

PR #2163, opened 2022-03-17
Experiment: immer patch merge

gbalint:
### Summary
Because we collect the commands of the individual keypresses in StrategyState.accumulatedCommands, if you do a long press, this is gonna be an evergrowing list of commands to apply, and on every new keydown we apply all the commands here.
Overall this means this is a quadratic algorithm, and it becomes unusable in a couple of seconds.

### Solution idea:
Do not apply all the commands in accumulatedCommands and their patches in every dispatch, but let's just store a merged patch and modify it on every dispatch so it contains the effect of the current command too.

### Main problem:
I could not find a readymade solution to merge immutability-helper patches, and it doesn't seem to be a trivial problem. If we only support $set commands, then it is not that complex, but the interaction between all the different commands is quite complex.
*Note:* the naive solution to just merge the patch objects is not corrent, e.g. when you have a $set command to a part of an object and another set for a subkey of the same part.

#### Use immer

Immer has a really nice feature that it can produce patches for a given change. If you apply multiple changes it will still produce a compact patch.
So if we would modify the commands to return immer patches, we could just apply all the patches from all the commands and get back a compact patch list containing all the effects. We could just store this compact patch list in StrategyState, and next time apply that, apply all the commands too, and store the overall merged patch in StrategyState, etc. So basically immer knows what we need.
Downsides:
- we need to rewrite all commands to generated immer patches (or recipe functions, and we can produce the patches themselves using immer)
- immer has a performance overhead, but for our editorstate it doesn't seem to be more than an extra 0.2-0.3 ms

### Notes about the draft
- only the keyboard absolute move is implemented and even that is buggy

### DIFF
--- a/editor/src/components/canvas/commands/adjust-number-command.ts
+++ b/editor/src/components/canvas/commands/adjust-number-command.ts
@@ -267,20 +267,11 @@ export function applyValuesAtPath(
           value: success.imports,
         },
       ]
-      const [_, patcccse] = produceWithPatches(editorState, (draft) => {
-        ;((((draft.projectContents['src'] as ProjectContentDirectory).children[
-          'app.js'
-        ] as ProjectContentFile).content as TextFile).fileContents
-          .parsed as ParseSuccess).topLevelElements = success.topLevelElements
-      })
-
-      console.log(immerPatches, patcccse)
 
       return immerPatches
     },
   )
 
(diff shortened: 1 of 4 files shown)


gbalint:
I'm closing this, because the other option won: https://github.com/concrete-utopia/utopia/pull/2165

PR #2164, opened 2022-03-18
keydown on modifier keys updates the interaction

balazsbajorics:
**Problem:**
Releasing a modifier key would correctly re-run the strategies code with the new set of modifier keys, as expected. However, pressing down a modifier key would not update the set of modifier keys, it would only happen on first mouse move. 

**Fix:**
When pressing down a modifier key, and if a canvas interaction session exists, update the list of modifier keys on the canvas interaction data, which will also trigger the re-execution of the strategies.

**Commit Details:**
- add `updateInteractionViaKeyboard` to `editor-component@onWindowKeyDown` when `Keyboard.keyIsModifier` and `existingInteractionSession != null`

### DIFF
--- a/editor/src/components/editor/editor-component.tsx
+++ b/editor/src/components/editor/editor-component.tsx
@@ -144,7 +144,22 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
 
         // TODO: maybe we should not whitelist keys, just check if Keyboard.keyIsModifer(key) is false
         const existingInteractionSession = editorStoreRef.current.editor.canvas.interactionSession
-        if (Keyboard.keyIsInteraction(key)) {
+        if (Keyboard.keyIsModifier(key) && existingInteractionSession != null) {
+          editorStoreRef.current.dispatch(
+            [
+              CanvasActions.createInteractionSession(
+                updateInteractionViaKeyboard(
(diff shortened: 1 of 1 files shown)


enidemi (approved):

editor/src/components/editor/editor-component.tsx
```diff
@@ -143,8 +143,8 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
         const key = Keyboard.keyCharacterForCode(event.keyCode)
 
         // TODO: maybe we should not whitelist keys, just check if Keyboard.keyIsModifer(key) is false
+        const existingInteractionSession = editorStoreRef.current.editor.canvas.interactionSession
```
Rheeseyb, line 146: I think we should probably take the approach suggested in the above `TODO`, i.e. if `Keyboard.keyIsModifier(key)` we only update existing interactions, whereas other keys can create a new interaction

balazsbajorics: the todo is not _quite_ about that, but I think I should flip the if-else order because this is risking a regression if we extend the whitelist in Keyboard.keyIsInteraction

Rheeseyb (approved):

seanparsons (approved):

PR #2166, opened 2022-03-18
rename the absolute strategies

balazsbajorics:
**Problem:**
We have two overlapping resize strategies, and they were badly named Resize and Multiselect Resize.

**Fix:**
Rename the two strategies to `absolute-resize-delta-strategy` (the user visible name is `Absolute Resize (Delta-based)`) and `absolute-resize-bounding-box-strategy` (user visible name printed as `Absolute Resize`)

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
@@ -30,7 +30,7 @@ import {
   runLegacyAbsoluteResizeSnapping,
 } from './shared-absolute-move-strategy-helpers'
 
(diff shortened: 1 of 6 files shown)


editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
```diff
@@ -31,8 +31,8 @@ import {
 } from './shared-absolute-move-strategy-helpers'
 
 export const multiselectAbsoluteResizeStrategy: CanvasStrategy = {
```
enidemi, line 33: I think you want to rename this too

enidemi (approved):

Rheeseyb (approved):

Issue #2059, opened 2021-12-20
wrapping a component in React.memo breaks the spy/dom-walker/uid logic, making the element (and its children) unselectable


balazsbajorics:
given this example code app.js:

```javascript
import * as React from 'react'
import { Card } from '/src/card.js'
import { View } from 'utopia-api'

export var Kutya = (props) => {
  const renderCountRef = React.useRef(0)
  renderCountRef.current++
  return (
    <View
      style={{
        backgroundColor: '#0091FFAA',
        width: 150,
        height: 150,
      }}
      data-uid='kutya-root'
    >
      Render {renderCountRef.current}
    </View>
  )
}

const CicaTemplate = (props) => {
  const renderCountRef = React.useRef(0)
  renderCountRef.current++
  return (
    <View
      style={{
        backgroundColor: '#0091FFAA',
        width: 150,
        height: 150,
      }}
      data-uid='cica-root'
    >
      Render {renderCountRef.current}
      {props.children}
    </View>
  )
}
const CicaWrappedFunc = (props) => (
  <CicaTemplate {...props} />
)
const CicaWrappedMemo = React.memo(CicaTemplate)

export var App = (props) => {
  const renderCountRef = React.useRef(0)
  renderCountRef.current++
  return (
    <div
      data-uid='app-outer-div'
      style={{
        position: 'relative',
        width: '100%',
        height: '100%',
        backgroundColor: '#FFFFFF',
        display: 'flex',
        justifyContent: 'center',
        alignItems: 'flex-start',
      }}
    >
      Render {renderCountRef.current}
      <CicaWrappedFunc data-uid='cica-func'>
        <Kutya />
      </CicaWrappedFunc>
      <CicaWrappedMemo data-uid='cica-memo'>
        <Kutya />
      </CicaWrappedMemo>
    </div>
  )
}
```

the dom-walker correctly finds the dimensions for `CicaWrappedFunc`, but not for `CicaWrappedMemo`.

Selecting `Kutya` inside `CicaWrappedMemo` is possible via the navigator, but then trying to drag it on the canvas deselects it (probably because its parent's bounds is `null`)

My hunch is that the problem comes from `patchedCreateReactElement` not handling the React.memo case.

PR #2168, opened 2022-03-21
Stop printing all paths to DOM

Rheeseyb:
This PR is a pre-cursor to the solution to #2059 

**Problem:**
We print all paths for an element onto the rendered dom element via the `data-paths` attribute. The reason for this is so that we can tie that rendered element to all of the components in the React tree that actually directly resulted in that dom element (for the sake of metadata, and selection following). This approach was fine so far, but the solution to the above ticket makes that approach a lot trickier, and we realised that we only need the final path, since we can extrapolate the other paths via that.

**Fix:**
Only print the final (full) path to the dom element, and then use a function in `element-path.ts` to provide the extrapolation algorithm.

The core changes exist in `element-path.ts` and `canvas-react-utils.ts`.

**Commit Details:**
- Created a function for extrapolating the paths of all elements directly result in the dom element rendered at a given path in `element-path.ts`
- Only attach the resultant path to the element in `canvas-react-utils.ts` rather than appending them all
- Renamed `DATA_UTOPIA_PATHS` to `DATA_UTOPIA_PATH`, and `data-paths` to `data-path`
- Updated all of the test snapshots broken by this, ensuring that only the printed paths had been updated and nothing more.

### DIFF
--- a/editor/src/core/model/performance-scripts.tsx
+++ b/editor/src/core/model/performance-scripts.tsx
@@ -179,7 +179,7 @@ function useTriggerHighlightPerformanceTest(key: 'regular' | 'all-elements'): ()
 
     const targetPath = allPaths.current[0]
 
-    const htmlElement = document.querySelector(`*[data-path~="${EP.toString(targetPath)}"]`)
+    const htmlElement = document.querySelector(`*[data-path*="${EP.toString(targetPath)}"]`)
     if (htmlElement == null) {
       console.info(`HIGHLIGHT_${allCapsKey}_TEST_ERROR_NO_ELEMENT`)
       return
@@ -252,7 +252,7 @@ export function useTriggerSelectionPerformanceTest(): () => void {
 
     const targetElement = forceNotNull(
       'Target element should exist.',
-      document.querySelector(`*[data-path~="${EP.toString(targetPath)}"]`),
+      document.querySelector(`*[data-path*="${EP.toString(targetPath)}"]`),
     )
     const originalTargetBounds = targetElement.getBoundingClientRect()
     const leftToTarget =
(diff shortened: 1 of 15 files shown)


editor/src/core/shared/element-path.ts
```diff
@@ -73,6 +73,28 @@ function getElementPathCache(fullElementPath: ElementPathPart[]): ElementPathCac
 const SceneSeparator = ':'
 const ElementSeparator = '/'
 
+function getComponentPathStringForPathString(path: string): string | null {
+  const indexOfLastSceneSeparator = path.lastIndexOf(SceneSeparator)
+  const indexOfLastElementSeparator = path.lastIndexOf(ElementSeparator)
+  if (indexOfLastSceneSeparator > indexOfLastElementSeparator) {
```
gbalint, line 79: It is really implicit here what happens if one (or both) of the indexes is/are -1
I would handle this even if it doesn't happen IRL

Rheeseyb: This is actually by design, but I have no issue making it explicit :+1:

editor/src/core/model/performance-scripts.tsx
```diff
@@ -179,7 +179,7 @@ function useTriggerHighlightPerformanceTest(key: 'regular' | 'all-elements'): ()
 
     const targetPath = allPaths.current[0]
 
-    const htmlElement = document.querySelector(`*[data-paths~="${EP.toString(targetPath)}"]`)
+    const htmlElement = document.querySelector(`*[data-path~="${EP.toString(targetPath)}"]`)
```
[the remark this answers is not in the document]
Rheeseyb: Ha, I literally just pushed a fix to that! But, the fix is now using `*=`, which _should_ return the same element based on the path...

editor/src/core/shared/uid-utils.ts
```diff
@@ -238,7 +238,8 @@ function getSplitPathsStrings(pathsString: string | null): Array<string> {
   if (pathsString == null) {
     return []
   } else {
-    return pathsString.split(' ')
+    const firstPath = pathsString.split(' ')[0]
```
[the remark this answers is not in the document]
Rheeseyb: Done :+1:

editor/src/core/shared/element-path.ts
seanparsons, line 79: Could you add a quick note as to why this condition holds?

seanparsons (approved):

PR #2170, opened 2022-03-22
Feature/canvas strategy reparent element

enidemi:
This PR fixes the absolute reparent strategy and adds tests.
It was not far from complete as the reparent and move are already handled by this strategy, the missing feature is that only elements with top/left were offset for the new parent, it needs offset for the bottom/right style properties too.
The other issue I noticed is the result of calculating the offset deltas with elements with border, the borders are included in the canvas frame, but for pin positions the inner content frame is needed.

The new parent target is also restricted to elements that are not flow using the `providesBoundsForChildren` from the metadata.

**Commit Details:**
- fix the absolute reparent strategy offset
- extended metadata to calculate the inner content box, the default border-box box-sizing includes the borders in the frame too.
- tests

### DIFF
--- a/editor/src/components/editor/store/store-deep-equality-instances.ts
+++ b/editor/src/components/editor/store/store-deep-equality-instances.ts
@@ -802,6 +802,10 @@ export function SpecialSizeMeasurementsKeepDeepEquality(): KeepDeepEqualityCall<
     const displayEquals = oldSize.display === newSize.display
     const htmlElementNameEquals = oldSize.htmlElementName === newSize.htmlElementName
     const renderedChildrenCount = oldSize.renderedChildrenCount === newSize.renderedChildrenCount
+    const globalContentBoxEquals = nullableDeepEquality(CanvasRectangleKeepDeepEquality)(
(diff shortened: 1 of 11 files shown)


editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -802,6 +802,10 @@ export function SpecialSizeMeasurementsKeepDeepEquality(): KeepDeepEqualityCall<
     const displayEquals = oldSize.display === newSize.display
     const htmlElementNameEquals = oldSize.htmlElementName === newSize.htmlElementName
     const renderedChildrenCount = oldSize.renderedChildrenCount === newSize.renderedChildrenCount
+    const globalContentBoxEquals = nullableDeepEquality(CanvasRectangleKeepDeepEquality)(
```
seanparsons, line 805: This should be getting the value of `areEqual` on the end.

seanparsons (approved):

gbalint (approved):

Rheeseyb (commented):
Nice!

PR #2171, opened 2022-03-22
Add parametrized keyboard interaction tests

gbalint:
Introducing parametrized tests with `it.each`
This way the same test can the keyboard interaction behavior for different key and modifier combinations.


enidemi (approved):

Rheeseyb (commented):
:+1:

PR #2172, opened 2022-03-23
Support for tests to provide their own project definitions.

seanparsons:
**Problem:**
Our performance tests are constrained to a single project which has to exist in the target environment.

**Fix:**
Changed the performance tests from a model of loading a given project upfront and then running all of the tests against that project. To one where each performance tests loads a project potentially specific to that test.

**Commit Details:**
- Provide a specific ID for the `VSCodeLoadingScreen` component so it can
  be tracked.
- Added `loadProject` to `performance-scripts.tsx` which does the work of
  loading the project and ensuring the editor is in a state which means
  the tests can be run.
- Ensure that `useTriggerResizePerformanceTest` has an element selected.
- Ensure that `useTriggerAbsoluteMovePerformanceTest` switches the editor
  to the correct mode.
- Removed the old hardcoded project ID from the performance tests.
- Separated the empty dispatch and calculate PI performance tests from
  being so wedded to each other.
- Removed `initialiseTestsReturnScale` as it wasn't necessary anymore
  and was the cause of some long pauses at the start of the performance
  tests.
- `consoleDoneMessage` also unregisters the handler from the `console`
  messages when it finishes.
- Removed `initialiseTests` as it wasn't necessary anymore and caused
  a long pause during the performance test.
- Put a sizable project definition into a file so that it can be utilised without
  having to make sure the project exists in an environment.

### DIFF
--- a/editor/src/core/model/performance-scripts.tsx
+++ b/editor/src/core/model/performance-scripts.tsx
@@ -74,7 +74,7 @@ function measureStep(prefix: string, framesPassed: number): void {
   )
 }
 
-const CANVAS_POPULATE_WAIT_TIME_MS = 20 * 1000
+const CANVAS_POPULATE_WAIT_TIME_MS = 60 * 1000
 
 async function loadProject(
   dispatch: DebugDispatch,
--- a/editor/src/core/model/performance-scripts.tsx
+++ b/editor/src/core/model/performance-scripts.tsx
@@ -10,6 +10,7 @@ import {
   setProp_UNSAFE,
   switchEditorMode,
   unsetProperty,
+  updateEditorMode,
 } from '../../components/editor/actions/action-creators'
 import { useEditorState, useRefEditorState } from '../../components/editor/store/store-hook'
 import {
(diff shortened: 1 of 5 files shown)


editor/src/core/model/performance-scripts.tsx
```diff
@@ -66,15 +74,120 @@ function measureStep(prefix: string, framesPassed: number): void {
...
+  await load(dispatch, persistentModel, 'Test', '999999', builtInDependencies, false)
+
+  // Wait for the editor to stabilise, ensuring that the canvas can render for example.
+  const startWaitingTime = Date.now()
+  let editorReady: boolean = false
+  let itemSelected: boolean = false
+  let canvasPopulated: boolean = false
+  let codeEditorPopulated: boolean = false
+  let codeEditorLoaded: boolean = false
+  while (startWaitingTime + 20000 > Date.now() && !editorReady) {
```
gbalint, line 118: I would extract all the magic time numbers to named constants.

gbalint (approved):

editor/src/core/model/performance-scripts.tsx
seanparsons: Yeah, that's a good shout.

seanparsons: Fixed in 2d46045d79166ca6ca771186e5192e00c42b06a0.

Rheeseyb (commented):
:+1:

PR #2173, opened 2022-03-24
Fix keyboard interaction timeout

gbalint:
# Bug

Keyboard interaction timeout was never triggered

# Root cause

Probably a buggy code completion, we just called a completely different clear function than what was needed.....

If I was there, I also converted the function to a hook, so the keyboard timeout handler ref can be created inside the hook. Because the function had to be called conditionally, I created a hook, which returns a function, which can be called inside the conditional

### DIFF
--- a/editor/src/components/editor/editor-component.tsx
+++ b/editor/src/components/editor/editor-component.tsx
@@ -60,7 +60,7 @@ import {
   createInteractionViaKeyboard,
   updateInteractionViaKeyboard,
 } from '../canvas/canvas-strategies/interaction-state'
-import { setupClearKeyboardInteraction } from '../canvas/controls/select-mode/select-mode-hooks'
+import { useClearKeyboardInteraction } from '../canvas/controls/select-mode/select-mode-hooks'
 
 function pushProjectURLToBrowserHistory(projectId: string, projectName: string): void {
   // Make sure we don't replace the query params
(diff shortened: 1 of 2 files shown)


editor/src/components/editor/editor-component.tsx
```diff
@@ -192,7 +193,7 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
         editorStoreRef.current.dispatch,
       )
     },
-    [editorStoreRef, namesByKey],
+    [editorStoreRef, namesByKey, setClearKeyboardInteraction],
```
balazsbajorics, line 196: the function returned by the hook is not memoized, so it'll always be a new reference, so this will always trip the reference equality check

seanparsons (approved):

editor/src/components/editor/editor-component.tsx
gbalint: True! I wonder which is nicer, memoizing the hook, or just converting it back to its original form.

gbalint: I fixed it keeping the hook

enidemi (approved):

balazsbajorics (commented):
👍

Rheeseyb (commented):
:+1:

PR #2175, opened 2022-03-24
Keyboard resize strategy

gbalint:
New strategy to resize with keyboard arrows using the cmd modifier.
It always resizes using be moving the left or the bottom side of the element.
If you press shift too, it uses bigger movements.

Implementation note:
I had to realize that when the cmd button is pressed, browsers do not trigger keyup events for other buttons!!!!!!
https://stackoverflow.com/questions/27380018/when-cmd-key-is-kept-pressed-keyup-is-not-triggered-for-any-other-key
This means it is not easy to differentiate between these cases:
1) You pressed cmd and then one arrow, and after the another arrow
2) You pressed cmd and then one arrow, and then without releasing that you pressed another arrow too.

As a workaround, only in the case of cmd modifier, I automatically release the a button if there are no keydown event for that key in the last 500ms. 
Shout if you have a better solution!


editor/src/components/canvas/canvas-strategies/keyboard-absolute-resize-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+        metadata,
+        canvasState.selectedElements[0],
+      )
+
+      return elementMetadata?.specialSizeMeasurements.position === 'absolute'
+    } else {
+      return false
+    }
+  },
+  controlsToRender: [], // Uses existing hooks in select-mode-hooks.tsx
```
enidemi, line 26: I think the `AbsoluteResizeControl` should be included here

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -44,9 +44,11 @@ import { isFeatureEnabled } from '../../../../utils/feature-switches'
 import {
   createInteractionViaMouse,
   KeyboardInteractionTimeout,
+  updateInteractionViaKeyboard,
 } from '../../canvas-strategies/interaction-state'
-import { Modifier } from '../../../../utils/modifiers'
+import { Modifier, Modifiers } from '../../../../utils/modifiers'
 import { pathsEqual } from '../../../../core/shared/element-path'
+import { getNodeMajorVersion } from 'typescript'
```
enidemi, line 51: this is an unused import

enidemi (approved):

editor/src/components/canvas/canvas-strategies/keyboard-absolute-resize-strategy.tsx
```diff
@@ -0,0 +1,146 @@
...
+        canvasState,
+        interactionState,
+        sessionState.startingMetadata,
+      ) &&
+      interactionState.interactionData.type === 'KEYBOARD'
+    ) {
+      const { interactionData } = interactionState
+
+      const arrowKeyPressed = interactionData.keysPressed.some(Keyboard.keyIsArrow)
+      const cmdAndOptionallyShiftModifier =
```
seanparsons, line 41: This mentions `Shift` in the name, but the conditions don't check the shift key?

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/keyboard-absolute-resize-strategy.tsx
gbalint: The name wanted to express that it is _exactly_ cmd and optionally shift, so no alt, no ctrl. And the shift is not the in the condition, because it is optional :) I'm open to a better name

Rheeseyb: I think it would have been better to name this to indicate that it is explicitly rejecting `ctrl` and `alt` in that case

Rheeseyb:
> As a workaround, only in the case of cmd modifier, I automatically release the a button if there are no keydown event for that key in the last 500ms.
Shout if you have a better solution!

Honestly, this seems like the best solution to me too :+1:

PR #2186, opened 2022-03-31
Aspect ratio locked resize 

gbalint:
This PR implements the aspect ratio locked resize (using the mouse)

Aspect ratio locked resize has 2 distinct cases:
- when you drag a corner we resize the element in a way that the mouse cursor stays on the sides of the bounding box
Mar-31-2022 12-57-57

- when you drag a side we do the same, but we also keep the center of the rectangle in the same position, so it grows/shrinks in two ways to the two sides.
Mar-31-2022 12-58-46

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-resize-delta-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-resize-delta-strategy.tsx
@@ -73,7 +73,7 @@ export const absoluteResizeDeltaStrategy: CanvasStrategy = {
         drag,
         edgePosition,
         canvasState.scale,
-        false,
+        null,
       )
 
       const commandsForSelectedElements = canvasState.selectedElements.flatMap(
@@ -122,7 +122,7 @@ function snapDrag(
   drag: CanvasVector,
   edgePosition: EdgePosition,
   canvasScale: number,
-  keepAspectRatio: boolean,
+  lockedAspectRatio: number | null,
 ): {
   snappedDragVector: CanvasVector
   guidelinesWithSnappingVector: Array<GuidelineWithSnappingVector>
@@ -137,7 +137,7 @@ function snapDrag(
     originalBoundingBox,
     drag,
     edgePosition,
-    false,
+    null,
     false,
   )
   const { snapDelta, guidelinesWithSnappingVector } = runLegacyAbsoluteResizeSnapping(
@@ -146,7 +146,7 @@ function snapDrag(
     edgePosition,
     resizedUnsnappedBounds,
     canvasScale,
-    keepAspectRatio,
+    lockedAspectRatio,
     false,
   )
   const snappedDragVector = offsetPoint(drag, snapDelta)
--- a/editor/src/components/canvas/canvas-strategies/absolute-resize-delta-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-resize-delta-strategy.tsx
@@ -32,7 +32,8 @@ export const absoluteResizeDeltaStrategy: CanvasStrategy = {
   isApplicable: (canvasState, interactionState, metadata) => {
     if (
       canvasState.selectedElements.length === 1 &&
-      !interactionState?.interactionData.modifiers.alt
+      !interactionState?.interactionData.modifiers.alt &&
+      !interactionState?.interactionData.modifiers.shift // shift is aspect ratio locked resize implemented in absolute-resize-bounding-box-strategy.tsx
     ) {
       return canvasState.selectedElements.every((element) => {
         const elementMetadata = MetadataUtils.findElementByElementPath(metadata, element)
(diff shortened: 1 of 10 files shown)


editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
```diff
@@ -191,6 +201,7 @@ export function resizeBoundingBox(
   boundingBox: CanvasRectangle,
   drag: CanvasPoint,
   edgePosition: EdgePosition,
+  keepAspectRatio: boolean,
```
Rheeseyb, line 204: Please can you change these boolean flags to string unions so that it's clearer which flag is being set

enidemi (approved):

editor/src/components/canvas/canvas-strategies/absolute-resize-delta-strategy.tsx
```diff
@@ -121,7 +111,7 @@ function snapDrag(
   drag: CanvasVector,
   edgePosition: EdgePosition,
   canvasScale: number,
-  keepAspectRatio: boolean,
+  lockedAspectRatio: number | null,
```
balazsbajorics, line 114: I believe this is always going to be null for this strategy!

editor/src/components/canvas/canvas-strategies/shared-absolute-resize-strategy-helpers.ts
```diff
@@ -79,3 +88,175 @@ function pinsForEdgePosition(edgePosition: EdgePosition): AbsolutePin[] {
...
+    cornerEdgePosition = { x: edgePosition.x, y: 1 }
+  }
+
+  const oppositeCornerPoint = pickPointOnRect(boundingBox, startingCornerPosition)
+  const draggedCorner = pickPointOnRect(boundingBox, cornerEdgePosition)
+  let newCorner = offsetPoint(draggedCorner, dragToUse)
+
+  if (lockedAspectRatio != null) {
+    const horizontalLineB = canvasPoint({
+      x: newCorner.x + 100,
```
balazsbajorics, line 127: what's this magic +100 here?

is it a problem that you maybe reassign newCorner at line 136?

gbalint: This just creates a horizonal and a vertical line here. It doesn't matter if this is 100 or a different constant, what matters is just that the other coordinate remains the same.
Line 136 is ok, that is what calculates the horizontal/vertical projection of this to the diagonal (which will the good newCorner)

editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
gbalint: done

balazsbajorics (approved):

enidemi (approved):

gbalint:
Let's refactor first

gbalint:
Cleaned up the code

editor/src/components/canvas/canvas-strategies/shared-absolute-resize-strategy-helpers.ts
```diff
@@ -79,3 +88,372 @@ function pinsForEdgePosition(edgePosition: EdgePosition): AbsolutePin[] {
 
   return [...horizontalPins, ...verticalPins]
 }
+
+type IsCenterBased = 'center-based' | 'non-center-based'
```
seanparsons, line 92: Probably worth exporting this and using it in the function parameters.

editor/src/core/shared/math-utils.ts
```diff
@@ -741,6 +741,22 @@ export function lineIntersection<C extends CoordinateMarker>(
   } as Point<C>
 }
 
+export function sideOfLine<C extends CoordinateMarker>(
+  lineA: Point<C>,
+  lineB: Point<C>,
+  p: Point<C>,
+): 'left' | 'right' | 'on-line' {
+  const d = (p.x - lineA.x) * (lineB.y - lineA.y) - (p.y - lineA.y) * (lineB.x - lineA.x)
```
seanparsons, line 749: Would it be possible to have a little comment here explaining what's going on here please?

seanparsons (approved):

editor/src/core/shared/math-utils.ts
gbalint: Added comments in 86186a0b08b4e3fec1e02751247e6e86f33d819d

editor/src/components/canvas/canvas-strategies/shared-absolute-resize-strategy-helpers.ts
gbalint: I'm not sure which function you mean. I use it in all new `resizeBoundingBox...` and `snapping` functions, and I don't know of other functions which would need this.

seanparsons: I can see it here for example: https://github.com/concrete-utopia/utopia/pull/2186/files#diff-95ecc07766bb6e984683db3702bb6b6d359a383a34ff254bff8f672c24193924R201

PR #2192, opened 2022-04-01
Run the performance and screenshot tests without using headless.

seanparsons:
**Problem:**
The baseline test appears to be flaky in relation to whether or not the tests are run headless. That is when running with the UI showing it appears to work flawlessly repeatedly, but in local testing switching `HEADLESS=true` on made it fail about 50% of the time.

**Fix:**
Completely disable the "headless" operation of Puppeteer as it seemingly works different somehow.

**Commit Details:**
- Put invocations of `xvfb-run` into the various calls that start any
  puppeteer tests.
- Disable the `HEADLESS` environment variable handling.
- Fix up the readme to represent that `HEADLESS` can't be used.
- Added a function `retryPageCalls` so that a specific failing performance test can be attempted more than once.
- Remove some unncessary cleanup logic from the absolute move test.
- Removed an unnecessary wait from `retryPageCalls`.
- Remove the `runBeforeUnload` option from `page.close` as it appears to do
  nothing.
- Added dialog dismissal callback so that any `beforeunload` handling is immediately
  dismissed so that the tests can continue.
- Use UUID V4 for project IDs to avoid possible collisions with using old ones.
- Removed some logic that would cause cross-origin errors as it was attempting to access the code editor iframe.

### DIFF
--- a/puppeteer-tests/README.md
+++ b/puppeteer-tests/README.md
@@ -3,10 +3,9 @@
 1. npm install
 2. Create a project on localhost and copy its URL.
 3. run the performance test like this in the terminal: `EDITOR_URL=<project-url-you-just-copied> npm run start`
-4. can provide an env var HEADLESS=false to see what the editor is doing: `HEADLESS=false EDITOR_URL=<project-url-you-just-copied> npm run start`
-5. It's possible to provide a browser path if the bundled one is no use: `BROWSER='google-chrome' EDITOR_URL=<project-url-you-just-copied> npm run start`
-6. you will see console output with the frame numbers.
-7. That's all you need for debugging existing tests!
+4. It's possible to provide a browser path if the bundled one is no use: `BROWSER='google-chrome' EDITOR_URL=<project-url-you-just-copied> npm run start`
+5. you will see console output with the frame numbers.
+6. That's all you need for debugging existing tests!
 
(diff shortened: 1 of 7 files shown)


puppeteer-tests/README.md
```diff
@@ -3,10 +3,9 @@
 1. npm install
 2. Create a project on localhost and copy its URL.
 3. run the performance test like this in the terminal: `EDITOR_URL=<project-url-you-just-copied> npm run start`
-4. can provide an env var HEADLESS=false to see what the editor is doing: `HEADLESS=false EDITOR_URL=<project-url-you-just-copied> npm run start`
-5. It's possible to provide a browser path if the bundled one is no use: `BROWSER='google-chrome' EDITOR_URL=<project-url-you-just-copied> npm run start`
-6. you will see console output with the frame numbers.
-7. That's all you need for debugging existing tests!
+4. It's possible to provide a browser path if the bundled one is no use: `BROWSER='google-chrome' EDITOR_URL=<project-url-you-just-copied> npm run start`
```
Rheeseyb, line 6: Minor: all of these should be using `pnpm` rather than `npm`. I have no issue making that change afterwards since it's completely unrelated to this PR

Rheeseyb (approved):

balazsbajorics (approved):

PR #2194, opened 2022-04-04
Ensure some ordering of localforage logic.

seanparsons:
**Problem:**
While doing some performance testing, it was observed that the LocaleForage initialisation would fairly often end up interspersed with code that was trying to access the LocalForage store. That invariably resulted in bizarre failures which were very race condition like in their variability.

**Fix:**
Now when accessing the store, the following checks are made:
- Wait for the first initialisation to complete.
- Wait for any currently executing initialisation to complete.

**Commit Details:**
- Have `initializeStore` chain off the back of (potential) previous
  invocations so that there's never more than one running simultaneously.
- `withSanityCheckedStore` now ensures that both a first initialisation
  has completed and that the current initialisation has completed.

### DIFF
--- a/utopia-vscode-common/src/fs/fs-core.ts
+++ b/utopia-vscode-common/src/fs/fs-core.ts
@@ -10,29 +10,42 @@ let thisDBName: string
 
 const StoreExistsKey = '.store-exists'
 
+/* eslint-disable-next-line @typescript-eslint/no-empty-function */
+let resolveFirstInitialize: () => void = () => {}
+const firstInitialize = new Promise<void>((resolve) => {
+  resolveFirstInitialize = resolve
+})
+let initializeStoreChain: Promise<void> = Promise.resolve()
+
 export async function initializeStore(
   storeName: string,
   driver: string = localforage.INDEXEDDB,
 ): Promise<void> {
-  thisDBName = `utopia-project-${storeName}`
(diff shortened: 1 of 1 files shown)


utopia-vscode-common/src/fs/fs-core.ts
```diff
@@ -10,29 +10,42 @@ let thisDBName: string
 
 const StoreExistsKey = '.store-exists'
 
+/* eslint-disable-next-line @typescript-eslint/no-empty-function */
+let resolveFirstInitialize: () => void = () => {}
+const firstInitialize = new Promise<void>((resolve) => {
+  resolveFirstInitialize = resolve
+})
```
balazsbajorics, line 17: we have a util to create a promise that can be externally resolved, called `defer` in `utils.ts`.
it encapsulates the promise and the externally accessible resolver function, so it lets you avoid creating a `let` in the module scope and mutating it.

I believe the code to use it would be this:
```
const firstInitialize = defer<void>()
```

and instead of calling `resolveFirstInitialize()`, you'd write
```
firstInitialize.resolve()
```

Rheeseyb: I _knew_ we were using this exact pattern somewhere else!

Rheeseyb (approved):

PR #2202, opened 2022-04-07
Parametrized tests for bounding box resize

gbalint:
I added parametrized tests for single select bounding box resize.

In my opinion these are more compact and way easier to read, they also cover the full functionality (all corners and sides with all modifier combinations).

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.spec.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.spec.tsx
@@ -7,7 +7,13 @@ import {
 } from '../../../core/shared/element-template'
 import { CanvasPoint, canvasPoint, canvasRectangle } from '../../../core/shared/math-utils'
 import { ElementPath } from '../../../core/shared/project-file-types'
-import { altModifier, emptyModifiers, Modifiers, shiftModifier } from '../../../utils/modifiers'
+import {
+  altModifier,
+  altShiftModifier,
+  emptyModifiers,
+  Modifiers,
+  shiftModifier,
+} from '../../../utils/modifiers'
 import { EditorState } from '../../editor/store/editor-state'
 import { EdgePosition } from '../canvas-types'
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.spec.tsx
```diff
@@ -48,754 +57,1177 @@ function multiselectResizeElements(
...
+        draggedBounding: { left: 29, top: 25, width: 292, height: 350 },
+      },
+    ],
+    [
+      'right side, center mode, aspect ratio locked',
+      {
+        edgePosition: { x: 1, y: 0.5 } as EdgePosition,
+        drag: testDrag,
+        modifiers: altShiftModifier,
+        bounding: testBounding,
```
balazsbajorics, line 411: I think it would be better if the testBounding rectangle would be here inline, I forgot the rectangle while I scrolled here, and now I have to scroll back a lot to find the rectangle size, memorize it, then scroll back here (find the test again) and then compare it with the result :)

enidemi (approved):

editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.spec.tsx
Rheeseyb: I'd also like to suggest renaming `draggedBounding` to `expectedBounding`. When first reading I _thought_ it was the expected result, but had to then read through the code itself to find where it was used to verify that.

Rheeseyb (approved):
Nice work! I'm a huge fan of this `it.each` style of creating tests for variant calls to a function!

editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.spec.tsx
gbalint: Ok, actually that was my first version, I just wanted to eliminate redundancy, but I agree with your points, I revert to inline literals.

PR #2208, opened 2022-04-12
Refactor/dom walker function

balazsbajorics:
Currently, this is how the update flow looks like on master:

Editor dispatch ➔ run update functions ➔ update zustand store ➔ triggers React render of the entire editor ➔ Canvas renders ➔ UseLayoutEffect in canvas root triggers dom walker ➔ DOM walker dispatches SAVE_DOM_REPORT ➔ run update functions ➔ update zustand store ➔ new react render triggered (elements that were queued to rerender but did not render before the DOM walker ran will only run once) ➔ ((we hope that the canvas this time will not trigger the dom walker)) ➔ the rest of the editor UI renders ➔ Chrome draws to screen.

**TLDR: the new flow is
Editor dispatch ➔ update functions ➔ update the canvas zustand store ➔ canvas renders, react flushSync updates the DOM of the Canvas (chrome does NOT draw yet!) ➔ run dom-walker ➔ save the dom walker jsxMetadata to the editor state ➔ update the "main" zustand store ➔ everything renders ➔ react flushSync updates the DOM of the Editor UI ➔ chrome draws on screen.**

The main problem with the old flow is that right in the middle there's an asynchronous UseLayoutEffect which triggers a second dispatch. This makes it very hard for us to write tests but it also makes it hard to improve the performance of the editor.

While digging through the performance measurements, I realized that when we synchronously trigger a react update (by updating the zustand store) it will synchronously return at a point where it's safe to inspect the contents of the DOM. It makes intuitive sense but for some reason this just wasn't a core assumption when we built the DOM walker. 
(Why? I _think_ the only reason is that the dom-walker was developed before we used contexts and zustand so it would have been very hard for us to implement a separate render pass for the canvas without turning it into its own React root. I hope I'm not missing a bigger, still relevant reason!)

With this in mind, we could get rid of the asynchronous useLayoutEffect, but that's not all. We could finally rework things so that we render only the canvas contents first, then do a DOM measurement, then render the rest of the editor. This significantly increases how consciously we are rendering things which makes it easier to implement performance improvements. One such performance improvement is to only re-render small parts of the canvas during an interaction.

## what this PR does
The following changes were kind of intertwined so I turned them into a single PR:
- turn the dom-walker into a regular function. it is triggered from editor.tsx at the end of the dispatch flow, right after we synchronously update the zustand store. At this stage the DOM has been updated by React but chrome did not render yet, so we can update everything on screen without drawing a frame with delayed controls.
- this in itself causes a flow of dispatch -> rendering everything -> measuring the canvas -> re-rendering everything, which is a significant performance _regression_, so in order to avoid that...
- ...introduce two zustand stores, one dedicated to only the canvas contents, the other for the rest of the editor. This is to avoid the performance regression.
- To make sure the zustand store update reached the DOM, we use the new React 18 API `ReactDOM.flushSync`. once a call to flushSync synchronously return, we can trust React that all the work has wrapped up and the DOM is up-to-date.
- The final flow is: Editor dispatch ➔ update functions ➔ update the canvas zustand store ➔ canvas renders, react flushSync updates the DOM of the Canvas (chrome does NOT draw yet!) ➔ run dom-walker ➔ save the dom walker jsxMetadata to the editor state ➔ update the "main" zustand store ➔ everything renders ➔ react flushSync updates the DOM of the Editor UI ➔ chrome draws on screen.
- I also updated `renderTestEditorWithModel` so that it replicates the same flow as editor.tsx, but the tests are not yet passing.

## Dom walker as regular function
`useDomWalker` is removed and turned into a regular function. The mutable state that was managed as `useEditorState` hooks has been turned into an editor-level Context. I used this as an opportunity to simplify some of the messiest code in the dom walker that was dealing with scene and element-level invalidation. The dom walker is maintaining and mutating Sets of invalidated ElementPaths. This has been going on for 1+ year, but I made it very explicit now instead of immutable-looking update functions that were in reality mutating.

## Separate zustand store for the Canvas contents
This PR creates two zustand stores. One store is dedicated to the Canvas root, the other store is for all the rest of the editor. The way it is implemented is that I put both stores into their own Context, and in the render tree near the canvas root I replace the "main" EditorStateContext with the contents of CanvasStateContext in a context override. 
This means that `useEditorState` still uses `EditorStateContext` everywhere, but when called inside the subtree belonging to the Canvas root it will see the canvas store, everywhere else will see the "main" store. I strongly believe this is the best solution because it prevents us from accidentally using the wrong store. The same `useEditorState` function will simply see a different store depending on where in the react render tree you are calling it from. This is good because accidentally using the "main" store in the canvas (or vice versa, using the canvas store in the rest of the editor) would trigger re-renders that are very hard to spot and debug, and regress performance significantly.

## ReactDOM.flushSync
The final piece of the puzzle is [flushSync](https://reactjs.org/docs/react-dom.html#flushsync), here's how the API looks like:
```javascript
flushSync(() => {
  storeHook.setState(...)
})
// At this point, we know the DOM is updated
// and we are safe to query it in the dom-walker
console.log(document.getElementById...
```

# TODO 
- [x] check and fix TODO left around in the new code
- [x] check tests that are not passing
- [x] the dom-walker tests are not updated yet


editor/src/components/canvas/dom-walker.ts
```diff
@@ -428,112 +249,227 @@ function mergeFragmentMetadata(
...
+    if (LogDomWalkerPerformance) {
+      performance.mark('DOM_WALKER_START')
+    }
+    // Get some base values relating to the div this component creates.
+    if (
+      ObserversAvailable &&
+      domWalkerMutableState.resizeObserver != null &&
+      domWalkerMutableState.mutationObserver != null
+    ) {
+      Array.from(document.querySelectorAll(`#${CanvasContainerID} *`)).map((elem) => {
```
seanparsons, line 331: The `NodeList` that comes back from `querySelectorAll` has a `forEach` function which should simplify this.
```suggestion
      document.querySelectorAll(`#${CanvasContainerID} *`).forEach((elem) => {
```

seanparsons:
Is there a "canvas renders" missing in the new flow? `Editor dispatch ➔ update functions ➔ update the canvas zustand store ➔ run dom-walker...`

balazsbajorics:
> Is there a "canvas renders" missing in the new flow? `Editor dispatch ➔ update functions ➔ update the canvas zustand store ➔ run dom-walker...`

thanks, I amended the flow by inserting "canvas renders" after "update the canvas zustand"

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -178,7 +178,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`2382`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`581`)
```
Rheeseyb, line 181: Wait what?! This is suspiciously good, but I guess intended...

Rheeseyb (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
balazsbajorics: this number changed when I added flushSync. my theory that I didn't bother to verify is that what we see here is the effect of batching the updates. but I'm not too fussed to figure this out, because this test is more of a canary to flag PRs that did not intend to change this number and initiate a conversation

editor/src/components/canvas/dom-walker.ts
seanparsons: *cough*

editor/src/core/performance/performance-regression-tests.spec.tsx
Rheeseyb: Well yeah kinda, but if the number fluctuates by too much it's also worthwhile investigating to make sure that's not the result of a significant regression

editor/src/components/canvas/dom-walker.ts
balazsbajorics: fixed!

gbalint (approved):

PR #2211, opened 2022-04-25
Fix: Reparent to different file

gbalint:
# Issue
You can not reparent from one file to another. (Exception is thrown and utopia crashes)

# Description

This is not really a bug, more like an unimplemented feature, the `runReparentElement` function is implemented in a way that it absolutely expects to target and the new parent element to be in the same file.

This PR just moves the target element to new file. That can easily cause problems if the element is dependent on variable from its source scope, or uses an import. (But this can cause problems even in a single file.) Full reparenting solution would mean we have to implement quite complex refactorings, so let's skip that for now.

One problem I needed to solve is that in case of multiple files we need to:
1. remove the target element from the old file
2. add the target element to the new file

Unfortunately, creating an editor patch is a quite manual effort (see https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/commands/reparent-element-command.ts#L67-L104 )

I extracted this logic into a function `getPatchForComponentChange` (maybe this can be reused elsewhere in other commands?)
However, I still had the issue that I have a separate patch for 1. and 2.
Merging editor state patches is a quite complex task I struggled with for a while, but finally I decided to refactor our commands in a way that they can include an array of editor state patches.

# Problems:

- This solution can reparent into a user defined component which doesn't render its children. In this case the target element just disappears. If we could recognize this situation, we could move it inside the root component inside the render function of the user defined component.
- The reparenting itself can easily result in exceptions, e.g. if variables are missing from the new scope.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.ts
+++ b/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.ts
@@ -62,21 +62,7 @@ export const absoluteReparentStrategy: CanvasStrategy = {
       newParent,
     )?.specialSizeMeasurements.providesBoundsForChildren
 
-    const selectedElementFiles = filteredSelectedElements.map((element) =>
-      getFileOfElement(element, projectContents, openFile),
-    )
-
-    const newParentFile = getFileOfElement(newParent, projectContents, openFile)
-
-    // Currently we only support reparenting into the same file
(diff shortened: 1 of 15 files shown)


balazsbajorics (approved):
can you make a new reparent test case for testing reparent across files?

enidemi (approved):

seanparsons (approved):

PR #2219, opened 2022-04-29
Multiselection for move and reparent strategies

gbalint:
# Issue

## Reparenting
Multiselection handling was not implemented in the absolute reparenting strategy.

## Moving
Multiselection was implemented in the absolute move strategy, but it did not filter out descendants, which caused them to move twice (their moved with their parent and also locally in the parent)

# Solution

I implemented multiselection in the reparenting strategy.
Filter out the descendants from the selection, they move with their parent anyway.
I added multiselection tests for both strategies

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
@@ -198,3 +198,10 @@ export function getFileOfElement(
     (_success, _element, _underlyingTarget, underlyingFilePath) => underlyingFilePath,
   )
 }
+
+// No need to include descendants in multiselection when dragging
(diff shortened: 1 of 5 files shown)


editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
```diff
@@ -198,3 +198,10 @@ export function getFileOfElement(
     (_success, _element, _underlyingTarget, underlyingFilePath) => underlyingFilePath,
   )
 }
+
+// No need to include descendants in multiselection when dragging
+export function getDragTargets(selectedViews: Array<ElementPath>): Array<ElementPath> {
+  return selectedViews.filter((view) =>
```
balazsbajorics, line 204: maybe worth adding a comment here to note that if you select lots of elements this may be slow

PR #2225, opened 2022-05-05
Remove the now redundant appending of UIDs

Rheeseyb:
**Problem:**
We previously relied on all UIDs for an element being appended to the `data-uid` prop as a space separated string, as that was required for building all paths for the rendered DOM element. However, this became redundant with #2168 which removed the requirement for all paths to be included (since we can calculate them from the longest path). The fact this became redundant was overlooked at the time, but is now causing issues with dynamically building the paths (a piece of ongoing work), so now is the time to fix it.

**Fix:**
Remove the UID appending behaviour, switching back to the previous behaviour (which for a root element is to use its UID if one is provided, or fall back to the UID in the component's props if one exists there).

**Commit Details:**
- Remove UID appending in `canvas-react-utils.ts` and in `ui-jsx-canvas-element-renderer-utils.tsx`
- Rename constants and variables to reflect that there is now only a single UID in the `data-uid` attribute
- Updated all affected `data-uid` values in tests

### DIFF
--- a/editor/src/components/canvas/__snapshots__/ui-jsx-canvas-bugs.spec.tsx.snap
+++ b/editor/src/components/canvas/__snapshots__/ui-jsx-canvas-bugs.spec.tsx.snap
@@ -20,11 +20,9 @@ exports[`UiJsxCanvas #747 - DOM object constructor cannot be called as a functio
         width: 375px;
         height: 812px;
       \\"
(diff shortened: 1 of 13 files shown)


seanparsons (approved):
The multiple paths has definitely confused me in the last couple of weeks so I'm all for this!

balazsbajorics (approved):

PR #2227, opened 2022-05-06
Flex reorder strategy

gbalint:
Flex reordering strategy.

No special controls or any visual effects, animation etc., it just the implements the flex reordering itself

Limitations:
- I expect the elements appear in the flex order in the screen. If they have manual offsets then this may not be true.
- It only supports single selection.


editor/src/components/canvas/canvas-strategies/flex-reorder-strategy.tsx
```diff
@@ -0,0 +1,160 @@
...
+      canvasPoint({
+        x: (frames[i].x + frames[i].width + frames[i + 1].x) / 2,
+        y: (frames[i].y + frames[i].height + frames[i + 1].y) / 2,
+      }),
+    )
+  }
+  return [firstPoint, ...middlePoints, lastPoint]
+}
+
+function reverseIndexIfRightToLeft(idx: number, length: number, isLeftToRight: boolean) {
```
balazsbajorics, line 158: nice

balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/flex-reorder-strategy.tsx
```diff
@@ -0,0 +1,160 @@
...
+    ) {
+      return []
+    }
+    const { selectedElements } = canvasState
+    const target = selectedElements[0]
+
+    const siblingsOfTarget = MetadataUtils.getSiblings(sessionState.startingMetadata, target).map(
+      (element) => element.elementPath,
+    )
+    const flexSiblingsOfTarget = siblingsOfTarget.filter((child) =>
```
seanparsons, line 54: Might be worth having a test for the case where one of the siblings returns false for this conditional.

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/flex-reorder-strategy.tsx
gbalint: Such a good point! I wrote it and realized that case has a bug! Fixing it now

gbalint: Fixed in 05f01159ddffc52b3a9d01283d7d027560861899

PR #2234, opened 2022-05-10
Feature/escape hatch strategy

enidemi:
I'm opening a PR for the escape hatch strategy before making it much larger, it's still a bit half ready but it does the conversion.

After having issues with using the strategy stacking this is a higher order strategy. I created 2 new commands, one sets the absolute position and the other sets the layout property. We only had the adjust number command and set fits better here.

**Included in this PR**
- introduced the strategy
- 2 new commands
- created tests
- the conversion is applied to the element and its siblings

**Not in this PR, it will be in the next**
- allow keeping % width, height values
- trigger logic to convert when reaching the parent bounds (and maintaining the state)
- timer
- controls


balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.tsx
```diff
@@ -0,0 +1,134 @@
...
+  )
+}
+
+function collectSiblingCommands(
+  selectedElements: Array<ElementPath>,
+  metadata: ElementInstanceMetadataMap,
+): Array<CanvasCommand> {
+  const siblings = selectedElements
+    .flatMap((path) => {
+      return [...MetadataUtils.getSiblings(metadata, path).map((element) => element.elementPath)]
```
seanparsons, line 82: Isn't the spread into the array redundant?

seanparsons (approved):

PR #2236, opened 2022-05-11
Bump React 18

balazsbajorics:
## Bumping to React 18
This PR updates our React dependency to 18.1.0. There shouldn't be any changes to behavior, this PR runs React as if it were React 17.

These were the main breaking changes:
- With the React 18 types, props.children are not implicit anymore. Library maintainers and also us need to specify if a component takes children or not. (btw This is a great change for Utopia, as this will be a useful data point for us in the future). So I had to go through a _bunch_ of our code and include props.children types.
- I had to patch some of our dependencies to include the children prop. I decided NOT to bump the version of these dependencies in this PR, to avoid mayhem.
- With the React 18 types, React.useCallback is not allowed to be accidentally `any`, this also broke a few things
- I also updated testing-library/react to the new version that's compatible with React 18. 
- testing-library/react-hooks have been merged into testing-library/react, but with a slightly changed API
- In all the places I use testing library with `{ legacyRoot: true }` which uses React 18 in "React 17 mode" by calling ReactDOM.render instead of the new API
- I bumped react-three-fiber and three to make them not crash violently.

Edit:
- There's a bunch of places where CSSTypes from Emotion is clashing with React's. I couldn't find a way to make both happy so I just gave up and in places where we spread props.style, I cast it to `any`


gbalint:
> With the React 18 types, props.children are not implicit anymore. Library maintainers and also us need to specify if a component takes children or not.

Great news for reparenting!

gbalint:
I think this is a good time to update the React 16 in the initial project's packages.json

gbalint:
I think this is a good time to update the React 16 in the initial project's packages.json

balazsbajorics:
> I think this is a good time to update the React 16 in the initial project's packages.json

I'll create a follow-up ticket

gbalint (approved):

seanparsons (approved):

PR #2237, opened 2022-05-11
Initial add of flex-resize height-width strategy

Seanmclem:
Needed Flex resize strategies, and a good use-case to test the strategy picker

added a strategy directly based on absolute resize delta, using the AbsoluteResizeControl. More customizations will be made specific to flex 

If there is a height css property, it will let you adjust it, if there is a width one -it will let you adjust that one, or both. 

flex


editor/src/components/canvas/canvas-strategies/flex-resize-width-height-strategy.tsx
```diff
@@ -0,0 +1,146 @@
...
+  isApplicable: (canvasState, interactionState, metadata) => {
+    if (
+      canvasState.selectedElements.length === 1 &&
+      !interactionState?.interactionData.modifiers.alt &&
+      !interactionState?.interactionData.modifiers.shift // shift is aspect ratio locked resize implemented in absolute-resize-bounding-box-strategy.tsx
+    ) {
+      return canvasState.selectedElements.every((element) => {
+        const elementMetadata = MetadataUtils.findElementByElementPath(metadata, element)
+
+        return elementMetadata?.specialSizeMeasurements.position !== 'absolute'
```
enidemi, line 32: I think here you should also check if the parent has flex layout, the `specialSizeMeasurements` contains data for this

editor/src/components/canvas/canvas-strategies/flex-resize-width-height-strategy.tsx
```diff
@@ -0,0 +1,146 @@
...
+  },
+  apply: (canvasState, interactionState, sessionState) => {
+    if (
+      interactionState.interactionData.type === 'DRAG' &&
+      interactionState.interactionData.drag != null &&
+      interactionState.activeControl.type === 'RESIZE_HANDLE'
+    ) {
+      const drag = interactionState.interactionData.drag
+      const edgePosition = interactionState.activeControl.edgePosition
+      const { snappedDragVector, guidelinesWithSnappingVector } = snapDrag(
```
enidemi, line 60: I think this snapping logic could be removed for now because flex snapping will be different from the absolute snapping

editor/src/components/canvas/canvas-strategies/flex-resize-width-height-strategy.tsx
Seanmclem: @enidemi so, the parent would need to be `display: flex;`?

editor/src/components/canvas/canvas-strategies/flex-resize-width-height-strategy.tsx
```diff
@@ -0,0 +1,146 @@
...
+} from './shared-absolute-resize-strategy-helpers'
+import { AbsoluteResizeControl } from '../controls/select-mode/absolute-resize-control'
+
+export const flexResizeWidthStrategy: CanvasStrategy = {
+  id: 'FLEX_RESIZE_WIDTH_HEIGHT',
+  name: 'Flex Resize (Width/Height)',
+  isApplicable: (canvasState, interactionState, metadata) => {
+    if (
+      canvasState.selectedElements.length === 1 &&
+      !interactionState?.interactionData.modifiers.alt &&
```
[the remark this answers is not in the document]
Seanmclem: ok I think I got them

editor/src/components/canvas/canvas-strategies/flex-resize-width-height-strategy.tsx
Seanmclem: Ok I think removed successfully

editor/src/components/canvas/canvas-strategies/flex-resize-width-height-strategy.tsx
enidemi: yes! it's called parentLayoutSystem

PR #2238, opened 2022-05-11
Feature/escape hatch strategy 2

enidemi:
Continuing #2234 

**Fix:**
The escape hatch strategy can handle if the element width and height values are set in % it keeps it in %, it also makes sure to set 2 frame pins at least on each side. It's possible that the element has bottom/right style props now we prefer to keep and update those.

**Commit Details:**
- extended the set css length command to support % values, for this the parent size is needed
- updated the escape hatch strategy to find the matching frame pins
- fixed a problem with `parseCSSNumber`, inside the command I'm using parseCSSPercent to check if the value is a number or a percent but `parseCSSPercent(10)` just returned right, which seems incorrect to me


editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.tsx
```diff
@@ -65,70 +75,134 @@ export const escapeHatchStrategy: CanvasStrategy = {
...
+    if (horizontalProps.length === 0) {
+      pinsToSet.push('left', 'width')
+    } else if (horizontalProps.length === 1) {
+      if (horizontalProps[0] !== 'width') {
+        pinsToSet.push(...horizontalProps, 'width')
+      } else {
+        pinsToSet.push('left', 'width')
+      }
+    } else {
+      pinsToSet.push(...horizontalProps)
```
Rheeseyb, line 179: How does this behave if more than 2 props are set in each direction? Does it update them all correctly, and then leave it to the browser to determine which prop to ignore?

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.tsx
enidemi: For more than 2 props set per direction all of them should be updated, I need to cover this part with more tests.

seanparsons (approved):

PR #2239, opened 2022-05-13
Update React DnD so that the navigator doesn't have some weird wrapper elements.

seanparsons:
**Problem:**
Currently if the navigator model changes slightly (see the `Edit Component` context menu option), the entire navigator gets re-rendered, which is caused by the old style React DnD class components.

**Fix:**
The React DnD dependencies were bumped to the latest version which removes the old class components anyway.

`NavigatorItemContainer` is where the work relates to primarily which used the old class components for drag and drop and now uses the `useDrag` and `useDrop` hooks. Parameters for the `onDrop` and `onHover` utility functions were renamed for clarity, especially because it's sometimes hard to know which item React DnD is referring to.

Because of a dubious approach to deprecation involving the use of invariant violations, the file items needed a surprising amount of refactoring which mostly involved switching the handling of the drop itself from the `useDrag` to the `useDrop` hook.

Bonus Point: Jest appears completely incapable of handling the new React DnD code so I had to mock out the two dependencies so that the tests could run.

**Commit Details:**
- Bumped `react-dnd` and `react-dnd-html5-backend` to `16.0.0`.
- Added `DndProvider` around the canvas controls because the global provider
  wasn't being reached somehow.
- `NAVIGATOR_REORDER` action now bumps `domWalkerInvalidateCount` so that the
  dom walker is re-run for the entire canvas.
- Reworked the hooks in `FileBrowserItem` as a result of the `begin` property
  being deprecated, but that deprecation triggering an invariant violation
  which crashes the entire editor. Moved the triggering of the change from
  the drag `end` field to the drop `drop` field.
- Removed a lot of the content from `drag-and-drop-utils.ts` as it is now redundant.
- Removed old React DnD class from the navigator items and replaced it with the
  `useDrag` and `useDrop` hooks.
- Renamed the parameters in `onDrop` and `onHover` to make it clearer what was
  being passed.
- Slight tweak to creating the `visibleNavigatorTargets` to eliminate the need
  to create and then explode an object.
- Mocked `react-dnd` and `react-dnd-html5-backend` for the jest tests otherwise
  jest implodes.


editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -64,7 +64,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`343`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`344`)
```
Rheeseyb, line 67: I wonder how accurate this is with the mocked ReactDnD we're having to use with jest 😬

Rheeseyb: The puppeteer tests don't appear to have deviated so I think we're good

Rheeseyb (approved):

enidemi (approved):

PR #2246, opened 2022-05-17
Feature/escape hatch timer

enidemi:
**Fix:**
This PR extends the escape hatch strategy with the timer, uses the strategy custom state to store if the timer is finished running or not and shows the timer animation on the canvas.
It also includes changes to the fitness function to trigger the escape hatch for the following scenarios:
- element is flex child without siblings
- when the element is flex child with siblings the cursor needs to be outside of the parents frame
- element is a flow element

**Commit Details:**
- the globalTime is added to dragInteractionData
- canvas action to update the dragInteractionData with the new time
- 2 new canvas controls: the timer animation and the dragged frame outline
- changes in the escape hatch strategy

 TODO in the next PR:
- more tests, especially about how the different fitness functions work together


editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -154,7 +157,9 @@ export function updateInteractionViaMouse(
       startedAt: currentState.startedAt,
     }
   } else {
-    return currentState
+    return {
```
balazsbajorics, line 160: is this shallow cloning needed here?

editor/src/components/editor/store/dispatch-strategies.spec.tsx
```diff
@@ -1,3 +1,5 @@
+jest.spyOn(Date, 'now').mockReturnValue(new Date(1000).getTime())
```
balazsbajorics, line 1: does this interfere with other tests? should this go in a `beforeAll` block (with a corresponding cleanup in `afterAll`)?

balazsbajorics: alternatively, instead of using Date.now() directly in the strategy code, you could make a strategy specific function that gets the date, and mock that. which means all the other places in the codebase that use Date.now() wouldn't be affected by this mock!

editor/src/components/canvas/controls/select-mode/pie-timer.tsx
```diff
@@ -0,0 +1,190 @@
...
+          elementRef.current.style.setProperty('display', 'none')
+        }
+      }
+    },
+    [elementRef],
+  )
+
+  useSelectorWithCallback((store) => {
+    if (store.editor.canvas.interactionSession?.interactionData.type === 'DRAG') {
+      if (store.editor.canvas.interactionSession?.interactionData.type === 'DRAG') {
```
seanparsons, line 167: These are the same two lines right?

seanparsons (approved):

balazsbajorics (approved):

PR #2249, opened 2022-05-18
Feature/escape hatch fitness tests

enidemi:
**Problem:**
There are no tests about which strategy will be activate for different layout types.

**Fix:**
I made tests for the escape hatch and flex reorder strategies with 'findCanvasStrategy', the goal is to find which strategy wins the fitness check. It's important to use realistic metadata as there are lots of details about the layout in the specialSizeMeasurements.

**Commit Details:**
- added tests


seanparsons (approved):

balazsbajorics (commented):
:+1:

PR #2255, opened 2022-05-20
Performance/canvas selective render

balazsbajorics:
**Problem:**
Whenever EditorState.projectContents changes, the entire canvas re-renders, including all of the user elements, every single time. For larger projects this kills the performance of interactions. Since we aim for a ~60 FPS mouse interaction performance, we need som way to only re-render the parts of the Canvas that actually change during an interaction.

**Fix:**
Standard React memoization of the ComponentRendererComponents is not applicable for two reasons: 
1. it would alter the user's code in a way that might surprise them, especially (only?) if the user code is not yet compatible with react 18 strict mode and expects a re-render in situations where React.Memo would prevent it.
2. More importantly, ComponentRendererComponents use ContextAtoms to access the ProjectContents to be able to update themselves when the underlying projectContext model for the target component changes. 

**TODO**
- [x] ⚠️ dragging Card component in the example project is broken
- [x] ❓ we only update the manipulated instance of a component, and the rest of the instances only update after the drag is released. Is this desirable? If we want all instances to re-render, I should create a useContextAtomSelector for ContextAtoms instead of my current solution.
- [x] 🐛 There seems to be a problem with focusing on components

**Commit Details:**
- added a new `ElementsToRerender` array to editor state
- `usePubSubAtomReadOnly` takes a callback function which decides if the hook's user component should be force-updated when the atom value changes. 
- I use the shouldUpdate callback in ComponentRendererComponent to prevent updates for components which are not on the `ElementsToRerender` list
- ComponentRendererComponent also calls `updateInvalidatedPaths` now (not sure if we need this?)
- new `SetElementsToRerenderCommand` changes the `ElementsToRerender` list
- using `SetElementsToRerenderCommand` in all strategies
- made a for-tests-only action `SET_ELEMENTS_TO_RERENDER` to fix some dom-walker tests


editor/src/core/shared/element-path.ts
```diff
@@ -271,6 +271,10 @@ export function parentPath(path: ElementPath): ElementPath {
   }
 }
 
+export function isParentComponentOf(maybeParent: ElementPath, maybeChild: ElementPath): boolean {
```
balazsbajorics, line 274: I don't remember why I made this function. should I just use `isParentOf`?

editor/src/components/canvas/canvas-strategies/keyboard-absolute-resize-strategy.tsx
```diff
@@ -52,45 +56,46 @@ export const keyboardAbsoluteResizeStrategy: CanvasStrategy = {
...
-            if (key == null) {
-              return []
-            }
-            const drag = getDragDeltaFromKey(key, interactionState.interactionData.modifiers)
-            const edgePosition = getEdgePositionFromKey(key)
-            if (drag.x !== 0 || drag.y !== 0) {
-              return canvasState.selectedElements.flatMap((selectedElement) => {
-                const element = withUnderlyingTarget(
+        commands: interactionState.interactionData.keysPressed.flatMap<
+          AdjustCssLengthProperty | SetElementsToRerenderCommand
```
seanparsons, line 60: Is this type really necessary?

balazsbajorics: not really. I removed it in bb5138a9

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -90,23 +92,40 @@ export function createComponentRendererComponent(params: {
 
     const mutableContext = params.mutableContextRef.current[params.filePath].mutableContext
 
-    const { topLevelElements, imports } = useGetTopLevelElementsAndImports(params.filePath)
-    const { code, highlightBounds } = useGetCodeAndHighlightBounds(params.filePath)
+    const shouldUpdate = () =>
+      !isFeatureEnabled('Canvas Selective Rerender') ||
+      ElementsToRerenderGLOBAL.current === 'rerender-all-elements' ||
+      ElementsToRerenderGLOBAL.current.findIndex((er) => {
```
seanparsons, line 98: Wouldn't `ElementsToRerenderGLOBAL.current.some(...` make more sense?

editor/src/core/shared/element-path.ts
seanparsons: It _looks_ equivalent to `isParentOf`.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
balazsbajorics: fixed in 1fd1b4c

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -90,23 +92,43 @@ export function createComponentRendererComponent(params: {
 
     const mutableContext = params.mutableContextRef.current[params.filePath].mutableContext
 
-    const { topLevelElements, imports } = useGetTopLevelElementsAndImports(params.filePath)
-    const { code, highlightBounds } = useGetCodeAndHighlightBounds(params.filePath)
+    const instancePath: ElementPath | null = tryToGetInstancePath(instancePathAny, pathsString)
+
+    const shouldUpdate = () =>
```
seanparsons, line 97: Can this be declared as a `function` without costing us anything?

seanparsons (approved):

gbalint (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
balazsbajorics: yep, pushed

PR #2257, opened 2022-05-20
feature(escape-hatch) use some of the absolute move strategy commands for snapping and guidelines

enidemi:
**Problem:**
After converting to absolute with the escape hatch the guidelines and snap points are missing.

**Fix:**
I can't directly use all the absolute move commands because escape hatch also creates these properties. Adding layout props and applying the drag is still part of the escape hatch. 

**Commit Details:**
- extending escape hatch with snapped drag delta and guidelines
- made common functions for escape hatch and absolute move strategies

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
+++ b/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
@@ -1,13 +1,23 @@
 import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import { ElementInstanceMetadataMap } from '../../../core/shared/element-template'
-import { CanvasPoint, offsetPoint } from '../../../core/shared/math-utils'
+import {
+  CanvasPoint,
+  offsetPoint,
+  zeroCanvasPoint,
+  zeroPoint,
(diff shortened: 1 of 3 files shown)


balazsbajorics (approved):

balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.ts
```diff
@@ -46,41 +52,64 @@ export const absoluteMoveStrategy: CanvasStrategy = {
...
     return emptyStrategyApplicationResult
   },
 }
 
+export function applyAbsoluteMoveCommon(
+  canvasState: InteractionCanvasState,
+  interactionState: InteractionSession,
+  strategyState: StrategyState,
+  getMoveCommands: (snappedDragVector: CanvasPoint) => Array<CanvasCommand>,
+) {
```
seanparsons, line 83: Would it be possible to have a return type for the function added please?

seanparsons (approved):

Rheeseyb (approved):

PR #2265, opened 2022-05-23
Only patch React in the Canvas

Rheeseyb:
**Problem:**

As our React monkey patch is applied to React globally, it was affecting the entire editor. This was causing problems when trying to fix up dynamic path creation in `React.memo` and `React.forwardRef` components, completely breaking the editor and hundreds of tests (I'm not entirely sure of the exact cause of that, but I have verified that this fixes that issue in the `fix/arbitrary-js-paths` branch).

**Fix:**

Rather than apply the patch to the imported version of React, we instead create a patched version that is then provided to the Canvas as part of the builtins. Then it was a case of making sure that patched version is used correctly in the tests (which were previously explicitly applying the monkey patch as part of their setup).

### DIFF
--- a/editor/src/utils/canvas-react-utils.ts
+++ b/editor/src/utils/canvas-react-utils.ts
@@ -315,8 +315,7 @@ export function isHooksErrorMessage(message: string): boolean {
   )
 }
 
-export const PatchedReact: typeof React & { isPatched: boolean } = {
+export const PatchedReact: typeof React = {
   ...React,
   createElement: patchedCreateReactElement,
-  isPatched: true,
 }


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import { PatchedReact as React } from '../../../utils/canvas-react-utils'
```
balazsbajorics, line 1: hah!

enidemi (approved):

balazsbajorics (approved):

Rheeseyb:
> > completely breaking the editor and hundreds of tests
> 
> wasn't the editor breakage indicative of some bug or oversight in the new monkey function?

I'm not overly convinced, but at the very least this helps narrow it down if it does

PR #2266, opened 2022-05-23
fix(components) disable focusing on utopia api components

enidemi:
**Problem:**
Double clicking on utopia-api components isolates them to component editing. 

**Fix:**
Changing `isFocusableComponentFromMetadata` to check if the element is imported from utopia-api

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -1258,6 +1258,11 @@ export const MetadataUtils = {
     if (isImported) {
       return false
     }
+    const isImportedFromUtopiaApi =
+      element != null ? isUtopiaAPIComponentFromMetadata(element) : false
+    if (isImportedFromUtopiaApi) {
+      return false
+    }
     const isComponent = elementName != null && !isIntrinsicElement(elementName)
     if (isComponent) {
       return true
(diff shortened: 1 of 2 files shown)


Rheeseyb (approved):

balazsbajorics (commented):
shouldn't we disable focus if the component is imported from any node_module?

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1258,6 +1258,11 @@ export const MetadataUtils = {
     if (isImported) {
       return false
     }
+    const isImportedFromUtopiaApi =
+      element != null ? isUtopiaAPIComponentFromMetadata(element) : false
```
gbalint, line 1262: Maybe it is just me, but for me `element == null || isUtopiaAPIComponentFromMetadata(element)` is way easier to process than this ternary.
Maybe because there are just too many negations in the original?

gbalint (approved):

maltenuhn:
> shouldn't we disable focus if the component is imported from any node_module?

Agreed - I thought that was the behaviour we settled on originally and it regressed (do we know why?)

enidemi:
> > shouldn't we disable focus if the component is imported from any node_module?
> 
> Agreed - I thought that was the behaviour we settled on originally and it regressed (do we know why?)

I know when it regressed, for imported components we show the inspector label about information about the package. The utopia-api is a bit special so we decided to not treat it as a real third party component and hide the npm package info.

seanparsons (approved):

PR #2273, opened 2022-05-24
`DerivedState` cleanup

seanparsons:
**Problem:**
`DerivedState` still potentially contains some cruft and given that we run it pretty regularly would benefit from any optimisation.

**Fix:**
Simplified the structure of `DerivedState` slightly and removed a field that wasn't being used by anything.

Made a small change to `createOrderedElementPathsFromElements` that results in one less array and loop of the array from needing to be created and run.

**Commit Details:**
- Flattened out `DerivedState` a tiny bit by removing the `canvas` structure within.
- `NewCanvasControlsInnerProps` now takes a property containing `TransientCanvasState`
  instead of `DerivedState`.
- Removed `descendantsOfHiddenInstances` from `DerivedState`.
- `createOrderedElementPathsFromElements` now creates and populates one less array.

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -753,22 +753,19 @@ export const MetadataUtils = {
           }
 
           const isCollapsed = EP.containsPath(path, collapsedViews)
+          const newCollapsedAncestor = collapsedAncestor || isCollapsed
 
-          let children: Array<ElementPathTree> = []
           let unfurledComponents: Array<ElementPathTree> = []
           fastForEach(subTree.children, (child) => {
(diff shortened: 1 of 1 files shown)


editor/src/components/editor/store/editor-state.ts
```diff
@@ -1876,15 +1870,12 @@ export function deriveState(
   const derived: DerivedState = {
     navigatorTargets: navigatorTargets,
     visibleNavigatorTargets: visibleNavigatorTargets,
-    canvas: {
-      descendantsOfHiddenInstances: editor.hiddenInstances, // FIXME This has been dead for like ever
```
Rheeseyb, line 1880: Fixed 😀

balazsbajorics (approved):

PR #2275, opened 2022-05-24
Fix useDomWalkerInvalidateCallbacks

gbalint:
useDomWalkerInvalidateCallbacks did not do anything, which broke the react-three-fiber test project, which draws content on canvas.

I fixed this and removed the unnecessary `invalidate` parameter.

### DIFF
--- a/editor/src/core/performance/performance-regression-tests.spec.tsx
+++ b/editor/src/core/performance/performance-regression-tests.spec.tsx
@@ -64,7 +64,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`721`)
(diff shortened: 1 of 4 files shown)


balazsbajorics:
and my canvas, you mean [Canvas](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API) :)

balazsbajorics (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -64,7 +64,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`721`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`422`)
```
seanparsons, line 67: That's a dramatic drop.

seanparsons (approved):

PR #2278, opened 2022-05-24
More Accurate Dynamic Path Building

### LINKED ISSUE #2059: wrapping a component in React.memo breaks the spy/dom-walker/uid logic, making the element (and its children) unselectable

Rheeseyb:
Fixes #2059 

**Problem:**
We build paths from the UIDs of elements to support selection and focusing of rendered elements, but those paths were built statically based on the parsed model, meaning that elements and components inside unparsed code or with any form of indirection (e.g. assigning to a different variable) either had incorrect or completely missing paths.

**Fix:**
The solution to this was to build the paths dynamically when rendering the elements, via the `createElement` patch in `canvas-react-utils.ts`. As the element paths are used to distinguish the relationship between a component and its root element (using a `:` separator) from the relationship between a parent element and its child (using a `/` separator), this logic has to be handled inside the various `mangle*Type` functions.

There are various pieces of special cased logic outlined in comments, but to capture them here:
- Intrinsic elements cannot be replaced with e.g. a functional component, as that can break user code (which was demonstrated with ReactDND which explicitly required intrinsic elements), so we can't directly mangle those, but instead must build the path as part of the `createElement` rather than the render function.
- `React.memo` stores the actual component in a `type` field, so to mangle those components we have to also mangle the component in that `type` field.
- `React.forwardRef` similarly stores the actual component in a `render` field, though I've had to pause that work as there are still issues with it (hence the check for forwardRef components being commented out)
- As exotic components themselves don't exist in the DOM, we have to explicitly pass on their paths to their children, whilst still honouring the above mentioned "component / root element" vs "parent / child" relationship.
- Class component mangling uses a very dirty `this.props` replacement to set the path. React does not like this. I don't have a better idea.

**Important:**
For some reason there were a bunch of cases where the patched version `createElement` wasn't being called when rendering elements in the canvas as a result of #2265. This isn't causing any issues prior to this PR as that only happens when rendering a component that Utopia has wrapped (and was therefore providing a `data-path` to separately), however the introduction of paths built via the patch caused edge cases there. I was unable to narrow down the exact cause of those issues, so as part of this PR I have restored the previous `applyUIDMonkeyPatch` flow. I can separate that out into a different PR if required.

**Notes:**
This PR makes it possible to select and adjust elements that were previously not selectable. It does not allow you to focus such elements, as that logic is elsewhere. That will now be possible thanks to the correct paths, but this work has run on for far too long, and I'm not sure how long the followup work to support that would be.

Also, we're still using `data-utopia-instance-path` for some things, and the `data-utopia-scene-id` is still being built using the old (often incorrect) logic. Nothing has regressed there, I have just left that behaviour as is so that this work can be merged.


seanparsons (approved):

enidemi (approved):

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
```diff
@@ -5762,13 +5762,13 @@ exports[`UiJsxCanvas render class component is available from arbitrary block in
       >
         <div
           data-uid=\\"ccc-unparsed-no-template-path\\"
-          data-path=\\"utopia-storyboard-uid/scene-aaa/app-entity:zzz/aaa\\"
+          data-path=\\"utopia-storyboard-uid/scene-aaa/app-entity:zzz/aaa:ccc-unparsed-no-template-path\\"
```
balazsbajorics, line 5765: I guess it's not unparsed-no-template-path anymore 🍾

Rheeseyb:
I've agreed with Balazs that I'm going to separate the reversion of #2265 into a separate PR and merge that first

editor/src/utils/canvas-react-utils.ts
```diff
@@ -119,14 +270,20 @@ const mangleFunctionType = Utils.memoize(
         if (MeasureRenderTimes) {
           performance.mark(`render_start_${uuid}`)
         }
-        let originalTypeResponse = (
+
+        const path = p?.[UTOPIA_PATH_KEY] ?? p?.[UTOPIA_UID_KEY]
```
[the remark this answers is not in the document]
Rheeseyb: Done

editor/src/components/canvas/ui-jsx-canvas-bugs.spec.tsx
```diff
@@ -241,6 +247,7 @@ export default function App(props) {
           data-utopia-root-element-path=\\"storyboard-entity\\"
         >
           <div
+            data-label=\\"Imported App\\"
```
balazsbajorics, line 250: a wild data-label appears?!

PR #2279, opened 2022-05-24
When running the performance test scripts in local dev, only iterate 5 times

balazsbajorics:
**Problem:**
I run into this problem 4 times a day: I click on one of the performance test buttons, only to realize I forgot to manually change the iteration count from 100 to something low. And now the test will run for around a minute locking up the editor and I have to restart everything and I'm sad.

**Fix:**
The editor loads with `NumberOfIterations` set to 5.
But it also exposes a function called `SetPerformanceScriptNumberOfIterations` on `window`. Puppeteer then uses the `page.evaluate()` API to call `SetPerformanceScriptNumberOfIterations(100)`

Result: for development purposes, we'll se 5 iterations everywhere (webpack, vite, utopia.app etc), and Puppeteer can easily switch to 100 iterations without having to rely on hacks.

### DIFF
--- a/editor/src/core/model/performance-scripts.tsx
+++ b/editor/src/core/model/performance-scripts.tsx
@@ -52,11 +52,11 @@ import { VSCodeLoadingScreenID } from '../../components/code-editor/vscode-edito
 import { v4 as UUID } from 'uuid'
 import { SmallSingleDivProjectContents } from '../../test-cases/simple-single-div-project'
 
-const NumberOfIterations = { current: 5 }
+let NumberOfIterations = 5
 if (window != null) {
   // we are exposing this function on window so it can be called from Puppeteer
   ;(window as any).SetPerformanceScriptNumberOfIterations = (value: number) => {
-    NumberOfIterations.current = value
+    NumberOfIterations = value
     return NumberOfIterations
   }
 }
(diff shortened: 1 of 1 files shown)


editor/src/core/model/performance-scripts.tsx
```diff
@@ -52,14 +52,21 @@ import { VSCodeLoadingScreenID } from '../../components/code-editor/vscode-edito
 import { v4 as UUID } from 'uuid'
 import { SmallSingleDivProjectContents } from '../../test-cases/simple-single-div-project'
 
+const NumberOfIterations = { current: 5 }
```
Rheeseyb, line 55:
```suggestion
let NumberOfIterations = 5
```
😄

puppeteer-tests/src/performance-test.ts
```diff
@@ -119,6 +119,15 @@ async function retryPageCalls<T>(
     await page.waitForXPath(`//div[contains(@id, "canvas-container")]`)
     await page.waitForXPath('//div[contains(@class, "item-label-container")]')
     try {
+      const iterations = await page.evaluate(() => {
+        try {
+          // @ts-expect-error – page.evaluate remotely calls this function in the running editor!!
+          SetPerformanceScriptNumberOfIterations(100)
+        } catch (e) {
+          // SetPerformanceScriptNumberOfIterations was not on window
```
Rheeseyb, line 127: After this PR is merged, should we stop squashing the error?

balazsbajorics: yep

PR #2282, opened 2022-05-26
Move props out of metatadata.

seanparsons:
**Problem:**
The spy props have long been effectively embedded inside `ElementInstanceMetadata` which leads to a lot of issues with them changing and forcing a lot of the model to need updating as well as those needing introspective deep equality which is slow to use on lots of small cases.

**Fix:**
Removed the `ElementInstanceMetadata.props` property and put the props into one dictionary style object at the root of `EditorState`.

**Commit Details:**
- `props` removed from `ElementInstanceMetadata`.
- Added `ElementProps` type for the props of an element.
- Added `AllElementProps` type for the props of the entire project.
- Added `EditorState.allElementProps` property.
- Updated the `EditorState` and `ElementInstanceMetadata` deep
  equality functions.
- Added an `allElementProps` parameter to `CanvasStrategy.isApplicable`.
- In numerous functions either added an `AllElementProps` parameter
  alongside an `ElementInstanceMetadataMap` parameter or in a few cases
  replaced it.
- In `property-control-hooks.spec.tsx` fixed a bug with the wrong
  path being assigned to the metadata of an element.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
@@ -50,7 +50,7 @@ export const absoluteResizeBoundingBoxStrategy: CanvasStrategy = {
         const elementProps = allElementProps[EP.toString(element)] ?? {}
         return (
           elementMetadata?.specialSizeMeasurements.position === 'absolute' &&
-          hasAtLeastTwoPinsPerSide(elementProps)
+          hasAtLeastTwoPinsPerSide(elementProps) // TODO should this use projectContents?
         )
       })
     } else {
--- a/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
@@ -12,7 +12,7 @@ import {
   transformFrameUsingBoundingBox,
 } from '../../../core/shared/math-utils'
 import { ElementPath } from '../../../core/shared/project-file-types'
-import { getElementFromProjectContents } from '../../editor/store/editor-state'
+import { AllElementProps, getElementFromProjectContents } from '../../editor/store/editor-state'
 import { stylePropPathMappingFn } from '../../inspector/common/property-path-hooks'
 import { EdgePosition } from '../canvas-types'
 import {
@@ -31,11 +31,12 @@ import {
   resizeBoundingBox,
   runLegacyAbsoluteResizeSnapping,
 } from './shared-absolute-resize-strategy-helpers'
+import * as EP from '../../../core/shared/element-path'
 
 export const absoluteResizeBoundingBoxStrategy: CanvasStrategy = {
   id: 'ABSOLUTE_RESIZE_BOUNDING_BOX',
   name: 'Absolute Resize',
-  isApplicable: (canvasState, interactionState, metadata) => {
+  isApplicable: (canvasState, interactionState, metadata, allElementProps) => {
     if (
       canvasState.selectedElements.length > 1 ||
       (canvasState.selectedElements.length >= 1 &&
@@ -44,9 +45,10 @@ export const absoluteResizeBoundingBoxStrategy: CanvasStrategy = {
     ) {
       return canvasState.selectedElements.every((element) => {
         const elementMetadata = MetadataUtils.findElementByElementPath(metadata, element)
+        const elementProps = allElementProps[EP.toString(element)] ?? {}
         return (
           elementMetadata?.specialSizeMeasurements.position === 'absolute' &&
-          hasAtLeastTwoPinsPerSide(elementMetadata.props)
+          hasAtLeastTwoPinsPerSide(elementProps)
         )
       })
     } else {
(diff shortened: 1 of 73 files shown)


editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
```diff
@@ -46,9 +47,10 @@ export const absoluteResizeBoundingBoxStrategy: CanvasStrategy = {
     ) {
       return canvasState.selectedElements.every((element) => {
         const elementMetadata = MetadataUtils.findElementByElementPath(metadata, element)
+        const elementProps = allElementProps[EP.toString(element)] ?? {}
         return (
           elementMetadata?.specialSizeMeasurements.position === 'absolute' &&
-          hasAtLeastTwoPinsPerSide(elementMetadata.props)
+          hasAtLeastTwoPinsPerSide(elementProps)
```
balazsbajorics, line 53: I think it's worth adding as a comment – I remember that using the spied props was a path of least resistance here, but I _think_ using the actual JSXElement would be more reliable, since we want to understand what are the props we can actually find in code so we can actually change them. 

```suggestion
          hasAtLeastTwoPinsPerSide(elementProps) // TODO should this use projectContents?
```

editor/src/components/canvas/controls/layout-parent-control.tsx
```diff
@@ -55,13 +55,14 @@ export const LayoutParentControl = React.memo((): JSX.Element | null => {
         store.editor.jsxMetadata,
         store.editor.selectedViews[0],
       )
+      const elementProps = store.editor.allElementProps[EP.toString(store.editor.selectedViews[0])]
```
Rheeseyb, line 58: Looks like this is using the wrong element, as the above code is using the metadata for the parent element

editor/src/components/canvas/controls/multiselect-resize-control.tsx
```diff
@@ -206,6 +208,8 @@ export class SingleSelectResizeControls extends React.Component<SingleSelectResi
     return this.props.selectedViews.map((view, index) => {
       const target = MetadataUtils.findElementByElementPath(this.props.componentMetadata, view)
       const frame = optionalMap((metadata) => metadata.globalFrame, target)
+      const targetProps =
+        target == null ? {} : this.props.allElementProps[EP.toString(target.elementPath)] ?? {}
```
Rheeseyb, line 212: I'm not sure if this makes a difference here (in theory it shouldn't?), but we're using `view` as the element path in line 209 so we should probably use that here too

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -154,7 +159,7 @@ export function getSelectableViews(
       const scene = MetadataUtils.findElementByElementPath(componentMetadata, path)
       const rootElements = MetadataUtils.getRootViewPaths(componentMetadata, path)
       if (
-        MetadataUtils.isSceneTreatedAsGroup(scene) &&
+        MetadataUtils.isSceneTreatedAsGroup(allElementProps, path) &&
```
balazsbajorics, line 162: AS GROUP! BY GOD! I think this should just be killed here (in a follow up PR)

editor/src/core/layout/layout-utils.spec.browser2.tsx
```diff
@@ -72,16 +73,21 @@ describe('maybeSwitchLayoutProps', () => {
...
+      style: {
+        width: 375,
+        height: 812,
+      },
+    }
+
+    const elementProps: ElementProps = {
+      'data-uid': NewUID,
+    }
+
```
balazsbajorics, line 86: these don't seem to be used later in the file

editor/src/core/tailwind/tailwind-options.tsx
```diff
@@ -298,6 +299,11 @@ export function useGetSelectedClasses(): {
     'ClassNameSelect elementPaths',
   )
 
+  const allElementProps = useEditorState(
+    (store) => store.editor.allElementProps,
```
balazsbajorics, line 303: `useEditorState` runs a shallow equality by default, if allElementProps is large, this will iterate over all the keys. maybe it's worth providing a strict equality check here? I'm not 100% sure about the tradeoff

balazsbajorics (approved):
I left some minor nitpicks, plus what Rheese said

editor/src/components/canvas/controls/layout-parent-control.tsx
seanparsons: Good spot.

editor/src/components/canvas/controls/multiselect-resize-control.tsx
seanparsons: Fixed in 3403d1aa5.

editor/src/core/layout/layout-utils.spec.browser2.tsx
seanparsons: Removed in 18ae1e3e6.

editor/src/core/tailwind/tailwind-options.tsx
seanparsons: That _sounds_ like something we should be doing in general for `useEditorState` if we're doing a deep equality pass?

balazsbajorics (approved):

Rheeseyb (approved):

PR #2284, opened 2022-05-26
Improve canvas valid path handling.

seanparsons:
**Problem:**
The `renderJSXElement` function is transforming the valid paths each and every time it is called.

**Fix:**
Perform the transformation on the valid paths when the scene context that holds them is created.

**Commit Details:**
- Change `validPaths` to be a `Set` instead of an `Array`.
- Apply `makeLastPartOfPathStatic` to the `validPaths` scene
  context value on creation.
- In `renderJSXElement` don't apply `makeLastPartOfPathStatic`
  as it has been done on creation.


balazsbajorics (approved):
nice catch

Rheeseyb (approved):

PR #2291, opened 2022-05-31
Restore previous applyUIDMonkeyPatch flow

Rheeseyb:
**Problem:**
As mentioned in #2278, the removal of the older `applyUIDMonkeyPatch()` flow in #2265 meant there were a bunch of cases where the patched version `createElement` wasn't being called when rendering elements in the canvas, and I've been unable to fully nail down some unknown amount of them. This isn't causing any issues prior to #2278 as that only happens when rendering a component that Utopia has wrapped (and was therefore providing a data-path to separately), but it prevents paths from being correctly built as part of the patched `createElement`.

**Fix:**
This PR simply reverts #2265 (without breaking the render count test)


balazsbajorics (approved):
I think it's worth creating a tracking issue in the backlog, maybe someday we could try this again...

Rheeseyb:
> I think it's worth creating a tracking issue in the backlog, maybe someday we could try this again...

Good shout 👍

PR #2294, opened 2022-05-31
Performance: Selective dom walker

gbalint:
# Optimization

This PR introduces a new dom walker mode: during interactions it is too slow to run the full dom walker in every frame.
Probably it is practically enough to update the metadata for exactly these elements, which are selectively rendered (see https://github.com/concrete-utopia/utopia/pull/2209)

For these elements we run a dom query to find them in the document, and collect their metadata only (we don't traverse the dom. This way the dom walker run time is decreased from 8-9 ms to ~1 ms.

I also update `EditorState.domMetadata` and converted its type to `ElementInstanceMetadataMap` (and not `ElementInstanceMetadata[]`)
This slightly decreased the run time of the `SAVE_DOM_REPORT` action, because the equality check seems to be faster for the map.

# Notes

### What happens if some elements are modified (e.g. flex siblings of a resized element), but not updated by the selective dom walker? 
We'll see, but it should not cause a real problem, because the interaction uses the starting metadata anyway, and the full dom walker runs when the interaction ends, and updates all metadata.

### `invalidatePaths` is problematic

Originally I wanted to use the invalidatedPaths to choose which metadata to update, but that caused a lot of trouble, because it seems the collection of invalidatedPaths is not reliable (see https://github.com/concrete-utopia/utopia/pull/2289)


editor/src/components/canvas/dom-walker.ts
```diff
@@ -287,19 +289,82 @@ interface RunDomWalkerParams {
...
+  selectedViews: Array<ElementPath>,
+  scale: number,
+  rootMetadataInStateRef: { readonly current: readonly ElementInstanceMetadata[] },
+  containerRectLazy: () => CanvasRectangle,
+): { metadata: ElementInstanceMetadata[]; cachedPaths: ElementPath[] } {
+  let workingMetadata: ElementInstanceMetadata[] = []
+
+  const canvasRootContainer = document.getElementById(CanvasContainerID)
+  if (canvasRootContainer != null) {
+    const validPaths: Array<ElementPath> | null = optionalMap(
```
seanparsons, line 310: You might find that putting these into a `Set` and later on using `has` will be quicker.

gbalint: done

seanparsons (approved):

editor/src/components/canvas/dom-walker.ts
```diff
@@ -528,38 +607,16 @@ function collectMetadata(
       )
     })
   if (shouldCollect && pathsForElement.length > 0) {
```
Rheeseyb, line 609: You can remove the `pathsForElement.length > 0` check here now

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1009,7 +1009,7 @@ export const MetadataUtils = {
     // wont make any difference.
     let workingElements: ElementInstanceMetadataMap = { ...fromSpy }
     let newlyFoundElements: Array<ElementPath> = []
-    fastForEach(fromDOM, (domElem) => {
+    fastForEach(Object.values(fromDOM), (domElem) => {
       const spyElem = MetadataUtils.findElementByElementPath(fromSpy, domElem.elementPath)
```
Rheeseyb, line 1013: Now that they're both `ElementInstanceMetadataMap`, they should both be using the same keys (which are the stringified paths), so you should be able to use the keys of `fromDOM` rather than any cases of `EP.toString` (which is what this `findElementByElementPath` will be doing, and is used below to insert into the `workingElements`)

editor/src/components/canvas/dom-walker.ts
gbalint: done

PR #2296, opened 2022-06-01
performance(editor) Assorted tuning, mostly relating to file paths.

seanparsons:
**Problem:**
There were still some places during the early parts of a canvas render which were a little wasteful.

**Fix:**
This has a collection of fixes, the main highlights are:
- Exploiting identity equality for `ElementPath` values because there should be no value which is identical equals but not deeply value equals.
- Using a LRU cache to hold the result of parsing `package.json` files.
- Caching constructed file paths in a structure similar to that which is done currently with `ElementPath`.

**Commit Details:**
- Added `lru-cache` dependency.
- `getContentsTreeFileFromElements` now loops instead of calling a
  recursive function.
- `clearSpyCollectorInvalidPaths` now takes a `Set<ElementPath` instead
  of an `Array<ElementPath>`.
- `useGetStoryboardRoot` returns both an `Array<ElementPath>` and
  a `Set<ElementPath`, as the former is more useful for generating
  a `string` of valid paths.
- Use `Object.assign` in `importResultFromImports` rather than
  spreading itself into the new object.
- `module-resolution.ts` now caches the result of parsing a `package.json`
  file rather than incurring a `JSON.parse` and then our internal parser
  near constantly.
- `normalizePath` now allows for entries containing an `Array<string` as well
  as a `string`, so that arrays need not be spread to be passed into the function.
- Created a couple of common values in places like `resolveNonRelativeModule` so
  as to not create the same array multiple times.
- Make `pathsEqual` for `ElementPath` just do an idendity check as there
  should never be values for which they are not identity equal if they
  are deeply value equals.
- `matchFile` in `default-loader.ts` now obtains the extension and then checks
  if it is in a stored `Set` of extensions.
- `matchFile` in `json-loader.ts` now doesn't create an unnecessary array.
- `makePathFromParts` now caches paths so as to not create the same string repeatedly.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas.tsx
@@ -278,14 +278,15 @@ function useClearSpyMetadataOnRemount(
 }
 
 function clearSpyCollectorInvalidPaths(
-  validPaths: Array<ElementPath>,
+  validPaths: Set<ElementPath>,
   spyCollectorContextRef: UiJsxCanvasContextData,
 ): void {
   const spyKeys = Object.keys(spyCollectorContextRef.current.spyValues.metadata)
   fastForEach(spyKeys, (elementPathString) => {
-    const elementPath = EP.fromString(elementPathString)
+    const elementPath =
+      spyCollectorContextRef.current.spyValues.metadata[elementPathString].elementPath
     const staticElementPath = EP.makeLastPartOfPathStatic(elementPath)
-    if (validPaths.every((validPath) => !EP.pathsEqual(validPath, staticElementPath))) {
+    if (!validPaths.has(staticElementPath)) {
       // we found a path that is no longer valid. let's delete it from the spy accumulator!
       delete spyCollectorContextRef.current.spyValues.metadata[elementPathString]
     }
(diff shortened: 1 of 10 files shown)


editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -278,14 +278,15 @@ function useClearSpyMetadataOnRemount(
...
 
 function clearSpyCollectorInvalidPaths(
-  validPaths: Array<ElementPath>,
+  validPaths: Set<ElementPath>,
   spyCollectorContextRef: UiJsxCanvasContextData,
 ): void {
   const spyKeys = Object.keys(spyCollectorContextRef.current.spyValues.metadata)
   fastForEach(spyKeys, (elementPathString) => {
-    const elementPath = EP.fromString(elementPathString)
+    const elementPath =
```
balazsbajorics, line 286: hah, I wonder how many places we could apply this quick hack

editor/src/core/webpack-loaders/json-loader.ts
```diff
@@ -1,7 +1,7 @@
 import { LoadModule, loadModuleResult, MatchFile, ModuleLoader } from './loader-types'
 
 const matchFile: MatchFile = (filename: string) => {
-  return ['.json'].some((extension) => filename.endsWith(extension))
+  return filename.endsWith('.json')
```
gbalint, line 4: this was seriously overcomplicated ...

gbalint (approved):

balazsbajorics (approved):

PR #2301, opened 2022-06-02
improved devtools logs

balazsbajorics:
Today I extensively used the Redux Devtool to log editor states while debugging. I made an assortment of improvements to it, here they go. I'll comment on the changes inline.

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -464,14 +464,7 @@ export function editorDispatch(
     builtInDependencies: storedState.builtInDependencies,
   }
 
-  if (!finalStore.nothingChanged) {
-    /**
-     * Heads up: we do not log dispatches that resulted in a NO_OP. This is to avoid clogging up the
-     * history with a million CLEAR_HIGHLIGHTED_VIEWS and other such actions.
-     *  */
-
-    reduxDevtoolsSendActions(actionGroupsToProcess, finalStore)
-  }
+  reduxDevtoolsSendActions(actionGroupsToProcess, finalStore)
 
   if (storedState.userState.loginState.type !== result.userState.loginState.type) {
     if (isLoggedIn(result.userState.loginState)) {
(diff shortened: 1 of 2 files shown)


editor/src/components/editor/store/dispatch.tsx
```diff
@@ -464,14 +464,7 @@ export function editorDispatch(
...
 
-  if (!finalStore.nothingChanged) {
-    /**
-     * Heads up: we do not log dispatches that resulted in a NO_OP. This is to avoid clogging up the
-     * history with a million CLEAR_HIGHLIGHTED_VIEWS and other such actions.
-     *  */
-
-    reduxDevtoolsSendActions(actionGroupsToProcess, finalStore)
-  }
+  reduxDevtoolsSendActions(actionGroupsToProcess, finalStore)
```
balazsbajorics, line 467: we no longer need to filter logs for !nothingChanged, because as part of an effort to clean up our dispatches, we no longer dispatch a million unnecessary actions

editor/src/core/shared/redux-devtools.ts
```diff
@@ -58,19 +58,15 @@ maybeDevTools?.subscribe((message) => {
...
-]
-
 let lastDispatchedStore: SanitizedState
 
 const PlaceholderMessage = '<<SANITIZED_FROM_DEVTOOLS>>'
 
-function simplifiedMetadata(elementMetadata: Partial<ElementInstanceMetadata>) {
+function simplifiedMetadata(elementMetadata: ElementInstanceMetadata) {
   return {
+    elementPath: EP.toString(elementMetadata.elementPath),
```
balazsbajorics, line 67: this is still a very small subset of the metadata, if anyone needs more stuff, they can add it here, on demand. but trying to log the full metadata chokes Devtools

editor/src/core/shared/redux-devtools.ts
```diff
@@ -81,12 +77,30 @@ function simplifiedMetadataMap(metadata: ElementInstanceMetadataMap) {
...
+  return {
+    selectedViews: editor.selectedViews.map(EP.toString) as any, // this is easier for human consumption
+    canvas: {
+      mountCount: editor.canvas.mountCount,
+      domWalkerInvalidateCount: editor.canvas.domWalkerInvalidateCount,
+      canvasContentInvalidateCount: editor.canvas.canvasContentInvalidateCount,
+      interactionSession: {
+        metadata: simplifiedMetadataMap(editor.canvas.interactionSession?.metadata ?? {}) as any,
+      },
+    } as Partial<EditorState['canvas']>,
```
seanparsons, line 92:
```suggestion
    } as Partial<EditorStateCanvas>,
```
As there's now a type for that field.

seanparsons (approved):

PR #2302, opened 2022-06-03
do not re-run handleStrategiesInner on SAVE_DOM_REPORT

balazsbajorics:
**Problem:**
For every action dispatched, we run handleStrategies twice: first for the actual action, then again for the SAVE_DOM_REPORT "action" which updates the metadatas.

**Fix:**
When a SAVE_DOM_REPORT is dispathced, always short-circuit handleStrategies and skip running handleStrategiesInner

### DIFF
--- a/editor/src/components/editor/store/dispatch-strategies.tsx
+++ b/editor/src/components/editor/store/dispatch-strategies.tsx
@@ -27,6 +27,7 @@ import {
 } from '../../canvas/canvas-strategies/canvas-strategy-types'
 import { isFeatureEnabled } from '../../../utils/feature-switches'
 import { PERFORMANCE_MARKS_ALLOWED } from '../../../common/env-vars'
+import { saveDOMReport } from '../actions/action-creators'
 
 interface HandleStrategiesResult {
   unpatchedEditorState: EditorState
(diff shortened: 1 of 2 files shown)


gbalint (approved):

enidemi (approved):

enidemi (changes requested):
with this change patchedEditorState will not contain the correct projectContents result for dragging

balazsbajorics:
> with this change patchedEditorState will not contain the correct projectContents result for dragging

shoot!

editor/src/components/editor/store/dispatch-strategies.tsx
```diff
@@ -563,13 +564,67 @@ function handleStrategiesInner(
...
+    // Unfortunately there's FIVE places we have metadata in EditorState
+    const oldPatchedEditorWithNewMetadata: EditorState = {
+      ...storedState.patchedEditor, // the "old" patched editor from the action dispatch that triggered SAVE_DOM_WALKER
+      jsxMetadata: result.unpatchedEditor.jsxMetadata, // the fresh metadata from SAVE_DOM_REPORT
+      domMetadata: result.unpatchedEditor.domMetadata,
+      spyMetadata: result.unpatchedEditor.spyMetadata,
+      canvas: {
+        ...storedState.patchedEditor.canvas,
+        interactionSession:
+          result.unpatchedEditor.canvas.interactionSession == null ||
```
gbalint, line 605: I try to wrap my head around this, but it is not clear for me why do we have to delete the interactionSession from the old patched editor when the new unpatchededitor doesn't have interactionsession.
Shouldn't we just leave it as it is in these cases?

balazsbajorics: that's why I throw the error a few lines above this if the null-ness of the interaction sessions doesn't match

editor/src/components/editor/store/dispatch-strategies.tsx
```diff
@@ -563,13 +564,67 @@ function handleStrategiesInner(
...
+          storedState.patchedEditor.canvas.interactionSession == null
+            ? null
+            : {
+                ...storedState.patchedEditor.canvas.interactionSession, // the "old" patched interactionSession
+                metadata: result.unpatchedEditor.canvas.interactionSession.metadata, // the fresh metadata from SAVE_DOM_REPORT
+              },
+        dragState:
+          result.unpatchedEditor.canvas.dragState == null ||
```
seanparsons, line 613: Could we pull these big fat ternary statements out of this somehow as it's all embedded in this one giant statement?

seanparsons (approved):
It would probably be worth adding tests that cover the behaviour of this even if just as a form of documentation for anyone landing on this code in the future.

balazsbajorics:
> It would probably be worth adding tests that cover the behaviour of this even if just as a form of documentation for anyone landing on this code in the future.

good shout! I added some tests in 4b4ca3a2

balazsbajorics:
I'm merging this one!

PR #2303, opened 2022-06-03
Use ElementInstanceMetadataMap everywhere in dom walker.

gbalint:
Inside the dom walker we use `ElementInstanceMetadata` arrays to access the old and collect the new metadata. However, in EditorState domMetadata is stored as an `ElementInstanceMetadataMap`. So when we run the dom walker and process its results, we need to convert the metadata from array to map and back.

So it makes sense to use maps everywhere in the dom walker, because then
1. We don't need to convert back and forth between arrays and maps
2. Accessing an element metadata is faster, and that is needed when merging the metadata of fragments

Originally all the metadata was collected in an array, and merged at the end. In this PR I continuously merge in the working metadata map, because I can not store multiple metadatas for the same element path in a map (unlike in an array, where it is possible). I wanted to make sure this will be performant with big metadata maps, so I implemented the merging with mutation. This is not absolutely necessary, but I wanted to be on the safe side regarding performance.

According to my measurements the performance is better, full dom walk was 8-9 ms before this PR, and it is around 6 ms after these changes

### DIFF
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -207,7 +207,7 @@ export interface DomWalkerProps {
 // it adds elementMetadata to metadataToMutate in way that it merges fragments with the same elementpath
 function mergeElementMetadataToMapWithFragments_MUTATE(
   metadataToMutate: ElementInstanceMetadataMap,
-  elementMetadata: ElementInstanceMetadata,
+  elementMetadata: Readonly<ElementInstanceMetadata>,
 ): void {
   const pathString = EP.toString(elementMetadata.elementPath)
   const existingMetadata = metadataToMutate[pathString]
@@ -244,7 +244,7 @@ function mergeElementMetadataToMapWithFragments_MUTATE(
 // it merges otherMetadata into metadataToMutate in way that it merges fragments with the same elementpath
 function mergeMetadataMapsWithFragments_MUTATE(
   metadataToMutate: ElementInstanceMetadataMap,
-  otherMetadata: ElementInstanceMetadataMap,
+  otherMetadata: Readonly<ElementInstanceMetadataMap>,
 ): void {
   fastForEach(Object.values(otherMetadata), (elementMetadata) => {
     mergeElementMetadataToMapWithFragments_MUTATE(metadataToMutate, elementMetadata)
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -16,7 +16,6 @@ import {
   StyleAttributeMetadata,
   emptyAttributeMetadatada,
   ElementInstanceMetadataMap,
-  emptyJsxMetadata,
 } from '../../core/shared/element-template'
 import { ElementPath } from '../../core/shared/project-file-types'
 import { getCanvasRectangleFromElement, getDOMAttribute } from '../../core/shared/dom-utils'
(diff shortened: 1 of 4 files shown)


seanparsons (approved):

editor/src/components/canvas/dom-walker.ts
```diff
@@ -349,21 +348,24 @@ function runSelectiveDomWalker(
...
+        mergeMetadataMapsWithFragments_MUTATE(workingMetadata, collectedMetadata)
       }
     })
-    const rootMetadataForOtherElements = rootMetadataInStateRef.current.filter(
-      (m) => !elementsToFocusOn.some((p) => EP.pathsEqual(p, m.elementPath)),
+    const otherElementPaths = Object.keys(rootMetadataInStateRef.current).filter(
+      (path) => !elementsToFocusOn.some((focusPath) => EP.toString(focusPath) === path),
     )
+    const rootMetadataForOtherElements = pick(otherElementPaths, rootMetadataInStateRef.current)
+    mergeMetadataMapsWithFragments_MUTATE(rootMetadataForOtherElements, workingMetadata)
```
Rheeseyb, line 358: I'm very uneasy about this, because we're very close to mutating the metadata stored in the editor state here. The only reason we're _not_ doing that right now is because the mutating functions above are replacing entire values, so please can you add comments to the mutating functions above making it explicitly clear that we should never mutate the metadata objects themselves.

Rheeseyb: Oooh good idea! We could also capture this in a test case for the ultimate in developer protection

gbalint: I don't think a test is necessary, in this PR I modified `ElementInstanceMetadataMap` to contain readonly `ElementInstanceMetadata` values, so if I try to mutate them, I see a compile error. E.g.:

image

See this change:
image

I'm happy to update the comment though.

gbalint: Added comments in 181c4649bef38075ce9f55c5b07a75c57caa8d7a

Rheeseyb: Ah ok yep that works for me! 👍

Rheeseyb (approved):

editor/src/components/canvas/dom-walker.ts
```diff
@@ -1178,14 +1189,14 @@ function walkElements(
       additionalElementsToUpdate,
     )
 
-    rootMetadataAccumulator = [...rootMetadataAccumulator, ...collectedMetadata]
+    mergeMetadataMapsWithFragments_MUTATE(rootMetadataAccumulator, collectedMetadata)
     cachedPathsAccumulator = [...cachedPathsAccumulator, ...cachedPaths]
     return {
       rootMetadata: rootMetadataAccumulator,
-      childPaths: collectedMetadata.map((metadata) => metadata.elementPath), // TODO why not extract childPaths from the metadata?
+      childPaths: Object.values(collectedMetadata).map((metadata) => metadata.elementPath), // TODO why not extract childPaths from the metadata?
```
balazsbajorics, line 1196: is this expensive here? shouldn't `collectMetadata` return a childPaths too?

editor/src/components/canvas/dom-walker.ts
```diff
@@ -212,44 +203,54 @@ export interface DomWalkerProps {
...
-      working[pathString] = merged
-    }
+// For performance reasons this function mutates the first parameter:
+// it merges otherMetadata into metadataToMutate in way that it merges fragments with the same elementpath
+// Note: it only mutates the map itself, it should never mutate the metadata instances inside the map!
+function mergeMetadataMapsWithFragments_MUTATE(
+  metadataToMutate: ElementInstanceMetadataMap,
+  otherMetadata: Readonly<ElementInstanceMetadataMap>,
+): void {
+  fastForEach(Object.values(otherMetadata), (elementMetadata) => {
```
balazsbajorics, line 251: can this Object.values be slow here? genuine question!

editor/src/components/canvas/dom-walker.ts
gbalint: Good idea, see fdf0c24a020642c922d6247af209aa31dad3432a

editor/src/components/canvas/dom-walker.ts
gbalint: Well, I somehow have to iterate through the elements and `Object.values` is still way faster then `Object.keys`

See my benchmark:
image

PR #2306, opened 2022-06-13
Trim outliers from the performance graphs.

seanparsons:
**Problem:**
We see a lot of variance in the performance tests, which can result in those results being very noisy.

**Fix:**
Trim off a percentage of the frame times, so that those outliers are ignored.

**Commit Details:**
- In `getFrameData`, only take the bottom 95% of the frame times.

### DIFF
--- a/puppeteer-tests/src/performance-test.ts
+++ b/puppeteer-tests/src/performance-test.ts
@@ -477,8 +477,8 @@ const getFrameData = (
 
   let frameTimesFixed = frameTimes.map((x) => Number(x.toFixed(1)))
   const sortedFrameTimes = frameTimesFixed.sort((a, b) => a - b)
-  // Trims the top 5% off.
-  let withoutOutliersLength = Math.round(sortedFrameTimes.length * 0.95)
+  // Trims the top 20% off.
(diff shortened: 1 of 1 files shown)


puppeteer-tests/src/performance-test.ts
```diff
@@ -473,20 +472,27 @@ const getFrameData = (
...
   let frameTimesFixed = frameTimes.map((x) => Number(x.toFixed(1)))
   const sortedFrameTimes = frameTimesFixed.sort((a, b) => a - b)
+  // Trims the top 5% off.
+  let withoutOutliersLength = Math.round(sortedFrameTimes.length * 0.95)
+  // Ensure we always have at least one result.
+  if (withoutOutliersLength <= 0) {
+    withoutOutliersLength = 1
+  }
+  const frameTimesWithoutOutliers = sortedFrameTimes.slice(0, withoutOutliersLength)
+  let totalFrameTimes = 0
```
Rheeseyb, line 487: This is definitely not going to produce the right `frameAvg`! But that doesn't matter because we don't even use `frameAvg`, so in the bin with it!

seanparsons: Ah, I've fixed this in 31ff7dbe3, we should probably remove that in a different PR.

PR #2310, opened 2022-06-14
Check drag threshold in all mouse strategies

gbalint:
1. Not all strategies check the drag threshold
2. The drag threshold is stored redundantly in `DragInteractionData`:
- in `dragThresholdPassed`
- `drag` is `null` when the threshold is not passed
3. There are no tests to check that the strategies should not emit commands when the threshold is passed

I removed the `dragThresholdPassed` flag from `DragInteractionData`, fixed the problematic strategies, and added a new test cases for all mouse strategies.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.tsx
@@ -92,6 +92,28 @@ function dragByPixels(
 }
 
 describe('Absolute Move Strategy', () => {
+  it('does not activate when drag treshold is not reached', async () => {
+    const targetElement = elementPath([
+      ['scene-aaa', 'app-entity'],
+      ['aaa', 'bbb'],
+    ])
+
+    const initialEditor: EditorState = prepareEditorState(
+      `
+    <View style={{ ...(props.style || {}) }} data-uid='aaa'>
+      <View
+        style={{ backgroundColor: '#0091FFAA', position: 'absolute', left: 50, top: 50, width: 250, height: 300 }}
+        data-uid='bbb'
+      />
+    </View>
+    `,
+      [targetElement],
+    )
+
+    const finalEditor = dragByPixels(initialEditor, canvasPoint({ x: 1, y: 1 }), emptyModifiers)
+
+    expect(finalEditor).toEqual(finalEditor)
+  })
   it('works with a TL pinned absolute element', async () => {
     const targetElement = elementPath([
       ['scene-aaa', 'app-entity'],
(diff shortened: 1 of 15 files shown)


editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.tsx
```diff
@@ -92,6 +92,28 @@ function dragByPixels(
...
+        data-uid='bbb'
+      />
+    </View>
+    `,
+      [targetElement],
+    )
+
+    const finalEditor = dragByPixels(initialEditor, canvasPoint({ x: 1, y: 1 }), emptyModifiers)
+
+    expect(finalEditor).toEqual(finalEditor)
```
enidemi, line 115: I think you wanted to check this against the initalEditor

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.tsx
```diff
@@ -109,7 +109,10 @@ export const escapeHatchStrategy: CanvasStrategy = {
     }
   },
   apply: (canvasState, interactionState, strategyState) => {
-    if (interactionState.interactionData.type === 'DRAG') {
+    if (
+      interactionState.interactionData.type === 'DRAG' &&
+      interactionState.interactionData.drag != null
```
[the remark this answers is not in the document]
gbalint: fixed in d87a1d89dc49df1f50e7af6b83ab1f1568b4dcb0

editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx
```diff
@@ -174,8 +221,8 @@ describe('Absolute Reparent Strategy', () => {
                 position: 'absolute',
                 width: 20,
                 height: 30,
-                top: 15,
-                left: 40,
+                top: 30,
```
balazsbajorics, line 224: were all these changes intentional?

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.spec.tsx
```diff
@@ -137,16 +137,21 @@ const mixedPinsMetadata = {
   } as ElementInstanceMetadata,
 }
 
-function dragBy15Pixels(
+function dragBy15Pixels(editorState: EditorState, metadata: ElementInstanceMetadataMap) {
```
Rheeseyb, line 140: `dragBy21.2132Pixels`

balazsbajorics: kidding aside, this was a significant readability improvement for these tests

editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx
gbalint: Yes. Until now this test did not really move the object, as it only tested reparenting itself, and that made the test cleaner. However, after this change the strategy does not reparent when drag is zero, so I needed to really drag, and update the coordinates according to that.

balazsbajorics (approved):

PR #2312, opened 2022-06-14
Feature: alt drag to duplicate

balazsbajorics:
**Problem:**

**Fix:**

**Commit Details:**
- Did [other things]

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-duplicate-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-duplicate-strategy.tsx
@@ -1,7 +1,9 @@
 import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import { generateUidWithExistingComponents } from '../../../core/model/element-template-utils'
 import * as EP from '../../../core/shared/element-path'
-import { duplicateElement } from '../commands/duplicate-element-command'
+import { ElementInstanceMetadataMap } from '../../../core/shared/element-template'
+import { ElementPath } from '../../../core/shared/project-file-types'
+import { DuplicateElement, duplicateElement } from '../commands/duplicate-element-command'
 import { updateSelectedViews } from '../commands/update-selected-views-command'
 import { ParentBounds } from '../controls/parent-bounds'
 import { ParentOutlines } from '../controls/parent-outlines'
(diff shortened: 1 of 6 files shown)


gbalint:
I think it is counterintuitive that the new element stays and the old one is left behind. We could say that this is an implementation detail, but I'm afraid this is going to leak to the user, e.g. with commenting (the comments will stay on the moved element), or visual scripting, or any extra data which relies on uids.

editor/src/components/canvas/canvas-strategies/absolute-resize-helpers.ts
```diff
@@ -5,7 +5,7 @@ export const verticalPins: Array<AbsolutePin> = ['top', 'height', 'bottom']
 
 export function hasAtLeastTwoPinsPerSide(props: { [key: string]: any }): boolean {
   return (
-    horizontalPins.filter((pin) => props['style'][pin] != null).length >= 2 &&
-    verticalPins.filter((pin) => props['style'][pin] != null).length >= 2
+    horizontalPins.filter((pin) => props['style']?.[pin] != null).length >= 2 &&
```
balazsbajorics, line 8: this is an unrelated fix but I kept triggering an error here during development

editor/src/components/canvas/canvas-strategies/absolute-duplicate-strategy.tsx
```diff
@@ -0,0 +1,112 @@
...
+      ...strategyState,
+      startingMetadata: withDuplicatedMetadata,
+    })
+
+    return {
+      commands: [
+        ...duplicateCommands,
+        ...moveCommands.commands,
+
+        setElementsToRerenderCommand([...canvasState.selectedElements, ...newPaths]),
```
seanparsons, line 104: Do the selected elements need to be re-rendered?

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/absolute-resize-helpers.ts
seanparsons, line 8: Is this clearer?
```suggestion
    horizontalPins.filter((pin) => props?.style?.pin != null).length >= 2 &&
```

balazsbajorics: hmm that's no good because pin is a variable, but I can remove the brackets from around style
```suggestion
    horizontalPins.filter((pin) => props.style?.[pin] != null).length >= 2 &&
```

balazsbajorics: pushed in bf62a4ec5

seanparsons: > hmm that's no good because pin is a variable, but I can remove the brackets from around style

OMG, proving my own point that it made me blind to that being a variable.

enidemi (approved):

PR #2320, opened 2022-06-16
fix(canvas) reparent strategy checks control type

enidemi:
**Problem:**
It's possible to trigger reparent strategy while resizing and pressing cmd.

**Fix:**
Other Absolute Move strategies are checking the active control type in their fitness functions, reparent missed this.

**Commit Details:**
- update `fitness` in `absolute-reparent-strategy`

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.tsx
@@ -46,6 +46,7 @@ export const absoluteReparentStrategy: CanvasStrategy = {
   fitness: (canvasState, interactionState) => {
     if (
       canvasState.selectedElements.length > 0 &&
+      interactionState.activeControl.type === 'BOUNDING_AREA' &&
       interactionState.interactionData.modifiers.cmd &&
       interactionState.interactionData.type === 'DRAG' &&
       interactionState.interactionData.drag != null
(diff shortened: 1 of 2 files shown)


gbalint (approved):

gbalint:
I think we should add fitness tests to the strategy unit tests! Right now we only test the apply function in those.
We even set the controlType to null: https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx#L44

I think a test like this for reparent fitness is quite easy to add, I would do that in this PR as an example

balazsbajorics:
> I think we should add fitness tests to the strategy unit tests! Right now we only test the apply function in those. We even set the controlType to null: https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx#L44
> 
> I think a test like this for reparent fitness is quite easy to add, I would do that in this PR as an example

I'm not sure a fitness _unit_ test would be useful. it's a short function with a single expression. the test wouldn't have caught that this is not filtering for bounds because whoever wrote the function didn't know they want to test for that case

enidemi:
> > I think we should add fitness tests to the strategy unit tests! Right now we only test the apply function in those. We even set the controlType to null: https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx#L44
> > I think a test like this for reparent fitness is quite easy to add, I would do that in this PR as an example
> 
> I'm not sure a fitness _unit_ test would be useful. it's a short function with a single expression. the test wouldn't have caught that this is not filtering for bounds because whoever wrote the function didn't know they want to test for that case
> 
> however, we should have end-to-end tests for the new interactions, but I think it'd be premature to add those now, while things are in flux

I added a test for this specific case here, we already had some fitness tests for how the escape hatch works.

PR #2322, opened 2022-06-16
Fixes some keyboard absolute move guideline bugs.

seanparsons:
**Problem:**
There are some minor issues with the absolute keyboard move strategy:
- Pressing the `alt` key prevents moving with the keyboard completely.
- Once a move has been started pressing the `alt` key doesn't show the distance guidelines.
- The snapping guidelines don't show at all when moving with the keyboard.

**Fix:**
The fixes for the above respectively are:
- Remove the check on the `alt` key from the fitness function of the strategy.
- Changed the check in the guidelines which prevented them from showing up if any interaction has been started.
- Used some existing code for creating guidelines from the mouse absolute move strategy to include the appropriate guidelines during a keyboard move.

Additionally an issue was found when the timeout cleared the interaction strategy but resulted in the snapping guidelines persisting because `foldAndApplyCommands` was accumulating the transient patches as well as permanent ones.

**Commit Details:**
- The `alt` key is now ignored by `keyboardAbsoluteMoveStrategy`, as it should
  be independent of toggling the distance guidelines.
- Includes the `activateSnap` guidelines while moving with the keyboard
  without snapping to them.
- `snapDrag` moved into `shared-absolute-move-strategy-helpers.ts`.
- `foldAndApplyCommands` does not accumulate patches for commands that are transient.
- Added `key` and `id` to the element underlying `GuidelineControl`.
- `DistanceGuidelineControl` changed from not displaying on any interaction being
  active, to not displaying on a drag interaction being active.
- `CLEAR_INTERACTION_SESSION` now wipes the entire `controls` field of
  `EditorStateCanvas`.

### DIFF
--- a/editor/src/templates/editor-canvas.tsx
+++ b/editor/src/templates/editor-canvas.tsx
@@ -47,6 +47,7 @@ import {
   ConsoleLog,
   DerivedState,
   EditorState,
+  editorStateCanvasControls,
   isOpenFileUiJs,
 } from '../components/editor/store/editor-state'
 import {
@@ -391,6 +392,7 @@ export function runLocalCanvasAction(
           ...model.canvas,
           interactionSession: null,
           domWalkerInvalidateCount: model.canvas.domWalkerInvalidateCount + 1,
+          controls: editorStateCanvasControls([], []),
         },
         jsxMetadata: {},
         domMetadata: {},
--- a/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.tsx
@@ -24,6 +24,7 @@ import {
   getAbsoluteMoveCommandsForSelectedElement,
   getDragTargets,
   getMultiselectBounds,
+  snapDrag,
 } from './shared-absolute-move-strategy-helpers'
 
 export const absoluteMoveStrategy: CanvasStrategy = {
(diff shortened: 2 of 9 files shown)


balazsbajorics:
> Changed the check in the guidelines which prevented them from showing up if any interaction has been started.

this may potentially collide with alt-drag-to-duplicate. I _guess_ we want to keep the distance guidelines hidden while alt-drag-to-duplicating? cc @maltenuhn

seanparsons:
> > Changed the check in the guidelines which prevented them from showing up if any interaction has been started.
> 
> this may potentially collide with alt-drag-to-duplicate. I _guess_ we want to keep the distance guidelines hidden while alt-drag-to-duplicating? cc @maltenuhn

From the commit details: 
> `DistanceGuidelineControl` changed from not displaying on any interaction being active, to not displaying on a drag interaction being active.

Rheeseyb (approved):

editor/src/templates/editor-canvas.tsx
```diff
@@ -391,6 +392,7 @@ export function runLocalCanvasAction(
           ...model.canvas,
           interactionSession: null,
           domWalkerInvalidateCount: model.canvas.domWalkerInvalidateCount + 1,
+          controls: editorStateCanvasControls([], []),
```
balazsbajorics, line 395: is this still needed here?

balazsbajorics (approved):

PR #2324, opened 2022-06-16
Introducing the Duplicate Cursor

lankaukk:
**Problem:**
With our new alt drag to duplicate functionality, we are currently using a grabbing glove as the cursor.

**Fix:**
Now we have the same familiar duplication cursor that designers see in Figma, and it is ready to be used in the new feature with the name "Duplicate".


Rheeseyb (approved):

balazsbajorics (approved):
when we actually start using it, let's make sure the top left is really at 4;4 :)

enidemi (approved):

seanparsons (approved):

PR #2326, opened 2022-06-17
Fix/canvas resize strategy

enidemi:
**Problem:**
There are 2 resize strategies but one of them is better (works with modifiers, multiselect), the other is killed in this PR. The plan was that will work nicely with partial props, but actually inserting missing props is needed for a good canvas behavior.

**Fix:**
Deleting the delta based resize strategy. Extending the other resize strategy to insert props if needed.

**Commit Details:**
- delete strategy
- update bounding box delta strategy `isApplicable` to include single select
- set missing prop if needed
- added tests for missing prop insertion


editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
```diff
@@ -36,29 +36,17 @@ import {
...
-        const elementProps = allElementProps[EP.toString(element)] ?? {}
-        return (
-          elementMetadata?.specialSizeMeasurements.position === 'absolute' &&
-          hasAtLeastTwoPinsPerSide(elementProps) // TODO should this use projectContents?
-        )
-      })
-    } else {
-      return false
-    }
+    return filteredSelectedElements.every((element) => {
```
gbalint, line 46: Shouldn't we still check that the number of selected elements is at least 1? I guess without that there is no control to drag though, so I'm unsure if it is necessary or not, but we still check that for other strategies.

editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.spec.tsx
```diff
@@ -1374,3 +1374,219 @@ describe('Absolute Resize Bounding Box Strategy single select', () => {
...
+            }}
+            data-uid='bbb'
+          >
+            <div data-uid='ccc' style={{width: 100, height: 80}} />
+          </div>
+        </div>`,
+      ),
+    )
+  })
+  it('works with element (no height, no width) resized from Left Edge, only adds width', async () => {
```
balazsbajorics, line 1452: this is cool!

balazsbajorics (approved):
very nice!

editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
enidemi: fixed

PR #2333, opened 2022-06-20
Revert ElementPath equality function

Rheeseyb:
**Problem:**
Our equality function for `ElementPath`s was recently optimised based on the assumption we were only creating one instance for any given path, since the factory functions for creating them all read from a cache before creating them. However, the cache used is cleared on worker updates, making that assumption invalid.

**Fix:**
Restore the previous deep equality function


balazsbajorics (approved):
:(

PR #2343, opened 2022-06-22
Make keyboard strategies resilient to repeated invocation.

seanparsons:
**Problem:**
Currently the keyboard strategies are implemented in an instantaneous style, such that they handle a keypress at a point in time. This can cause issues however when the strategies are invoked more than once when for example a modifier is pressed or released as it may cause the logic to internally move something a second time and then remove a guideline that was being shown.

**Fix:**
The primary change here is to record all the changes that happen to the keypresses since the interaction started. Then both the absolute move and resize strategies have a pass that accumulates repeated presses with counts. That is then followed by a strategy specific approach to combine those accumulated keypresses into a minimum number of commands for the strategy.

**Commit Details:**
- `KeyboardInteractionData` now records an array of `KeyState` elements,
  which include all the keypresses triggered since the interaction was first
  started.
- `interactionDataHardReset` now prunes the keyboard interaction key states
  to the latest.
- Extracted out `getDragDeltaFromKey` function from the two keyboard strategies.
- Added shared utility functions `getLastKeyPressState` and `accumulatePresses`.
- The move and resize strategies now accumulate repeated presses, populating a
  count of how many repeated presses for identical keypresses there are. Then
  the resultant drag is accumulated from those changes, so that the minimum of
  commands is produced by the strategy.
- Modified a few conditionals to correctly handle the case that now there is
  not a common `modifiers` field on both keyboard and mouse interactions.
- `handleAccumulatingKeypresses` now also culls the key states so that they
  are not carried over into a different strategy.
- Added some `keyup` dispatches to some test code, because we were firing repeated `keydown`
  events which to our logic looks like they are still depressed when we press
  another key.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
@@ -51,6 +51,8 @@ export const absoluteResizeBoundingBoxStrategy: CanvasStrategy = {
     } else {
       return false
     }
+
+    return false
   },
   controlsToRender: [
     { control: AbsoluteResizeControl, key: 'absolute-resize-control', show: 'always-visible' },
--- a/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
@@ -51,8 +51,6 @@ export const absoluteResizeBoundingBoxStrategy: CanvasStrategy = {
     } else {
       return false
     }
-
-    return false
   },
   controlsToRender: [
     { control: AbsoluteResizeControl, key: 'absolute-resize-control', show: 'always-visible' },
--- a/editor/src/components/canvas/canvas-strategies/interaction-state.ts
+++ b/editor/src/components/canvas/canvas-strategies/interaction-state.ts
@@ -37,13 +37,6 @@ export interface KeyState {
   modifiers: Modifiers
 }
 
-export function keyStatesEqual(first: KeyState, second: KeyState): boolean {
-  return (
-    setsEqual(first.keysPressed, second.keysPressed) &&
-    Modifier.equal(first.modifiers, second.modifiers)
-  )
-}
-
 export interface KeyboardInteractionData {
   type: 'KEYBOARD'
   keyStates: Array<KeyState>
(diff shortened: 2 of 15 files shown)


editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -340,6 +338,32 @@ export function hasDragModifiersChanged(
   )
 }
 
+export function hasKeyboardModifierChanged(
```
balazsbajorics, line 341: it looks like this function is not used anywhere, only imported in dispatch-strategies.tsx

editor/src/components/canvas/canvas-strategies/keyboard-absolute-move-resize.spec.browser2.tsx
```diff
@@ -44,6 +44,34 @@ describe('Keyboard Absolute Strategies E2E', () => {
     await expectElementLeftInPrintedCode(30)
   })
 
+  it('Pressing Shift + ArrowRight 3 times, then pressing ArrowRight 3 times', async () => {
```
balazsbajorics, line 47: yay more keypress tests! 😎

editor/src/components/editor/store/dispatch-strategies.tsx
```diff
@@ -406,50 +411,68 @@ function handleAccumulatingKeypresses(
...
-      customStrategyState: strategyResult.customState ?? strategyState.customStrategyState,
-      startingAllElementProps: strategyState.startingAllElementProps,
-    }
+    const interactionData = newEditorState.canvas.interactionSession.interactionData
+    if (isKeyboardInteractionData(interactionData)) {
+      const lastKeyState = last(interactionData.keyStates)
+      const updatedInteractionData: KeyboardInteractionData = {
+        ...interactionData,
+        keyStates: lastKeyState == null ? [] : [lastKeyState],
+      }
```
gbalint, line 420: I have ambigous feelings about this. Until now the data flow looked like this:
- user interactions changed the interaction state
- based on the editor state and the interaction state strategies modified the state, but mostly the projectcontents itself and/or the StrategyState

But in this code the strategy handling modifies the interaction state itself, so the interaction state is not purely a description of the user behavior. I believe when the strategies need state, that should be in the StrategyState not in the InteractionState. So how I would do it is to put this information into the StrategyState, or to be more precise, probably into the StrategyState.customStrategyState.

editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
```diff
@@ -51,6 +51,8 @@ export const absoluteResizeBoundingBoxStrategy: CanvasStrategy = {
     } else {
       return false
     }
+
+    return false
```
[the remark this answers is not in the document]
seanparsons: Sure is, fixed in 44845db84.

editor/src/components/editor/store/dispatch-strategies.tsx
seanparsons: Surely `interactionDataHardReset` would fall into that as well? As that is triggered from `interactionHardReset` and it modifies the `drag` and `prevDrag` values.

gbalint: Good argument! It seems the separation between interaction and strategy states is just not that clear

balazsbajorics: I'd definitely not put it in the `customStrategyState`, that should be reserved for things that are set by apply functions.

balazsbajorics (approved):

PR #2346, opened 2022-06-22
Handle missing left/top properties

gbalint:
# Issue

Dragging an absolutely positioned element should never fail because eg left and top are missing. If either of them is missing, it should be added in. We can make it smarter later (around which properties to set).

# Solution

This is a basic solution, if there is no `left` or `right` prop, I add a `left` (with initial value 0), if there is `top` or `bottom`, I add a `top` (with initial value 0)

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
@@ -71,17 +71,9 @@ function createMoveCommandsForElement(
       // TODO avoid using the loaded FramePoint enum
       framePointForPinnedProp(pin),
     )
+    const negative = pin === 'right' || pin === 'bottom'
 
-    const updatedPropValue = (() => {
-      const sizePin = pin === 'width' || pin === 'height'
-      // if the pin is width or height, that dimension was not defined so we can set it to 0
-      if (sizePin) {
-        return 0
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
```diff
@@ -82,7 +89,28 @@ function createMoveCommandsForElement(
...
+            drag.y,
+            elementParentBounds?.height,
+            true,
+          )
+        } else {
+          return null
+        }
       }
     },
     ['top', 'bottom', 'left', 'right'] as const,
```
balazsbajorics, line 116: Eni's https://github.com/concrete-utopia/utopia/pull/2326 has a slightly different approach to the same problem (adding missing props).

I think it'd make this code clearer if instead of getting `['top', 'bottom', 'left', 'right']` here, it'd only get the props to actually work on, and create the array of props to work on in a previous step. then while making the array you could ensure there's at least 1 prop per dimension (inserting left or top as needed). 
Kind of like how `ensureAtLeastTwoPinsForEdgePosition` works.

gbalint: Great! I refactored it according to your suggestions.

balazsbajorics (approved):

PR #2349, opened 2022-06-22
Fix adding classes on mouse move when className input focused

### LINKED ISSUE #2348: Moving mouse while typing classname selects class
**Describe the bug**
Moving mouse while typing classname selects class

**To Reproduce**
Steps to reproduce the behavior. For canvas tasks, take care to note down all relevant key holds, keys typed, selected state etc. E.g:

1. Start new project
2. Add tailwind to project
3. Begin typing a className
4. Move mouse

Result: The current suggestion is selected and added to the list

**Expected behaviour**
nothing

**Screenshots**
moving mouse selects class suggestion

BLamy:
Fixes #2348 

**Problem:**
Moving mouse while typing classname selects class

**Fix:**
Remove the deffered setState

**Commit Details:**
- Remove the deffered setState

### DIFF
--- a/editor/src/components/inspector/sections/style-section/className-subsection/className-subsection.tsx
+++ b/editor/src/components/inspector/sections/style-section/className-subsection/className-subsection.tsx
@@ -99,7 +99,6 @@ const menu: styleFn = (base) => ({
 })
 
 const AlwaysTrue = () => true
-let queuedDispatchTimeout: number | undefined = undefined
 let queuedFocusTimeout: number | undefined = undefined
 
 const focusedOptionAtom = atomWithPubSub<string | null>({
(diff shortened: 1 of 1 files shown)


balazsbajorics:
hey there! thanks so much for taking the time and making pull requests for us :) 

I locally checked out your branch and I could still reproduce the bug from #2348 
I'm a little bit uncertain because reproduction doesn't seem 100% reliable to me, and I still didn't find the root cause.

But while I was checking things out, I realized we recently touched this code and made a mistake, so I made a PR to fix that. could you try to see if that PR fixes your bug too? the PR is https://github.com/concrete-utopia/utopia/pull/2351

balazsbajorics:
we weren't able to reproduce #2348 since #2351 was merged, so it looks like the bug is fixed. I'm going to close this PR now.

thank you again for taking the time to create the bug ticket with the reproduction video!

PR #2356, opened 2022-06-24
Cull element path cache rather than trashing it

### LINKED ISSUE #2336: Replace ElementPath cache with a WeakMap
We cache `ElementPath`s for performance (since we create so many of them), but then explicitly clear the cache on an `UPDATE_FROM_WORKER` action, meaning it is purged for every single code change. Rather than doing this, we should investigate using a [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) for this cache.

Rheeseyb:
Fixes #2336

**Problem:**
Every code editor change will completely trash the element path cache

**Fix:**
Rather than trashing it repeatedly, instead use a timer to determine when the editor is idle, and use that cull the element path cache by removing paths for now non-existent UIDs. In theory this means that paths with entirely valid UIDs may no longer exist, but that would only happen in the most extreme of edge cases (thanks to the logic for attempting to retain UIDs which would mean if code shifts about it would generate a new UID, leaving the old path ripe for culling)

**Commit Details:**
- Added a function to `element-path.ts` which will remove paths from the various caches if they contain non-existent UIDs
- Added a timeout inside the handle for the `UPDATE_FROM_WORKER` action under the assumption that this is a good approximation to the editor being idle (I'm happy to move this elsewhere if anyone has a better suggestion - I was originally thinking inside the dispatch function itself, but we've often added actions that fire periodically, so I wanted to tie this to some action that indicates there was some deliberate interaction from the user, and this action fires after any form of update to the model whatsoever)
- Added unit tests to ensure the cache culling is working correctly, and an integration test to make sure it happens after the timeout when running the editor

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -334,10 +334,6 @@ export function resetDispatchGlobals(): void {
   applyProjectChangesCoordinator = Promise.resolve()
 }
 
-let cullElementPathCacheTimeoutId: number | undefined = undefined
-const CullElementPathCacheTimeout = 1000
-let lastProjectContents: ProjectContentTreeRoot = {}
-
 export function editorDispatch(
   boundDispatch: EditorDispatch,
   dispatchedActions: readonly EditorAction[],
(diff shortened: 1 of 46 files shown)


seanparsons (approved):

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -507,9 +512,33 @@ export function editorDispatch(
     storedState.workers.initWatchdogWorker(frozenEditorState.id)
   }
 
+  lastProjectContents = storedState.unpatchedEditor.projectContents
```
balazsbajorics, line 515: I'm worried this function is becoming a spaghetti, could you please move this new block to a new function? (we should also break down the code above this into sub-functions)

Rheeseyb: Yeah sure, I'll extract it. I wouldn't want it to live in `element-path.ts` since that pulls dependencies on the project model in with it, but it could certainly be tidied up here

balazsbajorics (approved):

PR #2357, opened 2022-06-24
Fix aspect ratio locked resize when dragging over the opposite edge

gbalint:
# Issue

Holding down shift while resizing a mirrored element should apply the aspect ratio correctly - not move it to its original position. 

# Fix

I realized that there is a simpler way to implement aspect ratio locked resize. Instead of having a whole separate logic to handle aspect ratio lock we can just allow normal resize to calculate the rectangle, and then extend the resulting rectangle so it has the correct aspect ratio. We just have to be careful how do we position the resulting rectangle, it should always have a fixed edge/corner, which is the opposite from the dragged one.
This implementation doesn't have the bug described in the report.

### DIFF
--- a/editor/src/core/shared/math-utils.ts
+++ b/editor/src/core/shared/math-utils.ts
@@ -741,24 +741,6 @@ export function lineIntersection<C extends CoordinateMarker>(
   } as Point<C>
 }
 
-// Returns which side of the line is point b. If looking from lineA to lineB it is on the left, it returns the value 'left', etc.
-export function sideOfLine<C extends CoordinateMarker>(
-  lineA: Point<C>,
-  lineB: Point<C>,
-  p: Point<C>,
-): 'left' | 'right' | 'on-line' {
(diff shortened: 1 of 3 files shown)


editor/src/core/shared/math-utils.ts
```diff
@@ -741,24 +741,6 @@ export function lineIntersection<C extends CoordinateMarker>(
   } as Point<C>
 }
 
-// Returns which side of the line is point b. If looking from lineA to lineB it is on the left, it returns the value 'left', etc.
```
balazsbajorics, line 744: why did you remove this? is it problematic?

balazsbajorics (approved):

editor/src/core/shared/math-utils.ts
gbalint: I removed this because it is not necessary for the new algorithm, and it was only wrote and used for this.

PR #2358, opened 2022-06-24
Resize with keyboard only shows guidelines when fully aligned.

seanparsons:
**Problem:**
Currently when resizing with the keyboard we have no guidelines, but we want them to appear similarly to when moving with the keyboard on full alignment.

**Fix:**
The logic for guidelines when moving with the keyboard was extracted into a function and used in the keyboard resize strategy.

**Commit Details:**
- Refactored out the guidelines logic from the keyboard absolute move strategy
  into a function called `getKeyboardStrategyGuidelines`.
- Keyboard absolute resize strategy calculates the new bounds of the
  resize so that the guidelines can be calculated using
  `getKeyboardStrategyGuidelines`.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/keyboard-absolute-resize-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/keyboard-absolute-resize-strategy.tsx
@@ -23,7 +23,7 @@ import { last } from '../../../core/shared/array-utils'
 import {
   AccumulatedPresses,
   accumulatePresses,
-  getDragDeltaFromKey,
+  getMovementDeltaFromKey,
   getKeyboardStrategyGuidelines,
   getLastKeyPressState,
 } from './shared-keyboard-strategy-helpers'
@@ -31,7 +31,7 @@ import { getMultiselectBounds } from './shared-absolute-move-strategy-helpers'
 import { setSnappingGuidelines } from '../commands/set-snapping-guidelines-command'
 
 interface VectorAndEdge {
-  drag: CanvasVector
+  movement: CanvasVector
   edge: EdgePosition
 }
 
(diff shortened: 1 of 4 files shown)


gbalint (approved):

editor/src/components/canvas/canvas-strategies/keyboard-absolute-resize-strategy.tsx
```diff
@@ -104,9 +113,22 @@ export const keyboardAbsoluteResizeStrategy: CanvasStrategy = {
     if (interactionState.interactionData.type === 'KEYBOARD') {
       const accumulatedPresses = accumulatePresses(interactionState.interactionData.keyStates)
       const dragsWithEdges = pressesToVectorAndEdges(accumulatedPresses)
+
+      // Start with the frame as it is at the start of the interaction.
+      let draggedFrame =
```
Rheeseyb, line 118: I really dislike that we are using `drag` everywhere in this file for a keyboard strategy

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/keyboard-absolute-resize-strategy.tsx
seanparsons: I'll see about doing some renaming.

enidemi (approved):

editor/src/components/canvas/canvas-strategies/keyboard-absolute-resize-strategy.tsx
seanparsons: I've renamed a bunch of those variables in 12d589a8b.

PR #2360, opened 2022-06-24
Cancel keyboard interaction on all keyboard shortcuts

gbalint:
# Issue

When there is an active keyboard interaction, other keyboard shortcuts (e.g. delete, paste, etc) dispatch their actions, but the interaction goes one, and everything is mixed into a single big undo step.

# Fix

This was already fixed for undo in https://github.com/concrete-utopia/utopia/pull/2342
I believe the best is to just cancel the keyboard interaction on any keyboard shortcuts. Although some of the do not have any effect on undo (e.g. zoom out), but it doesn't hurt to restart the interaction after a shortcut like that.

### DIFF
--- a/editor/src/components/editor/global-shortcuts.ts
+++ b/editor/src/components/editor/global-shortcuts.ts
@@ -711,12 +711,6 @@ export function handleKeyDown(
           : []
       },
       [UNDO_CHANGES_SHORTCUT]: () => {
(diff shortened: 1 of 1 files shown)


Rheeseyb (approved):

balazsbajorics:
@gbalint the failing test is actually catching a behavior change in your PR! we discord chatted about this with @maltenuhn and we think the new behavior is better, can you please update the test?

enidemi (approved):

PR #2362, opened 2022-06-24
Use toString for ElementPath equality checks

Rheeseyb:
**Problem:**
The `ElementPath` equality check is expensive for near misses (e.g. sibling paths), and we can't necessarily guarantee reference equality.

**Fix:**
Fall back to string equality if reference equality fails, as we are caching the stringified versions too.

### DIFF
--- a/editor/src/core/shared/element-path.ts
+++ b/editor/src/core/shared/element-path.ts
@@ -479,16 +479,8 @@ export function appendPartToPath(path: ElementPath, next: ElementPathPart): Elem
   return elementPath(updatedPathParts)
 }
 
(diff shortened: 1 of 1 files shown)


seanparsons (approved):

Rheeseyb:
> it looks like the only other place using `fullElementPathsEqual` is `notNullPathsEqual`. `notNullPathsEqual` should also be updated to use string equality, and `fullElementPathsEqual` should be removed

Good catch. `notNullPathsEqual` looks to be completely redundant, so I've completely removed both functions.

balazsbajorics (approved):

PR #2365, opened 2022-06-27
Fix dragging absolute position element with missing pins

gbalint:
# Issue

When the position of an absolute positioned element is not fully defined (horizontally or vertically), we add the missing `top`/`left` props on dragging, see https://github.com/concrete-utopia/utopia/pull/2346
However, this PR was buggy, because it assumed that the missing prop can be initialized with 0. But `absolute` elements with missing props can behave as `static` ones in layouting, so their position can be non-zero.

# Fix

Read the real position from the metadata (`localFrame`), and offset the drag coordinates with that.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.tsx
@@ -193,7 +193,7 @@ describe('Absolute Move Strategy', () => {
     )
   })
 
-  it('works with a T pinned absolute element, assumes L to 0', async () => {
+  it('works with a T pinned absolute element, assumes L to localframe position in metadata', async () => {
     const targetElement = elementPath([
       ['scene-aaa', 'app-entity'],
       ['aaa', 'bbb'],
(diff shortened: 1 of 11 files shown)


editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.tsx
```diff
@@ -217,7 +217,7 @@ describe('Absolute Move Strategy', () => {
       makeTestProjectCodeWithSnippet(
         `<View style={{ ...(props.style || {}) }} data-uid='aaa'>
         <View
-          style={{ backgroundColor: '#0091FFAA', position: 'absolute', top: 65, width: 250, height: 300, left: 15 }}
+          style={{ backgroundColor: '#0091FFAA', position: 'absolute', top: 65, width: 250, height: 300, left: 65 }}
```
balazsbajorics, line 220: could you make a karma test for this? the fake metadata is making this test a bit too fake!

also, in the karma test, could you please add margin to the parent? that should shift the static element and we should respect that

Rheeseyb: This is a good idea. It seems like we could add a few tests that shift the absolute positioned element to make sure we're not missing any other edge cases like this

editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
```diff
@@ -241,7 +262,7 @@ export function snapDrag(
 const horizontalPins: Array<AbsolutePin> = ['left', 'right']
 const verticalPins: Array<AbsolutePin> = ['top', 'bottom']
 
-function ensureAtLeastOnePinPerDimension(props: PropsOrJSXAttributes): Array<AbsolutePin> {
+function ensureAtLeastOnePinPerDimension(props: PropsOrJSXAttributes) {
```
Rheeseyb, line 265: Please can you add a return type to the function signature

enidemi (approved):

balazsbajorics (approved):

PR #2366, opened 2022-06-27
fix(canvas) use correct parent frame when dragging a root div

enidemi:
**Problem:**
When dragging a component's root div the parent controls are in a wrong place (parent bounds, parent outline and pin lines).

**Fix:**
Use `specialSizeMeasurements.immediateParentBounds` and `specialSizeMeasurements.coordinateSystemBounds` instead of parentPath and parent metadata.
There are many questions about how to solve the component and its root div issues, but at least with this fix it's visually better.

**Commit Details:**
- update parent-bounds, parent-outlines and pin lines


balazsbajorics:
this has been superseded by #2368 and can be closed, right?

balazsbajorics (approved):

PR #2371, opened 2022-06-27
Only follow code selection changes if code editor is focused

Rheeseyb:
**Problem:**
Occasionally changes made to an element via the canvas will result in the parent element being incorrectly selected.

**Fix:**
The issue here was that under certain (very hard to reproduce) circumstances the VS Code extension would believe that a selection change had been initiated from the code editor. We were using a variable to explicitly squash these changes in the extension, but I believe there was a race condition around setting and clearing that variable. Rather than using a global variable like that however, we can actually check with the VS Code window is focused, and ignore any `TextEditorSelectionChangeEvent`s when that is the case.

### DIFF
--- a/utopia-vscode-extension/src/extension.ts
+++ b/utopia-vscode-extension/src/extension.ts
@@ -351,7 +351,6 @@ function getFullConfig(): UtopiaVSCodeConfig {
 
 let currentDecorations: Array<DecorationRange> = []
 let currentSelection: BoundsInFile | null = null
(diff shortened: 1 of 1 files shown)


balazsbajorics (approved):
quite elegant idea!

Rheeseyb:
> quite elegant idea!

Quite un-elegant of previous Rheese to not think of this!

gbalint (approved):

PR #2372, opened 2022-06-27
Do not clear highlights on hover end during insertion

Rheeseyb:
**Problem:**
When dragging to insert, if you drag past the boundary of the scene you will end up targeting the canvas itself.

**Fix:**
Although we were correctly preventing highlighting new elements during insertion when hovering over them, we weren't doing the same when ending the hover, making it possible to unintentionally clear the highlighted element and then never replace it.


enidemi:
please add a test for this bugfix!

seanparsons (approved):

PR #2375, opened 2022-06-28
Do not show resize controls in insert mode

gbalint:
# Issue
When inserting an element the selected elements resize control is still visible and mousedown on it can start a resize which fails to clear on mouseup

# Fix
Do not render strategy controls in insert mode

### DIFF
--- a/editor/src/components/canvas/controls/new-canvas-controls.tsx
+++ b/editor/src/components/canvas/controls/new-canvas-controls.tsx
@@ -457,9 +457,9 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
       {when(
         resizeStatus !== 'disabled',
         <>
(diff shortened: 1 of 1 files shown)


enidemi (approved):

Rheeseyb (approved):

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -483,11 +478,17 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
...
           )}
         </>,
       )}
     </div>
   )
 }
+
+function isCanvasStrategyInSelectOrSelectLiteMode(mode: Mode): boolean {
+  return (
+    (isFeatureEnabled('Canvas Strategies') && mode.type === 'select') || mode.type === 'select-lite'
```
seanparsons, line 492: It could be how it reads just to me, but this condition is not what I'd expect from the function name.

seanparsons (approved):

PR #2383, opened 2022-06-29
Use the AABB for re-parenting

Rheeseyb:
**Problem:**
When cmd+dragging to re-parent, the result would flip-flop as we were using the live DOM to determine the new parent, but the unpatched editor state and metadata.

**Fix:**
Restore an AABB version of the `getAllTargetsAtPoint` function specifically for use in re-parenting. I've also tidied up the other function by removing the now unnecessary params.


balazsbajorics (approved):

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -141,8 +166,15 @@ export function getAllTargetsAtPoint(
     }),
   )
 
-  // TODO FIXME we should take the zero-sized elements from Canvas.getAllTargetsAtPoint, and insert them (in a correct-enough order) here. See PR for context https://github.com/concrete-utopia/utopia/pull/2345
-  return elementsFromDOM
+  return getElementsUnderPointFromAABB
+    .filter((foundElement) => {
+      if (!foundElement.canBeFilteredOut) {
```
seanparsons, line 171: This conditional could probably do with being flipped.

seanparsons (approved):

gbalint (approved):

PR #2389, opened 2022-06-30
Fix selection/dragging/hover behavior when there is an existing (multi)selection

gbalint:
When there is an existing (multi)selection, we expect the multiselection to be dragged when you drag the selection area. Even when the mouse is over a different overlapping element, or even when there is no element under the mouse
However, on mouse up, we should select the element under the mouse (or deselect if there is nothing there).
In other cases, we should still select on mouse down.

Note:
- I started from https://github.com/concrete-utopia/utopia/pull/2383 , because I modified the same part of the code, and use AABB to get the selection bounds.
- I added tests for the dragging cases, but I would like to wait with the selection tests until https://github.com/concrete-utopia/utopia/pull/2382 is merged.


editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -216,12 +217,31 @@ function useFindValidTarget(): (
...
-    (selectableViews: Array<ElementPath>, mousePoint: WindowPoint | null) => {
-      const { selectedViews } = storeRef.current
-      const validElementMouseOver: ElementPath | null = getValidTargetAtPoint(
-        selectableViews,
-        mousePoint,
-      )
+    (
+      selectableViews: Array<ElementPath>,
+      mousePoint: WindowPoint | null,
+      preferAlreadySelected: boolean,
```
Rheeseyb, line 223: Can you change this to a union of strings please to aid readability of the function calls

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -122,6 +123,109 @@ export function getAllTargetsAtPoint(
...
+    validElementPathsForLookup == 'no-filter'
+      ? 'no-filter'
+      : new Set(
+          validElementPathsForLookup.map((path) => EP.toString(EP.makeLastPartOfPathStatic(path))),
+        )
+  const elementsFromDOM = stripNulls(
+    elementsUnderPoint.map((element) => {
+      const foundValidelementPath = findFirstParentWithValidElementPath(validPathsSet, element)
+      if (foundValidelementPath != null) {
+        return foundValidelementPath
```
Rheeseyb, line 180:
```suggestion
      const foundValidElementPath = findFirstParentWithValidElementPath(validPathsSet, element)
      if (foundValidElementPath != null) {
        return foundValidElementPath
```
Extremely minor :)

Rheeseyb (approved):

enidemi (approved):

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -122,6 +123,103 @@ export function getAllTargetsAtPoint(
...
+    selectedViews,
+    hiddenInstances,
+    validElementPathsForLookup,
+    point,
+    canvasScale,
+    canvasOffset,
+    allElementProps,
+  )
+  if (targets === 'selection') {
+    return selectedViews[0] ?? null
```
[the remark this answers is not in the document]
gbalint: It doesn't matter, we don't make difference between the selected elements, because they are all moved anyway.

PR #2392, opened 2022-07-01
fix(insertion) insert menu should add the same props as shortcuts

enidemi:
**Problem:**
When you insert a View from the insert menu `position: absolute` is missing from the style props.

**Fix:**
The solution is to use the same default style on shortcut insertion and in insert menu too. 

**Commit Details:**
- update utopia-api component descriptor

### DIFF
--- a/editor/src/components/editor/defaults.ts
+++ b/editor/src/components/editor/defaults.ts
@@ -7,6 +7,7 @@ import {
   JSXElementChildren,
   emptyComments,
   jsxTextBlock,
+  JSXAttribute,
 } from '../../core/shared/element-template'
 import { NormalisedFrame } from 'utopia-api/core'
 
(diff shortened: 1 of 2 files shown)


Rheeseyb (approved):

editor/src/components/editor/defaults.ts
```diff
@@ -83,17 +92,21 @@ export function defaultTransparentViewElement(uid: string): JSXElement {
   )
 }
 
+export function defaultTextElementStyle(): JSXAttribute {
+  return jsxAttributeValue(
+    {
+      fontSize: 16,
```
gbalint, line 98: I would extract this 16 into a named constant

gbalint (approved):

balazsbajorics (approved):
could you also add position: 'absolute' to the Rectangle and Ellipse elements too?

PR #2394, opened 2022-07-01
Feature/do not show ghost snap guidelines

balazsbajorics:
**Problem:**
The snapping guidelines can show up even if the user attemps to drag an "undraggable" component, which doesn't read the props.style prop.

This happens because in the strategies we calculate the guidelines when we calculate the snapping, as part of calculating the intended new bounding box of the targeted elements. But then we actually issue the commands, calculate the patched EditorState and render the canvas, it can turn out that the element will simply remain in place, unmoved. The guidelines are still shown, because those were set by the strategy before we actually tried moving the element.

**Fix:**
My proposed solution to this problem is that all strategies that use guidelines must communicate their `intendedBounds`, which is an array of `CanvasFrameAndTarget`. 
Then the actual guidelineControl compares the intendedBounds to the actual resulting bounds, and only display the guidelines if the two match up.

**Commit Details:**
- Added new command called `pushIntendedBounds`
- I call `pushIntendedBounds` from every strategy that uses the `setSnappingGuidelines` command
- In some move/resize helper functions, I return some variant of ` { commands: CanvasCommand[]; intendedBounds: CanvasFrameAndTarget[] }`


Rheeseyb (approved):

enidemi (approved):

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.tsx
```diff
@@ -98,7 +110,10 @@ export function applyAbsoluteMoveCommon(
   canvasState: InteractionCanvasState,
   interactionState: InteractionSession,
   strategyState: StrategyState,
-  getMoveCommands: (snappedDragVector: CanvasPoint) => Array<CanvasCommand>,
+  getMoveCommands: (snappedDragVector: CanvasPoint) => {
```
seanparsons, line 113: Could we make a type for this return value? As it appears in at least two places fully specified.

editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
```diff
@@ -149,7 +150,8 @@ export const absoluteResizeBoundingBoxStrategy: CanvasStrategy = {
                   elementParentBounds,
                   edgePosition,
                 ),
-                setSnappingGuidelines('transient', guidelinesWithSnappingVector),
+                setSnappingGuidelines('transient', guidelinesWithSnappingVector), // TODO I think this will override the previous snapping guidelines
```
seanparsons, line 153: What's the TODO? :)

balazsbajorics: ah yeah, to investigate / make sure we are not making a mistake here

PR #2395, opened 2022-07-01
Highlight elements which will selected on mouse up

gbalint:
Since https://github.com/concrete-utopia/utopia/pull/2389 selection behavior changed, and inside an already selected selection area we only select elements on mouse up.
We also don't have highlight effects on these elements, only on those which are selectable by mouse down.

However, this feels wrong, because you can still select on mouse up, but there is no feedback that you can actually try. 

So in this PR I added back highlight on hover for these elements (and this behavior is in sync with figma too)


enidemi (approved):

seanparsons (approved):

balazsbajorics:
I'm merging this PR as it is approved!

PR #2396, opened 2022-07-01
Undo works in inspector

gbalint:
# Issue

We want undo and redo to work in the editor even when the focus is on an input in the inspector (or elsewhere)

# Solution

For the note, not all keyboard shortcuts are like this, e.g. copy/paste should work inside the text input.

I added a flag to the response from getShortcutActions, with which it is customizable one-by-one whether a shortcut should work when the editor is not the target of the keyboard event.


balazsbajorics (approved):

balazsbajorics:
I'm taking over and adding a test for this

Rheeseyb (approved):

enidemi (approved):

PR #2397, opened 2022-07-04
Drag pan while space is pressed.

seanparsons:
**Problem:**
We have no way currently to drag pan the canvas.

**Fix:**
While the first mouse button is held down with the `space` key also depressed, mouse moves will trigger our existing scroll canvas logic.

Any interaction that might have been started if the mouse down was applied first will be cancelled on the `space` key down. Along with any regular mouse handling being disabled while the `space` key is depressed.

**Commit Details:**
- In `useSelectOrLiveModeSelectAndHover` disable the handlers for
  onMouseMove/onMouseDown/onMouseUp if `space` is pressed down currently.
- In `EditorComponentInner` if `space` is pressed and an interaction session
  is in progress, clear the interaction session so that we don't pan
  and do whatever the interaction would've resulted in.
- Used the `Crosshair` cursor for now while pan dragging.
- In the `on` handler, if the first mouse button is held down along with
  `space`, then on a mouse move trigger the drag panning.
- Remapped the selection cycling previously assigned to the `space` key to the `q` key.

### DIFF
--- a/editor/src/components/canvas/canvas-panning.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-panning.spec.browser2.tsx
@@ -11,15 +11,15 @@ function createExampleProject(): Promise<EditorRenderResult> {
   )
 }
 
-describe(`pan while 'q' is held down`, () => {
-  it(`press 'q' first`, async () => {
+describe(`pan while 'space' is held down`, () => {
+  it(`press 'space' first`, async () => {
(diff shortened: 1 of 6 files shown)


maltenuhn:
@seanparsons see my comment in the paper doc, I think there was a misunderstanding about how to trigger this. Hopefully a minor swap? :)

balazsbajorics (approved):

PR #2398, opened 2022-07-06
Fix/canvas absolute conversion multiselect

enidemi:
**Problem:**
Absolute conversion on multiselect across hierarchy is not keeping the visual position.

**Fix:**
Reparent them the same way as the duplicate and paste does to a common ancestor. There was some confusion about how containing blocks are working in the browser, so `providesBoundsForChildren` is renamed to `providesBoundsForAbsoluteChildren` and has a special helper function to check if the element is the containing block or not. Based on https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block

**Commit Details:**
- escape hatch commands reparent too
- escape hatch adds frames based on the new target parent coordinateSystemBounds
- helper function for elementpaths to reparent the inner elements first
- new browser tests


editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.spec.browser2.tsx
```diff
@@ -0,0 +1,308 @@
...
+                data-uid='ggg'
+              />
+            </div>
+        </div>
+      </div>
+      `),
+    )
+  })
+  /* eslint-disable @typescript-eslint/no-empty-function */
+  xit('Converts element with absolute descendants', async () => {})
```
Rheeseyb, line 307: Is this a TODO?

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.tsx
```diff
@@ -199,55 +201,80 @@ function collectMoveCommandsForSelectedElements(
...
+//   let commands: Array<ConvertToAbsolute | SetCssLengthProperty> = []
+//   let intendedBounds: Array<CanvasFrameAndTarget> = []
+//   siblings.forEach((path) => {
+//     const elementResult = collectSetLayoutPropCommands(path, metadata, canvasState, null)
+//     intendedBounds.push(...elementResult.intendedBounds)
+//     commands.push(...elementResult.commands)
+//   })
+//   return { commands, intendedBounds }
+// }
+
```
Rheeseyb, line 258: Delete me?

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.spec.browser2.tsx
enidemi: yes it's a TODO to not forget writing tests after fixing the other bug with absolute elements

Rheeseyb (approved):

editor/src/core/shared/element-path.ts
```diff
@@ -979,3 +979,15 @@ export function isFocused(focusedElementPath: ElementPath | null, path: ElementP
     return pathUpToElementPath(focusedElementPath, lastPart, 'dynamic-path') != null
   }
 }
+
+export function getSortedPathsDescending(elementPaths: Array<ElementPath>): Array<ElementPath> {
```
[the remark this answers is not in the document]
enidemi: fixed, I renamed it

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -828,6 +828,20 @@ export const MetadataUtils = {
       }, Utils.asLocal(frame))
     }
   },
+  getFrameRelativeToTargetContainingBlock: function (
+    targetParent: ElementPath | null,
+    metadata: ElementInstanceMetadataMap,
+    frame: CanvasRectangle,
+  ): LocalRectangle {
+    const parent = this.findElementByElementPath(metadata, targetParent)
+    if (parent != null && parent.specialSizeMeasurements.coordinateSystemBounds != null) {
```
balazsbajorics, line 837: is `coordinateSystemBounds` the right measurement to use here? that's the bounds of `targetParent`'s containing block!

enidemi: yes, this needs to use the globalFrame of the absolute parent. I'm adding a test for this.

balazsbajorics (approved):

PR #2399, opened 2022-07-06
Handle object fields in an array correctly for property controls.

seanparsons:
**Problems:**
- With an attribute containing an array of objects, when adding a new entry once one field has been entered and the blur triggers the value in the model is updated. As this leaves a partially filled value, when the parse for the field runs it fails for the entry and the whole array in turn. This causes the inspector to lose all but the first entry.
- When clicking from the first field in a complex object held in the array control, a click into the second field appears to be wormholed to nowhere.

**Fixes:**
- The parser for object values has been made slightly less strict, allowing objects with a missing key to succeed.
- Handling for insertion was changed to create an extended array and use that in the inspector, instead of creating a special insertion row in the control. The click was being wormholed as a result of the special insertion row being removed and replaced with the new entry from the updated array, this meant that while nothing visually changed the input field is structurally different and React has destroyed the old HTML element and created a new one.

**Commit Details:**
- Change `RowForArrayControl` to instead use an extended array when inserting a
  new value instead of creating a structurally distinct insertion row.
- `unwrapAndParseObjectValues` checks for the case that the key is missing and
  if it is does not fail the parse and lets the parse continue without that key.
- Added `property-controls.spec.browser2.tsx`.

### DIFF
--- a/editor/src/components/inspector/sections/component-section/component-section.tsx
+++ b/editor/src/components/inspector/sections/component-section/component-section.tsx
@@ -301,17 +301,35 @@ const RowForArrayControl = React.memo((props: RowForArrayControlProps) => {
     () => getSectionHeight(controlDescription),
     [controlDescription],
   )
-  const transformedValue = Array.isArray(value) ? value : [value]
+
+  const [insertingRow, setInsertingRow] = React.useState(false)
+  const toggleInsertRow = React.useCallback(() => setInsertingRow((current) => !current), [])
+
+  // Ensure the value is an array, just in case.
+  const transformedValue = React.useMemo(() => {
+    return Array.isArray(value) ? value : [value]
+  }, [value])
+
+  // If we are inserting, extend the array with an `undefined` value and
+  // then let the handling of the elements cater for that, so that there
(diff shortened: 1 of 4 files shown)


enidemi (approved):

Rheeseyb (approved):

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -301,17 +301,35 @@ const RowForArrayControl = React.memo((props: RowForArrayControlProps) => {
...
   )
-  const transformedValue = Array.isArray(value) ? value : [value]
+
+  const [insertingRow, setInsertingRow] = React.useState(false)
+  const toggleInsertRow = React.useCallback(() => setInsertingRow((current) => !current), [])
+
+  // Ensure the value is an array, just in case.
+  const transformedValue = React.useMemo(() => {
+    return Array.isArray(value) ? value : [value]
+  }, [value])
```
balazsbajorics, line 311: is `value` coming from `useComponentPropsInspectorInfo` referentially stable?

balazsbajorics (commented):
👍

PR #2405, opened 2022-07-07
fix(canvas) absolute conversion keeps absolute descendants in same position

enidemi:
**Fix:**
Converting static elements to absolute where child elements are already absolute often make the child elements jump out of screen. This fix is trying to keep their visual position.

TODO
- [x] filter absolute descendants that already have positioned parents
- [x] filter component children where the selected view is outside of the component

**Commit Details:**
- updated tests to check element position with global frames before and after conversion
- update escape-hatch strategy

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.spec.browser2.tsx
@@ -1,464 +1,344 @@
-import { act, fireEvent } from '@testing-library/react'
+import {
+  renderTestEditorWithCode,
+  TestAppUID,
+  TestScenePath,
+  TestSceneUID,
+} from '../ui-jsx.test-utils'
+
+import * as Prettier from 'prettier/standalone'
+import { PrettierConfig } from 'utopia-vscode-common'
 import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import * as EP from '../../../core/shared/element-path'
 import { localRectangle, LocalRectangle } from '../../../core/shared/math-utils'
-import { runEscapeHatch } from '../../editor/actions/action-creators'
+import { fastForEach } from '../../../core/shared/utils'
+import { runEscapeHatch, setFocusedElement } from '../../editor/actions/action-creators'
 import {
-  getPrintedUiJsCode,
-  makeTestProjectCodeWithSnippet,
-  renderTestEditorWithCode,
-  TestScenePath,
-} from '../ui-jsx.test-utils'
+  BakedInStoryboardUID,
+  BakedInStoryboardVariableName,
+} from '../../../core/model/scene-utils'
+import { ElementPath } from '../../../core/shared/project-file-types'
+import { mapArrayToDictionary } from '../../../core/shared/array-utils'
 
-const absoluteDescendantProject = (absoluteFrame: LocalRectangle) => {
-  return `
-    <div
-      style={{
-        position: 'absolute',
-        width: 100,
-        height: 200,
-        left: 10,
-        top: 20,
-      }}
-      data-uid='aaa'
-    >
+const complexProject = () => {
+  const code = `
+  import * as React from 'react'
+  import { Scene, Storyboard, View } from 'utopia-api'
+
+  export const Card = (props) => {
+    return (
       <div
+        data-uid='card-inner'
         style={{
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.spec.browser2.tsx
```diff
@@ -339,6 +339,247 @@ describe('Convert to Absolute/runEscapeHatch action', () => {
...
+            }}
+            data-uid='ccc'
+          >
+            <div style={{ display: 'flex' }} data-uid='ddd'>
+              <div
+                data-uid='eee'
+                style={{
+                  position: 'absolute',
+                  width: 100,
+                  top: 138,
```
balazsbajorics, line 440: it's very very hard to see that the actual expected change is this one line!
I wonder how we could improve these tests. 
maybe the test should independently assert the props.style.top of aaa/ccc/ddd/eee is 218 at the beginning and becomes 138 after the dispatch

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.tsx
```diff
@@ -416,3 +405,105 @@ function collectHighlightCommand(
...
+  selectedElements: Array<ElementPath>,
+  metadata: ElementInstanceMetadataMap,
+): Array<ElementPath> {
+  const isSelectionInsideFocusedComponent = selectedElements.every((path) =>
+    EP.isInsideFocusedComponent(path),
+  )
+  return mapDropNulls((element) => {
+    const elementPath = element.elementPath
+    const isInsideFocusedComponent = EP.isInsideFocusedComponent(elementPath)
+    const allInsideComponentOrOutside =
```
balazsbajorics, line 419: I think instead of this you want to make sure that the `path.parts` fully match except for the last part, where they are in an ancestor-descendant relationship

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.tsx
```diff
@@ -416,3 +405,105 @@ function collectHighlightCommand(
...
+    const sortedAncestors = EP.getOrderedPathsByDepth(selectedElements).filter((selection) =>
+      EP.isDescendantOf(elementPath, selection),
+    )
+    if (
+      MetadataUtils.isPositionAbsolute(element) &&
+      !EP.isRootElementOfInstance(elementPath) &&
+      sortedAncestors.length > 0 &&
+      allInsideComponentOrOutside
+    ) {
+      const closestAncestor = sortedAncestors[0]
```
balazsbajorics, line 431: isn't this the same as EP.parentPath(elementPath) ?

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.tsx
```diff
@@ -416,3 +405,105 @@ function collectHighlightCommand(
...
+      const containingBlockPath = MetadataUtils.findContainingBlock(metadata, elementPath)
+      const closestAncestorContainingBlockPath = MetadataUtils.findContainingBlock(
+        metadata,
+        closestAncestor,
+      )
+      if (EP.pathsEqual(containingBlockPath, closestAncestorContainingBlockPath)) {
```
Rheeseyb, line 437: I honestly have no idea what the reason is for this check. Please can you add some commenting here.

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.tsx
```diff
@@ -416,3 +405,105 @@ function collectHighlightCommand(
...
+        return elementPath
```
balazsbajorics, line 438: please add some comments to this code about what are the intentions here and what is the logic behind it!

what does it mean that the closest ancestor of the element and the element share the same containing block? why does it mean these are the absolutes we want to update?

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.tsx
Rheeseyb: I was wondering the same thing here. Maybe a victim of some refactoring?

Rheeseyb (approved):
Nice! Thank you very much for taking the time to clarify the code and add those test cases!

PR #2408, opened 2022-07-07
Introducing the Open Hand Cursor

lankaukk:
There is now an open hand cursor available to be used on hover while spacebar is pressed but mouse-button is not down, or in whatever other circumstances we need.


seanparsons (approved):

Rheeseyb (commented):
👍

PR #2414, opened 2022-07-12
Protect against reparenting when a value is defined elsewhere.

seanparsons:
**Problem:**
If an element refers to a value defined externally to the element, then reparenting the element is almost guaranteed to fail as the variable will not exist in the scope of the new location.

**Fix:**
For any potential target of the reparenting logic, first check if the element has any values defined elsewhere.

**Commit Details:**
- Added `attributeReferencesElsewhere`, `jsxAttributesPartReferencesElsewhere`,
  and `elementReferencesElsewhere` functions to capture if any value is defined
  outside of an element.
- `absoluteReparentStrategy` checks if the element attempting to be dragged has
  any values defined elsewhere via `elementReferencesElsewhere`.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.tsx
@@ -1,3 +1,5 @@
+import { foldEither } from '../../../core/shared/either'
+import { elementReferencesElsewhere } from '../../../core/shared/element-template'
 import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import * as EP from '../../../core/shared/element-path'
 import { CSSCursor } from '../canvas-types'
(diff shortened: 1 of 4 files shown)


editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.tsx
```diff
@@ -30,7 +32,19 @@ export const absoluteReparentStrategy: CanvasStrategy = {
       return filteredSelectedElements.every((element) => {
         const elementMetadata = MetadataUtils.findElementByElementPath(metadata, element)
 
-        return elementMetadata?.specialSizeMeasurements.position === 'absolute'
+        const referencesExternalValue =
+          elementMetadata == null
```
Rheeseyb, line 36: Should this strategy just return `false` if there is no metadata? I know it will implicitly because of the `position` check, but I'm wondering if we should make some form a deliberate decision here.

editor/src/core/shared/element-template.ts
```diff
@@ -763,6 +763,63 @@ export function setJSXAttributesAttribute(
...
+export function attributeReferencesElsewhere(attribute: JSXAttribute): boolean {
+  switch (attribute.type) {
+    case 'ATTRIBUTE_VALUE':
+      return false
+    case 'ATTRIBUTE_OTHER_JAVASCRIPT':
+      return (
+        attribute.definedElsewhere.length > 0 ||
+        Object.values(attribute.elementsWithin).some((element) => {
+          return elementReferencesElsewhere(element)
+        })
```
Rheeseyb, line 775: I'm pretty much certain we can skip this inner check, because IIRC the inner `definedElsewhere` will either be coming from this JS block or will be coming from this block's `definedElsewhere`

Rheeseyb (approved):

editor/src/core/shared/element-template.ts
seanparsons: That sounds like an axiom which should be covered by a unit test.

seanparsons: Unit test confirms this axiom to be true: https://github.com/concrete-utopia/utopia/pull/2416

PR #2415, opened 2022-07-12
Easier default project + beaches v3

balazsbajorics:
**Bonus Bonus**
**I'm making the public default project the beaches one as well!**

**Problem:**
It was a very manual and involved task to update the default project.

**Fix:**
The default project can now by copy-pasted from the contents.json of any utopia project. The part that needs copy pasting is the projectContents.

**Bonus:**
I'm also updating the default project to https://utopia.app/p/0fb75a46-public-beaches!-v3/ Beaches V3!

**Note**
I modified RevisionsState so its no longer an enum, so the copy-pasted code is not throwing a parse error.

### DIFF
--- a/editor/src/sample-projects/sample-project-utils.ts
+++ b/editor/src/sample-projects/sample-project-utils.ts
@@ -142,6 +142,9 @@ export function complexDefaultProject(): PersistentModel {
 }
 
 function createBeachesProjectContents(): ProjectContentTreeRoot {
+  // to update the default project here, just delete the big object that is returned here, access the new project's contents.json
+  // using the link https://utopia.pizza/v1/project/<project-id>/contents.json and paste the contents of projectContents down below
+
   return {
     src: {
       directory: {


editor/src/core/shared/project-file-types.ts
```diff
@@ -515,23 +515,25 @@ export function isParsedJSONFailure(result: ParsedJSONResult): result is ParsedJ
 
 export type ParsedJSONResult = ParsedJSONSuccess | ParsedJSONFailure
 
-export enum RevisionsState {
-  ParsedAhead = 'PARSED_AHEAD',
-  CodeAhead = 'CODE_AHEAD',
-  BothMatch = 'BOTH_MATCH',
-}
```
Rheeseyb, line 522: Oh wow, we still had `enum`s?! 😬

Rheeseyb (approved):

gbalint (approved):

PR #2416, opened 2022-07-13
Check that the definedElsewhere property is propagated.

seanparsons:
**Problem:**
As mentioned here:
[https://github.com/concrete-utopia/utopia/pull/2414#discussion_r919086082](https://github.com/concrete-utopia/utopia/pull/2414#discussion_r919086082)
It would be preferable to actually check this axiom.

**Fix:**
Added a unit test which checks the axium that the `definedElsewhere` property does propagate upwards out of `elementsWithin`.

**Commit Details:**
- Added a test to confirm that the `definedElsewhere` property is
  propagated upwards from `elementsWithin`.


Rheeseyb (approved):
Thank you!

PR #2417, opened 2022-07-14
Capture closest offset parent for use in absolute conversion

Rheeseyb:
**Problem:**
Absolute conversion of an element will sometimes require descendants to be offset if those descendants are also positioned absolutely. However, the current logic doesn't correctly capture which element in the dom is the containing block, resulting in edge cases where descendants will be incorrectly offset.

**Fix:**
Capture the closest offset parent path as part of the dom walking, and use that as part of the absolute conversion to determine if a given element's position will be affected by the conversion (by checking if the closest offset parent path is a descendant of the target element).

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.spec.browser2.tsx
@@ -347,61 +347,8 @@ describe('Convert to Absolute/runEscapeHatch action', () => {
 
 describe('Convert to Absolute', () => {
   it('Correctly uses the captured closestOffsetParentPath to determine which elements to update', async () => {
-    const renderResult = await renderTestEditorWithCode(
-      `
-      import * as React from 'react'
-      import { Scene, Storyboard } from 'utopia-api'
-
-      export var App = (props) => {
-        return (
-          <div data-uid='app-root'>
-            <div
-              data-uid='inner-div'
-              style={{ position: 'absolute', top: 100 }}
-            >
-              <div data-uid='immediate-parent'>
-                {props.children}
-              </div>
-            </div>
-          </div>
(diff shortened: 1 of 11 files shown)


enidemi (approved):

editor/src/components/canvas/canvas-strategies/escape-hatch-strategy.spec.browser2.tsx
```diff
@@ -342,3 +344,117 @@ describe('Convert to Absolute/runEscapeHatch action', () => {
     },
   )
 })
+
+describe('Convert to Absolute', () => {
+  it('Correctly uses the captured closestOffsetParentPath to determine which elements to update', async () => {
```
balazsbajorics, line 349: there's already a `describe('Convert to Absolute/runEscapeHatch action'` in this file, which uses the `escapeHatchTestCases`. Can't you make this test be one of those?

I'm asking because Eni and I ran into the problem that the before-after for the large blocks of code is very hard to read – it's almost like a snapshot, I don't know what are the relevant assertions. The comment  `// Converting App should not result in any changes to 'sb/scene/app/child'` is the only clue we can rely on. 
The `escapeHatchTestCases` try to capture the relevant assertions instead. (of course, there's always room for improvement)

Rheeseyb: I've just clarified the test significantly. Hopefully that should help. I was against adding it to the existing block of tests as I feel like that actually complicates this case, but also the assertions around the frame wouldn't always work here (sometimes the frame remains the same even though the element has been incorrectly converted, as it was just given an offset of 0 in both directions). Instead the test now makes it absolutely clear which part of the code is intended to change and how.

balazsbajorics (approved):

PR #2421, opened 2022-07-18
fix(canvas) default to edit mode

enidemi:
**Problem:**
It tricks us frequently that projects open in select-only mode.

**Fix:**
Open projects in edit mode on staging and production too.

**Commit Details:**
- change `editorModelFromPersistentModel` and `createEditorState` to use `selectMode` instead of `selectLiteMode`


seanparsons (approved):

seanparsons:
In other news I often find `selectLiteMode` really confusing.

Rheeseyb:
> In other news I often find `selectLiteMode` really confusing.

I vote we bin it off completely

PR #2426, opened 2022-07-19
Fix/canvas cursor mousedown

enidemi:
**Problem:**
Starting an absolute drag without moving a mouse and pressing alt doesn't switch the cursor to show the duplicate cursor.
If an interaction changes the cursor on mousedown double click selection will stop selecting further elements.

**Fix:**
Fix strategy fitness functions to allow changing the cursor without passing the drag threshold.
The selection fix is a bit trickier, since we allow dragging when there are overlapping siblings and there is already one selected mousedown on it should allow dragging instead of selecting the top element, in this case the selection updates only on mouseup. This is a bit problematic for double click selection in a hierarchy, for this mousedown should be event that updates the selection. To fix that mouseup won't change selection after an interaction is to check the drag threshold is passed already.
I also found a problem of the select-mode-hook where the mouseup created a fresh interaction session.

**Commit Details:**
- added duplicate strategy and flex reorder strategy cursor command without drag threshold
- updated select mode hook
- moved the cursor component to the canvas controls
- to have the cursor component with full screen size it uses a portal

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
@@ -502,6 +502,22 @@ export function useHighlightCallbacks(
   return { onMouseMove }
 }
 
+function getPreferredSelectionForEvent(
+  eventType: 'mousedown' | 'mouseup' | string,
+  isDoubleClick: boolean,
+): 'prefer-selected' | 'dont-prefer-selected' {
+  // mousedown keeps selection on a single click to allow dragging overlapping elements and selection happens on mouseup
+  // with continuous clicking mousedown should select
+  switch (eventType) {
+    case 'mousedown':
+      return isDoubleClick ? 'dont-prefer-selected' : 'prefer-selected'
+    case 'mouseup':
+      return isDoubleClick ? 'prefer-selected' : 'dont-prefer-selected'
+    default:
+      return 'prefer-selected'
+  }
+}
+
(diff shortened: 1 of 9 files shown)


editor/src/components/editor/editor-component.tsx
```diff
@@ -456,7 +456,7 @@ export function EditorComponent(props: EditorProps) {
   )
 }
 
-const EditorCursorComponent = React.memo(() => {
+export const EditorCursorComponent = React.memo(() => {
```
Rheeseyb, line 459: This should probably be moved out of this file now and into the `canvas/controls` folder somewhere

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -567,7 +586,7 @@ function useSelectOrLiveModeSelectAndHover(
               const start = windowToCanvasCoordinates(
                 windowPoint(point(event.clientX, event.clientY)),
               ).canvasPositionRounded
-              if (event.button !== 2) {
+              if (event.button !== 2 && event.type !== 'mouseup') {
```
Rheeseyb, line 589: Well caught!

Rheeseyb (approved):

editor/src/components/editor/editor-component.tsx
enidemi: fixed

seanparsons (approved):

PR #2428, opened 2022-07-20
Reparent drag from absolute to flex.

seanparsons:
**Problem:**
We want to be able to reparent from absolute layouts into a flex parent.

**Fix:**
This follows a combination of the reparent and the flex re-ordering strategies, by moving the element to the flex parent and then (if required) reordering it. Supporting both reordering into the flex elements _and_ a straight reparent into the flex container, the reparent logic attempts to identify if the pointer is hovering over a child of a flex container or a flex container itself.

**Commit Details:**
- Added `absoluteReparentToFlexStrategy` strategy.
- Added `DeleteProperties` canvas command for removing properties
  from a given element.
- Slight formatting tweaks in `dom-lookup.ts`.
- Exported `getReorderIndex`.
- Made `getAllTargetsAtPointAABB` immune to DOM changes.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-reparent-to-flex-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-reparent-to-flex-strategy.tsx
@@ -41,7 +41,7 @@ function reparentTargetFromInteractionSession(
   const reparentResult = getReparentTarget(
     filteredSelectedElements,
     filteredSelectedElements,
-    interactionSession.metadata,
+    strategyState.startingMetadata,
     [],
     canvasState.scale,
     canvasState.canvasOffset,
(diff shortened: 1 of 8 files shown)


editor/src/components/canvas/canvas-strategies/absolute-reparent-to-flex-strategy.tsx
```diff
@@ -0,0 +1,248 @@
...
+    canvasState.openFile,
+    strategyState.startingAllElementProps,
+  )
+  if (reparentResult.newParent == null) {
+    return {
+      ...reparentResult,
+      shouldReorder: false,
+    }
+  } else {
+    // The target is in a flex container.
```
balazsbajorics, line 58: so if the found target parent is _in_ a flex container, then we return the target parent's parent as the actual target parent? 
could you extend the comment here with a brief explanation? I _think_ we will have to change the code in `getReparentTarget` as we move forward with this work

balazsbajorics (approved):

balazsbajorics (approved):

PR #2429, opened 2022-07-20
More Distinct Scene Styling

Rheeseyb:
This PR is based on the previous spike from #2404, but with a header that scales, and slightly different styling.

To be precise, whereas the spike used lighter styling for the header (and made the canvas background lighter to accommodate that), this implementation uses a slightly darker styling. It also removes the highlighting effect from the scene's header (giving the header a background at all times), but still uses a darker background for selection.

image

I've extracted the `checkerboardBackground` from the Inspector and moved it into `UtopiaStyles` (which also includes other backgrounds) so that it could be used by the `SceneComponent`. This change meant having to update a billion snapshots, hence the size of the PR.

**Note:**
There is no scaling applied to the checkerboard background when zooming, meaning the squares will grow when zooming, like so:

image

We can apply scaling if we want to, but it would probably require storing the scale in an atom as the `SceneComponent` doesn't (and arguably shouldn't) have access to the `EditorState`.

### DIFF
--- a/editor/src/components/canvas/__snapshots__/ui-jsx-canvas-bugs.spec.tsx.snap
+++ b/editor/src/components/canvas/__snapshots__/ui-jsx-canvas-bugs.spec.tsx.snap
@@ -16,6 +16,16 @@ exports[`UiJsxCanvas #747 - DOM object constructor cannot be called as a functio
         position: absolute;
         background-color: rgba(255, 255, 255, 1);
         box-shadow: 0px 0px 1px 0px rgba(26, 26, 26, 0.3);
+        background-image: linear-gradient(
+            to bottom left,
+            #e7e7e7 25%,
+            transparent 25%
+          ),
+          linear-gradient(to bottom left, transparent 75%, #e7e7e7 75%),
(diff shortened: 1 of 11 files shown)


seanparsons (approved):

editor/src/components/canvas/__snapshots__/ui-jsx-canvas-bugs.spec.tsx.snap
```diff
@@ -16,6 +16,16 @@ exports[`UiJsxCanvas #747 - DOM object constructor cannot be called as a functio
         position: absolute;
         background-color: rgba(255, 255, 255, 1);
         box-shadow: 0px 0px 1px 0px rgba(26, 26, 26, 0.3);
+        background-image: linear-gradient(
```
balazsbajorics, line 19: image
I think you murdered the github highlighter somehow

Rheeseyb: This wasn't even a github thing - the regular `git diff` did the same! I have never seen this before.

PR #2431, opened 2022-07-21
fix(canvas) removing select-lite

enidemi:
**Fix:**
Removing Select Lite mode completely! On the canvas you can choose between 'edit' and 'live' options.

### DIFF
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -1021,13 +1021,17 @@ export function editorState(
   }
 }
 
+export const StoredStateVersion = 1
+
 export interface StoredEditorState {
+  version: number
   selectedViews: Array<ElementPath>
   mode: PersistedMode | null
 }
 
 export function storedEditorStateFromEditorState(editor: EditorState): StoredEditorState {
   return {
+    version: StoredStateVersion,
     selectedViews: editor.selectedViews,
     mode: convertModeToSavedMode(editor.mode),
   }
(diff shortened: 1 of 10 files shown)


editor/src/components/editor/store/editor-state.ts
```diff
@@ -1043,7 +1043,7 @@ export function mergeStoredEditorStateIntoEditorState(
     return {
       ...editor,
       selectedViews: storedEditorState.selectedViews,
-      mode: storedEditorState.mode ?? EditorModes.selectLiteMode(),
+      mode: storedEditorState.mode ?? EditorModes.selectMode(),
```
Rheeseyb, line 1046: Looks like this will cause issues with existing projects that happened to be left in `SelectLiteMode`, so please can you add a safety check here so that we can default to `SelectMode` if the `storedEditorState.mode` is unrecognised?

enidemi: that's true, fixed!

Rheeseyb (approved):

editor/src/components/editor/store/editor-state.ts
```diff
@@ -1040,10 +1047,16 @@ export function mergeStoredEditorStateIntoEditorState(
   if (storedEditorState == null) {
     return editor
   } else {
+    // for unknown mode types, like select-lite
```
balazsbajorics, line 1050: maybe nitpicky, but I think this would be better as a migration step!

seanparsons (approved):

balazsbajorics (approved):

editor/src/components/editor/store/editor-state.ts
enidemi: I reverted this and added `version` to StoredState instead. `loadStoredState` won't load the stored state if the version is not matching to the latest storedstate version causing it to reset.

PR #2420, opened 2022-07-15
Flex Reparent To Absolute Strategy

balazsbajorics:
This PR enables the Flex to Absolute and Flex to Flex reparent strategies.
The PR still uses cmd - drag as the reparent gesture. (originally the PR also changed some of this behavior, but I changed this back and we'll deal with that in a separate PR).

There's now four reparent strategies: Absolute -> Absolute, Absolute -> Flex, Flex -> Absolute, Flex -> Flex.
To decide which strategy should be active, I made a new helper function called `findReparentStrategy`. 

The PR incorporates three things  from @seanparsons 's PR #2428:
1.  `reparentTargetFromInteractionSession ` is moved to the reparent helpers and renamed to `getReparentTargetForFlexElement`
2. the Flex -> Flex reparent strategy is plagiarizing the Absolute -> Flex strategy :) 
3. the browser2 tests for the two new reparent strategies is a slightly modified copy-pasted version of the Absolute -> Flex test file.

**OH, and** I disabled the Escape Hatch strategy, because cmd - drag on flex elements were clashing with the reparent and reorder!!

Test project (once the staging deploy is done): https://utopia.pizza/p/a62d93a9-imaginary-cotija/?branch_name=feature-flex-reorder-reparent

PR #2432, opened 2022-07-21
getReparentTargetAtPosition doesn't use WindowMousePosition anymore

balazsbajorics:
**Problem:**
As I was debugging tests for #2420 I realized that deep in the reparent code we use the globally stored variable for the current mouse position. This is _incredibly_ bad, because it means the tests do not control this key parameter. This probably also led to some flaky tests, especially when running multiple tests in the same session.

**Fix:**
Just use a property instead.

**Commit Details:**
- Asking for `pointOnCanvas` instead of calculating the point from WindowMousePositionRaw + canvasScale + canvasOffset
- fixing absolute reparent strategy tests that broke because they didn't provide a dragStart value. these were the only tests that failed


gbalint (approved):

enidemi (approved):

Rheeseyb (commented):
👍

PR #2435, opened 2022-07-21
Cater for unreparentable elements.

seanparsons:
**Problem:**
Some elements should or cannot be reparented, like generated elements as they do not potentially make any sense to be reparented.

**Fix:**
Provide a check against elements that are the target of being reparented so that canvas strategies can indicate that the element can't be reparented instead of either failing. This is powered by the helper function `ifAllowedToReparent` which provides the failure case for the reparent operation.

Additionally a fix was required to `anyUnknownOrGeneratedElements` which was not catering for dynamic element paths due to a function it was using supplying a static path for something that should have been a dynamic path.

**Commit Details:**
- Added `ifAllowedToReparent` helper function that falls back to
  a provided function for the regular function if something can
  be reparented.
- Used `ifAllowedToReparent` in the currently existing reparent
  strategies.
- Added `ReparentNotPermitted` cursor entry.
- Added `normalisedDynamicPath` to `NormalisePathSuccess`.
- Modified `normalisePathToUnderlyingTarget` to include the dynamic
  path by including it in the newly added field in `NormalisePathSuccess`.
- Fixed a bug in `anyUnknownOrGeneratedElements` where it was using
  the normalised static path when invoking `getElementOriginType` instead
  of the normalised dynamic path which allows it to distinguish a
  generated value.
- Tweaked the parameter types of `dropFirstPathElement` and
  `DropFirstPathElementResultType` to permit non-static paths.


editor/src/components/canvas/canvas-utils-scene.spec.browser2.tsx
```diff
@@ -198,6 +198,8 @@ describe('moving a scene/rootview on the canvas', () => {
       await dispatchDone
     })
 
+    await wait(5)
```
Rheeseyb, line 201: ?

editor/src/components/canvas/canvas-strategies/absolute-reparent-to-flex-strategy.spec.browser2.tsx
```diff
@@ -404,4 +458,134 @@ describe('Absolute Reparent To Flex Strategy', () => {
...
+    const dragDelta = windowPoint({
+      x: firstFlexChildCenter.x - absoluteChildCenter.x,
+      y: firstFlexChildCenter.y - absoluteChildCenter.y,
+    })
+    await act(() => dragElement(renderResult, 'generatedabsolutechild', dragDelta, cmdModifier))
+
+    await renderResult.getDispatchFollowUpActionsFinished()
+
+    expect(getPrintedUiJsCode(renderResult.getEditorState())).toEqual(
+      makeTestProjectCodeWithSnippet(`
```
Rheeseyb, line 494: It's unclear what (if anything) has changed here from the starting code. Has anything changed?

editor/src/components/canvas/canvas-utils-scene.spec.browser2.tsx
seanparsons: Ah yes, this fixed a weird issue that arose in this branch where the editor was treating the second mouse move as having been handled. Which as designed appears to be correct.

editor/src/components/canvas/canvas-strategies/absolute-reparent-to-flex-strategy.spec.browser2.tsx
seanparsons: Nothing should have changed in this case.

Rheeseyb: In that case can you change this to compare the result against the starting code please so that that is explicit?

editor/src/components/canvas/canvas-utils-scene.spec.browser2.tsx
Rheeseyb: Ah right, can you add a mega short comment here then please so that nobody comes back here and "tidies" it up!

Rheeseyb (approved):

editor/src/components/canvas/canvas-utils-scene.spec.browser2.tsx
balazsbajorics: good shout, otherwise this might end up being copy-pasted to other places where it's not needed too

seanparsons: Fixed in 71215f224.

editor/src/components/canvas/canvas-strategies/absolute-reparent-to-flex-strategy.spec.browser2.tsx
seanparsons: Fixed in 71215f224.

PR #2436, opened 2022-07-22
Fix double click handling on the canvas

Rheeseyb:
**Problem:**
Repeated double clicking on the canvas will often skip over elements, and will also focus a component at the same time as selecting it.

**Fix:**
There were a few related problems here:
- Firstly, `doubleClick` would be `true` on both a mouse up and a mouse down (as we are re-using the mouse handler for both), so I have restricted `doubleClick` to only ever be `true` for a mouse down event.
- As well as this, we were only checking if `event.detail > 0`, i.e. there had been more than one click. This meant that clicking 4 times in a row would be treated as 3 double clicks (rather than two), meaning you will very easily unintentionally skip over elements you were trying to select
- Finally, we use a double click to focus a component, but don't first check whether that component is already selected, so a double click to select the component would instantly focus it too. This should be a separate, deliberate action, so we now first check if the element is selected before focusing it.

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
@@ -567,7 +567,7 @@ function useSelectOrLiveModeSelectAndHover(
 
       // Skip all of this handling if 'space' is pressed or a mousemove happened in an interaction
       if (!(isSpacePressed || hasInteractionSessionWithMouseMoved)) {
-        const doubleClick = event.detail > 1 // we interpret a triple click as two double clicks, a quadruple click as three double clicks, etc
+        const doubleClick = event.type === 'mousedown' && event.detail > 0 && event.detail % 2 === 0
         const selectableViews = getSelectableViewsForSelectMode(event.metaKey, doubleClick)
         const preferAlreadySelected = getPreferredSelectionForEvent(event.type, doubleClick)
         const foundTarget = findValidTarget(
@@ -611,7 +611,9 @@ function useSelectOrLiveModeSelectAndHover(
             updatedSelection = foundTarget != null ? [foundTarget.elementPath] : []
           }
 
-          if (foundTarget != null && doubleClick) {
+          const foundTargetIsSelected = foundTarget?.isSelected ?? false
+
+          if (foundTarget != null && foundTargetIsSelected && doubleClick) {
             // for components without passed children doubleclicking enters focus mode
             const isFocusableLeaf = MetadataUtils.isFocusableLeafComponent(
               foundTarget.elementPath,
@@ -622,7 +624,7 @@ function useSelectOrLiveModeSelectAndHover(
             }
           }
 
-          if (!(foundTarget?.isSelected ?? false)) {
+          if (!foundTargetIsSelected) {
             // first we only set the selected views for the canvas controls
             setSelectedViewsForCanvasControlsOnly(updatedSelection)
 
--- a/editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
@@ -68,100 +68,106 @@ function fireSingleClickEvents(target: HTMLElement, clientX: number, clientY: nu
   )
 }
 
-function fireDoubleClickEvents(target: HTMLElement, clientX: number, clientY: number) {
-  fireEvent(
-    target,
-    new MouseEvent('mousemove', {
-      bubbles: true,
-      cancelable: true,
-      clientX: clientX,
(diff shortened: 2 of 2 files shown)


editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -567,7 +567,7 @@ function useSelectOrLiveModeSelectAndHover(
 
       // Skip all of this handling if 'space' is pressed or a mousemove happened in an interaction
       if (!(isSpacePressed || hasInteractionSessionWithMouseMoved)) {
-        const doubleClick = event.detail > 1 // we interpret a triple click as two double clicks, a quadruple click as three double clicks, etc
+        const doubleClick = event.type === 'mousedown' && event.detail > 0 && event.detail % 2 === 0
```
gbalint, line 570: I believe the standard click/doubleclick handling is to listen to mouse up. That is why you can "undo" a mouse down, if you move away before releasing the button. You can check this e.g. on mac is, all click events are happening on mouse up.
Or we could just use the `click` and `dblclick` events, is there a problem with those?

Rheeseyb: Our selection logic differs between `mousedown` and `mouseup`, so if we tie this to `mouseup` only it completely breaks double click selection. If we _don't_ restrict this to just one of the two then the focusing behaviour is still incorrect (as the `mousedown` "double click" will select the element, and then the `mouseup` "double click" will immediately focus it).

If we instead switch to using `dblclick` we have to add extra logic to check if the related `mousedown` or `mouseup` events changed the selection, which is a whole new can of worms.

gbalint: I checked and even figma uses mouse down to detect double click...
This is strange for me, but I accept it :)

gbalint:
I think it would be nice to have a test for this behavior, I'm not sure how hard is that.

gbalint (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
Rheeseyb: I think this is because of all of the nuances around canvas editing in general - e.g. wanting to select and drag in the same action, or wanting to drag an already selected (but overlapped) element vs wanting to select the overlapping element when there is no drag.

Rheeseyb:
> I think it would be nice to have a test for this behavior, I'm not sure how hard is that.

I'll see what I can do. We already have double click selection tests but I think they missed the problematic behaviour because of the synthetic way of firing the events in those tests

Rheeseyb:
> I think it would be nice to have a test for this behavior, I'm not sure how hard is that.

Done 👍

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -244,16 +250,18 @@ describe('Select Mode Selection', () => {
 
     const canvasControlsLayer = renderResult.renderedDOM.getByTestId(CanvasControlsContainerID)
 
+    const fireDoubleClickEvents = createDoubleClicker()
+
     const doubleClick = async () => {
       await act(async () => {
-        const dispatchDone = renderResult.getDispatchFollowUpActionsFinished()
+        // const dispatchDone = renderResult.getDispatchFollowUpActionsFinished()
```
seanparsons, line 257: Can these comments be removed?

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -507,24 +639,9 @@ describe('Select Mode Advanced Cases', () => {
...
 
-    expect(renderResult.getEditorState().editor.selectedViews).toEqual([
-      EP.fromString(
-        'sb/scene-CardList/CardList-instance:CardList-Root/CardList-Col/CardList-Card~~~1:Card-Root/Card-Row-Buttons/Card-Button-3',
-      ),
-    ])
+    expect(renderResult.getEditorState().editor.selectedViews).toEqual([desiredPath])
   })
 })
-function waitForAnimationFrame(): Promise<void> {
```
seanparsons, line 517: Praise be that this nightmare is gone!

seanparsons (approved):

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
Rheeseyb: Oops! I meant to remove those!

Rheeseyb: Fixed

PR #2442, opened 2022-07-27
Reparent includes imports.

seanparsons:
**Problem:**
Currently if an element is moved to a different file, it can potentially end up failing to render because a required import for the element is not in the file the element ends up in. Which terminates the reparenting interaction at that instant.

**Fix:**
The fix for the above issue is in essence pretty simple, which is to enumerate the imports used by the element being reparented and anything within it. Those imports are adjusted to use absolute paths and then added to the file being reparented into.

**Limitations:**
If a component is not exported and local to a file, but is used in an instance that gets moved, then currently this will still fail. Potentially a solution here is to make it exported, but this should be done in later work.

**Commit Details:**
- Added utility function `emptyStaticElementPathPart` to replace some inline type
  casting of empty arrays.
- Implemented `getImportsFor` which can produce an `Imports` value from
  an existing set of imports, to satisfy importing a particular value.
- Created `createProjectContentsPatch` to handle the difficulty of constructing
  a patch for the `ProjectContentsTreeRoot` type.
- Added a new canvas command of `AddImportsToFile`.
- Added `getReparentCommands` which includes the existing `ReparentElement` command
  as well as for adding imports via the `AddImportsToFile` command.
- Included a value of type `NodeModules` in `InteractionCanvasState` so that it
  can be included alongside `ProjectContentsTreeRoot` as a few functions require
  both.
- Incorporated `getReparentCommands` in various strategies which previously
  just created a `ReparentElement` command.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy-multi-file.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy-multi-file.spec.browser2.tsx
@@ -1,0 +1,389 @@
+import {
+  EditorRenderResult,
+  formatTestProjectCode,
+  getPrintedUiJsCode,
+  renderTestEditorWithCode,
+  renderTestEditorWithProjectContent,
+  TestAppUID,
+  TestSceneUID,
+} from '../ui-jsx.test-utils'
+import { act, fireEvent } from '@testing-library/react'
+import { CanvasControlsContainerID } from '../controls/new-canvas-controls'
+import { offsetPoint, windowPoint, WindowPoint } from '../../../core/shared/math-utils'
+import { cmdModifier, Modifiers } from '../../../utils/modifiers'
+import { PrettierConfig } from 'utopia-vscode-common'
+import * as Prettier from 'prettier/standalone'
+import {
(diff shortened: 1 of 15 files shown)


editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -64,7 +64,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`404`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`382`)
```
Rheeseyb, line 67: This is unexpected

Rheeseyb (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
seanparsons: Yeah, only came up after I merged master into the branch too, but I can't see anything that should've affected it.

enidemi (approved):

PR #2443, opened 2022-07-27
Correctly handle which intrinsics support children.

seanparsons:
**Problem:**
Elements like `img` can be inserted into in the editor which is not correct and causes the content to not show with an error in it's place.

**Fix:**
Prevent the parsed as well as unparsed variants of intrinsic elements that don't support children from being treated as if they can have children.

**Commit Details:**
- In `targetElementSupportsChildren` expand the checking of intrinsic elements
  to parsed HTML elements.
- Added `span` to intrinsincHTMLElementNamesThatSupportChildren`.
- Expanded the tests covering `targetElementSupportsChildren` to include
  parsed as well as unparsed elements.

### DIFF
--- a/editor/src/core/shared/dom-utils.ts
+++ b/editor/src/core/shared/dom-utils.ts
@@ -200,6 +200,7 @@ export const intrinsicHTMLElementNamesThatSupportChildren: Array<string> = [
   'menu',
   'nav',
   'section',
+  'span',
 ]
 
 export function getDOMAttribute(element: Element, attributeName: string): string | null {
--- a/editor/src/core/model/element-metadata-utils.spec.ts
+++ b/editor/src/core/model/element-metadata-utils.spec.ts
@@ -317,30 +317,55 @@ describe('targetElementSupportsChildren', () => {
     }
   }
 
-  it('Returns true for a utopia-api View', () => {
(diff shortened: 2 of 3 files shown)


editor/src/core/shared/dom-utils.ts
```diff
@@ -200,6 +200,7 @@ export const intrinsicHTMLElementNamesThatSupportChildren: Array<string> = [
   'menu',
   'nav',
   'section',
+  'span',
```
balazsbajorics, line 203: this _just_ came up in conversation with @maltenuhn that we should probably not allow inserting/reparenting into a span if it only has text. 
If we don't want to build detection for "if it only has text" right now, I would say that as a first approximation, I would simply not allow inserting/reparenting into a span, period. (which seems to be the case on master?)

maltenuhn: Is this specific to spans or all elements?

seanparsons: So it's only not allowing you to insert into a span because of what has been observed earlier which is that intrinsic elements are being parsed into the `Right` of the `Either`, not the `Left` as the original version of the test for spans were expecting. I'd argue it would then be easier to add in this special case after this rather than changing this PR for now.

PR #2447, opened 2022-07-29
fix(canvas) fix selective render with react hooks

enidemi:
**Problem:**
When you drag an element that has child components with react hooks all child elements are not rendered until you release and there is a console error with `Rendered fewer hooks than expected.`

**Fix:**
Changing the ui-jsx-renderer to always rerender when a component has arbitrary js blocks.

**Video of the problem:** 
https://user-images.githubusercontent.com/4403069/181762748-0784683e-2e98-49fa-b882-f6e80f51bee2.mov

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
@@ -182,7 +182,7 @@ export function createComponentRendererComponent(params: {
       })
     }
 
-    if (utopiaJsxComponent.arbitraryJSBlock != null && shouldUpdate()) {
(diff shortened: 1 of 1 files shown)


balazsbajorics (approved):
just to be absolutely clear, this doesn't change what re-renders, right? just that we always run the "javascript block" of a component that was rendered to begin with, right? but then this component renderer component will return with the previous result, still blocking the render of the subtree, right?

enidemi:
> just to be absolutely clear, this doesn't change what re-renders, right? just that we always run the "javascript block" of a component that was rendered to begin with, right? but then this component renderer component will return with the previous result, still blocking the render of the subtree, right?

I checked it locally with the react dev tools and it's okay. Maybe tests for this problem would be even better.

seanparsons (approved):

Rheeseyb (approved):

PR #2448, opened 2022-07-29
Children property check for reparenting.

seanparsons:
**Problem:**
Assuming every component can have children added to instances of the component can result in components not rendering the children as expected and a reparented value disappearing before the users eyes.

**Fix:**
Now in addition to known good intrinsic elements, we take a heuristic approach to user created components inspecting their definition to see if the `children` property is used within them.

**Commit Details:**
- Implemented `componentUsesProperty` and a series of associated functions
  that walk down through a component definition to identify if a property
  from the `props` parameter is utilised.
- `targetElementSupportsChildren` extended to lookup an underlying
  component and if it exists in the project, to check it with
  `componentUsesProperty` to see if it utilises the `children` property.
- A few smaller changes mostly around passing `projectContents` and
  `nodeModules` through to `MetadataUtils.targetSupportsChildren`.


Rheeseyb (approved):

balazsbajorics (commented):
as we talked in Discord, I think targetSupportsChildren should not take nodeModules as a parameter, and pass in {} internally

enidemi (approved):

balazsbajorics (approved):

PR #2450, opened 2022-08-01
Fix Code Editor Loading Issues

Rheeseyb:
**Problem:**
It is possible for the code editor to finish initialising before VSCode bridge, which can lead to either the loading screen persisting, or the code editor loading but with selection following broken.

**Fix:**
The previous logic for removing the loading screen and marking the code editor as "ready" was reliant on the assumption that the bridge would be up and running before VSCode (which at the time was correct). However, after some optimisations we were now in a situation where there was a race between the two, and if VSCode started first it would mean the sequence of messages sent between the two was incorrect.

This PR introduces the following:
- A "ready" message for both sides, which allows both sides to ensure the other is actually up and running
- A `ClearLoadingScreen` message, which the VS Code extension will send when attempting to load any file regardless of the outcome (the previous logic would send a message after opening the file or failing too, but if that file was already open it would mean the `FileOpened` message might not send, meaning the loading screen would persist)
- A timestamp capturing when a given mailbox was last cleared, which is checked when reading the mailbox to ensure no new messages have been missed (since we filter all messages before the last consumed message number)

These changes make the editor resilient to the above race, so now it doesn't matter whether the VSCode bridge or the extension is initialised first.

### DIFF
--- a/utopia-vscode-common/src/mailbox.ts
+++ b/utopia-vscode-common/src/mailbox.ts
@@ -19,6 +19,7 @@ let outbox: Mailbox
 let onMessageCallback: (message: any) => void
 let lastSentMessage: number = 0
 let lastConsumedMessage: number = -1
+let mailboxLastClearedTimestamp: number = Date.now()
 let queuedMessages: Array<ToVSCodeMessage | FromVSCodeMessage> = []
 const MIN_POLLING_TIMEOUT = 8
 const MAX_POLLING_TIMEOUT = MIN_POLLING_TIMEOUT * Math.pow(2, 6) // Max out at 512ms
@@ -44,6 +45,10 @@ function resetPollingFrequency() {
 
 function lastConsumedMessageKey(mailbox: Mailbox): string {
   return `/${mailbox}_LAST_CONSUMED`
+}
+
+function mailboxClearedAtTimestampKey(mailbox: Mailbox): string {
+  return `/${mailbox}_CLEARED`
 }
 
 function pathForMailbox(mailbox: Mailbox): string {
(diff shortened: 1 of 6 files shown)


utopia-vscode-common/src/mailbox.ts
```diff
@@ -113,7 +118,17 @@ async function waitForPathToExist(path: string, maxWaitTime: number = 5000): Pro
   }
 }
 
+async function checkAndResetIfMailboxCleared(mailbox: Mailbox): Promise<void> {
+  const mailboxClearedAtTimestamp = await getMailboxClearedAtTimestamp(inbox)
+  if (mailboxClearedAtTimestamp > mailboxLastClearedTimestamp) {
+    lastConsumedMessage = -1
```
balazsbajorics, line 124: could you write a comment here, explaining what is happening here? I never touched mailbox.ts and I don't quite understand the implications of lastConsumedMessage being set to -1 here

balazsbajorics (approved):

utopia-vscode-common/src/mailbox.ts
Rheeseyb: I've just added one that should hopefully help here

balazsbajorics: did you push it? :)

balazsbajorics (approved):

utopia-vscode-common/src/mailbox.ts
```diff
@@ -113,10 +118,25 @@ async function waitForPathToExist(path: string, maxWaitTime: number = 5000): Pro
   }
 }
 
+async function checkAndResetIfMailboxCleared(mailbox: Mailbox): Promise<void> {
+  const mailboxClearedAtTimestamp = await getMailboxClearedAtTimestamp(inbox)
+  if (mailboxClearedAtTimestamp > mailboxLastClearedTimestamp) {
+    // The mailbox was cleared since we last polled it, meaning our last consumed message
+    // count is now invalid, and we need to start consuming messages from the beginning again.
+    lastConsumedMessage = -1
```
seanparsons, line 126: I think I'd feel better if this was set to `null` or some other sentinel value like `'never'` to indicate a reset, something that actually indicates the meaning we intend.

seanparsons (approved):

gbalint (approved):

enidemi (approved):

utopia-vscode-common/src/mailbox.ts
Rheeseyb: Good shout. I'm going to do this separately, as I need to merge this now so that I can pull it into the scenes experiment PRs.

PR #2451, opened 2022-08-01
Double click fix for non-dom components

gbalint:
# Issue

If a component returns a fragment as the root element, the element _and_ the component do not appear in the DOM. This causes trouble with double-click to select, it is impossible to enter this component and select its children on the canvas with double click

# Root cause

Fragments do not appear in the dom. Moreover, when the root element of a component is a fragment, the component itself does not appear in the dom.
`getSelectionOrAllTargetsAtPoint`, which returns the possible targets for the mouse position only searches in the dom, so it can not find the component with the root fragment.

# Solution

The solution is not trivial, because searching statically using the ElementPath structure is incorrect, because we need to take z-order into account, which is hard to predict from the parsed model.
So we still need to rely on the dom, but somehow extend the search to other element paths, which do not appear in the dom. So I implemented a hybrid solution: with the old method I try to find the best target. However, that can be too high in the hierarchy.
I implemented another method, which tries to find a target which is not in the dom, but it has a descendant which is in the dom and is a good target.
The better target is the one which is deeper in the hierarchy.

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
@@ -641,12 +641,10 @@ describe('Select Mode Double Clicking', () => {
 })
 
 describe('Select Mode Double Clicking With Fragments', () => {
-  // The below tests are to ensure we're not inadvertently handling clicks too many times,
-  // which could either mean focusing something that wasn't meant to be focused, or skipping
-  // over the target element and selecting something further down the hierarchy.
-  // Each double click should _either_ select the next element down the hierarchy, _or_ focus
-  // the currently selected element. Also, we specifically skip over Scenes, meaning a single
-  // double click with nothing selected will select the first child of a Scene
+  // The below tests are similar to tests in 'Select Mode Double Clicking', but
+  // they use a test project which contain components which renders fragment root
+  // elements. These are special cases because these fragments and their components
+  // do not appear in the dom.
 
   before(() => {
     viewport.set(2200, 1000)
--- a/editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
@@ -640,6 +640,220 @@ describe('Select Mode Double Clicking', () => {
   })
 })
 
+describe('Select Mode Double Clicking With Fragments', () => {
+  // The below tests are to ensure we're not inadvertently handling clicks too many times,
+  // which could either mean focusing something that wasn't meant to be focused, or skipping
+  // over the target element and selecting something further down the hierarchy.
+  // Each double click should _either_ select the next element down the hierarchy, _or_ focus
+  // the currently selected element. Also, we specifically skip over Scenes, meaning a single
+  // double click with nothing selected will select the first child of a Scene
+
+  before(() => {
+    viewport.set(2200, 1000)
+  })
+
+  it('One double clicks to select Card Instance', async () => {
+    // prettier-ignore
+    const desiredPath = EP.fromString(
+      'sb' +            // Skipped as it's the storyboard
+      '/scene-2' +      // Skipped because we skip over Scenes
+      '/Card-instance', // <- First double click
+    )
+
+    const renderResult = await renderTestEditorWithCode(
+      TestProjectAlpineClimb,
+      'await-first-dom-report',
+    )
+
+    const cardSceneRoot = renderResult.renderedDOM.getByTestId('card-scene')
+    const cardSceneRootBounds = cardSceneRoot.getBoundingClientRect()
+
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -640,6 +640,218 @@ describe('Select Mode Double Clicking', () => {
...
+    expect(renderResult.getEditorState().editor.selectedViews).toEqual([desiredPath])
+  })
+
+  it('Two double clicks to select Card Scene Root', async () => {
+    // prettier-ignore
+    const desiredPath = EP.fromString(
+      'sb' +             // Skipped as it's the storyboard
+      '/scene-2' +       // Skipped because we skip over Scenes
+      '/Card-instance' + // <- First double click
+      ':Card-Root',      // <- Second double click, as the instance is automatically focused by the scene
```
balazsbajorics, line 694: oh this is _very nice_ and descriptive

Rheeseyb: Oh thanks :)

gbalint: I agree! I always copy paste from the best ones!

balazsbajorics: > Oh thanks :)

somehow I completely missed #2436, but kudos Rheese, it's _very nice_

Rheeseyb (approved):

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -640,6 +640,218 @@ describe('Select Mode Double Clicking', () => {
...
+      })
+    }
+
+    await doubleClick()
+    await doubleClick()
+    await doubleClick()
+    await doubleClick()
+    await doubleClick()
+    await doubleClick()
+    await doubleClick()
```
enidemi, line 849: The `desiredPath` description says 8 double clicks and I see only 7 here

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
enidemi: this is very easy to read!

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -61,25 +89,43 @@ export function findFirstParentWithValidElementPath(
...
+      if (allowDescendantsFromPath) {
+        for (const staticAndDynamic of staticAndDynamicTargetElementPaths) {
+          if (EP.isDescendantOfOrEqualTo(staticAndDynamic.staticPath, EP.fromString(validPath))) {
+            const depthDiff =
+              EP.navigatorDepth(staticAndDynamic.staticPath) -
+              EP.navigatorDepth(EP.fromString(validPath))
+            let ancestor = staticAndDynamic.dynamic
+            for (let i = 0; i < depthDiff; i++) {
+              ancestor = EP.parentPath(ancestor)
+            }
```
balazsbajorics, line 101: @gbalint I think it's worth turning this into a helper function in element-path.ts, and figuring out what was wrong with the function that's already there and seemed totally broken

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
gbalint: That is just a bug in the comments on the path, I have just fixed it.
Note: we need one less click than with fragments than with divs, because fragment support is not included in this PR. After fragment support is done we can increase this back to 8

editor/src/components/canvas/dom-lookup.ts
gbalint: I extracted a function here: d611beeb58efe534fc680632df3b733bb35a4042
I believe the original `getNthParent` function is not broken, just does something completely different: it goes n levels up in the component hierarchy.
I did not wanna touch that because maybe that is what is needed there. However, I still believe we should make our terminology clearer, let's discuss that offline.

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -61,25 +89,43 @@ export function findFirstParentWithValidElementPath(
...
+          (staticAndDynamic) => staticAndDynamic.static === validPath,
+        )?.dynamic
+      }
     },
     [...validStaticElementPaths],
   )
 
   if (filteredValidPathsMappedToDynamic.length > 0) {
     const sortedFilteredPaths = filteredValidPathsMappedToDynamic.sort(
-      (l, r) => EP.depth(l) - EP.depth(r),
```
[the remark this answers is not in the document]
gbalint: Same as in https://github.com/concrete-utopia/utopia/pull/2451#discussion_r934676621
Let's discuss the terminology offline.

balazsbajorics (approved):

PR #2453, opened 2022-08-02
Fix globalFrame and localFrame for components with root fragments

gbalint:
# Issue

If a component returns a fragment as the root element, the element and the component do not appear in the DOM. This means the dom walker can not reach these elements, and they do not appear in the `domMetadata`
Without `domMetadata`, they don't have `localFrame` and `globalFrame` neither.

# Solution

The solution is to extend the merge logic in `mergeComponentMetadata`.
If an element is not in the dom metadata, but available in the spy metadata, we try to reconstruct its bounding box from its children (or deeper descendants) in the dom.

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -1103,69 +1103,12 @@ export const MetadataUtils = {
       }
     })
 
-    const childrenInDomCache: { [pathStr: string]: Array<ElementInstanceMetadata> } = {}
+    const spyOnlyElements = fillSpyOnlyMetadataWithFramesFromChildren(fromSpy, fromDOM)
 
-    const findChildrenInDomRecursively = (pathStr: string): Array<ElementInstanceMetadata> => {
-      const existing = childrenInDomCache[pathStr]
-
-      if (existing != null) {
-        return existing
-      }
-
-      const spyElem = fromSpy[pathStr]
-      const childrenFromSpy = MetadataUtils.getChildren(fromSpy, spyElem.elementPath)
-      const childrenFromDom = MetadataUtils.getChildren(fromDOM, spyElem.elementPath)
-      const childrenNotInDom = childrenFromSpy.filter((childNotInDom) =>
-        childrenFromDom.every(
-          (childInDom) => !EP.pathsEqual(childNotInDom.elementPath, childInDom.elementPath),
-        ),
-      )
-      const recursiveChildren = childrenNotInDom.flatMap((c) => {
-        return findChildrenInDomRecursively(EP.toString(c.elementPath))
-      })
-      const children = [...childrenFromDom, ...recursiveChildren]
-
-      childrenInDomCache[pathStr] = children
-
-      return children
+    return {
+      ...workingElements,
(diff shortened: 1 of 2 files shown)


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1362,6 +1370,78 @@ export const MetadataUtils = {
...
+    if (children.length === 0) {
+      return
+    }
+    let workingMetadata: ElementInstanceMetadata = { ...spyElem }
+
+    fastForEach(children, (childMetadata) => {
+      const globalFrame =
+        workingMetadata.globalFrame != null
+          ? boundingRectangle(
+              workingMetadata.globalFrame ?? zeroCanvasRect,
```
Rheeseyb, line 1422: Annoying that TS didn't calculate that this `??` isn't necessary here

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1362,6 +1370,78 @@ export const MetadataUtils = {
...
+    const children = findChildrenInDomRecursively(pathStr)
+    if (children.length === 0) {
+      return
+    }
+    let workingMetadata: ElementInstanceMetadata = { ...spyElem }
+
+    fastForEach(children, (childMetadata) => {
+      const globalFrame =
+        workingMetadata.globalFrame != null
+          ? boundingRectangle(
```
Rheeseyb, line 1421: I believe these calculations will be wrong if we're falling back to using `zeroCanvasRect` at any point, since that will mean the bounding rect will always include the point `(0, 0)`, which probably won't be the case most of the time

seanparsons: Do any of the tests cover this case?

seanparsons (approved):

editor/src/core/model/element-metadata-utils.ts
gbalint: The tests cover this case, and it is not a problem, because globalFrame is always null, but childMetadata is never null, because we explicitly choose the children in a way that they have data from the dom walker.
However, you are right that we should fallback to null instead of zero rect.

gbalint: Better and way simpler solution: f569e233e38e07734beded4ab85186b194c3fe8b
I realized there is a function to calculate the bounding box for an array of rectangles (or nulls)

editor/src/core/model/element-metadata-utils.ts
gbalint: Not relevant after f569e233e38e07734beded4ab85186b194c3fe8b

editor/src/core/model/element-metadata-utils.ts
Rheeseyb: Perfect 👍

Rheeseyb (approved):

editor/src/core/model/element-metadata-utils.ts
seanparsons: Oh the implementation in https://github.com/concrete-utopia/utopia/commit/f569e233e38e07734beded4ab85186b194c3fe8b is lovely!

PR #2454, opened 2022-08-03
When reparenting cater for imports of built-in dependencies.

seanparsons:
**Problem:**
When reparenting an element across files, if the file it was being reparented from had an import which related to a built in dependency the reparent operation would throw an exception. This was due to a function being unable to identify the path for the module because built in dependencies do not exist inside the `NodeModules` object.

**Fix:**
The primary fix for this was to create `resolveModulePathIncludingBuiltIns` to allow for these built in dependency cases. However it highlighted that `SampleNodeModules` used in a lot of tests did not represent what `NodeModules` would look like in the running editor as it was full of built in dependencies and no regular dependencies. A sample project needed updating to use a regular dependency so this case was tested project. Our `renderTestEditorWithModel` test function was then also tweaked to handle rendering the project without throwing an exception due to the new requirement on a non-builtin dependency,

Mysteriously with all this a test relating to resizing a scene started breaking again, so the previous fix was reworked to avoid the timing pitfalls of `await wait(10)`.

**Commit Details:**
- Added `resolveModulePathIncludingBuiltIns` to extend on `resolveModulePath`
  as built in dependencies sit outside regular modules in our editor.
- Modified `getImportsFor` to use `resolveModulePathIncludingBuiltIns` so
  that dependencies using imports like `react` don't fail to find them.
- Passed through the `BuiltInDependencies` into a multitude of places
  in the canvas strategies code.
- `renderTestEditorWithModel` was tweaked to incorporate updating the node modules
  alongside loading the project to prevent errors that might occur after the project was
  loaded but didn't have the node modules update yet.
- Fixed some imports that were done from the root by my editor.
- Changed `resetMouseStatus` so that it stops the old request animation
  from and starts another.
- Instead of pausing for an arbitrary amount of time, have a test trigger
  `resetMouseStatus` instead.
- Made the sample project used in a lot of tests use a pretend
  `react-spring` component.
- `SampleNodeModules` is now representative of what would be loaded in
  a real loaded editor, as it previously held a bunch of elements which
  related to built in dependencies.
- Small fixes mostly relating to element paths in tests.

### DIFF
--- a/editor/src/components/canvas/canvas-utils-scene.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-utils-scene.spec.browser2.tsx
@@ -16,6 +16,7 @@ import { BakedInStoryboardUID } from '../../core/model/scene-utils'
 import { CanvasControlsContainerID } from './controls/new-canvas-controls'
 import { wait } from '../../utils/utils.test-utils'
 import { isFeatureEnabled, setFeatureEnabled } from '../../utils/feature-switches'
+import { resetMouseStatus } from '../mouse-move'
 
 describe('moving a scene/rootview on the canvas', () => {
   let originalValue = isFeatureEnabled('Canvas Strategies')
@@ -200,7 +201,7 @@ describe('moving a scene/rootview on the canvas', () => {
 
     // Ensures that the two mouse moves are handled without the check against
     // didWeHandleMouseMoveForThisFrame not triggering.
-    await wait(5)
+    resetMouseStatus()
 
     await act(async () => {
       const dispatchDone = renderResult.getDispatchFollowUpActionsFinished()
--- a/editor/src/components/canvas/ui-jsx.test-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx.test-utils.tsx
@@ -322,7 +322,7 @@ export async function renderTestEditorWithModel(
         numberOfCommits++
       }}
     >
-      <FailJestOnCanvasError />
+      {/* <FailJestOnCanvasError /> */}
       <EditorRoot
         api={storeHook}
         useStore={storeHook}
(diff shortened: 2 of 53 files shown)


editor/src/components/canvas/canvas-utils-scene.spec.browser2.tsx
```diff
@@ -200,7 +201,7 @@ describe('moving a scene/rootview on the canvas', () => {
 
     // Ensures that the two mouse moves are handled without the check against
     // didWeHandleMouseMoveForThisFrame not triggering.
-    await wait(5)
+    resetMouseStatus()
```
balazsbajorics, line 204: 👌

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -322,7 +322,7 @@ export async function renderTestEditorWithModel(
         numberOfCommits++
       }}
     >
-      <FailJestOnCanvasError />
+      {/* <FailJestOnCanvasError /> */}
```
balazsbajorics, line 325: I think this is a leftover change from debugging

seanparsons: Ah, you were too quick! I was still running the tests locally with this uncommented, fixed in 736376f57.

editor/src/components/mouse-move.ts
```diff
@@ -13,8 +13,21 @@ export function mouseWheelHandled() {
   didWeHandleWheelForThisFrame = true
```
balazsbajorics, line 13: hmmm, I wonder if we can now safely delete this whole chrome workaround... we should create a ticket to check if it's still needed

balazsbajorics: I made a ticket

balazsbajorics (approved):
very nice!

editor/src/components/canvas/canvas-utils-scene.spec.browser2.tsx
Rheeseyb: So much better, thank you!

Rheeseyb (approved):

PR #2457, opened 2022-08-03
Automatic Reparent During Drag to Move

balazsbajorics:
https://utopia.pizza/p/25422703-groovy-diplodocus/?branch_name=spike-flex-reparent-target-zones

This PR makes it so that holding CMD is no longer needed to trigger a reparent. Dragging by default starts hunting for possible new parents, and if a reparent is possible, the reparent strategies take over.

The PR introduces a new function `newGetReparentTarget` (needs better name ASAP), a single unified function which is tasked with finding the best reparent target under the cursor, given the array of dragged elements and whether cmd is held (which changes the behavior, cmd is treated as "forced reparent", ie allows reparent to parents that are smaller than the dragged element).
It also tweaks `findReparentStrategy` so that it doesn't initiate reparent to targets that were under the mouse at drag start.

**Details**
- Reparenting to flex now turns `position: absolute` into `position: relative` instead of deleting the position prop.
- Fixing some tests that were related to this
- Brand new `newGetReparentTarget` which handles both flex and absolute reparenting.
- Instead of the flex reparent happening realtime, it uses an insertion placeholder (`FlexReparentTargetIndicator`) to indicate where the dropped element will be inserted. (HMM maybe I should turn this into its own PR?)
- Related to this, a new `displayNoneInstances` is added similar to `hiddenInstances`, which lets us add `display: none` on any element. This means when we reparent _out_ of a flex parent into one of our siblings, we can leave the original element there but with display: none, which means it'll leave an element sized gap so that things don't shift around.
- New `InteractionSession.startingTargetParentToFilterOut` field stores the reparent target under the mouse when a drag begins. We filter this out from the array of potential parents so that a simple drag-to-nudge doesn't immediately turn into a reparent
-  Filtering out `startingTargetParentToFilterOut` broke some tests which were relying on [the case we want to avoid] reparenting to what is under the mouse at drag start.
- Fixing some other tests that broke and it felt like the reason they broke wasn't valid (either they relied on undefined behavior, or outright wrong behavior, unrelated to what they were actually testing)

**Questions for tomorrow Balazs**
- [ ] turn FlexReparentTargetIndicator into a dedicated PR?
- [ ] review some tests where I added `position: relative`, maybe some of those can be reverted now


editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -88,12 +92,13 @@ export function interactionSession(
...
     startedAt: startedAt,
     allElementProps: allElementProps,
+    startingTargetParentToFilterOut: startingTargetParentToFilterOut,
   }
 }
 
 export type InteractionSessionWithoutMetadata = Omit<
   InteractionSession,
-  'metadata' | 'allElementProps'
+  'metadata' | 'allElementProps' | 'startingTargetParentToFilterOut'
```
seanparsons, line 101: I wonder if this is reaching the point where we should just define `InteractionSession` as being this type plus those fields.

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -163,121 +186,361 @@ export function findReparentStrategy(
...
-    filteredSelectedElements,
-    filteredSelectedElements,
-    strategyState.startingMetadata,
-    [],
-    pointOnCanvas,
-    canvasState.projectContents,
-    canvasState.openFile,
-    strategyState.startingAllElementProps,
+  const filteredElementsUnderPoint = allElementsUnderPoint.filter(
+    (target) =>
```
seanparsons, line 304: This condition could do with being broken down a bit as it's giganta-huge.

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -163,121 +186,361 @@ export function findReparentStrategy(
...
+
+      const normalizedStart = Math.min(start, end)
+      const normalizedEnd = Math.max(start, end)
+
+      flexInsertionTargets.push(
+        rectFromTwoPoints(
+          {
+            [leftOrTop]: normalizedStart,
+            [leftOrTopComplement]: parentBounds[leftOrTopComplement],
+          } as any as CanvasPoint,
```
seanparsons, line 507: This typecast makes me very nervous.

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -287,68 +550,295 @@ export function applyFlexReparent(
...
+          updateSelectedViews('on-complete', [newPath]),
         ]
 
         const commandsAfterReorder = [
           ...commandToRemoveProperties,
-          setElementsToRerenderCommand([newPath]),
-          updateHighlightedViews('mid-interaction', []),
-          setCursorCommand('mid-interaction', CSSCursor.Move),
+          setElementsToRerenderCommand([target, newPath]),
+          // updateHighlightedViews('on-complete', []), // TODO WELP
```
seanparsons, line 584: Does this not work anymore?

balazsbajorics: omg it was so long ago. I wish I wrote something more useful here :D I'll try to uncomment it and see what broke

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -287,68 +550,295 @@ export function applyFlexReparent(
...
+          commands: [...midInteractionCommands, ...interactionFinishCommands], // TODO REVIEW
           customState: strategyState.customStrategyState,
         }
       }
     }
     return emptyStrategyApplicationResult
   })
 }
+
+// const absolutePropsToRemove: Array<PropertyPath> = [
```
seanparsons, line 745: Is this completely defunct now?

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -435,6 +447,14 @@ function renderJSXElement(
 function isHidden(hiddenInstances: ElementPath[], elementPath: ElementPath | null): boolean {
   return elementPath != null && hiddenInstances.some((path) => EP.pathsEqual(path, elementPath))
 }
+function isDisplayNoned(
```
seanparsons, line 450:
```suggestion
function anyDisplayNoneInstances(
```

seanparsons (approved):

gbalint:
`newGetReparentTarget` name idea: `getReparentTargetUnified`

enidemi (approved):

PR #2460, opened 2022-08-04
Handle disappearing reparented elements

seanparsons:
**Problem:**
Sometimes when reparenting an element it ends up inside a component that doesn't render its children, potentially based on some condition. So even though we might detect that it utilises the `children` property those children aren't rendered.

**Fix:**
If the element(s) reparented cannot be detected when they have been reparented then the cursor will change to indicate this. The detection is done against the `spyMetadata`, as that loses elements that have not been rendered, by checking against an array produced by the strategy.

**Commit Details:**
- Added `reparentedToPaths` to `CustomStrategyState`.
- Removed the null part of the type union for
  `StrategyApplicationResult.customState`, if a strategy
  wants to return the current custom state it's simpler
  to return it. The `null` made it difficult to clean up
  a value like `reparentedToPaths` if the strategy changed.
- Added `cursorForMissingReparentedItems` to perform the lookup
  associated with a reparented element suddenly not being rendered
  anymore.
- `CursorOverlay` now takes account of the missing reparented item
  case.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
@@ -12,24 +12,24 @@ export interface CustomStrategyState {
   escapeHatchActivated: boolean
   lastReorderIdx: number | null
   duplicatedElementNewUids: { [elementPath: string]: string }
+  reparentedToPaths: Array<ElementPath>
 }
 
-export function defaultCustomStrategyState(): CustomStrategyState {
-  return {
-    escapeHatchActivated: false,
-    lastReorderIdx: null,
-    duplicatedElementNewUids: {},
-  }
+export const defaultCustomStrategyState: CustomStrategyState = {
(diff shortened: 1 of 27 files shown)


editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
```diff
@@ -12,24 +12,24 @@ export interface CustomStrategyState {
...
+  reparentedToPaths: Array<ElementPath>
 }
 
-export function defaultCustomStrategyState(): CustomStrategyState {
-  return {
-    escapeHatchActivated: false,
-    lastReorderIdx: null,
-    duplicatedElementNewUids: {},
-  }
+export const defaultCustomStrategyState: CustomStrategyState = {
```
Rheeseyb, line 18: Can you `deepFreeze` this please so that we catch any accidental mutations that might happen

Rheeseyb (approved):

PR #2462, opened 2022-08-05
Handle disappearing reparented elements.

seanparsons:
**Problem:**
Sometimes when reparenting an element it ends up inside a component that doesn't render its children, potentially based on some condition. So even though we might detect that it utilises the `children` property those children aren't rendered.

**Fix:**
If the element(s) reparented cannot be detected when they have been reparented then the cursor will change to indicate this. The detection is done against the `spyMetadata`, as that loses elements that have not been rendered, by checking against an array produced by the strategy.

**Commit Details:**
- Added `reparentedToPaths` to `EditorStateCanvasControls`.
- Added `cursorForMissingReparentedItems` to perform the lookup
  associated with a reparented element suddenly not being rendered
  anymore.
- Added a new `CanvasCommand` of `SetReparentedToPaths`, along with
  the associated support functions.
- Pulled out the logic for selecting a cursor out of `CursorOverlay`
  into a function `getCursorForOverlay` for easier testing.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.browser2.tsx
@@ -16,12 +16,19 @@ import {
   BakedInStoryboardVariableName,
   BakedInStoryboardUID,
 } from '../../../core/model/scene-utils'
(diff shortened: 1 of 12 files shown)


balazsbajorics (approved):
nice!

Rheeseyb (approved):

enidemi (approved):

PR #2464, opened 2022-08-05
For absolute moves and resizes, check that the target supports style property.

seanparsons:
**Problem:**
Components which don't pass through the `style` property in any way means that those cannot be repositioned/resized by changing the `style` props assigned to an instance of that component.

**Fix:**
The absolute strategies now have an additional filter to their `isApplicable` check which is that the underlying components use the `style` property.

**Commit Details:**
- Added `supportsStyle` utility function to make it convenient
  to check for supporting the style prop in canvas strategies.
- Slightly refactored some code to create the `targetUsesProperty`
  function in `MetadataUtils`.
- Added a `supportsStyle` invocation to all the absolute canvas strategies.
- Added a test case that confirms that resize is disabled for the
  absolute cases.


Rheeseyb (approved):
Nice!

PR #2468, opened 2022-08-09
Correct the ordering of setting the storedState during dispatches

Rheeseyb:
**Problem:**
During the editor's `boundDispatch`, we trigger some zustand `updateStore` calls before updating the value of `this.storedState`. These `updateStore` calls can result in new renders, which can then in turn immediately dispatch new actions, which would be handled before `this.storedState` had been updated, meaning they would be handled based on a previous version of the editor store.

**Fix:**
Update `this.storedState` immediately, and ensure that we are always using that rather than a (potentially outdated) dispatch result throughout the rest of the `boundDispatch` function.

As a side note, we also realised that this is potentially a problem for all actions with handlers that use a passed in `boundDispatch` as they could then be updating the store in a way that would be immediately lost, but when digging into this we found that the only concrete cases of this were calling the passed in `boundDispatch` in an async manner, circumventing the issue.

### DIFF
--- a/editor/src/templates/editor.tsx
+++ b/editor/src/templates/editor.tsx
@@ -370,10 +370,10 @@ export class Editor {
         if (PerformanceMarks) {
           performance.mark(`update canvas ${updateId}`)
         }
-        ElementsToRerenderGLOBAL.current = dispatchResult.patchedEditor.canvas.elementsToRerender // Mutation!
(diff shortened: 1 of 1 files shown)


seanparsons (approved):

balazsbajorics (approved):
this must have been incredibly painful to debug!

seanparsons:
> this must have been incredibly painful to debug!

Yes.

gbalint (approved):

PR #2469, opened 2022-08-10
assets(pointers) - add magic hand pointer

maltenuhn:
**Problem:**
We need a mouse pointer to indicate spooky action at a distance - eg when you're moving something but it's affecting something the parent or grandparent.

**Fix:**
Introducing a mouse pointer for magic, so we can test it.
cursor-magic-hand@2x
.


enidemi (approved):

maltenuhn:
@lankaukk @enidemi I'll just merge this so we can test it now

PR #2470, opened 2022-08-10
Distinguish between item double clicks and label double clicks.

seanparsons:
**Problem:**
Currently double clicking on the navigator item label activates an `onDoubleClick` handler for both the label and the item, which focuses it and activates renaming.

**Fix:**
The renaming is now tied to the `alt` key as a modified, whereas the focusing on the item only works if the `alt` key is not pressed down.

**Notes:**
This will function slightly weirdly until https://github.com/concrete-utopia/utopia/pull/2468 is merged.

**Commit Details:**
- Renamed `ItemLabelProps.canRename` to `ItemLabelProps.selected` as
  that's what it actually represents.
- `ItemLabel` `onDoubleClick` handler requires the alt key to be depressed.
- `NavigatorItem` `onDoubleClick` handler requires that the alt key is not
  depressed to work.

### DIFF
--- a/editor/src/components/navigator/navigator-item/item-label.tsx
+++ b/editor/src/components/navigator/navigator-item/item-label.tsx
@@ -12,7 +12,7 @@ interface ItemLabelProps {
   dispatch: EditorDispatch
   target: ElementPath
   isDynamic: boolean
-  canRename: boolean
+  selected: boolean
   name: string
   suffix?: string
   inputVisible: boolean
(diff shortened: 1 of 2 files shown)


enidemi (approved):

editor/src/components/navigator/navigator-item/item-label.tsx
```diff
@@ -12,7 +12,7 @@ interface ItemLabelProps {
   dispatch: EditorDispatch
   target: ElementPath
   isDynamic: boolean
-  canRename: boolean
+  selected: boolean
```
Rheeseyb, line 15: This prop doesn't appear to be used anywhere?

Rheeseyb (approved):

editor/src/components/navigator/navigator-item/item-label.tsx
seanparsons: Hah, I might as well just delete it then.

PR #2473, opened 2022-08-10
Chore/strategy lifecycle

balazsbajorics:
**Problem:**
Today commands can either be 'transient' or 'permanent'. Transient commands are only active during the interaction, 'permanent' commands run during the interaction, and at the end of the interaction too.

During the reparent spike, I realized that there is a case for third type of command which only runs at the end of the interaction, but not _during_ the interaction.

**Fix:**
This PR changes `transientOrNot` to `whenToRun`, and extends the type:
`'transient'` becomes `'mid-interaction'`
`'permanent'` becomes `'always'`
and I introduce a new `'on-complete'`.

**Commit Details:**
- `transientOrNot` -> `whenToRun`
- adding `commandLifecycle: 'mid-interaction' | 'end-interaction',` to the foldAndApplyCommands... family of functions

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx
@@ -186,7 +186,7 @@ function reparentElement(
     [],
     [],
     strategyResult.commands,
-    'always',
+    'end-interaction',
   ).editorState
 
   return finalEditor
--- a/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx
@@ -186,7 +186,7 @@ function reparentElement(
     [],
     [],
     strategyResult.commands,
-    'permanent',
+    'always',
   ).editorState
 
   return finalEditor
--- a/editor/src/components/canvas/canvas-strategies/flex-reparent-to-absolute-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/flex-reparent-to-absolute-strategy.tsx
@@ -2,7 +2,7 @@ import { BuiltInDependencies } from '../../../core/es-modules/package-manager/bu
 import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import { canvasPoint, offsetPoint, rectContainsPoint } from '../../../core/shared/math-utils'
 import { EditorState, EditorStatePatch } from '../../editor/store/editor-state'
-import { foldAndApplyCommandsInner, TransientOrNot } from '../commands/commands'
+import { foldAndApplyCommandsInner, WhenToRun } from '../commands/commands'
 import { updateFunctionCommand } from '../commands/update-function-command'
 import { ParentBounds } from '../controls/parent-bounds'
 import { ParentOutlines } from '../controls/parent-outlines'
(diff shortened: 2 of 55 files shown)


editor/src/components/canvas/canvas-strategies/flex-reparent-to-absolute-strategy.tsx
```diff
@@ -104,7 +104,7 @@ function runAbsoluteReparentStrategyForFreshlyConvertedElement(
   editorState: EditorState,
   strategyState: StrategyState,
   interactionState: InteractionSession,
-  transient: TransientOrNot,
+  commandLifecycle: 'mid-interaction' | 'end-interaction',
```
Rheeseyb, line 107: Should this be `commandLifecycle: 'mid-interaction' | 'always',`?

Rheeseyb: Ah sorry, I misunderstood. Disregard!

Rheeseyb (approved):

editor/src/components/canvas/commands/commands.ts
```diff
@@ -51,10 +51,10 @@ export interface CommandFunctionResult {
 
 export type CommandFunction<T> = (editorState: EditorState, command: T) => CommandFunctionResult
 
-export type TransientOrNot = 'transient' | 'permanent'
+export type WhenToRun = 'mid-interaction' | 'always' | 'on-complete'
```
balazsbajorics, line 54: this file has most of the real changes that are not just renames

enidemi (approved):

editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx
```diff
@@ -186,7 +186,7 @@ function reparentElement(
     [],
     [],
     strategyResult.commands,
-    'permanent',
```
seanparsons, line 189: A few of these specs have something like this that has changed from `'permanent'` to `'end-interaction'`, should they really be `'mid-interaction'`?

seanparsons: Correction, `'always'`, not `'mid-interaction'`.

editor/src/components/canvas/commands/commands.ts
```diff
@@ -163,22 +163,33 @@ export function foldAndApplyCommandsInner(
   let workingCommandDescriptions: Array<CommandDescription> = []
 
   const runCommand = (command: CanvasCommand, shouldAccumulatePatches: boolean) => {
-    if (transient === 'transient' || command.transient === 'permanent') {
+    const shouldRunCommand: boolean = (() => {
```
seanparsons, line 166: I think this is clearer just making a `let` variable and assigning it in either branch of the if statement.

editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.tsx
balazsbajorics: we used the same type for the command and for the foldAndApplyCommand... functions.

in this PR I split the type: 
1. The command object has `WhenToRun = 'mid-interaction' | 'always' | 'on-complete'`, 
2. and the actual command _runner_ functions now have `commandLifecycle: 'mid-interaction' | 'end-interaction'`.

seanparsons: Ah, I think them both having `mid-interaction` is the root of my confusion.

balazsbajorics: > Ah, I think them both having `mid-interaction` is the root of my confusion.

should I try to unify the 'on-complete' and 'end-interaction' names somehow? I think 'on-complete' doesn't work as the lifecycle's name....

seanparsons: No I think it makes sense as it is, the terms make sense in their respective locations.

gbalint (approved):

editor/src/components/canvas/commands/commands.ts
balazsbajorics: sigh, I keep wishing that this (()=>{})() pattern somehow becomes better, but it never does

PR #2474, opened 2022-08-10
adjusted pin icons and magic cursor

lankaukk:
Tweaked the Pin icons for better optics. Also tweaked the Magic Hand cursor for better placement and contrast. See the difference in _Files changed_


maltenuhn:
That's it :)

seanparsons (approved):

maltenuhn (approved):

PR #2475, opened 2022-08-11
Fix/canvas strategies delay

enidemi:
**Problem:**
When double click selecting elements in a deep hierarchy strategy canvas controls show up immediately when jumping through elements. It should be behind a short timeout.

**Fix:**
Added a new hook to show the controls for the active strategy with a small delay before it appears the first time (switching strategies by the user should show controls immediately). Updated `useSelectorWithCallback` to work with null values.

**Commit Details:**
- updated `useSelectorWithCallback`
- added new test + fix editorstate in other store-hook tests
- new hook for canvas controls `useDelayedCurrentStrategy`

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
@@ -259,6 +259,7 @@ export const useDelayedCurrentStrategy = () => {
   const maybeDelayedCallback = React.useCallback(
     (currentStrategy: CanvasStrategyId | null) => {
       if (currentStrategy != null && delayedStrategyValue == null) {
+        // timer reset is not needed because the hierarchy selection mouseup clears interactionsession and activestrategy
         if (timer == null) {
           setTimer(
             window.setTimeout(() => {
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
@@ -235,7 +235,7 @@ export function applyCanvasStrategy(
   return strategy.apply(canvasState, interactionSession, strategyState)
 }
 
-const useDelayedCurrentStrategy = () => {
+export const useDelayedCurrentStrategy = () => {
   /**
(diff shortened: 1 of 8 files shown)


balazsbajorics:
I've tried it with this project https://utopia.pizza/p/48201b1e-habitual-bicycle/?branch_name=fix-canvas-strategies-delay

so double click drilling is definitely fixed and not blinky. that's good...
however, when drag-to-moving the element around, it feels weird that the parent outline appears with a delay. in this scenario I think all the controls should be immediate

it's kind of a tug-of-war between these two desired behaviors.

I'd suggest trying a few things:
1. as soon as the mouse drag threshold is exceeded, immediately show all the relevant controls, because the strategy is now "active active". 
2. maybe this fix should specifically be about the double clicks not being blinky, so the delay should be 0 _unless_ the user recently changed the selected element.

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -234,12 +235,41 @@ export function applyCanvasStrategy(
   return strategy.apply(canvasState, interactionSession, strategyState)
 }
 
+const useDelayedCurrentStrategy = () => {
+  const [currentStrategyValue, setCurrentStrategyValue] = React.useState<CanvasStrategyId | null>(
```
seanparsons, line 239: Maybe slightly rename this, as `currentStrategyValue` seems a bit too close to `currentStrategy` to me.
```suggestion
  const [previousStrategy, setPreviousStrategy] = React.useState<CanvasStrategyId | null>(
```

seanparsons: Maybe `previous` isn't a great prefix, perhaps `delayed`?

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
enidemi: thanks! updated the names in ab1b91ed7

PR #2476, opened 2022-08-11
Unified reparent-to-flex apply function

balazsbajorics:
**Problem:**
The flex-reparent-to-flex and absolute-reparent-to-flex apply functions were almost identical, except the absolute-reparent-to-flex strategy strips absolute props. 

**Fix:**
Create a shared apply function with a flag to strip absolute props or not.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
@@ -164,7 +164,7 @@ export function getReparentTargetForFlexElement(
   }
 }
 
-const propertiesToRemove: Array<PropertyPath> = [
+const absolutePropsToRemove: Array<PropertyPath> = [
   PP.create(['style', 'position']),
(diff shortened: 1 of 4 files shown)


editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -148,3 +163,104 @@ export function getReparentTargetForFlexElement(
     }
   }
 }
+
+const propertiesToRemove: Array<PropertyPath> = [
```
seanparsons, line 167: Now this isn't in a specific strategy, it could probably do with a better name.

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
balazsbajorics: fixed!

enidemi (approved):

PR #2477, opened 2022-08-11
Validate the use of style props during reparenting.

seanparsons:
**Problem:**
If a component does not carry positional properties (like `left`) from the `style` attribute into the root element making that component up, then it very likely will not be repositionable. We want something that can't be repositioned to also not be reparentable as well.

**Fix:**
Some very basic static analysis is performed to see if the appropriate properties are carried into the root element's style attribute or if the `style` attribute of the component props are spread into the `style` attribute of the root element.

**Commit Details:**
- Implemented `componentHonoursPropsPosition` and `componentHonoursPropsSize`
  which check if a component honours the position and size properties
  respectively by passing them into the root element.
- Created `propsStyleIsSpreadInto` which checks if the attributes provided
  have `props.style` spread into them.
- Added `propertyComesFromPropsStyle` which analyses the attributes of
  an element to see if a certain property has been utilised from the props
  parameter.
- Added `targetHonoursPropsSize` and `targetHonoursPropsPosition` to
  `MetadataUtils` which provide a higher level call over the
  `componentHonours...` functions.
- Added `honoursPropsPosition` and `honoursPropsSize` to simplify
  checking for canvas strategies.
- Prevent component instances from being reparented if they do not honour
  the position properties in the root element of the component.

### DIFF
--- a/editor/src/core/model/element-template-utils.spec.ts
+++ b/editor/src/core/model/element-template-utils.spec.ts
@@ -378,6 +378,66 @@ export const TestComponent = (props) => {
     const result = componentHonoursPropsPosition(component)
     expect(result).toEqual(true)
   })
+  it('returns true for a component that uses top and right directly', () => {
+    const component = getComponentFromCode(
+      'TestComponent',
+      `
+export const TestComponent = (props) => {
+  return <div style={{position: 'absolute', right: props.style.right, top: props.style.top}}>The Best Test Component</div>
+}
+    `,
+    )
+    const result = componentHonoursPropsPosition(component)
+    expect(result).toEqual(true)
+  })
+  it('returns false for a component that uses left and right only', () => {
+    const component = getComponentFromCode(
+      'TestComponent',
+      `
+export const TestComponent = (props) => {
+  return <div style={{position: 'absolute', left: props.style.left, right: props.style.right}}>The Best Test Component</div>
+}
+    `,
+    )
+    const result = componentHonoursPropsPosition(component)
+    expect(result).toEqual(false)
+  })
+  it('returns false for a component that uses top only', () => {
+    const component = getComponentFromCode(
+      'TestComponent',
+      `
+export const TestComponent = (props) => {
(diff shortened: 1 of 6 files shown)


editor/src/core/model/element-template-utils.spec.ts
```diff
@@ -362,3 +364,117 @@ export const TestComponent = (props) => {
...
   })
 })
+
+describe('componentHonoursPropsPosition', () => {
+  it('returns true for a component that uses top and left directly', () => {
+    const component = getComponentFromCode(
+      'TestComponent',
+      `
+export const TestComponent = (props) => {
+  return <div style={{position: 'absolute', left: props.style.left, top: props.style.top}}>The Best Test Component</div>
```
balazsbajorics, line 374: aww The Best Test Component

editor/src/core/model/element-template-utils.spec.ts
```diff
@@ -362,3 +364,117 @@ export const TestComponent = (props) => {
     expect(result).toEqual(true)
   })
 })
+
+describe('componentHonoursPropsPosition', () => {
```
balazsbajorics, line 368: could you insert a few test cases for false results? just to make sure the function doesn't accidentally regress into "always true"

balazsbajorics (approved):
nice! (just in case you don't plan it already) I think a follow-up PR should also disable moving and resizing strategies for elements that don't respect the relevant props

seanparsons:
> nice! (just in case you don't plan it already) I think a follow-up PR should also disable moving and resizing strategies for elements that don't respect the relevant props. IIRC the Delta-based absolute move doesn't need the size props, and there used to be a delta-based absolute resize but maybe that has been removed.

That seems like a natural progression, I thought I would keep it out of this PR since that's not the purpose of it.

editor/src/core/model/element-template-utils.spec.ts
seanparsons: Additional test cases added in 6744435ba.

PR #2478, opened 2022-08-11
Fix/visible except when other strategy

balazsbajorics:
**Problem:**
In figma and sketch, outline and resize controls disappear during drag to move.
https://screenshot.click/figma_drag_hides_canvas_controls.mp4

In utopia, they do not: https://screenshot.click/what_about__utopia.mp4

**Fix:**
https://screenshot.click/outline_visible_during_drag.mp4

**Commit Details:**
- Fix previously broken implementation for `visible-except-when-other-strategy-is-active` control type
- AbsoluteResizeControl and ZeroSizeResizeControlWrapper becomes `visible-except-when-other-strategy-is-active`
- PinLines should only be visible if there's no active strategy

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-resize-bounding-box-strategy.tsx
@@ -58,11 +58,15 @@ export const absoluteResizeBoundingBoxStrategy: CanvasStrategy = {
     }
   },
   controlsToRender: [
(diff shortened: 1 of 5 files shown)


enidemi (approved):

seanparsons:
Could you include a test(s) that checks this new case?

seanparsons (approved):

balazsbajorics:
> Could you include a test(s) that checks this new case?

I'll write tests in a follow-up PR, but it turns out to write nice tests I'll have to move some things around

PR #2480, opened 2022-08-16
Handle dragging into the gap between flex elements.

seanparsons:
**Problem:**
Dragging a component instance into a flex parent in the spaces where no flex child resides currently results in the instance being reparented in on the end of the existing flex children. But there are spaces like this between elements if attributes like `row-gap` are set, which results in the reparented instances jumping around.

**Fix:**
Rather than checking if the mouse position coincides with an existing element, we check for the closest existing element and then identify if the dragged element falls on the left or top "side" versus the right or bottom "side" as that means the reorder might need to be shifted to after the target versus being before the target.

**Commit Details:**
- `getReparentTargetForFlexElement` now indicates that reparenting
  into a flex parent should trigger a reorder.
- `getReorderIndex` now determines the index based on the distance
  between the mouse and the center point of the siblings.
- Corrected some tests that didn't line up with the changes introduced.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
@@ -1,7 +1,12 @@
 import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import * as EP from '../../../core/shared/element-path'
 import { ElementInstanceMetadataMap } from '../../../core/shared/element-template'
-import { CanvasVector, offsetPoint, rectContainsPoint } from '../../../core/shared/math-utils'
+import {
+  CanvasPoint,
+  CanvasVector,
+  offsetPoint,
+  rectContainsPoint,
+} from '../../../core/shared/math-utils'
 import { ElementPath, PropertyPath } from '../../../core/shared/project-file-types'
 import * as PP from '../../../core/shared/property-path'
 import { CSSCursor } from '../canvas-types'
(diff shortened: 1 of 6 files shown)


editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -22,22 +27,88 @@ import { InteractionSession, StrategyState } from './interaction-state'
...
-    return (
+    if (
       frame != null &&
-      rectContainsPoint(frame, point) &&
       MetadataUtils.isParentYogaLayoutedContainerAndElementParticipatesInLayout(sibling, metadata)
-    )
-  })
+    ) {
+      const centerPoint = Utils.getRectCenter(frame)
+      const distance = Utils.distance(point, centerPoint)
```
Rheeseyb, line 78: I believe this distance should only take into account the relevant dimension based on the value of `rowOrColumn`.

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
seanparsons: That only works if there's only a single row or column, if you imagine 9 elements that wrap to 3 rows of 3 elements then just the horizontal distance will make it jump around between the 3 rows.

Rheeseyb: Ah yeah ok that makes sense 👍

PR #2481, opened 2022-08-17
Resize and move strategies check if properties are honoured.

seanparsons:
**Problem:**
Resize and move strategies currently only check if an underlying component utilises the `style` property. But recent work for reparenting has gone beyond that and now checks for the use of the appropriate position and size properties.

**Fix:**
Replaced the old `supportsStyle` function usage with the new functions recently implemented for reparenting.

**Commit Details:**
- `absoluteMoveStrategy`, `absoluteResizeBoundingBoxStrategy`,
  `keyboardAbsoluteMoveStrategy` and `keyboardAbsoluteResizeStrategy`
  now utilise `honoursPropsPosition` and `honoursPropsSize` instead of
  `supportsStyle`.


Rheeseyb (approved):
Nice!

PR #2484, opened 2022-08-19
Navigator reparent/reorder now re-uses the canvas strategies logic.

seanparsons:
**Problem:**
In the editor there's a divergence of behaviour where the navigator reparent/reorder logic is different to that used by the new canvas strategies. Ideally this would not be the case so that changes don't need to be repeated or potentially not updated in one or more locations.

**Fix:**
These changes take two main parts:
- Surface level functionality around checking if elements can be reparented or reparented into, which uses existing functions to perform those checks before a reparent/reorder interaction.
- Handling of the reparent/reorder interaction after it has been triggered which has involved rewriting the function that does most of the work in handling the existing navigator reparent editor action.

**Todo:**
While it will not change the implementation itself meaningfully, some browser tests will be added to this or in another PR subsequently to extend the existing navigator reparent tests.

**Commit Details:**
- Added utility function `shiftIndexPositionForRemovedElement` to handle
  the change that is necessary when using an `IndexPosition` which relates
  to a collection where the reordered element exists already in the array.
- Corrected the `canReparent` value used in the navigator item dnd container
  `onHover` function as it was actually checking the dragged item not the target
  item.
- Implemented `canDrag` and `canDrop` for `NavigatorItemContainer` to have them
  match up with the same checks that are implemented in the canvas.
- `editorMoveMultiSelectedTemplates` reimplemented to use `getReparentOutcome`
  and `reorderElement` to handle the update which originates from the navigator
  action.
- `ReorderElement` now uses `IndexPosition` instead of an absolute index.
- `reorderComponent` now uses `shiftIndexPositionForRemovedElement` to adjust
  the index position for the case when the element being removed makes the position
  invalid.
- `getReparentOutcome` is the newly renamed `getReparentCommands`, now avoids
  creating the commands to move something if it doesn't actually move. As well it
  now also returns the new path for the element in its reparented position.
- Refactored `isAllowedToReparent` to make it not be constrained to the use of
  canvas strategies alone.

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
@@ -21,6 +21,9 @@ import { BasePaddingUnit, getElementPadding, NavigatorItem } from './navigator-i
 import { NavigatorHintBottom, NavigatorHintTop } from './navigator-item-components'
 import { JSXElementName } from '../../../core/shared/element-template'
 import { DropTargetHint, ElementWarnings } from '../../editor/store/editor-state'
+import { useRefEditorState } from '../../../components/editor/store/store-hook'
+import { isAllowedToReparent } from '../../../components/canvas/canvas-strategies/reparent-helpers'
+import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 
 const BaseRowHeight = 35
 const PreviewIconSize = BaseRowHeight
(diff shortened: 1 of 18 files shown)


enidemi (approved):

Rheeseyb (approved):

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -134,13 +141,7 @@ function onHover(
     const targetAction = propsOfDraggedItem.highlighted
       ? []
       : [EditorActions.setHighlightedView(propsOfDraggedItem.elementPath)]
-    const canReparent = propsOfDraggedItem
-      .getDragSelections()
-      .every(
-        (dragSelectedItem: DragSelection) =>
-          !EP.pathsEqual(propsOfDraggedItem.elementPath, dragSelectedItem.elementPath) &&
```
Rheeseyb, line 141: We probably still want this specific line to prevent accidentally dropping some selection onto one of the selected elements

Rheeseyb: (Except it should definitely be checking against the drop target...)

seanparsons: It checking the wrong thing was at the root of this not working correctly.

Rheeseyb (approved):

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
seanparsons: Fixed in 28f7c8d06.

PR #2485, opened 2022-08-22
Click to insert

gbalint:
This is a very simple click-to-insert implementation based on the already existing drag-to-insert solution.
Click to insert was already allowed for utopia-api text and image components. For everything else, it was explicitly disallowed.
I just allowed it for everything, and added a default width-height, so they are not inserted it with zero dimensions.
I also added a new test case for click to insert.

### DIFF
--- a/editor/src/components/canvas/controls/insert-mode-control-container.tsx
+++ b/editor/src/components/canvas/controls/insert-mode-control-container.tsx
@@ -54,6 +54,9 @@ import { isSceneFromMetadata } from '../../../core/model/project-file-utils'
 import { RightMenuTab } from '../../editor/store/editor-state'
 import { stylePropPathMappingFn } from '../../inspector/common/property-path-hooks'
 
+const DefaultWidth = 100
+const DefaultHeight = 100
+
 // I feel comfortable having this function confined to this file only, since we absolutely shouldn't be trying
 // to set values that would fail whilst inserting elements. If that ever changes, this function should be binned
 // and we should handle those failures explicitly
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/controls/insert-mode.spec.browser2.tsx
```diff
@@ -182,7 +182,7 @@ describe('Inserting', () => {
                 position: 'absolute',
                 left: 5,
                 top: 5,
-                width: 20,
+                width: 100,
```
enidemi, line 185: why did the width change here?

enidemi (approved):

editor/src/components/canvas/controls/insert-mode.spec.browser2.tsx
Rheeseyb: Yeah... this is suspicious

editor/src/components/canvas/controls/insert-mode-control-container.tsx
```diff
@@ -249,6 +252,11 @@ export class InsertModeControlContainer extends React.Component<
     )
     const localFrame = Utils.getLocalRectangleInNewParentContext(parentOrigin, dragFrame)
 
+    if (localFrame.width === 0 && localFrame.height === 0) {
+      localFrame.width = DefaultWidth
+      localFrame.height = DefaultHeight
+    }
```
Rheeseyb, line 258: Presumably in this case we're inserting with the top-left matching the click point. If that's the case, we should probably centre the inserted element on the click here.

editor/src/components/canvas/controls/insert-mode.spec.browser2.tsx
gbalint: Yes, that was commited accidentally (and broke the tests)

editor/src/components/canvas/controls/insert-mode-control-container.tsx
gbalint: Good point, fixed in a8d97c735ed175d05d861d8c2035cec0daceae60 and 3983901e60b13544f97ac334ee6fdf4a002f22f1

seanparsons (approved):

Rheeseyb (approved):

PR #2487, opened 2022-08-24
Feature/canvas selection lock

enidemi:
**Fix:**
Introducing selection locking on the canvas, it can be accessed in the navigator with the lock icon. 
Toggling has 3 modes:
1. selectable - unlocked icon
2. only the toggled element is locked (but children are selectable)
3. all of its descendants are locked too and they are not selectable on the canvas.

Scenes cannot be locked because the icon is not shown in the navigator for them.

**Commit Details:**
- added action `TOGGLE_SELECTION_LOCK`
- 2 browser tests
- new component in navigator `SelectionLockedIndicator`
- update `getSelectableViews` with locking and cleaning up old dynamic scene related filtering as it's not used anymore
- also cleaned up helper function `isSceneTreatedAsGroup`, this was also for the old dynamic scenes

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
@@ -26,6 +26,7 @@ import {
   AllElementProps,
   EditorState,
   EditorStorePatched,
+  LockedElements,
 } from '../../../editor/store/editor-state'
 import { useEditorState, useRefEditorState } from '../../../editor/store/store-hook'
 import CanvasActions from '../../canvas-actions'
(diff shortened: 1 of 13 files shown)


editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -140,60 +141,91 @@ function filterHiddenInstances(
...
 
     candidateViews = uniqueSelectableViews
   }
 
-  return filterHiddenInstances(hiddenInstances, candidateViews)
+  const nonSelectableElements = [
+    ...hiddenInstances,
+    ...getAllLockedElementPaths(componentMetadata, lockedElements),
+  ]
+  return filterHiddenInstances(nonSelectableElements, candidateViews)
```
Rheeseyb, line 228: Very minor, but this function should now be renamed

Rheeseyb (approved):

gbalint (approved):

PR #2495, opened 2022-08-26
Change the paste handler to use the canvas strategies logic.

seanparsons:
**Problem:**
Similar to what was done for the navigator we want pasting to use the underlying canvas strategies transformations when an element is pasted into a parent which potentially has a wildly different setting for `display`.

**Fix:**
The work for this consists of the following main elements:

1) Checks if elements can be copied, displaying a toast if not.
    - This was just a case of wrapping the existing checks with a new toast handler on the copy shortcut trigger.
2) Fixing and reworking the copy-paste handling as it relates to the data that moves across the clipboard.
    - A small bug fix was needed to ensure the correct metadata made it across.
    - Another bug fix prevents the side effect copy operation from not being invoked regardless of it it should be.
    - The data itself was refactored to help associate the original path with the element and add the imports from the original file.
3) Changing the paste handler to utilise the logic from the canvas strategies and how they handle reparenting.
    - The paste handler treats some of the process as being like a straight reparent and this resulted in a function handling an `ElementPath` or a `JSXElementChild`.
    - Since the canvas strategies involve certain types a bit more refactoring was done to functions so that the core of their functionality could be invoked from the paste operation.

**Notes:**
- Handling of the imports is quite primitive in this PR and should probably be improved subsequently, as it just globs across all the imports blindly which works but isn't great.

**Commit Details:**
- Fix the capitalisation of the `AddToReparentedPathsCommand` type.
- Export the `ReparentStrategy` type.
- Extracted the return type of `findReparentStrategy`.
- Split `findReparentStrategy` into two functions with the new one handling
  the core logic related to `ReparentStrategy` being called `reparentStrategyForParent`.
- Extracted the type for the `stripAbsoluteProperties` parameter from
  `applyFlexReparent`.
- Refactored `applyFlexReparent` to use the newly added
  `getFlexReparentPropertyChanges` function.
- Moved `getAbsoluteOffsetCommandsForSelectedElement` to `reparent-strategy-helpers.ts`
  and renamed it to `getAbsoluteReparentPropertyChanges`.
- Added `getFlexReparentPropertyChanges` which just extracts some existing
  logic.
- Implemented `getReparentPropertyChanges` to wrap `getAbsoluteReparentPropertyChanges`
  and `getFlexReparentPropertyChanges`.
- Created a new type `ToReparent`, which can either represent an `ElementPath`
  or `JSXElementChild` and some accompanying `Imports`.
- Updated `getReparentOutcome` to handle the updated parameter using the `ToReparent`
  type instead of just an `ElementPath`.
- Added `getReparentImports` as an early attempt at capturing the imports necessary
  for the reparent operation.
- Created `AddElement` canvas command to capture inserting a raw `JSXElementChild`
  into the hierarchy of the project.
- Slightly reworked the `PasteJSXElements` type with an array of `ElementPaste`
  and a target path to paste into.
- `editorMoveMultiSelectedTemplates` slightly refactored to handle
  getReparentOutcome` potentially returning a `null`.
- Added `toastOnUncopyableElementsSelected` and `toastOnUncopyableElementsTargeted`
  utility functions.
- Rewrote the `PASTE_JSX_ELEMENTS` action handler to use functions like
  `getReparentOutcome` and `getReparentPropertyChanges`.
- Shifted the side effecting `setClipboardData` inside the action parameter of
  `toastOnUncopyableElementsSelected` so that we don't run the side effect
  regardless of if it's allowed or not.
- Changed a browser test to be attributed to `pasteJSXElements` instead of
  `maybeSwitchLayoutProps` as that made more sense.
- Updated `getActionsForClipboardData` to handle the new `ElementPaste` entries.
- Fixed a bug with `filterMetadataForCopy` which meant that the targeted element
  was not included in the metadata.


enidemi (approved):

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -155,33 +203,16 @@ export function findReparentStrategy(
     strategyState,
   )
 
-  const newParentPath = reparentResult.newParent
```
balazsbajorics, line 158: I feel like a juicy conflict with #2457 is coming..

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1462,6 +1480,52 @@ function toastOnGeneratedElementsTargeted(
   return result
 }
 
+function toastOnUncopyableElementsSelected(
```
balazsbajorics, line 1483: what's the point of this function if it just calls `toastOnUncopyableElementsTargeted`?

editor/src/components/editor/store/editor-update.tsx
```diff
@@ -116,10 +116,8 @@ export function runSimpleLocalEditorAction(
     case 'CLOSE_POPUP':
       return UPDATE_FNS.CLOSE_POPUP(action, state)
     case 'PASTE_JSX_ELEMENTS':
-      return UPDATE_FNS.PASTE_JSX_ELEMENTS(action, state, dispatch)
+      return UPDATE_FNS.PASTE_JSX_ELEMENTS(action, state, dispatch, builtInDependencies)
     case 'COPY_SELECTION_TO_CLIPBOARD':
-      // side effect 😟
-      setClipboardData(createClipboardDataFromSelection(state))
```
balazsbajorics, line 122: is this moved somewhere else? it's not clear for me from the PR description

seanparsons: That's what this relates to:
"Shifted the side effecting `setClipboardData` inside the action parameter of `toastOnUncopyableElementsSelected` so that we don't run the side effect regardless of if it's allowed or not."

editor/src/components/editor/actions/actions.tsx
seanparsons: It was mirroring some existing code, as it seemed likely to be needed similarly where we can supply the paths being targeted separately.

seanparsons: Removed in 9783c0bca.

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
seanparsons: *clicks "Squash and merge"*

balazsbajorics (approved):

PR #2499, opened 2022-09-01
Feature/flow reorder strategies

enidemi:
**Problem:**
Static elements cannot be dragged on the canvas.

**Fix:**
Allow dragging static elements, it triggers reorder between sibling static elements. Based on https://github.com/concrete-utopia/utopia/pull/2494
Reordering from block to inline-block is possible with converting the element's display prop. If the display prop would be set to the default it is removed instead.

There are 3 strategies offered to the user on drag as the intention can be quite different:
1. automatically convert when inserting to rows
2. no conversion, keeps the original display prop
3. restrict reorder to only the display type the element has

Improvements for the future: insert into the end of row is not working currently, workaround is to use the same-type restricted strategy, and later support for style 'flow: right', dragging relative elements should allow setting the offset.

**Commit Details:**
- new strategy type
- strategy helper functions to find the new position and new display type
- new command to convert to block and inline-block
- 2 new canvas controls:
     - dragged outline differently colored for block and inline elements
     - for reorder restricted to rows(inline) or columns(blocks) shows target area with striped background on sibling elements
- new metadatautil: `isPositionedByFlow`
- 3 basic tests

Monosnap screencast 2022-09-01 16-35-34

Test it on this project: https://utopia.pizza/project/03ed1611/?branch_name=feature-flow-reorder-strategies


lankaukk:
Screen Shot 2022-09-01 at 12 43 00 PM
minor thing - but this looks off

seanparsons (approved):

editor/src/components/canvas/commands/convert-inline-block-command.ts
```diff
@@ -0,0 +1,49 @@
+import { emptyComments, jsxAttributeValue } from '../../../core/shared/element-template'
+import { ValueAtPath } from '../../../core/shared/jsx-attributes'
+import { ElementPath } from '../../../core/shared/project-file-types'
+import { EditorState } from '../../editor/store/editor-state'
+import { stylePropPathMappingFn } from '../../inspector/common/property-path-hooks'
+import { applyValuesAtPath } from './adjust-number-command'
+import { BaseCommand, CommandFunction, WhenToRun } from './commands'
+
+export interface ConvertInlineBlock extends BaseCommand {
```
balazsbajorics, line 9: my PR introduces a generic set property command – is there a merit to having this display prop specific one?

editor/src/core/shared/math-utils.ts
```diff
@@ -291,6 +291,19 @@ export function product<C extends CoordinateMarker>(a: Point<C>, b: Point<C>): n
   return a.x * b.x + a.y * b.y
 }
 
+export function pointIsClockwiseFromLine<C extends CoordinateMarker>(
+  targetPoint: Point<C>,
+  linePointA: Point<C>,
+  linePointB: Point<C>,
+): boolean {
+  // The order of line points a and b are important, as they determine which direction the line is pointing in,
```
balazsbajorics, line 299: if the line is pointing towards a direction, shouldn't we call it a vector instead?

editor/src/core/shared/dom-utils.ts
```diff
@@ -274,3 +275,13 @@ export function getDomRectCenter(rect: DOMRect): { x: number; y: number } {
     y: rect.y + rect.height / 2,
   }
 }
+
+export function defaultDisplayTypeForHTMLElement(elementName: string): 'inline' | 'block' | null {
```
balazsbajorics, line 279: hmmm, this only works if the user doesn't overwrite the defaults with a css. 
I _think_ in most/all places where we store browser defaults, we only use them to inform visual indicators, and not to make decisions based on them, but maybe I am totally wrong about this

enidemi: This prevents setting `display: block` on divs with the autoconversion dragging

balazsbajorics: could you leave a comment here please to indicate that this will need to be smarter if the user has global css overrides?

PR #2503, opened 2022-09-02
Escape cancels more of insert mode.

seanparsons:
**Problem:**
Because insertion is now powered by some newer logic, the older actions it used to fire when cancelling insert mode now no longer apply.

**Fix:**
Added an action to clear the interaction session and another to switch the right menu tab to the inspector. As well as also applying these actions if the insert menu is option at all, as well as if the current mode is insert mode.

**Commit Details:**
- Added `clearInteractionSession` and `setRightMenuTab` to the actions
  triggered by `esc` in insert mode.
- Wrapped up the actions in a function so it can be reused and to clean
  up the code a little.
- Also trigger the actions when the insert menu is open, not just when
  some kind of insert is in progress.

### DIFF
--- a/editor/src/components/editor/global-shortcuts.ts
+++ b/editor/src/components/editor/global-shortcuts.ts
@@ -110,10 +110,20 @@ import {
   TOGGLE_FOCUSED_OMNIBOX_TAB,
   FOCUS_CLASS_NAME_INPUT,
 } from './shortcut-definitions'
-import { DerivedState, EditorState, getOpenFile } from './store/editor-state'
+import { DerivedState, EditorState, getOpenFile, RightMenuTab } from './store/editor-state'
 import { CanvasMousePositionRaw, WindowMousePositionRaw } from '../../utils/global-positions'
 import { getDragStateStart } from '../canvas/canvas-utils'
 import { isFeatureEnabled } from '../../utils/feature-switches'
+
+function cancelInsertModeActions(): Array<EditorAction> {
+  return [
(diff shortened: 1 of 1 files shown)


editor/src/components/editor/global-shortcuts.ts
```diff
@@ -110,11 +110,21 @@ import {
...
 import { isFeatureEnabled } from '../../utils/feature-switches'
 
+function cancelInsertModeActions(): Array<EditorAction> {
+  return [
+    EditorActions.switchEditorMode(EditorModes.selectMode()),
+    CanvasActions.clearDragState(false),
+    EditorActions.clearHighlightedViews(),
+    CanvasActions.clearInteractionSession(false),
+    EditorActions.setRightMenuTab(RightMenuTab.Inspector),
+  ]
```
gbalint, line 125: It seems we could use this function here too: https://github.com/concrete-utopia/utopia/blob/master/editor/src/templates/editor-canvas.tsx#L188
And maybe here too: https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/controls/insert-mode-control-container.tsx#L432

seanparsons: There's a subsequent task on the list which involves another case for cancelling insert mode, I think it makes sense to do that refactoring there as my first attempt resulted in circular references.

balazsbajorics (approved):

gbalint (approved):

enidemi (approved):

PR #2509, opened 2022-09-05
Fix: flex reparent/insert target indicators

balazsbajorics:
**Problem:**
The flex reparent/insert target indicators are attached to the flex sibling element. This is because the indicators were actually just a placeholder implementation while I worked on the reparent, but then in lieu of more time, I merged them to master.

**Fix:**
The indicators now respect the flex gap property.

**Commit Details:**
- Added flex gap property to specialSizeMeasurements
- Using the measured flex gap to offset the indicators

**Note:** 
- The indicators still don't have the size of the dragged element
- if the computed style's flex gap is somehow different from the visual one, they will fail

Here's how it looks like:
https://screenshot.click/05-08-rugm8-5xq16.mp4

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
@@ -56,6 +56,7 @@ import {
   adjustCssLengthProperty,
 } from '../commands/adjust-css-length-command'
 import { updatePropIfExists } from '../commands/update-prop-if-exists-command'
+import { FlexReparentIndicatorSize } from '../controls/select-mode/flex-reparent-target-indicator'
 
 interface ReorderElement {
   distance: number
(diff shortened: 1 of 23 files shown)


editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -723,16 +738,16 @@ export function applyFlexReparent(
               const targetLineBeginningOfParent: CanvasRectangle =
                 newParentFlexDirection === 'row'
                   ? canvasRectangle({
-                      x: parentRect.x,
+                      x: parentRect.x, // TODO offset the insert line by half the flex gap?
```
seanparsons, line 741: Not sure if this TODO needs doing now.

editor/src/components/canvas/dom-walker.ts
```diff
@@ -893,6 +893,13 @@ function getSpecialMeasurements(
     height: globalFrame.height - border.top - border.bottom,
   })
 
+  // TODO is there a nicer way to get the number only if it's in pixels, and otherwise set it to null?
```
seanparsons, line 896: You could `flatMapEither` it and then `defaultEither` it, but it's arguable which is more comprehensible.

seanparsons (approved):

enidemi:
Based on flex-gap it's possible that the insertion indicator will be outside of the flex area.

Screenshot 2022-09-06 at 11 21 51

balazsbajorics:
closing it in favor of #2518

PR #2512, opened 2022-09-05
Show snapping guidelines during an absolute reparent

Rheeseyb:
**Problem:**
When reparenting an absolute positioned element, no snapping guidelines would be shown.

**Fix:**
The issue is that we are applying the move strategy first to update the frame, and then applying the re-parenting, but this means that the snapping was using outdated element paths (but was still being applied), and then the guidelines were not being rendered. To address this, the reparenting now provides a path map via the `customStrategyState`, which is used to determine which elements to use for snapping, as well as for the target of `intendedBounds` (which are used to determine whether or not to render the applied guidelines).

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
@@ -13,6 +13,7 @@ export interface CustomStrategyState {
   escapeHatchActivated: boolean
   lastReorderIdx: number | null
   duplicatedElementNewUids: { [elementPath: string]: string }
+  updatedTargetPaths: Array<ElementPath> | null
 }
 
 export function defaultCustomStrategyState(): CustomStrategyState {
@@ -20,6 +21,7 @@ export function defaultCustomStrategyState(): CustomStrategyState {
     escapeHatchActivated: false,
     lastReorderIdx: null,
     duplicatedElementNewUids: {},
+    updatedTargetPaths: null,
   }
 }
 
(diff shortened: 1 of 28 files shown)


editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
```diff
@@ -13,13 +13,15 @@ export interface CustomStrategyState {
   escapeHatchActivated: boolean
   lastReorderIdx: number | null
   duplicatedElementNewUids: { [elementPath: string]: string }
+  updatedTargetPaths: Map<ElementPath, ElementPath>
```
seanparsons, line 16: This will rely on the `ElementPath` instances being unique for a given path right?

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
balazsbajorics, line 16: instead of updatedTargetPaths being in the CustomStrategyState, I think they'd be better in the `InteractionSession`. Reason: the `CustomStrategyState` should be used to store state-like data, values that can be modified as a side effect during a long-running interaction. (example: whether some threshold was passed, such as the `escapeHatchActivated`) If a strategy flips a value, that should be preserved for the duration of the interaction.
Since what you want is to mask a value in a sort of inner context, I think that should go to some "prop like" part of the strategy API, and the closest we have to that is the `InteractionSession`. Last week I put `startingTargetParentToFilterOut` there in the reparent PR

balazsbajorics (approved):
see my comment about updatedTargetPaths should be moved to InteractionSession, otherwise looks good

editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
Rheeseyb: Hmmm yeah so we mostly ensure that, but actually I'll change this to use a stringified version for the key (since the `toString` function we created is fast anyway)

Rheeseyb: Yeah that's a fair comment. I'll move it to there 👍

PR #2517, opened 2022-09-06
No reorder for generated elements

gbalint:
Reordering is disallowed for elements which are root elements of generated content. 
This can be detected by checking if dynamic uid is the same as the static one. 
To be safe, I also disallowed reordering when any of the siblings are root elements of generated content (to make sure we don't try to reorder between two dynamically generated instances of the same static element)

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/reorder-utils.ts
+++ b/editor/src/components/canvas/canvas-strategies/reorder-utils.ts
@@ -1,0 +1,12 @@
+import * as EP from '../../../core/shared/element-path'
+import { ElementPath } from '../../../core/shared/project-file-types'
+
+export function isReorderAllowed(siblings: Array<ElementPath>) {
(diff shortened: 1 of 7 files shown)


seanparsons (approved):

editor/src/components/canvas/canvas-strategies/reorder-utils.ts
```diff
@@ -0,0 +1,12 @@
+import * as EP from '../../../core/shared/element-path'
+import { ElementPath } from '../../../core/shared/project-file-types'
+
+export function isReorderAllowed(siblings: Array<ElementPath>) {
+  return siblings.every((sibling) => !isRootOfGeneratedElement(sibling))
+}
+
+function isRootOfGeneratedElement(target: ElementPath): boolean {
```
seanparsons, line 8: Ah, clever!

seanparsons (approved):

PR #2518, opened 2022-09-06
flex insert/reparent indicator is drawn at the midpoint between targe…

balazsbajorics:
**Problem:**
The flex reparent/insert target indicators are attached to the flex sibling element. This is because the indicators were actually just a placeholder implementation while I worked on the reparent, but then in lieu of more time, I merged them to master.

**Fix:**
The indicators are now drawn in the geometric center between the two target siblings.

**Note:**
This is a variant to https://github.com/concrete-utopia/utopia/pull/2509 which uses the flex gap value

**This is how it looks like:**
https://screenshot.click/05-08-sklcu-x5x8g.mp4

**Details:**
- Calculating the indicator using the two target siblings instead of prepending it before the target succeeding sibling
- Adjusting the line width by canvas scale
- **Sneaked in fix** also using the canvas scale to alter the size of the flex insertion detection zones, so users have more controls over zoomed in canvases

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
@@ -57,6 +57,8 @@ import {
 } from '../commands/adjust-css-length-command'
 import { updatePropIfExists } from '../commands/update-prop-if-exists-command'
 
+const FlexReparentIndicatorSize = 2
+
 export type ReparentStrategy =
   | 'FLEX_REPARENT_TO_ABSOLUTE'
   | 'FLEX_REPARENT_TO_FLEX'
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -530,51 +555,78 @@ export function applyFlexReparent(
             setCursorCommand('mid-interaction', CSSCursor.Move),
           ]
 
-          const newParentFlexDirection = MetadataUtils.getFlexDirection(
-            MetadataUtils.findElementByElementPath(strategyState.startingMetadata, newParent),
-          )
+          const pseudoElementBefore: CanvasRectangle =
+            newParentFlexDirection === 'row' // TODO handle row-reverse or col-reverse
```
Rheeseyb, line 559: Can you add these TODOs to the paper doc please

Rheeseyb (approved):

enidemi (approved):

PR #2521, opened 2022-09-06
Mark edge intersections with xmarks

bkrmendy:
Mark edge intesections of snappes and snapper with x-marks-the-spot, like Figma
xmarks++


maltenuhn:
Hmm - just realised we normally inline the SVGs - but I don't actually remember why!

Do these behave and position correctly when the canvas is zoomed in / out?

editor/src/components/canvas/controls/guideline-controls.tsx
```diff
@@ -26,6 +41,35 @@ export const GuidelineControls = React.memo(() => {
...
+  const metadata = useEditorState((store) => store.editor.jsxMetadata, 'JSX metadata')
+
+  const { strategyIntendedBounds, snappingGuidelines } = useEditorState(
+    (store) => store.editor.canvas.controls,
+    'Strategy intended bounds and snapping guidelines',
+  )
+
+  const parentAndSiblings = collectParentsAndSiblings(
+    metadata,
+    strategyIntendedBounds.map((_) => _.target),
```
Rheeseyb, line 53: Please don't use underscores here. We typically use those to denote a param that isn't used.

editor/src/components/canvas/controls/guideline-controls.tsx
```diff
@@ -153,3 +201,129 @@ function useGuideline<T = HTMLDivElement>(
   }, [innerCallback])
   return controlRef
 }
+
+function framesFromMetadata(
+  metadata: ElementInstanceMetadataMap,
+  paths: Array<ElementPath>,
+): Array<CanvasRectangle> {
+  return mapDropNulls((path) => metadata[EP.toString(path)]?.globalFrame ?? null, paths)
```
Rheeseyb, line 209: Please use the functions in `element-metadata-utils.ts` for accessing anything in the metadata

editor/src/components/canvas/controls/guideline-controls.tsx
```diff
@@ -153,3 +201,129 @@ function useGuideline<T = HTMLDivElement>(
...
+    case 'YAxisGuideline': {
+      const a = canvasPoint({ x: guideline.xLeft, y: guideline.y })
+      const b = canvasPoint({ x: guideline.xRight, y: guideline.y })
+      return canvasSpan(a, b)
+    }
+    case 'CornerGuideline':
+      return null
+    default:
+      const _: never = guideline
+      throw new Error(`Unknown guideline found`)
```
Rheeseyb, line 237:
```suggestion
      throw new Error(`Unknown guideline of type ${guideline.type} found`)
```
Just so that we have a bit of extra info if this ever does happen

bkrmendy: what you suggested was my first idea too

editor/src/components/canvas/controls/guideline-controls.tsx
bkrmendy: fixed!

editor/src/components/canvas/controls/guideline-controls.tsx
Rheeseyb: Aha of course! Ok this is fine then.

editor/src/components/canvas/controls/guideline-controls.tsx
bkrmendy: thanks, fixed!

editor/src/components/canvas/controls/guideline-controls.tsx
```diff
@@ -153,3 +208,158 @@ function useGuideline<T = HTMLDivElement>(
...
+// https://stackoverflow.com/a/9997374
+function spansIntersect(a: CanvasSpan, b: CanvasSpan): boolean {
+  function ccw(p1: CanvasPoint, p2: CanvasPoint, p3: CanvasPoint): boolean {
+    return (p3.y - p1.y) * (p2.x - p1.x) <= (p2.y - p1.y) * (p3.x - p1.x)
+  }
+
+  return ccw(a.a, b.a, b.b) !== ccw(a.b, b.a, b.b) && ccw(a.a, a.b, b.a) !== ccw(a.a, a.b, b.b)
+}
+
+function spansIntersect2(a: CanvasSpan, b: CanvasSpan): boolean {
```
gbalint, line 299: Having a `spansIntersect` and a `spansIntersect2` and `spanIntersection` is confusing, what is the difference? Can we give them more meaningful names?

editor/src/components/canvas/controls/guideline-helpers.ts
```diff
@@ -17,9 +17,31 @@ import {
 import { ElementInstanceMetadataMap } from '../../../core/shared/element-template'
 import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import { EdgePosition } from '../canvas-types'
-import { pluck } from '../../../core/shared/array-utils'
 import { defaultIfNull } from '../../../core/shared/optional-utils'
 
+export function collectParentsAndSiblings(
```
gbalint, line 22: This looks like something which could be in MetadataUtils, it is independent from the guidelines

bkrmendy: moved, thanks!

editor/src/components/canvas/controls/guideline-controls.tsx
bkrmendy: while trying to explain the code in a comment, i actually found a better solution, so I pushed that instead (and the site that "inspired" the solution)

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.browser2.tsx
```diff
@@ -913,4 +914,34 @@ describe('Absolute Move Strategy Canvas Controls', () => {
...
+        `),
+      'await-first-dom-report',
+    )
+
+    const target = EP.appendNewElementPath(TestScenePath, ['aaa', 'bbb'])
+    const dragDelta = canvasPoint({ x: 29, y: -23 }) // 'bbb' will snap to bottom right corner of 'ccc'
+
+    await startDragUsingActions(renderResult, target, dragDelta)
+
+    expect(renderResult.renderedDOM.getByTestId('xmark-0').style.display).toBeTruthy()
```
seanparsons, line 941: I would recommend checking the actual value you want with `toEqual(..)` as `toBeTruthy() will match pretty much anything that isn't a `null` or `false`.

seanparsons (approved):

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1149,78 +1139,211 @@ export function collectGuidelines(
...
-                point.x + imageDimension.width,
-                point.y - lowHalfHeight,
-                point.y + highHalfHeight,
-              ),
-              yAxisGuideline(point.y - lowHalfHeight, point.x - imageDimension.width, point.x),
-              yAxisGuideline(point.y - lowHalfHeight, point.x, point.x + imageDimension.width),
-              yAxisGuideline(point.y + highHalfHeight, point.x - imageDimension.width, point.x),
-              yAxisGuideline(point.y + highHalfHeight, point.x, point.x + imageDimension.width),
+              {
+                guideline: xAxisGuideline(
```
bkrmendy, line 1212: I just realized that there's a way to write these with no duplication (basically a function that returns the guideline + the points based on the guildelin endpoints), I can add it on in this PR or in one immediately after this one (so the diff is more understandable)

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.browser2.tsx
```diff
@@ -202,6 +202,10 @@ export var ${BakedInStoryboardVariableName} = (props) => {
 `
 }
 
+function positioningFromCss({ left, top }: CSSStyleDeclaration) {
```
balazsbajorics, line 205: super minor nitpick, but I _think_ we generally avoid object destructuring in function parameters. I'm not sure if it's a rule, or we only avoid when in combination with type annotations (because that makes it look silly)

bkrmendy: I also think it looks a bit strange but at the same time it's somewhat useful in the case of these selector type functions

bkrmendy: refactoring

editor/src/components/canvas/canvas-strategies/shared-keyboard-strategy-helpers.ts
```diff
@@ -93,54 +93,58 @@ export function getKeyboardStrategyGuidelines(
...
+          const snappingVector = Guidelines.getOffsetToSnapToXGuideline(
+            horizontalPoints,
+            guideline,
+            null,
+          )
+          if (Utils.magnitude(snappingVector) === 0) {
+            return {
+              guideline: guideline,
+              snappingVector: snappingVector,
+              pointsOfRelevance: [],
```
Rheeseyb, line 117: From the PR alone I'm not clear on the previous intention of `activateSnap` here, and the result of replacing it with an empty `pointsOfReference` array 🤔 Is this changing the snapping behaviour here when using the keyboard?

editor/src/components/canvas/canvas-strategies/shared-keyboard-strategy-helpers.ts
```diff
@@ -93,54 +93,58 @@ export function getKeyboardStrategyGuidelines(
...
+            } as GuidelineWithSnappingVectorAndPointsOfRelevance
```
Rheeseyb, line 118: Please can you use a factory function or some other helper here rather than casting. We've found that it makes refactoring much easier.

editor/src/components/canvas/canvas-strategies/shared-keyboard-strategy-helpers.ts
bkrmendy: so I saw that `GuidelineWithSnappingVector` didn't include `activateSnap`, but this didn't cause a type error due to structural typing (since the object had everything `GuidelineWithSnappingVector` needed, plus `activateSnap`), so I removed `activateSnap` as cleanup

bkrmendy: keyboard snapping seems to be fine

Rheeseyb: Aha, so my request to use a factory function is spot on then 😀

editor/src/components/canvas/canvas-strategies/shared-keyboard-strategy-helpers.ts
bkrmendy: done!

editor/src/components/canvas/canvas-utils.ts
Rheeseyb: I'd say add it in a separate PR to keep this one as clear as possible

balazsbajorics: a follow-up PR sounds good!

editor/src/utils/utils.ts
```diff
@@ -626,6 +626,10 @@ function assert(errorMessage: string, predicate: boolean | (() => boolean)): voi
   throw new Error(`Assert failed: ${errorMessage}`)
 }
 
+function assertNever(n: never): never {
+  throw new Error(`Expected \`never\`, got ${JSON.stringify(n)}`)
+}
```
Rheeseyb, line 631: This is good, but we've stopped using this specific file as it became a massive grab bag for anything that didn't have a specific place. Instead we have moved to using more specific utils files in `core/shared`

editor/src/utils/utils.ts
```diff
@@ -920,9 +924,23 @@ export function isOptionsType<T extends OptionTypeBase>(
...
+  const results: Array<T> = []
+  for (const t of ts) {
+    const k = key(t)
+    if (!seen.has(k)) {
+      results.push(t)
+      seen.add(k)
+    }
+  }
+  return results
+}
```
Rheeseyb, line 938: Nice add, but we already have a function for exactly this, called `uniqBy` in `src/core/shared/array-utils.ts` 😁

Rheeseyb (approved):
Nice work!

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.browser2.tsx
gbalint: TBH I like them, because they express in the signature which part of the parameter is really used in the function body.

balazsbajorics (approved):

PR #2524, opened 2022-09-06
Navigator updates with reordering live.

seanparsons:
**Problem:**
Currently due to the way the tree is built for the navigator, it leans on the insertion order of the metadata, which results in incorrect ordering if elements are only reordered and the metadata is updated instead of getting reconstructed.

**Fix:**
For cases which can be reordered by the user, the tree structure gets reordered based on the element structure relating to those elements. This is done by taking the ordering of the children in the elements and then applying that ordering back to the path tree.

**Commit Details:**
- Implemented `reorderTree` which takes the result produced by `buildTree` and
  corrects the ordering based on elements found within elements.
- Added a call to `reorderTree` to `createOrderedElementPathsFromElements`.

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -83,7 +83,7 @@ import {
 } from '../../components/editor/store/editor-state'
 import { ProjectContentTreeRoot } from '../../components/assets'
 import { memoize } from '../shared/memoize'
-import { buildTree, ElementPathTree, getSubTree } from '../shared/element-path-tree'
+import { buildTree, ElementPathTree, getSubTree, reorderTree } from '../shared/element-path-tree'
 import { findUnderlyingTargetComponentImplementation } from '../../components/custom-code/code-file'
 
 const ObjectPathImmutable: any = OPI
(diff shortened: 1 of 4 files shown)


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -807,7 +807,11 @@ export const MetadataUtils = {
     } => {
       // Note: This will not necessarily be representative of the structured ordering in
       // the code that produced these elements.
```
Rheeseyb, line 809: Is this note still valid?

seanparsons: Yes, because `reorderTree` is quite conservative, so things like blocks of generated elements will potentially still be out of order.

editor/src/core/shared/element-path-tree.ts
```diff
@@ -54,6 +59,56 @@ export function buildTree(elementPaths: Array<ElementPath>): ElementPathTreeRoot
...
+  } else {
+    return foldEither(
+      () => {
+        return tree
+      },
+      (elementChild) => {
+        switch (elementChild.type) {
+          case 'JSX_ELEMENT': {
+            const allChildrenAreElements = elementChild.children.every(isJSXElement)
+            if (allChildrenAreElements) {
```
Rheeseyb, line 78: It seems we don't need to restrict this to cases where all children are JSX Elements? Couldn't we just skip over children that aren't?

Rheeseyb (approved):

editor/src/core/shared/element-path-tree.ts
seanparsons: If you have an element with children of element, arbitrary block, element then it can end up doing odd things with reordering the last element into the middle of the generated elements produced by the arbitrary block.

PR #2527, opened 2022-09-06
Auto-scroll to selected element in the navigator

Rheeseyb:
**Problem:**
The navigator doesn't auto-scroll to the selected element.

**Fix:**
Use the [`scrollToItem`](https://react-window.vercel.app/#/examples/list/scroll-to-item) API of react-window to scroll to the first selected element on rendering via a `useEffect`.

Note that unfortunately it seems like this component library is no longer maintained, and has some issues with the auto-scrolling. Attempting to use the `"center"` option for scrolling the list to center the element doesn't work, so I've gone with the default behaviour. Scrolling to the very first element is also broken.


maltenuhn:
Is this triggering for every interaction (code editor, canvas selection)? Or only specific ones? Any other gotchas / expected behaviours?

bkrmendy (approved):

enidemi (approved):

Rheeseyb:
> Is this triggering for every interaction (code editor, canvas selection)? Or only specific ones? Any other gotchas / expected behaviours?

This will trigger whenever the selection index changes

PR #2529, opened 2022-09-07
fix(canvas) delay cursor and strategy picker too

enidemi:
**Problem:**
Strategy picker and canvas controls are delayed when starting a drag, but canvas cursors still show up immediately.

**Fix:**
Delay canvas cursors too when a strategy starts. If the user moves the mouse it appears immediately.

**Commit Details:**
- add new hook to use the same delay as the strategy picker

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
@@ -236,7 +236,7 @@ export function applyCanvasStrategy(
 }
 
 export function useDelayedStrategyWithCallback<T>(
-  innerCallback: StateSelector<EditorStorePatched, T | null>,
+  selector: StateSelector<EditorStorePatched, T | null>,
 ): T | null {
   /**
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -233,58 +235,69 @@ export function applyCanvasStrategy(
   return strategy.apply(canvasState, interactionSession, strategyState)
 }
 
-export const useDelayedCurrentStrategy = () => {
+export function useDelayedStrategyWithCallback<T>(
+  innerCallback: StateSelector<EditorStorePatched, T | null>,
```
Rheeseyb, line 239: Please can you rename this, as it is a selector and not a callback (so this could lead to some amount of confusion)

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
enidemi: fixed

seanparsons (approved):

PR #2530, opened 2022-09-07
holding cmd allows absolute reparent to static parents

balazsbajorics:


balazsbajorics:
Malte and I played with it a bit, and concluded that I'll shelf the spike for now, because it only solves a specific edge case, but I just realized for example it would stop working correctly if we introduce Flow/Static reparent into the unified reparent lookup function.

Instead of this spike, we should probably have a modifier or other way to turn an element into absolute, regardless of the parent's context

PR #2532, opened 2022-09-07
Use the closest coordinate system bounds when re-parenting

Rheeseyb:
**Problem:**
Re-parenting from a deep hierarchy to the canvas can offset the dragged element if the original parent doesn't provide the coordinate system bounds for the dragged element

**Fix:**
Find the closest coordinate system bounds (`globalContentBox` or `coordinateSystemBounds`) in the hierarchy above the old and new parents, and use those when calculating the offset to apply.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
@@ -719,11 +719,11 @@ export function getAbsoluteReparentPropertyChanges(
     return []
   }
 
-  const currentParentContentBox = MetadataUtils.getClosestParentCoordinateSystemBounds(
+  const currentParentContentBox = MetadataUtils.getParentCoordinateSystemBounds(
     EP.parentPath(target),
     targetStartingMetadata,
   )
-  const newParentContentBox = MetadataUtils.getClosestParentCoordinateSystemBounds(
+  const newParentContentBox = MetadataUtils.getParentCoordinateSystemBounds(
     newParent,
     newParentStartingMetadata,
   )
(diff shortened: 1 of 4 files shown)


enidemi (approved):

seanparsons (approved):

balazsbajorics (commented):
Do we know if we can trust `closestOffsetParentPath`? What if the _real_ closest offset parent is in the middle of a component's internal hierarchy?

for the spike #2530 I did this:
```typescript

  const elementOrContainingParent = providesBoundsForAbsoluteChildren
    ? element
    : (element.offsetParent as HTMLElement)

  const globalFrame = globalFrameForElement(elementOrContainingParent, scale, containerRectLazy)
  const globalContentBox = canvasRectangle({
    x: globalFrame.x + border.left,
    y: globalFrame.y + border.top,
    width: globalFrame.width - border.left - border.right,
    height: globalFrame.height - border.top - border.bottom,
  })
```

which uses the `element.offsetParent` and then dom measures that, so even if that is an element without a valid elementPath, we can still measure it

Rheeseyb:
> Do we know if we can trust `closestOffsetParentPath`? What if the _real_ closest offset parent is in the middle of a component's internal hierarchy?
> 
> for the spike #2530 I did this:
> 
> ```ts
>   const elementOrContainingParent = providesBoundsForAbsoluteChildren
>     ? element
>     : (element.offsetParent as HTMLElement)
> 
>   const globalFrame = globalFrameForElement(elementOrContainingParent, scale, containerRectLazy)
>   const globalContentBox = canvasRectangle({
>     x: globalFrame.x + border.left,
>     y: globalFrame.y + border.top,
>     width: globalFrame.width - border.left - border.right,
>     height: globalFrame.height - border.top - border.bottom,
>   })
> ```
> 
> which uses the `element.offsetParent` and then dom measures that, so even if that is an element without a valid elementPath, we can still measure it

Good shout, done :+1:

gbalint (approved):

balazsbajorics (approved):

PR #2534, opened 2022-09-07
Navigator shows flex reparent live

balazsbajorics:
**Problem:**
The navigator doesn't update live during a flex reparent interaction.

The reason is that we don't actually reparent, only draw an insertion target indicator.

**Solution:**
Actually _do_ run the flex reparent, but then make the element displayNone.

**Note:**
The navigator correctly picks up the reparented element, but draws it with a yellow warning triangle because it's invisible. This could be improved by using some sort of whitelist of elements that shouldn't have the warning triangle.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
@@ -527,7 +527,7 @@ export function applyFlexReparent(
           canvasState.openFile,
           pathToReparent(target),
           newParent,
-          'on-complete',
+          'always',
         )
 
         if (outcomeResult != null) {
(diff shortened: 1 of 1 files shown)


gbalint (approved):

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -802,8 +804,8 @@ export function getFlexReparentPropertyChanges(
 ) {
   return stripAbsoluteProperties
     ? [
-        deleteProperties('on-complete', newPath, propertiesToRemove),
-        updatePropIfExists('on-complete', newPath, PP.create(['style', 'position']), 'relative'), // SPIKE TODO only insert position: relative if there was a position nonstatic prop before
+        deleteProperties('always', newPath, propertiesToRemove),
+        updatePropIfExists('always', newPath, PP.create(['style', 'position']), 'relative'), // SPIKE TODO only insert position: relative if there was a position nonstatic prop before
```
Rheeseyb, line 808: I appreciate this is from previous work, but `SPIKE TODO`?

Rheeseyb (approved):

seanparsons (approved):

bkrmendy (approved):

PR #2535, opened 2022-09-07
Refactor/custom strategy state patch

gbalint:
Strategies can store state in `StrategyState.customStrategyState`
To make this possible `StrategyApplicationResult` contains a `customState` field, which can be `null`, to express that the strategy doesn't want to change the custom state. That is an ugly pattern, its goal was to not force strategies to always return the previous custom state.

This PR introduces a new way to solve this problem: strategies can just return a patch to the state, so they only have to return with the fields what they want to change.

At the same time I removed the `success` field from `CustomStrategyState`: it was not a good place for that, because `success` is not something what we ever want to keep: this should be always the result of the last application of the current strategy.

I added a constructor function for `StrategyApplicationResult` to make it easier to create results in the apply functions and in the helpers. Mostly we just return an array of commands, so the other parameters are optional.

We had this `emptyStrategyApplicationResult` which were always used in error cases, so I decided to just call that `failedStrategyApplicationResult` and set the `status` in it to `failure`

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.tsx
@@ -17,7 +17,7 @@ import { determineConstrainedDragAxis } from '../controls/select-mode/move-utils
 import { ConstrainedDragAxis, GuidelineWithSnappingVector } from '../guideline'
 import {
   CanvasStrategy,
-  failedStrategyApplicationResult,
+  emptyStrategyApplicationResult,
   getTargetPathsFromInteractionTarget,
   InteractionCanvasState,
   strategyApplicationResult,
@@ -111,7 +111,7 @@ export const absoluteMoveStrategy: CanvasStrategy = {
       )
     }
     // Fallback for when the checks above are not satisfied.
-    return failedStrategyApplicationResult
+    return emptyStrategyApplicationResult
   },
 }
 
@@ -174,6 +174,6 @@ export function applyAbsoluteMoveCommon(
     }
   } else {
     // Fallback for when the checks above are not satisfied.
-    return failedStrategyApplicationResult
+    return emptyStrategyApplicationResult
   }
 }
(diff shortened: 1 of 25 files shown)


editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
```diff
@@ -6,33 +6,47 @@ import { ElementPath, NodeModules } from '../../../core/shared/project-file-type
...
 
-export const emptyStrategyApplicationResult = {
+export const failedStrategyApplicationResult: StrategyApplicationResult = {
   commands: [],
-  customState: null,
+  customStatePatch: {},
+  status: 'failure',
+}
+
+export function strategyApplicationResult(
```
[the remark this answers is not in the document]
gbalint: The main reason for me to introduce a strategy function is that in 95% of the cases you only want return an array of commands, and do not want change the custom state, or set the status to failure. So we would need a lot of 
```
  customStatePatch: {},
  status: 'success',
}
```
boilerplate lines. But with the factory function i could make these parameters optional. However, I can just remove it and use object literals everywhere (although we use constructor functions everywhere, so this is pretty standard in the code)

editor/src/components/canvas/canvas-strategies/absolute-move-strategy.tsx
```diff
@@ -110,7 +111,7 @@ export const absoluteMoveStrategy: CanvasStrategy = {
       )
     }
     // Fallback for when the checks above are not satisfied.
-    return emptyStrategyApplicationResult
+    return failedStrategyApplicationResult
```
Rheeseyb, line 114: Hmmm this raises an interesting semantical point here, because I would interpret a "failed" result to imply that we tried to apply it but something went wrong. That makes sense in the re-parenting strategies when the target doesn't accept children. But in this case we haven't tried to apply the strategy as such, because there has been no drag yet. If we mark this as failed (and grey out the text in the picker) I think that risks giving the user the impression that they can't apply this strategy. Maybe it would be better for this to return an empty successful result?

balazsbajorics: oooooh that's a very nice catch! would this show the failed state for all strategies before they reach the drag threshold? that's definitely a mistake!

gbalint: fixed in b62e7b05842dadd58dca881eba7c34d100649e48

balazsbajorics (approved):

PR #2538, opened 2022-09-08
new cursors for indicating what you CANT do

lankaukk:
The previous "cursor-no-reparent" was in black, and not _strikingly_ clear enough. Let's try it in red! I've also changed it to use this symbol: 🚫​ instead. 

I've also added a cursor with a red ​❌​ , named "cursor-no" in case we want to try that or use it for anything else.


bkrmendy (approved):

Rheeseyb (approved):

Rheeseyb:
Looks good!

PR #2539, opened 2022-09-09
Fix/canvas flow reorder fix

enidemi:
The 2 issues with the flow reorder (when dragging the element, not the slider):
1. it's based on the closest sibling and since then we changed the flex reorder to use the sibling under cursor, the flow reorder should use the same
2. it's using the starting metadata, but compared to a flex reorder here the layout can change so much with mix of blocks and inline-blocks that the target areas are incorrect without the latest metadata.

Using the element under cursor would cause a flip-flop `previousReorderTargetSiblingUnderMouse` is there to stop it.

This PR also removes extra flow reorder strategies: the goal is that there will be 2 reorder strategies, 1 flex-like reorder and 1 fallback (slider).

There is a TODO about filtering the siblings in `isValidSibling`, I plan to to do it in a different PR

https://utopia.pizza/p/24a97552-closed-rainforest/?branch_name=fix-canvas-flow-reorder-fix


balazsbajorics (approved):
shouldn't the 1-d flow and 1-d flex reorders actually be the exact same strategy? flex layouts can also have wildly differently sized elements.

enidemi:
> shouldn't the 1-d flow and 1-d flex reorders actually be the exact same strategy? flex layouts can also have wildly differently sized elements.

or at least use the same helpers

balazsbajorics (approved):

seanparsons (approved):
Less code and works better, what's not to like? :)

PR #2540, opened 2022-09-09
Use the same position: absolute setting in all types of insertion

gbalint:
Adding `position: absolute` to the insertion subject when it is initialized in the insert menu and added to the editor mode in editor state.
With this change, we don't need to set it when dragging to insert, or when we create the jsx elements from html elements.

Note: when we drag into a flex component, this will be overridden. This is just a better default than nothing, but it can be changed during the actual insertion.

### DIFF
--- a/editor/src/components/editor/actions/actions.spec.tsx
+++ b/editor/src/components/editor/actions/actions.spec.tsx
@@ -1450,7 +1450,7 @@ describe('INSERT_INSERTABLE', () => {
           export var Card = (props) => {
             return (
               <div style={{ ...props.style }}>
-                <div style={{ position: 'absolute' }}>
+                <div>
                   <div
                     style={{
                       position: 'absolute',
--- a/editor/src/components/shared/__snapshots__/project-components.spec.ts.snap
+++ b/editor/src/components/shared/__snapshots__/project-components.spec.ts.snap
@@ -52,26 +52,7 @@ Array [
               "propertyElements": Array [],
             },
           },
-          "props": Array [
-            Object {
-              "comments": Object {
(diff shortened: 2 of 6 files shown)


editor/src/components/editor/actions/actions.spec.tsx
```diff
@@ -1450,7 +1450,7 @@ describe('INSERT_INSERTABLE', () => {
           export var Card = (props) => {
             return (
               <div style={{ ...props.style }}>
-                <div style={{ position: 'absolute' }}>
+                <div>
```
Rheeseyb, line 1453: Is this change expected?

Rheeseyb (approved):

seanparsons (approved):

editor/src/components/shared/__snapshots__/project-components.spec.ts.snap
```diff
@@ -52,26 +52,7 @@ Array [
               "propertyElements": Array [],
             },
           },
-          "props": Array [
```
balazsbajorics, line 55: are these changes expected?

gbalint: Yes, this is just a revert of last week's change which made even inserting with the A button and the Add Element popup using absolute positioning: https://github.com/concrete-utopia/utopia/commit/c4c81aeb1db424b42ccd82a8a7f1727cfe319769 (which is a followup to https://github.com/concrete-utopia/utopia/commit/6d99bce4830e6484937ac1c808693ee5ccc1bf22 )

So basically this PR is also a fix of a recent regression.

PR #2544, opened 2022-09-09
Non-resizable controls on elements that can't be resized.

seanparsons:
**Problem:**
Elements that are not resizable just lack the controls for resizing, but don't indicate specifically that they don't allow resizing.

**Fix:**
When the applicable strategies lack a case which supports resizing, a set of non-resizing corner indicator controls are added in.

**Commit Details:**
- Split out some parts of `useGetApplicableStrategyControls` into smaller functions.
- Added `isResizableStrategy` to determine if a strategy means something is resizable.
- `useGetApplicableStrategyControls` now adds in `notResizableControls` if no resizable strategy is currently applicable.
- Added `NonResizableControl` and NonResizablePoint`.
- Corrected some minor typos on flow strategy variable names.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
@@ -19,6 +19,8 @@ import { selectComponents } from '../../editor/actions/action-creators'
 import CanvasActions from '../canvas-actions'
 import { AllElementProps } from '../../editor/store/editor-state'
 import { CanvasControlsContainerID } from '../controls/new-canvas-controls'
+import { PrettierConfig } from 'utopia-vscode-common'
+import * as Prettier from 'prettier/standalone'
 
 const baseStrategyState = (
   metadata: ElementInstanceMetadataMap,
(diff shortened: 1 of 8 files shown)


editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -300,20 +301,59 @@ export const useDelayedStrategyCursor = () => {
...
+        strategyControls,
+        (l, r) => l.control === r.control,
       )
-      return addAllUniquelyBy(working, filteredControls, (l, r) => l.control === r.control)
-    }, [])
+    }
+    // Special case controls.
+    if (!isResizable) {
+      applicableControls.push(notResizableControls)
+    }
```
Rheeseyb, line 355: Hmmm this means that these controls show up during all dragging interactions, which also means they render whilst you mouse down on a selected resizeable element. I think the approach here needs to be that we explicitly choose when to include these controls, rather than implicitly adding them

seanparsons: Hmmm, showing up during drags is definitely unexpected. It's a pain because of it being the inverse of an actual condition, where other things are better at determining that case.

seanparsons: This should now be fixed in 7ea4e4003.

editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
```diff
@@ -697,3 +699,222 @@ describe('Snapping guidelines for absolutely moved element', () => {
...
+    ></div>
+  </div>
+  <div style='position: absolute; pointer-events: none; left: 100px; top: 30px;'>
+    <div
+      data-testid='non-resizable-1-1'
+      style='position: relative; width: 6px; height: 6px; top: -3px; left: -3px; background: rgb(255, 255, 255); border: 1px solid rgba(0, 0, 0, 0.5); border-radius: 50%;'
+    ></div>
+  </div>
+</div>
+`)
```
balazsbajorics, line 900: could you replace this with one or more more specific assertions? this will break every time we update the design of these indicators, and it's not a snapshot so it'll be a pain to update.

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -300,21 +301,79 @@ export const useDelayedStrategyCursor = () => {
   return useDelayedStrategy<CSSCursor | null>(selector)
 }
 
+const notResizableControls: ControlWithKey = {
```
balazsbajorics, line 304: I wonder if this could be made more "ceremonial" by turning into the official array of "NonApplicableControls" and become an array with one element

editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
seanparsons: Sure.

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
seanparsons: It was an array at one point, but I suspect because any other cases will also be similarly specific that they'll need to be separate instances.

Rheeseyb (approved):

PR #2547, opened 2022-09-12
Chore/strategy metadata test

balazsbajorics:
**Problem:**
1. I forgot that `InteractionSession.metadata` was the latest metadata, and not some starting value. 
2. When looking into it, I realized that `InteractionSession.allElementProps` is accidentally only the starting value. The code that maintains this is a bit convoluted and not tested.

**Fix:**
1. Renamed InteractionSession.metadata to InteractionSession.latestMetadata, and InteractionSession.allElementProps to InteractionSession.latestAllElementProps
2. Made a new test case that checks that startingMetadata and latestMetadata diverge during an interaction session


editor/src/components/editor/store/dispatch.tsx
```diff
@@ -674,7 +676,8 @@ function editorDispatchInner(
               ...result.unpatchedEditor.canvas,
               interactionSession: {
                 ...result.unpatchedEditor.canvas.interactionSession,
-                metadata: reconstructJSXMetadata(result.unpatchedEditor),
+                latestMetadata: reconstructJSXMetadata(result.unpatchedEditor),
+                latestAllElementProps: result.unpatchedEditor.allElementProps,
```
balazsbajorics, line 680: heads up this is a change in behavior. on master, interactionSession.allElementProps are never updated, so it's effectively a startingAllElementProps

seanparsons (approved):

bkrmendy (approved):

PR #2548, opened 2022-09-12
Prevent reparenting when target only contains text

Rheeseyb:
**Problem:**
Elements that only contain text as a child (e.g. `<span>Hello</span>`) are rarely the intended targets of a reparent interaction

**Fix:**
Prevent those elements as reparent targets by adding a check inside `MetadataUtils.targetElementSupportsChildren`. I've also added a number of tests to capture this.

This broke a number of unrelated tests, which exposed a couple of issues in the navigator's restriction of drop targets, and the cursor rendering when a target is invalid. I fixed those as part of this PR because they were blocking this. If absolutely necessary those fixes can be separated from this PR.

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
+++ b/editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
@@ -5,8 +5,7 @@ import { useEditorState } from '../../../editor/store/store-hook'
 import { cursorForMissingReparentedItems } from '../../canvas-strategies/reparent-utils'
 import { CSSCursor } from '../../canvas-types'
 import { getCursorFromDragState } from '../../canvas-utils'
-import Utils from '../../../../utils/utils'
-import { useDelayedStrategyCursor } from '../../canvas-strategies/canvas-strategies'
+import { useDelayedEditorState } from '../../canvas-strategies/canvas-strategies'
 
 export function getCursorForOverlay(editorState: EditorState): CSSCursor | null {
   const forMissingReparentedItems = cursorForMissingReparentedItems(
@@ -19,10 +18,9 @@ export function getCursorForOverlay(editorState: EditorState): CSSCursor | null 
 }
 
 export const CursorOverlay = React.memo(() => {
-  const strategyCursor = useDelayedStrategyCursor()
-  const cursor = useEditorState((store) => {
-    return Utils.defaultIfNull(strategyCursor, getCursorFromDragState(store.editor))
-  }, 'CursorOverlay cursor')
+  const cursor = useDelayedEditorState((store) => {
+    return getCursorForOverlay(store.editor)
+  })
 
   const styleProps = React.useMemo(() => {
     let workingStyleProps: React.CSSProperties = {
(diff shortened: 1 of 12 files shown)


editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
```diff
@@ -19,10 +18,9 @@ export function getCursorForOverlay(editorState: EditorState): CSSCursor | null
 }
 
 export const CursorOverlay = React.memo(() => {
-  const strategyCursor = useDelayedStrategyCursor()
-  const cursor = useEditorState((store) => {
-    return Utils.defaultIfNull(strategyCursor, getCursorFromDragState(store.editor))
-  }, 'CursorOverlay cursor')
+  const cursor = useDelayedEditorState((store) => {
+    return getCursorForOverlay(store.editor)
```
Rheeseyb, line 22: This previously meant that the canvas overlay wouldn't show the "not permitted" cursor for disappearing elements (though the test for this was returning that cursor!)

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -338,16 +338,19 @@ export const NavigatorItemContainer = React.memo((props: NavigatorItemDragAndDro
...
-          item.elementPath,
-        )
+        const isReparentTarget = item.appropriateDropTargetHint?.type === 'reparent'
+        const childrenSupportedIfRequired =
+          !isReparentTarget ||
+          MetadataUtils.targetSupportsChildren(
+            editorState.projectContents,
+            editorState.canvas.openFile?.filename,
+            editorState.jsxMetadata,
+            props.elementPath,
```
Rheeseyb, line 348: `item.elementPath` is the dragged item here, so previously we were incorrectly preventing the reparenting of an element that doesn't accept children (rather than testing if the drop target accepts children)!

seanparsons (approved):

editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
balazsbajorics: dod this change the other delayed curso behaviors though?

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
balazsbajorics: oh wow, nice catch!

editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
Rheeseyb: Nah I checked this. It was previously using the exact same logic, but without updating the cursor for missing dragged items

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
Rheeseyb: This was the result of a failing test case that confused the hell out of me!

PR #2549, opened 2022-09-13
Remove unused BoundingArea.target

gbalint:


editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -1656,8 +1656,9 @@ export const EdgePositionKeepDeepEquality: KeepDeepEqualityCall<EdgePosition> =
     edgePosition,
   )
```
balazsbajorics, line 1658: there's some other places in this file that also include a line calling the factory function, to make sure the build brakes if we later change boundingArea to _have_ parameters in the future

```suggestion
boundingArea() // this is here to break if the definition of boundingArea changes
```

for example: https://github.com/concrete-utopia/utopia/blob/e85cc7349ff40a8b83e29b40328d75f5369a9cbf/editor/src/components/editor/store/store-deep-equality-instances.ts#L2203-L2207

balazsbajorics (approved):

enidemi (approved):

Rheeseyb (approved):

editor/src/components/editor/store/store-deep-equality-instances.ts
Rheeseyb: Just you wait until a parameter is added, and the factory function is blindly updated without taking note of the comment 😁

PR #2550, opened 2022-09-13
Use reparent position logic for the flex insertion indicators

seanparsons:
**Problem:**
The logic for the flex insertion indicators works differently to the reparent/reorder logic and we really want those to be consistent.

**Fix:**
The calculations utilised in the reparent/reorder logic were extracted out as far as reasonably possible into two functions `getSiblingMidPointPosition` and `siblingAndPseudoPositions`, which have then been used in the flex insertion indicators.

Alongside this some tweaks needed to be made to the indicators as their new positions resulted in them either being inconsistent with each other or were offset from where they needed to be visually.

**Limitations:**
Some of the existing logic does not currently handle `row-reverse` and `column-reverse`, which has been carried forward into this. This should be addressed in a later PR.

**Screenshots:**
New positions:
image

**Commit Details:**
- Pulled out `getSiblingMidPointPosition` refactoring the position logic out
  of `applyFlexReparent`.
- Changed `ButtonControlProps.key` to be `ButtonControlProps.identifier, as the old
  name conflicted a bit with the React `key` and could end up getting removed
  from the props.
- `InsertionControls` uses `getBetweenChildrenPosition`, which in turn calls
  `getSiblingMidPointPosition` for those cases where it is calculating the position
  between flex elements, as opposed to the opposite axis.
- Changed `BlueDot` to match `PlusButton` as closely as it can so that it gets rendered
  in the same position as the latter.
- Corrected the positioning slightly of the `Line` component as it was rendering one
  pixel out from where it shoould be.
- Added more `data-testid` values for easier debugging.
- Extracted out `siblingAndPseudoPositions` so that
  the bounds can be easily calculated for the positions
  either side of the child elements.
- Slight refactor to extract `midInteractionCommandsForTarget`
  so that the near identical code in two places can be collapsed.
- `InsertionControls` now uses `siblingAndPseudoPositions`
  to construct the frames, looping over those instead of the children.

### DIFF
--- a/editor/src/core/performance/performance-regression-tests.spec.tsx
+++ b/editor/src/core/performance/performance-regression-tests.spec.tsx
@@ -64,7 +64,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`402`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`400`)
   })
 
   it('Clicking on opacity slider with a less simple project', async () => {
@@ -124,7 +124,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`391`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`389`)
   })
 
   it('Changing the selected view with a simple project', async () => {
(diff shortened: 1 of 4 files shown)


editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -64,7 +64,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`402`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`400`)
```
Rheeseyb, line 67: Any idea why each of these dropped by 2? They seem very unrelated...

Rheeseyb (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
seanparsons: Not particularly, I thought maybe it was returning nulls in something and now only returns 1, but glancing over it I can't particularly see anything like that.

PR #2551, opened 2022-09-13
Feature/canvas flow strategy slider

enidemi:
Canvas Flow Slider Strategy and Control

Try it here: https://utopia.pizza/p/4ae3c4c5-giant-beetle/?branch_name=experiment-canvas-flow-strategy-slider

**The Control:**
- it is placed in the middle of the flow elements frame
- interacting with the control shows the slider where you can drag left and right to reorder the selected element
- mousedown enables pointer lock mode in the browser https://developer.mozilla.org/en-US/docs/Web/API/Document/pointerLockElement
  - this has an advantage: compared to dragging which stops when reaching the side of the window this can be infinite that restarts the reorder from the other side when reaching the end
  - with this enabled clientX/Y are not changing during the mousemove events, interaction session is updated with movementX/Y, those are too small changes for reaching the drag threshold so this interaction starts even with 1px movement
  - unfortunately this has no tests as it requires a real user input
- there is a blue indicator moving with your mouse, the goal is that it's giving you feedback how close you are to trigger a reorder

Improvements in the next PRs
- enabling dragging vertically if it's nice to use
- fix wrap around for the indicator when reaching the edges

**The Strategy:**
- for now the fitness is quite high if it's applicable
- similarly how the flow reorder strategy works this strategy also automatically converts between inline block and block


editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -1705,6 +1705,11 @@ export const CanvasControlTypeKeepDeepEquality: KeepDeepEqualityCall<CanvasContr
         return KeyboardCatcherControlKeepDeepEquality(oldValue, newValue)
       }
       break
+    case 'FLOW_SLIDER':
+      if (newValue.type === oldValue.type) {
+        return keepDeepEqualityResult(oldValue, true)
```
balazsbajorics, line 1710: to make sure that we update the code if the flowSlider() factor function ever gets a parameter, in this file we usually add a line calling the factory function in the root with a comment saying if this breaks, update the code. look at lines 1675-1676

editor/src/core/shared/math-utils.ts
```diff
@@ -1020,3 +1020,7 @@ export function segmentIntersection(
   }
   return null
 }
+
+export function mod(n: number, m: number) {
+  return ((n % m) + m) % m
```
balazsbajorics, line 1025: what is this? why isn't % enough? is this some JavaScript madness here?

editor/src/templates/editor-canvas.tsx
```diff
@@ -1221,23 +1224,47 @@ export class EditorCanvas extends React.Component<EditorCanvasProps> {
...
-            cursor: null,
-            nativeEvent: event,
-            interactionSession: updateInteractionViaMouse(
-              this.props.editor.canvas.interactionSession,
-              newDrag,
-              Modifier.modifiersForEvent(event),
-              null,
-            ),
-          })
+          if (document.pointerLockElement) {
```
balazsbajorics, line 1227: what is document.pointerLockElement? is it a boolean?

enidemi: oops, fixed

editor/src/core/shared/math-utils.ts
enidemi: unfortunately the % is the remainder operator in JS which works differently from modulo for negative values. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder#description

editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -185,10 +185,13 @@ export function updateInteractionViaMouse(
   drag: CanvasVector,
   modifiers: Modifiers,
   sourceOfUpdate: CanvasControlType | null, // If null it means the active control is the source
+  ignoreThreshold: 'ignore-threshold' | 'use-threshold' = 'use-threshold',
```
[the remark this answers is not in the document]
enidemi: I'm still fixing this, the clumsy hands is not solved

Rheeseyb (approved):

balazsbajorics (approved):

PR #2553, opened 2022-09-13
Always enable absolute reparenting

Rheeseyb:
**Problem:**
Sometimes you just want to drag the thing, but also reparent it!

**Fix:**
Remove the requirement for `specialSizeMeasurements.providesBoundsForAbsoluteChildren` when checking if a reparent can happen, and give both the `absoluteReparentStrategy` and `flexReparentToAbsoluteStrategy` fallback weights of 2 if the best fitting reparent strategy isn't `'do-not-reparent'` (meaning we will still give the best preferred strategy the higher weighting).

As part of this, I have also fixed an issue with absolute reparenting - whereas all forms of flex reparenting wrap the apply function in a call to `ifAllowedToReparent` (which will trigger the "not allowed" cursor for generated elements), that was missing from the absolute reparenting strategy, so I have added it there.

Sample project: https://utopia.pizza/p/b53fe139-fluttering-settee-(forked)-(forked)/?branch_name=feature-force-absolute-reparent


editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.browser2.tsx
```diff
@@ -326,9 +326,7 @@ describe('Absolute Reparent Strategy', () => {
 
     await renderResult.getDispatchFollowUpActionsFinished()
 
-    expect(getPrintedUiJsCode(renderResult.getEditorState())).toEqual(
-      createCodeForProject(-960, -950),
-    )
+    expect(getPrintedUiJsCode(renderResult.getEditorState())).toEqual(createCodeForProject(40, 50))
```
balazsbajorics, line 329: this line looks sus to me

editor/src/components/canvas/canvas-strategies/absolute-reparent-to-flex-strategy.spec.browser2.tsx
```diff
@@ -357,7 +357,7 @@ describe('Absolute Reparent To Flex Strategy', () => {
     await renderResult.getDispatchFollowUpActionsFinished()
 
     const dragDelta = windowPoint({
-      x: firstFlexChildEnd.x - absoluteChildCenter.x + 50,
+      x: firstFlexChildEnd.x - absoluteChildCenter.x,
```
balazsbajorics, line 360: why was this change needed?

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -87,25 +85,19 @@ export function reparentStrategyForParent(
...
-    newParentMetadata?.specialSizeMeasurements.providesBoundsForAbsoluteChildren ?? false
-
   const parentIsFlexLayout = MetadataUtils.isFlexLayoutedContainer(newParentMetadata)
-  const parentIsStoryboard = EP.isStoryboardPath(newParent)
 
   if (allDraggedElementsAbsolute) {
     if (parentIsFlexLayout) {
       return { strategy: 'ABSOLUTE_REPARENT_TO_FLEX', newParent: newParent }
-    }
-    if (parentProvidesBoundsForAbsoluteChildren || parentIsStoryboard) {
```
balazsbajorics, line 100: this is definitely an unfinished part of the behavior. we shouldn't automatically reparent (without the user pressing _something_) into an element that doesn't provide bounds. 

we should leave at least a TODO comment here, but I wonder if this is a merge blocker.

balazsbajorics (approved):
I don't like the fact that we now allow the _default_ absolute reparent to include parents that don't provide bounds for children

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/absolute-reparent-strategy.spec.browser2.tsx
Rheeseyb: This is because of the `ifAllowedToReparent` that was previously missing from the absolute reparent strategy (but present in all flex reparenting strategies). Previously the absolute reparent would actually be applied, but wouldn't _actually_ reparent, meaning it was acting as an absolute move instead (whilst the picker would still claim it was performing a reparent). Now this will instead show the "not allowed" cursor, and will leave the element as it is (unless you explicitly switch to the "Move" strategy)

editor/src/components/canvas/canvas-strategies/absolute-reparent-to-flex-strategy.spec.browser2.tsx
Rheeseyb: This was dragging over the centre of an element that is now a valid parent, so now it is instead being dragged just before it

editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
Rheeseyb: Yeah, this is what I meant when I said that the absolute conversion is too easy to trigger, as this is still the same case (right now) when dragging a non-absolute element to move it when it is already in a parent that doesn't provide bounds. This is what triggered the other spike around finding other valid targets for an interaction. I wouldn't say this is a merge blocker, but it's definitely something to keep at the forefront of our minds whilst we work on that spike.

PR #2554, opened 2022-09-13
Fix: make the Unified Reparent Fitness Function Tests run on the CI

balazsbajorics:
This PR enables the two existing unified reparent strategy picker / fitness function tests.

There was an issue with Sinon fakeTimers (in `keyboard-absolute-move-resize.spec.browser2.tsx`) that somehow broke these tests. I couldn't figure out what is the problem, so I'm going with the hackiest solution to sidestep the issue: renaming this test file so it runs _before_ the keyboard test file. I'm sorry but at least I get to resume work on the actual tests. I left a comment on those fake timers that we need to fix them.


Rheeseyb (approved):
This is horrible and I don't like it. Approved.

enidemi (approved):

PR #2555, opened 2022-09-14
fix long paths in insert menu

bkrmendy:
## This change:
image

## Pizza:
image

### DIFF
--- a/editor/src/components/canvas/ui/floating-insert-menu.tsx
+++ b/editor/src/components/canvas/ui/floating-insert-menu.tsx
@@ -4,13 +4,13 @@ import React from 'react'
 import { jsx } from '@emotion/react'
 import { CSSObject } from '@emotion/serialize'
 import WindowedSelect, {
-  ActionMeta,
   InputActionMeta,
+  OptionProps,
   StylesConfig,
   ValueType,
 } from 'react-windowed-select'
 
-import { getControlStyles } from '../../../uuiui-deps'
+import { getControlStyles, Utils } from '../../../uuiui-deps'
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/ui/floating-insert-menu.tsx
```diff
@@ -297,13 +299,7 @@ function useComponentSelectorStyles(): StylesConfig<InsertMenuItem, false> {
...
         return {
           // ...styles,
-          color: colorTheme.fg7.value,
-          height: 25,
-          right: 12,
-          position: 'absolute',
-          display: 'flex',
-          alignItems: 'center',
-          pointerEvents: 'none',
+          display: 'none',
```
Rheeseyb, line 302: I'm assuming this was unintentionally committed? I don't believe we want to remove the group headings.

balazsbajorics (approved):

editor/src/components/canvas/ui/floating-insert-menu.tsx
bkrmendy: I think we do! the reason is that group headings are something that react-select gives us, but the way we want the insert menu to look like it just doesn't work as a heading (since we want a clickable element in every row). So to make this happen, I removed the group heading via the css here, and added the same label via `CustomOption`

bkrmendy: there might be betters ways to do this, I think this is the most straightforward and leads to the fewest surprises

Rheeseyb (approved):

PR #2560, opened 2022-09-15
look for parent applicable strategy

bkrmendy:
# Changes made
- [x] add a new superstrategy that walks up the dom tree to look for a strategy which is not `Move (Abs)`, that will be applied to the parent of the element being interacted with
- [x] update tests accordingly

# TODO
- [x] display outline with dragged parent
- [x] highlight selection in navigator
- [x] highlight all affected elements in navigator, with a different color


seanparsons:
Are those TODOs for a subsequent PR?

editor/src/components/canvas/canvas-strategies/look-for-applicable-parent.tsx
```diff
@@ -0,0 +1,275 @@
...
+      strategyState.startingMetadata,
+      strategyState.startingAllElementProps,
+    )
+
+    if (result == null || result.strategies.length === 0) {
+      return defaultName
+    }
+
+    const fittestStrategy = result.strategies.sort(
+      // TODO: better/idiomatic way of sorting
```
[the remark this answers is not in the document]
seanparsons: In fact we do exactly this in `getApplicableStrategiesOrderedByFitness`.

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/look-for-applicable-parent.tsx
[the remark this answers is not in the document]
bkrmendy: perfect, thanks, fixing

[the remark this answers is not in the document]
bkrmendy: fixed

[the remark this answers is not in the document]
bkrmendy: and no todos remaining as far as I can see

editor/src/components/canvas/canvas-strategies/look-for-applicable-parent.tsx
```diff
@@ -0,0 +1,298 @@
...
+      patchedCanvasState,
+      interactionSession,
+      strategyState,
+    )
+
+    return strategyApplicationResult(
+      [
+        ...chosenStrategyApplicationResult.commands,
+        highlightElementsCommand(effectiveTarget),
+        updateSelectedViews('mid-interaction', originalTarget),
```
balazsbajorics, line 135: is this line still needed here? is this defensive against `chosenStrategyApplicationResult.commands` overriding the selectedViews? if so, I'd change it from 'mid-interaction' to 'always'

editor/src/components/canvas/canvas-strategies/look-for-applicable-parent.tsx
```diff
@@ -0,0 +1,298 @@
...
+  return {
+    ...canvasState,
+    interactionTarget: {
+      type: 'TARGET_PATHS',
+      elements: path,
+    },
+  }
+}
+
+function isApplicableInner(
```
balazsbajorics, line 175: I think this could use a better name, if I understand correctly, this is the function that returns a potential ancestor's applicable strategy. the name implies to me that it only returns isApplicable ~boolean

editor/src/components/canvas/canvas-strategies/look-for-applicable-parent.tsx
```diff
@@ -0,0 +1,298 @@
...
+        updateSelectedViews('mid-interaction', originalTarget),
+        setCursorCommand('mid-interaction', CSSCursor.MovingMagic),
+      ],
+      chosenStrategyApplicationResult.customStatePatch,
+      chosenStrategyApplicationResult.status,
+    )
+  },
+}
+
+function* elementAncestry(path: ElementPath) {
```
balazsbajorics, line 144: wow the first generator function in our codebase!

editor/src/components/canvas/canvas-strategies/look-for-applicable-parent.tsx
```diff
@@ -0,0 +1,298 @@
...
+  canvasState: InteractionCanvasState,
+  interactionSession: InteractionSession,
+  metadata: ElementInstanceMetadataMap,
+  allElementProps: AllElementProps,
+): IsApplicableTraverseResult | null {
+  if (interactionSession.interactionData.type !== 'DRAG') {
+    return null
+  }
+
+  const strategiesMinusTraverse = RegisteredCanvasStrategies.filter(
```
balazsbajorics, line 185: I'm still a bit heartbroken you didn't make a recursive lookup instead :)

editor/src/components/canvas/canvas-strategies/look-for-applicable-parent.tsx
bkrmendy: the reason that I included it is that in this specific case the generator function doens't add a lot of cognitive overhead (besides the yield statement), and it can bail out as soon as a suitable parent was found, no need to "materialize" all the parents in the hierarchy, although very deep hierarchies are probably an edge case

balazsbajorics (approved):
looks good so far

PR #2564, opened 2022-09-15
feat(editor) More flexible fallback to absolute reparenting

Rheeseyb:
**Problem:**
With the newly enabled forcible absolute reparenting, it is a bit too easy to accidentally perform the wrong kind of reparenting when dragging.

**Fix:**
Allow more fine grained control over whether or not a forced absolute reparent happens, by providing 2 versions of each absolute reparent strategy (absolute to absolute, and flex to absolute).

In the base case strategy we only allow absolute reparenting onto targets that provide the coordinate system bounds to their children. As this is highly likely to be a desired outcome, we give this case a high weight (3), but also allow it to be a fallback option with a lower weight (2) if it isn't the _most_ likely desired outcome (so e.g. reparenting into a flex container that also provides the coord bounds most likely means the user intends it to be inserted as a flex element, but they can still hit tab to insert it with absolute positioning).

In the forced strategy we allow reparenting onto any target that accepts children, but give this a much lower weight (0.5) as it is highly likely that it wouldn't be the desired outcome, but we _do_ still want to provide it as an option.

I've also added some tests that check that these strategies are possible, but aren't the default option. To support those tests I made a tweak to the `renderTestEditorWithModel` function to allow tests to define which strategies are available during the test (this functionality already existed for the test dispatch function, but these tests don't call that function directly).

**Edit:**
There was an issue with the previous version of this, as the `startingTargetParentToFilterOut` was now no longer capturing all of the data required. The issue is that really we now need to capture _both_ possible starting target parents (based on whether or not the reparent will require coordinate system bounds). This has now been addressed.


editor/src/components/canvas/canvas-strategies/reparent-strategy-helpers.ts
```diff
@@ -241,6 +298,19 @@ export function getReparentTargetUnified(
   )
 
   const filteredElementsUnderPoint = allElementsUnderPoint.filter((target) => {
+    let validParentForFlexOrAbsolute = cmdPressed
```
balazsbajorics, line 301: I wonder if pressing cmd should "force" the other strategy to win instead of letting the strict strategy become permissive

editor/src/templates/editor-canvas.tsx
```diff
@@ -434,13 +434,15 @@ export function runLocalCanvasAction(
             action.interactionSession.interactionData.originalDragStart,
             action.interactionSession.interactionData.drag ?? zeroCanvasPoint,
           )
+          // FIXME This probably means we need both
```
balazsbajorics, line 437: yeah, it totally means that! I think this should be fixed before merge, because this can lead to subtle but incredibly annoying bugs. Unfortunately I still didn't write the test cases yet, but I'll make a note to explicitly write a test that would break because of this change

balazsbajorics (commented):
this is really nice! I think the FIXME comment needs fixing before the merge, but otherwise :chefs_kiss:

enidemi (approved):

editor/src/components/canvas/canvas-strategies/absolute-reparent-to-flex-strategy.spec.browser2.tsx
```diff
@@ -357,7 +354,7 @@ describe('Absolute Reparent To Flex Strategy', () => {
     await renderResult.getDispatchFollowUpActionsFinished()
 
     const dragDelta = windowPoint({
-      x: firstFlexChildEnd.x - absoluteChildCenter.x,
+      x: firstFlexChildEnd.x - absoluteChildCenter.x + 50,
```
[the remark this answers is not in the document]
Rheeseyb: Yeah, it was removed in #2553 because with that PR it would result in an undesired re-parent. This PR fixes that by making the forced re-parent the lower weighted strategy (as the target doesn't provide coordinate bounds)

editor/src/templates/editor-canvas.tsx
Rheeseyb: Ok this has now been fixed

editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -59,6 +59,23 @@ export function isKeyboardInteractionData(
 
 export type UpdatedPathMap = { [oldPathString: string]: ElementPath }
 
+export interface ReparentTargetsToFilter {
+  'use-strict-bounds': ReparentTarget
+  'allow-missing-bounds': ReparentTarget
+}
```
balazsbajorics, line 65: omg you monster :D

balazsbajorics (approved):

PR #2566, opened 2022-09-16
a strategy proposal

balazsbajorics:


editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -60,6 +67,24 @@ export const RegisteredCanvasStrategies: Array<CanvasStrategy> = [
...
+  interactionSession: InteractionSession | null,
+  metadata: ElementInstanceMetadataMap,
+  allElementProps: AllElementProps,
+) => {
+  // this is where we could run helper function to determine what reparent strategies are active
+  // or we can just return a strategy object that we generate dynamically
+
+  // or I can just return an existing superstrategy but under a changed name!
+  return [{ ...absoluteMoveStrategy, name: 'Absolute Mojave Strategy' }]
+}
```
balazsbajorics, line 81: this is the interesting part of the proposal. you could have a meta canvas strategy that is the _only place_ that runs `reparentStrategyForParent`, and based on the results returns one _or more_ reparent strategies. The returned reparent strategies could be objects that are created dynamically, which means their apply function could be containing flags in their closure. that would mean that the `getReparentTargetUnified` would also be called once, instead of 1. first in the fitness function then 2. again in the apply function.

balazsbajorics: to give a more concrete example: https://github.com/concrete-utopia/utopia/pull/2564/files?diff=split&w=1#diff-3aea0ce258632d3874fceeef2dd91289ac9440ae8c8c2899b662b5b8afc078ecR31

the `getAbsoluteReparentStrategy` could be called from _inside_ this function, to return the two flavors of strategy.

manually explicitly listing the two flavors of reparent strategies is fine if there's two flavors, but it may mean that we are discouraged from exploring behaviors that involve two or more flags (which would lead to a permutational explosion of these flavored strategies)

balazsbajorics: on the other hand, the downside is that we'd lose a static list of allowed strategies, making some debugging harder. 

previously we agreed that it's probably a _good_ thing that strategies must be explicitly listed and it's not a problem if we have to list a dozen strategy permutations, because it makes debugging easier.

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -84,15 +109,17 @@ function getInteractionTargetFromEditorState(editor: EditorState): InteractionTa
...
   metadata: ElementInstanceMetadataMap,
   allElementProps: AllElementProps,
 ): Array<CanvasStrategy> {
-  return strategies.filter((strategy) => {
-    return strategy.isApplicable(canvasState, interactionSession, metadata, allElementProps)
-  })
+  return metaStrategies
+    .flatMap((ms) => ms(canvasState, interactionSession, metadata, allElementProps))
+    .filter((strategy) => {
+      return strategy.isApplicable(canvasState, interactionSession, metadata, allElementProps)
```
bkrmendy, line 121: probably checking whether a strategy is applicable can be rolled into the meta strategy as well?

bkrmendy (approved):

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
bkrmendy: i mean to comment not to approve 🤦‍♂️

PR #2568, opened 2022-09-19
highlight component instance in navigator

bkrmendy:
Command to highlight components in the navigator that are not selected in the canvas

### DIFF
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -380,6 +380,7 @@ export interface NavigatorState {
   dropTargetHint: DropTargetHint
   collapsedViews: ElementPath[]
   renamingTarget: ElementPath | null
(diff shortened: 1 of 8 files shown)


editor/src/components/editor/store/editor-state.ts
```diff
@@ -380,6 +380,7 @@ export interface NavigatorState {
   dropTargetHint: DropTargetHint
   collapsedViews: ElementPath[]
   renamingTarget: ElementPath | null
+  highlightedTargets: ElementPath[]
```
enidemi, line 383:
```suggestion
  highlightedTargets: Array<ElementPath>
```

enidemi (approved):

Rheeseyb (commented):
This is good, but we really shouldn't have referred to them as "highlighted", since we're already using that term elsewhere, meaning this will cause confusion later on. We need to rename this to something else.

PR #2571, opened 2022-09-20
Added Import From URL functionality.

seanparsons:
**Problem:**
We want to be able to copy projects from one environment to another easily so that sample projects can be created and then migrated across those environments.

**Fix:**
Similar to how the old Github import was implemented, a handler on load of the editor looks for an appropriate query parameter and then initiates the import flow.

The persistent model is pulled from the project via the project download endpoint, then a new project is created with that content and once the project content is available then the assets and images are pulled from the old project and uploaded into the new one.

**Notes:**
Until the code has been deployed to both source and target environments, it will not work across existing deployments due to the `Access-Control-Allow-Origin` header change that was needed.

**Commit Details:**
- Added `githuboptions` to `LeftMenuPanel` so that it can be included in focus checks, which fixes an issue with not being able to paste into the import from URL input field.
- Added the input field and button to hold the URL and trigger the import to `GithubPane`.
- Brought `GithubPane` to be more like the other panes in its use of utility components.
- Added `importFromProjectURL` for handling the lookup of the data involved in importing a project.
- Implemented `reuploadAssets` and `reuploadAsset` utility functions to upload the assets from the old project into the new one.
- Added `waitUntil` utility function.
- Added handling for the presence of the `import_url` query parameter which triggers the import itself.
- Added the `Access-Control-Allow-Origin` header for the endpoint that provides the `content.json` download so that it can be invoked across domains.

### DIFF
--- a/editor/src/utils/utils.ts
+++ b/editor/src/utils/utils.ts
@@ -933,6 +933,20 @@ function deduplicateBy<T>(key: (t: T) => string, ts: Array<T>): Array<T> {
   return results
 }
 
+export async function waitUntil(limitMS: number, check: () => boolean): Promise<boolean> {
+  if (check()) {
+    return true
+  } else if (limitMS <= 0) {
+    return false
+  } else {
+    return new Promise((resolve) => {
+      setTimeout(resolve, 5)
+    }).then(() => {
+      return waitUntil(limitMS - 5, check)
+    })
+  }
+}
+
 export default {
   generateUUID: generateUUID,
   assert: assert,
@@ -1094,4 +1108,5 @@ export default {
   findLastIndex: findLastIndex,
   timeLimitPromise: timeLimitPromise,
   deduplicateBy: deduplicateBy,
+  waitUntil: waitUntil,
 }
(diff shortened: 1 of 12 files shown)


editor/src/utils/utils.ts
```diff
@@ -933,6 +933,20 @@ function deduplicateBy<T>(key: (t: T) => string, ts: Array<T>): Array<T> {
...
+    return false
+  } else {
+    return new Promise((resolve) => {
+      setTimeout(resolve, 5)
+    }).then(() => {
+      return waitUntil(limitMS - 5, check)
+    })
+  }
+}
+
```
Rheeseyb, line 949: Please can you move this function to the `/src/core/shared/` directory rather than this file (which we really need to bin off)

Rheeseyb (approved):

balazsbajorics (approved):

editor/src/utils/utils.ts
seanparsons: Fixed in 68683d84b.

PR #2572, opened 2022-09-21
padding resize controls

bkrmendy:
## Try it [here](https://utopia.pizza/p/33698e8e-scrawny-ironclad/?branch_name=feature/padding-controls)

padddddd

## Problem
Paddings cannot be adjusted via canvas controls (a la Figma)

## Solution
This PR implements padding canvas controls

## Commit details
- Strategy to drive padding resize
- Canvas controls for padding resize
- boilerplate for canvas controls


editor/src/components/canvas/padding-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+    paddingLeft: parts.paddingLeft ?? whole.paddingLeft ?? defaults.paddingLeft,
+    paddingRight: parts.paddingRight ?? whole.paddingRight ?? defaults.paddingRight,
+  }
+}
+
+const pxValue = (number: CSSNumber): number | undefined =>
+  number.unit === 'px' || number.unit == null ? number.value : undefined
+
+function pxValueFromEither(value: Either<string, CSSNumber | undefined>): number | undefined {
+  if (isLeft(value) || value.value == null) {
```
seanparsons, line 76: If `pxValue` also returned an `Either` and handled a null input, you could flatMap it and use `defaultEither` to get the result.

seanparsons (approved):

editor/src/components/canvas/padding-utils.ts
bkrmendy: thanks! I've updated the code along your suggestion

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
```diff
@@ -0,0 +1,215 @@
...
+  'paddingRight',
+]
+
+export const setPaddingStrategy: CanvasStrategyFactory = (
+  canvasState,
+  interactionSession,
+  customStrategyState,
+) => {
+  const selectedElements = getTargetPathsFromInteractionTarget(canvasState.interactionTarget)
+  if (selectedElements.length === 0) {
```
Rheeseyb, line 54: We should filter here for `selectedElements.length !== 1`, as otherwise we show padding controls when we have multiple elements selected (but just don't act when they are being dragged)

bkrmendy: great point! updated here and in the rest of the strategy as well

editor/src/components/canvas/controls/select-mode/padding-resize-step-indicator.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const PaddingResizeStepIndicatorTestId = 'PaddingResizeStepIndicatorTestId'
+
+export interface PaddingResizeStepIndicatorProps {
+  currentPaddingValue: number
+  activeEdge: EdgePiece
+  dragDelta: CanvasVector
+  dragStart: CanvasPoint
+}
+
+export const PaddingResizeStepIndicator =
```
Rheeseyb, line 19: I think this needs to be renamed, as the word `step` in here implies that it is an indicator for a discrete value that the user can jump to, whereas this indicator is for showing the current padding value

bkrmendy: done!

editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
```diff
@@ -0,0 +1,262 @@
...
+  React.forwardRef<HTMLDivElement, ResizeContolProps>((props, ref) => {
+    const scale = useEditorState((store) => store.editor.canvas.scale, 'PaddingResizeControl scale')
+    const dispatch = useEditorState((store) => store.dispatch, 'PaddingResizeControl dispatch')
+    const canvasOffsetRef = useRefEditorState((store) => store.editor.canvas.roundedCanvasOffset)
+    const { maybeClearHighlightsOnHoverEnd } = useMaybeHighlightElement()
+
+    const colorTheme = useColorTheme()
+
+    const [hidden, setHidden] = React.useState<boolean>(true)
+    const [hoverStart, hoverEnd] = useHoverWithDelay(200, (h) => setHidden(!h))
```
Rheeseyb, line 60: Shouldn't this just be calling `setHidden(false)`?

Rheeseyb: Ok nopr, it seems I've misunderstood what the param for `useHoverWithDelay` was

bkrmendy: should I come up with a better name for `useHoverWithDelay`?

Rheeseyb (approved):

Rheeseyb:
Nice work!

PR #2575, opened 2022-09-21
wrapping the flow reorder indicator

enidemi:
**Problem:**
The flow reorder indicator is not that accurate, when reaching the end of the slider it will restart from the beginning and it's not visible from the UI.

**Fix:**
Show 2 reorder indicators on the edges in this case, with overflow set to hidden.

**Commit Details:**
- updating the flow slider control with optionally showing 2 indicators
- rename variables as there are no more Animations

### DIFF
--- a/editor/src/components/canvas/controls/flow-slider-control.tsx
+++ b/editor/src/components/canvas/controls/flow-slider-control.tsx
@@ -125,8 +125,10 @@ export const FlowSliderControl = React.memo(() => {
               display: 'flex',
               alignItems: 'center',
               boxSizing: 'content-box',
+              overflow: 'hidden',
             }}
           >
+            <ReorderIndicators startingIndex={startingIndex} siblings={siblings} />
             {siblings.map((s) => {
               return (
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/controls/flow-slider-control.tsx
```diff
@@ -179,20 +168,36 @@ const AnimatedReorderIndicator = React.memo((props: AnimatedReorderIndicatorProp
     }
   }, 'FlowSliderControl indicatorOffset')
 
+  // when reaching the end of the slider it will restart from the beginning, a second indicator is shown
+  if (indicatorOffset > siblings.length - 1) {
```
balazsbajorics, line 172: techinally you could _always_ render the two indicators :D one would be hidden most of the time because of the overflow anyways

balazsbajorics (approved):

bkrmendy (approved):

Rheeseyb (commented):
Nice!

PR #2580, opened 2022-09-22
Click to insert as a strategy

gbalint:
**Problem:**
Click to insert was broken after draw to insert was introduced as a strategy. Furthermore, click to insert always inserted with absolute positioning, even into a flex container.

**Fix:**
Extend the draw-to-insert strategy so it can insert with clicking too.

We can detect that a "drag" is a "click" when the drag is null in the interaction, and the interaction is just finished (so a mouse up has happened without move)

To make this possible, we need to know in the `apply` function if it is running when interaction is finished or earlier. To achieve this I added a `lifecycle` parameter to the `apply` function, and I fill that in in `dispatch-strategies`.

This is very similar to the command lifecycle, but not the same. This lifecycle can make it possible to emit different commands from the strategy in the different lifecycle phases. 

Theoretically we could remove lifecycle management from commands, and always only emit the commands which are relevant in that lifecycle.

E.g. instead of 

```typescript
     return strategyApplicationResult([
        ...commandsForSelectedElements.commands,
        pushIntendedBounds(commandsForSelectedElements.intendedBounds),
        updateHighlightedViews('mid-interaction', []),
        setElementsToRerenderCommand(selectedElements),
        setCursorCommand('mid-interaction', CSSCursor.Select),
      ])
```

we could write:

```typescript
   const alwaysCommands = [
        ...commandsForSelectedElements.commands,
        pushIntendedBounds(commandsForSelectedElements.intendedBounds),
        setElementsToRerenderCommand(selectedElements),
        setCursorCommand('mid-interaction', CSSCursor.Select),
      ])

  return lifecycle === 'mid-interaction' ? strategyApplicationResult([
        ...alwaysCommands,
       updateHighlightedViews([]),
    ]) :  strategyApplicationResult(alwaysCommands)
```

But I believe this is just way more complicated to use, so I decided to keep lifecycle management inside commands too.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
@@ -83,7 +83,7 @@ export const dragToInsertStrategy: CanvasStrategy = {
       ? 1
       : 0
   },
-  apply: (canvasState, interactionState, strategyState, lifecycle) => {
+  apply: (canvasState, interactionState, strategyState, strategyLifecycle) => {
     const insertionSubjects = getInsertionSubjectsFromInteractionTarget(
       canvasState.interactionTarget,
     )
@@ -105,7 +105,7 @@ export const dragToInsertStrategy: CanvasStrategy = {
             interactionState,
             transient,
             insertionCommands,
-            lifecycle,
+            strategyLifecycle,
           )
         },
       )
--- a/editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
@@ -82,7 +82,7 @@ export const dragToInsertStrategy: CanvasStrategy = {
       ? 1
       : 0
   },
-  apply: (canvasState, interactionState, strategyState) => {
+  apply: (canvasState, interactionState, strategyState, lifecycle) => {
     const insertionSubjects = getInsertionSubjectsFromInteractionTarget(
       canvasState.interactionTarget,
     )
@@ -104,6 +104,7 @@ export const dragToInsertStrategy: CanvasStrategy = {
             interactionState,
             transient,
             insertionCommands,
+            lifecycle,
           )
         },
       )
@@ -196,6 +197,7 @@ function runTargetStrategiesForFreshlyInsertedElement(
   interactionState: InteractionSession,
   commandLifecycle: 'mid-interaction' | 'end-interaction',
   insertionSubjects: Array<{ command: InsertElementInsertionSubject; frame: CanvasRectangle }>,
+  strategyLifeCycle: 'mid-interaction' | 'end-interaction',
 ): Array<EditorStatePatch> {
   const canvasState = pickCanvasStateFromEditorState(editorState, builtInDependencies)
 
@@ -264,6 +266,7 @@ function runTargetStrategiesForFreshlyInsertedElement(
       patchedCanvasState,
       interactionState,
       patchedStrategyState,
+      strategyLifeCycle,
     ).commands
 
     return foldAndApplyCommandsInner(editorState, [], [], reparentCommands, commandLifecycle)
--- a/editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
@@ -5,6 +5,7 @@ import {
   emptyStrategyApplicationResult,
   getInsertionSubjectsFromInteractionTarget,
   InteractionCanvasState,
+  InteractionLifecycle,
   strategyApplicationResult,
   targetPaths,
 } from './canvas-strategy-types'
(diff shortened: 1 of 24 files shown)


editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
```diff
@@ -196,6 +197,7 @@ function runTargetStrategiesForFreshlyInsertedElement(
   interactionState: InteractionSession,
   commandLifecycle: 'mid-interaction' | 'end-interaction',
   insertionSubjects: Array<{ command: InsertElementInsertionSubject; frame: CanvasRectangle }>,
+  strategyLifeCycle: 'mid-interaction' | 'end-interaction',
```
bkrmendy, line 200: can we factor this out into a type? (to avoid duplication)

gbalint: Good point, fixed in 31f76c49593651c9919a44ad73edddbbe00d2720

enidemi (approved):

balazsbajorics:
> But I believe this is just way more complicated to use, so I decided to keep lifecycle management inside commands too.

so when I added lifecycle to commands, I started from adding lifecycle to the apply function. but then I realized the code can be much more succinct if the lifecycle is in the commands. but yeah, it is best if we expose this information to the apply function _and_ the commands as well

editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
```diff
@@ -194,8 +196,9 @@ function runTargetStrategiesForFreshlyInsertedElement(
   editorState: EditorState,
   strategyState: StrategyState,
   interactionState: InteractionSession,
-  commandLifecycle: 'mid-interaction' | 'end-interaction',
+  commandLifecycle: InteractionLifecycle,
   insertionSubjects: Array<{ command: InsertElementInsertionSubject; frame: CanvasRectangle }>,
+  strategyLifeCycle: InteractionLifecycle,
```
balazsbajorics, line 201: oh this will break my brain the next time I read this code :D I understand the purpose of the two lifecycles in this situation, but somehow their name is not distinct enough... 🤔 
But I don't have a better name to propose

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
```diff
@@ -60,6 +59,31 @@ async function fireDragEvent(
...
+      new MouseEvent('mousedown', {
+        bubbles: true,
+        cancelable: true,
+        clientX: point.x,
+        clientY: point.y,
+        buttons: 1,
+      }),
+    )
+
+    fireEvent(
```
balazsbajorics, line 75: I wonder if it'd be worth dispatching a click as well (because in real life a click event would follow). to make the test less synthetic. but I know it can never be 100% simulation because you are dispatching the events only on the canvasControlsLayer

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
```diff
@@ -191,6 +215,73 @@ describe('Inserting into absolute', () => {
...
+
+    const targetElement = renderResult.renderedDOM.getByTestId('bbb')
+    const targetElementBounds = targetElement.getBoundingClientRect()
+    const canvasControlsLayer = renderResult.renderedDOM.getByTestId(CanvasControlsContainerID)
+
+    const point = slightlyOffsetWindowPointBecauseVeryWeirdIssue({
+      x: targetElementBounds.x + 65,
+      y: targetElementBounds.y + 55,
+    })
+
```
balazsbajorics, line 230: I think if you put here a mouse move event dispatch and then check the highlightedViews, that would be a pretty cheap way to check that we are highlighting the target parent. maybe it's also worth making a specific test that shows an edge case in which highlight differs from regular select mode highlight

balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
Rheeseyb: I've mentally wrestled with this one too. We really should be dispatching all of the events that would fire (we should probably just write helper functions since we do this so much anyway), but it will still be a very artificial situation. Though at the least we do know that the events will bubble, so it will still be closer to real life anyway.

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
gbalint: More consistent naming in 5e101ff7b481769539deeba0b2bda173ff843195

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
gbalint: Ok, I added dispatching click event too in d86dfd425b6faad711e0fb82164b33eb0cfc71d0

seanparsons (approved):

PR #2583, opened 2022-09-26
introducing Meta Strategies

bkrmendy:
**Problem**
1. The way the getReparentTargetUnified and findReparentStrategy work is that we have 4 (or more) reparent strategies, and their fitness functions are all calling the same getReparentTargetUnified function, and check if the returned strategy name matches their own name. Then the apply function of the winning reparent strategy will again call the same getReparentTargetUnified to get the target elementPath.
2. A superstrategy such as the new insert mode currently doesn't support switching sub-strategies, and currently it could only support switching strategies if the superstrategy itself would be manually added to the RegisteredCanvasStrategies array multiple times, with strategies that are only different from each other in a name + a flag for what internal strategy they wrap around.
3. The `lookForApplicableParentStrategy` runs the same `lookForParentApplicableStrategy` helper function in the fitness function and apply function, and in the name function. It renders a static `controlsToRender` array which may not contain the controls of the applicable parent strategy it found. So we should need to turn controlsToRender into a function too, and then `lookForApplicableParentStrategy` would also have to reuse the same lookup for a fourth time.

**Fix**
Create a new type, 
```typescript
export type MetaCanvasStrategy = (
  canvasState: InteractionCanvasState,
  interactionSession: InteractionSession | null,
) => Array<CanvasStrategy>
```
The MetaCanvasStrategy is a function that takes the same props as the `CanvasStrategy.isApplicable`, and returns an array of applicable CanvasStrategies. 
This achieves three things:
1. it obsoletes CanvasStrategy.isApplicable, 
2. but it also gives us an opportunity to dynamically assemble the CanvasStrategy object. This means that strategies that rely on logic in helper functions can call that helper function once, at the time the MetaCanvasStrategy function is called. The MetaCanvasStrategy function can then take the values returned by the helper function, and use that same value to populate the name field, the controlsToRender field, and it can also create the `apply` function on the fly, which means the apply function can use values stored in its closure, so it also doesn't need to call the helper function again.
3. It also means that it allows multiple strategies that are currently orchestrated by a shared helper function to be explicitly orchestrated by a single MetaCanvasStrategy function. So for example the various reparenting edge cases can be much more explicit now, instead of coded implicitly into fitness values across multiple files.

**Also**
The PR introduces the new `lookForApplicableParentMeta` strategy, our first real metaStrategy, courtesy of @bkrmendy


enidemi (approved):

editor/src/components/canvas/canvas-strategies/look-for-applicable-parent-strategy.tsx
```diff
@@ -1,153 +1,88 @@
...
-    )
-      ? 1
-      : 0
-  },
+function tweakStrategy(
+  strategy: CanvasStrategy,
+  effectiveTarget: Array<ElementPath>,
+  componentsInSubtree: Array<ElementPath>,
+): CanvasStrategy {
+  const { controlsToRender } = strategy
```
enidemi, line 62: as the canvas controls might use `editorState.selectedViews` this might not be correct, depending on the control

editor/src/components/canvas/canvas-strategies/look-for-applicable-parent-strategy.tsx
```diff
@@ -1,153 +1,88 @@
...
 
-    if (result == null) {
-      return emptyStrategyApplicationResult
-    }
+  const fitness: CanvasStrategy['fitness'] = () => 1
 
-    const { strategies, effectiveTarget, componentsInSubtree } = result
-    if (strategies.length < 1) {
-      return emptyStrategyApplicationResult
+  const apply: CanvasStrategy['apply'] = (c, i, s, l) => {
```
seanparsons, line 68: Probably preferable to give these parameters their longer names.

seanparsons (approved):

gbalint (approved):

editor/src/components/canvas/canvas-strategies/look-for-applicable-parent-strategy.tsx
balazsbajorics: I made a note for a follow-up fix

Rheeseyb (approved):
Nice work!

PR #2587, opened 2022-09-27
Still honour startingTargetParentsToFilterOut when holding cmd

Rheeseyb:
**Problem:**
We were ignoring `startingTargetParentsToFilterOut` during reparenting when `cmd` was held. This was erroneously added logic, and meant that holding `cmd` would immediately reparent an element to whatever was behind it at the start of a drag.

**Fix:**
Remove the logic that was using `cmd` in this way. This did cause an issue with insertion, as that is using the same reparenting logic, so to fix that I have patched the `interactionState` to set `startingTargetParentsToFilterOut` to `null` before calling the reparenting logic during an insertion.

I've also added a number of tests for `startingTargetParentsToFilterOut`, as well as a couple of insertion tests to ensure we can still insert into an element that is smaller than the inserted element.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
@@ -250,6 +250,7 @@ function runTargetStrategiesForFreshlyInsertedElement(
   const patchedInteractionState = {
     ...interactionState,
     interactionData: patchedInteractionData,
+    startingTargetParentsToFilterOut: null,
   }
 
   const { strategy } = findCanvasStrategy(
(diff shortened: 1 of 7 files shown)


editor/src/components/canvas/canvas-strategies/drag-to-insert-strategy.tsx
```diff
@@ -265,7 +266,7 @@ function runTargetStrategiesForFreshlyInsertedElement(
   } else {
     const reparentCommands = strategy.strategy.apply(
       patchedCanvasState,
-      interactionState,
+      patchedInteractionState,
```
gbalint, line 269: Nice catch! 🙇

gbalint (approved):

PR #2589, opened 2022-09-27
insert image from finder

bkrmendy:
## Problem
Image insertion seems to be regressed (both drag and drop from Finder and drag and drop from the Utopia file browser). This PR aims to fix these problems.

## Fixes
Images can be dragged and dropped from Finder into the hierarchy, by leveraging the drag to insert superstrategy (and the whole strategy mechanism).  This behaviour is slightly different when the user doing the drag drop is logged out, in this case, the base64 data from the image is inserted as src and the image is not uploaded (since that would require a logged in user).

Images can also be dragged and dropped to the canvas from the Utopia File Browser Tab (UFTB).

## Limitations
- on drag and drop, only one dropped image is inserted, even if multiple images are dropped - turn the image drop logic into a "for loop" over the dropped images
- dragging from the Utopia file browser is only possible via a "drag handle", because otherwise dnd clashes with react DnD. We should look for a way to integrate this with react dnd
- when base64 data is inserted, the editor becomes really slow (probably due to the size of the base64 string in the project)
- inserting two images with the same name overwrites the exiting one


seanparsons (approved):

editor/src/templates/image-dnd.spec.browser2.tsx
```diff
@@ -0,0 +1,403 @@
...
+      items: fileList.map((f) => ({ kind: 'file', getAsFile: () => f })),
+      files: {
+        item: (itemIndex: number) => fileList[itemIndex],
+        length: fileList.length,
+      },
+    },
+  })
+
+  return fileDropEvent
+}
```
Rheeseyb, line 395: This function living here is fine for now, but if we find ourselves needing to produce more drag and drop events in other tests we should definitely abstract part of it into `event-helpers.test-utils`

editor/src/templates/image-dnd.spec.browser2.tsx
```diff
@@ -0,0 +1,403 @@
...
+      makeDragEvent('drag', canvasControlsLayer, { x: 5, y: 5 }, [file]),
+    )
+
+    fireEvent(canvasControlsLayer, makeDragEvent('drag', canvasControlsLayer, endPoint, [file]))
+
+    fireEvent(canvasControlsLayer, makeDragEvent('drop', canvasControlsLayer, endPoint, [file]))
+
+    await editor.getDispatchFollowUpActionsFinished()
+
+    await wait(1000) // read the image
```
Rheeseyb, line 324: This makes me very sad. At the very least, can we reduce this timeout?

Rheeseyb (approved):

editor/src/templates/image-dnd.spec.browser2.tsx
bkrmendy: moved it for good measure

PR #2590, opened 2022-09-27
Feature/canvas flow reorder filter

enidemi:
**Fix:**
When reordering static elements we should skip these siblings as they are not a bit standing somewhere outside of the flow layout:
- absolute
- float
- relative with at least one position property set

The fallback strategy still shows the slider control for these (except for absolute)

**Commit Details:**
- added float to specialSizeMeasurements
- filter selected element in `isApplicable`
- filter potential reorder sibling target when dragging an element over these special siblings
- test

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/flow-reorder-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/flow-reorder-helpers.ts
@@ -12,23 +12,13 @@ import { stylePropPathMappingFn } from '../../inspector/common/property-path-hoo
 import { DeleteProperties, deleteProperties } from '../commands/delete-properties-command'
 import { SetProperty, setProperty } from '../commands/set-property-command'
 
-export function isValidFlowReorderTarget(
-  elementMetadata: ElementInstanceMetadata | null,
-  allElementProps: AllElementProps,
-): boolean {
+export function isValidFlowReorderTarget(elementMetadata: ElementInstanceMetadata | null): boolean {
   if (MetadataUtils.isPositionAbsolute(elementMetadata)) {
     return false
   } else if (elementMetadata?.specialSizeMeasurements.float !== 'none') {
     return false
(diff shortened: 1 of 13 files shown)


editor/src/components/canvas/canvas-strategies/flow-reorder-helpers.ts
```diff
@@ -7,26 +7,45 @@ import {
...
+export function isValidFlowReorderTarget(
+  elementMetadata: ElementInstanceMetadata | null,
+  allElementProps: AllElementProps,
+): boolean {
+  if (MetadataUtils.isPositionAbsolute(elementMetadata)) {
+    return false
+  } else if (elementMetadata?.specialSizeMeasurements.float !== 'none') {
+    return false
+  } else if (MetadataUtils.isPositionRelative(elementMetadata) && elementMetadata != null) {
+    const styleProps = allElementProps[EP.toString(elementMetadata.elementPath)]?.style
```
balazsbajorics, line 24: allElementProps uses the spy as the data source, which means that it doesn't guarantee the computedStyle won't have a left top right or bottom set. I think we should use the specialSizeMeasurements here too (I don't know if it need a new field or not)

balazsbajorics (commented):
should we apply the same sibling exclusion for Flex reorder? I think so!

balazsbajorics (commented):
should we apply the same sibling exclusion for Flex reorder? I think so!

enidemi:
> should we apply the same sibling exclusion for Flex reorder? I think so!

yes! relatives in flex need to be excluded from reorder too :)

editor/src/components/canvas/canvas-strategies/flow-reorder-helpers.ts
enidemi: I updated `specialSizeMeasurements` to contain a new field for this

balazsbajorics (approved):

Rheeseyb (commented):
👍

PR #2592, opened 2022-09-28
Save To Github

seanparsons:
**Problem:**
Currently projects are held solely inside the Utopia server and can't be easily retrieved without a lot of manual copying and pasting should the user want to use their project outside our editor.

**Fix:**
We now provide a field and button which when filled in and triggered respectively will create a branch in the repository containing the current content of the project.

The steps taken are as follows:
- Create a git tree containing the files within the project.
- Create a git commit which references the git tree created.
- Create a git ref (the real name of a branch) which references the git commit created.

**Commit Details:**
- Added tests for the `clientmodel` package.
- Replaced `PartialPersistentModel` with `PersistentModel` in `clientmodel` so that it can full cycle the representation.
- Simplified `getProjectContentsTreeFile`.
- Created `SaveToGithub` action in the frontend that triggers the side effecting action of saving to Github.
- Added `githubSettings` field to `PersistentModel`.
- Added model migration to add the `githubSettings` field.
- Added section to `GithubPane` for adding the target repository to save to.
- Made `EditorState.appID` not possible to omit as a field.
- Added `002.sql` database migration which makes `refresh_token` and `expires_at` nullable as the OAuth apps do not supply these at all.
- `getAuthorizationURI` now includes the scope to request the access token with.
- Modified `oauth2TokenToGithubAuthenticationDetails` now caters for the refresh token and expiry fields being optional.
- Added `accessTokenFromRefreshToken`.
- Some functions have been reworked as it makes more sense to parse `ProjectContentsTreeRoot` out from the model JSON rather than decoding to `PersistentModel` and extracting the project contents from that.
- Added `githubSaveEndpoint` as the entry to the calls for the logic that creates the Github branch.
- Created `callGithub` utility function for the REST API which does the common functionality for all the REST calls which is invoked by the specific functions like `createGitTree`.
- Added `useAccessToken` for handling the potential need for getting a new access token from the refresh token, which is done transparently to the action passed into it.

### DIFF
--- a/clientmodel/lib/src/Utopia/ClientModel.hs
+++ b/clientmodel/lib/src/Utopia/ClientModel.hs
@@ -3,7 +3,7 @@
 -}
 module Utopia.ClientModel where
 
-import           Control.Lens
+import           Control.Lens                                  hiding (children)
 import           Control.Monad.Fail
 import           Data.Aeson
 import           Data.Aeson.Lens
@@ -11,10 +11,19 @@ import           Data.Aeson.Types
 import           Data.Data
 import           Data.Generics.Product
 import           Data.Generics.Sum
-import qualified Data.HashMap.Strict   as M
-import           Data.Text             hiding (foldl', reverse)
+import qualified Data.HashMap.Strict                           as M
+import           Data.Text                                     hiding (foldl',
+                                                                reverse)
 import           Data.Typeable
 import           Relude
+import           Test.QuickCheck
+import           Test.QuickCheck.Arbitrary.Generic
+import           Test.QuickCheck.Instances.Text
+import           Test.QuickCheck.Instances.UnorderedContainers
+
+-- Slightly nonsense Arbitrary instance, but one that suffices for our needs.
+instance Arbitrary Value where
+  arbitrary = fmap (toJSON :: Text -> Value) arbitrary
 
 textToJSON :: Text -> Value
 textToJSON = toJSON
@@ -31,6 +40,10 @@ instance FromJSON ElementPath where
 
 instance ToJSON ElementPath where
   toJSON = genericToJSON defaultOptions
+
+instance Arbitrary ElementPath where
+  arbitrary = genericArbitrary
+  shrink = genericShrink
 
 data RevisionsState = ParsedAhead
                     | CodeAhead
@@ -52,6 +65,10 @@ instance ToJSON RevisionsState where
   toJSON CodeAhead   = textToJSON "CODE_AHEAD"
   toJSON BothMatch   = textToJSON "BOTH_MATCH"
 
+instance Arbitrary RevisionsState where
+  arbitrary = genericArbitrary
+  shrink = genericShrink
+
 data ParseFailure = ParseFailure
                   { diagnostics   :: Maybe [Value]
                   , parsedJSON    :: Maybe Value
@@ -65,6 +82,10 @@ instance FromJSON ParseFailure where
 
 instance ToJSON ParseFailure where
   toJSON = genericToJSON defaultOptions
+
+instance Arbitrary ParseFailure where
+  arbitrary = genericArbitrary
+  shrink = genericShrink
 
 data ParseSuccess = ParseSuccess
                   { imports                        :: Value
@@ -82,6 +103,10 @@ instance FromJSON ParseSuccess where
 instance ToJSON ParseSuccess where
   toJSON = genericToJSON defaultOptions
 
+instance Arbitrary ParseSuccess where
+  arbitrary = genericArbitrary
+  shrink = genericShrink
+
 data Unparsed = Unparsed
               deriving (Eq, Show, Generic, Data, Typeable)
 
@@ -90,6 +115,10 @@ instance FromJSON Unparsed where
 
 instance ToJSON Unparsed where
   toJSON = const $ object []
+
+instance Arbitrary Unparsed where
+  arbitrary = genericArbitrary
+  shrink = genericShrink
 
 data ParsedTextFile = ParsedTextFileFailure ParseFailure
                     | ParsedTextFileSuccess ParseSuccess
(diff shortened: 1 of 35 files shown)


clientmodel/lib/src/Utopia/ClientModel.hs
```diff
@@ -3,18 +3,27 @@
...
 import           Data.Typeable
 import           Relude
+import           Test.QuickCheck
+import           Test.QuickCheck.Arbitrary.Generic
+import           Test.QuickCheck.Instances.Text
+import           Test.QuickCheck.Instances.UnorderedContainers
+
+-- Slightly nonsense Arbitrary instance, but one that suffices for our needs.
+instance Arbitrary Value where
+  arbitrary = fmap (toJSON :: Text -> Value) arbitrary
```
Rheeseyb, line 26: Can all of these be moved to the `test` directory?

clientmodel/lib/src/Utopia/ClientModel.hs
```diff
@@ -197,6 +267,24 @@ instance FromJSON ProjectContentDirectory where
...
+  [ fmap ProjectContentsTreeDirectory $ resize 3 $ generateProjectContentDirectory depth
+  , fmap ProjectContentsTreeFile arbitrary
+  ]
+
+generateProjectContentDirectory :: Int -> Gen ProjectContentDirectory
+generateProjectContentDirectory depth = do
+  fullPath <- arbitrary
+  children <- liftArbitrary $ generateProjectContentsTree (depth - 1)
+  let directory = Directory
+  pure ProjectContentDirectory{..}
```
Rheeseyb, line 282: Likewise, can these also be moved please? It's hard to tell at first glance what code exists purely for tests vs what is the actual code

editor/src/components/editor/actions/actions.tsx
```diff
@@ -5006,6 +5010,25 @@ export const UPDATE_FNS = {
...
+      const updatedEditor = {
+        ...editor,
+        githubSettings: {
+          ...editor.githubSettings,
+          targetRepository: updatedRepo,
+        },
+      }
+      // Side effect - Pushing this to the server to get that to save to Github.
+      const persistentModel = persistentModelFromEditorModel(updatedEditor)
+      saveProjectToGithub(persistentModel)
```
Rheeseyb, line 5027: Is there a plan to show the status of this in the UI yet?

clientmodel/lib/src/Utopia/ClientModel.hs
seanparsons: So in Haskell creating orphan instances (typeclass instances not in the same file as their associated type) is frowned upon because it makes them difficult to use from multiple locations as you have to know which places to import them from.

clientmodel/lib/src/Utopia/ClientModel.hs
seanparsons: As these are used in the `Arbitrary` instances they should probably stay where they are for previously mentioned reasons.

clientmodel/lib/src/Utopia/ClientModel.hs
Rheeseyb: Urgh, that's a pain then. In that case, can I suggest we move all test code to the bottom of the file with a comment above stating that it is purely test code from that point onwards?

editor/src/components/editor/actions/actions.tsx
seanparsons: I'm currently working on a tweak that will give some feedback in the form of some toasts for success and failure. Which for the scope of this work should suffice, but once the work for loading is completed then we'll probably look at how to make more of the flow more git like.

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -2988,6 +2992,21 @@ export const UtopiaVSCodeConfigKeepDeepEquality: KeepDeepEqualityCall<UtopiaVSCo
     },
   )
 
+export const GithubRepoKeepDeepEquality: KeepDeepEqualityCall<GithubRepo> = combine2EqualityCalls(
+  (repo) => repo.owner,
+  createCallWithTripleEquals(),
```
balazsbajorics, line 2997: not strictly related to the PR but I wonder if we should make dedicated StringKeepDeepEquality and NumberKeepDeepEquality functions, because now if we change repo.owner to, say, an object, this will not break here and we will then run === equality on objects, which will always fail, and then that will bubble up to always fail the editor state deep equality too.

clientmodel/lib/src/Utopia/ClientModel.hs
seanparsons: I think that's just having two files in one file! :) While `Arbitrary` instances are used in test code they're not specifically test code themselves.

editor/src/components/editor/store/store-deep-equality-instances.ts
seanparsons: Or at least pass an explicit type to that function.

Rheeseyb (approved):

editor/src/components/editor/actions/actions.tsx
Rheeseyb: Yeah that makes sense 👍

editor/src/components/editor/store/store-deep-equality-instances.ts
Rheeseyb: Very very good shout!

clientmodel/lib/src/Utopia/ClientModel.hs
Rheeseyb: Yes, but, crucially it means someone can read the code and know immediately whether or not the code they are looking at exists purely for tests (without having to resort to knowing specific conventions)

seanparsons: I'm struggling to understand why that would be beneficial.

Rheeseyb: Code readability + code file organisation. For the same reasons we typically have separate test files rather than mixing the tests in with the rest of the code. If a year from now I need to delve into this code to remind myself how the logic works and to make a change to the code, I don't want to spend extra time reading and understanding irrelevant code. I can make the choice to either start from the actual code, or start from the tests, rather than having to read functions before understanding that they are test specific functions.

PR #2593, opened 2022-09-28
Browser Tests QOL Improvements

Rheeseyb:
**Problem:**
Our browser tests were inconsistent in their use of mouse and keyboard events, as well as their setting of the viewport, leading to many lost hours debugging incorrect simulated interactions.

**Fix:**
Create a helper file for firing keyboard and mouse events, as well as a setup file for calling `viewport.set(2200, 1000)` at the start of every browser test.

The goal of this work was to remove all other uses of `fireEvent`, but there are still 4 remaining test files that are calling it in slightly different (sometimes questionable) ways. To switch those over still requires further debugging, but I decided to leave those as they are now (and will strongly suggest that any future tests either use these helpers or add new helpers :) )


editor/mocha-setup-beforeall.js
```diff
@@ -0,0 +1,3 @@
+beforeEach(function () {
```
seanparsons, line 1: Glorious!

seanparsons (approved):
Cannot approve enough!

editor/patches/keycode+2.2.1.patch
```diff
@@ -0,0 +1,23 @@
...
+--- a/node_modules/keycode/index.js
++++ b/node_modules/keycode/index.js
+@@ -137,6 +137,10 @@ var aliases = exports.aliases = {
+   'spacebar': 32,
+   'pgup': 33,
+   'pgdn': 34,
++  'arrowleft': 37,
++  'arrowup': 38,
++  'arrowright': 39,
++  'arrowdown': 40,
```
balazsbajorics, line 20: should we submit a PR to keycode? :)

enidemi (approved):

PR #2595, opened 2022-09-30
reset drop target on drag end

bkrmendy:
## Problem
When a file browser dnd session ends outside the file browser, the drop target highlight doesn't get reset, and gives the impression of getting "stuck".

## Solution
Use the `end` callback from React DnD to reset the drop target at the end of each drop session.

## Commit details
- Add the `end` callback to the React DnD drag specification to reset the drop target to null


Rheeseyb (approved):

balazsbajorics (approved):
thanks for the PR description :)

PR #2596, opened 2022-09-30
reset interaction state when it ends outside the canvas

bkrmendy:
## Problem:
When a drag to move interaction finished outside the visible canvas bounds in insert mode (such as dragging an element from the insert menu but then dragging it back), the interaction is still committed as if it was intentionally aimed there.

This can lead to "surprises", such as finding a stray element *below* the right side bar.

## Fix:
In `handleCanvasEvent` in `editor-canvas.tsx`, if the event is a mouse up event in insert mode, check whether the event happened on or off the canvas, and whether the active control is `BOUNDING_AREA` (since we want to commit changes made through other controls). Do not apply changes if the event happened off the canvas. This way, an interaction can be cancelled by dragging out of the canvas and releasing the mouse.

The logic behind is is essentially to **discard** changes if `!isInsideCanvas && boundingAreaActive && isInsertModeActive` .

## Commit Details:
- add a new parameter to `handleCanvasEvent` to signal whether the event happened on or off the canvas
- in `handleCanvasEvent` in `editor-canvas.tsx`, add a check to see if the event happened on or off the canvas
- add test for the cancellation behaviour


gbalint (approved):

balazsbajorics (approved):

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
```diff
@@ -458,6 +459,71 @@ describe('Inserting into absolute', () => {
...
+    const endPoint = slightlyOffsetWindowPointBecauseVeryWeirdIssue({
+      x: targetElementBounds.x + 1005,
+      y: targetElementBounds.y + 1005,
+    })
+
+    mouseMoveToPoint(canvasControlsLayer, startPoint)
+
+    await startInsertMode(renderResult.dispatch)
+
+    mouseDragFromPointToPointNoMouseDown(canvasControlsLayer, startPoint, endPoint)
```
Rheeseyb, line 484: I'm confused - this wouldn't trigger an insertion anyway if there were no mousedown event. It seems more than likely that you should be triggering a full drag here. If you need the mouse events to fire at different DOM elements then it might make sense to call the `mouseDownAtPoint`, `mouseMoveToPoint` and `mouseUpAtPoint` events separately so that you can do that.

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
```diff
@@ -458,6 +459,71 @@ describe('Inserting into absolute', () => {
...
+
+    await renderResult.getDispatchFollowUpActionsFinished()
+
+    // Check that the inserted element is a child of bbb
```
Rheeseyb, line 488: This comment seems wrong :)

enidemi (approved):

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
```diff
@@ -458,6 +459,71 @@ describe('Inserting into absolute', () => {
...
+    expect(getPrintedUiJsCode(renderResult.getEditorState())).toEqual(
+      makeTestProjectCodeWithSnippet(`
```
balazsbajorics, line 490: if there's no expected changes, you should be able to reuse `inputCode` here

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
balazsbajorics: does this test break (as it should) on master?

bkrmendy: `Should drag to insert into targets smaller than the element` does

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
bkrmendy: fixed, thanks!

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
bkrmendy: so my understanding is that when I'm inserting an element via the insert menu, when I `mousedown` on an insert menu item, the insertion is initiated by the mousedown event handler,  and I wanted to model the same process. 

Is it better approach to 
- enter insert mode
- fire a mousedown event on the insert menu
- move the cursor to the canvas
- move the cursor somewhere far out
- fire mouseup and check whether the interaction was cancelled?
?

bkrmendy: OK I got it, I modified the test so that it break on master but succeeds on this branch, with all the required mouse events

Rheeseyb (approved):

PR #2602, opened 2022-10-03
Use lint-full command for type checked linting in CI

Rheeseyb:
**Problem:**
Type checked linting is slow, but we want it for catching floating promises

**Fix:**
Add a separate eslint config for type checked linting, and use that in CI (but not the pre-commit hook)


Rheeseyb:
Actually, change of plan, I'm going to switch this from "regular" and "full" to "lite" and "regular", using the lite version for the commit hooks

PR #2604, opened 2022-10-03
Feature/relative move

ruggi:
# Problem

Having a strategy to handle the movement of elements with `position: relative`.

# Fix

Added the `RELATIVE_MOVE` strategy.

## Notes & gotchas:

- It only triggers when the element has an explicit `position: "relative"` property. 
- If the element has defined offsets (at least one TLBR), the strategy will have a higher fitness score
- If the element doesn't have defined offsets, the strategy will have a lower fitness score (e.g. it will default to `FLOW_REORDER`)
- the offsets are set with TL priority
  - if the horizontal or vertical offset is missing, it will be added (T or L)
  - if the BR offsets are set, they will be updated as per the movement delta
- ⚠️ the condition to check whether the flow reorder strategy applies are changed to be applicable even if the element has `position: "relative"` so that the two strategies can be used alternatively on the same element. This _should_ be fine, but it'd be great if somebody with more experience on this could double check it's not gonna introduce issues
- ⚠️ I'm not super sure about which controls should be rendered for the strategy, I left it empty for now

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/relative-move-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/relative-move-strategy.spec.browser2.tsx
@@ -1,0 +1,393 @@
+import React from 'react'
+import { createBuiltInDependenciesList } from '../../../core/es-modules/package-manager/built-in-dependencies-list'
+import { elementPath } from '../../../core/shared/element-path'
+import {
+  ElementInstanceMetadata,
+  ElementInstanceMetadataMap,
+  SpecialSizeMeasurements,
+} from '../../../core/shared/element-template'
+import {
+  canvasPoint,
+  canvasRectangle,
+  CanvasVector,
+  localRectangle,
+} from '../../../core/shared/math-utils'
+import { ElementPath } from '../../../core/shared/project-file-types'
+import { emptyModifiers } from '../../../utils/modifiers'
+import { EditorState } from '../../editor/store/editor-state'
+import { foldAndApplyCommands } from '../commands/commands'
+import {
+  getEditorState,
+  makeTestProjectCodeWithSnippet,
+  testPrintCodeFromEditorState,
+} from '../ui-jsx.test-utils'
+import { absoluteMoveStrategy } from './absolute-move-strategy'
+import { pickCanvasStateFromEditorState } from './canvas-strategies'
+import { defaultCustomStrategyState } from './canvas-strategy-types'
+import { InteractionSession, StrategyState } from './interaction-state'
+import { createMouseInteractionForTests } from './interaction-state.test-utils'
+
+const prepareEditorState = (
+  codeSnippet: string,
+  selectedViews: Array<ElementPath>,
+): EditorState => {
+  return {
+    ...getEditorState(makeTestProjectCodeWithSnippet(codeSnippet)),
+    selectedViews: selectedViews,
+  }
+}
+
+const defaultMetadata: ElementInstanceMetadataMap = {
+  'scene-aaa': {
+    elementPath: elementPath([['scene-aaa']]),
+  } as ElementInstanceMetadata,
+  'scene-aaa/app-entity': {
+    elementPath: elementPath([['scene-aaa', 'app-entity']]),
+  } as ElementInstanceMetadata,
+  'scene-aaa/app-entity:aaa': {
+    elementPath: elementPath([['scene-aaa', 'app-entity'], ['aaa']]),
+  } as ElementInstanceMetadata,
+  'scene-aaa/app-entity:aaa/bbb': {
+    elementPath: elementPath([
+      ['scene-aaa', 'app-entity'],
+      ['aaa', 'bbb'],
+    ]),
+    specialSizeMeasurements: {
+      immediateParentBounds: canvasRectangle({ x: 0, y: 0, width: 400, height: 400 }),
+      coordinateSystemBounds: canvasRectangle({ x: 0, y: 0, width: 400, height: 400 }),
+    } as SpecialSizeMeasurements,
+    globalFrame: canvasRectangle({ x: 50, y: 50, width: 250, height: 300 }),
+    localFrame: localRectangle({ x: 50, y: 50, width: 250, height: 300 }),
+  } as ElementInstanceMetadata,
+}
+
+function dragByPixels(
+  editorState: EditorState,
+  vector: CanvasVector,
+  metadata: ElementInstanceMetadataMap = defaultMetadata,
+): EditorState {
+  const interactionSession: InteractionSession = {
+    ...createMouseInteractionForTests(
+      null as any, // the strategy does not use this
+      emptyModifiers, // the strategy does not use this
+      null as any, // the strategy does not use this
+      vector,
+    ),
+    latestMetadata: null as any, // the strategy does not use this
+    latestAllElementProps: null as any, // the strategy does not use this
+    startingTargetParentsToFilterOut: null,
+  }
+
+  const strategyResult = absoluteMoveStrategy.apply(
+    pickCanvasStateFromEditorState(editorState, createBuiltInDependenciesList(null)),
+    interactionSession,
+    {
+      currentStrategy: null as any, // the strategy does not use this
+      currentStrategyFitness: null as any, // the strategy does not use this
+      currentStrategyCommands: null as any, // the strategy does not use this
+      accumulatedPatches: null as any, // the strategy does not use this
+      commandDescriptions: null as any, // the strategy does not use this
+      sortedApplicableStrategies: null as any, // the strategy does not use this
+      startingMetadata: metadata ?? defaultMetadata,
+      startingAllElementProps: {},
+      customStrategyState: defaultCustomStrategyState(),
+    } as StrategyState,
+    'end-interaction',
+  )
+
+  expect(strategyResult.customStatePatch).toEqual({})
+  expect(strategyResult.status).toEqual('success')
+
+  const finalEditor = foldAndApplyCommands(
+    editorState,
+    editorState,
+    [],
+    [],
+    strategyResult.commands,
+    'end-interaction',
+  ).editorState
+
+  return finalEditor
+}
+
+const makeParentView = (uid: string, children: (() => string)[]) => {
+  return `
+    <View data-uid='${uid}'>
+      ${children.map((c) => c()).join('\n')}
+    </View>
+  `
+}
+
+const makeView = (uid: string, style: React.CSSProperties) => () => {
+  const defaultStyle: React.CSSProperties = {
+    backgroundColor: '#f0f',
+    width: 200,
+    height: 200,
+  }
+  return `
+    <View
+      data-uid='${uid}'
+      style={${JSON.stringify({
+        ...defaultStyle,
+        ...style,
+      }).replace(/"/g, "'")}}
+    />
+  `
+}
+
+describe('Relative move', () => {
+  describe('when the element position is relative', () => {
+    it('does not trigger when the threshold is not met', async () => {
+      const targetElement = elementPath([
+        ['scene-foo', 'app-entity'],
+        ['foo', 'bar'],
+      ])
+
+      const initialEditor = prepareEditorState(
+        makeParentView('foo', [
+          makeView('bar', {
+            position: 'relative',
+          }),
+        ]),
+        [targetElement],
+      )
+
+      const finalEditor = dragByPixels(initialEditor, canvasPoint({ x: 1, y: 1 }))
+      expect(testPrintCodeFromEditorState(finalEditor)).toEqual(
+        makeTestProjectCodeWithSnippet(
+          makeParentView('foo', [
+            makeView('bar', {
+              position: 'relative',
+            }),
+          ]),
+        ),
+      )
+    })
+
+    describe('when the element has no offsets', () => {
+      it('sets the TL offsets', async () => {
+        const targetElement = elementPath([
+          ['scene-foo', 'app-entity'],
+          ['foo', 'bar'],
+        ])
+
+        const initialEditor = prepareEditorState(
+          makeParentView('foo', [
+            makeView('bar', {
+              position: 'relative',
+            }),
+          ]),
+          [targetElement],
+        )
+
+        const finalEditor = dragByPixels(initialEditor, canvasPoint({ x: 15, y: 15 }))
+        expect(testPrintCodeFromEditorState(finalEditor)).toEqual(
+          makeTestProjectCodeWithSnippet(
+            makeParentView('foo', [
+              makeView('bar', {
+                position: 'relative',
+                left: 15,
+                top: 15,
+              }),
+            ]),
+          ),
+        )
+      })
+    })
+
+    describe('when the element has offsets', () => {
+      it('updates the offsets', async () => {
+        const targetElement = elementPath([
+          ['scene-foo', 'app-entity'],
+          ['foo', 'bar'],
+        ])
+
+        const initialEditor = prepareEditorState(
+          makeParentView('foo', [
+            makeView('bar', {
+              position: 'relative',
+              left: 10,
+              top: 10,
+            }),
+          ]),
+          [targetElement],
+        )
+
+        const finalEditor = dragByPixels(initialEditor, canvasPoint({ x: 15, y: 15 }))
+        expect(testPrintCodeFromEditorState(finalEditor)).toEqual(
+          makeTestProjectCodeWithSnippet(
+            makeParentView('foo', [
+              makeView('bar', {
+                position: 'relative',
+                left: 25,
+                top: 25,
+              }),
+            ]),
+          ),
+        )
+      })
+
+      describe('when vertical or horizontal offsets are missing', () => {
+        it('sets the missing offset', async () => {
+          const targetElement = elementPath([
+            ['scene-foo', 'app-entity'],
+            ['foo', 'bar'],
+          ])
+
+          const initialEditor = prepareEditorState(
+            makeParentView('foo', [
+              makeView('bar', {
+                position: 'relative',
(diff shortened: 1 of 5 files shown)


editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+    // should we also support absolute elements, for which we'll do something like `CONVERT_TO_RELATIVE`?
+    return meta.specialSizeMeasurements.position === 'relative'
+  },
+
+  controlsToRender: [],
+
+  fitness: (canvasState, interactionState, _sessionState) => {
+    const { interactionData, activeControl } = interactionState
+    if (!(interactionData.type === 'DRAG' && interactionData.drag !== null)) {
```
gbalint, line 43: For null checks we prefer the `!=` operator to catch `undefined` too (except if you have special reason to only check for null, which you should not have :)
For all other values we use  `!==`

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
```diff
@@ -0,0 +1,191 @@
...
+  isApplicable: (canvasState, _interactionState, metadata) => {
+    const selectedElements = getTargetPathsFromInteractionTarget(canvasState.interactionTarget)
+    if (selectedElements.length === 0) {
+      return false
+    }
+
+    const filteredSelectedElements = getDragTargets(selectedElements)
+    const last = filteredSelectedElements[filteredSelectedElements.length - 1]
+    const meta = MetadataUtils.findElementByElementPath(metadata, last)
+    if (!meta) {
```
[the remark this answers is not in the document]
ruggi: https://github.com/concrete-utopia/utopia/pull/2604/commits/8fd13803c78258f257a885f773868dbac98239cd

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/2604/commits/ac2a475cfd0a477db8d7720dad35be1d83e47cfa

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
```diff
@@ -0,0 +1,188 @@
...
+      return false
+    }
+
+    // should we also support absolute elements, for which we'll do something like `CONVERT_TO_RELATIVE`?
```
Rheeseyb, line 35: Good question, but I think the answer is probably no here... To convert to relative we'd have to perform some calculations to figure out the correct offset given that the element would now be positioned by flow. That _sounds_ like either a separate strategy, or something that relies on the meta-strategies work.

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
```diff
@@ -0,0 +1,188 @@
...
+    return meta.specialSizeMeasurements.position === 'relative'
+  },
+
+  controlsToRender: [],
```
Rheeseyb, line 39:
```suggestion
  controlsToRender: [
    {
      control: ParentOutlines,
      key: 'parent-outlines-control',
      show: 'visible-only-while-active',
    },
    {
      control: ParentBounds,
      key: 'parent-bounds-control',
      show: 'visible-only-while-active',
    },
  ],
```
That should cover it. The bounding box comes from `select-mode-hooks.tsx`, so all this will need is the above parent controls.

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
```diff
@@ -0,0 +1,188 @@
...
+      return 0
+    }
+
+    const selectedElements = getTargetPathsFromInteractionTarget(canvasState.interactionTarget)
+    if (selectedElements.length === 0) {
+      return 0
+    }
+    const filteredSelectedElements = getDragTargets(selectedElements)
+    const last = filteredSelectedElements[filteredSelectedElements.length - 1]
+    const meta = MetadataUtils.findElementByElementPath(interactionState.latestMetadata, last)
```
Rheeseyb, line 56: Minor, but can we stick with `metadata` for this please, as we use that (rather than `meta`) everywhere

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
```diff
@@ -0,0 +1,188 @@
...
+    if (!offsets) {
+      return 0
+    }
+
+    const hasOffsets =
+      offsets.left != null || offsets.top != null || offsets.right != null || offsets.bottom != null
+
+    return hasOffsets
+      ? 4 // +1 than reorder flow
+      : 1 // there should be a more structured way to define priorities (:
```
Rheeseyb, line 70: Yes, there should :)

editor/src/components/canvas/canvas-strategies/relative-move-strategy.spec.browser2.tsx
```diff
@@ -0,0 +1,393 @@
...
+  codeSnippet: string,
+  selectedViews: Array<ElementPath>,
+): EditorState => {
+  return {
+    ...getEditorState(makeTestProjectCodeWithSnippet(codeSnippet)),
+    selectedViews: selectedViews,
+  }
+}
+
+const defaultMetadata: ElementInstanceMetadataMap = {
```
balazsbajorics, line 40: I think this was copied from a jest test (.spec.tsx) and not a karma test (.spec.browser2.tsx)
the main difference between jest tests and karma tests is that in Karma tests we don't need to fake the metadata. 

take for example https://github.com/concrete-utopia/utopia/blob/d25efb0553ad480604732d753e98087cfd87e9c9/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.browser2.tsx#L248-L287 where we use `await renderTestEditorWithCode` which will return with a full metadata. 

we started writing the strategies with these fake metadata tests, but we realised mid-summer (when we were refactoring some strategies) that we are much better off if we write karma tests that run the full editor, they are much more refactor-proof and should be less finicky to write too.

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
ruggi: sure, thanks for pointing that out - it makes it a lot easier for me to pick up the small nit conventions 🙏

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
ruggi: thanks for letting me know about this, will do in the future as well 👍

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
```diff
@@ -0,0 +1,188 @@
...
+      return emptyStrategyApplicationResult
+    }
+
+    const { drag } = interactionData
+
+    const commands: Array<CanvasCommand> = []
+
+    // should set a direction only if it's either defined in the original offsets or if the counterpart (horiz/vert) is not defined
+    // vertical
+    commands.push(
```
balazsbajorics, line 101: I think this should use/share code with `applyAbsoluteMoveCommon` (and probably the Absolute should be removed from the name of the shared function :)) because that function also provides some common drag-to-move behavior such as snapping, holding shift locks the move to one axis, and holding cmd disables snapping.

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
```diff
@@ -0,0 +1,188 @@
...
+    attrs: PropsOrJSXAttributes,
+  ): number | null => {
+    const prop = getLayoutProperty(name, attrs, ['style'])
+    if (!isRight(prop)) {
+      return null
+    }
+    if (!prop.value) {
+      return null
+    }
+    return prop.value.value
```
Rheeseyb, line 169: You might benefit from using [`foldEither`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/either.ts#L41-L51) here (or some of the other functions in that file)

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
Rheeseyb: Yeah this is a good point. We don't want to lose / overlook that common functionality here.

editor/src/components/canvas/canvas-strategies/relative-move-strategy.spec.browser2.tsx
balazsbajorics: the flipside (or the do not fix if it ain't broke approach) is that we can consider converting this to a proper Jest test (ie removing .browser2 from the file name), and leave it as it is. – but I suspect that we will see the shortcomings of the fake metadata once you want to test snapping for example

editor/src/components/canvas/canvas-strategies/relative-move-strategy.spec.browser2.tsx
```diff
@@ -0,0 +1,393 @@
...
+    specialSizeMeasurements: {
+      immediateParentBounds: canvasRectangle({ x: 0, y: 0, width: 400, height: 400 }),
+      coordinateSystemBounds: canvasRectangle({ x: 0, y: 0, width: 400, height: 400 }),
+    } as SpecialSizeMeasurements,
+    globalFrame: canvasRectangle({ x: 50, y: 50, width: 250, height: 300 }),
+    localFrame: localRectangle({ x: 50, y: 50, width: 250, height: 300 }),
+  } as ElementInstanceMetadata,
+}
+
+function dragByPixels(
```
Rheeseyb, line 64: I think rather than using this pattern which creates and updates the `InteractionSession`s directly, we would be better off to use the pattern where we fire mouse events. e.g. we in the absolute move tests we fire one to select the element, and then another to drag it here: https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/canvas-strategies/absolute-move-strategy.spec.browser2.tsx#L50-L58

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
ruggi: something like this? https://github.com/concrete-utopia/utopia/pull/2604/commits/1c4e6b29855e726829728377778956dd55d19261

balazsbajorics: yep, perfect!

editor/src/components/canvas/canvas-strategies/relative-move-strategy.spec.browser2.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/2604/commits/a22bbd3f6e432073143d829bd39db8fd64a93f96 <- should this be acceptable? if so it should also address @balazsbajorics comment above

editor/src/components/canvas/canvas-strategies/relative-move-strategy.spec.browser2.tsx
```diff
@@ -0,0 +1,457 @@
...
+  modifiers?: { shiftPressed?: boolean },
+): Promise<EditorRenderResult> => {
+  const targetElement = renderResult.renderedDOM.getByTestId(elementId)
+  const targetElementBounds = targetElement.getBoundingClientRect()
+  const canvasControlsLayer = renderResult.renderedDOM.getByTestId(CanvasControlsContainerID)
+
+  const startPoint = windowPoint({ x: targetElementBounds.x + 5, y: targetElementBounds.y + 5 })
+  const dragDelta = windowPoint({ x, y })
+
+  act(() => dragElement(canvasControlsLayer, startPoint, dragDelta, !!modifiers?.shiftPressed))
```
enidemi, line 81: I think here you can use the helper `mouseDragFromPointWithDelta` or `mouseDragFromPointToPoint`

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
```diff
@@ -0,0 +1,144 @@
...
+    attrs: PropsOrJSXAttributes,
+  ): number | null => {
+    return foldEither(
+      (_) => null,
+      (v) => (v != null ? v.value : null),
+      getLayoutProperty(name, attrs, ['style']),
+    )
+  }
+
+  if (!isRight(meta.element)) {
```
seanparsons, line 128:
```suggestion
  if (isLeft(meta.element)) {
```

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/relative-move-strategy.spec.browser2.tsx
ruggi: oh nice, I did not spot those

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
ruggi: you're absolutely _right_ (ba-dum-tss) 🏅

editor/src/components/canvas/canvas-strategies/relative-move-strategy.spec.browser2.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/2604/commits/c71735577000ef8b3bb16d46cac415eadafec617

enidemi (approved):

editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
```diff
@@ -30,22 +23,130 @@ import {
...
+  ): {
+    commands: Array<AdjustCssLengthProperty>
+    intendedBounds: Array<CanvasFrameAndTarget>
+  } => {
+    const selectedElements = getTargetPathsFromInteractionTarget(canvasState.interactionTarget)
+    const filteredSelectedElements = getDragTargets(selectedElements)
+    let commands: Array<AdjustCssLengthProperty> = []
+    let intendedBounds: Array<CanvasFrameAndTarget> = []
+    filteredSelectedElements.forEach((selectedElement) => {
+      const elementResult = getAbsoluteMoveCommandsForSelectedElement(
```
Rheeseyb, line 74: This function name is making me raise my eyebrow given that `getAdjustMoveCommands` is being used by the relative positioning strategy. I took a quick look at the function itself and it looks like it is still fine to use for relative positioning, in which case please can we rename it. Also, these functions should be moved to a more generically named file since they're now being used by non-absolute strategies.

ruggi: yeah good points, will do

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
```diff
@@ -0,0 +1,144 @@
...
+    const filteredSelectedElements = getDragTargets(selectedElements)
+    const last = filteredSelectedElements[filteredSelectedElements.length - 1]
+    const metadata = MetadataUtils.findElementByElementPath(sessionState.startingMetadata, last)
+    if (metadata == null) {
+      return emptyStrategyApplicationResult
+    }
+    const offsets = getStyleOffsets(metadata)
+    if (!offsets) {
+      return emptyStrategyApplicationResult
+    }
```
Rheeseyb, line 105: It seems that with the use of `applyMoveCommon` this code is now effectively just duplicating the `isApplicable` or `fitness` functions, in which case we should probably just directly reuse those at the start of this function (we're already following that pattern in a bunch of other strategies)

editor/src/components/canvas/canvas-strategies/relative-move-strategy.spec.browser2.tsx
```diff
@@ -0,0 +1,401 @@
...
+  modifiers?: Modifiers,
+): Promise<EditorRenderResult> => {
+  const renderResult = await renderTestEditorWithCode(project, 'await-first-dom-report')
+
+  const targetElement = renderResult.renderedDOM.getByTestId(elementId)
+  const targetElementBounds = targetElement.getBoundingClientRect()
+  const canvasControlsLayer = renderResult.renderedDOM.getByTestId(CanvasControlsContainerID)
+
+  const startPoint = windowPoint({ x: targetElementBounds.x + 5, y: targetElementBounds.y + 5 })
+  const dragDelta = windowPoint({ x, y })
```
Rheeseyb, line 26: Heads up here - the use of `windowPoint` for these tests is a remnant of some older code that happened to require those types because of the utility functions we were using for them. The newer helper functions in `event-helpers.test-utils.ts` don't require `windowPoint`

editor/src/components/canvas/canvas-strategies/shared-absolute-move-strategy-helpers.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/2604/commits/b355cdb0a31eadada31eda5ae4723ce613f2aa00 note: aside from renaming the mentioned function, I just renamed the whole helpers file to `shared-move-strategies-helpers` as imho it makes sense to not split it into two (e.g. one for the relative and one for the absolute movements) if one just includes both

editor/src/components/canvas/canvas-strategies/relative-move-strategy.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/2604#:~:text=98067a4-,reuse%20fitting%20logic%20when%20applying,-b635766

Rheeseyb (approved):

PR #2605, opened 2022-10-03
JsxMetadata dispatch fixes

balazsbajorics:
**Context:**
I am currently making a pull request that moves `startingMetadata` and `startingAllElementProps` from `StrategyState` to `InteractionCanvasState`. In order for _that_ PR to work correctly, there's two fixes to how we manage metadata that I wanted to put in their dedicated PR to reduce noise.

**Problem:**
The PR fixes two problems:
1. CLEAR_INTERACTION_SESSION clears `unpatchedEditorState.jsxMetadata` and `unpatchedEditorState.domMetadata` before the strategies are executed which is bad for us because then we can't rely on unpatchedEditorState.jsxMetadata as our starting metadata. 
2. While `unpatchedEditorState.jsxMetadata` is the _starting_ metadata during the course of an interaction session, `allElementProps` is always fresh. The two should behave the same way.

**Fix:**
1. Do not clear jsxMetadata/domMetadata in CLEAR_INTERACTION_SESSION, instead only clear it in `interactionCancel`/`interactionFinished`. For the rest of the editor, this change should be transparent, but for strategies it means we can rely on unpatchedEditorState.jsxMetadata as the starting metadata.
2. added _currentAllElementProps_KILLME to store the latest allElementProps, and only update allElementProps when we update jsxMetadata


bkrmendy (approved):

gbalint (approved):

enidemi (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -734,7 +734,7 @@ function switchAndUpdateFrames(
     case 'flex':
       withUpdatedLayoutSystem = {
         ...withUpdatedLayoutSystem,
-        allElementProps: MetadataUtils.setPropertyDirectlyIntoMetadata(
+        _currentAllElementProps_KILLME: MetadataUtils.setPropertyDirectlyIntoMetadata(
```
Rheeseyb, line 737: If you don't kill this in the next few days I will give you my extreme disappointment face

Rheeseyb (commented):
👍 

If you ever desecrate the underscore prefix like this again I'm going to head straight over to your house and deliver to you a _very_ strongly worded letter

PR #2606, opened 2022-10-03
tests(performance) - eliminate small deltas from automated PR annotations

maltenuhn:
**Problem:**
Our perf tests show all results, and we suppress details if the absolute percentage difference is less than 5. This generates a spammy level of lines. We already show all results in the perf graph anyway (though maybe even that is overkill)

**Fix:**
Only show lines where abs delta is > 5%

_NB: rage fix. Also, Gresham's Law stimulus for a / a better solution_


seanparsons:
Does this completely eliminate all of the graphical chart generation?

seanparsons (approved):

PR #2608, opened 2022-10-03
Feature/flow reorder likely works

enidemi:
**Fix:**
Only allow flex-like flow reorder in cases where the elements are in one type of layout like all blocks or inline elements without wrapping to multiple columns/rows, possibly all siblings are placed in one visible column or row. For other flow layouts the fallback slider can be used.

**Commit Details:**
- added a new flow reorder helper to `isApplicable`
- changed tests as automatic conversion happens only with the slider control now


editor/src/components/canvas/canvas-strategies/flow-reorder-helpers.ts
```diff
@@ -33,6 +29,55 @@ export function isValidFlowReorderTarget(
...
+
+function areNonWrappingSiblings(
+  frames: Array<CanvasRectangle>,
+  direction: 'horizontal' | 'vertical',
+): boolean {
+  // wrapping or columns set: a sibling frame has a larger x/y then the right/bottom edge of its siblings
+  return frames.every((frame) => {
+    return frames.every((frameToCompare) => {
+      if (direction === 'horizontal') {
+        return frameToCompare.y < frame.y + frame.height
```
balazsbajorics, line 68: I'm not sure but isn't horizontal and vertical swapped here? shouldn't horizontal be comparing the x coordinate?

balazsbajorics (commented):
please add test cases with examples for allowed and not-allowed layouts. I suspect we will tweak the allowlist until we find the right set of heuristics, and if this code is untested, we will just make regressions every time we change it.

editor/src/components/canvas/canvas-strategies/flow-reorder-helpers.ts
enidemi: the horizontal direction is for the layout type, I fixed the variable name here. What it checks here is if the elements are in an inline layout with wrapping texts some siblings will have their y below the siblings bottom edge

enidemi: > hmmmm I'm not sure I like that check, because it would break for this case: image
> 
> I think it's more reliable and less strict to verify that in a left-to-right row layout, no element should be to the left to their preceding sibling. so if element n wraps, that will jump to the left edge of the layout, meaning it will be to the left of element n-1. image

the second version is nicer and less strict, but for the 1st image how do you know where to drag if we use the starting frames?

balazsbajorics: > the second version is nicer and less strict, but for the 1st image how do you know where to drag if we use the starting frames?

if we know that we are thinking in terms of a 1-dimensional restricted drag, we could compare just the X or just the Y coordinates. so in this case, the entire area divided by red could trigger the reorder, not just the element itself:
image


or a bit more strictly like this:
image

editor/src/components/canvas/canvas-strategies/flow-reorder-helpers.ts
```diff
@@ -33,6 +29,66 @@ export function isValidFlowReorderTarget(
...
+  layoutDirection: 'horizontal' | 'vertical',
+): boolean {
+  if (layoutDirection === 'horizontal') {
+    // all elements are on the right side of the previous sibling
+    return frames.every((frame, i) => {
+      if (i === 0) {
+        return true
+      } else {
+        const prevFrame: CanvasRectangle = frames[i - 1]
+        return frame.x > prevFrame.x
```
balazsbajorics, line 71: if you put the if (layoutDirection === 'horizontal') inside here that saves you from having to write the loop code twice

balazsbajorics (approved):
thanks for the test cases!

editor/src/components/canvas/canvas-strategies/flow-reorder-helpers.ts
enidemi: fixed

seanparsons (approved):

PR #2609, opened 2022-10-03
Strategy Interface Cleanup

balazsbajorics:
**Problem:**
startingMetadata and startingAllElementProps should not live in StrategyState. The full `StrategyState` object shouldn't even be exposed to the strategies, it makes the mental model harder (it acts as a sort of "loopback" but no strategy is actually introspecting the values anyways). The strategies are only interested in the CustomStrategyState.

**Fix:**
Since unpatchedEditor.jsxMetadata and unpatchedEditor.allElementProps are the starting metadata and starting props anyways, the most logical place is to move the startingMetadata is the `InteractionCanvasState`. 

**Commit Details:**
- startingMetadata and startingAllElementProps now come from canvasState in all strategy APIs
- strategyState has been narrowed down and we only pass down customStrategyState to the strategy code
- tests have been updated, most of the diffs are a removal of fake strategyState
- unfortunately patchedEditor.jsxMetadata is the latest, so to let canvas controls access the startingMetadata, I kept `strategyState.startingMetadata` alive. it is not exposed to the strategies though, but there's definitely room for improvement there, but I think that work should be combined with a cleanup of the reconstructMetadata stuff and where we are stashing the latestMetadata in the editor

**More PRs incoming**
This is a PR in a string of PRs that are cleaning up the strategy API. Next up: the isApplicable function shouldn't receive an interactionSession, StrategyState should probably be renamed to StrategyResult to signal that it is downstream (derived) of EditorState+InteractionSession. I _think_ there's still some leftover stuff around `commandsToAccumulate` that can now go in the bin (I hope).


editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -92,6 +93,26 @@ export function pickCanvasStateFromEditorState(
     openFile: editorState.canvas.openFile?.filename,
     scale: editorState.canvas.scale,
     canvasOffset: editorState.canvas.roundedCanvasOffset,
+    startingMetadata: editorState.jsxMetadata,
+    startingAllElementProps: editorState.allElementProps,
```
balazsbajorics, line 97: this here is a very important assumption of this PR.

enidemi (approved):

bkrmendy (approved):

gbalint:
Minor issues:

I think the renaming is incomplete, there are several strategies where the apply function's third parameter is still called `strategyState`, e.g. `flex-reparent-to-flex-strategy`, `flow-reorder-strategy`, etc

I think this would be a good time to rename `interactionSession` parameter to `interactionState` in those few strategies where it is still used, e.g. `flex-reparent-to-flex-strategy`

gbalint:
Also in the fitness function we should try to use consistent parameter naming, e.g. in `convert-to-absolute-move-strategy` we still call the `customStrategyState` parameter `sessionState`
In most of the strategies it is called `strategyState` which is also kinda misleading, because its type is not `StrategyState`

balazsbajorics:
good point @gbalint , I'll roll in the variable name fixes

Rheeseyb (approved):

gbalint (approved):

PR #2610, opened 2022-10-03
Load project contents from a Github branch.

seanparsons:
**Problem:**
It's possible for us to save to Github, but not currently to retrieve those projects back from Github.

**Fix:**
This implements a flow of listing branches for a given repository, as well as the functionality to load the content from a given branch.

It was found while implementing this that Github use a MIME Base64 encoding, rather than the more regular style of Base64 encoding

When loading the content, in a slight reversal from what is possible when creating a git tree, the tree has to be recursively loaded for each directory in the tree (which is itself another tree). As well as individual files needing to be retrieved by looking up the value of each blob.

**Commit Details:**
- Added `fullPathFromProjectContentsTree` utility function.
- Added `UPDATE_PROJECT_CONTENTS` editor action which replaces the entirety of the project contents in `EditorState`.
- Added server endpoint to retrieve the branches for a given repository
- Added server endpoint that pulls the content for a given branch.
- Moved the types out of `Github.hs` as it was starting to get swamped by them.
- Generalised `callGithub` with a `MakeGithubRequest` parameter which allows us to make `GET` requests without a request body.
- Improved the error handling in `callGithub`, also disabling the default `checkResponse` so that any status code response will return.
- Added service types for getting the branch listing and for retrieving the content of a branch.
- `getRecursiveGitTreeAsContent` fronts the functionality of handling of blobs or being itself invoked for a tree within the tree.

### DIFF
--- a/editor/src/components/navigator/left-pane.tsx
+++ b/editor/src/components/navigator/left-pane.tsx
@@ -813,7 +813,7 @@ const GithubPane = React.memo(() => {
 
   React.useEffect(() => {
     if (parsedTargetRepository != null) {
-      getBranchesForGithubRepository(parsedTargetRepository).then((result) => {
+      void getBranchesForGithubRepository(parsedTargetRepository).then((result) => {
         setBranchesForRepository(result)
       })
     }
@@ -838,7 +838,7 @@ const GithubPane = React.memo(() => {
               {branchesForRepository.branches.map((branch, index) => {
                 function loadContentForBranch() {
                   if (parsedTargetRepository != null) {
-                    getBranchContent(dispatch, parsedTargetRepository, branch.name)
+                    void getBranchContent(dispatch, parsedTargetRepository, branch.name)
                   }
                 }
                 return (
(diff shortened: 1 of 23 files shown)


editor/src/components/navigator/left-pane.tsx
```diff
@@ -777,22 +784,81 @@ const GithubPane = React.memo(() => {
...
+
+  React.useEffect(() => {
+    if (parsedTargetRepository != null) {
+      void getBranchesForGithubRepository(parsedTargetRepository).then((result) => {
+        setBranchesForRepository(result)
+      })
+    }
+  }, [parsedTargetRepository])
+
+  const getBranchesUI = React.useCallback(() => {
```
Rheeseyb, line 822: Looks like `React.useMemo` would be better suited to this

Rheeseyb (approved):

Rheeseyb (approved):

editor/src/components/navigator/left-pane.tsx
seanparsons: Fixed in 8fc83dd4a.

PR #2620, opened 2022-10-05
Don't set project URL during browser test debug

ruggi:
## Problem

When running browser tests in debug mode, the editor sets the URL for the project in the browser window; this means that repeated tests won't run automatically, forcing us to manually reset the URL to the root.

## Fix

This PR tries the most conservative approach: if we're in the test environment _and_ an explicit new env var is set to true (`REACT_APP_BROWSER_TEST_DEBUG`), the editor won't push the project URL to the history, so that tests can re-run automatically when a file changes (typical use case: a browser window with Karma tests running open to the side of the IDE where we're writing the tests).

This could be done in a more or less granular way, but this should be flexible enough so that i.e. we can explicitly disable it in certain cases, etc. For example, I could have made the `IS_BROWSER_TEST_DEBUG` a function returning a boolean instead of a constant value (so i.e. a spec can disable it), but I don't want to do preemptive engineering if we're not benefiting from it right now :)


balazsbajorics (approved):

editor/src/components/editor/editor-component.tsx
```diff
@@ -37,7 +37,11 @@ import { isLiveMode, dragAndDropInsertionSubject, EditorModes, isSelectMode } fr
 import { Toast } from '../common/notices'
 import { chrome as isChrome } from 'platform-detect'
 import { applyShortcutConfigurationToDefaults } from './shortcut-definitions'
-import { PROPERTY_CONTROLS_INFO_BASE_URL } from '../../common/env-vars'
+import {
+  IS_BROWSER_TEST_DEBUG,
+  IS_TEST_ENVIRONMENT,
```
Rheeseyb, line 42:
```suggestion
```
This import wasn't used in the end it seems

editor/src/components/editor/editor-component.tsx
```diff
@@ -282,6 +286,9 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
   }, [projectName])
 
   React.useEffect(() => {
+    if (IS_BROWSER_TEST_DEBUG) {
+      return
+    }
```
Rheeseyb, line 291: Perfect!

Rheeseyb (approved):

enidemi (approved):

gbalint (approved):

Rheeseyb (approved):

PR #2623, opened 2022-10-05
chore: introducing ControlWithProps

balazsbajorics:
**Problem:**
The strategies' canvas controls/indicators cannot have props. We worked around this limtation for a while by using `EditorState.canvas.controls` as a place to put information for these controls via Commands. But with the "find valid ancestor strategy" strategy, we hit a wall: all current controls operate assuming `EditorState.selectedViews` are going to be the target of their strategy counterparts. This means that if I have element `A/B/C` selected, but the found relevant strategy is targeting A/B, the controls will still believe the target is A/B/C. 

**Fix:**
The controls should be able to have an explicit target property. 
To achieve _that_, the controls should be able to have props in the first place.

**Details:**
- `CanvasStrategy.controlsToRender` array now has a props field for each control.
- The controls receive these props when they are rendered
- To make sure we can look up all the react components that are used by strategies as controls, I added a type `ControlForStrategy`
- To be able to easily mark a component as a "control for a strategy", I added a helper function `controlForStrategyMemoized`. This is the same as React.memo but also wraps the control in a typed object.
- There should be NO FUNCTIONAL CHANGES introduced by this PR, at this point all of the props fields are an empty {}. I previously had a branch that  updates the controls' use of EditorState.selectedViews to a targets prop, but it was incredibly noisy. I will do that in a follow-up pull request.


editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -364,16 +365,17 @@ export const useDelayedCurrentStrategy = () => {
...
   key: 'not-resizable-control',
   show: 'visible-except-when-other-strategy-is-active',
-}
+})
 
 export function getApplicableControls(
   currentStrategy: CanvasStrategyId | null,
   strategy: CanvasStrategy,
-): Array<ControlWithKey> {
+): Array<ControlWithProps<any>> {
```
Rheeseyb, line 378: Please can we use `unknown` rather than `any` where possible

editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
```diff
@@ -49,13 +50,29 @@ export function strategyApplicationResult(
...
 
-export interface ControlWithKey {
-  control: React.FC<React.PropsWithChildren<unknown>>
+export interface ControlForStrategy<P> {
+  type: 'ControlForStrategy'
+  control: React.FC<P>
+}
+
+export function controlForStrategyMemoized<P>(control: React.FC<P>): ControlForStrategy<P> {
+  return { type: 'ControlForStrategy', control: React.memo<any>(control) }
```
Rheeseyb, line 59: I'm pretty sure we can use `React.memo<P>` here

balazsbajorics: image

I was fucking around with it for like an hour before I gave up. This <any> doesn't leak outside of the function, so this is not affecting anything in practice. It looks silly but it isn't worth more effort trying to type this up

Rheeseyb: Oh ffs

Rheeseyb (approved):

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
balazsbajorics: let me see!

editor/src/components/canvas/canvas-strategies/canvas-strategy-types.ts
```diff
@@ -49,13 +50,29 @@ export function strategyApplicationResult(
...
+}
+
+export type WhenToShowControl =
+  | 'always-visible'
+  | 'visible-only-while-active'
+  | 'visible-except-when-other-strategy-is-active'
+
+export interface ControlWithProps<P> {
+  control: ControlForStrategy<P>
+  props: P
```
ruggi, line 69: maybe a silly question, but: given the usage (always `{}`), why aren't we making this optional and defaulting it to `{}` when using it in the memoized wrapper?
```ts
export interface ControlWithProps<P> {
  control: ControlForStrategy<P>
  props?: P
  ...
}
```

balazsbajorics: that's actually a good question! I erred on the side that _for now_ these are {} but hopefully soon it will be filled up. 
but it would certainly make for a smaller diff if these would be optional...

ruggi: imho it would also make things less confusing: if I see code everywhere where `{}` is used, I have to wonder why that's the case 🤔 but of course maybe that's just me!

balazsbajorics: oh you gave me a Typescript riddle!

if P is {} we want props to be optional. but if P is actually a non-empty props interface, we want to enforce props

ruggi: 🤯

balazsbajorics: I cannot come up with a good solution. But I promise that a follow-up PR is coming with non-empty props objects!

PR #2631, opened 2022-10-06
Remove Text component from utopia-api

gbalint:
**Problem:**
We no longer use Text from utopia-api, it doesn't really work anymore (no insertion, no editing)

**Fix:**
Remove it!
Note: I did not remove Text from parser-printer and ts-worker tests, which contain lot of test code with `Text` imported. However, importing Text is still syntactically valid JS code, it doesn't cause any problems, so we can just keep the tests intact.


enidemi (commented):
There is a shortcut 'T' in the editor that inserts a Text element from utopia-api

seanparsons (approved):

enidemi (approved):

PR #2633, opened 2022-10-07
Fix/canvas insertion image aspect ratio

enidemi:
**Fix:**
Inserting an image with draw-to-insert should keep the aspect ratio. I'm using a custom strategy state variable to store aspect ratio from `insertionSubject.defaultSize` that can be used by resize strategies.

Only for absolute resize now. Aspect ratio lock is missing from flex resize strategy.

**Commit Details:**
- update absolute-resize strategy
- tests
- removed commit hash from test runner env vars otherwise the test results can change with commits


ruggi (approved):

editor/src/components/canvas/canvas-strategies/draw-to-insert-strategy.spec.browser2.tsx
```diff
@@ -1814,3 +1817,121 @@ describe('Inserting into flex column', () => {
...
+          >
+            <img              
+              style={{
+                width: 300,
+                height: 300,
+                position: 'absolute',
+                left: 5,
+                top: 5,
+              }}
+              src='/editor/icons/favicons/favicon-128.png?hash=nocommit'
```
enidemi, line 1918: this is where the commit hash breaks the tests

balazsbajorics: thanks!

editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -258,6 +263,7 @@ export function updateInteractionViaDragDelta(
       userPreferredStrategy: currentState.userPreferredStrategy,
       startedAt: currentState.startedAt,
       updatedTargetPaths: currentState.updatedTargetPaths,
+      aspectRatioLock: null,
```
balazsbajorics, line 266: I think this shouldn't reset the aspectRatioLock, and just use currentState.aspectRatio lock. same with the other `update...` functions

enidemi: oops, fixed :)

gbalint (approved):

PR #2639, opened 2022-10-10
Refactor: simplify insertion subjects

gbalint:
With the new insert mode `InsertionSubjects` types can be simplified:

- `DragAndDropInsertionSubject` is not used anymore
- `ElementInsertionSubjects` are just an array of `ElementInsertionSubject` 
- Because only `ElementInsertionSubject` remained in `InsertionSubject` we don't need a union type anymore

Note:
I did not refactor the insert case in `produceCanvasTransientState`, because insert mode without the Canvas strategies feature switch is fully broken anyway.

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -1874,57 +1874,6 @@ export function produceCanvasTransientState(
     const editorMode = editorState.mode
     switch (editorMode.type) {
       case 'insert':
-        if (
-          insertionSubjectIsJSXElement(editorMode.subject) &&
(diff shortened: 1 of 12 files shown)


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1874,57 +1873,6 @@ export function produceCanvasTransientState(
     const editorMode = editorState.mode
     switch (editorMode.type) {
       case 'insert':
-        if (
```
balazsbajorics, line 1877: I think it's probably worth throwing a big fat error here, just in case anyone tries to use this

bkrmendy (approved):

PR #2636, opened 2022-10-10
Make flex resize respect aspect ratio

ruggi:
## Problem

The basic flex strategy does not respect aspect ratio when `shift` is pressed.

## Fix

Make the strategy handle `shift` locking the aspect ratio of the resized element.

**Notes:**
- Exported the `getLockedAspectRatio` function and moved it to the helpers file
- Renamed the helpers file to a generic `resize-helpers` instead of being tied to "absolute"

PR #2642, opened 2022-10-11
Flex resize use resizeBoundingBox

ruggi:
This PR is a followup to https://github.com/concrete-utopia/utopia/pull/2636

⚠️ This PR is only performing a refactor to use `resizeBoundingBox` in the flex resize strategy. The scope has changed, but I'm keeping the old description below for historical purposes.

---

**Problem:**

Images should be resized with a fixed aspect ratio.

**Fix:**

This PR forces the resizing strategy to keep the aspect ratio locked for images.

Notes:
- The same thing could have also been done by faking a `shift` modifier for a quick win when calling `getLockedAspectRatio`, but honestly that would be ugly as hell.
- As a bonus point, got rid of the `let` assignments when calculating the new dimensions ✨

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/flex-resize-basic-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/flex-resize-basic-strategy.tsx
@@ -205,11 +205,7 @@ const getSpecializedLockedAspectRatio = (
   rectangle: CanvasRectangle,
   metadata: ElementInstanceMetadata,
 ) => {
-  if (rectangle.width === 0 || rectangle.height === 0) {
-    // if dimensions are missing, there's no point in calculating the ratio
-    return null
-  }
-  if (MetadataUtils.isImg(metadata)) {
+  if (MetadataUtils.isImg(metadata) && rectangle.width !== 0 && rectangle.width !== 0) {
     return rectangle.width / rectangle.height
   }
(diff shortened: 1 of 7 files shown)


editor/src/components/canvas/canvas-strategies/strategies/flex-resize-basic-strategy.tsx
```diff
@@ -200,6 +199,18 @@ export function flexResizeBasicStrategy(
   }
 }
 
+const getSpecializedLockedAspectRatio = (
+  interactionSession: InteractionSession,
+  modifiers: Modifiers,
+  rectangle: CanvasRectangle,
+  metadata: ElementInstanceMetadata,
+) => {
+  if (MetadataUtils.isImg(metadata) && rectangle.width !== 0 && rectangle.width !== 0) {
```
Rheeseyb, line 208:
```suggestion
  if (MetadataUtils.isImg(metadata) && rectangle.width !== 0 && rectangle.height !== 0) {
```

Copy pasta error

Rheeseyb (approved):

gbalint (approved):

ruggi:
note: the tests are updated to reflect the changes introduced by using the logic of `resizeBoundingBox` instead of using width as the dominant measure

balazsbajorics (approved):

PR #2644, opened 2022-10-11
Add targets fields for strategy controls

### LINKED ISSUE #2655: Add an explicit `targets: Array<ElementPath>` prop to the Canvas Controls that use the currently selected views
All strategy controls / indicators use editorState.selectedViews to find targets and calculate position/size/color etc. But with the "look for applicable ancestor strategy" strategy, the target of the strategy may not be selectedViews anymore. Therefore we should be explicitly providing the targets to those controls.

balazsbajorics:
_Updated by Rheese on 13/10/2022_

Fixes #2655 

**Problem:**
All strategy controls / indicators use editorState.selectedViews to find targets and calculate position/size/color etc. But with the "look for applicable ancestor strategy" strategy, the target of the strategy may not be selectedViews anymore.

**Fix:**
Make `targets: Array<ElementPath>` an explicit prop on these controls.

This PR also adds some more explicit parent controls, which take `targetParent: ElementPath` in their props, as the previous (name renamed) ones were using the `immediateParentBounds` of the targets. This means that those new controls will always correctly reflect the target of the reparent (even in cases where we prevent the reparenting), and make them useable for insertion.

**Commit Details:**
- Added `targets: Array<ElementPath>` or `target: ElementPath` as prop on strategy controls / indicators that were using EditorState.selectedViews
- Removed all reference to EditorState.selectedViews in these controls / indicators
- Renamed `ParentOutlines` and `ParentBounds` controls to `ImmediateParentOutlines` and `ImmediateParentBounds` to indicate that they are using the `immediateParentBounds` of the target elements
- Added new `ParentOutlines` and `ParentBounds` controls which take a `targetParent: ElementPath` prop and are rendered based on that element's frame

**Note:**
This change increased the render count tests slightly. This is because the `RenderControlMemoized` component we use to render these controls only runs a shallow equality check on the props, which these props will now fail. I profiled whilst dragging an element to repeatedly trigger both flex and absolute reparent strategies on staging and on this deployed branch to see if there was any impact, and there was none, so I don't suggest we switch to a deeper equality check for this.


editor/src/components/canvas/canvas-strategies/strategies/drag-to-insert-strategy.tsx
```diff
@@ -57,20 +57,20 @@ export function dragToInsertStrategy(
...
+        control: ImmediateParentOutlines,
+        props: { targets: [] }, // <<<- TODO feed it with targetParents
         key: 'parent-outlines-control',
         show: 'visible-only-while-active',
       }),
       controlWithProps({
-        control: ParentBounds,
-        props: {},
+        control: ParentBoundsForInsertion,
+        props: { targetParents: [] }, // <<<- TODO feed it with real props
```
seanparsons, line 67: Are these TODOs for subsequent work?

Rheeseyb: Yeah, there is a point about this in the Insertion doc, as this is only possible once insertion becomes a meta strategy

editor/src/components/canvas/canvas-strategies/strategies/relative-move-strategy.tsx
```diff
@@ -30,7 +29,10 @@ export function relativeMoveStrategy(
   const filteredSelectedElements = getDragTargets(selectedElements)
   const last = filteredSelectedElements[filteredSelectedElements.length - 1]
   const metadata = MetadataUtils.findElementByElementPath(canvasState.startingMetadata, last)
-  if (metadata == null || metadata.specialSizeMeasurements.position !== 'relative') {
+  if (
+    metadata == null ||
+    metadata.specialSizeMeasurements.position !== 'relative' // I think this check should be filteredSelectedElements.every
```
seanparsons, line 34: Looks like it.

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/relative-move-strategy.tsx
Rheeseyb: I'm going to leave this for 2 reasons:
1) This comment was made by @balazsbajorics :D
2) I don't want to change any unrelated behaviour in this PR.

enidemi (approved):

PR #2645, opened 2022-10-11
Creating a reparenting metastrategy

Rheeseyb:
**Problem:**
Reparenting strategies shared a lot of duplicated code, which was called multiple times during each strategy's lifecycle (to determine if it was applicable, what its fitness was, and how to apply it), and which is also heavily relied upon by insertion. On top of this, since the insertion strategies are currently having to find the reparenting strategy with the highest fitness, it means there is currently no way to select different types of insertion (e.g. to insert an absolute positioned element into a flex container).

**Fix:**
The fix here is to introduce a reparenting meta strategy, which performs the task of finding which types of reparent are available for a given point on the canvas, and then generate all of the reparent strategies for that.

This is achieved via a new function in `reparent-strategy-helpers.ts` called `findPartialReparentStrategies`, which determines what available parents are at that point, what context you would be reparenting into (currently just "absolute" or "flex"), and whether or not that reparenting action would require `'allow-missing-bounds'`.

Because we now capture all of this information upfront in the metastrategy (we were previously already calling the same logic, but then just not capturing all of the results, meaning we'd be calling pieces of it repeatedly), it means we can determine the fitness of each of the possible strategies, as well as shortcut parts of their `apply` function (since we don't need to find the target parent again).

I've updated all of the tests, but note that the non-browser tests with faked metadata are just getting worse and worse, so we should take a pitstop in the very near future to turn those into browser tests.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
@@ -1,12 +1,21 @@
+import { isHorizontalPoint } from 'utopia-api/core'
+import { getLayoutProperty } from '../../../../core/layout/getLayoutProperty'
+import {
+  framePointForPinnedProp,
+  LayoutPinnedProp,
+} from '../../../../core/layout/layout-helpers-new'
+import {
+  flexDirectionToFlexForwardsOrBackwards,
+  flexDirectionToSimpleFlexDirection,
+  FlexForwardsOrBackwards,
+  SimpleFlexDirection,
+} from '../../../../core/layout/layout-utils'
 import { MetadataUtils } from '../../../../core/model/element-metadata-utils'
 import { getStoryboardElementPath } from '../../../../core/model/scene-utils'
-import { mapDropNulls, reverse, stripNulls, uniqBy } from '../../../../core/shared/array-utils'
+import { mapDropNulls, reverse, stripNulls } from '../../../../core/shared/array-utils'
+import { isRight, right } from '../../../../core/shared/either'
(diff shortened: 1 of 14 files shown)


editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
```diff
@@ -43,226 +60,119 @@ import {
...
+  cmdPressed: boolean,
+  pointOnCanvas: CanvasPoint,
+): Array<FindReparentStrategyResult> {
+  const metadata = canvasState.startingMetadata
+
+  const reparentSubjects =
+    canvasState.interactionTarget.type === 'INSERTION_SUBJECTS'
+      ? newReparentSubjects()
+      : existingReparentSubjects(
+          getDragTargets(getTargetPathsFromInteractionTarget(canvasState.interactionTarget)), // uhh
```
seanparsons, line 133: `// uhh` ?

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
balazsbajorics: it was meant to signify that this could is not production quality 🤣

balazsbajorics: I think it's better if reparentSubjects is a parameter of `findReparentStrategies`, because it kinda feels like the reparent and the insertion codes are leaking and intermixing in here

gbalint (approved):

balazsbajorics (approved):

PR #2662, opened 2022-10-13
Feature/draw to insert metastrategy

### LINKED ISSUE #2647: Turn draw to insert into a metastrategy.
Use the reparent metastrategy work as a jumping off point. 
The goal is to be able to allow the same options as for reparent (especially the forced-not forced options) in the strategy picker. 
To be able to expose multiple strategy options in the picker, we need multiple strategies in the picker.
While there's multiple ways to achieve this, I believe the cleanest and most future proof way if to build on the reparent metastrategy, and map those results into individual draw-to-insert strategies.

balazsbajorics:
Fixes #2647 

**Problem:**
Draw to insert (and drag-to-insert) only offer a single strategy, which selects the most fit reparent startegy to use during the insertion. This means during insertion the users cannot pick between reparenting to the nearest absolute parent vs reparenting to the nearest parent (ie `Reparent (Abs)` vs `Reparent (Abs, Forced)`).

**Solution**
Since #2645 we have a unified reparent metastartegy. By refactoring that metastrategy a bit, we get a reuseable inner function that can be called from the new draw-to-insert metastrategy.
The draw to insert metastrategy gets a list of applicable reparent strategies, and returns a draw to insert strategy for each applicable reparent method. 
The result is that we have a list of draw-to-insert options to chose from in the strategy picker, such as `Draw to Insert (Flex)`, `Draw to Insert (Abs)` and `Draw to Insert (Abs, Forced)`.

**Details**
- New helper file getApplicableReparentFactories
- New drawToInsertMetaStrategy
- drawToInsertMetaStrategy maps over the result of getApplicableReparentFactories and uses `drawToInsertStrategyFactory` to create a strategy that has a specific name, fitness and associated reparent strategy to use.
- runTargetStrategiesForFreshlyInsertedElementToReparent no longer uses `findCanvasStrategy`, but instead explicitly uses `reparentStrategyToUse`, making the code easier to debug and less abstract.

**TODO**
- Next up: Do the same for drag-to-insert, turn that into a metastrategy too
- minor: in the future when we offer multiple competing resize strategies, runTargetStrategiesForFreshlyInsertedElementToResize should also have a locked in associated resize strategy (and allow the user to pick between them)


editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-metastrategy.tsx
```diff
@@ -48,22 +51,88 @@ import {
...
+      ? interactionSession.interactionData.originalDragStart
+      : interactionSession.interactionData.point
+
+  const cmdPressed = interactionSession.interactionData.modifiers.cmd
+
+  const applicableReparentFactories = getApplicableReparentFactories(
+    canvasState,
+    pointOnCanvas,
+    cmdPressed,
+    true, // <- TODO this is an important assumption, make sure this is still true when inserting into a flex storyboard
```
seanparsons, line 83: Sounds like a case for a test?

balazsbajorics: now that you mentioned it... there's a ton of things here that should be tested - this PR unlocks a bunch of new behavior!!!

seanparsons:
There's a few TODOs sprinkled through this, should they be resolved before merging?

editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-metastrategy.tsx
```diff
@@ -96,8 +165,8 @@ export function drawToInsertStrategy(
       ((interactionSession.interactionData.type === 'DRAG' &&
         interactionSession.activeControl.type === 'RESIZE_HANDLE') ||
         interactionSession.interactionData.type === 'HOVER')
-        ? 1
-        : 0,
+        ? fitness
+        : 0, // TODO is this still needed here?
```
Rheeseyb, line 169: I suspect we can just use the fitness value directly now

editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-metastrategy.tsx
```diff
@@ -48,22 +51,88 @@ import {
...
   }
   const insertionSubject = insertionSubjects[0]
   return {
-    id: 'DRAW_TO_INSERT',
-    name: 'Draw to insert',
+    id: name, // TODO review this before merge
+    name: name,
     controlsToRender: [
       // TODO the controlsToRender should instead use the controls of the actual canvas strategy -> to achieve that, this should be a function of the StrategyState here
       controlWithProps({
```
Rheeseyb, line 138: You can update this to provide the `targetParent` to `ParentOutlines` and `ParentBounds` after #2644 is merged

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-metastrategy.tsx
```diff
@@ -48,22 +51,88 @@ import {
...
 ): CanvasStrategy | null {
   const insertionSubjects = getInsertionSubjectsFromInteractionTarget(canvasState.interactionTarget)
   if (insertionSubjects.length !== 1) {
     return null
   }
   const insertionSubject = insertionSubjects[0]
   return {
-    id: 'DRAW_TO_INSERT',
-    name: 'Draw to insert',
+    id: name, // TODO review this before merge
```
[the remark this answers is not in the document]
balazsbajorics: yeah, I think it's an appropriate ID. I'll delete the comment

editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-metastrategy.tsx
balazsbajorics: done!

bkrmendy (approved):

PR #2663, opened 2022-10-13
Disable flex resize for missing dimensions

ruggi:
**Problem:**

The current flex resize can set missing dimensions, even in situations when it would not be correct to do so (e.g. when there's `flexBasis` set).

**Fix:**

This PR disables the resizing completely if dimensions are missing, and only allows resizing dimensions which are explicitly set already.

**Notes:**
- @balazsbajorics I also thought about hiding the controls at all, but _probably_ they should be there anyways and do nothing, which _seems_ to me it's less confusing. Feel free to disagree of course and lmk if you prefer otherwise.


enidemi (approved):

balazsbajorics (approved):
I tried it and think this resizing behavior is good to go! 😊

(it seems like it regressed some insertion tests, for insertion we do want to keep the behavior as it was on master)

seanparsons (approved):

PR #2672, opened 2022-10-14
Clicking on files in the explorer does not open them in the editor

bkrmendy:
Fixes https://github.com/concrete-utopia/utopia/issues/2671

## Problem:
See https://github.com/concrete-utopia/utopia/issues/2671.

In a technical sense, there two mouseDown event handlers were created, and one of them was "dropped" during the refactors.

## Fix:
Merge the two mouse event handlers.

**Commit Details:**
- Merge `onMouseDown` and `onItemMouseDown`

### DIFF
--- a/editor/src/components/filebrowser/fileitem.tsx
+++ b/editor/src/components/filebrowser/fileitem.tsx
@@ -558,7 +558,7 @@ class FileBrowserItemInner extends React.PureComponent<
     })
   }
 
-  onMouseDown = (e: React.MouseEvent) => {
+  onMouseClick = (e: React.MouseEvent) => {
     if (e.button === 0) {
       if (this.props.fileType !== 'ASSET_FILE' && this.props.fileType !== 'IMAGE_FILE') {
         this.props.setSelected(this.props)
@@ -567,7 +567,9 @@ class FileBrowserItemInner extends React.PureComponent<
         }
       }
     }
+  }
 
+  onMouseDown = (e: React.MouseEvent) => {
     if (this.props.imageFile == null) {
       return
     }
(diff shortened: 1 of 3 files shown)


Rheeseyb (approved):

editor/src/components/filebrowser/fileitem.tsx
```diff
@@ -574,7 +558,16 @@ class FileBrowserItemInner extends React.PureComponent<
     })
   }
 
-  onMouseDown = () => {
+  onMouseDown = (e: React.MouseEvent) => {
+    if (e.button === 0) {
+      if (this.props.fileType !== 'ASSET_FILE' && this.props.fileType !== 'IMAGE_FILE') {
+        this.props.setSelected(this.props)
+        if (this.props.fileType != null && this.props.fileType !== 'DIRECTORY') {
+          this.props.dispatch([EditorActions.openCodeEditorFile(this.props.path, true)], 'everyone')
```
gbalint, line 566: Why on mouse down and not click? 
In real vscode you have to click on a file to open it, and I believe that is more standard than doing actions on mouse down.

Rheeseyb: This would have been because that was the previous logic, where we were tending to use mousedown rather than click events as we found it to be far more reliable at the time (years ago). This probably would be better suited to a click event now though.

bkrmendy: the only difference is that this feels slightly slower (since mousedown happens before the click event)

bkrmendy: (moved to logic to `onClick`)

enidemi (approved):

seanparsons (approved):

PR #2675, opened 2022-10-14
Turn drag-to-insert into a metastrategy

### LINKED ISSUE #2665: Turn drag to insert into a metastrategy
Follow on from #2647 

We now have a draw-to-insert metastrategy, so the goal of this is to use that to inform a very similar drag-to-insert one.

Rheeseyb:
Fixes #2665 

**Problem:**
Because insertion requires use of the reparenting strategies, we are currently picking the best fitting reparent strategy when using drag-to-insert.

**Fix:**
Turn drag-to-insert into a meta strategy, allowing us to provide different insertion options (via the strategy picker) based on the target parent of the element being inserted.

Because of this, we are no longer hard coding the `cmd` modifier during the drag to insert, meaning that we also apply snapping during those insertions. As a result of this I had to fix some existing image insertion tests, which are now triggering this snapping behaviour.

**Note:**
This PR also unlocks some further tests that can be added (based on the different reparent options). I'll be adding those in a follow up PR, as this PR now also fixes a test that is currently flakey on master.


enidemi (approved):

editor/src/components/canvas/canvas-strategies/strategies/drag-to-insert-strategy.spec.browser2.tsx
```diff
@@ -26,6 +27,7 @@ async function setupInsertTest(inputCode: string): Promise<EditorRenderResult> {
 
 async function dragFromInsertMenuDivButtonToPoint(
   targetPoint: { x: number; y: number },
+  modifiers: Modifiers,
```
gbalint, line 30: I guess we should rename this file too to `drag-to-insert-metastrategy.spec.browser2.tsx`

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
```diff
@@ -187,13 +187,16 @@ export function reparentTarget(
 
 type ReparentSubjects = NewReparentSubjects | ExistingReparentSubjects
 
+// FIXME Does it ever make sense for this to refer to more than one element?
```
gbalint, line 190: I decided to allow multiple subjects because I thought maybe we can reuse this later for pasting.

gbalint (approved):

editor/src/components/canvas/canvas-strategies/strategies/drag-to-insert-strategy.spec.browser2.tsx
Rheeseyb: Ha yes! We also should rename the draw-to-insert one too, but I'll do that separately incase someone else is making changes to it

editor/src/components/canvas/canvas-strategies/strategies/drag-to-insert-metastrategy.tsx
```diff
@@ -19,58 +26,131 @@ import {
...
+      ? offsetPoint(interactionData.originalDragStart, interactionData.drag ?? zeroCanvasPoint)
+      : interactionData.point
+
+  const cmdPressed = interactionData.modifiers.cmd
+
+  const applicableReparentFactories = getApplicableReparentFactories(
+    canvasState,
+    pointOnCanvas,
+    cmdPressed,
+    true, // <- TODO this is an important assumption, make sure this is still true when inserting into a flex storyboard
```
seanparsons, line 86: Later work?

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/drag-to-insert-metastrategy.tsx
Rheeseyb: Yeah. This is a duplicate `TODO` from the draw-to-insert-metastrategy, as both need to be addressed together. I'll make sure it's tracked in the doc actually

PR #2648, opened 2022-10-13
Drag and drop image to the file browser through the canvas

bkrmendy:
## Problem:
See detailed description in https://github.com/concrete-utopia/utopia/issues/2650

## Fix:
Disable the cursor overlay when a drag session is in progress, so that it doesn't capture mouse events. This can be done by creating a new state on EditorState (`dragSessionState`) and the associated action (`SET_DRAG_SESSION_STATE`). This state is maintained when starting/ending drag sessions, and it can be checked by the cursor overlay.

**Commit Details:**
- Modify the cursor overlay so that it doesn't capture mouse events
- Create a new state on EditorState (`dragSessionState`) and update the boilerplate
- Create a new action (`SET_DRAG_SESSION_STATE`) with the associated boilerplate
- read `dragSessionState` in `CursorOverlay` to catch/pass pointer events
- set ``dragSessionState` in `FileItem` and `EditorCanvas` during drag sessions
- update tests to work with the new setup

PR #2677, opened 2022-10-17
Cursor overlay: only swallow pointer events when the cursor is set

bkrmendy:
## Problem:
Due to a code change introduced by https://github.com/concrete-utopia/utopia/pull/2648, CursorOverlay swallows pointer events whenever there's no image drag session in progress (either dragging from Finder or form UTFB)

## Fix:
Only set `pointerEvents` to `all` on the cursor overlay if `cursor` is set to a non-null value and no image drag is in progress

**Commit Details:**
- Update `cursor-overlay.tsx` according to Fix, with a heuristic function that determines whether pointer events are caught or not
- add tests to check editor state management of cursor and image dragging session
- reset image dragging session in `onDrop`


maltenuhn:
@bkrmendy it's probably nothing, but I have vague memories that the overlay was at one point also used to to the right thing w/r/t dragging around sliders and colour picker controls in the inspector. I.e. if you do drag those around and over-drag onto the canvas, the mouse movement shouldn't trigger canvas highlights. 

I don't think we had test coverage for that, so it's entirely possible that "unrelated" changes like this could regress is - just give it a quick check before opening up the PR as non-draft please.

editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
```diff
@@ -53,3 +53,21 @@ export const CursorOverlay = React.memo(() => {
     portalDiv,
   )
 })
+
+export interface WorldState {
```
Rheeseyb, line 57: Ha! Please rename this 😁 Or tbh I'm not even sure it's needed? We could just provide the two params separately to the function below?

Rheeseyb (approved):

balazsbajorics (approved):

PR #2679, opened 2022-10-17
Fix/drag to insert not permitted

### LINKED ISSUE #2667: dragging in non-image files from the filebrowser should set the cursor to nonpermitted

enidemi:
Fixes #2667

**Problem:**
When dragging any non-image over the canvas we show the default cursor. It should show the not-permitted cursor.

**Fix:**
I made changes to `DraggingFromSidebar` in the editorstate to store if a non-image is a dragged from the filebrowser. This way inside the `onDragOver` I am aware if there is a filebrowser drag event that contains other file type. From here through the drag-insertion-strategy I can set the editor cursor, in the not-allowed case the insertion subject is an empty array, I had to make changes to insertion strategies because it had errors. 

The tests had a mysterious error with react-dnd so I removed the extra Html5backend.

Also the test-utils for dragging is extended with `types: ['Files']`, this triggers opening the left pane automatically. I changed the tests to open manually first because it shifts the results and disabled the animation in case it's a test environment.

Monosnap screencast 2022-10-17 13-40-16


balazsbajorics:
I'm not entirely sure what should I see on the gif! is the cursor the dragged file entry thingy? or that mickey mouse glove thing?

one thing I'd say is that no matter what the solution, we should _definitely_ stick to the cursor being a function of EditorState. so even if we end up putting them on `document.body.style.cursor`, those places where you manually change it or reset to `initial` look bad to me, and instead we should have a central place that takes the editorState and updates `document.body.style.cursor` as needed (and sets it to initial as well).

bkrmendy (approved):

seanparsons (approved):

gbalint (approved):

PR #2682, opened 2022-10-17
Parse image file name parts

bkrmendy:
## Problem:
Parsing the image multiplier (stuff@2x.png) and drag dropping two images with the same name clash in a surprising way (since stuff@2x.png will be postfixed to `stuff@2x_2.png`). This causes the `@2x` modifier not to be picked up, and the image is inserted in it's original size (instead of the expected, scaled-down size).

## Fix:
Instead of using a regex to determine the image multiplier, parse the image file name into a JS object, which can be manipulated in a more structured way.

## Commit Details:
- add file name interface/parser/printer
- use the new filename splitting machinery when looking for duplicate file names
- use the new filename splitting machinery when determining image multiplier
- tests


editor/src/components/images.ts
```diff
@@ -54,14 +48,88 @@ export function getImageSrc(
...
+  return {
+    rest: parts.slice(0, -1).join(sepBy),
+    part: made,
+  }
+}
+
+const safeParseInt = (raw: string): number | null => {
+  const result = Number.parseInt(raw)
+  return isNaN(result) ? null : result
+}
```
Rheeseyb, line 94: We have a `safeParseInt` in [`math-utils.ts`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/math-utils.ts#L915) which is truly awful, and thankfully unused. Please can you replace that version with this one

editor/src/components/images.ts
```diff
@@ -54,14 +48,88 @@ export function getImageSrc(
...
+  return `${filename}${dedupeString}${multiplierString}.${extension}`
+}
+
+export interface LastPartSeparatedByResult<T> {
+  part: T
+  rest: string
+}
+
+function lastPartSeparatedBy<Separator extends string, T>(
+  sepBy: Separator,
```
Rheeseyb, line 71: Is there any benefit here to creating a specific type for this?

bkrmendy: well this can go either way, I created the type so it's easier to see what the function separates along, and so that the first parameter is not a simple string

I'm ok with converting the function into one that takes an object instead of two strings, if that would be an improvement

editor/src/core/model/project-file-utils.ts
```diff
@@ -644,44 +639,44 @@ export function switchToFileType(from: ProjectFile, to: ProjectFileType): Projec
...
+  const startingIDCorrected = correctProjectContentsPath(filename)
+  const fileWithSameNameExistsAlready =
+    getContentsTreeFileFromString(projectContents, startingIDCorrected) != null
+
+  if (!fileWithSameNameExistsAlready) {
+    return startingIDCorrected
+  }
+
+  const parts = getFilenameParts(startingIDCorrected)
+  if (parts == null) {
```
Rheeseyb, line 654: It looks like both paths of this `if` statement are doing almost exactly the same thing, with the only difference being the value of `possibleNewID`, in which case you could just set that via a ternary

bkrmendy: or, I could generalise `getFilenameParts` so that it doesn't fail if a file doesn't have an extension, in which case the two branches will do the exact same work (I wanted to do this anyway but decided against because it felt like scope creep), would work instead?

editor/src/components/images.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+      [{ filename: 'stuff', extension: 'png', multiplier: 2 }, 'stuff@2x.png'],
+      [
+        { filename: 'stuff', extension: 'png', multiplier: 2, deduplicationSeqNumber: 2 },
+        'stuff_2@2x.png',
+      ],
+      [
+        { filename: 'stuff@3', extension: 'png', multiplier: 2, deduplicationSeqNumber: 2 },
+        'stuff@3_2@2x.png',
+      ],
+    ]
```
balazsbajorics, line 49: nit: for these tests, I think it'd read much better if you'd flip the input array to be [[input, result]] !

editor/src/components/images.ts
balazsbajorics: yeah I don't think it adds valuable information to make the Separator a generic type like this

editor/src/components/images.ts
bkrmendy: thanks, done!

editor/src/components/images.ts
bkrmendy: thanks for the feedback, I removed the unnecessary generic param

editor/src/components/images.spec.ts
bkrmendy: absolutely, fixed!

Rheeseyb (approved):

balazsbajorics (approved):

Issue #2692, opened 2022-10-18
Drag-and-dropping an image to the canvas takes 2 undo steps to undo


bkrmendy:
## Describe the bug
Drag-and-dropping an image to the canvas from the file system takes 2 undo steps to undo

## Expected behaviour
A single undo should revert the effect of the drop

PR #2693, opened 2022-10-19
DnD'ing an image to the canvas takes 2 undos to undo

bkrmendy:
Fixes https://github.com/concrete-utopia/utopia/issues/2692

## Problem:
See title and https://github.com/concrete-utopia/utopia/issues/2692

## Reason
When an image is dropped, a new `` element is inserted into the DOM, and the corresponding image asset is uploaded and added to the `assets/` directory. Both of these steps are undoable (non-transient), so the first undo will undo adding the image file to the `assets/` directory, and the second undo undoes adding the `` to the DOM

## Fix:
In DropHandlers.onDrop, upload assets using `saveAssets`, instead of using `SAVE_ASSETS` action. This way, all dispatched actions are in the same `dispatch` call (instead of spread over multiple ones)

### DIFF
--- a/editor/src/templates/image-drop.tsx
+++ b/editor/src/templates/image-drop.tsx
@@ -19,7 +19,9 @@ import { generateUidWithExistingComponentsAndExtraUids } from '../core/model/ele
 import React from 'react'
 import { CanvasPositions } from '../components/canvas/canvas-types'
 import { EditorState, notDragging } from '../components/editor/store/editor-state'
-import { uniqueProjectContentID } from '../core/model/project-file-utils'
+import { imageFile, uniqueProjectContentID } from '../core/model/project-file-utils'
+import { AssetToSave, saveAssets } from '../components/editor/server'
+import { notice } from '../components/common/notice'
 
 export async function getPastedImages(dataTransfer: DataTransfer): Promise<ImageResult[]> {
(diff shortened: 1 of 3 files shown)


bkrmendy:
This PR is more like a proposal, I totally need some context around why `UPLOAD_FILE` is an undoable action on master

gbalint (approved):

editor/src/templates/image-drop.tsx
```diff
@@ -113,6 +115,14 @@ async function onDrop(
       'autoincrement',
     )
 
+    void saveAssets(context.editor.id!, assetInfo)
```
balazsbajorics, line 118: should you await this before dispatching `saveImageActions: [EditorActions.updateFile...`?

bkrmendy: it could be awaited, my thinking was that

- the stuff you dragged to the canvas should appear ASAP, with minimal latency
- if something goes wrong, it's super easy to get back to the previous state with a ⌘-Z
no strong opinions though

balazsbajorics (approved):

Rheeseyb (commented):
I think we should handle the different cases where the broken image icon can appear here (failed upload, or successful upload that took longer than rendering the `` tag), but otherwise 👍

Issue #2689, opened 2022-10-18
drag-to-insert doesn’t show element outline


gbalint:
drag-to-insert doesn’t show element outline (this is visible when you drag into flex, because then we don’t render the element just the position indicators)

PR #2694, opened 2022-10-19
Outline control for drag to insert

gbalint:
Fixes https://github.com/concrete-utopia/utopia/issues/2689

**Problem:**
Drag outline control is not visible during drag to insert

**Root cause:**
`DragOutlineControl` calculates the bounding box of the target elements using the `startingMetadata`. But that metadata does not contain the element which will be just inserted during this interaction

**Fix:**
`DragOutlineControl` can receive either an array of element paths or an array of frames to render the outline. In the  drag-to-insert strategy we can just pass the frames, so the control does not need to refer to the metadata to get the bounds.

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/drag-outline-control.tsx
+++ b/editor/src/components/canvas/controls/select-mode/drag-outline-control.tsx
@@ -1,50 +1,96 @@
 import React from 'react'
+import {
+  boundingRectangleArray,
+  CanvasRectangle,
+  offsetRect,
+} from '../../../../core/shared/math-utils'
 import { ElementPath } from '../../../../core/shared/project-file-types'
+import { assertNever } from '../../../../core/shared/utils'
 import { useColorTheme } from '../../../../uuiui'
 import { useEditorState } from '../../../editor/store/store-hook'
 import { controlForStrategyMemoized } from '../../canvas-strategies/canvas-strategy-types'
 import { getMultiselectBounds } from '../../canvas-strategies/strategies/shared-move-strategies-helpers'
 import { CanvasOffsetWrapper } from '../canvas-offset-wrapper'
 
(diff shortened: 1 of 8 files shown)


bkrmendy (approved):

editor/src/components/canvas/controls/select-mode/drag-outline-control.tsx
```diff
@@ -1,50 +1,96 @@
...
+export function dragTargetsFrame(frames: Array<CanvasRectangle>): DragTargetsFrames {
+  return {
+    type: 'frames',
+    frames: frames,
+  }
+}
+
+type DragOutlineControlProps = DragTargetsElementPaths | DragTargetsFrames
+
+export const DragOutlineControl = controlForStrategyMemoized((props: DragOutlineControlProps) => {
```
balazsbajorics, line 43: probably nitpick, but it feels weird to me that `props` _is_ the tagged union type, with a `props.type` field on it.
Somehow it'd feel more natural to me if `props.targets` would be the union type. Again, nitpick, but it just sort of breaks my brain that this is the only props object in Utopia which is _also_ a tagged union.

balazsbajorics: hmm but at the same time, if props.targets would be `DragTargetsElementPaths | DragTargetsFrames` then it'd bust the shallow equality. aaa! okay I changed my mind. it's good as it is :D

balazsbajorics (approved):

editor/src/components/canvas/controls/select-mode/drag-outline-control.tsx
Rheeseyb: Agree, and agree 😀

Rheeseyb (commented):
👍

PR #2696, opened 2022-10-19
No outline for image insert

gbalint:
**Problem:**
https://github.com/concrete-utopia/utopia/pull/2694 broke image insert, we should not show an outline when dragging an image to insert.

**Fix:**
Filter out images from the insertion subjects


balazsbajorics (approved):

bkrmendy (approved):

Rheeseyb (commented):
👍

PR #2699, opened 2022-10-20
canvas cursor without screensize overlay

enidemi:
https://utopia.pizza/p?branch_name=spike-canvas-cursors-no-overlay

Trying to get rid of the big cursor overlay causing so many problems recently :)  

I added everything that's needed to the current CursorOverlay to keep them in one place.

I checked it with canvas interactions, filebrowser drag and drop and also looked at if it keeps the canvas interaction cursor over input fields. So far it's good.

About the original Cursor Overlay: it was introduced in 2019 at that point our canvas interactions were different, it was a solution to [this issue](https://github.com/concrete-utopia/dystopia/issues/848)


editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
```diff
@@ -18,56 +18,30 @@ export function getCursorForOverlay(editorState: EditorState): CSSCursor | null
...
-      workingStyleProps.cursor = cursor
+      document.body.classList.add('customCursor')
+      document.body.style.cursor = cursor
+    } else {
+      document.body.classList.remove('customCursor')
+      document.body.style.cursor = 'unset'
+    }
+
+    if (editor != null) {
+      editor.style.pointerEvents = 'initial'
```
Rheeseyb, line 35: This could just go in the `index.html` unless there's a reason to set it via the `useEffect`

enidemi: I want to avoid deleting it accidentally

editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
```diff
@@ -18,56 +18,30 @@ export function getCursorForOverlay(editorState: EditorState): CSSCursor | null
 
 export const CursorOverlay = React.memo(() => {
```
balazsbajorics, line 19: I think this should be renamed, since it's not an overlay anymore :)

gbalint (approved):

editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
Rheeseyb: I don't understand?

Rheeseyb (approved):

seanparsons (approved):
Nice!

editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
enidemi: well, I was thinking that we frequently change inline styles and it's easy to break this, but I realized that we rarely touch the html, so I moved it there now

editor/src/components/canvas/controls/select-mode/cursor-overlay.tsx
enidemi: fixed

PR #2701, opened 2022-10-20
Merge flex reparent strategies

gbalint:
**Problem:**
absolute-reparent-to-flex and flex-reparent-to-flex strategies are very similar, they are easy to merge.
I'm working on reparent-to-flow, and I would like to start from reparent-to-flex strategies, but I don't want to separate these strategies by their original layout, because that would result in just too many very similar strategies.

**Fix:**
I merged the two strategies into a single one.


editor/src/components/canvas/canvas-strategies/strategies/reparent-to-flex-strategy.tsx
```diff
@@ -35,22 +36,29 @@ export function baseAbsoluteReparentToFlexStrategy(
...
+    const isOriginallyAbsolute =
       MetadataUtils.findElementByElementPath(
         canvasState.startingMetadata,
         filteredSelectedElements[0],
-      )?.specialSizeMeasurements.position !== 'absolute'
-    ) {
+      )?.specialSizeMeasurements.position === 'absolute'
+
+    const isOriginallyFlex =
+      MetadataUtils.isParentYogaLayoutedContainerAndElementParticipatesInLayout(
```
Rheeseyb, line 46: We could probably prefix this with `!isOriginallyAbsolute && `

editor/src/components/canvas/canvas-strategies/strategies/reparent-to-flex-strategy.tsx
```diff
@@ -35,22 +36,29 @@ export function baseAbsoluteReparentToFlexStrategy(
...
+        selectedElements[0],
```
Rheeseyb, line 47: I think this should be `filteredSelectedElements`

editor/src/components/canvas/canvas-strategies/strategies/reparent-to-flex-strategy.tsx
```diff
@@ -75,12 +83,14 @@ export function baseAbsoluteReparentToFlexStrategy(
       ],
       fitness: fitness,
       apply: () => {
-        return applyFlexReparent(
-          'strip-absolute-props',
-          canvasState,
-          interactionSession,
-          reparentTarget,
-        )
+        return interactionSession == null
```
Rheeseyb, line 86: This check should be unnecessary as we've already ensured that `interactionSession` is not null earlier on in the factory

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-to-flex-strategy.tsx
gbalint: The source of mistake is that I started from the flex-reparent-to-flex strategy, and that did not use filteredSelectedElements at all

PR #2714, opened 2022-10-24
Choose Github repositories from those accessible to the user.

### LINKED ISSUE #2712: Github target repository selection should be possible via a combo-box.
By default the combo box should be populated with the repositories the user has access to.

seanparsons:
Fixes #2712

**Problem:**
Currently the user must remember what repositories they have access to and type the URL in of the one they wish to operate on by hand. Which is a somewhat error prone process.

**Fix:**
Now we retrieve the public repositories that the user has access to from Github and provide those in a dropdown for the user to select them from.

**Commit Details:**
- Added call in the backend to get a page of the accessible repositories.
- Added `getGithubUsersPublicRepositories` to the backend which paginates the repository listing and combines them all into one.
- Updated `getGithubBranches` to use a slightly better function with `unfoldM` that avoids making an extra call to the server.
- Added endpoint for retrieving user's repositories.
- Added `getUsersPublicGithubRepositories` to the frontend which handles triggering a callback on a successful result.
- Change the control from a `StringInput` to a `CreatableSelect` to allow the user to choose the repository from the list as well as potentially specify a new one.

### DIFF
--- a/editor/src/components/navigator/left-pane.tsx
+++ b/editor/src/components/navigator/left-pane.tsx
@@ -861,7 +861,7 @@ const GithubPane = React.memo(() => {
 
   React.useEffect(() => {
     if (githubAuthenticated) {
-      getUsersPublicGithubRepositories(dispatch, setUsersRepositoriesCallback)
+      void getUsersPublicGithubRepositories(dispatch, setUsersRepositoriesCallback)
     }
   }, [githubAuthenticated, dispatch, setUsersRepositoriesCallback])
 
--- a/editor/src/components/navigator/left-pane.tsx
+++ b/editor/src/components/navigator/left-pane.tsx
@@ -17,7 +17,7 @@ import { isParseSuccess, isTextFile, ProjectFile } from '../../core/shared/proje
 import { NO_OP } from '../../core/shared/utils'
 import { auth0Url, BASE_URL, FLOATING_PREVIEW_BASE_URL } from '../../common/env-vars'
 import { shareURLForProject } from '../../core/shared/utils'
-import Utils from '../../utils/utils'
+import Utils, { isOptionType } from '../../utils/utils'
 import {
   useColorTheme,
   UtopiaTheme,
@@ -37,7 +37,7 @@ import {
   Icons,
   Avatar,
 } from '../../uuiui'
-import { SelectOption, User } from '../../uuiui-deps'
+import { getControlStyles, SelectOption, User } from '../../uuiui-deps'
 import { setFocus } from '../common/actions'
 import { EditorDispatch, LoginState } from '../editor/action-types'
 import * as EditorActions from '../editor/actions/action-creators'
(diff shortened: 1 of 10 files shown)


editor/src/components/navigator/left-pane.tsx
```diff
@@ -959,14 +1060,36 @@ const GithubPane = React.memo(() => {
...
-            <StringInput
-              testId='saveToGithubInput'
-              value={targetRepository}
-              disabled={!githubAuthenticated}
+            <CreatableSelect
+              autoFocus={false}
+              placeholder={
+                usersRepositories == null ? 'Loading repositories...' : 'owner/repository'
+              }
+              value={currentTargetRepositoryValue}
```
Rheeseyb, line 1073: I think this code implies we will show dubious values if the project isn't yet connected to a github repo?

editor/src/components/navigator/left-pane.tsx
```diff
@@ -959,14 +1060,36 @@ const GithubPane = React.memo(() => {
...
+              inputValue={targetRepository}
+              isDisabled={false}
               onChange={onChangeTargetRepository}
+              onInputChange={onInputChangeTargetRepository}
+              components={{
+                IndicatorsContainer: () => null,
+                Input,
+              }}
+              className='className-inspector-control'
```
ruggi, line 1082: copypasta 🍝

server/src/Utopia/Web/Executors/Common.hs
```diff
@@ -390,11 +390,11 @@ createTreeAndSaveToGithub githubResources awsResource logger metrics pool userID
     pure (branchName, view (field @"url") branchResult, commitSha)
   pure $ either responseFailureFromReason responseSuccessFromBranchNameAndURL result
 
-convertBranchesResultToUnfold :: Int -> GetBranchesResult -> Maybe (GetBranchesResult, Int)
+convertBranchesResultToUnfold :: Int -> GetBranchesResult -> Maybe (GetBranchesResult, Maybe Int)
 convertBranchesResultToUnfold page result =
-  case Protolude.null result of
-    True  -> Nothing
-    False -> Just (result, page + 1)
+  -- We expect some X elements per page, if we get less than that, assume we've hit the end and should not
```
ruggi, line 395: 👍

server/src/Utopia/Web/Executors/Common.hs
```diff
@@ -431,6 +432,31 @@ getGithubBranch githubResources awsResource logger metrics pool userID owner rep
       pure (projectContent, commitSha)
     pure $ either getBranchContentFailureFromReason getBranchContentSuccessFromContent result
 
+convertUsersRepositoriesResultToUnfold :: Int -> GetUsersPublicRepositoriesResult -> Maybe (GetUsersPublicRepositoriesResult, Maybe Int)
+convertUsersRepositoriesResultToUnfold page result =
+  -- We expect some X elements per page, if we get less than that, assume we've hit the end and should not
+  -- continue paginating.
+  Just (result, if Protolude.length result < userRepositoriesPerPage then Nothing else Just (page + 1))
```
ruggi, line 439: Could this logic be extracted and reused (e.g. in the `convertBranchesResultToUnfold` as well)?

editor/src/components/navigator/left-pane.tsx
seanparsons: The value will be `undefined` in the case where the project isn't yet connected, which results in the placeholder being shown.

editor/src/components/navigator/left-pane.tsx
seanparsons: Bah, good spot.

server/src/Utopia/Web/Executors/Common.hs
seanparsons: Quite possibly, but since there's a constant in there I didn't think there was much gain in generalising it given it's basically a one-liner.

PR #2715, opened 2022-10-24
Add flow cases to reparent metastrategy

### LINKED ISSUE #2704: Extend the reparent metastrategy to include cases for flow reparent

balazsbajorics:
Fixes #2704 

**Problem**
Currently on the canvas you can drag-to-reparent between elements that are absolutely positioned in a flow layout (called Absolute Layout) and elements that are in a flex layout (called Flex Layout). 
You cannot reparent into a flow layout parent as a positon: static element (called Flow Layout). Also, you cannot currently take elements that are in Flow Layout and reparent them to any parent.

**Fix**
I added new test cases based on the paper doc, then added a function called `flowParentAbsoluteOrStatic` which takes a display: flow target parent, and tells (based on a ruleset) if it should be an Absolute reparent, or a Flow reparent.
* During this step of the reparent, we actually only care if we reparent something as absolute, or as static, so I changed the reparent target type from `'REPARENT_TO_ABSOLUTE' | 'REPARENT_TO_FLEX'` to `'REPARENT_AS_ABSOLUTE' | 'REPARENT_AS_STATIC'`. 
* Removed some extra isApplicable checks from the individual reparent strategies, to allow reparent _from_ Flow.

**Note:** For legacy reasons we talk about Absolute Layout, Flex Layout, and Flow Layout, but in reality there's Flow and Flex as layouts, and an element may participate in this layout, or break out using position: absolute (regardless of the parent layout type). There's some things that we could further clear up, for example an absolute child in a display: flex parent should be treated as an Absolute Layout element, I think there's some cases today where this is not the case.


bkrmendy (approved):

editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
```diff
@@ -274,7 +274,7 @@ describe('Strategy Fitness', () => {
           data-uid='bbb'
         />
         <div
-          style={{ backgroundColor: '#0091FFAA', width: 100, height: 50 }}
+          style={{ backgroundColor: '#0091FFAA', width: 100, height: 50, contains: 'layout' }}
```
Rheeseyb, line 277: This should be `contain`, not `contains`

editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-to-flex-strategy.spec.browser2.tsx
```diff
@@ -315,7 +315,7 @@ describe('Absolute Reparent To Flex Strategy', () => {
     await renderResult.getDispatchFollowUpActionsFinished()
 
     const dragDelta = windowPoint({
-      x: firstFlexChildEnd.x - absoluteChildCenter.x + 50,
+      x: firstFlexChildEnd.x - absoluteChildCenter.x + 10,
```
Rheeseyb, line 318: Ha! I remember this one :D

editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
balazsbajorics: so how does it pass lol

editor/src/components/canvas/canvas-strategies/strategies/flex-reparent-to-absolute-strategy.tsx
```diff
@@ -47,15 +47,6 @@ export function baseFlexReparentToAbsoluteStrategy(
     customStrategyState: CustomStrategyState,
   ): CanvasStrategy | null => {
     const selectedElements = getTargetPathsFromInteractionTarget(canvasState.interactionTarget)
```
Rheeseyb, line 49: Should this file / strategy be renamed now?

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-strategy.tsx
```diff
@@ -65,7 +65,7 @@ export function flowReorderStrategy(
       interactionSession != null &&
       interactionSession.interactionData.type === 'DRAG' &&
       interactionSession.activeControl.type === 'BOUNDING_AREA'
-        ? 3
+        ? 1
```
Rheeseyb, line 68: Why has this fitness value changed?

editor/src/components/canvas/canvas-strategies/strategies/forced-absolute-reparent-strategies.spec.browser2.tsx
```diff
@@ -732,22 +743,10 @@ describe('Forced Absolute Reparent Strategies', () => {
...
-            top: 200,
-          }}
-          data-uid='flexchild1'
-          data-testid='flexchild1'
-        />
       </div>`),
     )
   })
-  it('Absolute to absolute on a target parent element without size', async () => {
+  it('Absolute to static on a target parent element without size', async () => {
```
Rheeseyb, line 749: Interesting... was this change deliberate?

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
```diff
@@ -97,17 +96,86 @@ export function reparentStrategyForParent(
...
+  const children = MetadataUtils.getChildren(metadata, parent)
+
+  const storyboardRoot = EP.isStoryboardPath(parent)
+  if (storyboardRoot) {
+    // always reparent as absolute to the Storyboard
+    return 'REPARENT_AS_ABSOLUTE'
+  }
+
+  if (parentMetadata == null) {
+    // should this throw an error?
```
Rheeseyb, line 132: If there is the possibility of some race conditions that could lead to this, then we should either fail gracefully (and not reparent at all), or we should throw an error.

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
```diff
@@ -129,41 +197,39 @@ export function findReparentStrategies(
...
-    forcedTarget != null &&
-    EP.pathsEqual(forcedTarget.newParent, strictTarget.newParent)
-  const convertToAbsolute = sameTargets && strictStrategy?.strategy === 'REPARENT_TO_FLEX'
-  const skipForcedTarget = sameTargets && !convertToAbsolute
-  const forcedStrategy =
-    forcedTarget == null || skipForcedTarget
-      ? null
-      : reparentStrategyForReparentTarget(metadata, forcedTarget, convertToAbsolute)
+  if (targetParent == null) {
+    return [] // TODO is canvas a null target parent?
```
Rheeseyb, line 217: Ha I don't know, but please find out before merging!

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
```diff
@@ -129,41 +197,39 @@ export function findReparentStrategies(
...
+  }
+
+  const strategy = reparentStrategyForReparentTarget(metadata, targetParent, false)
+
+  const fallbackStrategy: FindReparentStrategyResult =
+    strategy.strategy === 'REPARENT_AS_ABSOLUTE'
+      ? {
+          isFallback: true, // TODO is this true?
```
Rheeseyb, line 225: I'm very confused about what's happening here and why!

editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
Rheeseyb: You tell me! https://developer.mozilla.org/en-US/docs/Web/CSS/contain

editor/src/components/canvas/canvas-strategies/strategies/absolute-flex-reparent-strategy-helpers.spec.browser2.tsx
```diff
@@ -854,6 +852,738 @@ describe('Unified Reparent Fitness Function Tests', () => {
...
+                data-uid='draggedElement'
+                data-testid='draggedElement'
+              />
+            </div>
+          </div>
+      `),
+      )
+    })
+
+    it('target parent display: flow with any children in flow', async () => {
```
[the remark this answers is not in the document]
Rheeseyb: Jeez do I have to do everything for you?!
```suggestion
    it('target parent with any children in flow', async () => {
```

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-strategy.tsx
gbalint: Flow reorder allows you to leave the parent, so this way reparent to flow would never win, even when you drag over a new parent.

editor/src/components/canvas/canvas-strategies/strategies/forced-absolute-reparent-strategies.spec.browser2.tsx
balazsbajorics: if you reparent into a sizeless display:flow target parent, according to the new rules that always leads the child being converted to static. since this test is _primarily_ testing sizeless parents, and the new correct behavior is a static reparent, I changed the test to reflect that.

editor/src/components/canvas/canvas-strategies/strategies/flex-reparent-to-absolute-strategy.tsx
balazsbajorics: no, we decided that we still keep the two "...toAbsolute" reparents separate, and maybe unify them later, but they are doing different enough things to warrant a separate strategy. but now the metaStrategy decides which one sshould run, so this applicability check was redundant and useless

editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
balazsbajorics: fixed

editor/src/components/canvas/canvas-strategies/strategies/absolute-flex-reparent-strategy-helpers.spec.browser2.tsx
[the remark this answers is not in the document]
balazsbajorics: fixed! I renamed all occurences of "display: flow" to "flow layout"

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
balazsbajorics: I changed it to throw an error. it should be an impossible condition to trigger it, let's see :)

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
balazsbajorics: I added some clarification comments in the code, and removed the TODO questions

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
Rheeseyb: Thank you!

PR #2716, opened 2022-10-24
Resizing images respects their aspect ratio

bkrmendy:
## Problem:
Resizing images/elements that have `data-aspect-ratio-locked` set doesn't respect their aspect ratio by default, which makes it easy to deform an image/aspect ratio locked element.

## Fix / Changes compared to `master`:
- When resizing a single image, the image's aspect ratio is preserved by default. If `shift` is held, the aspect ratio is disregarded, and the image can be resized in an arbitrary way.
- When resizing an element that has `data-aspect-ratio-locked` set, the element's aspect ratio is preserved by default. If `shift` is held, the aspect ratio is disregarded, and the element  can be resized in an arbitrary way.
- In a multiselection, if any element in the multiselection is an image or has `data-aspect-ratio-locked` set, the whole multiselection is resized with preserving the aspect ratio of the multiselection bounding box (this can be overridden by holding `shift` like above).


editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
```diff
@@ -212,6 +212,49 @@ export var ${BakedInStoryboardVariableName} = (props) => {
...
+  <Storyboard data-uid='0cd'>
+    <div
+      data-aspect-ratio-locked
+      style={{
+        backgroundColor: '#0091FFAA',
+        position: 'absolute',
+        left: 33,
+        top: 307,
+        width: 363,
+        height: 426,
```
balazsbajorics, line 228: I think it'd make the readability better if these numbers would be round, like 300x40

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
```diff
@@ -364,6 +407,462 @@ describe('Absolute Resize Strategy', () => {
       `),
     )
   })
+  it('resizes absolute element with aspect ratio locked, keeping aspect ratio', async () => {
+    const renderResult = await renderTestEditorWithCode(
+      projectWithAspectRatioLockedElement,
+      'await-first-dom-report',
+    )
+
+    const target = EP.elementPath([['0cd', '7a0']])
```
balazsbajorics, line 416: if an element has data-uid prop, it means you can give it an arbitrary uid (so long as it's unique), so you could make this selection ['aaa', 'bbb'], or something similar human readable, like ['sb', 'aspect-ratio-locked-element']

editor/src/components/canvas/canvas-strategies/strategies/resize-helpers.ts
```diff
@@ -99,16 +107,51 @@ export function getLockedAspectRatio(
...
+}
+
+export function isAnySelectedElementAspectRatioLocked(
+  jsxMetadata: ElementInstanceMetadataMap,
+  selectedElements: Array<ElementPath>,
+): boolean {
+  return MetadataUtils.findElementsByElementPath(jsxMetadata, selectedElements).some((e) => {
+    const h = isElementImage(e)
+    const w = isElementAspectRatioLocked(e)
+    return h || w
```
balazsbajorics, line 151: h and w? w the f? :D

balazsbajorics: I think `return isElementImage(e) || isElementAspectRatioLocked(e)` is better. 
no need to push temporary variables in the PR reader's mental stack only to pop them immediately!

bkrmendy: absolutely, that was the original version, the one that you commented on got pushed by accident

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
```diff
@@ -364,6 +407,462 @@ describe('Absolute Resize Strategy', () => {
...
+    const dragDelta = windowPoint({ x: 29, y: -23 })
+
+    await renderResult.dispatch([selectComponents([target], false)], true)
+    resizeElement(renderResult, dragDelta, EdgePositionTopLeft, emptyModifiers)
+    await renderResult.getDispatchFollowUpActionsFinished()
+
+    expect(getPrintedUiJsCode(renderResult.getEditorState()))
+      .toEqual(`import * as React from 'react'
```
balazsbajorics, line 424: I don't like something about these tests: 
because they share the same starter project, the output is very noisy. in these tests you are basically asserting 1-2-4 lines of change, but you hide those lines in 45 lines of project content. 
plus, because the project code is declared up top, you have to scroll a lot if you want to compare the project contents before-and-after. so for tests at the bottom, you scroll up 800 lines, memorize the project contents, scroll to the test, try to find the one line of change.

I think this would be better off if each test would have its own tailored starting project contents, even if that makes the test file longer. for test cases that only need a single element, you could use a projectContents with only that one element, making the diff much easier to see.

bkrmendy: would it be OK to create a "template" that can be instantiated with differenct top/left/width/height values (such as https://github.com/concrete-utopia/utopia/blob/f887383496127e184284320648bd1b9cf4a5807f/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx#L258)?

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
bkrmendy: thanks! fixed

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
balazsbajorics: yeah that would be awesome!

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
bkrmendy: thanks! updated the uid

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
Rheeseyb: You can cut the template down quite a lot too by using `makeTestProjectCodeWithSnippet`

Rheeseyb: There's a good example of this actually that @ruggi introduced: https://github.com/concrete-utopia/utopia/blob/d39e5225c50a7359eff05cbb2e83316fa5aa3515/editor/src/components/canvas/canvas-strategies/strategies/flex-resize-basic-strategy.spec.browser2.tsx#L708-L761

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
bkrmendy: done!

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
bkrmendy: I followed Balazs's suggestion and created minimal test fixtures

balazsbajorics (approved):

PR #2717, opened 2022-10-25
feature(canvas) reorder strategy uses extended sibling frame area

### LINKED ISSUE #2725: flow reorder: extend target area to use parent dimension in cross-axis

enidemi:
Closes #2725
Changing reorder strategies to instead of just the sibling frame use the parent size in cross-axis. Solves the issue with siblings with different flex alignments and flow inline blocks which are automatically offset based on their text children.

https://utopia.pizza/p/d18a3362-foil-tarn/?branch_name=feature-canvas-reorder-area

Updating flex reorder tests to use karma and real metadata.


bkrmendy (approved):

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
```diff
@@ -75,8 +75,10 @@ function areNonWrappingSiblings(
   })
 }
 
-function getElementDirection(element: ElementInstanceMetadata): 'vertical' | 'horizontal' {
-  const displayValue = element.specialSizeMeasurements.display
+export function getElementDirection(
+  element: ElementInstanceMetadata | null,
+): 'vertical' | 'horizontal' {
+  const displayValue = element?.specialSizeMeasurements.display
   return displayValue === 'inline' || displayValue === 'inline-block' ? 'horizontal' : 'vertical'
```
balazsbajorics, line 82: not part of this PR, but we should actually use `displayValue.includes('inline')` to also include inline-flex and inline-grid and whatever else
```suggestion
  return displayValue.includes('inline') ? 'horizontal' : 'vertical'
```

enidemi: 👍

balazsbajorics (approved):

PR #2718, opened 2022-10-25
remove aspect ratio locked prop instead of setting it to false

bkrmendy:
## Problem
When toggling the aspect ratio lock, it's set to `false` in the JSX markup. Instead, removing the prop would be a slight improvement since it reduces clutter in the JSX markup and removes a utopia-specific artefact from the code.

## Fix 
Remove the aspect ratio lock prop when it's set to `false`

### DIFF
--- a/editor/src/components/inspector/inspector.spec.browser2.tsx
+++ b/editor/src/components/inspector/inspector.spec.browser2.tsx
@@ -1,5 +1,6 @@
 import { r } from 'tar'
 import { elementPath } from '../../core/shared/element-path'
+import { assertNever } from '../../core/shared/utils'
 import { mouseClickAtPoint } from '../canvas/event-helpers.test-utils'
 import { getPrintedUiJsCode, renderTestEditorWithCode } from '../canvas/ui-jsx.test-utils'
 import { selectComponents } from '../editor/actions/action-creators'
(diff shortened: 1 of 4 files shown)


editor/src/components/inspector/inspector.spec.browser2.tsx
```diff
@@ -0,0 +1,99 @@
...
+      }}
+      data-uid='7a0'
+    />
+  </Storyboard>
+)
+`)
+  })
+
+  it('toggle aspect ratio lock on', async () => {
+    const codeAfterToggle = await runToggleAspectRatioLockTest(false)
```
balazsbajorics, line 33: please make `aspectRatioLocked` a string literal union, to improve the readability here

bkrmendy: thanks, fixed!

balazsbajorics (approved):

editor/src/components/inspector/inspector.tsx
```diff
@@ -312,8 +313,15 @@ export const Inspector = React.memo<InspectorProps>((props: InspectorProps) => {
...
-      return setAspectRatioLock(path, !aspectRatioLocked)
+    const aspectRatioLockPropPath = PP.create(['data-aspect-ratio-locked'])
+    const actions: EditorAction[] = selectedViews.flatMap((path): EditorAction[] => {
+      if (aspectRatioLocked) {
+        return [unsetProperty(path, aspectRatioLockPropPath)]
+      }
+      return [
+        setProperty(path, aspectRatioLockPropPath, jsxAttributeValue(true, emptyComments)),
+        setAspectRatioLock(path, true),
+      ]
```
Rheeseyb, line 324: I think this could be simplified significantly by modifying the action handler itself: https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/editor/actions/actions.tsx#L4156 (this would also make the behaviour more consistent if we ever fire that action elsewhere in the editor)

bkrmendy: done!

Rheeseyb (approved):

PR #2721, opened 2022-10-25
Updated Ancestor Metastrategy

Rheeseyb:
**Problem:**
Projects will often contain nested elements, and sometimes when dragging an element to move / reparent / reorder it the intention is actually to apply that change to an ancestor of the selected element (e.g. in the case of `<div><div>Drag Me</div></div>`). We already have a strategy for doing that, but it was built before our concept of meta strategies and strategy factories had evolved to a suitable point, and so was only able to find the best fitting strategy for the ancestor and apply that.

**Fix:**
 Replace the `lookForApplicableParentStrategy` with `ancestorMetaStrategy`. This new meta strategy will only apply when the target of the interaction is a single element with static positioning, and that element is the only child of its parent. As before, this will keep bubbling up until we reach the nearest ancestor where this isn't the case, and will then return all of the strategies applicable to that element.

As we only want the ancestor strategies to kick in when dragging the element itself (rather than dragging e.g. resize controls on the element), this PR groups together the strategies that are actually compatible with the ancestor meta strategy in `canvas-strategies.tsx` (creating a `moveOrReorderStrategies: MetaCanvasStrategy` along the way).

When the ancestor metastrategy returns applicable strategies, it alters the name (the altered name here includes a prefix stating how many levels up the hierarchy the affected element will be), increases the fitness (as these strategies should always take priority), and applies the same cursor change and target highlighting as before.

When an ancestor triggers reparenting, the `startingTargetParentsToFilterOut` will be invalid, which would trigger undesirable reparenting behaviour. To address this, rather than capturing those targets at the start of the interaction, we now capture them in the reparent meta strategy, and memoise it based on the starting metadata and interaction targets.

Finally, I've had to make a change to the running of the `SetElementsToRerenderCommand`, so that it also includes descendants of the targets. This is because otherwise the selective DOM walker will only run for the provided targets, meaning that if an ancestor strategy is being applied it won't update the live metadata for the original target being dragged (which leads to an incorrect outline control being rendered).

**Note:**
I did try adjusting the heuristics used to determine when the ancestor strategy should be applicable, as I was trying to check if the parent and child are contiguous elements (i.e. the frames match), but that isn't the case when a parent doesn't specify dimensions, as that leads to the parent filling the grandparent in one dimension. We previously used a check `isParentZeroSized` for this, but I'm not sure we still want / need that here. Either way, it is pretty easy to add in if we do want it.

**Test Project:**
https://utopia.pizza/p/eef4b92b-rain-buffalo/?branch_name=spike-ancestor-metastrategy


editor/src/components/canvas/canvas-strategies/strategies/reparent-metastrategy.tsx
```diff
@@ -106,6 +121,75 @@ export function getApplicableReparentFactories(
...
+
+const getStartingTargetParentsToFilterOut = memoize(getStartingTargetParentsToFilterOutInner, {
+  maxSize: 5,
+  equals: (
+    l: InteractionCanvasState | InteractionSession,
+    r: InteractionCanvasState | InteractionSession,
+  ) => {
+    // We only need to re-calculate the targets to filter if the interaction targets or starting metadata have changed
+    if (isCanvasState(l) && isCanvasState(r)) {
+      // FIXME What about insertion?
```
Rheeseyb, line 181: I've realised that insertion shouldn't be filtering any initial targets. Will fix that now.

Rheeseyb: Fixed

bkrmendy (approved):

enidemi (approved):

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -389,14 +404,9 @@ export function getApplicableControls(
...
 export function isResizableStrategy(canvasStrategy: CanvasStrategy): boolean {
-  switch (canvasStrategy.id) {
-    case 'ABSOLUTE_RESIZE_BOUNDING_BOX':
-    case 'KEYBOARD_ABSOLUTE_RESIZE':
-    case 'FLEX_RESIZE_BASIC':
-      return true
-    default:
-      return false
-  }
+  return ['ABSOLUTE_RESIZE_BOUNDING_BOX', 'KEYBOARD_ABSOLUTE_RESIZE', 'FLEX_RESIZE_BASIC'].some(
```
seanparsons, line 407: This feels a little fragile, I wonder if it would be better to attach a flag to the strategy itself.

editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy-canvas.spec.tsx
```diff
@@ -167,7 +167,7 @@ function reparentElement(
   return finalEditor
 }
 
-describe('Absolute Reparent Strategy', () => {
+xdescribe('Absolute Reparent Strategy', () => {
```
seanparsons, line 170: Will these tests be re-enabled subsequently?

editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    if (targets.length !== 1) {
+      return []
+    }
+
+    const target = targets[0]
+
+    // Avoid children of the storyboard
+    if (
+      EP.isEmptyPath(targets[0]) ||
```
seanparsons, line 33:
```suggestion
      EP.isEmptyPath(target) ||
```

balazsbajorics:
> Finally, I've had to make a change to the running of the SetElementsToRerenderCommand, so that it also includes descendants of the targets. This is because otherwise the selective DOM walker will only run for the provided targets, meaning that if an ancestor strategy is being applied it won't update the live metadata for the original target being dragged (which leads to an incorrect outline control being rendered).

I worry that this has dire performance implications if you drag something that is near the root of a deep hierarchy. Instead of this change, what if you'd just blanked insert a SetElementsToRerenderCommand for the selected element in the ancestor strategy?

editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.tsx
```diff
@@ -0,0 +1,119 @@
...
+          },
+        })),
+      )
+    }
+  }
+}
+
+function applyLevelSuffix(name: string, level: number): string {
+  // FIXME What should we use for the label here?
+  const newSuffix = `(Up ${level})`
```
seanparsons, line 115: This makes me think the name should be some kind of type which describes the level it is at, so we can just add one to the level rather than repeatedly strip and append the suffix.

editor/src/components/canvas/canvas-strategies/strategies/reparent-metastrategy.tsx
```diff
@@ -106,6 +123,80 @@ export function getApplicableReparentFactories(
...
+    canvasState.startingAllElementProps,
+    'allow-missing-bounds',
+    allowSmallerParent,
+  )
+
+  return reparentTargetsToFilter(strictBoundsResult, missingBoundsResult)
+}
+
+function isCanvasState(
+  v: InteractionCanvasState | InteractionSession,
```
seanparsons, line 174:
```suggestion
  canvasState: InteractionCanvasState | InteractionSession,
```

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -500,6 +500,13 @@ export const MetadataUtils = {
     }
     return result
   },
+  getDescendantPaths(
+    elements: ElementInstanceMetadataMap,
+    target: ElementPath,
+  ): Array<ElementPath> {
+    const allPaths = Object.keys(elements).map(EP.fromString)
```
seanparsons, line 507: If I got this right it'll save on the need to call `EP.fromString`:
```suggestion
    const allPaths = Object.values(elements).map(elem => elem.elementPath)
```

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
Rheeseyb: This code was to support resizing via the Ancestor meta strategy, which we have since decided is undesirable, so I'll revert it. I do agree though that maybe this is something that should be handled in a different way in general (either via a flag or perhaps by checking that it was produced by a certain meta strategy)

editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy-canvas.spec.tsx
Rheeseyb: They either need to go in the bin or be replaced with browser tests. One of those two will be happening in a follow up PR

editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.tsx
Rheeseyb: Ha! I think this was the result of some code being shuffled around

editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.tsx
Rheeseyb: Yeah I was wondering this. The name for all strategies could be created by their factory functions. Alternatively though it maybe would be enough for the name to be `${name} (ancestor)`. I think this is something to tweak and test out after merging this PR.

editor/src/components/canvas/canvas-strategies/strategies/reparent-metastrategy.tsx
Rheeseyb: I'm going to go with `value` because we don't know if it is or isn't a canvas state yet!

editor/src/core/model/element-metadata-utils.ts
Rheeseyb: Good shout! We could go even further and combine with the next line using a `mapDropNulls` too, so I'm going to do that as well

editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
```diff
@@ -290,10 +290,10 @@ describe('Strategy Fitness', () => {
 
     const interactionSession: InteractionSession = {
       ...createMouseInteractionForTests(
-        canvasPoint({ x: 0, y: 0 }),
+        canvasPoint({ x: 10, y: 10 }),
```
balazsbajorics, line 293: what is the reason for this change?

editor/src/core/model/element-metadata-utils.ts
Rheeseyb: I've added a new function `mapAndFilter` since we use this pattern a lot

editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
Rheeseyb: This test was previously faking the `startingTargetParentsToFilterOut` as part of creating the `InteractionSession`, but that is no longer applicable as we calculate those via the reparent metastrategy. Because this test also used `0,0` as the drag's starting point, that was causing the `getReparentTargetUnified` to pick up the the canvas as the target to filter out, meaning no reparent strategy was applicable. Moving the starting point inside the actual dragged element ensure the correct target is ignored.

Rheeseyb:
> > Finally, I've had to make a change to the running of the SetElementsToRerenderCommand, so that it also includes descendants of the targets. This is because otherwise the selective DOM walker will only run for the provided targets, meaning that if an ancestor strategy is being applied it won't update the live metadata for the original target being dragged (which leads to an incorrect outline control being rendered).
> 
> I worry that this has dire performance implications if you drag something that is near the root of a deep hierarchy. Instead of this change, what if you'd just blanked insert a SetElementsToRerenderCommand for the selected element in the ancestor strategy?

Yeah I think that sounds like a better approach, since we would only care about the selected element and the target ancestor, and no elements between (or below) them. I'll do that 👍

Rheeseyb:
Didn't mean to close with that comment 😩

editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -103,7 +103,7 @@ export interface InteractionSession {
 
   startedAt: number
 
-  startingTargetParentsToFilterOut: ReparentTargetsToFilter | null
+  startingTargetParentsToFilterOut: ReparentTargetsToFilter | null // FIXME Delete in a follow up PR
```
balazsbajorics, line 106: do we no longer need this?

editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.tsx
```diff
@@ -0,0 +1,115 @@
...
+    // Avoid a cyclic dependency by explicitly passing the other metastrategies when creating the next layer's meta strategy
+    const nextAncestorResult = ancestorMetaStrategy(allOtherStrategies, level + 1)(
+      adjustedCanvasState,
+      interactionSession,
+      customStrategyState,
+    )
+
+    if (nextAncestorResult.length > 0) {
+      // A length of > 0 means that we should be bubbling up to the next ancestor
+      return nextAncestorResult
```
balazsbajorics, line 79: can you move the call of the recursive ancestorMetaStrategy in this if branch to make it lazy?

editor/src/components/canvas/canvas-strategies/strategies/reparent-metastrategy.tsx
```diff
@@ -106,6 +123,80 @@ export function getApplicableReparentFactories(
...
+}
+
+function isCanvasState(
+  value: InteractionCanvasState | InteractionSession,
+): value is InteractionCanvasState {
+  return (value as InteractionCanvasState).startingMetadata != null
+}
+
+const getStartingTargetParentsToFilterOut = memoize(getStartingTargetParentsToFilterOutInner, {
+  maxSize: 5,
```
balazsbajorics, line 180: wouldn't we benefit if we increase this number?

balazsbajorics (approved):

Rheeseyb:
> > Finally, I've had to make a change to the running of the SetElementsToRerenderCommand, so that it also includes descendants of the targets. This is because otherwise the selective DOM walker will only run for the provided targets, meaning that if an ancestor strategy is being applied it won't update the live metadata for the original target being dragged (which leads to an incorrect outline control being rendered).
> 
> I worry that this has dire performance implications if you drag something that is near the root of a deep hierarchy. Instead of this change, what if you'd just blanked insert a SetElementsToRerenderCommand for the selected element in the ancestor strategy?

Ok I have now almost exactly this, but this ended up being a lot trickier than I had expected since that command overwrites the existing value, and sometimes we apply that command via the `updateFunctionCommand`, meaning that attempting to merge all instances of that command in the result becomes very brittle, and the command couldn't be changed to append to the existing value since the default starting value is `'render-all-elements'`. To deal with this I've added another command `AppendElementsToRerenderCommand`, which will use the `$apply` patch function to append to the end of the existing `elementsToRerender`.

editor/src/components/canvas/canvas-strategies/interaction-state.ts
Rheeseyb: Nope!

editor/src/components/canvas/canvas-strategies/strategies/reparent-metastrategy.tsx
Rheeseyb: It has pros and cons. This is enough for someone to drag an element that affects an ancestor 5 levels up. I'm not sure how likely that is in practice tbh. We could certainly increase this to 10 or more.

Rheeseyb: I've increased it to 10

editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy-canvas.spec.tsx
gbalint: I already deleted this test, because a same was implemented as a browser test already

PR #2724, opened 2022-10-25
Disable performance test feature flags in staging by default

Rheeseyb:
**Problem:**
We rely on some sample projects in staging when developing new features, but it is extremely easy to accidentally overwrite a project by accident by clicking one of the performance test triggers

**Fix:**
Disable those feature flags by default in staging, meaning the buttons won't be rendered unless they are explicitly enabled


enidemi (approved):

Rheeseyb:
Closing this because it's not necessary anymore

PR #2730, opened 2022-10-26
Only show padding controls when padding is already set

bkrmendy:
## Problem:
We currently show padding controls for elements that can have a size and have children that are not positioned absolute or sticky. While this filters all elements to some degree, the padding controls still add visual clutter when shown on an element that doesn't have a padding set in the code (meaning that it might never have padding applied).

## Fix:
check whether padding is already set in some form on the selected element, and only show the padding controls if it is.

## Quirks
This PR uses `element.specialSizeMeasurements.padding` to check if padding is set on an element, but this brings a tradeoff with it:
- on the plus side, using element.specialSizeMeasurements.padding tells us the padding even if it's set in units other than px, or it's set from css
- on the minus side, padding values coming from element.specialSizeMeasurements.padding are 0 if the padding is not set, so to filter this out, we can only show the padding controls if the padding value is > 0. This edge case where padding is set to 0 but the controls are not shown.
The upsides outweigh the downsides here, but I still wanted to call out the behaviour, so that the edge case above will be a known bug

## Also
Also adding a further idea from the discussion from [discord[(https://discord.com/channels/647102474519576589/1026865812864249926/1034858760230944779):
> if they are dimensioned and empty, show padding controls always even if no padding

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
@@ -21,7 +21,52 @@ import {
 } from '../../ui-jsx.test-utils'
 import { SetPaddingStrategyName } from './set-padding-strategy'
 
+const edgePieces: Array<EdgePiece> = ['top', 'bottom', 'left', 'right']
+
 describe('Padding resize strategy', () => {
+  it('Padding resize handle is not present for elements that have no padding set', async () => {
+    const editor = await renderTestEditorWithCode(
+      makeTestProjectCodeWithSnippet(`<div
+      data-testid='mydiv'
+      style={{
+        backgroundColor: '#0091FFAA',
+        position: 'absolute',
+        left: 28,
+        top: 28,
+        width: 612,
+        height: 461,
+      }}
+      data-uid='24a'
+    >
+      <div
+        style={{
+          backgroundColor: '#0091FFAA',
+          width: '100%',
+          height: '100%',
+        }}
+        data-uid='002'
+      />
+    </div>`),
+      'await-first-dom-report',
+    )
+
+    const canvasControlsLayer = editor.renderedDOM.getByTestId(CanvasControlsContainerID)
+    const div = editor.renderedDOM.getByTestId('mydiv')
+    const divBounds = div.getBoundingClientRect()
+    const divCorner = {
+      x: divBounds.x + 5,
+      y: divBounds.y + 4,
+    }
+
+    mouseClickAtPoint(canvasControlsLayer, divCorner, { modifiers: cmdModifier })
+
+    const paddingControls = edgePieces.flatMap((edge) =>
+      editor.renderedDOM.queryAllByTestId(paddingControlTestId(edge)),
+    )
+
+    expect(paddingControls).toEqual([])
+  })
+
   it('Padding resize is present', async () => {
     const editor = await renderTestEditorWithCode(
       makeTestProjectCodeWithStringPaddingValues(
@@ -45,7 +90,6 @@ describe('Padding resize strategy', () => {
 
     mouseClickAtPoint(canvasControlsLayer, divCorner, { modifiers: cmdModifier })
 
-    const edgePieces: Array<EdgePiece> = ['top', 'bottom', 'left', 'right']
     edgePieces.forEach((edge) => {
       const paddingControlOuter = editor.renderedDOM.getByTestId(paddingControlTestId(edge))
       expect(paddingControlOuter).toBeTruthy()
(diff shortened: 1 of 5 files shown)


editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
```diff
@@ -21,7 +21,88 @@ import {
 } from '../../ui-jsx.test-utils'
 import { SetPaddingStrategyName } from './set-padding-strategy'
 
+const edgePieces: Array<EdgePiece> = ['top', 'bottom', 'left', 'right']
```
balazsbajorics, line 24: I _think_ we have a convention to make these global consts capitalized, ie EdgePieces

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
```diff
@@ -21,7 +21,88 @@ import {
...
+
+    const canvasControlsLayer = editor.renderedDOM.getByTestId(CanvasControlsContainerID)
+    const div = editor.renderedDOM.getByTestId('mydiv')
+    const divBounds = div.getBoundingClientRect()
+    const divCorner = {
+      x: divBounds.x + 50,
+      y: divBounds.y + 40,
+    }
+
+    mouseClickAtPoint(canvasControlsLayer, divCorner, { modifiers: cmdModifier })
```
balazsbajorics, line 61: as far as I understand, your only goal is to select the element, you can achieve that by dispatching 

```
await renderResult.dispatch([selectComponents([targetPath], false)], false)
```

this reduces the potential flakyness of the test, since now you are testing two things: click to select, _and_ padding controls

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
```diff
@@ -174,19 +176,51 @@ function supportsPaddingControls(metadata: ElementInstanceMetadataMap, path: Ele
...
+  const paddingProps = (p: keyof CSSPadding) => PP.create(['style', p])
+
+  return (
+    jsxAttributeExists(PP.create(['style', 'padding'])) ||
+    jsxAttributeExists(paddingProps('paddingBottom')) ||
+    jsxAttributeExists(paddingProps('paddingTop')) ||
+    jsxAttributeExists(paddingProps('paddingLeft')) ||
+    jsxAttributeExists(paddingProps('paddingRight'))
+  )
+}
```
balazsbajorics, line 222: there's element.specialSizeMeasurements.padding that should give you 4 values that come from computedStyle – meaning we could _show_ the padding controls even if they are set from CSS or wherever.

bkrmendy: I tried this while working the PR, and it seems like that padding from specialSizeMeasurements is 0 when there's no padding set in the CSS.

We can make this work by only showing padding controls if any of the padding values is >0, but this leaves us with the slim edge case where all paddings are set to 0, but the still controls aren't shown. I felt this to go against the original spirit of the PR (to make props manipulable visually), so I went for the more verbose but more accurate solution.

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
bkrmendy: thanks, updated!

bkrmendy: thanks, updated!

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
Rheeseyb: I'd say keep this as is because the browser tests are closer to integration tests than unit tests, so we want this to be closer to real life use

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
```diff
@@ -21,7 +21,88 @@ import {
...
+
+    const paddingControls = edgePieces.flatMap((edge) => [
+      // ...editor.renderedDOM.queryAllByTestId(paddingControlTestId(edge)),
```
Rheeseyb, line 64: Delete me?

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
bkrmendy: I just updated it 🤦‍♂️

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
bkrmendy: I readded it instead, it makes for a more comprehensive check

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
bkrmendy: I refactored the strategy along your suggestion, and I added the tradeoff in the PR description for posterity

seanparsons (approved):

PR #2733, opened 2022-10-27
Github MUD indicators

### LINKED ISSUE #2732: Github MUD indicators

ruggi:
Fixes #2732 

**Problem:**

After working off of a Git branch, we should have indicators in the UTFB and the menu tiles showing which files diverged from the upstream ones, so that it's clear what changed when changes are commished.

**Fix:**

This PR achieves three goals:

1. It displays MUD indicators in the UTFB (note: placeholder colors).
Screen Shot 2022-10-27 at 12 44 08 PM
2. It shows a badge on the octocat menu tile with the number of total changed files (added, changed, removed).
Screen Shot 2022-10-27 at 12 45 28 PM
3. It shows a list of the changed files in the Github pane above the `Save to Github` button so the user knows what's about to be saved.
Screen Shot 2022-10-27 at 12 52 16 PM

**Notes**
- The changed files section in the Github pane is absolutely temporary and purely utilitarian while we finalize/work on the designs but it should still be useful in the meantime
- The colors for the Git statuses are just placeholder values and are not representing anything other than giving a visual hint. We should definitely update them (as noted in the code) once we have finalized the design (including checking for accessibility - e.g. making sure those colors are colorblind-friendly, etc.)
- The Github checksums are stored in the persistent model so that they will remain in place between project reloads (this should be useful in the future as well to keep pairing between Utopia projects and Github repos/branches, although most likely we'll end up shuffling around the fields a bit)
- The octocat badge uses a simple cap so that for a _huge_ number of file changes it will only show the cap. This may be a premature optimization considering the upcoming design work, but it's probably a nice touch to at least remember it's something we might keep into account.
badge


editor/src/components/assets.ts
```diff
@@ -40,6 +41,50 @@ export function getAllProjectAssetFiles(
...
+export function deriveGithubFileChanges(
+  projectChecksums: GithubChecksums,
+  githubChecksums: GithubChecksums | null,
+): GithubFileChanges | null {
+  if (githubChecksums == null) {
+    return null
+  }
+
+  const projectFiles = Object.keys(projectChecksums)
+  const githubFiles = Object.keys(githubChecksums)
```
bkrmendy, line 73: could be turned into a Set to speed up the includes calls?

editor/src/components/menubar/menubar.tsx
```diff
@@ -232,6 +270,26 @@ export const Menubar = React.memo(() => {
...
+  )
+
+  const githubFileChangesCountLabel = React.useMemo(() => {
+    if (githubFileChangesCount <= 0) {
+      return undefined
+    }
+    if (githubFileChangesCount < GITHUB_FILE_CHANGES_BADGE_LIMIT) {
+      return `${githubFileChangesCount}`
+    }
+    return `${GITHUB_FILE_CHANGES_BADGE_LIMIT}+`
```
bkrmendy, line 290: chef's kiss 👌

bkrmendy (approved):

editor/src/components/assets.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/2733/commits/6f95b442e8a942bdc7aff83b506c23a09936bbe6

editor/package.json
```diff
@@ -259,6 +260,7 @@
     "@types/react-window": "1.8.5",
     "@types/scheduler": "0.16.1",
     "@types/semver": "7.3.4",
+    "@types/sha.js": "^2.4.0",
```
Rheeseyb, line 263: This one too please :D

editor/src/components/assets.ts
```diff
@@ -40,6 +41,65 @@ export function getAllProjectAssetFiles(
...
+  githubChecksums: GithubChecksums | null,
+): GithubFileChanges | null {
+  if (githubChecksums == null) {
+    return null
+  }
+
+  const projectFiles = new Set(Object.keys(projectChecksums))
+  const githubFiles = new Set(Object.keys(githubChecksums))
+
+  const untracked: string[] = []
```
Rheeseyb, line 75: This is a very minor one, but we have a convention where we use the `let` keyword to signify that a variable is going to be mutated somewhere down the line. In this case it's extremely minor since it's obvious to see that, but our use of `let` makes it even more immediately obvious.

editor/src/components/assets.ts
```diff
@@ -40,6 +41,65 @@ export function getAllProjectAssetFiles(
...
+      untracked.push(f)
+    }
+  })
+
+  const modified: string[] = []
+  projectFiles.forEach((f) => {
+    if (githubFiles.has(f) && githubChecksums[f] !== projectChecksums[f]) {
+      modified.push(f)
+    }
+  })
```
Rheeseyb, line 87: This could be rolled into the previous for loop to save a bit of time on large projects

editor/src/components/editor/actions/action-creators.ts
```diff
@@ -1441,6 +1443,45 @@ export function updateGithubOperations(
   }
 }
 
+export type GithubChecksums = { [filename: string]: string } // key = filename, value = sha1 hash of the file
```
Rheeseyb, line 1446: Thanks for adding the clarifying comment here!

editor/src/components/filebrowser/filebrowser.tsx
```diff
@@ -108,7 +120,21 @@ function collectFileBrowserItems(
       }
     }
   })
-  return fileBrowserItems
+  return fileBrowserItems.map((f) => {
```
Rheeseyb, line 123: It looks like we could roll this into the above `walkContentsTree` call (unless I've missed something)?

editor/src/components/filebrowser/fileitem.tsx
```diff
@@ -205,6 +205,20 @@ const isFile = (fileBrowserItem: FileBrowserItemInfo) => {
   return fileBrowserItem.type === 'file'
 }
 
+export const getGithubFileStatusColor = (type: GithubFileStatus): string => {
+  // NOTE: these are placeholder colors, we should update them once we finalize the design
```
Rheeseyb, line 209: Don't worry about it for the sake of this PR since the design is still a WIP, but we have colour themes in [`theme.ts`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/uuiui/styles/theme.ts), and a hook [`useColorTheme`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/uuiui/styles/theme.ts#L314) to access the current theme in functional components.

editor/src/components/filebrowser/fileitem.tsx
```diff
@@ -278,6 +292,32 @@ class FileBrowserItemInner extends React.PureComponent<
 
   toggleCollapse = () => this.props.toggleCollapse(this.props.path)
 
+  renderGithubStatus = () => {
+    if (this.props.githubStatus != undefined) {
+      const c = this.props.githubStatus.at(0)?.toUpperCase()
```
Rheeseyb, line 297: Super minor, but can you change the name of this const to `statusChar` please? I had to do a fair amount of looking about to figure out what this `c` actually represents to find out it is the first character of the git status

editor/src/core/shared/string-utils.ts
```diff
@@ -2,6 +2,10 @@ export function capitalize(value: string): string {
   return value.charAt(0).toUpperCase() + value.slice(1)
 }
 
+export function pluralize(value: string, count: number): string {
+  return value + (count !== 1 ? 's' : '')
```
Rheeseyb, line 6: Oh I wish 😁

Rheeseyb (approved):

editor/src/components/assets.ts
```diff
@@ -40,6 +41,65 @@ export function getAllProjectAssetFiles(
...
+  const githubFiles = new Set(Object.keys(githubChecksums))
+
+  const untracked: string[] = []
+  projectFiles.forEach((f) => {
+    if (!githubFiles.has(f)) {
+      untracked.push(f)
+    }
+  })
+
+  const modified: string[] = []
```
balazsbajorics, line 82: nit: we use the `Array<T>` form instead of `T[]`

editor/src/components/filebrowser/fileitem.tsx
ruggi: noice 👌

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -2416,6 +2417,53 @@ export function ProjectContentTreeRootKeepDeepEquality(): KeepDeepEqualityCall<P
...
+  if (oldAttribute == null) {
+    return keepDeepEqualityResult(newAttribute, false)
+  }
+  if (newAttribute == null) {
+    return keepDeepEqualityResult(oldAttribute, false)
+  }
+  return objectDeepEquality(StringKeepDeepEquality)(oldAttribute, newAttribute)
+}
+
+const GithubFileChangesKeepDeepEquality: KeepDeepEqualityCall<GithubFileChanges | null> = (
```
balazsbajorics, line 2439: I think this is now unused! if so, to the bin with it!

ruggi: oh, good catch

ruggi: https://github.com/concrete-utopia/utopia/pull/2733/commits/c4c4672691ab949ef4917343733281da9baf95e4 (and https://github.com/concrete-utopia/utopia/pull/2733/commits/bfdc39cfcdb9e0436673ec3f6f34ea6df387dad8 which moves the selectors and `GithubFileChanges` to the GH helpers file)

balazsbajorics (approved):

Issue #2735, opened 2022-10-27
Image upload fails when image is dropped onto a scene from Finder


bkrmendy:
## Describe the bug
When an image is dropped on a scene from Finder, image upload fails.

## To Reproduce
1. Drag an image from Finder
2. onto a Scene

Result: Image upload fails

## Expected behaviour
Image upload succeeds

## Probable root cause / technical background
Currently images are inserted with a two-step process:
1. the base64 content of the image is inlined into `src`
2. the image is uploaded to s3
3. once the image has uploaded, the `src` is updated to the relative path of the image resource

To determine the element path of the newly inserted image, a function called `updateImageSrcsActions` scans through `projectContents` to find `img` tags that have to have their uids updated (see https://github.com/concrete-utopia/utopia/blob/e50ca378700ce231f10093affd2f71f74aee515f/editor/src/templates/image-drop.tsx#L319-L348 for full definition). However, when dispatch is called with the `src` update actions, the action executor errors out with the explanation that the element at the given path was not found.

PR #2736, opened 2022-10-28
Cannot drop image into component file

bkrmendy:
Fixes https://github.com/concrete-utopia/utopia/issues/2735

## Problem:
See https://github.com/concrete-utopia/utopia/issues/2735 for full description and technical background

## Fix:
Get the element paths for the update actions from `allElementPaths` instead of `projectContents`

### DIFF
--- a/editor/src/templates/image-drop.tsx
+++ b/editor/src/templates/image-drop.tsx
@@ -14,32 +14,18 @@ import { ElementPath } from '../core/shared/project-file-types'
 import { fastForEach } from '../core/shared/utils'
 import { createDirectInsertImageActions, parseClipboardData } from '../utils/clipboard'
 import { imagePathURL } from '../common/server'
-import { ProjectContentTreeRoot, walkContentsTreeForParseSuccess } from '../components/assets'
+import { ProjectContentTreeRoot } from '../components/assets'
 import { createJsxImage, getFrameAndMultiplier } from '../components/images'
 import { generateUidWithExistingComponentsAndExtraUids } from '../core/model/element-template-utils'
 import React from 'react'
 import { CanvasPositions } from '../components/canvas/canvas-types'
 import { AllElementProps, EditorState, notDragging } from '../components/editor/store/editor-state'
-import {
-  getUtopiaJSXComponentsFromSuccess,
-  imageFile,
-  uniqueProjectContentID,
-} from '../core/model/project-file-utils'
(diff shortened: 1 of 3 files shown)


editor/src/utils/utils.ts
```diff
@@ -935,6 +935,18 @@ function deduplicateBy<T>(key: (t: T) => string, ts: Array<T>): Array<T> {
   return results
 }
 
+function groupBy<Element>(
```
Rheeseyb, line 938: Can you move this into [`array-utils.ts`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/array-utils.ts) please

editor/src/utils/utils.ts
```diff
@@ -935,6 +935,18 @@ function deduplicateBy<T>(key: (t: T) => string, ts: Array<T>): Array<T> {
...
+  elements: Array<Element>,
+  key: (_: Element) => string,
+): { [key: string]: Element } {
+  let result: { [key: string]: Element } = {}
+  for (const element of elements) {
```
Rheeseyb, line 943: Please use `fastForEach` here

editor/src/templates/image-drop.tsx
```diff
@@ -319,33 +321,23 @@ interface SrcSubstitutionData {
...
-                PP.create(['src']),
-                jsxAttributeValue(subsititutionPath, emptyComments),
-              ),
-            )
-          }
-        }
-      }
-    })
-  })
+  for (const [path, props] of Object.entries(allElementProps)) {
```
Rheeseyb, line 329: This can use `fastForEach` as well

Rheeseyb (approved):

editor/src/utils/utils.ts
bkrmendy: done!

editor/src/utils/utils.ts
bkrmendy: done!

editor/src/templates/image-drop.tsx
bkrmendy: done!

editor/src/utils/utils.ts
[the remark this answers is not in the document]
bkrmendy: ended up not using the function at all

balazsbajorics (approved):

PR #2737, opened 2022-10-28
Feature/canvas reorder 1d row reverse

enidemi:
**Problem:**
There are inconsistencies about how flow reorder and flex reorder becomes the active strategy when dragging an element. For example: flow is not enabled without siblings, flex is allowed etc.

**Fix:**
Only allow flex reorder for non-wrapping rows/columns and with this change the frame calculations are extended with reverse flex layout and rtl direction too (rtl reverses the layout for inline flow).
Next PR will add the reorder slider for wrapping flex layouts.

**Commit Details:**
- extended metadata with text direction, unfortunately the computed style was late on select+drag interactions
- update `areAllSiblingsInOneDimension` to work with flex or flow, this is one bigger function now but it's easier not to forget to change both flex and flow in the future
- there were reverse tests for flex, added rtl reversed flow layout reorder test
- snapshot in tests updated with new style prop in special size measurements.


editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
```diff
@@ -27,43 +28,70 @@ export function isValidFlowReorderTarget(
   }
 }
 
-export function areAllSiblingsInOneDimension(
+export function areAllSiblingsInOneDimensionFlexOrFlow(
   target: ElementPath,
   metadata: ElementInstanceMetadataMap,
+  targetLayout: 'flex' | 'flow',
```
balazsbajorics, line 34: I wonder if this should just be based on the metadata for `target`, since you are using the metadata anyways to get subsequent information, such as flexDirection

balazsbajorics (approved):

Rheeseyb (approved):

PR #2738, opened 2022-10-28
Reparent to 1D flow is similar to flex

gbalint:
Fixes https://github.com/concrete-utopia/utopia/issues/2741

**Problem:**
Allow reordering in reparent to flow when flow is 1 dimensional.

**Fix:**
We modified reparent to flex in a way that instead of checking for flex properties we also allow reordering for one dimensional flow containers.

**More details**:

Static layouts have a few important properties:
- is the layout flex or flow
- are the child elements layouted in one dimension
- is the direction row or column
- is the direction forward or reverse
- are there other absolute children
- how many flex/flow children are there

These properties all affect whether we can detect a correct position index for the reparent target in the new flex/flow parent.
For flex layouts we kept the old behavior and we always allow reordering, but we know that this does not work well in cases when the children are not on one dimension. (todo for later, flex reparent was not the scope of this PR)

For flow layouts we wanted to allow reordering only in cases when we can reliably detect the proper position index.
For this:
- we need to detect the row/column layout for flow containers: for flex this is a property on the container, but for flow we need to check the display property of the children. (This means flow containers have to have already existing children to detect these properties.)
A flow layout is a column by default, it is a row when all kids have `display: inline*` properties. We need to make sure to add these properties to newly reparented/inserted elements. 

- we need to detect the forward/reverse direction for flow containers: it is reverse when the container has `textDirection: rtl` property

- we need to check if the elements are on one dimension, and just fallback to reparent to the last position when they are not.

**Todo:**

We needed to cherry-pick and merge some of the functionality from flex reparent and some from flow reorder. Code sharing between reparent and reorder strategies is not the strength of our code, and it was quite hard to decide how deep should we go in trying to make this clean and non-redundant.
I think we stopped at a good compromise, but probably a followup, refactor-only PR is needed to create a nice API to gather information about static containers, their children and their properties.


editor/src/components/canvas/canvas-strategies/strategies/flex-reparent-to-flow-strategy.spec.browser2.tsx
```diff
@@ -0,0 +1,637 @@
...
+          data-uid='flexchild4'
+          data-testid='flexchild4'
+        />
+      </div>
+    </div>
+  `),
+    )
+  })
+
+  // TODO: enable when insertion to zero position is fixed
```
enidemi, line 448: what is this TODO about?

enidemi (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
```diff
@@ -367,22 +369,29 @@ export function getReparentTargetUnified(
...
   // if the mouse is over the canvas, return the canvas root as the target path
 
-  const flexElementsUnderPoint = [...filteredElementsUnderPoint]
-    .reverse()
-    .filter((element) =>
-      MetadataUtils.isFlexLayoutedContainer(
-        MetadataUtils.findElementByElementPath(metadata, element),
-      ),
-    )
+  const staticContainersUnderPoint = mapDropNulls((element) => {
```
enidemi, line 372: I'm a bit confused by this name, without reading the direction helper I would expect it to handle only flow layouts

editor/src/components/canvas/canvas-strategies/strategies/flex-reparent-to-flow-strategy.spec.browser2.tsx
gbalint: Sorry, outdated comment, the test is already enabled under it

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
gbalint: I tried to use the terminology that static == flex or flow
Isn't that a good idea?

enidemi: I think my confusion comes from `position: static`, but I don't have any good suggestion :(

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
```diff
@@ -34,51 +35,97 @@ export function areAllSiblingsInOneDimensionFlexOrFlow(
...
-    const shouldReverse = flexDirection?.includes('reverse') ?? false
-    const frames = mapDropNulls((sibling) => {
+  return staticContainerDirections(EP.parentPath(target), metadata) !== 'non-1d-static'
+}
+
+export function staticContainerDirections(
+  container: ElementPath,
+  metadata: ElementInstanceMetadataMap,
+):
+  | {
```
seanparsons, line 50: Could we separate out this into a type somewhere?

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
```diff
@@ -34,51 +35,97 @@ export function areAllSiblingsInOneDimensionFlexOrFlow(
...
+  const children = MetadataUtils.getChildren(metadata, container)
+  if (containerElement == null) {
+    return 'non-1d-static'
+  }
+
+  const layoutSystem = containerElement.specialSizeMeasurements.layoutSystemForChildren
+
+  if (layoutSystem === 'flex') {
+    const flexDirection = MetadataUtils.getSimpleFlexDirection(containerElement)
+    const targetDirection = flexDirection.direction === 'row' ? 'horizontal' : 'vertical' // TODO unify row and horizontal types
```
seanparsons, line 66: Future work TODO?

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
```diff
@@ -954,12 +976,24 @@ export function getAbsoluteReparentPropertyChanges(
 export function getStaticReparentPropertyChanges(
   newPath: ElementPath,
   targetOriginalStylePosition: CSSPosition | null,
+  targetOriginalDisplayProp: string | null,
+  convertToInline: 'row' | 'column' | 'do-not-convert',
```
seanparsons, line 980: This type looks to be in a bunch of places, perhaps it should be a separately defined type?

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
```diff
@@ -987,6 +1022,51 @@ export function getReparentPropertyChanges(
...
+        convertDisplayInline,
+      )
+  }
+}
+
+function getDirectionsForValidStaticReparentTarget(
+  path: ElementPath,
+  metadata: ElementInstanceMetadataMap,
+):
+  | {
```
seanparsons, line 1050: Is this the same return type that `staticContainerDirections` has?

seanparsons (approved):
There's quite a few TODOs in this, do any of those really need doing before this is merged?

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
balazsbajorics: future work!

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
balazsbajorics: we deliberately did not want to make this an explicit type, because we wanted to avoid looking like this is part of an API or a sensible collection of values. it's mostly necessity-driven that these data must come from one function call

seanparsons: https://github.com/concrete-utopia/utopia/pull/2738#discussion_r1018958369

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
```diff
@@ -367,22 +369,29 @@ export function getReparentTargetUnified(
...
   // if the mouse is over the canvas, return the canvas root as the target path
 
-  const flexElementsUnderPoint = [...filteredElementsUnderPoint]
-    .reverse()
-    .filter((element) =>
-      MetadataUtils.isFlexLayoutedContainer(
-        MetadataUtils.findElementByElementPath(metadata, element),
-      ),
-    )
+  const singleAxisAutoLayotuContainersUnderPoint = mapDropNulls((element) => {
```
enidemi, line 372: typo

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
balazsbajorics: fixed!

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
balazsbajorics: fixed!

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
balazsbajorics: renamed!

editor/src/components/canvas/canvas-strategies/strategies/reparent-strategy-helpers.ts
balazsbajorics: I decided to leave this one type alone, there's a follow up task to rename this to horizontal | vertical anyways

PR #2743, opened 2022-10-31
Allow reverting file changes and go back to the branch content

### LINKED ISSUE #2742: Reverting file changes for GH integration

ruggi:
Fixes #2742 

**Problem:**

When a project uses content from a Github branch it should be possible to amend changes and go back to the branch original content.

**Fix:**

This PR does a bunch of things:
1. It pins the loaded branch name to the project's github settings, to it's possible to know from which branch the project descends. The Github pane now also displays the current branch name and an indicator (right-pointing arrow) with the same meaning in the branches list. Again, purely temporary and utilitarian until we work on the final designs.
current branch

2. When the branch is loaded and pinned, its contents are also stored behind the scenes. This is to allow to perform individual (or global) change reverts, without having to fetch the contents from upstream again. This could be also done asynchronously by pulling a blob by sha, but I _think_ the overhead is too much (at least until we have evidence of the opposite).

3. Files can be reverted individually from either the UTFB with a right click or via the Github pane. All changes can be reverted in bulk from the Github pane as well.
revert

4. As a bonus point, branches get refreshed when the pinned branch is updated, so for example `Save to Github` triggers a branch refresh. Note: this is a bit cumbersome at the moment since we're always refreshing branches synchronously, but it'll go away once we move that part to the background.


editor/src/components/navigator/left-pane.tsx
```diff
@@ -1376,21 +1416,114 @@ const GithubPane = React.memo(() => {
...
+        textOverflow: 'ellipsis',
+      }}
+      title={title}
+    >
+      {children}
+    </div>
+  )
+}
+
+const handleClickRevertAllFiles = (dispatch: EditorDispatch) => (e: React.MouseEvent) => {
```
seanparsons, line 1443: These should be defined in the component where they're used wrapped in `React.useCallback` to avoid lots of instances being created as a result of re-renders.

editor/src/core/shared/github.ts
```diff
@@ -319,6 +328,15 @@ export const githubFileChangesSelector = createSelector(
   },
 )
 
+export type GithubFileStatus =
+  | 'modified'
+  | 'deleted'
+  | 'untracked'
+  | 'added' // unused, keeping for future reference
```
seanparsons, line 335: I would err on the side of not including these if they're not being used as it's easy for them to end up living forever in the code.

editor/src/core/shared/github.ts
```diff
@@ -348,3 +366,63 @@ export function githubFileChangesEquals(
...
+        const branchContentsTree = treeToContents(branchContents)
+        newContents[filename] = branchContentsTree[filename]
+        const newTree = contentsToTree(newContents)
+        dispatch([updateProjectContents(newTree)], 'everyone')
+        break
+      default:
+        break
+    }
+  }
+}
```
Rheeseyb, line 428: Rather than these two functions taking the `EditorDispatch` as a param and producing a side effect, please can they instead return the actions that need to be dispatched, so that the code calling them can then dispatch those actions.

editor/src/components/navigator/left-pane.tsx
```diff
@@ -1376,21 +1416,114 @@ const GithubPane = React.memo(() => {
...
+
   return (
-    <div style={{ color: getGithubFileStatusColor(type), marginTop: 4, marginBottom: 4 }}>
-      <div>
-        {capitalize(type)} files ({files.length})
-      </div>
-      {files.map((f) => (
-        <div key={f} style={{ marginLeft: 8 }}>
-          &rarr; {f}
+    // Note: this is completely temporary until we finalize the design
```
Rheeseyb, line 1501: I don't think there's any need to include comments like this, just as long as we're tracking the UI work

editor/src/core/shared/github.ts
```diff
@@ -348,3 +366,63 @@ export function githubFileChangesEquals(
...
+): void {
+  if (branchContents != null) {
+    switch (status) {
+      case 'untracked':
+        dispatch([deleteFile(filename)], 'everyone')
+        break
+      case 'deleted':
+      case 'modified':
+        const newContents = treeToContents(projectContents)
+        const branchContentsTree = treeToContents(branchContents)
```
seanparsons, line 419: Can use `getContentsTreeFileFromString` to retrieve a file rather than converting the structures and `addFileToProjectContents` to add it into another.

editor/src/components/editor/actions/migrations/migrations.ts
```diff
@@ -361,3 +362,23 @@ function migrateFromVersion8(
     }
   }
 }
+
+function migrateFromVersion9(
+  persistentModel: PersistentModel,
+): PersistentModel & { projectVersion: 10 } {
+  if (persistentModel.projectVersion != null && persistentModel.projectVersion !== 8) {
```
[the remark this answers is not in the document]
ruggi: https://github.com/concrete-utopia/utopia/pull/2743/commits/8fcada15df5530b206c12cbc58ebf99985e24bc9

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -2420,6 +2426,37 @@ export function ProjectContentTreeRootKeepDeepEquality(): KeepDeepEqualityCall<P
   return objectDeepEquality(ProjectContentsTreeKeepDeepEquality())
 }
 
+const NullableGithubFileStatusKeepDeepEquality: KeepDeepEqualityCall<GithubFileStatus | null> = (
```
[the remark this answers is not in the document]
ruggi: https://github.com/concrete-utopia/utopia/pull/2743/commits/b9a1aea5937de2c822ca70518bca34e903fc4052

editor/src/components/navigator/left-pane.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/2743/commits/7f64023515b2de51952acdd1a0e80747d9c48019

editor/src/core/shared/github.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/2743/commits/ac58fb13f1b3a147153f72f414a99425d28cfdf1

editor/src/components/navigator/left-pane.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/2743/commits/b231d6f0df69151c4a2743c63b39233dcb459924

editor/src/core/shared/github.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/2743/commits/96b748b746c7283d34266309049dc424841ea20c

seanparsons (approved):

Rheeseyb (approved):

PR #2744, opened 2022-10-31
Github repository filtering

seanparsons:
**Problem:**
The input field for choosing the Github repository of interest is a little clunky for a user, as they need to go out of their way to input the exact repository they want.

**Fix:**
Below the input field, we now have a dynamic list of the repositories that the user can select which they want to import from without having to fill in the entire input field.

**Commit Details:**
- Added `react-timeago` and its `@types` dependency.
- Added a `loadRepositories` Github operation.
- Created a new `RepositoryRow` component which handles the individual elements within the repository listing.
- Added `RepositoryListing` component which enumerates the repositories, filters and renders them.
- Refactored some of the hooks in the `GithubPane` component, removing some and making others more sensible for their needs.
- Added many more fields to `RepositoryEntry` for all the information needed in the repository listing.
- `getUsersPublicGithubRepositories` dispatches the actions for starting and completing the Github operation specific to it.
- Added extra fields to `UsersRepository` and renamed it from `UsersPublicRepository`.


editor/src/components/navigator/left-pane.tsx
```diff
@@ -763,6 +768,249 @@ const GithubSpinner = () => {
...
+          background: colorTheme.buttonBackground.value,
+          boxShadow: 'none',
+          border: 'none',
+          height: 22,
+          color: colorTheme.inlineButtonColor.value,
+          borderRadius: 2,
+        }}
+        onMouseUp={importRepository}
+      >
+        Import
```
ruggi, line 854: for whatever reason beyond me, this does not use `cursor: pointer` when hovered 🧐

editor/src/components/navigator/left-pane.tsx
```diff
@@ -763,6 +768,249 @@ const GithubSpinner = () => {
...
+          backgroundRepeat: 'no-repeat',
+        }}
+      />
+      <div style={{ flexGrow: 1 }}>
+        <span style={{ fontWeight: 600, textOverflow: 'ellipsis' }}>
+          {props.name ?? props.fullName}
+        </span>{' '}
+        <br />
+        <span style={{ opacity: 0.5 }}>
+          {props.private ? 'private' : 'public'} · <TimeAgo date={props.updatedAt} />
```
ruggi, line 839: when a repository is not found, `private` is shown and the `TimeAgo` will start going up from `0` every second. e.g.:
Screen Shot 2022-10-31 at 4 48 48 PM because of how the `additionalEntry` item is added.

ruggi: note: with the current implementation it's impossible to see (and so import) public repos which are not owned by the GH user. Imho it would be nice to be able to do this (but of course `Save to Github` should be disabled if write access is not available.

editor/src/components/navigator/left-pane.tsx
```diff
@@ -763,6 +768,249 @@ const GithubSpinner = () => {
...
+              filteredRepositoriesWithSpecialCases == null
+                ? 'Loading repositories...'
+                : 'owner/repository'
+            }
+            onChange={onInputChangeTargetRepository}
+            list={'repositories-list'}
+            id={'repositories-input'}
+            testId={'repositories-input'}
+            name={'repositories-input'}
+            value={targetRepository}
```
ruggi, line 979: watch out - this does not update when the repo is imported for the first time

ruggi (commented):
This looks very nice! 👌 
Left a couple of suggestions

editor/src/components/navigator/left-pane.tsx
seanparsons: Yeah, I think it's going to be better to hide the time for that fake entry.

editor/src/components/navigator/left-pane.tsx
seanparsons: I smell a dependency being missed out.

editor/src/components/navigator/left-pane.tsx
maltenuhn: I'm thinking we might have a global override for that somewhere, meaning `button` by default does not use the pointer cursor. It's one of those "application / editor UX" vs "web UX" clashes.

editor/src/components/navigator/left-pane.tsx
seanparsons: Since this is for "connecting" to a repo, it makes more sense that these are repositories that you can modify.

editor/src/components/navigator/left-pane.tsx
seanparsons: Fixed in 1a05bc536.

editor/src/components/navigator/left-pane.tsx
seanparsons: Fixed in 1a05bc536.

editor/src/components/navigator/left-pane.tsx
seanparsons: Fixed in 1a05bc536.

Rheeseyb (approved):

editor/src/components/navigator/left-pane.tsx
```diff
@@ -763,6 +768,259 @@ const GithubSpinner = () => {
...
+          color: props.importPermitted
+            ? colorTheme.inlineButtonColor.value
+            : colorTheme.inlineButtonColor.shade(50).value,
+          borderRadius: 2,
+          cursor: 'pointer',
+        }}
+        disabled={!props.importPermitted}
+        onMouseUp={importRepository}
+      >
+        Import
```
ruggi, line 864: One last thing: we should make sure that this is:
1. disabled while there are ongoing github operations
2. showing the spinner during the import like the other buttons

This requires some minor changes, here's a suggestion:

In `left-pane.tsx`:
```ts
const RepositoryRow = (props: RepositoryRowProps) => {
  // ...
  const githubOperations = useEditorState(
    (store) => store.editor.githubOperations,
    'Github operations',
  )

  const githubWorking = React.useMemo(() => githubOperations.length > 0, [githubOperations])

  const [importing, setImporting] = React.useState(false)

  React.useEffect(() => {
    if (
      !githubOperations.some(
        (op) => op.name === 'loadBranch' && githubRepoName(op.repository) === props.fullName,
      )
    ) {
      setImporting(false)
    }
  }, [githubOperations, props.fullName])

  // ...

  const importRepository = React.useCallback(() => {
    const parsedTargetRepository = parseGithubProjectString(props.fullName)
    if (parsedTargetRepository == null || props.defaultBranch == null) {
      // ...
    } else {
      setImporting(true)
      void getBranchContent(
      // ...
    }

    // ...
    <button
       // ...
        disabled={!props.importPermitted || githubWorking}
       // ...
      >
        {importing ? <GithubSpinner /> : 'Import'}
      </Button>
```

In `editor-state.ts`:
```ts
export type GithubOperation =
  | { name: 'commish' }
  | { name: 'listBranches' }
  | { name: 'loadBranch'; repository: GithubRepo; branchName: string }
  | { name: 'loadRepositories' }
```

In `github.ts`:
```ts
export async function getBranchContent(
  dispatch: EditorDispatch,
  githubRepo: GithubRepo,
  projectID: string,
  branchName: string,
): Promise<void> {
  const operation: GithubOperation = {
    name: 'loadBranch',
    repository: githubRepo,
    branchName: branchName,
  }
  // ...
```

editor/src/components/navigator/left-pane.tsx
```diff
@@ -763,6 +768,259 @@ const GithubSpinner = () => {
...
+          {props.private ? 'private' : 'public'}
+          {props.updatedAt == null ? null : (
+            <>
+              {' · '}
+              <TimeAgo date={props.updatedAt} />
+            </>
+          )}
+        </span>
+      </div>
+      <button
```
ruggi, line 848: Could we maybe use `Button` here for consistency?

```ts
      <Button
        spotlight
        highlight
        disabled={!props.importPermitted}
      // ...
      >
        // ...
      </Button>
```

editor/src/components/navigator/left-pane.tsx
seanparsons: Fixed in 19dd7446e.

editor/src/components/navigator/left-pane.tsx
seanparsons: Fixed in 19dd7446e.

ruggi: probably worth to add some custom `padding` since now it's all constrained 
Screen Shot 2022-11-01 at 12 49 29 PM

editor/src/components/navigator/left-pane.tsx
ruggi: it'd be good to add the `importing` logic as well, or all the import buttons will be spinning while github is working (see the example snippet above for my naive take on it)

Screen Shot 2022-11-01 at 12 48 43 PM

ruggi (approved):
LGTM 🎖️ it'd be nice to have the spinner tied to only one of the import buttons - but this is cool 👏

seanparsons:
> LGTM medal_military it'd be nice to have the spinner tied to only one of the import buttons - but this is cool clap

Added repo specific spinner functionality in de9b81e61.

editor/src/components/navigator/left-pane.tsx
seanparsons: Fixed padding in de9b81e61.

PR #2745, opened 2022-10-31
Phase one title bar redesign

thetylerwolf:
Phase one update to Title Bar for redesign

**Problem:**
The current UI element configuration for the menu bar has outlived its usefulness. Time to redesign.

**Fix:**
Consolidated/removed various sections of the currently existing left pane tabs, added a Title bar with useful functionality.

**Commit Details:**
- Removed notice bar information related to logged in/out state
- Added <TitleBar> component with logged in and logged out states
- Consolidated sharing and project panes into the settings pane
- menu bar is now hidden when the left pane is hidden
- Shattered (refactored) left-pane.tsx and spread it in 8 directions


editor/src/components/canvas/canvas-strategies/strategies/absolute-flex-reparent-strategy-helpers.spec.browser2.tsx
```diff
@@ -2156,7 +2156,7 @@ describe('Reparent indicators', () => {
     await renderResult.getDispatchFollowUpActionsFinished()
 
     // Check the indicator presence and position.
-    await checkReparentIndicator(renderResult, 432, 637, 2, 123)
+    await checkReparentIndicator(renderResult, 388, 645, 2, 123)
```
Rheeseyb, line 2159: FYI to others viewing this - these hard coded values are based on exact pixel coordinates, and so have all had to be manually offset as a result of the menubar change. It seems we have quite a few tests along these lines, which we need to address separately using values calculated based on actual DOM elements.

editor/src/components/titlebar/Buttons.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const RoundedButton: React.FC<React.PropsWithChildren<TitleButtonProps>> = ({
+  onClick,
+  color,
+  children,
+}) => {
+  const ButtonEl = styled(TitleButton)((props) => ({
+    padding: '3px 12px 3px 8px',
+    height: 36,
+    borderRadius: 18,
+    ...(color && { backgroundColor: color }),
```
seanparsons, line 56: Wait, what's going on here? Is this spreading a boolean?

seanparsons (approved):
Looks good, but could we lowercase the filenames for things like `Buttons.tsx` please?

editor/src/components/editor/notification-bar.tsx
```diff
@@ -43,16 +40,9 @@ export const LoginStatusBar = React.memo(() => {
...
     case 'NOT_LOGGED_IN':
-      return (
-        <NotificationBar
-          level='PRIMARY'
-          message={'Welcome to Utopia. Click here to sign in and save your projects.'}
-          onClick={onClickLoginNewTab}
-          style={{ cursor: 'pointer' }}
-        />
-      )
+      return null
```
Rheeseyb, line 45: We still have other notification bars - do we still want to keep those for the moment?

editor/src/components/editor/store/editor-state.ts
```diff
@@ -256,8 +229,7 @@ export interface UserState extends UserConfiguration {
 export type GithubOperation =
   | { name: 'commish' }
   | { name: 'listBranches' }
-  | { name: 'loadBranch'; branchName: string; githubRepo: GithubRepo }
-  | { name: 'loadRepositories' }
+  | { name: 'loadBranch'; branchName: string }
```
Rheeseyb, line 232: It looks like some of the newer GH code was lost during a merge. I think the only code in this file coming from your branch is around the `export enum LeftMenuTab` above, so everything below that point should be unchanged

editor/src/components/titlebar/Buttons.tsx
```diff
@@ -0,0 +1,64 @@
+import styled from '@emotion/styled'
```
Rheeseyb, line 1: Please can this file be renamed to `buttons.tsx`

editor/src/components/titlebar/Buttons.tsx
thetylerwolf: It's valid syntax! If color is set, you get the spread object, but if it it's false, you get nothing. But yes, you can actually spread a boolean.

editor/src/components/editor/notification-bar.tsx
thetylerwolf: Yes, for now. Will likely be replaced by future title bar updates

editor/src/components/titlebar/Buttons.tsx
Rheeseyb: The `&&` operator will return the right value if the left value is truthy, so this is checking if `color` is defined, and then returning the object on the right of that `&&`, and then spreading that. Please can we avoid this syntax, as it can be pretty confusing (and we are very much against truthiness as it has bit us on many occasions), and instead be more explicit:
```ts
const bgColor = color == null ? {} : { backgroundColor: color }
const ButtonEl = styled(TitleButton)((props) => ({
    padding: '3px 12px 3px 8px',
    height: 36,
    borderRadius: 18,
    ...bgColor,
}))
```

editor/src/components/editor/editor-component.tsx
```diff
@@ -345,26 +317,27 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
...
             >
+              <SimpleFlexColumn
+                style={{
+                  height: '100%',
+                  width: MenuBarWidth,
+                  backgroundColor: colorTheme.leftMenuBackground.value,
+                }}
+              >
+                <Menubar />
+              </SimpleFlexColumn>
```
ruggi, line 340: Commenting here because I cannot comment directly on the related lines (`473` and onwards) - the `LockerOverlay` component should also be updated to apply the left skew in the style props only when the sidebar is open or it will be incorrectly positioned:

Screen Shot 2022-11-02 at 1 37 18 PM

editor/src/components/navigator/left-pane/forks-given.tsx
```diff
@@ -0,0 +1,168 @@
...
+    </Button>
+  )
+})
+
+export const ForksGiven = React.memo(() => {
+  const colorTheme = useColorTheme()
+
+  const { id, projectName, description, isLoggedIn, forkedFrom } = useEditorState((store) => {
+    return {
+      dispatch: store.dispatch,
```
gbalint, line 52: This is unused

editor/src/components/canvas/design-panel-root.tsx
```diff
@@ -13,7 +13,6 @@ import {
 import { useEditorState } from '../editor/store/store-hook'
 import { InspectorEntryPoint } from '../inspector/inspector'
 import { CanvasWrapperComponent } from './canvas-wrapper-component'
-import { InsertMenuPane } from '../navigator/left-pane'
```
balazsbajorics, line 16: ultra nitpick, just FYI: we try to avoid changes to unrelated files in a PR, to reduce noise for the reviewers.

editor/src/components/navigator/left-pane/github-pane/github-spinner.tsx
```diff
@@ -0,0 +1,45 @@
...
+        from {
+          transform: rotate(0deg);
+        }
+        to {
+          transform: rotate(360deg);
+        }
+      `
+
+  return (
+    <FlexColumn>
```
gbalint, line 19: Why do you need to put this single svg into a FlexColumn?

editor/src/components/canvas/design-panel-root.tsx
Rheeseyb: The import statement has changed

Rheeseyb: (I thought the same, that's why I knew the answer straight away :D)

editor/src/components/navigator/left-pane/settings-pane.tsx
```diff
@@ -0,0 +1,205 @@
...
+  )
+
+  const handleBlurProjectName = React.useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      updateProjectName(e.target.value)
+    },
+    [updateProjectName],
+  )
+
+  const handleBlurProjecDescription = React.useCallback(
```
gbalint, line 92: Typo: handleBlurProjec**t**Description

editor/src/components/canvas/design-panel-root.tsx
balazsbajorics: ooooh, sorry! I thought it's a leftover Organize Imports. sorry for the noise then!!

editor/src/components/navigator/left-pane/github-pane/github-spinner.tsx
thetylerwolf: This element is used in multiple files and that's how it was implemented, so I didn't want to disturb it

editor/src/components/navigator/left-pane/forks-given.tsx
thetylerwolf: good catch!

editor/src/components/titlebar/title-bar.tsx
```diff
@@ -0,0 +1,118 @@
...
+      </div>
+
+      <div>
+        <ProjectTitle>{projectName}</ProjectTitle>
+      </div>
+
+      <div style={{ display: 'flex', height: '100%' }}>
+        <TestMenu />
+        {/* <TextButton onClick={exportToGithub}>Fork</TextButton> */}
+        {loggedIn ? null : (
```
balazsbajorics, line 110: the loginState has multiple statuses (eg COOKIES_OR_LOCALFORAGE_UNAVAILABLE) where showing the "Sign in to Save" button is not the adequate UI choice. I think we should at least put a TODO here to make sure we don't forget that this needs improvements, or create a second const that is explicitly about showing the Sign In button

editor/src/components/navigator/left-pane/settings-pane.tsx
thetylerwolf: I will fix it

editor/src/components/titlebar/title-bar.tsx
thetylerwolf: Noted. We have a bunch of states in notification-bar.tsx (not sure if it's exhaustive), which is still being used as well.

balazsbajorics (approved):

Issue #2751, opened 2022-11-02
Padding controls only work when padding values are specified in `px`


bkrmendy:
## Problem
Padding controls only work when padding values are specified in `px`

## Solution
Padding manipulation code should be generalised to handle padding values that are specified in `%`, `em`, etc

PR #2753, opened 2022-11-02
Padding controls - make it work when padding is specified in values other than `px`

bkrmendy:
https://github.com/concrete-utopia/utopia/issues/2751

## Try it [here](https://utopia.pizza/p/26178fdd-maize-mulberry/?branch_name=feature/padding-control-non-px-units)

padding-units

## Problem
Padding can only be adjusted by dragging the padding controls if the padding is specified in `px` units. We want to make this more general and make it work with other units (such as `em` and `%`) as well.

## Implementation
As far as padding controls were concerned, padding used to be represented as a simple `number`, since we only handled padding values that were specified in `px`. Now that it can be any unit, the padding data structure is comprised of 
- the actual rendered size of the padding (taken from `specialSizeMeasurements`)
- the padding value as a `CSSNumber`, taken from `props`

Once we have the actual size and the value from CSS, the resize interaction calculates how many "units" a given drag vector means, and updates the padding with that much. For example, if `paddingLeft` is `2em`, and an `em` unit happens to be `15px`, dragging will adjust `paddingLeft` in increments of `15px`.

## Changes compared to `master`
- the padding value indicator shows the unit as well, in addition to the value of the given padding
- the padding interaction adjusts the padding using the units the padding was originally specified in


editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
```diff
@@ -141,8 +142,8 @@ export const PaddingResizeControl = controlForStrategyMemoized(() => {
 
   const elementMetadata = useRefEditorState((store) => store.editor.jsxMetadata)
 
-  const widthWithtDefaultPx = (w: number) => w + 'px'
-  const heightWithtDefaultPx = (h: number) => h + 'px'
+  const numberToPxValue = (n: number) => n + 'px'
+  const cssNumberToString = (n: CSSNumber) => n.value + (n.unit ?? 'px')
```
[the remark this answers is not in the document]
Rheeseyb: ...saying that, the `CSSNumber` import is already coming from that same file too, so maybe just leave it all there and we can choose whether or not to move those functions "later"

editor/src/components/canvas/controls/select-mode/padding-value-indicator.tsx
```diff
@@ -38,7 +46,7 @@ export const PaddingValueIndicator = controlForStrategyMemoized<PaddingValueIndi
             color: 'white',
           }}
         >
-          {actualPaddingValue}
+          {actualPaddingValue.value + (actualPaddingValue.unit ?? 'px')}
```
Rheeseyb, line 49: This appears to be inlining that same css number printing code

editor/src/components/canvas/padding-utils.ts
```diff
@@ -5,48 +5,71 @@ import { ElementInstanceMetadataMap, isJSXElement } from '../../core/shared/elem
...
 import { EdgePiece } from './canvas-types'
 
 type CSSPaddingKey = keyof CSSPadding
-export type SimpleCSSPadding = { [key in CSSPaddingKey]: number }
+type CSSPaddingMappedValues<T> = { [key in CSSPaddingKey]: T }
+export type CSSPaddingMeasurements = CSSPaddingMappedValues<PaddingMeasurement>
+
+export interface PaddingMeasurement {
+  value: CSSNumber
+  renderedDimension: number
```
Rheeseyb, line 17: Please can you rename this to something like `renderedValue` or `renderedValuePx`? It took me a while to figure out what this value represented and why we were dividing by it later on

Rheeseyb (approved):

editor/src/components/canvas/controls/select-mode/padding-value-indicator.tsx
bkrmendy: thanks! I've replaces this with a call to `printCSSNumber`

editor/src/components/canvas/padding-utils.ts
bkrmendy: done!

thetylerwolf (approved):

PR #2760, opened 2022-11-03
fix(reorder) elements with transform props are not fit for reorder

enidemi:
**Problem:**
Exclude elements from dragging reorder strategies when they have transform style props as transform: translate can create an offset.

**Fix:**
Added this information to the `specialSizeMeasurements`.

**Commit Details:**
- update tests and snapshots
- extend metadata
- filter in flex and flow reorder strategies

### DIFF
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -912,6 +912,7 @@ function getSpecialMeasurements(
     !positionValueIsDefault(elementStyle.right) ||
     !positionValueIsDefault(elementStyle.bottom) ||
     !positionValueIsDefault(elementStyle.left)
+  const hasTransform = elementStyle.transform !== 'none'
 
   const flexGapValue = parseCSSLength(parentElementStyle?.gap)
   const parsedFlexGapValue = isRight(flexGapValue) ? flexGapValue.value.value : 0
@@ -943,6 +944,7 @@ function getSpecialMeasurements(
     elementStyle.float,
     hasPositionOffset,
     elementStyle.direction,
+    hasTransform,
   )
 }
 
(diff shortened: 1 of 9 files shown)


Rheeseyb (approved):

bkrmendy (approved):

editor/src/core/shared/element-template.ts
```diff
@@ -1607,6 +1607,7 @@ export interface SpecialSizeMeasurements {
   float: string
   hasPositionOffset: boolean
   textDirection: string
```
gbalint, line 1609: Instead of `hasTransform`, I would add a `transform: 'translate' | 'scale' | 'rotate'`... etc
It is just as easy to extract this information, and maybe in the future we need more than a boolean.

editor/src/components/canvas/dom-walker.ts
```diff
@@ -912,6 +912,7 @@ function getSpecialMeasurements(
     !positionValueIsDefault(elementStyle.right) ||
     !positionValueIsDefault(elementStyle.bottom) ||
     !positionValueIsDefault(elementStyle.left)
+  const hasTransform = elementStyle.transform !== 'none'
```
seanparsons, line 915: Is `transform` never `undefined` or `null`?

seanparsons (approved):

editor/src/components/canvas/dom-walker.ts
enidemi: I checked the initial computed value, it is always 'none' https://developer.mozilla.org/en-US/docs/Web/CSS/transform

editor/src/core/shared/element-template.ts
enidemi: That would be useful, unfortunately the computed value is a matrix form which is not handled by our css parser yet. It could be a nice improvement for later!

PR #2762, opened 2022-11-03
Store Github branches and repos, refresh with buttons

### LINKED ISSUE #2757: Store GH branches and repos, refresh contents via buttons

ruggi:
Fixes #2757 

**Context:**

After the initial exploration work, I'm splitting the work for the GH background refresh (#2758) in incremental PRs. This is the first one that lays the groundwork for the background refreshing.

**Fix:**

1. Store the user's public repos list and the repos branches in the editor state **(not persisted),** so that they can be retrieved and updated independently. This will be used later for the background refresh.
2. When the authentication state changes, refresh the repos list.
3. When the imported repo changes, refresh the branches list.
4. When the contents are pushed to GH, refresh the branches list.
5. Do not lock the editor when refreshing repos and/or branches.
6. Add the ability to refresh the repos manually with a refresh button
refresh-repo
7. Add the ability to refresh the branches manually with a refresh button
refresh-branches

### DIFF
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -163,7 +163,7 @@ import { GuidelineWithSnappingVectorAndPointsOfRelevance } from '../../canvas/gu
 import { MouseButtonsPressed } from '../../../utils/mouse'
 import { UTOPIA_LABEL_KEY } from '../../../core/model/utopia-constants'
 import { FileResult } from '../../../core/shared/file-utils'
-import { GithubFileStatus } from '../../../core/shared/github'
+import { GithubBranch, GithubFileStatus, RepositoryEntry } from '../../../core/shared/github'
 
 const ObjectPathImmutable: any = OPI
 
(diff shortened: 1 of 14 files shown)


Rheeseyb (approved):

editor/src/components/editor/store/editor-state.ts
```diff
@@ -1010,6 +1019,20 @@ export function githubRepo(owner: string, repository: string): GithubRepo {
...
+}
+
+export function githubRepoEquals(a: GithubRepo | null, b: GithubRepo | null): boolean {
+  if (a == null && b == null) {
+    return true
+  }
+  if (a == null || b == null) {
+    return false
+  }
+  return githubRepoName(a) === githubRepoName(b)
```
seanparsons, line 1033: It would be more preferable to check the two fields `owner` and `repository` against each other here, as `githubRepoName` is making 2 new strings for each comparison.

seanparsons (approved):

editor/src/components/editor/store/editor-state.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/2762/commits/a542d5a15696df0a91dbe111ae58fca53c3bc46f

PR #2766, opened 2022-11-03
Surface merge conflicts.

### LINKED ISSUE #2759: Surface merge conflicts in files.

seanparsons:
Fixes #2759

**Problem:**
If a user makes changes to the project and changes are also made in git, there's currently no way to deal with the merge conflicts that may arise with those changes and pull the branch updates into Utopia.

**Fix:**
This PR introduces a user flow originating with a new button `Update Against Github` which performs the following steps:
- Retrieves the content of the branch at the originating commit.
- Retrieves the current content of the branch.
- With the two variants retrieved and the current status in the editor, performs a basic 3 way merge of the project contents.

**Limitations:**
- The 3 way merge potentially dangerously ignores anything that could be a tree conflict and goes with the current project contents for that part of the tree.
- The uploading of assets when retrieving the branch contents has been completely removed, resulting in the endpoint that does that being one that can be safely called in parallel. In a subsequent piece of work, what should be implemented is an endpoint that can save those assets. In this PR blob SHAs have been added to `ImageFile` and `AssetFile` to support this change, so for example in a tree conflict outcome we would only want to save the assets that survive the merge.

**Commit Details:**
- Include `gitBlobSha` in `ImageFile` and `AssetFile`.
- Added `node-diff3` dependency to the editor.
- Added `UPDATE_AGAINST_GITHUB` action so that merging the project contents changes is properly coordinated with our regular chain of actions.
- Added `updateAgainstBranch` entry to `GithubOperation`.
- Renamed `getBranchContent` function to `updateProjectWithBranchContent` to more accurately represent what it is doing.
- Added `Update Against Github` button to trigger the operation of checking what changes have happened and how to merge them.
- Refactored out some logic into `getBranchContentFromServer` as the same call was used in multiple places.
- Added `mergeProjectContents` and `mergeProjectContentsTree` that perform a 3-way merge of the project contents, with the limitation that they do not currently handle tree conflicts.
- Changed `getGithubBranchContentEndpoint` to take a commit SHA instead of a project ID and switched it over to the `GET` verb from the `POST` verb.
- Refactored `getGithubBranch` to support getting the latest commit for a branch or in the presence of a commit SHA to get the content of that commit.
- `getGithubBranch` does not now upload assets while retrieving a branch so that simultaneous calls are possible and wont end up writing over the top of assets pulled by the others.
- Added `getGitCommit` to pull the details of the tree related to a particular commit.


ruggi:
Looks neat! One thing we have to take into account tho is that the merge will replace the indentation and formatting, introducing all sorts of diff changes that may be very confusing.

You can get an idea of this by doing this:
1. load a branch
2. immediately update against github
3. you get diff changes on basically _all_ source files

https://user-images.githubusercontent.com/1081051/200040113-28c36f7b-6dc4-44e3-a87b-40e8225cc07d.mov

Do you think we should tackle this later? Is this outside of the scope of the PR?

seanparsons:
> Looks neat! One thing we have to take into account tho is that the merge will replace the indentation and formatting, introducing all sorts of diff changes that may be very confusing.
> 
> You can get an idea of this by doing this:
> 
> 1. load a branch
> 2. immediately update against github
> 3. you get diff changes on basically _all_ source files
> 
>  Screen.Recording.2022-11-04.at.6.34.43.PM.mov 
> Do you think we should tackle this later? Is this outside of the scope of the PR?

Aha, I did see this once and then didn't subsequently I assume because I was using the same project. I _think_ that we might be missing a Prettier format pass on the project contents that we start with and at some point those files are getting Prettier applied to them which is then shunting a bunch of stuff around.

seanparsons:
> Looks neat! One thing we have to take into account tho is that the merge will replace the indentation and formatting, introducing all sorts of diff changes that may be very confusing.
> 
> You can get an idea of this by doing this:
> 
> 1. load a branch
> 2. immediately update against github
> 3. you get diff changes on basically _all_ source files
> 
>  Screen.Recording.2022-11-04.at.6.34.43.PM.mov 
> Do you think we should tackle this later? Is this outside of the scope of the PR?

This should now be fixed by 8474c64, the `stringSeparator` field value that the node-diff3 library gives in its examples collapses together runs of line feeds.

ruggi (approved):
I tried to break it to no avail. This is so good 👌 

A bonus point for the future (maybe good to write it down somewhere) would be to make the merge indicators on the editor non-selectable, but this is _super_ nice.

PR #2767, opened 2022-11-04
feature(canvas) drag metastrategy

enidemi:
I'm creating this metastrategy that collects base dragging strategies together to solve the following problem:

1. there are cases where no dragging is available and there is no visual feedback but if you keep dragging this element when reaching the parents edge it triggers a reparent strategy
3. it's possible that the escape hatch is automatically activated without user input in these cases, the goal of this metastrategy to be an alternative for that with a higher fitness

Today this has effect on multiline flex and flow layouts where the reorder is restricted but there are available reparent strategies when leaving the parent.

The fallback strategy is only shown in the strategy picker when there are no reparent or real move/drag strategies.


editor/src/components/canvas/canvas-strategies/strategies/move-reorder-metastrategy.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+// + ancestor strategy !!???
+const baseMoveStrategyFactories: Array<CanvasStrategyFactory> = [
+  absoluteMoveStrategy,
+  flexReorderStrategy,
+  flowReorderStrategy,
+  relativeMoveStrategy,
+]
+
+export const moveReorderMetaStrategy: MetaCanvasStrategy = (
```
balazsbajorics, line 33: it sounds like this is all drag-to-move strategies: move, reorder, reparent. I don't think this is too much at all! actually, I think this is a pretty elegant solution.

```suggestion
export const dragToMoveMetaStrategy: MetaCanvasStrategy = (
```

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/drag-to-move-metastrategy.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export function doNothingStrategy(
+  canvasState: InteractionCanvasState,
+  interactionSession: InteractionSession | null,
+  customStrategyState: CustomStrategyState,
+): CanvasStrategy {
+  const selectedElements = getTargetPathsFromInteractionTarget(canvasState.interactionTarget)
+  return {
+    id: 'DO_NOTHING',
+    name: 'Placeholder',
```
gbalint, line 82: I think we should find a better name..... Placeholder is pretty cryptic.

enidemi: oh yes, maybe 'Move' would fit better :)

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/drag-to-move-metastrategy.tsx
```diff
@@ -0,0 +1,108 @@
...
+    interactionSession.interactionData.modifiers.alt
+  ) {
+    return []
+  }
+
+  const ancestorStrategies = lookForApplicableParentStrategy(
+    canvasState,
+    interactionSession,
+    customStrategyState,
+  )
```
Rheeseyb, line 56: This can now be removed

PR #2768, opened 2022-11-04
Refresh Github data asynchronously

### LINKED ISSUE #2758: Refresh GH data in the background

ruggi:
Fixes #2758 

**Context:**

This is continues from https://github.com/concrete-utopia/utopia/pull/2762 by adding the background refresh functionality for the Github data (repositories and branches).

**Fix:**

Every 30s, refresh the GH data (if there are no other GH operations happening at the same time).

**Notes:**
- This is done with a dummy component to separate the code pieces a bit since the main component is already _huge_.

### DIFF
--- a/editor/src/components/editor/editor-component.tsx
+++ b/editor/src/components/editor/editor-component.tsx
@@ -428,7 +428,7 @@ const GithubDataManager = () => {
   // 1. perform a straight refresh then the repo or the auth change
   React.useEffect(() => {
     void refreshGithubData(dispatch, { githubAuthenticated, githubRepo })
-  }, [githubAuthenticated, githubRepo, dispatch])
+  }, [dispatch, githubAuthenticated, githubRepo])
 
   // 2. schedule a repeat refresh every GITHUB_REFRESH_INTERVAL
   React.useEffect(() => {
@@ -443,7 +443,7 @@ const GithubDataManager = () => {
     return function () {
       clearInterval(refreshInterval)
     }
-  }, [githubAuthenticated, githubRepo, dispatch, githubOperations])
+  }, [dispatch, githubAuthenticated, githubRepo, githubOperations])
 
   return null
 }
--- a/editor/src/components/editor/editor-component.tsx
+++ b/editor/src/components/editor/editor-component.tsx
@@ -415,7 +415,7 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
   )
 })
 
-const useGithubData = () => {
+const useGithubData = (): void => {
   const dispatch = useEditorState((store) => store.dispatch, 'Dispatch')
   const { githubAuthenticated, githubRepo, githubOperations } = useEditorState(
     (store) => ({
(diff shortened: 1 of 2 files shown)


editor/src/components/editor/editor-component.tsx
```diff
@@ -418,11 +408,46 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
         <ModalComponent />
         <ToastRenderer />
         <LockedOverlay />
+        <GithubDataManager />
       </SimpleFlexRow>
     </>
   )
 })
 
+const GithubDataManager = () => {
```
bkrmendy, line 417: I think it would be worth it to call this `useGithubData` or similar, because it uses hooks inside, and if this is used in an `if` or after an early return the app will break with React errors (as outlined in https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level) (and the linter can help enforce these rules if the function's name starts with `use...`

ruggi: yeah, but I was unsure given that that hook would not return anything.
would you think it'd be fine to do just call the hook function without caring about the return value?

e.g.
```ts
const useGithubData = () => {
    // ...
}

const EditorComponentInner = React.memo() => {
  // ...
  useGithubData()
  // ...
  return <>...</>
}
```

seanparsons: It's definitely worth making it honour the hooks rules, even if just to keep us safe from future refactoring introducing something that causes bugs.

bkrmendy: it's not about the return value, but about how hooks are implemented (there's a good description [here](https://stackoverflow.com/questions/54673188/how-does-react-implement-hooks-so-that-they-rely-on-call-order), it's up to you to decide how good an idea it was to implement hooks this way 😬).

ruggi: something like this? https://github.com/concrete-utopia/utopia/pull/2768/commits/0fdc67cce16a086d78d2e495125ca1f74961213f

seanparsons (approved):

editor/src/components/editor/editor-component.tsx
gbalint: BTW hooks doesn't have to return anything, see e.g. `React.useEffect`

editor/src/components/editor/editor-component.tsx
```diff
@@ -423,6 +415,40 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
...
+      return
+    }
+
+    let interval = setInterval(refresh, GITHUB_REFRESH_INTERVAL_MILLISECONDS)
+    return function () {
+      clearInterval(interval)
+    }
+  }, [refresh, githubOperations])
+
+  return null
```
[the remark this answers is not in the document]
gbalint: Sorry I just realized you never return anything meaningful, so I would just keep it a void function and skip the whole `return null` from the last line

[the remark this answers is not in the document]
ruggi: https://github.com/concrete-utopia/utopia/pull/2768/commits/660f6861c84ee83092a251529266371ad14f5bd7 (note: the return values were returned by two different things)

bkrmendy (approved):

gbalint (approved):

PR #2778, opened 2022-11-07
fix(canvas) show zero size controls on reparent

### LINKED ISSUE #2773: show zero size control during reparent

enidemi:
Fixes #2773

https://utopia.pizza/p/3eb80fb7-shocking-diagnostic/?branch_name=fix-canvas-zero-size-controls

**Problem:**
There is a bug that it's possible to reparent into zero size elements but there are no controls shown before it happens.

**Fix:**
Showing zero size controls on dragging an element and reparenting. All zero size controls are shown that can be reparent targets.

When the element is selected only the children zero size controls are visible.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
@@ -16,6 +16,7 @@ import {
   getAdjustMoveCommands,
   getDragTargets,
 } from './shared-move-strategies-helpers'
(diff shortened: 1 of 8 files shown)


balazsbajorics:
I just tried it and it feels pretty intuitive for me! if it turns out to be too noisy in practice, we can tone it down, but I think this is a good first step!

bkrmendy (approved):

seanparsons (approved):

Rheeseyb (approved):

PR #2783, opened 2022-11-08
Remove MissingBoundsHandling

### LINKED ISSUE #2764: Bin MissingBoundsHandling
Now that we have static reparenting, we no longer need to handle missing bounds during reparenting

Rheeseyb:
Fixes #2764 

**Problem:**
We no longer need to use `MissingBoundsHandling` to allow reparenting an element with absolute positioning as a fallback option.

**Fix:**
Delete all uses of `MissingBoundsHandling`. This has the added bonus of simplifying a lot of the reparent logic (especially the `ReparentTargetsToFilter` since we no longer need to cater to two different types of filtered targets).


editor/src/components/canvas/canvas-strategies/strategies/forced-absolute-reparent-strategies.spec.browser2.tsx
```diff
@@ -322,7 +325,7 @@ describe('Forced Absolute Reparent Strategies', () => {
       dragDelta,
       cmdModifier,
       function midDragCallback() {
-        pressKey('3') // this should select the Reparent (Abs, Forced) strategy
+        pressKey('2') // this should select the Reparent (Abs) strategy
```
balazsbajorics, line 328: did we lose a strategy here?

balazsbajorics (approved):

seanparsons (approved):

PR #2792, opened 2022-11-08
Editor theme selection is persisted on the user between sessions

thetylerwolf:
Editor theme selection is persisted on the user between sessions

**Problem:**
After the user selects a color theme (light or dark), leaves Utopia, then returns, the theme selection is not persisted.

**Fix:**
Theme selection is now persisted on the `UserConfiguration` and stored on the server.

**Commit Details:**
- Moved `theme` from `editorState` to `UserConfiguration` and renamed `themeConfig`
- Added functionality on the server to receive updated `UserConfiguration` and store values in the database

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -4621,7 +4621,7 @@ export const UPDATE_FNS = {
     }
     // Side effect.
     void saveUserConfiguration(updatedUserConfiguration)
-    return { ...userState, themeConfig: action.theme }
+    return { ...userState, ...updatedUserConfiguration }
   },
   FOCUS_CLASS_NAME_INPUT: (editor: EditorModel): EditorModel => {
     return {
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -974,7 +974,6 @@ function restoreEditorState(currentEditor: EditorModel, history: StateHistory): 
     vscodeReady: currentEditor.vscodeReady,
     focusedElementPath: currentEditor.focusedElementPath,
     config: defaultConfig(),
-    theme: currentEditor.theme,
     vscodeLoadingScreenVisible: currentEditor.vscodeLoadingScreenVisible,
     indexedDBFailed: currentEditor.indexedDBFailed,
     forceParseFiles: currentEditor.forceParseFiles,
@@ -4368,6 +4367,7 @@ export const UPDATE_FNS = {
     updatedShortcutConfig[action.shortcutName] = [action.newKey]
     const updatedUserConfiguration: UserConfiguration = {
       shortcutConfig: updatedShortcutConfig,
+      themeConfig: userState.themeConfig,
     }
     // Side effect.
     void saveUserConfiguration(updatedUserConfiguration)
@@ -4614,11 +4614,14 @@ export const UPDATE_FNS = {
       forkedFromProjectId: action.id,
     }
   },
-  SET_CURRENT_THEME: (action: SetCurrentTheme, editor: EditorModel): EditorModel => {
-    return {
-      ...editor,
-      theme: action.theme,
+  SET_CURRENT_THEME: (action: SetCurrentTheme, userState: UserState): UserState => {
+    const updatedUserConfiguration: UserConfiguration = {
+      shortcutConfig: userState.shortcutConfig,
+      themeConfig: action.theme,
     }
+    // Side effect.
+    void saveUserConfiguration(updatedUserConfiguration)
+    return { ...userState, themeConfig: action.theme }
   },
   FOCUS_CLASS_NAME_INPUT: (editor: EditorModel): EditorModel => {
     return {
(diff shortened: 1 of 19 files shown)


bkrmendy (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -4614,11 +4614,14 @@ export const UPDATE_FNS = {
       forkedFromProjectId: action.id,
     }
   },
-  SET_CURRENT_THEME: (action: SetCurrentTheme, editor: EditorModel): EditorModel => {
-    return {
-      ...editor,
-      theme: action.theme,
+  SET_CURRENT_THEME: (action: SetCurrentTheme, userState: UserState): UserState => {
+    const updatedUserConfiguration: UserConfiguration = {
+      shortcutConfig: userState.shortcutConfig,
```
balazsbajorics, line 4619: I think it's better if you spread the ...userState here, so if someone adds a third field, this code does not need updating here

balazsbajorics: oh but I see, UserState !== UserConfiguration. eek!

editor/src/components/editor/store/editor-update.tsx
```diff
@@ -320,8 +320,8 @@ export function runSimpleLocalEditorAction(
       return UPDATE_FNS.SET_FILEBROWSER_DROPTARGET(action, state)
     case 'SET_FORKED_FROM_PROJECT_ID':
       return UPDATE_FNS.SET_FORKED_FROM_PROJECT_ID(action, state)
-    case 'SET_CURRENT_THEME':
-      return UPDATE_FNS.SET_CURRENT_THEME(action, state)
+    // case 'SET_CURRENT_THEME':
```
balazsbajorics, line 323: delete me

editor/src/components/editor/actions/actions.tsx
Rheeseyb: Agreed, and then the below return statement should also spread the result of this (so that you're not having to make the same change twice, even though it is minor right now) like so:
```ts
return {
  ...userState,
  ...updatedUserConfiguration,
}
```

balazsbajorics (approved):

editor/src/components/editor/actions/actions.tsx
Rheeseyb: Urgh, yes, you are right, and we aren't storing the `UserConfiguration` separately. I think it still makes sense to spread the `updatedUserConfiguration` into the return value though

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -112,6 +112,11 @@ function processAction(
       ...working,
       userState: UPDATE_FNS.SET_SHORTCUT(action, working.userState),
     }
+  } else if (action.action === 'SET_CURRENT_THEME') {
+    return {
+      ...working,
+      userState: UPDATE_FNS.SET_CURRENT_THEME(action, working.userState),
+    }
```
Rheeseyb, line 119: Ah, it seems I missed that action handlers had started creeping into here. Because there are so many of them, and we don't want to fill `dispatch.tsx` with noise, we call the action handlers here <https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/editor/store/editor-update.tsx#L48>. Please can you move this to there.

Also, or action handling code follows the pattern of always returning the updated `EditorState`, so please can you update the handler that you've added to `actions.tsx` to return the full `EditorState`.

editor/src/components/editor/store/editor-update.tsx
Rheeseyb: No no, use me instead of calling `UPDATE_FNS.SET_CURRENT_THEME` inside `dispatch.tsx`

Rheeseyb (approved):
Some very minor changes requested, but otherwise great stuff!

editor/src/components/navigator/left-pane/settings-pane.tsx
```diff
@@ -36,26 +36,26 @@ const themeOptions = [
...
+    }, 'SettingsPane')
 
   const isMyProject = useIsMyProject(projectId)
 
-  const [theme, setTheme] = React.useState<SelectOption>({
-    label: 'Light',
-    value: 'light',
-  })
+  const [theme, setTheme] = React.useState<SelectOption>(
+    themeOptions.find((option) => option.value === themeConfig) || defaultTheme,
```
seanparsons, line 57: Better to use the nullish coalescing operator:
```suggestion
    themeOptions.find((option) => option.value === themeConfig) ?? defaultTheme,
```

seanparsons (approved):

editor/src/components/editor/store/editor-update.tsx
thetylerwolf: I moved it from here because functions here expect a return type of `EditorState`, but this function modifies the `UserState`

editor/src/components/editor/store/dispatch.tsx
thetylerwolf: I moved the action handler here because the one in `editor-update.tsx` expects a return type of `EditorState`, which doesn't contain `UserState`, so this seemed to be the only option

thetylerwolf: But maybe I'm missing something here?

Rheeseyb: Sorry, I am a total idiot! I’m so unused to us updating anything other then EditorState via the actions that I completely forgot this isn’t even touching that!

PR #2800, opened 2022-11-09
Padding controls can show up on elements that have no business having a padding 

bkrmendy:
## Problem:
Padding controls can show up on elements that have no business having a padding (such as `img`).

## Fix:
Tweak the rules so that
1. if the element has a non-zero padding value, we show the padding controls
2. if the element has any children that are not positioned `absolute` or `sticky` we show the padding controls
3. otherwise, we don't show the padding controls

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
@@ -185,9 +185,15 @@ function supportsPaddingControls(metadata: ElementInstanceMetadataMap, path: Ele
     return false
   }
 
+  const { top, right, bottom, left } = element.specialSizeMeasurements.padding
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
```diff
@@ -185,9 +185,15 @@ function supportsPaddingControls(metadata: ElementInstanceMetadataMap, path: Ele
     return false
   }
 
+  const { top, right, bottom, left } = element.specialSizeMeasurements.padding
+  const elementHasNonzeroPadding = [top, right, bottom, left].some((s) => s != null && s > 0)
```
seanparsons, line 189: Is negative padding valid?

seanparsons (approved):

enidemi (approved):

PR #2803, opened 2022-11-09
removed shade from utopiColor, removed all instances of .shade() and moved actual colors to the light and dark themes

thetylerwolf:
**Problem:**
UtopiColor.shade() is no longer desirable functionality.

**Fix:**
Removed the .shade() function from UtopiColor and moved all colors created with it that exist outside of `theme.ts` into the `theme.ts`

**Commit Details:**
- Removed `shade` function from `utopi-color-helpers.ts`
- Removed tests using the `shade` function

### DIFF
--- a/editor/src/components/titlebar/title-bar.tsx
+++ b/editor/src/components/titlebar/title-bar.tsx
@@ -8,8 +8,6 @@ import { MenuTile } from './menu-tile'
 import { FullHeightButton, RoundedButton, TextButton } from './buttons'
 import { TestMenu } from './test-menu'
 
-interface TitleBarProps {}
-
 const AppLogo: React.FC<{ onClick: () => void }> = ({ onClick }) => (
   <div
     onClick={onClick}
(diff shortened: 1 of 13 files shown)


editor/src/components/titlebar/title-bar.tsx
```diff
@@ -28,8 +26,8 @@ const ProjectTitle: React.FC<React.PropsWithChildren<ProjectTitleProps>> = ({ ch
   return <div style={{ fontWeight: 400, fontSize: 12, padding: '0 10px' }}>{children}</div>
 }
 
-const TitleBar: React.FC<React.PropsWithChildren<TitleBarProps>> = () => {
-  const { loginState, projectName, isCodeEditorVisible, isPreviewPaneVisible, dispatch } =
+const TitleBar = React.memo(() => {
```
balazsbajorics, line 29: this seems to be an unrelated fix that sneaked in :)

enidemi (approved):

seanparsons (approved):

editor/src/components/titlebar/title-bar.tsx
thetylerwolf: eh yeah I caught it after I pushed 😅

PR #2811, opened 2022-11-11
Fix/reorder helpers switch displayprop

enidemi:
Fixes [#2807](https://github.com/concrete-utopia/utopia/issues/2807)

**Problem:**
When reordering a flow layout from keyboard or slider and inserting into a row of inline elements the element is converted to inline too. The issue is that it always added `display: inline-block`, even if it was `display: flex` before, the flex property was lost.

**Commit Details:**
- added tests to convert between flex and inline-flex, block and inline-block
- using `getOptionalCommandToConvertDisplayInlineBlock` which solved the same problem for reparent

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/reorder-slider-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/reorder-slider-strategy.spec.browser2.tsx
@@ -11,8 +11,9 @@ import { selectComponents } from '../../../editor/actions/action-creators'
 import { IconSize } from '../../controls/reorder-slider-control'
 import { ReorderChangeThreshold } from './flow-reorder-helpers'
 import { mouseDragFromPointWithDelta } from '../../event-helpers.test-utils'
+import { MetadataUtils } from '../../../../core/model/element-metadata-utils'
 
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-strategies/strategies/reorder-slider-strategy.spec.browser2.tsx
```diff
@@ -333,12 +335,104 @@ describe('Reorder Slider Strategy', () => {
...
+    ])
+
+    await renderResult.getDispatchFollowUpActionsFinished()
+
+    const draggedElementMetadata = MetadataUtils.findElementByElementPath(
+      renderResult.getEditorState().editor.jsxMetadata,
+      target,
+    )
+    const newDisplayValue = draggedElementMetadata?.specialSizeMeasurements.display
+    expect(newDisplayValue).toEqual('flex')
```
balazsbajorics, line 431: nice!

balazsbajorics (approved):

bkrmendy (approved):

Rheeseyb (approved):

seanparsons (approved):

PR #2813, opened 2022-11-11
Fix keyboard reorder for flex reverse and rtl text direcion

gbalint:
**Problem:**
This was started as a refactor, because `getDirectionFlexOrFlow` was very similar to `singleAxisAutoLayoutContainerDirections`, and I wanted to unify them.

However, it turned out we also have some bugs.

Until now we treated `reverse` for flex layouts and `rtl`/`inline` for flow layouts as something as similar concepts. We thought we just have two cases, the forward and the reverse case.

However, the situation is way more complex than that:
- `rtl` works in flex too, and switches horizontal direction for rows, but doesn't affect columns. It still switches the order of wrapped columns!
- in flow parents, `rtl` switches the direction of `inline-block` rows, but leaves the direction of everything which is in a vertical layout! So it can actually create a permutation of the original order which is neither forward nor reverse.
- flex `reverse` changes the order in columns and rows too!
- Fun fact: having a `row-reverse` and `rtl` reverses the direction twice, so it remains the original

Another important note:
- if you have `rtl` layout, you expect the `right` key to go backward in order and the `left` key to go forward in order, even if you have a column layout. This is a mindset difference if you have `rtl` text direction.

**Fix:**
Instead of trying to find out if we should reverse the direction or not (`shouldReverse`), I added a new function which creates an `arrow button` -> `index change` map depending on the layout. This is necessary because we don't have just 2 options (reverse or not reverse).

In non-reverse flex and in flow:
- `up` and `left` moves backward
- `down` and `right` moves forward

In reverse flex everything is the opposite:
- `up` and `left` moves forward
- `down` and `right` moves backward

If you have `rtl` text direction on top of any layouts, you have to switch the effect of the `left` and `right` keys (but leave `up` and `down` as it is).

I added new test cases for the new logic.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/keyboard-reorder-strategy.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/keyboard-reorder-strategy.ts
@@ -1,5 +1,7 @@
+import { flexDirectionToFlexForwardsOrBackwards } from '../../../../core/layout/layout-utils'
 import { MetadataUtils } from '../../../../core/model/element-metadata-utils'
 import * as EP from '../../../../core/shared/element-path'
+import { ElementInstanceMetadata } from '../../../../core/shared/element-template'
 import { KeyCharacter } from '../../../../utils/keyboard'
 import { absolute } from '../../../../utils/utils'
 import { reorderElement } from '../../commands/reorder-element-command'
@@ -15,7 +17,7 @@ import {
 } from '../canvas-strategy-types'
 import { InteractionSession } from '../interaction-state'
 import { getOptionalDisplayPropCommandsForFlow } from './flow-reorder-helpers'
-import { getDirectionFlexOrFlow, isReorderAllowed } from './reorder-utils'
+import { isReorderAllowed } from './reorder-utils'
 import { accumulatePresses } from './shared-keyboard-strategy-helpers'
 
 export function keyboardReorderStrategy(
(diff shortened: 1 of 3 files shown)


bkrmendy (approved):

enidemi (approved):

editor/src/components/canvas/canvas-strategies/strategies/keyboard-reorder-strategy.ts
```diff
@@ -101,15 +105,59 @@ export function keyboardReorderStrategy(
...
-      return delta
-    default:
-      return 0
-  }
+// This function creates the map which describes which keyboard cursor button should move the index which way.
+// In standard layouts keyboard up and left moves backward and keyboard down and right moves forward.
+// In flex reverse layouts this is fully reversed: keyboard up and left moves forward and keyboard down and right moves backward.
+// If you have rtl text direction on top of any layouts, that should switch the effect of the left and right keys (but leave up and down as it is)
+function getIndexChangesForArrowKeys(element: ElementInstanceMetadata | null): {
+  [key: string]: number
```
balazsbajorics, line 113: nitpick but I think it's worth explicitly listing up right bottom left here

balazsbajorics (approved):

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/keyboard-reorder-strategy.ts
gbalint: I thought about this but then I have to cast the `KeyCharacter` to this union, and that just did not feel right.

gbalint: Fixed in acb802445b9968ef11286a5291bbb7340fa58b8a

PR #2816, opened 2022-11-12
fix gap/padding control handle sizes

bkrmendy:
On Saturday, 12 November an attempt was made to finally make padding/gap control handles look good by making them `2px` across on the inside with a 1px border

image

image

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
+++ b/editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
@@ -56,9 +56,9 @@ function sizeFromOrientation(orientation: Orientation, desiredSize: Size): Size 
 
 export const PaddingResizeControlHoverTimeout: number = 200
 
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
```diff
@@ -56,9 +56,9 @@ function sizeFromOrientation(orientation: Orientation, desiredSize: Size): Size
 
 export const PaddingResizeControlHoverTimeout: number = 200
 
-const PaddingResizeControlWidth = 2
+const PaddingResizeControlWidth = 4
```
Rheeseyb, line 59: Presumably these `const`s should be used in `flex-gap-control.tsx` then if we're aiming to keep the styling of both types of controls in sync (maybe in which case they could be renamed 🤷 )

Rheeseyb (approved):

seanparsons (approved):

PR #2819, opened 2022-11-14
Drag to remove gap/padding props

bkrmendy:
`TODO`

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
@@ -3,7 +3,7 @@ import { ElementInstanceMetadataMap } from '../../../../core/shared/element-temp
 import { optionalMap } from '../../../../core/shared/optional-utils'
 import { ElementPath } from '../../../../core/shared/project-file-types'
 import { assertNever } from '../../../../core/shared/utils'
-import { ParsedCSSProperties } from '../../../inspector/common/css-utils'
+import { ParsedCSSProperties, printCSSNumber } from '../../../inspector/common/css-utils'
 import { stylePropPathMappingFn } from '../../../inspector/common/property-path-hooks'
 import { CSSCursor, EdgePiece } from '../../canvas-types'
 import { deleteProperties } from '../../commands/delete-properties-command'
@@ -14,8 +14,8 @@ import { updateHighlightedViews } from '../../commands/update-highlighted-views-
 import { isZeroSizedElement } from '../../controls/outline-utils'
 import { PaddingResizeControl } from '../../controls/select-mode/padding-resize-control'
 import {
-  FloatingCSSNumberIndicator,
-  FloatingCSSNumberIndicatorProps,
+  FloatingIndicator,
+  FloatingIndicatorProps,
 } from '../../controls/select-mode/floating-number-indicator'
 import {
   deltaFromEdge,
@@ -90,7 +90,7 @@ export const setPaddingStrategy: CanvasStrategyFactory = (canvasState, interacti
     (props) => [
       resizeControl,
       controlWithProps({
-        control: FloatingCSSNumberIndicator,
+        control: FloatingIndicator,
         props: props,
         key: 'padding-value-indicator-control',
         show: 'visible-except-when-other-strategy-is-active',
@@ -214,7 +214,7 @@ function paddingValueIndicatorProps(
   canvasState: InteractionCanvasState,
   interactionSession: InteractionSession | null,
   selectedElements: ElementPath[],
-): FloatingCSSNumberIndicatorProps | null {
+): FloatingIndicatorProps | null {
   const filteredSelectedElements = getDragTargets(selectedElements)
 
   if (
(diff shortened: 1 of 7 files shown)


editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
```diff
@@ -141,23 +144,78 @@ export const setPaddingStrategy: CanvasStrategyFactory = (canvasState, interacti
...
+
+      if (rawDelta < PaddingTearThreshold) {
+        const paddingPropToBeRemoved = paddingPropForEdge(edgePiece)
+        return strategyApplicationResult([
+          ...basicCommands,
+          deleteProperties('always', selectedElement, [
+            StylePaddingProp,
+            stylePropPathMappingFn(paddingPropToBeRemoved, ['style']),
+          ]),
+          ...IndividualPaddingProps.filter(
```
seanparsons, line 177: Could this be broken out into a variable? I think with some sensible name it would make the purpose of this a little clearer.

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
```diff
@@ -258,6 +259,155 @@ describe('Padding resize strategy', () => {
...
+      }),
+      'await-first-dom-report',
+    )
+
+    await testPaddingResizeForEdge(editor, dragDeltaToZero, 'top', 'precise')
+    await editor.getDispatchFollowUpActionsFinished()
+
+    expect(getPrintedUiJsCode(editor.getEditorState())).toEqual(
+      makeTestProjectCodeWithLongHandPaddingValues({
+        paddingTop: `10`,
```
Rheeseyb, line 404: It seems we have some inconsistent behaviour here. If we have 3 padding values set (without units) and then drag to add the 4th, we squash them into a single `padding` prop that explicitly provides `px` as the unit. Then remove two prop and the padding is split into the individual props with `px` still. Add one back in and that new one is unitless. Add the 4th again and they all have `px` again. It's a minor thing, but I think it's worth either always explicitly adding `px` or never adding it.

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
```diff
@@ -235,16 +293,31 @@ function paddingValueIndicatorProps(
...
+    Math.max(-currentPadding.renderedValuePx, delta),
+    precisionFromModifiers(interactionSession.interactionData.modifiers),
+  )
+
+  const message = (): string | number => {
+    if (updatedPaddingMeasurement.renderedValuePx >= PaddingTearThreshold) {
+      return printCSSNumber(maxedPaddingMeasurement.value, null)
+    }
+    return '\u2014' // emdash
+  }
```
Rheeseyb, line 317: It looks like this function should be shared between the padding and gap controls to ensure we keep them in sync

editor/src/components/canvas/controls/select-mode/floating-number-indicator.tsx
```diff
@@ -1,50 +1,36 @@
...
-    const colorTheme = useColorTheme()
+export const FloatingIndicator = controlForStrategyMemoized<FloatingIndicatorProps>((props) => {
+  const { value, position } = props
+  const colorTheme = useColorTheme()
 
-    const actualPaddingValue: CSSNumber = {
-      unit: value.unit,
-      value: Math.max(0, value.value),
-    }
+  const scale = useEditorState((store) => store.editor.canvas.scale, 'PaddingValueIndicator scale')
```
Rheeseyb, line 20:
```suggestion
  const scale = useEditorState((store) => store.editor.canvas.scale, 'FloatingIndicator scale')
```

Very minor

bkrmendy: thanks!

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
bkrmendy: and more importantly, border radius controls will be able to reuse it as well

added it!

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
bkrmendy: thanks, done!

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.spec.browser2.tsx
bkrmendy: it was easier to implement it this way but it is in fact better UX if `px` values are used consistently, I added it

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
bkrmendy: done!

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
bkrmendy: done!

PR #2821, opened 2022-11-14
Determine the mime type from the content not the filename.

seanparsons:
**Problem:**
In this PR:
https://github.com/concrete-utopia/utopia/pull/2815
It was observed that as a result of identifying mime types by file extension, it is actually difficult to create a situation where a text file becomes an image causing a conflict.

**Fix:**
Instead of using the file extension, now `libmagic` is used to instead identify the mime type from the content.

**Commit Details:**
- Added `magic` library to the dependencies.
- Replaced `blobEntryTypeFromFilename` with `blobEntryTypeFromContents`.
- Removed `expandedMimeMap` as it's not needed anymore.
- Reformatted some code as `style-project` runs on the whole project.

### DIFF
--- a/clientmodel/lib/src/Utopia/ClientModel.hs
+++ b/clientmodel/lib/src/Utopia/ClientModel.hs
@@ -181,11 +181,11 @@ instance Arbitrary TextFile where
   shrink = genericShrink
 
 data ImageFile = ImageFile
-               { imageType :: Maybe Text
-               , base64    :: Maybe Text
-               , width     :: Maybe Double
-               , height    :: Maybe Double
-               , hash      :: Int
+               { imageType  :: Maybe Text
+               , base64     :: Maybe Text
+               , width      :: Maybe Double
+               , height     :: Maybe Double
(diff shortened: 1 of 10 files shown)


ruggi (commented):
This looks fine, but my 2c: let's keep the old flow in place and combine it with the magic bytes - if the file (as it could be) is small enough, let's decode it and use the magic bytes to figure out what it is (falling back to the `looksLikeXXX`), but if the file is big enough just do the mimetype/file extension "guesstimate".

seanparsons:
> This looks fine, but my 2c: let's keep the old flow in place and combine it with the magic bytes - if the file (as it could be) is small enough, let's decode it and use the magic bytes to figure out what it is (falling back to the `looksLikeXXX`), but if the file is big enough just do the mimetype/file extension "guesstimate".

The problem is that it can then still fall foul of the issue this PR should resolve, but now it does that contingent on the size of the file, which would be highly confusing to a user.

Rheeseyb (approved):

PR #2824, opened 2022-11-14
Sync the VSCode theme with the editor theme

Rheeseyb:
**Problem:**
We had no way of syncing the VSCode theme with the editor's theme, meaning dark mode wouldn't be dark.

**Fix:**
Added a message type for instructing VSCode which theme to use, using the default "Dark+" and "Light+" themes for our dark and light themes respectively. The VSCode extension then updates the user config directly with this value when receiving that message.

To ensure the correct theme is used, we send this message in 3 separate places:
1) When successfully fetching the user configuration from the server
2) After the VS Code bridge has been set up, when VS Code then requests the initialisation data
3) When toggling the theme in the editor
This ensures we are always setting it at the first possible opportunity.

**Note:**
When creating the `theme` column in postgres, we hadn't set the default value, or set it to `NOT NULL`. To combat this I've provided the default value when fetching the stored user config. I'll probably replace that with an SQL migration script to backfill that default value and make the column not null.

**Update:**
I added a migration script to set the default theme value and backfill the existing table.

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -4651,11 +4651,13 @@ export const UPDATE_FNS = {
       shortcutConfig: userState.shortcutConfig,
       themeConfig: action.theme,
     }
-    // Side effect - store the setting on the server
-    void saveUserConfiguration(updatedUserConfiguration)
 
     // Side effect - update the setting in VS Code
     void sendSetVSCodeTheme(action.theme)
+
+    // Side effect - store the setting on the server
+    void saveUserConfiguration(updatedUserConfiguration)
+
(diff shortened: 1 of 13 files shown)


bkrmendy (approved):

enidemi (approved):

seanparsons:
> To combat this I've provided the default value when fetching the stored user config. I'll probably replace that with an SQL migration script to backfill that default value and make the column not null.

For the database side ideally we should create these new columns as nullable because setting a default value creates a table lock and if the table is very large PostgreSQL will just sit there happily adding the default values while the app gets blocked and grinds to a halt.

seanparsons (approved):

PR #2827, opened 2022-11-15
GH pane initial redesign

### LINKED ISSUE #2826: Scaffold GH pane redesign with the existing functionality

ruggi:
Fixes #2826 

This is the initial work to implement the new designs for the GH pane.
The idea is to have the main portions of the designs integrated with the existing components/functionality, so that we can built on top of that.


editor/src/components/navigator/left-pane/github-pane/github-file-changes-list.tsx
```diff
@@ -167,52 +181,72 @@ export const GithubFileChangesList: React.FC<{
...
-            <>
-              <UIGridRow padded variant='|--16px--|<--------auto-------->'>
+      <FlexColumn
+        style={{
+          border: '1px solid #2D2E33',
+          borderRadius: 2,
+        }}
+      >
+        {list.map((i) => {
+          const conflicting = conflicts?.includes(i.filename) || false
```
seanparsons, line 213: I don't think this should have changed, plus `x || false` should be equivalent to `x`.

seanparsons: Ah, if it's falsey (like in the null/undefined case), then `||` will fail over to the RHS, we should definitely avoid using `||` for that.

editor/src/core/shared/github.ts
```diff
@@ -374,7 +382,14 @@ export async function updatePullRequestsForBranch(
         break
       case 'SUCCESS':
         dispatch(
-          [updateGithubData({ currentBranchPullRequests: responseBody.pullRequests })],
+          [
+            updateGithubData({
+              currentBranchPullRequests: responseBody.pullRequests.map((pr) => ({
+                ...pr,
+                number: getPullRequestNumberFromUrl(pr.htmlURL),
```
seanparsons, line 389: There is a ` number` field in the response Github gives us that we can use for this.

editor/src/components/navigator/left-pane/github-pane/github-file-changes-list.tsx
ruggi: I have no idea why it changed o.O I guess it got lost during the big merges and I incorrectly re-typed it ¯\_(ツ)_/¯

editor/src/core/shared/github.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/2827/commits/9392aa53365b8455d504b6fc9b578fd0d22c9215

editor/src/components/navigator/left-pane/github-pane/block.tsx
```diff
@@ -78,7 +79,7 @@ export const Block = React.memo((props: BlockProps) => {
             justifyContent: 'flex-start',
             height: '100%',
             display: 'grid',
-            gridTemplateRows: '12px 10px 1fr',
+            gridTemplateRows: '13px 10px 1fr',
```
lankaukk, line 82: as long as this is the correct length to make the indicator aligned withe the title 👍

ruggi: yeah, it was off by a couple pixels and it was driving me nuts

editor/src/components/navigator/left-pane/github-pane/index.tsx
```diff
@@ -2,290 +2,340 @@
 /** @jsx jsx */
```
Rheeseyb, line 2: This file should definitely be split up, as it contains multiple components, but that can happen later and shouldn't block this PR

Rheeseyb (approved):

PR #2829, opened 2022-11-15
Implemented CSS Vars for color themes, removed colors from UtopiaTheme object, small fixes to theming

thetylerwolf:
**Problem:**
The current theming solution is less than ideal. Class components cannot be easily styled with a consistent theme

**Fix:**
Began implementing the use of CSS variables for styling so that all components can be styled using the same pattern.

**Commit Details:**
- Created ColorThemeStyleComponent to generate CSS variables based on the colorTheme object
- Cleaned up theming in a few places (title-bar, settings-pane, among others)
- Small fixes to make current use of `UtopiaTheme` work with the new setup
- Removed colors from `UtopiaTheme`

NOTE: Began converting the `UtopiaTheme` object into css variables as well, but things got messy quickly. Will implement that in a future PR


editor/src/uuiui/styles/theme/index.ts
```diff
@@ -0,0 +1,70 @@
+import { Theme } from '../../../components/editor/store/editor-state'
+import { useEditorState } from '../../../components/editor/store/store-hook'
+import { dark } from './dark'
+import { light } from './light'
+import { ColorTheme, colorTheme, darkColorTheme, UtopiaTheme } from './utopia-theme'
+import type { UtopiColor } from '../utopi-color-helpers'
+
+export const colorThemeCssVariables = Object.keys(light).reduce((prev, curr) => {
+  const parts = curr.replace(/([A-Z]|[0-9]+)/g, '-$1'),
+    varName = `--${parts.toLowerCase()}`
```
Rheeseyb, line 10:
```suggestion
  const parts = curr.replace(/([A-Z]|[0-9]+)/g, '-$1')
  const varName = `--${parts.toLowerCase()}`
```
Please don't use the combined initialisation syntax for variables

editor/src/uuiui/styles/css-vars.tsx
```diff
@@ -0,0 +1,30 @@
...
+      {':root {'}
+      {Object.entries(colorThemeCssVariables).map((entry) => {
+        const [key, variable] = entry
+        const cssValue = currentTheme[key as keyof typeof currentTheme].cssValue
+        return `${variable}:${cssValue};`
+      })}
+      {/* {Object.entries(utopiaThemeCssValues).map((entry) => {
+        const [variable, value] = entry
+        return `${variable}:${value};`
+      })} */}
```
Rheeseyb, line 26: Please can you remove the commented code from this PR completely. We don't really gain anything from having the code committed and commented out, especially if it's going to be uncommented (and quite possibly changed / moved) in another PR following shortly after this.

editor/src/uuiui/styles/theme/index.ts
```diff
@@ -0,0 +1,70 @@
+import { Theme } from '../../../components/editor/store/editor-state'
+import { useEditorState } from '../../../components/editor/store/store-hook'
+import { dark } from './dark'
+import { light } from './light'
+import { ColorTheme, colorTheme, darkColorTheme, UtopiaTheme } from './utopia-theme'
+import type { UtopiColor } from '../utopi-color-helpers'
+
+export const colorThemeCssVariables = Object.keys(light).reduce((prev, curr) => {
```
Rheeseyb, line 8: I think what you want to use here is `mapValues` from https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/object-utils.ts#L92-L107

seanparsons (approved):

thetylerwolf:
NB: I am currently working on test fixes, but the code itself is unlikely to change

editor/src/components/canvas/controls/reorder-slider-control.tsx
```diff
@@ -127,7 +127,7 @@ export const ReorderSliderControl = controlForStrategyMemoized(
                 background: colorTheme.bg0.value,
                 boxShadow: `inset 0px 0px 0px ${0.5 / scale}px ${colorTheme.border3.value} , 0px ${
                   2 / scale
-                }px ${4 / scale}px 0px ${colorTheme.fg6.o(50).value}`,
+                }px ${4 / scale}px 0px ${colorTheme.fg650.value}`,
```
Rheeseyb, line 130: I think this naming convention is going to cause confusion. I know it's pretty verbose, but maybe it would be better to go with `fg6opacity50` or `fg6o50`? I don't think it's important though, and these keys are in a specific part of the theme that will (hopefully) go soon, so I'm easy either way tbh.

editor/src/uuiui/styles/css-vars.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export const ColorThemeStyleComponent = React.memo(() => {
+  const colorTheme = useColorThemeVariables(),
+    { inverted, ...currentTheme } = colorTheme
+
+  return (
+    <style>
+      {':root {'}
+      {Object.entries({ ...inverted, ...currentTheme }).map((entry) => {
+        const [variable, value] = entry
```
Rheeseyb, line 12: You can deconstruct the entry directly in the inlined function's params like so
```suggestion
      {Object.entries({ ...inverted, ...currentTheme }).map(([variable, value]) => {
```

editor/src/uuiui/styles/theme/utopia-theme.ts
```diff
@@ -111,7 +136,7 @@ const noticeStyles: { [styleName: string]: React.CSSProperties } = {
   },
   info: {
     backgroundColor: '#f1f1f1',
-    color: colorTheme.darkPrimary.value,
+    color: colorTheme.darkPrimary.value as string,
```
Rheeseyb, line 139: Is this `as string` necessary here? I believe this is something you said was necessary at some point during this work but has since become redundant.

editor/src/uuiui/styles/utopi-color-helpers.ts
```diff
@@ -1,47 +1,34 @@
...
 
-export interface UtopiColor {
+export class UtopiColor {
   /**
-   * css-compatible rgba string: `rgba(255,0,255,1)` format
+   * css variable associated with this color: var(--variable-name) format
    */
-  value: string
-
+  value: string = 'var(--utopitheme-not-set)'
```
Rheeseyb, line 8: We don't need to give these variables any values here since we're setting them in the constructor

editor/src/uuiui/styles/utopi-color-helpers.ts
```diff
@@ -1,47 +1,34 @@
...
+    this.value = `var(${path || cssVar})`
+    this.cssValue = hexWithAlpha
+    return this
   }
-  if (opacitycache[this.value][value] == null) {
+
+  /**
+   * Opacity or Alpha. Value goes from 0 to 100
+   */
+  o(value: number): UtopiColor {
```
Rheeseyb, line 26: Looks like this function can be binned now

Rheeseyb (commented):
There appears to be an empty file added:
`editor/src/uuiui/styles/utopi-colors.ts`
Please can you remove it.

Rheeseyb (approved):

editor/src/uuiui/styles/css-vars.tsx
thetylerwolf: I was getting type issues when I tried to do this. I'll try again

editor/src/uuiui/styles/theme/utopia-theme.ts
thetylerwolf: Yes no longer needed

editor/src/uuiui/styles/utopi-color-helpers.ts
```diff
@@ -1,47 +1,34 @@
...
-   * Opacity or Alpha. Value goes from 0 to 100
+   * css-compatible rgba string: `rgba(255,0,255,1)` format
    */
-  o: (value: number) => UtopiColor
-}
+  cssValue: string = '--utopitheme-not-set'
 
-type ColorHex = string
+  constructor(baseColor: string, path?: string) {
+    const hexWithAlpha = Chroma(baseColor).css('rgba')
```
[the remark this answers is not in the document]
thetylerwolf: It made sense when we were catching every color value so that there were no unnecessary duplicates (like an hsl and rgb version of the same color). But now I don't think it is needed. In fact, now that I think about it, this class can be converted back to a single function.

editor/src/components/canvas/controls/reorder-slider-control.tsx
maltenuhn: Anything but 650, please :D Probably more verbose is better since we've got autocomplete

thetylerwolf: went verbose

editor/src/uuiui/styles/css-vars.tsx
thetylerwolf: worked this time

editor/src/uuiui/styles/utopi-color-helpers.ts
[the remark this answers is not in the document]
thetylerwolf: Removed UtopiColor class and `createUtopiColor` is now just a very simple function

editor/src/uuiui/styles/css-vars.tsx
Rheeseyb: Perhaps you did `.map([k,v] =>` which would have failed, as the parens are required for the deconstruction to work

editor/src/uuiui/styles/utopi-color-helpers.ts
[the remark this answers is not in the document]
Rheeseyb: Nice!

PR #2830, opened 2022-11-15
Don't show all canvas controls on small elements

bkrmendy:
## Problem
On small element, canvas controls can be "clutter" that overshadow the actual element

## Fix
Only show control that are the "most" important based on the actual selected element


bkrmendy:
wrong approach

bkrmendy:
rolled into https://github.com/concrete-utopia/utopia/pull/2867

PR #2832, opened 2022-11-15
adding more icons related to github import/export

lankaukk:
Adding a regular cloud icon (online), and a cloud with a slash through it (offline). These icons are placed so the baselines of the clouds align with the other clouds we already have for upload and download.


seanparsons (approved):

Rheeseyb (approved):
☁️

PR #2834, opened 2022-11-16
Represent padding values that are not defined in `style` as `undefined`

bkrmendy:
## Problem:
Currently, padding values that are not defined in `style` are treated as if they were 0. Due to this, we can't differentiate between padding values that were set to 0 intentionally, and padding values that are simply not defined. 

So far this has been an edge case, since the padding controls normalised the `padding` prop to the 4-value form (`padding: "top right bottom left"`). Now that we have the capability to fully remove padding from any of the sides, not knowing whether a padding is 0 or has been removed became a problem. For example, if there's only `paddingTop`, `paddingBottom`, and `paddingRight` present in `style`, and `paddingRight` we removed, how do we know if `paddingLeft` is 0 or not there at all (without reading `style` again and again)?

## Fix:
Refactor the function that reads what padding is set for an element to return `undefined` for paddings that are not set.

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/controls-common.tsx
+++ b/editor/src/components/canvas/controls/select-mode/controls-common.tsx
@@ -10,7 +10,7 @@ export interface CSSNumberWithRenderedValue {
   renderedValuePx: number
 }
 
-export const pureCSSNumberWithRenderedValue = (
+export const unitlessCSSNumberWithRenderedValue = (
   renderedValuePx: number,
 ): CSSNumberWithRenderedValue => ({
   value: { value: renderedValuePx, unit: null },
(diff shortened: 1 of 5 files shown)


editor/src/components/canvas/controls/select-mode/controls-common.tsx
```diff
@@ -10,6 +10,13 @@ export interface CSSNumberWithRenderedValue {
   renderedValuePx: number
 }
 
+export const pureCSSNumberWithRenderedValue = (
```
balazsbajorics, line 13: pure sounds odd to me, would unitless be better?

```suggestion
export const unitlessCSSNumberWithRenderedValue = (
```

balazsbajorics (approved):

editor/src/components/canvas/controls/select-mode/controls-common.tsx
bkrmendy: great idea, done!

PR #2836, opened 2022-11-16
When saving to Github now the branch and commit message can be supplied.

seanparsons:
**Problem:**
Previously the branch name and commit messages were dynamic based on the date and a constant respectively, which is not that useful for real world situations.

**Fix:**
The save request endpoint now accepts `branch_name` and `commit_message` query parameters to set the values of those two when making the calls to save the current project.

**Notes:**
This was tested by supplying values to the function `saveProjectToGithub` and afterwards they were replaced with `null`.

**Commit Details:**
- `createGitCommit` now takes an optional branch name and optional commit message.
- `githubSaveEndpoint` now permits the branch name and commit message to be supplied via optional query parameters.
- `saveProjectToGithub` now takes `branchName` and `commitMessage` parameters which then get added to the fetch request as query parameters.

### DIFF
--- a/editor/src/core/shared/github.ts
+++ b/editor/src/core/shared/github.ts
@@ -173,6 +173,8 @@ export async function saveProjectToGithub(
   projectID: string,
   persistentModel: PersistentModel,
   dispatch: EditorDispatch,
+  branchName: string | null,
+  commitMessage: string | null,
 ): Promise<void> {
   const operation: GithubOperation = { name: 'commish' }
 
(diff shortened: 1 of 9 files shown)


bkrmendy (approved):

ruggi (approved):

editor/src/core/shared/github.ts
```diff
@@ -173,15 +173,30 @@ export async function saveProjectToGithub(
   projectID: string,
   persistentModel: PersistentModel,
   dispatch: EditorDispatch,
+  branchName: string | null,
+  commitMessage: string | null,
```
Rheeseyb, line 177: This typing makes me very uneasy, because it leaves us susceptible to accidentally messing up the order when calling the function. I think we should replace it with some form of (probably nullable) `options` object.

PR #2840, opened 2022-11-16
Retrieve Github user details.

seanparsons:
**Problem:**
We want to be able to show some of the Github user information when the user is logged in.

**Fix:**
Added a service endpoint that gets the appropriate information and a function that can compose with the authentication functions to perform a lookup of the user details when the user is authenticated.

**Commit Details:**
- Added `GithubUser` type and included it in `GithubData.githubUserDetails`.
- Added `updateUserDetailsWhenAuthenticated` as a function that can wrap around a function that returns a boolean indicating the user is authenticated with Github. If they are authenticated it pulls the user's details from the server.
- Used `updateUserDetailsWhenAuthenticated` in the Github polling and during the startup flow for initialising the editor.
- Added `getGithubUserEndpoint` to handle the calls for looking up the user.
- Added `getGithubUser` and associated types to perform the actual work of retrieving the user details.


ruggi (approved):

editor/src/templates/editor.tsx
```diff
@@ -309,7 +310,10 @@ export class Editor {
         // Ensure we have the correct theme set in VS Code
         void sendSetVSCodeTheme(getCurrentTheme(userState))
 
-        void isAuthenticatedWithGithub(loginState).then((authenticatedWithGithub) => {
+        void updateUserDetailsWhenAuthenticated(
+          this.boundDispatch,
+          isAuthenticatedWithGithub(loginState),
+        ).then((authenticatedWithGithub) => {
```
Rheeseyb, line 316: Does this imply we should also be updating the code here? https://github.com/concrete-utopia/utopia/blob/8c7722e5b8dc9bb2d36d12e3bb45a4aec5bc90c9/editor/src/components/editor/server.ts#L393-L400

seanparsons: Now that you've merged that change, yeah! :)

server/src/Utopia/Web/Github.hs
```diff
@@ -294,6 +294,17 @@ listPullRequests accessToken owner repository page possibleHead = do
   let queryParameters = [("page", show page)] <> headParameter
   callGithub getFromGithub queryParameters listPullRequestsErrorCases accessToken repoUrl ()
 
+getGithubUserErrorCases :: (MonadIO m) => Status -> ExceptT Text m a
+getGithubUserErrorCases status | status == notModified304                   = liftIO $ fail "Not modified returned for Github user."
```
Rheeseyb, line 298: Why is the 304 an explicitly handled failure case?

seanparsons: I just handle the cases like this that Github says they might return.

editor/src/templates/editor.tsx
seanparsons: Fixed in bc55a5974.

Rheeseyb (approved):

PR #2841, opened 2022-11-16
Pull stored config and github auth status on polled login

Rheeseyb:
**Problem:**
When opening the app as a logged in user we would pull down the user's stored config, and update their github authentication status in `editor.tsx` [here](https://github.com/concrete-utopia/utopia/blob/master/editor/src/templates/editor.tsx#L299). However, if the user isn't logged in when first opening the app, but logs in via the "Sign in" button in the title bar, we will update the login status via a polled check in `server.ts` [here](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/editor/server.ts#L372-L403), without then pulling the stored config or github auth status

**Fix:**
Inside the polled check, if the login state has changed, and the user is now logged in, we fire off two fetch requests to pull the stored config and github auth status, dispatched actions that update the `UserState` on the results of those. Those requests are fired immediately (but not `await`ed) so that we can update with the results as soon as possible.

### DIFF
--- a/editor/src/components/editor/server.ts
+++ b/editor/src/components/editor/server.ts
@@ -18,8 +18,15 @@ import { addFileToProjectContents, AssetFileWithFileName, walkContentsTree } fro
 import { isLoginLost, isNotLoggedIn } from '../../common/user'
 import { notice } from '../common/notice'
 import { EditorDispatch, isLoggedIn } from './action-types'
-import { setLoginState, showToast, removeToast } from './actions/action-creators'
+import {
+  setLoginState,
+  showToast,
+  removeToast,
+  setUserConfiguration,
+  setGithubState,
+} from './actions/action-creators'
 import { getFileExtension } from '../../core/shared/file-utils'
+import { isAuthenticatedWithGithub } from '../../utils/github-auth'
 
 export { fetchProjectList, fetchShowcaseProjects, getLoginState } from '../../common/server'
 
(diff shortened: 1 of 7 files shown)


balazsbajorics (approved):

ruggi (approved):

editor/src/components/editor/server.ts
```diff
@@ -377,7 +384,29 @@ export function startPollingLoginState(
   setInterval(async () => {
     const loginState = await getLoginState('no-cache')
     if (previousLoginState.type !== loginState.type) {
+      if (isLoggedIn(loginState)) {
+        // Fetch the user configuration
+        void getUserConfiguration(loginState).then((userConfig) =>
```
seanparsons, line 389: We probably should `await` these, otherwise if they do a go slow or block, this could pile up multiple calls all on top of each other. I believe that they wont block the main thread if done that way either.

seanparsons: I'm not sure if `setInterval` is smart enough to wait until Xms _after_ the last one fired or not....

seanparsons (approved):

thetylerwolf:
I am curious - why not `await` those requests?

thetylerwolf (approved):

editor/src/components/editor/server.ts
Rheeseyb: These only fire if the login state has changed. My thinking was to not `await` them so that we don't have to block updating the login state, user config, or github authentication state - they can each be updated as soon as we have the data for them.

seanparsons: That's reasonable.

PR #2844, opened 2022-11-17
When local changes have been made, add a warning to the PR pane.

seanparsons:
**Problem:**
Ideally a user does not create a PR when there are local changes present.

**Fix:**
Present a warning to the user in the pull request UI when local changes exist.

**Commit Details:**
- Added a message to `PullRequestPane`, conditionally based on the number of file changes being above 0.


ruggi (approved):
Makes sense, we should include this message in the new designs

bkrmendy (approved):

PR #2846, opened 2022-11-17
Prevent all strategies from applying when a root element is selected

Rheeseyb:
**Problem:**
It is far too easy on the canvas to accidentally update the root element of a component when you meant to update the instance instead. Once that mistake is made, it is hard to correct it if you aren't aware that is what has happened.

**Fix:**
Prevent all strategies from being applicable if a root element is selected. Also, I've updated the ancestor strategy to be applicable at all times if the only element selected is a root element.


bkrmendy (approved):

enidemi (approved):

balazsbajorics (approved):
looks good to merge, but please
1. add a test!!!
2. how do we make sure that if someone adds a new metaStrategy, they will remember to filter out root elements? I wonder if there's a way to do this one level above the registered strategies

Rheeseyb:
> I wonder if there's a way to do this one level above the registered strategies

Hmmm yeah I think this is something that should go into the registered strategies and inside the ancestor strategy (when checking each ancestor), as really the _only_ strategy that should fire here is the ancestor strategy

Rheeseyb:
> 2. how do we make sure that if someone adds a new metaStrategy, they will remember to filter out root elements? I wonder if there's a way to do this one level above the registered strategies

Ok I've tweaked this a bit, so now we use a wrapper metastrategy everywhere that we want to prevent the wrapped metastrategy from being applicable on a root element, with all of that logic being applied directly in and around the `RegisteredCanvasStrategies`

enidemi (approved):

PR #2848, opened 2022-11-17
assets(icons) - various updates and a couple of additions

maltenuhn:
**Problem:**
- Some icons were incorrectly sized or unavailable
- we did not have a pyramid logo without the triangular shadow underneath it
- Some icons were not available in all colours (but sans dark mode we did not notice this before

**Fix:**
Replaces and adds various icons


thetylerwolf (approved):
Truly remarkable!

Rheeseyb (approved):

PR #2855, opened 2022-11-17
style fix for menu tabs

lankaukk:
**Problem:**
The Left Pain Menu Tabs were bothering me, not filling their UIGridRow container, looking super awkward (especially in dark mode)
bad-tabs

**Fix:**
So I adjusted their height and padding. 
good-tabs

### DIFF
--- a/editor/src/uuiui/menu-tab.tsx
+++ b/editor/src/uuiui/menu-tab.tsx
@@ -40,7 +40,6 @@ export const MenuTab: React.FunctionComponent<React.PropsWithChildren<MenuTabPro
       },
       cursor: 'pointer',
       flexGrow: 1,
-      height: '100%',
       justifyContent: 'center',
     }
 
--- a/editor/src/uuiui/menu-tab.tsx
+++ b/editor/src/uuiui/menu-tab.tsx
@@ -32,9 +32,7 @@ export const MenuTab: React.FunctionComponent<React.PropsWithChildren<MenuTabPro
     const label = defaultIfNull<React.ReactElement | string>('', props.label)
 
     const baseStyle = {
-      paddingLeft: 4,
-      paddingRight: 4,
-      paddingBottom: 6,
+      padding: '4px 6px',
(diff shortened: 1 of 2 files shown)


Rheeseyb (approved):

editor/src/uuiui/menu-tab.tsx
```diff
@@ -32,16 +32,15 @@ export const MenuTab: React.FunctionComponent<React.PropsWithChildren<MenuTabPro
     const label = defaultIfNull<React.ReactElement | string>('', props.label)
 
     const baseStyle = {
-      paddingLeft: 4,
-      paddingRight: 4,
-      paddingBottom: 6,
+      padding: '4px 6px',
```
maltenuhn, line 35: the asymmetric styling top vs bottom was intentional to account for the inset box shadow and still visually centre the text - pls check you're happy with the result everywhere we use the tabs in the editor

editor/src/uuiui/menu-tab.tsx
```diff
@@ -32,16 +32,15 @@ export const MenuTab: React.FunctionComponent<React.PropsWithChildren<MenuTabPro
...
       transition: 'all .05s ease-in-out',
       '&:hover': {
         backgroundColor: colorTheme.tabHoveredBackground.value,
         boxShadow: `inset 0px -2px 0px 0px ${colorTheme.primary.value}`,
       },
       cursor: 'pointer',
       flexGrow: 1,
+      height: '100%',
```
maltenuhn, line 43: We try to avoid `height: 100%` because it can have unintended consequences. To get the tabs to full height, you should instead configure the layout system that's rendering them to handle the sizing. 

In this case, that means the container row for the tabs should `stretch` them via `alignItems`.

maltenuhn (changes requested):
Great initiative! Kindly avoid `height: 100%` for re-usable components like tabs. Solution suggested inline.

maltenuhn (approved):
💯 LGTM!

PR #2858, opened 2022-11-18
assets(icons) - light gray uses light gray instead of color w alpha

maltenuhn:
**Problem:**
🎶 Is that a gray I see
    no, it's just transparenc-eeee 🎶

**Fix:**
Uses a solid light grey colour for some of our icons that previously were using a heavily alpha-value'd #000


Rheeseyb (approved):
I just gray myself?

maltenuhn:
There’s a good chance I may have committed some light colours

PR #2861, opened 2022-11-18
Fix/strategy indicator icon delay

enidemi:
**Problem:**
The strategy indicator icons appear with a delay.

**Fix:**
Show the indicator with `opacity: 0` on mousedown. Also fixed in move/ancestor/reparent metastrategies to add the Indicator Flags command properly, originally the `appendCommandsToApplyResult` only added commands when there were real commands from the strategy, but here that means in some strategies this is only the case after the drag threshold passed, but some behave differently, it was inconsistent.
Minor fix for relative move strategy, it wasn't fit before the drag threshold the icon delay was visible.

**Commit Details:**
- update `strategy-indicator`
- update move/ancestor/reparent metastrategies


bkrmendy (approved):

Rheeseyb (approved):

editor/src/components/canvas/controls/select-mode/strategy-indicator.tsx
```diff
@@ -159,3 +177,14 @@ const Divider = React.memo(() => {
   const colorTheme = useColorTheme()
   return <div style={{ height: '100%', width: 1, backgroundColor: colorTheme.fg8.value }} />
 })
+
+interface VisibilityWrapperProps {
+  condition: boolean
```
balazsbajorics, line 182: can we name this `visible`?

balazsbajorics (approved):

editor/src/components/canvas/controls/select-mode/strategy-indicator.tsx
enidemi: fixed

PR #2862, opened 2022-11-18
on component instances, only show padding/border radius controls for explicitly props

bkrmendy:
## Problem:
In the case of padding/border-radius, props defined in `style` on compoenent instances can differ from what is measured by `specialSizeMeasurements` (if the component set it's own padding/borderRadius internally).

## Fix:
on component instances, only show padding/border radius controls for explicitly props

## `TODO`
- [x] border radius
- [ ] padding


bkrmendy:
rolled into https://github.com/concrete-utopia/utopia/pull/2873

PR #2867, opened 2022-11-18
padding/border radius controls overlap

bkrmendy:
## Problem
Border radius controls shouldn't render over the resize controls
Padding background can be rendered over other controls, preventing controls under them from being dragged

## Fix
- add a new editorState, `hoveredViews` that contain the views that are being hovered over
- check this in the padding/gap border radius controls, render the controls if the selected element is among the elements being hovered
- padding background: slap a `z-index: 1` on it

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
+++ b/editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
@@ -147,7 +147,6 @@ const PaddingResizeControlI = React.memo(
           display: 'flex',
           alignItems: 'center',
           justifyContent: 'center',
-          zIndex: 0,
           border: isDragging ? `${dragBorderWidth}px solid ${color}` : undefined,
           ...(hidden ? {} : StripedBackgroundCSS(stripeColor, scale)),
         }}
@@ -159,6 +158,7 @@ const PaddingResizeControlI = React.memo(
           onMouseLeave={hoverEnd}
           onMouseUp={onMouseUp}
           style={{
+            pointerEvents: 'all',
             visibility: shown ? 'visible' : 'hidden',
             position: 'absolute',
             padding: hitAreaWidth,
--- a/editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
+++ b/editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
@@ -30,10 +30,8 @@ import {
   CSSNumberWithRenderedValue,
   PillHandle,
   StripedBackgroundCSS,
-  StripeOpacity,
   useHoverWithDelay,
 } from './controls-common'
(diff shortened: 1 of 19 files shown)


Rheeseyb (approved):

editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
```diff
@@ -164,15 +154,16 @@ const PaddingResizeControlI = React.memo(
...
           onMouseEnter={hoverStart}
           onMouseLeave={hoverEnd}
           onMouseUp={onMouseUp}
           style={{
+            pointerEvents: 'all',
             visibility: shown ? 'visible' : 'hidden',
             position: 'absolute',
             padding: hitAreaWidth,
             cursor: cursor,
+            zIndex: 1,
```
maltenuhn, line 166: obligatory "does this cause conflicts with any modals we display?" question? FWIW I don't think it would.

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -451,12 +472,29 @@ export function useCalculateHighlightedViews(
...
+    maybeClearHighlightsOnHoverEnd,
+    maybeHoverOnHover,
+    maybeClearHoveredViewsOnHoverEnd,
+  } = useMaybeHighlightElement()
   const findValidTarget = useFindValidTarget()
   return React.useCallback(
     (targetPoint: WindowPoint, eventCmdPressed: boolean) => {
       const selectableViews: Array<ElementPath> = getHighlightableViews(eventCmdPressed, false)
       const validElementPath = findValidTarget(selectableViews, targetPoint, 'dont-prefer-selected')
+      const validElementPathForHover = findValidTarget(
```
balazsbajorics, line 486: aha!

balazsbajorics (approved):

editor/src/components/canvas/gap-utils.ts
```diff
@@ -82,13 +83,34 @@ function paddingControlContainerBoundsFromChildBounds(
   }))
 }
 
+function inset(sidess: Sides, rect: CanvasRectangle): CanvasRectangle {
```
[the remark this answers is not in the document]
bkrmendy: not good because `sides` is the constructor function for `Sides`, and applying your suggestion would result in a lint error, since it happens to be in scope (I also named the param `sides` when I created the function)

editor/src/components/canvas/controls/select-mode/padding-resize-control.tsx
bkrmendy: I checked, as far as I can see it doesn't

PR #2868, opened 2022-11-18
deleted blue icons that were white, added white files

lankaukk:
**Problem:**
white icons were incorrectly named blue. 

**Fix:**
exported and added them with the proper naming. also added a couple others needed for dark mode


maltenuhn:
@lankaukk can you check that we weren't using the blue icons somewhere (and they were incorrectly showing up as white)? An easy way to do that is to set the property they are matching to something from code, and looking at the inspector, eg for 

`const customWrap = 'no-wrap'`

and then
```
<div style={{ flexWrap: 'no-wrap' >Hello</div>
<div style={{ flexWrap: customWrap>Hello i should be blue in the inspector</div>
```

maltenuhn:
>  There are still copies of the blue icons that are actually blue.

:excellent:

maltenuhn (approved):

balazsbajorics (approved):

PR #2870, opened 2022-11-18
Convert percent to pixel on reparent

gbalint:
**Problem:**
Elements with percent width/height change their size after reparent depending the new parent (or even collapse to zero.

**Fix:**
Quick fix: just set explicit size when it is a percent.


editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.tsx
```diff
@@ -425,6 +425,7 @@ function createUpdatePinsCommands(
         path,
         stylePropPathMappingFn(framePin, ['style']),
         pinValue,
+        'keep-percent',
```
balazsbajorics, line 428: maybe we should put a TODO here to decide

balazsbajorics (approved):

bkrmendy (approved):

PR #2871, opened 2022-11-18
Restore the old contents.json endpoint implementation.

seanparsons:
**Problem:**
The newer `contents.json` endpoint implementation only includes the values inside of `projectContents`, which is a bit rubbish.

**Fix:**
Replaced the newer implementation with the old one as it is better.

**Commit Details:**
- Reverted the implementation of `downloadProjectEndpoint`.


balazsbajorics:
😍

PR #2873, opened 2022-11-18
For components don't show the padding or radius controls if those props aren't explicitly set already on that instance 2

bkrmendy:
## Problem

If components define their own `padding` or `borderRadius` in the component implementation, as opposed to what is set in `style` on the component instance itself, padding/border-radius controls are not rendered the correct way.

## Fix
We decide whether to show padding/border-radius controls based on the following ruleset:
```
if (padding set in `style`) {
    show all controls, based on what is set in `style`
}

if (padding not set in `style`) and (specialSizeMeasurements are all 0) {
    show all controls
}

if (padding not set in `style`) and (any of specialSizeMeasurements is non-0) {
    show no controls
}

show all controls
```
https://github.com/concrete-utopia/utopia/blob/d670ab59882ca1a599ed397646729ca5a6fbd9c1/editor/src/components/canvas/controls/select-mode/controls-common.tsx#L217-L232

## Failure modes
1. if `style` on the component instance is not spread onto `style` in the root element of the component, whatever is set by the padding/border radius controls doesn't affect the rendered component 
2. if `style` on the component instance is spread like {{ ...props.style, (rest of props)}} the style that was set on the instance is overridden, which is completely valid js/jsx/react


editor/src/components/canvas/canvas-strategies/strategies/set-border-radius-strategy.tsx
```diff
@@ -207,12 +214,38 @@ function borderRadiusFromElement(
...
+    () => false,
+    (e) =>
+      isJSXElement(e) &&
+      (isIntrinsicElement(e.name) || jsxElementNameEquals(e.name, jsxElementName('Scene', []))),
+    element.element,
+  )
+
+  if (
+    !elementIsIntrinsicElementOrScene &&
+    !shouldShowControls({
```
balazsbajorics, line 229: I think `!(elementIsIntrinsicElementOrScene || !shouldShowControls` would be easier on my brain

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
```diff
@@ -265,9 +272,38 @@ function supportsPaddingControls(metadata: ElementInstanceMetadataMap, path: Ele
     return false
   }
 
+  const padding = simplePaddingFromMetadata(metadata, path)
   const { top, right, bottom, left } = element.specialSizeMeasurements.padding
```
balazsbajorics, line 276: I think some name cleanup could be useful here, to indicate what comes from props and what comes from computedStyle

balazsbajorics (approved):
thanks for the tests! I left some readability nits

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
bkrmendy: done!

editor/src/components/canvas/canvas-strategies/strategies/set-border-radius-strategy.tsx
bkrmendy: agreed to do it as a separate PR

gbalint (approved):

PR #2875, opened 2022-11-18
Dragging interactions won't trigger strategies until the threshold has been passed

Rheeseyb:
**Problem:**
Dragging strategies are treated as "active" even before a drag has started, meaning certain canvas controls are incorrectly hidden during selection, and causing the strategy picker to show up as though the user had been dragging.

**Fix:**
Don't start applying strategies on dragging interactions until the drag threshold has been surpassed, using early returns in the functions for creating and updating interactions if the threshold hasn't been passed.


bkrmendy (approved):

editor/src/components/editor/store/dispatch-strategies.spec.tsx
```diff
@@ -836,66 +838,37 @@ describe('only update metadata on SAVE_DOM_REPORT', () => {
...
-            if (interactionSession == null) {
-              return strategyApplicationResult([])
-            }
-            expect(canvasState.startingMetadata).not.toBe(interactionSession.latestMetadata)
-            expect(canvasState.startingAllElementProps).not.toBe(
-              interactionSession.latestAllElementProps,
-            )
-
-            // first we make sure the _starting_ metadata and startingAllElementProps have the original undefined backgroundColor
-            expect(
```
balazsbajorics, line 876: removing these expects defeats the point of this test. please instead create a setup where we already have a passing drag interaction so this TEST_STRATEGY can become active

Rheeseyb: I've just spoken to Balazs about this. I tried doing this and was seeing some unexpected and hard to debug issues around missing `latestMetadata` for the affected element. I think it ultimately stems from this not being a browser test. Either way this needs further investigation. I'm going to add a FIXME into this test and capture this in one of the paper docs so that it's not forgotten.

gbalint (approved):

PR #2877, opened 2022-11-19
Cleanup/bars and borders

maltenuhn:
**Problem:**
- With the new toolbar especially, we started duplicating a bunch of functionality, which added clutter to the editor
- We have more panels in the editor that sometimes didn't use borders to subdivide from each other
- the formula bar wants a lot of attention, but in its current incarnation doesn't merit it

**Fix:**
- Moved things about
- Slimmed down the topmenu (but didn't rename it) and moved it into the canvas root
- tidied up some spacing and added borders where there used to be
- gave the formula bar a very light look, and removed the distracting placeholder text (however: it has issues).

image

(It's a draft PR because the omnibar passes clicks through, but I subsequently realised _all_ of our on-canvas menus do that - including the toolbar)


balazsbajorics:
oh god the tests fail because we have some baked-in global x and y coordinates we are comparing against

balazsbajorics:
> oh god the tests fail because we have some baked-in global x and y coordinates we are comparing against

I went in and manually updated the values. I feel dirty

balazsbajorics (approved):

enidemi (approved):

PR #2880, opened 2022-11-20
border radius controls on rotated element

bkrmendy:
## Problem:
When an element is rotated via `translate: transform`, border radius controls show up in the wrong place

## Fix:
use `element.globalFrame` instead of `element.specialSizeMeasurements.{width/height}` when calculating element bounding box

### After
image

### Before
image


balazsbajorics (approved):

maltenuhn:
NB we've got the same issue with the reorder control

Rheeseyb (approved):

enidemi (approved):

PR #2888, opened 2022-11-21
make border radius not overlap with resize controls

bkrmendy:
## Problem:
Border radius controls might overlap with resize controls

## Fix:
Position border radius controls at a minimum offset of 10px from the corner

**Commit Details:**
- Did [other things]


balazsbajorics (approved):
I think the minimum distance could be even larger, maybe 12px?

Rheeseyb (approved):

seanparsons (approved):

PR #2889, opened 2022-11-21
Switch between creating or updating a reference appropriately.

seanparsons:
**Problem:**
The Github APIs are sensitive to two things during our existing saving flow:
a) The repository being completely empty causes some endpoints to just outright fail.
b) There are distinct create and update endpoints for references which fail if the reference does or does not exist respectively.

**Fix:**
The above issues are fixed in the following ways:
a) An initial commit is pushed to the default branch of the repository if the repository appears to be completely empty.
b) The existence of the reference is checked for and the appropriate create or update reference endpoint is called.

**Commit Details:**
- Added `name` to `RepositoryEntry`.
- Change `private` to `isPrivate` on `RepositoryEntry` to avoid keyword collision.
- Modify `callGithub` to log errors when a non-2XX response comes back.
- Renamed some functions in `Github.hs` which are prefixed with `createGitTree` to make them more consistent with others that have the same prefix.
- Added `updateGitFile`, `getReference`, `updateGitBranch` and `getRepository` functions for calling Github API endpoints.
- Added `createInitialCommitIfNecessary` to handle the empty repository case of creating a pointless commit in a way that seems to work for Github.
- `updated_at` is now required in `RepositoryEntry`.
- `createTreeAndSaveToGithub` now triggers the initial commit creation and switches between the create and update endpoints depending on the existence of the reference.

### DIFF
--- a/server/src/Utopia/Web/Github.hs
+++ b/server/src/Utopia/Web/Github.hs
@@ -112,13 +112,19 @@ gitTreeEntriesFromProjectContent loadAsset createBlob projectContents = do
   entries <- mapConcurrently (gitTreeEntriesFromProjectContentsTree loadAsset createBlob) $ M.elems projectContents
   pure $ join entries
 
-createGitTreeFromProjectContent :: (MonadIO m, MonadBaseControl IO m) => LoadAsset -> SimpleCreateBlob m -> ProjectContentTreeRoot -> ExceptT Text m CreateGitTree
-createGitTreeFromProjectContent loadAsset createBlob projectContents = fmap (CreateGitTree Nothing) $ gitTreeEntriesFromProjectContent loadAsset createBlob projectContents
+gitTreeFromProjectContent :: (MonadIO m, MonadBaseControl IO m) => LoadAsset -> SimpleCreateBlob m -> ProjectContentTreeRoot -> ExceptT Text m CreateGitTree
+gitTreeFromProjectContent loadAsset createBlob projectContents = fmap (CreateGitTree Nothing) $ gitTreeEntriesFromProjectContent loadAsset createBlob projectContents
 
 type MakeGithubRequest a = WR.Options -> String -> a -> IO (WR.Response BL.ByteString)
 
 postToGithub :: (ToJSON a) => MakeGithubRequest a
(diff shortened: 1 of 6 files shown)


server/src/Utopia/Web/Github.hs
```diff
@@ -133,9 +139,9 @@ callGithub makeRequest queryParameters handleErrorCases accessToken restURL requ
               & WR.params .~ queryParameters
   -- Make the request.
   result <- liftIO $ makeRequest options (toS restURL) request
-  -- Uncomment the next line if you want to see the headers.
-  -- liftIO $ print $ view WR.responseHeaders result
   let status = view WR.responseStatus result
+  -- Uncomment the next line if you want to see the headers.
```
ruggi, line 143: probably good to update this comment

server/src/Utopia/Web/Github.hs
```diff
@@ -197,7 +207,46 @@ createGitCommitForTree :: (MonadIO m) => AccessToken -> PersistentModel -> Text
 createGitCommitForTree accessToken PersistentModel{..} treeSha possibleCommitMessage parentCommits = do
   let possibleGithubRepo = targetRepository githubSettings
   case possibleGithubRepo of
-    Just repo -> createGitCommit accessToken repo treeSha possibleCommitMessage parentCommits
+    Just GithubRepo{..} -> createGitCommit accessToken owner repository treeSha possibleCommitMessage parentCommits
+    Nothing   -> throwE "No repository set on project."
+
+updateGitFileHandleErrorCases :: (MonadIO m) => Status -> ExceptT Text m a
+updateGitFileHandleErrorCases status | status == notFound404             = throwE "Repository or tree not found."
```
ruggi, line 214: 😍

ruggi (approved):

server/src/Utopia/Web/Github.hs
seanparsons: Good spot.

seanparsons: Fixed in 18d2365fa.

Rheeseyb (approved):

balazsbajorics (approved):

PR #2899, opened 2022-11-21
Performance regression fix

balazsbajorics:
**Problem:**
The `TitleBar` regressed performance by using getProjectContentsChecksums without the Low Priority Store.

**Fix:**
I made sure it's impossible to accidentally use this without the low priority store ever again:
image

**Commit Details:**
- The stores have a storeName field
- getProjectContentsChecksums -> useGithubFileChanges hook to be reused in the codebase
- useGithubFileChanges explicitly throws an error if it's called from outside the low-priority-store
- wrap the TitleBar in a LowPriorityStoreProvider

### DIFF
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -388,16 +388,23 @@ export type EditorStoreFull = EditorStoreShared & {
   patchedDerived: DerivedState
 }
 
+type StoreName = 'editor-store' | 'canvas-store' | 'low-priority-store'
+
 export type EditorStorePatched = EditorStoreShared & {
+  storeName: StoreName
   editor: EditorState
   derived: DerivedState
 }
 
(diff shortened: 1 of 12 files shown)


editor/src/components/editor/store/editor-state.ts
```diff
@@ -388,16 +388,23 @@ export type EditorStoreFull = EditorStoreShared & {
   patchedDerived: DerivedState
 }
 
+type StoreName = 'editor-store' | 'canvas-store' | 'low-priority-store'
```
Rheeseyb, line 391: After this PR (to prevent blocking it) please can we rename these to better clarify the differences between the `editor-store` and `canvas-store`

balazsbajorics: yep! absolutely

balazsbajorics:
I'm going to merge this pr with one review, to fix performance

PR #2912, opened 2022-11-22
Chore/remove canvas strategies fs

gbalint:
**Problem:**
The editor doesn't work anymore without the Canvas Strategies feature switch, time to remove it.

**Fix:**
I tried to delete all unused code, but this is not easy. I remove a select-mode-control-container and a lot of unused controls too though.


editor/src/components/canvas/controls/size-box.tsx
```diff
@@ -87,84 +87,23 @@ class ResizeControl extends React.Component<ResizeControlProps> {
...
-            [],
-          ),
-          start,
-          null,
-          targetProperty,
-          enableSnapping,
-          centerBasedResize,
-          keepAspectRatio,
-        )
+      const startPoint = windowToCanvasCoordinates(
```
enidemi, line 90: I think this is used only by the multiselect-resize-control which is a deleted file

enidemi (approved):

editor/src/components/canvas/controls/size-box.tsx
gbalint: thanks, deleted!

Rheeseyb (approved):

seanparsons (approved):

PR #2920, opened 2022-11-23
Make the Github polling more resilient.

seanparsons:
**Problem:**
There were a couple of issues with the previous polling:
- It was being (re)triggered through changes in a React `useEffect` call, which could trigger a lot of `clearInterval` and `setInterval` calls as various dependencies of the call changed.
- As multiple parts of the polling were updating a `GithubData` value, they were all setting their own value for `lastUpdatedAt` which resulted in the timer in the UI bouncing around near 0s for a few seconds.
- The various promises triggered in the polling weren't `await`ed or anything like that so in the case of a slow call could dogpile the server.

**Fix:**
The primary part of improving the polling was to have all the various parts of the polling have their results be accumulated in the polled function, which had the following effect:
- With a tweak to the action, all the updates to `GithubData` are now happening in quick succession so the `lastUpdatedAt` value doesn't visibly get bounced about.
- Now they are all effectively being waited on to get their results, the polling function only succeeds once all of the promises have finished.

The triggering of the polling has been changed to live completely outside the sphere of React as well, with it scheduling the next run only once the previous poll job has finished.

**Commit Details:**
- Moved updating of `GithubData.lastUpdatedAt` to the action handler from the action creator so that it should be changing monotonically. As well as that being the exact time it does change.
- Removed React based polling from `editor-component.tsx` and added a new function for polling based on the time when the last poll completed, which now accesses the Zustand store more directly.
- New polling function `startGithubPolling` is triggered from the `Editor` component.
- Added `dispatchPromiseActions` utility function.
- Refactored a lot of the functions called during the polling to return the actions that they previously individually dispatched for the result of processing them. As they are now collated together and dispatched all together.
- Refactored out `updateUpstreamChanges` from the previous polling code.

### DIFF
--- a/editor/src/core/shared/github.ts
+++ b/editor/src/core/shared/github.ts
@@ -1218,17 +1218,22 @@ export async function refreshGithubData(
   }
 
   // Resolve all the promises.
-  await Promise.all(promises)
-    .then((collatedActions) => {
-      // Dispatch all the actions from all the polling functions.
-      dispatch(
-        collatedActions.flatMap((arr) => arr),
-        'everyone',
-      )
-    })
-    .catch((error) => {
-      console.error(`Error whie polling Github: ${error}`)
-    })
+  await Promise.allSettled(promises).then((results) => {
+    let actions: Array<EditorAction> = []
+    for (const result of results) {
+      switch (result.status) {
+        case 'rejected':
+          console.error(`Error whie polling Github: ${result.reason}`)
+          break
+        case 'fulfilled':
+          actions.push(...result.value)
+          break
+      }
+    }
+
+    // Dispatch all the actions from all the polling functions.
+    dispatch(actions, 'everyone')
+  })
 }
 
 async function updateUpstreamChanges(
--- a/editor/src/core/shared/github.ts
+++ b/editor/src/core/shared/github.ts
@@ -63,6 +63,13 @@ import { emptySet } from './set-utils'
 import { trimUpToAndIncluding } from './string-utils'
 import { arrayEquals } from './utils'
 
+export function dispatchPromiseActions(
+  dispatch: EditorDispatch,
+  promise: Promise<Array<EditorAction>>,
+): Promise<void> {
+  return promise.then((actions) => dispatch(actions, 'everyone'))
+}
+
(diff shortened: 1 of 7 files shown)


editor/src/core/shared/github.ts
```diff
@@ -355,66 +367,61 @@ export async function updatePullRequestsForBranch(
...
             ),
-          ],
-          'everyone',
-        )
-        break
-      case 'SUCCESS':
-        dispatch(
-          [
+          ]
+          break
```
ruggi, line 402: not needed

editor/src/core/shared/github.ts
```diff
@@ -355,66 +367,61 @@ export async function updatePullRequestsForBranch(
...
             }),
-          ],
-          'everyone',
-        )
-        break
-      default:
-        const _exhaustiveCheck: never = responseBody
-        throw new Error(`Unhandled response body ${JSON.stringify(responseBody)}`)
+          ]
+          break
```
ruggi, line 412: not needed

ruggi (commented):
LGTM! There's only one issue though, which is that now repos and branches don't refresh either on load or when they change. I'll push a small fix itmt

ruggi:
> refresh repos/branch on load/change

this one → https://github.com/concrete-utopia/utopia/pull/2920/commits/aa435632a402cfe7a6a7c4915c33706bfdcb6ed6

ruggi (approved):

seanparsons:
> > LGTM! There's only one issue though, which is that now repos and branches don't refresh either on load or when they change. I'll push a small fix itmt
> 
> this one → [aa43563](https://github.com/concrete-utopia/utopia/commit/aa435632a402cfe7a6a7c4915c33706bfdcb6ed6)

Good spot, thank you!

Rheeseyb (approved):

editor/src/core/shared/github.ts
seanparsons: Fixed in cc6fda6c71313001c3045ec5dd87329b0039f5b4.

editor/src/core/shared/github.ts
seanparsons: Fixed in cc6fda6c71313001c3045ec5dd87329b0039f5b4.

PR #2916, opened 2022-11-23
Tweak catchment areas for canvas controls

bkrmendy:
## Problem:
Canvas controls are sometimes hard to click/drag

## Fixes:
- [x] create a padding around border radius handles to make them easier to catch
- [x] use `magnitude(drag)` when calculating whether the drag threshold has been exceeded (and the drag interaction can be started)
- [x] update said drag threshold from `3px` to `2px`
- [x] use `roundTo` instead of `Math.floor` when calculating the border radius drag delta
- [x] use a custom cursor for border radius/flex gap controls
- [x] make the "catchment area" around padding controls the same width as the edge resize controls
- [x] refine show/hide logic of corner/padding/gap controls to be consistent with each other

PR #2922, opened 2022-11-23
Debug slider to play with the drag threshold

bkrmendy:
image


bkrmendy:
rolled into https://github.com/concrete-utopia/utopia/pull/2916

PR #2923, opened 2022-11-23
padding/gap/border radius controls should always be calculated based on the latest metadata

bkrmendy:
## Problem
padding/gap/border radius controls are calculated based on `startingMetadata`. Because of this, if padding/gap/border radius is changed through the inspector, the controls goes out of sync

## Fix
Calculate padding/gap/border radius controls are calculated based on `latestMetadata` in the control


bkrmendy:
this whole problem might be caused by https://github.com/concrete-utopia/utopia/pull/2930

Plan is to merge https://github.com/concrete-utopia/utopia/pull/2930 first, and then check if the problem persists

bkrmendy:
most of the problems seem to be fixed, the rest are fixed on https://github.com/concrete-utopia/utopia/pull/2940

PR #2927, opened 2022-11-24
Only send selection changed messages if vscode isn't focused

### LINKED ISSUE #2926: Selection following in VS Code breaks after loading a project without the currently open file
If the user opens a code file in the code editor, and then loads a project or branch from github that does not contain the currently open file, then selection following will temporarily be broken _if_ only non-existent files were previous open.

Rheeseyb:
Fixes #2926 

**Problem:**
Loading a branch or repo from github that doesn't contain the file(s) currently open in the code editor will break selection following temporarily (provided the code editor doesn't have _any_ files open that exist in the loaded branch / repo)

**Fix:**
The problem here was that as part of removing those files, VS Code will throw an error complaining that it can no longer read the open file, and then close that file. If this happens and there is no remaining valid open file, VS Code will then believe that it is focused (which we check in the extension via `vscode.window.state.focused`).

The reason we check if VS Code is focused before updating the selection in the VS Code extension is because we don't want selection changes coming from the code editor to then update the user's cursor in the code editor.

To counter this, we now instead perform the focus check on the Utopia side, checking if the iframe used to load VS Code is the `document.activeElement`. We use that to prevent sending `selectedElementChanged` messages, meaning we can now remove the focus check on the VS Code side.

I've also deleted an unused function `sendSelectedElementChangedMessage` so that only one place ever constructs these messages.

### DIFF
--- a/editor/src/components/code-editor/code-editor-container.tsx
+++ b/editor/src/components/code-editor/code-editor-container.tsx
@@ -5,8 +5,7 @@ import { createIframeUrl } from '../../core/shared/utils'
 import { getUnderlyingVSCodeBridgeID } from '../editor/store/editor-state'
 import { VSCodeLoadingScreen } from './vscode-editor-loading-screen'
 import { getEditorBranchNameFromURL, setBranchNameFromURL } from '../../utils/branches'
-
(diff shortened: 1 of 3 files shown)


ruggi (approved):
noice

seanparsons (approved):

PR #2928, opened 2022-11-24
Memoize Canvas.getFramesInCanvasContext.

seanparsons:
**Problem:**
`getFramesInCanvasContext` is called indirectly from a few places from within `handleStrategies` and is comparatively expensive.

**Fix:**
Memoized `getFramesInCanvasContext` as it would appear the parameters aren't changing within a single frame, so we end up saving a couple of ms when calling `handleStrategies`.

**Commit Details:**
- Memoized calls to Canvas.getFramesInCanvasContext.

### DIFF
--- a/editor/src/components/canvas/canvas.ts
+++ b/editor/src/components/canvas/canvas.ts
@@ -19,6 +19,7 @@ import * as EP from '../../core/shared/element-path'
 import { buildTree, ElementPathTree, getSubTree } from '../../core/shared/element-path-tree'
 import { objectValues } from '../../core/shared/object-utils'
 import { fastForEach } from '../../core/shared/utils'
+import { memoize } from '../../core/shared/memoize'
 
 export enum TargetSearchType {
   ParentsOfSelected = 'ParentsOfSelected',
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/canvas.ts
```diff
@@ -46,84 +126,7 @@ const Canvas = {
...
-    return storyboardChildren.flatMap((storyboardChild) => {
-      const subTree = getSubTree(projectTree, storyboardChild)
-      if (subTree == null) {
-        return []
-      } else {
-        return recurseChildren(subTree).frames
-      }
-    })
-  },
+  getFramesInCanvasContext: memoize(getFramesInCanvasContextUncached, { maxSize: 1 }),
```
balazsbajorics, line 129: maybe it's worth making the maxSize 2 or more, because I can imagine some scenario where we call it with the unpatched editor and the patched editor state within the same frame

balazsbajorics (approved):

gbalint (approved):

gbalint (approved):

editor/src/components/canvas/canvas.ts
seanparsons: Yeah, I put 1 in and it made a bunch of calls disappear, but it's probably worth giving it that bit of extra wiggle room.

Changed in c20b2a94c.

PR #2931, opened 2022-11-24
Fix/inspector performance selective rerender

enidemi:
**Problem:**
Inspector controls are slow (sliders, color picker, drag interactions). 

**Fix:**
Only rerender selected elements while dragging inspector controls.

**Commit Details:**
- optional `elementsToRerender` field for transient actions
- `fixElementsToRerender` already fixes it in some cases, now it's also handling transient actions
- `shouldUpdateLowPriorityUI` now correctly uses `elementsToRerender`

Todo:
- [ ] tests 

Before
Screenshot 2022-11-24 at 13 29 25
After
Screenshot 2022-11-24 at 13 26 52


bkrmendy (approved):

editor/src/components/editor/action-types.ts
```diff
@@ -255,6 +255,7 @@ export type SetZIndex = {
 export type TransientActions = {
   action: 'TRANSIENT_ACTIONS'
   transientActions: Array<EditorAction>
+  elementsToRerender: Array<ElementPath> | null
```
balazsbajorics, line 258: this could use the `export type ElementsToRerender = Array<ElementPath> | 'rerender-all-elements'` type here too for consistency

gbalint (approved):

editor/src/templates/editor.tsx
```diff
@@ -127,15 +129,43 @@ function replaceLoadingMessage(newMessage: string) {
   }
 }
 
+function collectElementsToRerenderTransient(
```
balazsbajorics, line 132:
```suggestion
function collectElementsToRerenderFromTransientActions(
```

editor/src/templates/editor.tsx
```diff
@@ -127,15 +129,43 @@ function replaceLoadingMessage(newMessage: string) {
...
 // for the very first pass to get another `rerender-all-elements`.
 let lastElementsToRerender: ElementsToRerender = 'rerender-all-elements'
-function fixElementsToRerender(currentElementsToRerender: ElementsToRerender): ElementsToRerender {
-  let elementsToRerender: ElementsToRerender = currentElementsToRerender
+function fixElementsToRerender(
+  currentElementsToRerender: ElementsToRerender,
+  dispatchedActions: readonly EditorAction[],
+): ElementsToRerender {
+  // while running transient actions there is an optional elementsToRerender
+  const elementsToRerenderTransient = uniqBy<ElementPath>(
```
balazsbajorics, line 159: I think I'd clean the name here too
```suggestion
  const elementsToRerenderFromTransientActions = uniqBy<ElementPath>(
```

editor/src/components/editor/action-types.ts
enidemi: that feels a bit like cheating, as the fixElements helper only works on filtering the result for transient actions

editor/src/templates/editor.tsx
enidemi: fixed

maltenuhn:
Late to the party but: i just played with this. WOW. This colour picker is nothing short of amazing, I don't think we've _ever_ had performance like this.

PR #2932, opened 2022-11-24
currentElementsToRender does not fall back to rerender-all-elements

balazsbajorics:
**Problem:**
All reparenting runs with selective rerender turned off (set to `rerender-all-elements`). This causes a very visible performance problem. 

**Fix:**
I tweaked `fixedElementsToRerender` to a new behavior: when elementsToRerender changes, we don't fall back to rerender-all-elements (as on master) but instead use the union of the old elementsToRerender array and the new elementsToRerender array.

**Other fix:**
The ui-jsx-canvas.tsx was the only place that consumed EditorState.canvas.elementsToRerender – which is not the same value as the output of fixedElementsToRerender. I changed it so the code in ui-jsx-canvas now uses ElementsToRerenderGLOBAL.current which is the same value as used by the ComponentRendererComponent.

### DIFF
--- a/editor/src/templates/editor.tsx
+++ b/editor/src/templates/editor.tsx
@@ -113,6 +113,7 @@ import { ProjectContentTreeRootKeepDeepEquality } from '../components/editor/sto
 import { waitUntil } from '../core/shared/promise-utils'
 import { sendSetVSCodeTheme } from '../core/vscode/vscode-bridge'
 import { refreshGithubData, updateUserDetailsWhenAuthenticated } from '../core/shared/github'
+import { uniqBy } from '../core/shared/array-utils'
 
 if (PROBABLY_ELECTRON) {
   let { webFrame } = requireElectron()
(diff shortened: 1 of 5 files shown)


bkrmendy (approved):

editor/src/templates/editor.tsx
```diff
@@ -128,35 +129,23 @@ function replaceLoadingMessage(newMessage: string) {
...
-      }
+  let fixedElementsToRerender: ElementsToRerender = currentElementsToRerender
+  if (
+    currentElementsToRerender !== 'rerender-all-elements' &&
+    lastElementsToRerender !== 'rerender-all-elements'
+  ) {
+    // if the current elements to rerender array doesn't match the previous elements to rerender array, for a single frame let's use the union of the two arrays
+    fixedElementsToRerender = uniqBy(
+      [...lastElementsToRerender, ...currentElementsToRerender],
+      EP.pathsEqual,
```
gbalint, line 144: I think this function is faster for element paths than uniqBy, maybe we should put it into somewhere more general and use it from here too: https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/commands/set-elements-to-rerender-command.ts#L84

gbalint (approved):

enidemi (approved):

editor/src/templates/editor.tsx
balazsbajorics: fixed

Rheeseyb (approved):

PR #2933, opened 2022-11-24
Detect asset checksums, detect image file changes

### LINKED ISSUE #2917: Include image files in the checksum list

ruggi:
Fixes #2917 

**Problem:**

Image files are not included in the detected GH file changes.

**Fix:**

This PR includes the image files in the github file changes listings (remote and local). A tricky bit was to calculate the _correct_ SHA checksum for the asset, which is now done in the same way `git` will calculate and return it.

### DIFF
--- a/editor/src/components/assets.ts
+++ b/editor/src/components/assets.ts
@@ -74,13 +74,11 @@ export function getProjectContentsChecksums(
         checksums[filename] = getSHA1Checksum(file.fileContents.code)
         break
       case 'ASSET_FILE':
-        if (file.base64 != undefined) {
-          checksums[filename] = getSHA1Checksum(file.base64)
-        }
-        break
       case 'IMAGE_FILE':
-        if (file.gitBlobSha) {
+        if (file.gitBlobSha != null) {
           checksums[filename] = file.gitBlobSha
+        } else if (file.base64 != undefined) {
+          checksums[filename] = getSHA1Checksum(file.base64)
         } else if (Object.keys(assetChecksums).includes(filename)) {
           checksums[filename] = assetChecksums[filename]
         }
--- a/editor/src/components/assets.ts
+++ b/editor/src/components/assets.ts
@@ -46,8 +46,13 @@ function getSHA1Checksum(contents: string | Buffer): string {
   return new sha1().update(contents).digest('hex')
 }
 
-export function inferGitBlobChecksum(buf: Buffer): string {
-  return getSHA1Checksum(Buffer.concat([Buffer.from(`blob ${buf.byteLength}\0`), buf]))
+export function inferGitBlobChecksum(buffer: Buffer): string {
+  // This function returns the same SHA1 checksum string that git would return for the same contents.
+  // Given the contents in the buffer variable, the final checksum is calculated by hashing
+  // a string built as "<prefix><contents>". The prefix looks like "blob <contents_length_in_bytes><null_character>".
+  const prefix = Buffer.from(`blob ${buffer.byteLength}\0`)
+  const wrapped = Buffer.concat([prefix, buffer])
+  return getSHA1Checksum(wrapped)
 }
 
 export function getProjectContentsChecksums(
--- a/editor/src/components/assets.ts
+++ b/editor/src/components/assets.ts
@@ -50,6 +50,7 @@ export function inferGitBlobChecksum(buffer: Buffer): string {
   // This function returns the same SHA1 checksum string that git would return for the same contents.
   // Given the contents in the buffer variable, the final checksum is calculated by hashing
   // a string built as "<prefix><contents>". The prefix looks like "blob <contents_length_in_bytes><null_character>".
+  // Ref: https://git-scm.com/book/en/v2/Git-Internals-Git-Objects
   const prefix = Buffer.from(`blob ${buffer.byteLength}\0`)
   const wrapped = Buffer.concat([prefix, buffer])
   return getSHA1Checksum(wrapped)
--- a/editor/src/components/assets.ts
+++ b/editor/src/components/assets.ts
@@ -47,7 +47,7 @@ function getSHA1Checksum(contents: string | Buffer): string {
 }
 
 export function inferGitBlobChecksum(buf: Buffer): string {
-  return getSHA1Checksum(Buffer.concat([Buffer.from(`blob ${buf.byteLength.toString()}\0`), buf]))
+  return getSHA1Checksum(Buffer.concat([Buffer.from(`blob ${buf.byteLength}\0`), buf]))
 }
 
 export function getProjectContentsChecksums(
--- a/editor/src/components/assets.ts
+++ b/editor/src/components/assets.ts
@@ -42,7 +42,7 @@ export function getAllProjectAssetFiles(
   return allProjectAssets
 }
 
-export function getSHA1Checksum(contents: string | Buffer): string {
+function getSHA1Checksum(contents: string | Buffer): string {
   return new sha1().update(contents).digest('hex')
 }
 
--- a/editor/src/components/assets.ts
+++ b/editor/src/components/assets.ts
@@ -65,16 +65,31 @@ export function getProjectContentsChecksums(
   const checksums: GithubChecksums = {}
   Object.keys(contents).forEach((filename) => {
     const file = contents[filename]
-    if (isTextFile(file)) {
-      checksums[filename] = getSHA1Checksum(file.fileContents.code)
-    } else if (isAssetFile(file) && file.base64 != undefined) {
-      checksums[filename] = getSHA1Checksum(file.base64)
-    } else if (isImageFile(file)) {
-      if (file.gitBlobSha) {
-        checksums[filename] = file.gitBlobSha
-      } else if (Object.keys(assetChecksums).includes(filename)) {
-        checksums[filename] = assetChecksums[filename]
-      }
+    if (file == null) {
+      return
+    }
+
+    switch (file.type) {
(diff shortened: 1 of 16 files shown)


editor/src/components/assets.ts
```diff
@@ -42,11 +42,18 @@ export function getAllProjectAssetFiles(
...
 
-export function getProjectContentsChecksums(tree: ProjectContentTreeRoot): GithubChecksums {
+export function inferGitBlobChecksum(buf: Buffer): string {
+  return getSHA1Checksum(Buffer.concat([Buffer.from(`blob ${buf.byteLength}\0`), buf]))
+}
+
+export function getProjectContentsChecksums(
+  tree: ProjectContentTreeRoot,
+  assetChecksums: GithubChecksums,
+): GithubChecksums {
```
bkrmendy, line 56: does this function handle each file type we have? or is it OK that some files won't be included in `checksums`

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2008,9 +2012,36 @@ export const UPDATE_FNS = {
     }
   },
   UPDATE_GITHUB_CHECKSUMS: (action: UpdateGithubChecksums, editor: EditorModel): EditorModel => {
+    const githubChecksums = action.checksums != null ? { ...action.checksums } : null
+    const assetChecksums = { ...editor.assetChecksums }
+    if (githubChecksums != null) {
+      // patch checksums
+      Object.keys(editor.assetChecksums).forEach((k) => {
+        if (githubChecksums[k] == undefined) {
```
bkrmendy, line 2020: nit: as far as I have seen, `== null` is the convention for null/undefined checks

editor/src/components/editor/server.ts
```diff
@@ -257,13 +257,13 @@ export async function saveAsset(
...
   imageId: string,
-): Promise<void> {
+): Promise<string | null> {
   try {
     return saveAssetRequest(projectId, fileType, base64, imageId)
   } catch (e) {
     // FIXME Client should show an error if server requests fail
     console.error(e)
-    return
+    return null
```
bkrmendy, line 266: is the `FIXME` here still up to date? as far as I know this function is either awaited or has a `.catch` at all call sites, so the try/catch might be removed from here (of course if this a proves to be a risky change it's not a big deal)

bkrmendy (approved):

editor/src/components/assets.ts
```diff
@@ -42,11 +42,18 @@ export function getAllProjectAssetFiles(
...
 }
 
-function getSHA1Checksum(contents: string): string {
+function getSHA1Checksum(contents: string | Buffer): string {
   return new sha1().update(contents).digest('hex')
 }
 
-export function getProjectContentsChecksums(tree: ProjectContentTreeRoot): GithubChecksums {
+export function inferGitBlobChecksum(buf: Buffer): string {
+  return getSHA1Checksum(Buffer.concat([Buffer.from(`blob ${buf.byteLength}\0`), buf]))
```
Rheeseyb, line 50: My eyes! Please can you add a comment (+ link if relevant) explaining this (I remember you mentioning it, but that will absolutely be forgotten in the near future)

editor/src/components/assets.ts
```diff
@@ -56,6 +63,12 @@ export function getProjectContentsChecksums(tree: ProjectContentTreeRoot): Githu
       checksums[filename] = getSHA1Checksum(file.fileContents.code)
     } else if (isAssetFile(file) && file.base64 != undefined) {
       checksums[filename] = getSHA1Checksum(file.base64)
+    } else if (isImageFile(file)) {
+      if (file.gitBlobSha) {
+        checksums[filename] = file.gitBlobSha
+      } else if (Object.keys(assetChecksums).includes(filename)) {
+        checksums[filename] = assetChecksums[filename]
+      }
```
Rheeseyb, line 71: I think this should become exhaustive and explicitly checked if possible (even if we do nothing in certain cases) so that the type checker can warn us if we introduce any new file types.

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2008,9 +2012,36 @@ export const UPDATE_FNS = {
...
+    }
+  },
+  SET_ASSET_CHECKSUM: (action: SetAssetChecksum, editor: EditorModel): EditorModel => {
+    const assetChecksums: GithubChecksums =
+      editor.assetChecksums == null ? {} : { ...editor.assetChecksums }
+    if (action.checksum == null) {
+      delete assetChecksums[action.filename]
+    } else {
+      assetChecksums[action.filename.replace(/^\.\//, '/')] = action.checksum
+    }
```
Rheeseyb, line 2040: Looks like there's a bug here, as we're using a potentially different key when updating a value in the object vs when deleting from it.

editor/src/components/editor/server.ts
```diff
@@ -281,12 +281,14 @@ export function assetToSave(fileType: string, base64: string, fileName: string):
...
+export async function saveAssets(
+  projectId: string,
+  assets: Array<AssetToSave>,
+): Promise<(string | null)[]> {
   const promises = assets.map((asset) =>
     saveAsset(projectId, asset.fileType, asset.base64, asset.fileName),
   )
-  await Promise.all(promises)
-  return
+  return await Promise.all(promises)
```
Rheeseyb, line 291: I believe this `await` is now redundant

bkrmendy: I think it's good to always `return await`, because the `await` "evaluates" the promise, and any errors thrown will show up in the function that returned the promise (omitting the `await` returns the promise, without actually "calling" it) (AFAIK)

Rheeseyb: ORLY? Interesting. I don't believe we're doing this anywhere, but that's very useful knowledge.

balazsbajorics: I believe we are back to this `await` being redundant :)

editor/src/components/assets.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/2933/commits/01521400b0a41bb5a0ab56d1ae07135a03603701 + https://github.com/concrete-utopia/utopia/pull/2933/commits/b282ef3ca384327bbb3e139968052f00a6606e41

editor/src/components/assets.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/2933/commits/3798a564ed4d59fce7c7275e4c16fad11dd2a678

editor/src/components/editor/actions/actions.tsx
ruggi: yep that's true, but I'm unable to find out who is not setting it correctly. maybe a pair of extra eyes could help here 👀

ruggi: actually found it → https://github.com/concrete-utopia/utopia/blob/65fa7b54943af7f97886464225e4054376c7a211/website-next/components/common/server.ts#L68-L70

editor/src/components/assets.ts
```diff
@@ -42,20 +42,54 @@ export function getAllProjectAssetFiles(
...
+      case 'TEXT_FILE':
+        checksums[filename] = getSHA1Checksum(file.fileContents.code)
+        break
+      case 'ASSET_FILE':
+        if (file.base64 != undefined) {
+          checksums[filename] = getSHA1Checksum(file.base64)
+        }
+        break
+      case 'IMAGE_FILE':
+        if (file.gitBlobSha) {
```
seanparsons, line 82:
```suggestion
        if (file.gitBlobSha != null) {
```
The same logic from here should also apply to `ASSET_FiLE` values as well.

editor/src/components/editor/server.ts
```diff
@@ -281,12 +281,14 @@ export function assetToSave(fileType: string, base64: string, fileName: string):
   }
 }
 
-export async function saveAssets(projectId: string, assets: Array<AssetToSave>): Promise<void> {
+export async function saveAssets(
+  projectId: string,
+  assets: Array<AssetToSave>,
+): Promise<(string | null)[]> {
```
seanparsons, line 287:
```suggestion
): Promise<Array<string | null>> {
```
This might be slightly clearer.

seanparsons (approved):

editor/src/components/assets.ts
ruggi: :+1: https://github.com/concrete-utopia/utopia/pull/2933/commits/f1a2dd824dec2a94cd48f6940380be822a48e9d4

editor/src/components/editor/server.ts
ruggi: oh that's true! 🙏 https://github.com/concrete-utopia/utopia/pull/2933/commits/e0fa1135b74406a4abc15194f70d06c0f3430b8e

editor/src/components/editor/server.ts
ruggi: so which one? :)

editor/src/components/editor/actions/actions.tsx
ruggi: fixed in this one https://github.com/concrete-utopia/utopia/pull/2933/commits/655d01e91cc6c98192d7834a0ff078ee91d6a647

Rheeseyb (approved):

editor/src/components/editor/server.ts
ruggi: FFR: as discussed, we'll stick with `return await`

PR #2936, opened 2022-11-24
Spike: number inputs on canvas

bkrmendy:
image


editor/src/components/canvas/controls/select-mode/flex-gap-control.tsx
```diff
@@ -305,3 +377,46 @@ function startInteraction(
...
+        onChange(currentValue)
+      }
+    },
+    [currentValue, onChange],
+  )
+
+  return (
+    <>
+      <input
+        onKeyDown={onKeyDown}
```
maltenuhn, line 413: If this input gets created / destroyed, you could add an `autofocus` here - there's a good chance it'll work as expected, so you can click & type.

PR #2937, opened 2022-11-24
Performance improvement for getAllTargetsAtPoint and getValidTargetAtPoint

gbalint:
**Problem:**
`getValidTargetAtPoint`, which is used for highlight and selection, uses `getAllTargetsAtPoint` to get the element under the mouse. `getAllTargetsAtPoint` get the html elements under a point from the browser, and then maps them to valid element paths. This calculation is quite cheap (~1ms), but it runs a lot, especially for highlight on mouse hover.

**Fix:**

1. When we only need the topmost element (`getValidTargetAtPoint`), it is unnecessary to calculate the closest valid element paths to all elements under point.
2. After we have the elements we are doing the calculation to get the valid element paths for all mouse events, but mostly we just calculate the same thing again and again (except when we move the mouse from an element to a different one). So it makes sense to memoize this calculation. The list of elements under a point is the same a lot of times (when we are hovering over the same top element), and even if we leave an element, sometimes we go back later. And in between these mouse move events nothing changes in the document. So it makes sense to use a `>1` maxsize, because we can go back and forth over elements with the mouse.

Measurements before the fix for the calculation:
Screenshot 2022-11-24 at 19 30 42

Measurements after the fix:

Screenshot 2022-11-24 at 19 28 07

### DIFF
--- a/editor/src/components/canvas/dom-lookup.ts
+++ b/editor/src/components/canvas/dom-lookup.ts
@@ -20,6 +20,7 @@ import Canvas, { TargetSearchType } from './canvas'
 import { CanvasPositions } from './canvas-types'
 import { AllElementProps } from '../editor/store/editor-state'
 import Utils from '../../utils/utils'
+import { memoize } from '../../core/shared/memoize'
 
 export function findParentSceneValidPaths(target: Element): Array<ElementPath> | null {
   const validPaths = getDOMAttribute(target, 'data-utopia-valid-paths')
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/dom-lookup.ts
```diff
@@ -149,6 +151,41 @@ export function getAllTargetsAtPoint(
     return []
   }
   const elementsUnderPoint = document.elementsFromPoint(point.x, point.y)
+  // TODO FIXME we should take the zero-sized elements from Canvas.getAllTargetsAtPoint, and insert them (in a correct-enough order) here. See PR for context https://github.com/concrete-utopia/utopia/pull/2345
+  return findFirstValidParentsForAllElements(validElementPathsForLookup, elementsUnderPoint)
+}
+
+const findFirstValidParentForSingleElement = memoize(findFirstValidParentForSingleElementUncached, {
```
balazsbajorics, line 158: what equality function does memoize use by default? Asking because I think the `Array<Elements>` is something we haven't tried memoizing before. Maybe it'd make sense to provide our own equality function here? just so it's very visible what happens under the hood?

balazsbajorics (approved):

editor/src/components/canvas/dom-lookup.ts
gbalint: We use our own `shallowEquals` by default: https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/memoize.ts#L10

And that uses our own `abstractEquals` which compares arrays by their values:
https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/equality-utils.ts#L11

https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/equality-utils.ts#L37

I can pass in `shallowEqual` by myself, but I think this has been deliberately provided as a good general default equality function for memoization, I don't really feel the need here to be honest.

Rheeseyb (approved):

bkrmendy (approved):

PR #2938, opened 2022-11-25
Don't overlay buttons over the file name in UTFB

bkrmendy:
## After
image

[Video](https://screenshot.click/05-57-80tg2-lf3a6.mp4)

## Before
image
(note that the `S3` marker is pushed out as well)

### DIFF
--- a/editor/src/components/navigator/navigator-item/expandable-indicator.tsx
+++ b/editor/src/components/navigator/navigator-item/expandable-indicator.tsx
@@ -25,7 +25,7 @@ export const ExpandableIndicator: React.FunctionComponent<
         color={props.selected ? 'on-highlight-main' : 'main'}
         style={{
           pointerEvents: props.visible ? 'all' : 'none',
-          opacity: props.visible ? 1 : 0,
+          visibility: props.visible ? 'visible' : 'hidden',
           cursor: 'pointer',
         }}
         onMouseDown={props.onMouseDown}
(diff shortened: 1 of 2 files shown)


editor/src/components/navigator/navigator-item/expandable-indicator.tsx
```diff
@@ -19,15 +19,18 @@ export const ExpandableIndicator: React.FunctionComponent<
...
-        />
-      ) : null}
+      <Icn
+        category='semantic'
+        type={`expansionarrow-${props.collapsed ? 'right' : 'down'}`}
+        color={props.selected ? 'on-highlight-main' : 'main'}
+        style={{
+          pointerEvents: props.visible ? 'all' : 'none',
+          opacity: props.visible ? 1 : 0,
+          cursor: 'pointer',
```
maltenuhn, line 29: Remind me - does this not trigger on elements if opacity is zero, or do you need to use `visibility: hidden` or `display: none`?

seanparsons (approved):

editor/src/components/navigator/navigator-item/expandable-indicator.tsx
bkrmendy: great idea, `opacity` does trigger, so `visibility` would be much better, added it

Rheeseyb (approved):

PR #2940, opened 2022-11-25
Flex gap control improvements

bkrmendy:
## Problems
- [x] Flex gap controls aren't positioned in the right place if the gap is adjusted via the inspector
- [x] Flex gap handle is shown during drag


maltenuhn:
(I've just tested this on the branch, it seems to fix the previously buggy behaviour. 🎉)

bkrmendy:
> (I've just tested this on the branch, it seems to fix the previously buggy behaviour. 🎉)

(🙏)

balazsbajorics (approved):

Rheeseyb (approved):

PR #2946, opened 2022-11-29
spike: overridden props

bkrmendy:
override props in parsed model, don't write them back to source


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -546,6 +547,7 @@ export function renderComponentUsingJsxFactoryFunction(
       throw new Error(`Unable to find factory function ${factoryFunctionName} in scope.`)
     }
   }
+  // console.log(fixedProps)
```
ruggi, line 550: 👋

PR #2957, opened 2022-12-01
Bugs 

bkrmendy:
## Problems:
When setting the layout system to `flex`...
1. `flexBasis` values are applied
2. `top`/`left` values are set to `infinity` when the element has no parent

## Fix:
1. Remove setting `flexBasis`
2. Check if the parent's frame contain `Infinity` values, and use a fallback frame if they do

## `TODO`
- [ ] Tests

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -606,7 +606,9 @@ export function updateFramesOfScenesAndComponents(
         parentFrame = optionalParentFrame
       }
       const parentOffset =
-        parentFrame == null
+        parentFrame == null ||
+        Math.abs(parentFrame.x) === Infinity ||
+        Math.abs(parentFrame.y) === Infinity
           ? ({ x: 0, y: 0 } as CanvasPoint)
           : ({ x: parentFrame.x, y: parentFrame.y } as CanvasPoint)
 
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -18,8 +18,6 @@ import {
 } from '../../core/layout/layout-helpers-new'
 import {
   maybeSwitchLayoutProps,
-  PinningAndFlexPoints,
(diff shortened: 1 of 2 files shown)


editor/src/core/layout/layout-utils.ts
```diff
@@ -286,12 +286,12 @@ function getLayoutFunction(
 
 export const PinningAndFlexPoints: Array<LayoutPinnedProp | 'flexBasis'> = [
   ...LayoutPinnedProps,
-  'flexBasis',
+  // 'flexBasis',
```
Rheeseyb, line 289:
```suggestion
```

editor/src/core/layout/layout-utils.ts
```diff
@@ -286,12 +286,12 @@ function getLayoutFunction(
...
 ]
 
 export const PinningAndFlexPointsExceptSize: Array<LayoutPinnedProp | 'flexBasis'> = [
   ...LayoutPinnedProps.filter((p) => p !== 'width' && p !== 'height'),
-  'flexBasis',
+  // 'flexBasis',
```
Rheeseyb, line 294:
```suggestion
```

Rheeseyb (approved):

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -606,7 +606,9 @@ export function updateFramesOfScenesAndComponents(
         parentFrame = optionalParentFrame
       }
       const parentOffset =
-        parentFrame == null
+        parentFrame == null ||
+        Math.abs(parentFrame.x) === Infinity ||
```
bkrmendy, line 610: actual solution: special case the canvas root

editor/src/core/layout/layout-utils.ts
bkrmendy, line 289: actual solution: remove the whole function

seanparsons (approved):

seanparsons:
I'm surprised that no tests were broken.

PR #2958, opened 2022-12-02
eye dropper

bkrmendy:
## Problem
Background colors can only be picked via the color picker

## Fix
Use the [Eyedropper API](https://developer.mozilla.org/en-US/docs/Web/API/EyeDropper_API) to make picking a background color easier.

If there are one or more elements selected, the color picker can be opened by pressing `ctrl + C` (same as in Sketch). Once a color has been picked, it is set as the `backgroundColor` prop of all selected elements. The interaction can be aborted with `Esc`.

[Demo](https://screenshot.click/hack.mp4)

### DIFF
--- a/editor/src/components/editor/global-shortcuts.ts
+++ b/editor/src/components/editor/global-shortcuts.ts
@@ -86,7 +86,7 @@ import {
   TOGGLE_FOCUSED_OMNIBOX_TAB,
   FOCUS_CLASS_NAME_INPUT,
   INSERT_DIV_SHORTCUT,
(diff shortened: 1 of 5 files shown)


maltenuhn:
What needs to happen here so this can be merged?
And does it fail gracefully (no-op, not crash) on Safari?

bkrmendy:
Pressing ctrl+c in any browser that doesn't support the eyedropper API results in a no-op (the shortcut doesn't do anything)

balazsbajorics (approved):

PR #2960, opened 2022-12-05
perf: do not run `createExecutionScope` unnecessarily

bkrmendy:
## Problem
`createExecutionScope` runs on every project contents change, which happens at every mouse move event during drag

## Solution
Memoize `createExecutionScope` and only re-run it when `ElementsToRerenderGLOBAL` is in `rerender-all-elements` mode


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-execution-scope.tsx
```diff
@@ -67,6 +55,7 @@ export function createExecutionScope(
   scope: MapLike<any>
   topLevelJsxComponents: Map<string | null, UtopiaJSXComponent>
   requireResult: MapLike<any>
+  jsxFactoryFunction: string | null
```
balazsbajorics, line 58: seems like this is a leftover, and never used. could you clean it up please?

seanparsons (approved):
Does this still work when someone reparents an element?

Rheeseyb:
> Does this still work when someone reparents an element?

Specifically when an element is reparented across files by dragging it on the canvas.

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -350,33 +331,37 @@ export const UiJsxCanvas = React.memo<UiJsxCanvasPropsWithErrorCallback>((props)
...
+  const projectContentsForRequireFn = maybeOldProjectContents.current
+  const requireFn = React.useMemo(
+    () => curriedRequireFn(projectContentsForRequireFn),
+    [curriedRequireFn, projectContentsForRequireFn],
+  )
+
+  const resolve = React.useMemo(
+    () => curriedResolveFn(projectContentsForRequireFn),
+    [curriedResolveFn, projectContentsForRequireFn],
+  )
```
Rheeseyb, line 353: We definitely need an important note in the comment here outlining that this means the require and resolve functions will be stale if `ElementsToRerenderGLOBAL.current !== 'rerender-all-elements'`, so if we ever design a system for updating javascript blocks that also happens to run without re-rendering all elements it would require changes here (perhaps unlikely, but that would be _horrendous_ to debug)

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -350,23 +331,46 @@ export const UiJsxCanvas = React.memo<UiJsxCanvasPropsWithErrorCallback>((props)
   )
   useClearSpyMetadataOnRemount(props.mountCount, props.domWalkerInvalidateCount, metadataContext)
 
+  const elementsToRerenderRef = React.useRef(ElementsToRerenderGLOBAL.current)
+  if (
+    ElementsToRerenderGLOBAL.current === 'rerender-all-elements' ||
+    elementsToRerenderRef.current === 'rerender-all-elements' || // once we get here, we know that `ElementsToRerenderGLOBAL.current` is an array, and `elementsToRerenderRef.current` isn't
+    !pathArraysEqual(ElementsToRerenderGLOBAL.current, elementsToRerenderRef.current) // once we get here, we know that neither `ElementsToRerenderGLOBAL.current` and `elementsToRerenderRef.current` are arrays
```
Rheeseyb, line 338: I believe you meant to say that we know they're not equal?

bkrmendy: just a typo (fixed in https://github.com/concrete-utopia/utopia/pull/2960/commits/0c26f3e1f1fa49aaae14b12ab3a8394165df0b22)

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -775,3 +793,16 @@ const CanvasContainer = React.forwardRef<
   )
 })
 CanvasContainer.displayName = 'CanvasContainer'
+
+function pathArraysEqual(one: Array<ElementPath>, other: Array<ElementPath>) {
+  if (one.length != other.length) {
+    return false
+  }
+
+  for (const path of one) {
```
Rheeseyb, line 802: There is a potential bug here as we're not enforcing uniqueness on either of the arrays, so e.g. if we called this with `pathArraysEqual([path1, path1], [path1, path2])` it would return. In this specific use case that would be fine, since we only care that all of the values of `ElementsToRerenderGLOBAL.current` are contained in `elementsToRerenderRef.current`, and not that they contain the exact same values (I believe?), so I think it's not necessarily a problem to leave it as it is (so that we don't have to enforce uniqueness here), but I'd suggest we call that out in a comment.

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -350,23 +331,45 @@ export const UiJsxCanvas = React.memo<UiJsxCanvasPropsWithErrorCallback>((props)
   )
   useClearSpyMetadataOnRemount(props.mountCount, props.domWalkerInvalidateCount, metadataContext)
 
+  const elementsToRerenderRef = React.useRef(ElementsToRerenderGLOBAL.current)
+  const shouldRerenderRef = React.useRef(false)
+  shouldRerenderRef.current =
+    ElementsToRerenderGLOBAL.current === 'rerender-all-elements' ||
+    elementsToRerenderRef.current === 'rerender-all-elements' || // once we get here, we know that `ElementsToRerenderGLOBAL.current` is an array, and `elementsToRerenderRef.current` isn't
```
balazsbajorics, line 338:
```suggestion
    elementsToRerenderRef.current === 'rerender-all-elements' || // TODO this means the first drag frame will still be slow, figure out a nicer way to immediately switch to true. probably this should live in a dedicated a function
```

balazsbajorics (approved):

PR #2961, opened 2022-12-05
Enable React 18 mode at last

balazsbajorics:
**Problem:**
We were running React in "React 17 compatibility mode" by using the old ReactDOM.render API instead of createRoot. As a first part of performance investigation, it's probably not the worst idea to bump react to the latest version, and enable proper react 18 mode.

The reason I did not switch to createRoot earlier was some bug in testing-library which failed a bunch of tests if we ran it in `{ legacyRoot: true }` mode (which is their API to use ReactDOM.render thus forcing react to React 17 compatibility mode).

**Fix:**
Bump React to 18.2.0 and testing-library/react to 13.4.0.
Replace ReactDOM.render calls with createRoot.
Remove `{ legacyRoot: true }` flags from the tests.

**Note**
Seems to work all right, but now the tests get an incredibly annoying `Warning: The current testing environment is not configured to support act` warning logged, and I already lost 2 hours trying to figure out how to make them go away. Hint: `globalThis.IS_REACT_ACT_ENVIRONMENT = true;` DOES NOT fix them, [even though that's what the docs say...](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#configuring-your-testing-environment)


editor/patches/react+18.2.0.patch
```diff
@@ -0,0 +1,408 @@
...
+-          var ReactVersion = '18.2.0';
++          // TODO: this is special because it gets imported during build.
++//
++// TODO: 18.0.0 has not been released to NPM;
++// It exists as a placeholder so that DevTools can support work tag changes between releases.
++// When we next publish a release, update the matching TODO in backend/renderer.js
++// TODO: This module is used both by the release scripts and to expose a version
++// at runtime. We should instead inject the version number as part of the build
++// process, and use the ReactVersions.js module as the single source of truth.
++var ReactVersion = '18.1.0';
```
seanparsons, line 148: This is ringing alarm bells.

seanparsons (approved):
Approving with some uncertainty about what the patch is actually changing in React.

balazsbajorics:
I'm closing this because I'm 100% sure the home-compiled version of react does not match the released 18.2.0. since I confirmed there were no performance benefits, this is not urgent

PR #2963, opened 2022-12-06
ux(github) - only show github settings to logged-in users

maltenuhn:
**Problem:**
- We show non-authenticated users the github pane with the "sign up" CTA - but if you're not signed up, you can't then authorise github to load/save projects. And we don't inform you of that limitation.

**Fix:**
- Only show the contents of the github pane if you're authenticated with Utopia already

**Notes:**
This is a quick fix, not a systemic one integrating the authentication systems.

### DIFF
--- a/editor/src/components/navigator/left-pane/github-pane/index.tsx
+++ b/editor/src/components/navigator/left-pane/github-pane/index.tsx
@@ -14,6 +14,7 @@ import {
   updateProjectWithBranchContent,
   useGithubFileChanges,
 } from '../../../../core/shared/github'
+import { User } from '../../../../uuiui-deps'
 import { startGithubAuthentication } from '../../../../utils/github-auth'
 import { unless, when } from '../../../../utils/react-conditionals'
 import {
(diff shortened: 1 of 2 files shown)


editor/src/components/navigator/left-pane/github-pane/index.tsx
```diff
@@ -951,8 +953,12 @@ const PullRequestBlock = () => {
 export const GithubPane = React.memo(() => {
   const githubUser = useEditorState(
     (store) => store.editor.githubData.githubUserDetails,
-    'Github user details',
+    'GithubPane githubUser',
   )
+  const isLoggedIn = useEditorState((store) => {
```
balazsbajorics, line 958: @maltenuhn I went in and split these into two useEditorState hooks. the reason is that the original githubUser hook may benefit from shallow equality checks, so putting it one level deeper may blow the equality and thus trigger rerenders of this component.

balazsbajorics (approved):

PR #2964, opened 2022-12-06
ux(title bar) - disappearing sign-in button, projects page navigation

maltenuhn:
**Problem:**
The Title Bar has a couple of usability issues:
- The sign-in button in the title bar "disappears" on hover
- You can't navigate back to the `/projects` page

**Fix:**
- Use the right invocation chant to prohibit the sign-in button returning to its ethereal, translucent form
- Make the avatar a link to `/projects` and hope our saved-state detection is good enough. Maybe this should just open in a new tab tho?

### DIFF
--- a/editor/src/components/titlebar/title-bar.tsx
+++ b/editor/src/components/titlebar/title-bar.tsx
@@ -180,11 +180,21 @@ const TitleBar = React.memo(() => {
       <div style={{ flex: '0 0 0px', paddingRight: 8 }}>
         {unless(
           loggedIn,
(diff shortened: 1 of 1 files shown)


bkrmendy:
opening it in a new tab is a good idea, I'd add that if it's super easy

bkrmendy (approved):

maltenuhn:
@bkrmendy 👍 : updated with `target="_blank"`

balazsbajorics (approved):

PR #2973, opened 2022-12-07
Nine-block control & inspector strategies POC

bkrmendy:
08-31-7zbki-0nu5e

## Problem
The inspector in its current for can only manipulate properties on a one-by-one basis. The drawback of this approach is that there exist some interactions, where multiple props have to be adjusted (such as converting a container to/from flex or changing the flex direction). 

## Solution
To fix this, we're trying a new approach that is geared towards compositionality/customisability. In this new approach, an interaction through a given control is run through a cascade of inspector strategies (inspired by the canvas strategies system that we already have). These inspector strategies detect if they are applicable, and if they are, they return an array of editor actions that are dispatched as the result of that UI interaction.

## This PR
This PR is a PoC/first step in the approach outlined above. It introduces
- a "nine-block" control for setting `align-items` and `justify-content` on a flex container
- the inspector strategies system

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -5163,8 +5163,9 @@ export const UPDATE_FNS = {
       imageDragSessionState: action.imageDragSessionState,
     }
   },
-  APPLY_COMMANDS: (action: ApplyCommandsAction, editor: EditorModel): EditorModel =>
-    foldAndApplyCommandsSimple(editor, action.commands),
+  APPLY_COMMANDS: (action: ApplyCommandsAction, editor: EditorModel): EditorModel => {
+    return foldAndApplyCommandsSimple(editor, action.commands)
+  },
 }
 
 /** DO NOT USE outside of actions.ts, only exported for testing purposes */
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -314,6 +314,7 @@ import {
   SetHoveredView,
   ClearHoveredViews,
   SetAssetChecksum,
+  ApplyCommandsAction,
 } from '../action-types'
 import { defaultSceneElement, defaultTransparentViewElement } from '../defaults'
 import { EditorModes, isLiveMode, isSelectMode, Mode } from '../editor-modes'
@@ -5162,6 +5163,8 @@ export const UPDATE_FNS = {
       imageDragSessionState: action.imageDragSessionState,
     }
   },
+  APPLY_COMMANDS: (action: ApplyCommandsAction, editor: EditorModel): EditorModel =>
+    foldAndApplyCommandsSimple(editor, action.commands),
 }
 
 /** DO NOT USE outside of actions.ts, only exported for testing purposes */
--- a/editor/src/components/inspector/nine-block-controls.tsx
+++ b/editor/src/components/inspector/nine-block-controls.tsx
@@ -3,20 +3,17 @@
 
 import { jsx } from '@emotion/react'
 import React from 'react'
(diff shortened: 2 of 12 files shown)


editor/src/utils/feature-switches.ts
```diff
@@ -27,6 +29,7 @@ let FeatureSwitches: { [feature in FeatureName]: boolean } = {
   'Re-parse Project Button': !(PRODUCTION_CONFIG as boolean),
   'Performance Test Triggers': !(PRODUCTION_CONFIG as boolean),
   'Canvas Strategies Debug Panel': false,
+  'Nine block control': true,
```
bkrmendy, line 32: `TODO` before merge: set to false

editor/src/components/editor/actions/actions.tsx
```diff
@@ -5162,6 +5163,8 @@ export const UPDATE_FNS = {
       imageDragSessionState: action.imageDragSessionState,
     }
   },
+  APPLY_COMMANDS: (action: ApplyCommandsAction, editor: EditorModel): EditorModel =>
+    foldAndApplyCommandsSimple(editor, action.commands),
```
balazsbajorics, line 5167: super minor: please use the { return ... } syntax, so it's easier to inject a console.log for debugging etc

balazsbajorics (approved):

editor/src/utils/feature-switches.ts
enidemi: I think a feature switch for 1 extra inspector section is not really necessary

editor/src/components/inspector/nine-block-controls.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  const [hovered, setHovered] = React.useState<number>(-1)
+
+  const [flexJustifyContent, flexAlignment] = detectFlexAlignJustifyContent(
+    metadata,
+    selectedViews[0],
+  )
+
+  const setAlignItemsJustifyContent = React.useCallback(
+    (intededFlexAlignment: FlexAlignment, intededJustifyContent: FlexJustifyContent) => {
```
enidemi, line 71: typo :)

editor/src/components/inspector/nine-block-controls.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const dispatch = useEditorState((store) => store.dispatch, 'NineBlockControl dispatch')
+  const metadata = useEditorState((store) => store.editor.jsxMetadata, 'NineBlockControl metadata')
+  const selectedViews = useEditorState(
+    (store) => store.editor.selectedViews,
+    'NineBlockControl selectedViews',
+  )
+
+  const [hovered, setHovered] = React.useState<number>(-1)
+
+  const [flexJustifyContent, flexAlignment] = detectFlexAlignJustifyContent(
```
enidemi, line 65: please add a TODO to detect if it's set via css only or code or jsx prop

editor/src/components/inspector/nine-block-controls.tsx
```diff
@@ -0,0 +1,162 @@
...
+      const strategies = setFlexAlignJustifyContentStrategies(
+        intededFlexAlignment,
+        intededJustifyContent,
+      )
+      runStrategies(dispatch, metadata, selectedViews, strategies)
+    },
+    [dispatch, metadata, selectedViews],
+  )
+
+  if (selectedViews.length === 0) {
```
enidemi, line 81: I think it should be visible only for flex layouts

editor/src/components/inspector/nine-block-controls.tsx
bkrmendy: thanks!

enidemi (approved):

PR #2975, opened 2022-12-08
fix(reparent) reorder inside the parent should not trigger reparent

enidemi:
**Problem:**
It's very easy to reparent to be the sibling of the direct parent in a flex in flex layout while only reordering between current siblings, because the cursor position already matches the parent padded area.

**Fix:**
Filter out ancestors while dragging inside the parent area. Added 2 tests for this.

**Commit Details:**
- fix `getReparentTargetUnified`
- tests

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
@@ -55,11 +55,28 @@ export function getReparentTargetUnified(
     allowSmallerParent,
   )
 
+  // with current parent under cursor filter ancestors from reparent targets
+  const currentParentUnderCursor =
+    reparentSubjects.type === 'EXISTING_ELEMENTS'
+      ? validTargetparentsUnderPoint.find((targetParent) =>
+          EP.isParentOf(targetParent, reparentSubjects.elements[0]),
+        ) ?? null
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
```diff
@@ -55,11 +55,28 @@ export function getReparentTargetUnified(
     allowSmallerParent,
   )
 
+  // with current parent under cursor filter ancestors from reparent targets
+  const currentParentUnderCursor =
+    reparentSubjects.type === 'EXISTING_ELEMENTS'
+      ? validTargetparentsUnderPoint.find((targetParent) =>
+          EP.isParentOf(targetParent, reparentSubjects.elements[0]),
+        ) ?? null
+      : null
```
Rheeseyb, line 64: It looks like this would make more sense as a boolean since we do nothing with this value other than check that it exists

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
balazsbajorics: it is used in line 68

balazsbajorics (approved):

PR #2976, opened 2022-12-09
Flex direction toggle

bkrmendy:
09-52-mckmz-z724w

Inspector strategies-backed flex-direction toggle

### DIFF
--- a/editor/src/components/inspector/flex-direction-control.tsx
+++ b/editor/src/components/inspector/flex-direction-control.tsx
@@ -5,11 +5,8 @@ import { Icons, useColorTheme } from '../../uuiui'
 import { EditorDispatch } from '../editor/action-types'
 import { useEditorState, useRefEditorState } from '../editor/store/store-hook'
 import { FlexDirection } from './common/css-utils'
-import {
-  filterKeepFlexContainers,
-  metadataSelector,
-  selectedViewsSelector,
-} from './inspector-common'
+import { metadataSelector, selectedViewsSelector } from './inpector-selectors'
+import { filterKeepFlexContainers } from './inspector-common'
 import {
   removeFlexDirectionStrategies,
   runStrategies,
--- a/editor/src/components/inspector/flex-direction-control.tsx
+++ b/editor/src/components/inspector/flex-direction-control.tsx
@@ -1,0 +1,99 @@
+import React from 'react'
+import { ElementInstanceMetadataMap } from '../../core/shared/element-template'
+import { ElementPath } from '../../core/shared/project-file-types'
+import { Icons, UNSAFE_getIconURL, useColorTheme } from '../../uuiui'
+import { EditorDispatch } from '../editor/action-types'
+import { useEditorState } from '../editor/store/store-hook'
+import { FlexDirection } from './common/css-utils'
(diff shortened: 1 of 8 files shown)


editor/src/components/inspector/flex-section.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { useEditorState } from '../editor/store/store-hook'
+import { FlexDirectionToggle } from './flex-direction-control'
+import { detectFlexDirection } from './inspector-common'
+import { NineBlockControl } from './nine-block-controls'
+
+export const FlexSection = React.memo(() => {
+  const metadata = useEditorState((store) => store.editor.jsxMetadata, 'FlexSection metadata')
```
balazsbajorics, line 8: store.editor.jsxMetadata is expected to change almost every frame during an interaction, so if your selector returns that, it will trigger a re-render of FlexSection for every frame.

instead, it is better to move detectFlexDirection _inside_ the selector. then useEditorState's built-in shallow equality check will compare the previous flexDirection with the current flexDirection, which means it will only trigger a re-render if flexDirection _actually_ changes. 

it's also important to keep in mind that the selector needs to be very very fast, if you have to make a slow selector, there's tools to mitigate that like reselect, but the first approximation is that selectors should almost take no time to run!

balazsbajorics (approved):

editor/src/components/inspector/flex-section.tsx
bkrmendy: fixed!

editor/src/components/inspector/flex-direction-control.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const colorTheme = useColorTheme()
+
+  const handleColumnClick = React.useCallback(
+    (e: React.MouseEvent) =>
+      maybeSetFlexDirection(
+        dispatch,
+        metadataRef.current,
+        selectedViewsRef.current,
+        e.button === 2 ? null : 'column',
```
Rheeseyb, line 36: Why not `e.button === 0 ? 'column' : null,`?

editor/src/components/inspector/inspector-common.tsx
```diff
@@ -73,3 +86,24 @@ export function filterKeepFlexContainers(
...
+}
+
+export const isFlexColumn = (flexDirection: FlexDirection): boolean =>
+  flexDirection.startsWith('column')
+
+export const metadataSelector = (store: EditorStorePatched): ElementInstanceMetadataMap =>
+  store.editor.jsxMetadata
+
+export const selectedViewsSelector = (store: EditorStorePatched): ElementPath[] =>
+  store.editor.selectedViews
```
Rheeseyb, line 109: These should probably be in a file for common selectors rather than Inspector specific ones

Rheeseyb (approved):

editor/src/components/inspector/flex-direction-control.tsx
bkrmendy: no reason, I refactored it to check the left mouse button

editor/src/components/inspector/inspector-common.tsx
bkrmendy: moved!

PR #2977, opened 2022-12-09
Fix race conditions around extension starting

Rheeseyb:
**Problem:**
There are some race conditions when starting the VS Code extension that can cause the file system provider to attempt to read files before they have been written, or break the messaging system that transfers instructions between VSCode and Utopia

**Fix:**
- Add some extra safety checking when attempting to read the last consumed message, as the value can be deleted during startup between checking that it exists and reading the value (as we clear the mailbox on startup)
- Wait for the project files to be written to the FS before registering the FS provider part of the extension, as VS Code will patiently wait for a suitable FS provider to be registered if none are available yet

### DIFF
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -143,7 +143,7 @@ import {
 } from './store-deep-equality-instances'
 import { forceNotNull } from '../../../core/shared/optional-utils'
 import * as EP from '../../../core/shared/element-path'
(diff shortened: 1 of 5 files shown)


balazsbajorics (approved):

enidemi (approved):

Rheeseyb:
Closing as I've discovered a possibly related bug

seanparsons (approved):

PR #2981, opened 2022-12-12
Bump reselect, remove factory

balazsbajorics:
**Problem:**
If you wanted to pass a parameter to reselect, you had to make ugly hacks, such as the `navigatorItemWrapperSelectorFactory`, which was a function that returned a selector. Since selectors are memoized _by instance_ (not globally), you had to be very careful to then wrap this factory in a useMemo. All in all, too finicky and hard to make robust.

**Fix:**
A new version of reselect supports parametrized selectors! Yay!

**Commit Details:**
- Made a .spec.ts tets file that demonstrates how reselect memoization works
- bumped reselect
- killed navigatorItemWrapperSelectorFactory

### DIFF
--- a/editor/src/components/editor/store/reselect-examples.spec.tsx
+++ b/editor/src/components/editor/store/reselect-examples.spec.tsx
@@ -92,7 +92,7 @@ describe('Reselect Investigation', () => {
     selectorFn()(store)
     expect(numberOfTimesCalled).toBe(4)
 
-    // uh oh!!! be very careful when craeting a selector with a curry function
+    // uh oh!!! be very careful when creating a selector with a curry function
   })
 
   it('selectors that take an argument are memoized', () => {
--- a/editor/src/components/editor/store/reselect-examples.spec.tsx
+++ b/editor/src/components/editor/store/reselect-examples.spec.tsx
@@ -1,0 +1,196 @@
+import { createSelector } from 'reselect'
+
+interface ExampleStore {
+  exNumber: number
+  exString: string
+}
+
+describe('Reselect Investigation', () => {
+  it('selectors are memoized', () => {
+    let numberOfTimesCalled = 0
+    const selector = createSelector(
+      (store: ExampleStore) => store.exNumber,
+      (store: ExampleStore) => store.exString,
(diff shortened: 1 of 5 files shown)


editor/src/components/editor/store/reselect-examples.spec.tsx
```diff
@@ -0,0 +1,196 @@
...
+    selector1(store)
+    selector1(store)
+    selector2(store)
+    selector2(store)
+
+    expect(numberOfTimesCalled.s1).toBe(1)
+    expect(numberOfTimesCalled.s2).toBe(1)
+  })
+
+  it('!!! WARNING: accidentally creating a new will not be memoized properly!!!!!', () => {
```
Rheeseyb, line 70:
```suggestion
  it('!!! WARNING: accidentally creating a new selector will not be memoized properly!!!!!', () => {
```

Also, It probably makes sense to move this to the very start of the test to increase visibility (though I'm sceptical that many will visit this file very often anyway, so this warning would probably make the most sense inside a comment in a share selectors file)

editor/src/components/editor/store/reselect-examples.spec.tsx
```diff
@@ -0,0 +1,196 @@
...
+    selectorFn()(store)
+    expect(numberOfTimesCalled).toBe(1)
+    selectorFn()(store)
+    expect(numberOfTimesCalled).toBe(2)
+    selectorFn()(store)
+    expect(numberOfTimesCalled).toBe(3)
+    selectorFn()(store)
+    expect(numberOfTimesCalled).toBe(4)
+
+    // uh oh!!! be very careful when craeting a selector with a curry function
```
Rheeseyb, line 95:
```suggestion
    // uh oh!!! be very careful when creating a selector with a curry function
```

Rheeseyb (approved):

Rheeseyb:
Nice work, and good idea adding the tests. I think it's worth adding a comment that points to those tests in `store-hook.tsx` or another shared selectors file to serve as a reminder

enidemi (approved):

PR #2982, opened 2022-12-12
detect flex direction/justify-content/align-items based on all selected elements

bkrmendy:
## Problem
In the inspector-strategies backed inspector controls, `flex-direction`, `justify-content`, `align-items` and `display(: flex)` were determined only by inspecting the first selected element, disregarding the rest. This lead to misleading UI states.

## Fix
Check all selected elements for the props in question, and
- if they are set to the same value, display that value on the UI
- otherwise, display a default value on the UI

## After
null_state

## Before
image

### DIFF
--- a/editor/src/components/inspector/inspector-common.tsx
+++ b/editor/src/components/inspector/inspector-common.tsx
@@ -1,7 +1,6 @@
 import { MetadataUtils } from '../../core/model/element-metadata-utils'
 import { ElementInstanceMetadataMap } from '../../core/shared/element-template'
 import { ElementPath } from '../../core/shared/project-file-types'
-import { EditorState, EditorStorePatched } from '../editor/store/editor-state'
 import { FlexDirection } from './common/css-utils'
 
 export type StartCenterEnd = 'flex-start' | 'center' | 'flex-end'
@@ -130,6 +129,15 @@ export function detectFlexDirection(
   return allElemsEqual(allDetectedMeasurement, (l, r) => l === r)
     ? allDetectedMeasurement[0]
     : DefaultFlexDirection
+}
+
+export function detectAreElementsInFlexLayout(
+  metadata: ElementInstanceMetadataMap,
+  elementPaths: Array<ElementPath>,
+): boolean {
+  return elementPaths.every((path) =>
+    MetadataUtils.isFlexLayoutedContainer(MetadataUtils.findElementByElementPath(metadata, path)),
+  )
 }
 
 export const isFlexColumn = (flexDirection: FlexDirection): boolean =>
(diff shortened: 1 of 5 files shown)


balazsbajorics (approved):

editor/src/components/inspector/inspector-common.tsx
```diff
@@ -61,23 +60,42 @@ function stringToFlexDirection(str: string | null): FlexDirection | null {
...
-export function detectFlexAlignJustifyContent(
+const DefaultJustifyContentFlexAlign: [FlexJustifyContent, FlexAlignment] = [
+  'flex-start',
+  'flex-start',
+]
+
+function detectFlexAlignJustifyContentOne(
   metadata: ElementInstanceMetadataMap,
   elementPath: ElementPath,
 ): [FlexJustifyContent, FlexAlignment] {
```
Rheeseyb, line 71: I must have overlooked or missed this before, and I don't want to block this PR on something unrelated, but please can we replace this return value with an object

editor/src/components/inspector/inspector-common.tsx
```diff
@@ -87,17 +105,48 @@ export function filterKeepFlexContainers(
...
+    ? allDetectedMeasurement[0]
+    : DefaultFlexDirection
+}
+
+export function detectAreElementsInFlexLayout(
+  metadata: ElementInstanceMetadataMap,
+  elementPaths: Array<ElementPath>,
+): boolean {
+  return elementPaths.every((path) =>
+    MetadataUtils.isFlexLayoutedContainer(MetadataUtils.findElementByElementPath(metadata, path)),
```
balazsbajorics, line 139: this is actually wrong! there's `MetadataUtils.isParentFlexLayoutedContainerForElement()` that should be used if you are talking about whether the element at path is _in_ a flex parent

but if you want to detect if it's in a flex parent -and- participates in the layout, we have the incredibly named `MetadataUtils.isParentYogaLayoutedContainerForElementAndElementParticipatesInLayout`!

Rheeseyb (approved):

editor/src/components/inspector/inspector-common.tsx
bkrmendy: done! (https://github.com/concrete-utopia/utopia/pull/2982/commits/39a3f3293b12d49d8beedcc159aaf27c091a025b)

editor/src/components/inspector/inspector-common.tsx
bkrmendy: (discussed this and it turned out that the name of the function was misleading, I renamed it to `detectAreElementsFlexContainers`)

PR #2985, opened 2022-12-12
unifying ImportInfo and ImportedFromWhereResult

balazsbajorics:
We had two semi-overlapping sources of information. The ElementInstanceMetadata has ImportInfo, and `importedFromWhere` returned with `ImportedFromWhereResult`. 

I realized that `findUnderlyingTargetComponentImplementation` (a function that is used _a lot_ during each update) can become quicker if it uses information straight from the metadata vs looking up import information from the projectContents. _(separate PR coming after this one!)_ 

This PR unifies the two types with a superset, which means places in the codebase that are using `importedFromWhere` can now alternatively look up the import info from the JsxMetadata. 

importedFromWhere itself is _not_ a slow function, but to give it the parameters it needs, you have to traverse ProjectContents, often using `normalisePathToUnderlyingTarget`, and _that_ part can be skipped entirely if you have metadata + element path at your disposal.


bkrmendy (approved):

enidemi (approved):

Rheeseyb (commented):
👍

PR #2987, opened 2022-12-13
findUnderlyingTargetComponentImplementationFromImportInfo

balazsbajorics:
This PR introduces `findUnderlyingTargetComponentImplementationFromImportInfo` which takes a projectContents and an importInfo and returns the component implementation the importInfo is referring to. I also changed `findUnderlyingTargetComponentImplementation` to use the new function instead of calling `lookupElementImport` and subsequent calls to `normalisePathToUnderlyingTarget`. 

The new function should already be faster than the one using `lookupElementImport`, but the main advantage is that in a follow up PR I will change call sites from using `findUnderlyingTargetComponentImplementation` to `findUnderlyingTargetComponentImplementationFromImportInfo`. If the call sites have access to jsxMetadata, they can get the importInfo "for free", without having to call `normalisePathToUnderlyingTarget` and `importedFromWhere`. 

This is very important, because a lot of selectors on the hot path want to access the underlying component information which today involves a whole bunch of tree walking and module lookups. 
With this change, I can make a subsequent PR that replaces those selectors with cheaper ones that call `findUnderlyingTargetComponentImplementationFromImportInfo` directly with importInfo from metadata.


editor/src/components/custom-code/code-file.ts
```diff
@@ -570,6 +572,38 @@ export function normalisePathToUnderlyingTargetForced(
...
+    return null
+  }
+
+  const projectContentsFlattened = treeToContents(projectContents)
+
+  const variableName =
+    importInfo.type === 'SAME_FILE_ORIGIN' ? importInfo.variableName : importInfo.exportedName
+
+  // we have to find the element based on the top level name
+  const file = projectContentsFlattened[importInfo.filePath]
```
seanparsons, line 589:
```suggestion
  const file = getContentsTreeFileFromString(projectContents, importInfo.filePath)
```

Then you don't need to invoke `treeToContents`.

seanparsons (approved):

editor/src/components/custom-code/code-file.ts
balazsbajorics: ah I actually wanted to ask this!!

balazsbajorics: thanks, I fixed it

PR #2988, opened 2022-12-13
fix(reparent) static reparent keeps the original in place hidden

enidemi:
**Problem:**
When reparenting a static element the parent size might change shifting possible reparent targets under the cursor, makes it difficult to target these elements.

**Fix:**
Reparent strategies create a duplicate element in place of the original one and they appear as hidden elements on the canvas. Only if the original elements was not absolute.

**Commit Details:**
- add duplicate element command for non-absolute reparent targets
- these elements are added to `hiddenInstances` and they are hidden in the navigator too
- fix reparent strategies to correctly update the duplicate element uid
- test to check if flex parent is not collapsed during reparent

Before https://utopia.pizza/p/97fce0d6-glory-xylocarp
After https://utopia.pizza/p/97fce0d6-glory-xylocarp/?branch_name=fix-reparent-flex-layout-collapse


editor/src/components/canvas/canvas-strategies/strategies/reparent-as-static-strategy.tsx
```diff
@@ -186,24 +190,28 @@ function applyStaticReparent(
 
             const commandsAfterReorder = [
               ...propertyChangeCommands,
-              setElementsToRerenderCommand([target, newPath]),
+              setElementsToRerenderCommand([target, newPath]), // TODO THIS LIST IS INCOMPLETE, KEEPS GHOSTS IN THE NAVIGATOR
```
Rheeseyb, line 193: I believe this is a bug with all of our existing re-parent strategies

editor/src/components/canvas/canvas-strategies/strategies/reparent-utils.ts
```diff
@@ -162,3 +174,44 @@ export function cursorForMissingReparentedItems(
...
+      duplicatedElementNewUids[selectedElementString] = newUid
+
+      const newPath = EP.appendToPath(EP.parentPath(elementPath), newUid)
+
+      return [
+        duplicateElement('mid-interaction', elementPath, newUid),
+        wildcardPatch('mid-interaction', {
+          hiddenInstances: { $push: [newPath] },
+        }),
+        hideInNavigatorCommand([newPath]),
```
Rheeseyb, line 210: So I guess we should use something like this for other re-parenting strategies (but using the old path in those)?

Rheeseyb (approved):

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-utils.ts
```diff
@@ -162,3 +174,44 @@ export function cursorForMissingReparentedItems(
...
+  let duplicatedElementNewUids: { [elementPath: string]: string } = {
+    ...customStrategyState.duplicatedElementNewUids,
+  }
+  const commands = filteredSelectedElements.flatMap((elementPath) => {
+    const element = MetadataUtils.findElementByElementPath(
+      canvasState.startingMetadata,
+      elementPath,
+    )
+    // we want to keep a placeholder element where the original dragged element was to avoid the new parent shifting around on the screen
+    const hasCommonAncestor =
```
balazsbajorics, line 193: I think it's worth adding a comment here that THIS is the entry point if we want a more finegrained heuristic for happy paths where we don't need the layout-keeping

balazsbajorics (approved):
nice!

PR #2994, opened 2022-12-13
Remove ElementOriginType

balazsbajorics:
**Problem:**
ElementOriginType was a type that we used in some places to determine if something is a generated element or not. It was pretty old and outdated: generated-ness can be determined just by looking at the ElementPath. This means that MetadataUtils.getElementOriginType and all the functions calling it were unnecessarily spinnning `withUnderlyingTarget`, and had to use selectors relying on projectContents _which necessarily changes every single interaction frame_.

**Fix:**
Introduce much more simplistic `MetadataUtils.isElementGenerated`, since that was the only thing we cared about anyways. Kill all uses of ElementOriginType :knife:


seanparsons (approved):

Rheeseyb (approved):
So much simpler!

enidemi (approved):

PR #2995, opened 2022-12-13
Carry over metadata changes when an interaction finishes.

seanparsons:
**Problem:**
When changing selection, values like the `computedStyle` for the newly selected element appear to be lost even though they have been captured.

**Fix:**
An interaction session was being started and cleared in the middle of processing the selection change, which resulted in the update to the metadata getting assigned into the patched editor state and then lost almost immediately by the interaction being finished.

This fix carries over the metadata from the patched editor state to the new editor when the interaction finishes and only specifically when an interaction is not viewed as being "in progress".

**Commit Details:**
- In `interactionFinish`, when closing off the interaction that isn't in progress, carry over the metadata from the patched editor.

### DIFF
--- a/editor/src/components/editor/store/dispatch-strategies.tsx
+++ b/editor/src/components/editor/store/dispatch-strategies.tsx
@@ -54,7 +54,7 @@ export function interactionFinished(
   storedState: EditorStoreFull,
   result: EditorStoreUnpatched,
 ): HandleStrategiesResult {
-  const newEditorState = result.unpatchedEditor
+  let newEditorState = result.unpatchedEditor
   const withClearedSession = createEmptyStrategyState(
     newEditorState.canvas.interactionSession?.latestMetadata ?? newEditorState.jsxMetadata,
     newEditorState.canvas.interactionSession?.latestAllElementProps ??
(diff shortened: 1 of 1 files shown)


editor/src/components/editor/store/dispatch-strategies.tsx
```diff
@@ -109,6 +109,14 @@ export function interactionFinished(
       newStrategyState: withClearedSession,
     }
   } else {
+    // Try to keep any updated metadata that may have been populated into here
+    // in the meantime.
+    newEditorState = {
```
balazsbajorics, line 114: sorry for the nitpick, but would you mind declaring a new const here? the `let newEditorState` in an almost 100 line long function takes a big toll on me when reading / refactoring

balazsbajorics (approved):

maltenuhn:
> When changing selection, values like the computedStyle for the newly selected element appear to be lost even though they have been captured.

If there is even a remote chance of this accidentally regressing again, this is _such_ a core editor behaviour that it probably wants some form of test - if that's possible?

Rheeseyb:
> > When changing selection, values like the computedStyle for the newly selected element appear to be lost even though they have been captured.
> 
> If there is even a remote chance of this accidentally regressing again, this is _such_ a core editor behaviour that it probably wants some form of test - if that's possible?

Yep, definitely agree, though that should probably come in a follow up PR so we can get this one straight out

PR #2998, opened 2022-12-14
Escape text in text editing mode

### LINKED ISSUE #2997: Escape HTML entities in text editing mode

ruggi:
Fixes #2997

**Problem:**
Typing HTML entities during text editing (e.g. an ampersand or angle brackets) breaks the components.

**Fix:**
Escape the HTML entities when dispatching edits, and unescape them in the rendered components to keep their original visual aspect.

escape


bkrmendy (approved):

editor/package.json
```diff
@@ -132,6 +132,7 @@
     "fontfaceobserver": "2.1.0",
     "friendly-words": "1.1.10",
     "graphlib": "2.1.1",
+    "he": "^1.2.0",
```
Rheeseyb, line 135: Please can you remove the `^` from the dependencies that have been added here. It doesn't look like this library is likely to be updated in the near future anyway, but we've had quite a few problems when building with dependencies that weren't pinned to a specific version

Rheeseyb (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -319,11 +319,12 @@ export function renderCoreElement(
       return <>{renderedChildren}</>
     }
     case 'JSX_TEXT_BLOCK': {
+      const unescaped = unescapeHTML(element.text)
       const parentPath = Utils.optionalMap(EP.parentPath, elementPath)
       if (parentPath == null || !EP.pathsEqual(parentPath, editedText)) {
-        return element.text
+        return unescaped
```
balazsbajorics, line 325: this line rings an alarm bell to me. how would react render this text outside of utopia? we shouldn't be applying any special escaping that doesn't happen in create-react-app/codesandbox/etc

gbalint (approved):

enidemi (approved):

balazsbajorics (changes requested):
let's talk about escaping the text when not in text-editing mode!

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
ruggi: That's a good point to which I don't have an obvious response 🤔 Should we maybe make a custom component that does the same unescaping?

ruggi: actually, it seems to be fine:

Screenshot 2022-12-14 at 2 02 40 PM

Rheeseyb: Was about to say, we should leave the text escaped 👍

balazsbajorics: I believe this line should just be the original `return element.text` and that _should_ just work fine on the canvas. Right?

ruggi: unfortunately not 😬 

Screenshot 2022-12-14 at 2 13 15 PM

PR #3004, opened 2022-12-14
Fixed up the distance guideline styling

### LINKED ISSUE #3000: Distance guidelines are hard to read
The distant guides are hard to read (or even see) in a lot of cases, especially when the distance is small. We should use the same format (white on pink) as the dimension and gap indicators

Rheeseyb:
Fixes #3000 

**Problem:**
The distance guidelines were often very hard to read or even see

**Fix:**
Use the same colours and styling as the floating number controls used for padding and gap controls. I've also moved these controls after the others so that they always appear on top.

**Before:**
image

image

image

**After:**
image

image

image


balazsbajorics (approved):

enidemi (approved):

maltenuhn:
@lankaukk any tweaks to suggest here?

PR #3011, opened 2022-12-15
`keyboardSetOpacityStrategy`

bkrmendy:
## Set opacity via the keyboard
This PR makes it possible to set the opacity elements by typing a number on the keyboard. This can be done by selecting the element, and typing the desired opacity value (such as `3-4`), which will set the `opacity` prop of the element to that value (`34%` in the example).

If only a single digit is typed (such as `7`), the opacity value will be that number * 10 (so `70%` if 7 is typed). If a single `0` or `100` is typed, opacity is set to `100%`.

The typed number is reset after 600ms, which means that if there's a brief pause between entering digits, the strategy will regard them as separate interaction (so in the case of `4 --- pause --- 7`, `opacity` will be set to `70%`, with 4 discarded).


gbalint (approved):

editor/src/components/canvas/canvas-strategies/strategies/keyboard-set-opacity-strategy.spec.browser2.tsx
```diff
@@ -0,0 +1,142 @@
+import { CanvasControlsContainerID } from '../../controls/new-canvas-controls'
+import { mouseClickAtPoint, pressKey } from '../../event-helpers.test-utils'
+import { EditorRenderResult, renderTestEditorWithCode } from '../../ui-jsx.test-utils'
+
+describe('adjust font size with the keyboard', () => {
+  describe('no font size specified', () => {
```
Rheeseyb, line 6: "font size" 👀

bkrmendy: 😬

bkrmendy: fixed

Rheeseyb (approved):

PR #3013, opened 2022-12-15
Only show re-order slider on hover

### LINKED ISSUE #3012: Only show re-order slider control on hover
We already only show the re-order slider control for suitable elements, but it still adds noise to the canvas, so we should only ever show it when either the element is hovered or the control is already being dragged

Rheeseyb:
Fixes #3012 

**Problem:**
We want to hide the re-order slider unless the element is hovered or the control is already being dragged

**Fix:**
Add a check to see if the element is either hovered, or the control is being dragged, plus some tests to capture that.

### DIFF
--- a/editor/src/components/canvas/controls/reorder-slider-control.tsx
+++ b/editor/src/components/canvas/controls/reorder-slider-control.tsx
@@ -46,10 +46,10 @@ export const ReorderSliderControl = controlForStrategyMemoized(
         store.editor.canvas.interactionSession.activeControl.type === 'REORDER_SLIDER',
       'ReorderSliderControl isDragging',
     )
-    const isSelectedElementHovered = useEditorState((store) => {
-      const { hoveredViews, selectedViews } = store.editor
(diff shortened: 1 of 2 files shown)


bkrmendy (approved):

editor/src/components/canvas/controls/reorder-slider-control.tsx
```diff
@@ -46,6 +46,10 @@ export const ReorderSliderControl = controlForStrategyMemoized(
         store.editor.canvas.interactionSession.activeControl.type === 'REORDER_SLIDER',
       'ReorderSliderControl isDragging',
     )
+    const isSelectedElementHovered = useEditorState((store) => {
+      const { hoveredViews, selectedViews } = store.editor
+      return hoveredViews.includes(selectedViews[0])
```
seanparsons, line 51: Should this check all of the possible values of `selectedViews`? I don't _think_ it should, but I wonder if that's more coincidental to how the related strategies are implemented.

PR #3017, opened 2022-12-15
Support pasting content during text editing

### LINKED ISSUE #3016: Support pasting content into the text editor

ruggi:
Fixes #3016 

**Problem:**
Currently pasting content is not possible in text edit mode.

**Fix:**
Handle clipboard events and insert the content in the right place inside the text.

Notes:
- it only works with keyboard-based paste, as right clicking with the mouse will exit the editing mode
- the caret gymnastics are required to make sure that the caret stays effectively where it should be (after the pasted content) instead of jumping around
- undo is a bit borked, but this should probably be fixed in a different PR

Kapture 2022-12-15 at 15 46 45

### DIFF
--- a/editor/src/components/text-editor/text-editor.tsx
+++ b/editor/src/components/text-editor/text-editor.tsx
@@ -62,8 +62,6 @@ export const TextEditor = React.memo(({ elementPath, text }: TextEditorProps) =>
       return
     }
 
-    setSelectionToEnd(currentElement)
-
     currentElement.focus()
 
     return () => {
@@ -79,6 +77,7 @@ export const TextEditor = React.memo(({ elementPath, text }: TextEditorProps) =>
       return
     }
     myElement.current.textContent = firstTextProp
+    setSelectionToEnd(myElement.current)
   }, [firstTextProp])
 
   const onKeyDown = React.useCallback(
--- a/editor/src/components/text-editor/text-editor.tsx
+++ b/editor/src/components/text-editor/text-editor.tsx
@@ -63,6 +63,7 @@ export const TextEditor = React.memo(({ elementPath, text }: TextEditorProps) =>
     }
 
     setSelectionToEnd(currentElement)
+
     currentElement.focus()
 
     return () => {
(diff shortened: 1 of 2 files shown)


editor/src/components/text-editor/text-editor.tsx
```diff
@@ -127,6 +158,25 @@ export const TextEditor: React.FC<TextEditorProps> = ({ elementPath, text }: Tex
...
+      const before = myElement.current.textContent.slice(0, caretPosition)
+      const after = myElement.current.textContent.slice(caretPosition)
+      myElement.current.textContent = before + pastedText + after
+
+      const newCaretPosition = caretPosition + pastedText.length
+      moveCaret(myElement.current, newCaretPosition)
+      setCaretPosition(newCaretPosition)
+    },
+    [caretPosition],
+  )
```
gbalint, line 178: I'm not sure about this, but isn't that easier to force the browser to actually insert the pasted text with `document.execCommand('insertText', pastedText)` , just like in the tests?
And then we would need this whole complexity and manual cursor management.
BTW do we know why paste doesn't work out of the box? It should, shouldn't it?

ruggi: > isn't that easier to force the browser to actually insert the pasted text

sure, but this felt simple enough to justify not using a deprecated API

> BTW do we know why paste doesn't work out of the box?
no idea ¯\_(ツ)_/¯

gbalint: To be honest I would try to investigate first why it doesn't work (maybe we have a `preventDefault` somewhere?), because this solution feels more fragile than the native paste handling of the browser.

ruggi: incredibly enough, it all boiled down to just doing a `stopPropagation()` on the onPaste event o.O

https://github.com/concrete-utopia/utopia/pull/3017/commits/7e9c54c43aa66d118e590554c3dddbecfe991420

I'm now working on the tests which are not working anymore, but the functionality is there :)

maltenuhn: > BTW do we know why paste doesn't work out of the box?

I'm pretty sure it's because Utopia grabs top-level paste events so you can eg paste something onto the canvas even if eg the navigator is focused.

PR #3020, opened 2022-12-15
Double click to edit text

gbalint:
Fixes #[ticket_number]

**Problem:**

**Fix:**

**Commit Details:**
- Did [other things]

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
@@ -23,6 +23,7 @@ import {
   selectComponents,
   setHoveredView,
   clearHoveredViews,
+  switchEditorMode,
 } from '../../../editor/actions/action-creators'
 import { cancelInsertModeActions } from '../../../editor/actions/meta-actions'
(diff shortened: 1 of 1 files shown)


ruggi (approved):

editor/src/components/canvas/controls/text-edit-mode/text-edit-mode.spec.browser2.tsx
```diff
@@ -64,14 +82,22 @@ describe('Click to choose target text for editing', () => {
     const editor = await renderTestEditorWithCode(projectWithNestedDiv, 'await-first-dom-report')
 
     pressKey('t')
-    await selectElement(editor, 'div')
+    await clickOnElement(editor, 'div')
```
maltenuhn, line 85: does this imply that eg `<button>Hello World</button>`  wouldn't be huntable / editable with `t` ?

Rheeseyb: That shouldn't make a difference since this is in select mode and not live mode, so the button itself wouldn't be handling the events

Rheeseyb (approved):
I agree it's hacky, but I'm not sure there's a better solution without a non-trivial amount of refactoring

PR #3021, opened 2022-12-15
Selection hot path improvements.

seanparsons:
**Problem:**
Selection is still quite expensive because of the numerous things that happen during selection.

**Fix:**
This is a collection of small fixes which improve the performance of changing selection.

**Results:**
The average time of 4 selection related mouse down events was dropped from 100.22ms to 91.12ms as measured by the Chrome profiler.

**Commit Details:**
- Memoized some of the lower level CSS parsing functions which are called from the inspector.
- `NavigatorItemDndWrapper` now doesn't unnecessarily spread `props.windowStyle`.
- Used `useKeepReferenceEqualityIfPossible` for the `windowStyle` property of `Item`, as it was changing unnecessarily.
- Use `traverseEither` instead of mapping an array and then invoking `sequenceEither`.
- Replaced `urljoin` with `appendToPath` as the former utilises a bunch of regexes and as a result is very slow.

### DIFF
--- a/editor/src/components/inspector/common/css-utils.ts
+++ b/editor/src/components/inspector/common/css-utils.ts
@@ -4836,10 +4836,12 @@ function parseValueFactory<T, K extends keyof T>(parserMap: {
   }
 }
 
-const parseMetadataValue = parseValueFactory(elementPropertiesParsers)
-const parseCSSValue = parseValueFactory(cssParsers)
-const parseOldLayoutValue = parseValueFactory(layoutParsers)
-const parseNewLayoutValue = parseValueFactory(layoutParsersNew)
+const parseMetadataValue = Utils.memoize(parseValueFactory(elementPropertiesParsers), {
+  maxSize: 1000,
+})
+const parseCSSValue = Utils.memoize(parseValueFactory(cssParsers), { maxSize: 1000 })
+const parseOldLayoutValue = Utils.memoize(parseValueFactory(layoutParsers), { maxSize: 1000 })
+const parseNewLayoutValue = Utils.memoize(parseValueFactory(layoutParsersNew), { maxSize: 1000 })
 
 function isMetadataProp(prop: unknown): prop is keyof ParsedElementProperties {
   return typeof prop === 'string' && prop in elementPropertiesEmptyValues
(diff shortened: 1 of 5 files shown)


editor/src/components/inspector/common/css-utils.ts
```diff
@@ -4836,10 +4836,12 @@ function parseValueFactory<T, K extends keyof T>(parserMap: {
   }
 }
 
-const parseMetadataValue = parseValueFactory(elementPropertiesParsers)
-const parseCSSValue = parseValueFactory(cssParsers)
-const parseOldLayoutValue = parseValueFactory(layoutParsers)
-const parseNewLayoutValue = parseValueFactory(layoutParsersNew)
+const parseMetadataValue = Utils.memoize(parseValueFactory(elementPropertiesParsers), {
+  maxSize: 1000,
```
balazsbajorics, line 4840: how close does the before-i-go project comes to hitting this maxSize per frame?

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -250,9 +250,7 @@ export class NavigatorItemDndWrapper extends PureComponent<
         key='navigatorItem'
         id={`navigator-item-${safeComponentId}`}
         data-testid={`navigator-item-${safeComponentId}`}
-        style={{
-          ...props.windowStyle,
-        }}
+        style={props.windowStyle}
```
balazsbajorics, line 253: 🤦

balazsbajorics (approved):
that maxSize: 1000 looks a bit scary to me, otherwise 👍

editor/src/components/inspector/common/css-utils.ts
seanparsons: I'll have a look and see, I set it high because the values involved should be quite small.

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
Rheeseyb: omg

editor/src/utils/hashed-assets.ts
```diff
@@ -35,7 +35,7 @@ function getPossiblyHashedURLInner(url: string): string {
 
 export function getPossiblyHashedURL(url: string): string {
   const relativeURL = getPossiblyHashedURLInner(url)
-  return urljoin(STATIC_BASE_URL, relativeURL)
+  return appendToPath(STATIC_BASE_URL, relativeURL)
```
Rheeseyb, line 38: Hmmm this is definitely less safe, so we'll have to make sure we're extra vigilant with uses of this function

Rheeseyb (approved):

PR #3031, opened 2022-12-21
draw to insert text strategy (wip)

ruggi:
Fixes #[ticket_number]

**Problem:**

**Fix:**

**Commit Details:**
- Did [other things]

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -1961,6 +1961,9 @@ export const UPDATE_FNS = {
     derived: DerivedState,
   ): EditorModel => {
     // same as UPDATE_EDITOR_MODE, but clears the drag state
+    if (action.unlessMode === editor.mode.type) {
+      return clearDragState(editor, derived, false)
+    }
     return clearDragState(setModeState(action.mode, editor), derived, false)
   },
   TOGGLE_CANVAS_IS_LIVE: (editor: EditorModel, derived: DerivedState): EditorModel => {
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
@@ -50,6 +50,7 @@ import * as EP from '../../../core/shared/element-path'
 import { keyboardSetFontSizeStrategy } from './strategies/keyboard-set-font-size-strategy'
 import { keyboardSetFontWeightStrategy } from './strategies/keyboard-set-font-weight-strategy'
 import { keyboardSetOpacityStrategy } from './strategies/keyboard-set-opacity-strategy'
+import { drawToInsertTextStrategy } from './strategies/draw-to-insert-text-strategy'
 
 export type CanvasStrategyFactory = (
   canvasState: InteractionCanvasState,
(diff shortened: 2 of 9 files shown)


gbalint (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1961,6 +1961,9 @@ export const UPDATE_FNS = {
     derived: DerivedState,
   ): EditorModel => {
     // same as UPDATE_EDITOR_MODE, but clears the drag state
+    if (action.unlessMode === editor.mode.type) {
```
bkrmendy, line 1964: I think instead of the `unlessMode` flag, `clearDragState` should be turned into a full-fledged action, and whenever `action.unlessMode` would be the same as `editor.mode.type`, the new `clearDragState` action should be called.

The reason for this would be that with `unlessMode`, the `SWITCH_EDITOR_MODE` action could be misused in a way that wouldn't actually switch the editor mode. While this is not a huge deal, solving this via composition, rather than adding a flag to the existing action would lead to a cleaner implementation.

bkrmendy (approved):

editor/src/components/editor/actions/actions.tsx
bkrmendy: turns out this isn't feasible in this PR (see https://github.com/concrete-utopia/utopia/pull/3031/commits/dbf885e89856e0cd7270d28c9ed8ea2a9e4c4160)

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-helpers.ts
```diff
@@ -54,6 +54,7 @@ export function findReparentStrategies(
   cmdPressed: boolean,
   pointOnCanvas: CanvasPoint,
   allowSmallerParent: AllowSmallerParent,
+  allowWithOnlyTextChildren?: boolean,
```
seanparsons, line 57: I'd be tempted to make this mandatory, to avoid `undefined` shenanigans.
```suggestion
  allowWithOnlyTextChildren: boolean,
```

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-helpers.ts
bkrmendy: good idea, but in this specific case I'd approach it from a different direction, looking at where this flag is used, I wonder if there could be a `MetadataUtils.targetElementSupportsChildrenAlsoText` that is same as `MetadataUtils.targetElementSupportsChildren` with `allowWithOnlyTextChildren: true`, but instead of boolean, returns `"supportsChildren' | 'hasOnlyTextChildren' | 'doesNotSupportChildren'`. 
`MetadataUtils.targetElementSupportsChildren` would call this function internally, and return true for supportsChildren, and false otherwise.

the upside of this would be that instead of threading an extra parameter through a bunch of functions, there would be a more general metadata utils function (the hypothetical MetadataUtils.targetElementSupportsChildrenAlsoText), which could be used by the text editing code, and one that could be easily reused by the existing `MetadataUtils.targetElementSupportsChildren`

bkrmendy: looks amazing!

PR #3034, opened 2023-01-02
`Fill container` improvements

bkrmendy:
This PR improves the fill/fixed/hug control by setting `flex-grow: 1` instead of `width: 100%`, when a flex child is set to `Fill container`. In the same vein, when determining the state of the fixed/fill/hug control, `flex-grow` is classified as `Fill container`, and the value of `flex-grow` is displayed in the number input (not editable yet).

### DIFF
--- a/editor/src/components/inspector/inspector-strategies.tsx
+++ b/editor/src/components/inspector/inspector-strategies.tsx
@@ -67,9 +67,9 @@ export const updateFlexDirectionStrategies = (
       return null
     }
 
-    return elements.flatMap((path) => [
+    return elements.map((path) =>
       setProperty('always', path, PP.create(['style', 'flexDirection']), flexDirection),
-    ])
+    )
   },
 ]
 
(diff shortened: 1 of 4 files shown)


editor/src/components/inspector/inspector-strategies.tsx
```diff
@@ -62,9 +67,9 @@ export const updateFlexDirectionStrategies = (
       return null
     }
 
-    return elements.map((path) =>
+    return elements.flatMap((path) => [
```
seanparsons, line 70: Isn't this unnecessary?

seanparsons (approved):

editor/src/components/inspector/inspector-strategies.tsx
bkrmendy: thanks, fixed! (it got stuck there from an earlier iteration)

ruggi (approved):

PR #3046, opened 2023-01-04
when removing flex layout, transform children to keep their position visually

bkrmendy:
remove_flex_layout

## Description
When flex layout is removed from a container
- all flex-related properties are pruned from the children and the root container
- all children are converted to `position: absolute`,
- for all children, `width`/`height` is set to their measured dimensions (based on `specialSizeMeasurements`)
- the parent container is sized to keep its visual size


editor/src/components/inspector/add-remove-layout-system-control.tsx
```diff
@@ -16,9 +16,9 @@ import { metadataSelector, selectedViewsSelector } from './inpector-selectors'
 import {
   addFlexLayoutStrategies,
   removeFlexLayoutStrategies,
-  runStrategies,
-} from './inspector-strategies'
+} from './inspector-strategies/inspector-strategies'
 import { detectAreElementsFlexContainers } from './inspector-common'
+import { runStrategies } from './inspector-strategies/inspector-strategy'
```
Rheeseyb, line 21: Unexpected auto refactor

Rheeseyb: It seems this has happened in a lot of these files

Rheeseyb: Oh sorry, my bad, I just noticed the slight filename difference!

editor/src/components/inspector/inspector-strategies/remove-flex-convert-to-absolute-strategy.tsx
```diff
@@ -0,0 +1,73 @@
...
+function removeFlexConvertToAbsoluteOne(
+  metadata: ElementInstanceMetadataMap,
+  elementPath: ElementPath,
+): Array<CanvasCommand> {
+  const children = MetadataUtils.getChildrenPaths(metadata, elementPath)
+  return [
+    ...pruneFlexPropsCommands(elementPath), // flex-related stuff is pruned
+    ...children.flatMap((c) => positionAbsoluteCommands(c)), // all children are converted to absolute,
+    ...children.flatMap((c) => sizeToVisualDimensions(metadata, c)), // with width/height based on measured dimensions
+    ...children.flatMap((c) => pruneFlexPropsCommands(c)),
```
Rheeseyb, line 64: I believe this means that if a child is also a flex container then we will be removing that prop, meaning it would break the layout of the grandchildren in that case

Rheeseyb: In fact, it would remove other props on the children that exist purely for the layout of their respective children

editor/src/components/inspector/inspector-strategies/inspector-strategy.tsx
```diff
@@ -0,0 +1,25 @@
...
+  metadata: ElementInstanceMetadataMap,
+  selectedViews: ElementPath[],
+  strategies: InspectorStrategy[],
+): void {
+  for (const strategy of strategies) {
+    const commands = strategy(metadata, selectedViews)
+    if (commands != null) {
+      dispatch([applyCommandsAction(commands)])
+    }
+    return
```
seanparsons, line 23: Doesn't this `return` statement cause it to exit after the first `strategy` call?

seanparsons (approved):

editor/src/components/inspector/inspector-strategies/inspector-strategy.tsx
bkrmendy: this is fully intentional, the strategies should be in order of specificity, with the most specific matching one applied. As far as I understand this is the same principle that the canvas strategies use (with the difference that inspector strategies don't have a strategy picker (yet))

Rheeseyb (approved):

PR #3052, opened 2023-01-05
Feature/text editing insert autosize text

enidemi:
**Problem:**
Inserting a new span with clicking adds a fix width and height of 100px. It should start from an unset size to allow it automatically grow when typing.

**Fix:**
Extended insertion subject to not always use a default or given size, but it's possible to skip all size props and only set top and left. Draw to insert still adds width and height.
When inserting a span without size the cursor is not visible, so in this case a 1px minwidth is set on the text editor wrapper.

**Commit Details:**
- updated insertion strategies
- both `t` shortcut and canvas toolbar button inserts a span without a size
- TextEditorWrapper adds minwidth/unsets maxwidth if size is 0, to always show a cursor

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-metastrategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-metastrategy.tsx
@@ -12,6 +12,7 @@ import { elementPath } from '../../../../core/shared/element-path'
 import {
   ElementInstanceMetadataMap,
   emptyComments,
+  JSXAttributes,
   jsxAttributeValue,
 } from '../../../../core/shared/element-template'
 import {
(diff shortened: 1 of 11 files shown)


editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-metastrategy.tsx
```diff
@@ -295,47 +302,74 @@ function getHighlightAndReorderIndicatorCommands(
...
+  insertionSubjectSize: Size | 'skip-size-props',
   sizing: 'zero-size' | 'default-size',
 ): { command: InsertElementInsertionSubject; frame: CanvasRectangle } | null {
   if (
     interactionSession.interactionData.type === 'DRAG' &&
     (sizing === 'default-size' || interactionSession.interactionData.drag != null)
   ) {
     const pointOnCanvas = interactionSession.interactionData.dragStart
+    if (insertionSubjectSize === 'skip-size-props') {
+      const frame = canvasRectangle({
```
Rheeseyb, line 314: It looks like this side of the condition is running the exact same code as if `sizing === 'zero-size'` below

ruggi (approved):

Rheeseyb (approved):

seanparsons (approved):

PR #3059, opened 2023-01-06
Double click on resize edge to set container to `hug contents`

bkrmendy:
dbl_click_hug

## Description
When double clicking on the resize edge of a container, it will be size to `Hug contents`, if applicable.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
@@ -18,13 +18,16 @@ import {
   WindowPoint,
   zeroCanvasPoint,
 } from '../../../../core/shared/math-utils'
-import { cmdModifier, emptyModifiers, Modifiers, shiftModifier } from '../../../../utils/modifiers'
+import { emptyModifiers, Modifiers, shiftModifier } from '../../../../utils/modifiers'
 import { ElementPath } from '../../../../core/shared/project-file-types'
 import {
   EdgePositionBottomRight,
   EdgePosition,
   EdgePositionLeft,
   EdgePositionTopLeft,
+  EdgePositionRight,
+  EdgePositionBottom,
+  EdgePositionTop,
 } from '../../canvas-types'
 import { wait } from '../../../../utils/utils.test-utils'
 import { ControlDelay } from '../canvas-strategy-types'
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/controls/select-mode/absolute-resize-control.tsx
```diff
@@ -234,13 +238,30 @@ const ResizeEdge = React.memo(
       [maybeClearHighlightsOnHoverEnd],
     )
 
+    const onEdgeDblClick = React.useCallback(
+      (event: React.MouseEvent<HTMLDivElement>) => {
+        if (event.detail != 2) {
```
ruggi, line 243: as discussed on discord, this can lead to weird behavior if there are children with position `absolute`, in which case it would make sense to just do nothing here

editor/src/components/canvas/controls/select-mode/absolute-resize-control.tsx
```diff
@@ -234,13 +238,30 @@ const ResizeEdge = React.memo(
...
+    )
+
     const lineSize = ResizeMouseAreaSize / scale
     const width = props.direction === 'horizontal' ? undefined : lineSize
     const height = props.direction === 'vertical' ? undefined : lineSize
     const offsetLeft = props.direction === 'horizontal' ? `0px` : `${-lineSize / 2}px`
     const offsetTop = props.direction === 'vertical' ? `0px` : `${-lineSize / 2}px`
     return (
       <div
+        onClick={onEdgeDblClick}
```
gbalint, line 264: Can't you just use `onDoubleClick`? And then there is no need to check the event.detail

enidemi (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
```diff
@@ -268,6 +273,69 @@ export var ${BakedInStoryboardVariableName} = (props) => {
...
+  const divCorner = {
+    x: divBounds.x + 50,
+    y: divBounds.y + 40,
+  }
+
+  mouseClickAtPoint(canvasControlsLayer, divCorner)
+
+  const nineBlockControlSegment = editor.renderedDOM.getByTestId(testId)
+
+  mouseClickAtPoint(nineBlockControlSegment, { x: 2, y: 30 }, { eventOptions: { detail: 2 } })
```
Rheeseyb, line 289: We have a `mouseDoubleClickAtPoint` which would be better to use here as it would fire all of the events

PR #3061, opened 2023-01-06
position caret with double click

### LINKED ISSUE #3060: Position caret under mouse when double-clicking a text editable

ruggi:
Fixes #3060 

**Problem:**

Currently double-clicking or click-to-insert-click a text-editable element positions the caret at the very end of the text contents. It would be nice to be able to position the caret under the mouse immediately.

**Fix:**

1. When double clicking, position the caret under the mouse (**Including multi-line**)
2. When click-to-insert, position the caret under the mouse (**including multi-line**)

click caret

The positioning is done with a best-effort logic, so it may happen that the cursor is placed before or after the exact character "kinda-under" the mouse depending on the window positioning (the boundaries are calculated relative to _the caret itself_).

The position is found with a linear search instead of a BS in order to keep the code readable and not overcomplicate things too much since we're going to be dealing with really small arrays anyways.

### DIFF
--- a/editor/src/components/text-editor/text-editor.tsx
+++ b/editor/src/components/text-editor/text-editor.tsx
@@ -174,8 +174,8 @@ async function setSelectionToOffset(
   scale: number,
   cursorPosition: Coordinates | null,
 ) {
-  const sel = window.getSelection()
-  if (sel == null) {
+  const selection = window.getSelection()
+  if (selection == null) {
     return
   }
   if (element.childNodes.length != 1) {
@@ -187,14 +187,14 @@ async function setSelectionToOffset(
   }
 
   const setRange = (start: number | null) => {
-    sel.removeAllRanges()
+    selection.removeAllRanges()
     const range = document.createRange()
     range.selectNodeContents(textNode)
     range.collapse(start != null)
     if (start != null) {
       range.setStart(textNode, start)
     }
-    sel.addRange(range)
+    selection.addRange(range)
     return range
   }
 
--- a/editor/src/components/text-editor/text-editor.tsx
+++ b/editor/src/components/text-editor/text-editor.tsx
@@ -211,17 +211,13 @@ async function setSelectionToOffset(
     // linear search is a tad slow, but it should be fine
     // and it's a lot easier when dealing with the scaling of the editor
     const targetX = cursorPosition.x / scale
+    const targetY = cursorPosition.y / scale
     for (let i = 0; i <= maxLength; i++) {
       const range = setRange(i)
       const rect = range.getBoundingClientRect()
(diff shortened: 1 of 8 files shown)


editor/src/components/text-editor/text-editor.tsx
```diff
@@ -158,15 +169,69 @@ export const TextEditorWrapper = React.memo((props: TextEditorProps) => {
...
+    if (start != null) {
+      range.setStart(textNode, start)
+    }
+    sel.addRange(range)
+    return range
+  }
+
+  const maxLength = setRange(null).endOffset
+
+  if (cursorPosition != null) {
```
bkrmendy, line 203: nit, but this check could be moved out of the function

ruggi: mmh not really because the same function can be used to just set the caret to the end of the sentence as a default (e.g. you press `t` on a selected editable)

editor/src/components/text-editor/text-editor.tsx
```diff
@@ -158,15 +169,69 @@ export const TextEditorWrapper = React.memo((props: TextEditorProps) => {
...
-function setSelectionToEnd(element: HTMLSpanElement) {
-  const range = document.createRange()
-  range.selectNodeContents(element)
-  range.collapse(false)
+async function setSelectionToOffset(
+  element: HTMLSpanElement,
+  scale: number,
+  cursorPosition: Coordinates | null,
+) {
+  const sel = window.getSelection()
```
Rheeseyb, line 177: Minor, but why us `sel` and not `selection`?

editor/src/components/text-editor/text-editor.tsx
```diff
@@ -158,15 +169,69 @@ export const TextEditorWrapper = React.memo((props: TextEditorProps) => {
...
+      }
+    }
+    const targetY = cursorPosition.y / scale
+    for (let i = 0; i <= maxLength; i++) {
+      const range = setRange(i)
+      const rect = range.getBoundingClientRect()
+      if (rect.y <= targetY && targetY <= rect.y + rect.height) {
+        validY.push(i)
+      }
+    }
```
Rheeseyb, line 228: Why not combine these rather than iterating over the full range twice?

editor/src/components/text-editor/text-editor.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/3061/commits/d0ea2b8d246bce78bacc55825e24fc5f4340632d

Rheeseyb (approved):

bkrmendy (approved):

PR #3064, opened 2023-01-06
Canvas overlay for text editing

gbalint:
**Problem:**
We don't want the canvas to be live while text editing, but we can not have the canvas control layer on top because
1) that would cover the text too, so it would be impossible to use the text editor with the mouse, e.g. setting selection
2) we don't wanna show the canvas controls.

**Fix:**
I made a new component called `TextEditCanvasOverlay` which covers the canvas all the time, except when you are over the text editor itself. I takes into account overflowing text
I also removed the mouse event handlers from the actual text component

**Issues:**
This is a best effort PR, interactivity can still leak in, e.g. if there is an interactive component under the edited element (the overlay can not cover that)


enidemi (approved):

ruggi (approved):
🏅

PR #3068, opened 2023-01-09
Delete new text editable elements when left empty

### LINKED ISSUE #3067: Delete new text editable elements when left empty

ruggi:
Fixes #3067 

**Problem:**

When a new text element is created but its contents are left empty, the new element should be removed from the project.

**Fix:**

When a _new_ element is left empty after exiting text edit mode (`Escape`, click away, etc.), it is removed from the tree.
Non-new elements that are edited and left empty, instead, are not removed.

Kapture 2023-01-09 at 13 36 57

### DIFF
--- a/editor/src/components/text-editor/text-editor.spec.browser2.tsx
+++ b/editor/src/components/text-editor/text-editor.spec.browser2.tsx
@@ -5,6 +5,7 @@ import { CanvasControlsContainerID } from '../canvas/controls/new-canvas-control
 import {
   mouseClickAtPoint,
   mouseDoubleClickAtPoint,
+  mouseDragFromPointToPoint,
   pressKey,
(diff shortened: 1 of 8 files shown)


editor/src/components/text-editor/text-editor.spec.browser2.tsx
```diff
@@ -417,3 +496,13 @@ export var storyboard = (
   </Storyboard>
 )
 `)
+
+const emptyProject = formatTestProjectCode(`import * as React from 'react'
+import { Storyboard } from 'utopia-api'
+
+
+export var storyboard = (
+  <Storyboard data-uid='sb'>
```
ruggi, line 505: note: this is intentionally left "ugly" to make sure the editor actually does its thing and cleans up the tree

bkrmendy (approved):

PR #3070, opened 2023-01-09
fix(canvas) br tags are not shown in the navigator

enidemi:
**Problem:**
Inside text elements there can be `<br>` tags, some of these are added automatically for new lines. The problem is that they show canvas controls when the text is selected and they should be excluded from the navigator too.

**Fix:**
I was looking at the list of [Void Elements](https://developer.mozilla.org/en-US/docs/Glossary/Void_element) to filter other text related tags from the navigator, only `br` and `wbr` looked like a good match for us.

**Commit Details:**
- don't show zero size control on canvas when the child element is not a reparent target
- filter special tags from visible navigator elements

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -955,7 +955,11 @@ export const MetadataUtils = {
           const path = subTree.path
           const isHiddenInNavigator = EP.containsPath(path, hiddenInNavigator)
           navigatorTargets.push(path)
-          if (!collapsedAncestor && !isHiddenInNavigator) {
+          if (
+            !collapsedAncestor &&
+            !isHiddenInNavigator &&
+            !MetadataUtils.isElementTypeHiddenInNavigator(path, metadata)
+          ) {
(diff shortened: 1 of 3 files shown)


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -993,6 +997,19 @@ export const MetadataUtils = {
       maxSize: 1,
     },
   ),
+  isElementTypeHiddenInNavigator(path: ElementPath, metadata: ElementInstanceMetadataMap): boolean {
+    const VoidElementsToFilter = ['br', 'wbr']
```
Rheeseyb, line 1001: We should probably move this list to `dom-utils.ts`.

Rheeseyb (approved):
I suspect there are few others that we'll want to add to this list, but there doesn't seem to be a category of HTML elements we can use that captures them all.

editor/src/core/model/element-metadata-utils.ts
enidemi: fixed!

bkrmendy (approved):

PR #3071, opened 2023-01-09
Hug contents with children not participating in the layout

bkrmendy:
## Description

This PR fixes an omission in the `Hug contents` inspector strategy, by checking if the element the strategy is being applied to acutally has any children that participate in the layout (otherwise the element can be collapsed to become 0-sized).

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
@@ -387,6 +387,31 @@ export var storyboard = (
 )
 `
 
+const projectForEdgeDblClickNoChildren = `import * as React from 'react'
+import { Storyboard } from 'utopia-api'
+
+export var storyboard = (
+  <Storyboard>
+    <div
+      data-testid='mydiv'
+      style={{
+        backgroundColor: '#3EA881FC',
+        position: 'absolute',
+        left: -231,
+        top: 221,
+        width: 637,
+        display: 'flex',
+        gap: 31,
+        alignItems: 'center',
+        justifyContent: 'center',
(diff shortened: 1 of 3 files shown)


ruggi (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
```diff
@@ -1109,4 +1160,44 @@ describe('Double click on resize edge', () => {
...
+
+  it('not applicable when children are positioned absolute', async () => {
+    const editor = await renderTestEditorWithCode(
+      projectForEdgeDblClickWithPosition('absolute', 'absolute'),
+      'await-first-dom-report',
+    )
+    const div = await doDblClickTest(editor, edgeResizeControlTestId(EdgePositionBottom))
+    expect(div.style.width).toEqual('637px')
+    expect(div.style.height).toEqual('445px')
+  })
```
Rheeseyb, line 1182: It doesn't seem like there is much point in having both of these tests as they're doing the same thing. Maybe it would be better to have one of these, and a test without any children?

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
bkrmendy: absolutely, fixing it

bkrmendy: https://github.com/concrete-utopia/utopia/pull/3071/commits/00192c37ce0a457f3a4f0a1f33947fa11d94f912

PR #3072, opened 2023-01-09
Small tweaks that make some selectors faster.

seanparsons:
**Problem:**
Found a couple of inefficiencies when looking at some selector performance.

**Fix:**
Removed an unused property from the navigator items and slightly reordered a utility function.

**Commit Details:**
- Removed `staticElementName` from navigator items, as it's never used.
- Slight reordering in `isFocusableComponentFromMetadata`.


balazsbajorics (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1514,6 +1510,10 @@ export const MetadataUtils = {
     if (isImported) {
       return false
     }
+    const elementName = MetadataUtils.getJSXElementName(maybeEitherToMaybe(element?.element))
```
Rheeseyb, line 1513: Super minor, but this line could even be moved after the below conditional

Rheeseyb (approved):

editor/src/core/model/element-metadata-utils.ts
seanparsons: Fixed in 0330ba2b4.

PR #3075, opened 2023-01-10
Substores

balazsbajorics:
**Problem:**
We have (depending on the number of navigator items on screen) 200-500 useEditorState hooks. For every single dispatch we have to pay a ~2ms fixed price to call each of the useEditorState callbacks, and their equality functions. The problem is the sheer amount of callbacks, as far as I understand the individual callbacks / equality functions are as fast as they can be. 

**Investigation:**
The inevitable solution is to reduce the number of callbacks that need to be called for a certain dispatch. We need some way of predicting what useEditorState hooks are guaranteed to be not interested in the update. We can then skip calling these callbacks.
I tried looking into an automated approach using `react-tracked` that uses proxies. The idea would have been that when the selector executes, it calls a series of getters starting from EditorStore. ie if a selector is picking `store.editor.selectedViews`, then react-tracked will remember this, and next time the editorState changes, it pre-checks if store.editor.selectedViews remained the same. and only call the callback if not. The approach was promising, but it had two problems: the pre-checking is not free, and the proxy wrapping was _very_ expensive. With the 500 selectors each needing a separate proxy, with some of the selectors going very very deep in the editor state, the spike ended up slower than master.
So then the next idea was to just suck it up and split the editor store into sub-stores. Then the dispatch would update only the relevant stores (ie the store that contains selectedViews) and so each sub-store would only have a fraction of the subscribers. Problem: we need overlapping types for these sub-stores, because most selectors are interested in the same-ish parts of the editorState. (Ie. most selectors reading metadata also need to know the selectedViews, but not vice-versa.) Plus this refactor would be prohibitively large.

**Solution:**
So with input from Rheese and Sean we ended up going with a third route: for the sake of useEditorState callbacks, we _pretend_ that there are multiple stores, we call them Substores. Each store must be a strict subset of the original editorStore. This makes it possible to create types that restrict access to the EditorStore to only those sub-slices that are relevant for the selector, without having to actually create these sub-stores as objects in memory. 
Then we have a set of equality functions that we make sure are checking if anything in the given Substore changed. 
If and only if the substore changed, then we notify the subscribers of the given substore. But we sneakily pass in the full editorStore to each of the substores, so we don't need to create a bunch of temporary objects in memory.

**Results:**
The following results were taken from the Chrome performance profiling using the Before I Go test project.

- Scroll the canvas:
    - Before: Update Editor: ~3ms, Update Main Store: ~1.7ms
    - After: Update Editor: ~0.9ms, Update Main Store: ~0.4ms // _we get the biggest win for dispatched actions that don't touch the actual project contents or metadata. A part of this 0.4ms goes to the remaining uses of fullOldStore, and could be further minimized, but we decided this is enough of an improvement already._

- Move an element:
    - Before: Update Editor: ~4ms, Update Main Store: ~2ms
    - After: Update Editor: ~4ms, Update Main Store: ~2ms // **_<<<----- Note:_** _you can see we don't have much of an improvement here, and that is because an _incredible_ amount of the editor hooks focus on metadata and projectContents – which makes sense because most of the content on the screen are related to these two. But I'd argue that the current changes are also a good first step to be able to make these things faster too. we now own and control more of this update loop, so we can insert further specific optimizations in the future._

- Resize an element:
    - Before: Update Editor: ~5ms, Update Main Store: ~3.7ms
    - After: Update Editor: ~4.5ms, Update Main Store: ~2.5ms

**How to use:**
The `useEditorState` hook has a new parameter: you have to specify which Substore you are keying into.

Before:
```typescript
useEditorState(
  (store) => store.editor.selectedViews, 
  'MySelector selectedViews',
)
```

After:
```typescript
useEditorState(
  Substores.selectedViews,
  (store) => store.editor.selectedViews,
  'MySelector selectedViews',
)
```

The first parameter changes the type of `store` passed in to the selector. This means typescript will break if you try to access anything that doesn't exist in the specified Substore.
There's two special substores:
`Substores.restOfStore` – this is basically a grab-bag of every key in the EditorState that we didn't make a specific selector for.
`Substores.fullOldStore` – this is an escape hatch that lets the selector access the _entire_ EditorStore. This means that these selectors _will_ run for each and every dispatch. Ideally, we should have none of these. After spending a few days removing `Substores.fullOldStore` uses from the codebase, we decided we are at diminishing returns, and for this PR we think we are good enough. But for the future, we should keep an eye out to remove as much as possible of these, and never add new ones.

**Notes:**
Most of useEditorState calls in the codebase have been re-indented, I recommend viewing the PR with "whitespace: none"


bkrmendy (approved):

editor/src/components/editor/store/store-hook.ts
```diff
@@ -224,3 +315,157 @@ export function useSelectorWithCallback<U>(
...
+  selectedViews: SelectedViewsSubstate
+  focusedElement: FocusedElementPathSubstate
+  highlightedHoveredViews: HighlightedViewsSubstate
+  projectContents: ProjectContentSubstate
+  canvas: CanvasSubstate
+  canvasOffset: CanvasOffsetSubstate
+  derived: { derived: DerivedState }
+  restOfEditor: { editor: RestOfEditorState }
+  restOfStore: Omit<EditorStorePatched, 'editor' | 'derived'>
+  fullOldStore: EditorStorePatched
```
balazsbajorics, line 330: note to self: fix name and mark as deprecated for VSCode

editor/src/components/editor/store/store-hook.ts
```diff
@@ -224,3 +315,159 @@ export function useSelectorWithCallback<U>(
...
+      b.editor,
+    )
+  },
+  selectedViews: (a: SelectedViewsSubstate, b: SelectedViewsSubstate) =>
+    keysEquality(['selectedViews'], a.editor, b.editor),
+  focusedElement: (a: FocusedElementPathSubstate, b: FocusedElementPathSubstate) =>
+    keysEquality(['focusedElementPath'], a.editor, b.editor),
+  highlightedHoveredViews: (a: HighlightedViewsSubstate, b: HighlightedViewsSubstate): boolean => {
+    return (
+      // a.editor.selectedViews === b.editor.selectedViews &&
```
Rheeseyb, line 365:
```suggestion
```

editor/src/components/editor/store/store-hook.ts
```diff
@@ -224,3 +315,159 @@ export function useSelectorWithCallback<U>(
...
+export interface StoresAndSetState {
+  stores: UtopiaStores
+  setState: (store: EditorStorePatched) => void
+  getState: () => EditorStorePatched
+}
+
+export const createStoresAndState = (initialEditorStore: EditorStorePatched): StoresAndSetState => {
+  let latestStoreState: EditorStorePatched = initialEditorStore
+
+  let substores: UtopiaStores = objectMap((_, key) => {
```
Rheeseyb, line 432: `key` is unnecessary here too

editor/src/components/editor/store/store-hook.ts
```diff
@@ -224,3 +315,159 @@ export function useSelectorWithCallback<U>(
...
+} = Substores
+
+export type UtopiaStores = { [key in StoreKey]: Store<Substates[key]> }
+export interface StoresAndSetState {
+  stores: UtopiaStores
+  setState: (store: EditorStorePatched) => void
+  getState: () => EditorStorePatched
+}
+
+export const createStoresAndState = (initialEditorStore: EditorStorePatched): StoresAndSetState => {
```
Rheeseyb, line 429: I think this file, and this function in particular, could do with copious amounts of comments to explain what is happening and why here, as at this point it requires a fair amount of knowledge of how zustand works

editor/src/components/editor/store/store-hook-selectors.ts
```diff
@@ -0,0 +1,211 @@
...
+
+export interface MetadataSubstate {
+  editor: {
+    selectedViews: Array<ElementPath> // duplicated from SelectedHighlightedViewsSubstate, for convenience!
+    focusedElementPath: ElementPath | null
+    spyMetadata: ElementInstanceMetadataMap // this is coming from the canvas spy report.
+    domMetadata: ElementInstanceMetadataMap // this is coming from the dom walking report.
+    jsxMetadata: ElementInstanceMetadataMap // this is a merged result of the two above.
+    allElementProps: AllElementProps // the final, resolved, static props value for each element. // This is the counterpart of jsxMetadata. we only update allElementProps when we update jsxMetadata
+    // _currentAllElementProps_KILLME: AllElementProps // This is the counterpart of domMetadata and spyMetadata. we update _currentAllElementProps_KILLME every time we update domMetadata/spyMetadata
```
Rheeseyb, line 31:
```suggestion
```

editor/src/components/editor/store/store-hook-selectors.ts
```diff
@@ -0,0 +1,211 @@
...
+
+export interface FocusedElementPathSubstate {
+  editor: {
+    focusedElementPath: ElementPath | null
+  }
+}
+
+export interface HighlightedViewsSubstate {
+  editor: {
+    // selectedViews: Array<ElementPath>
```
Rheeseyb, line 49:
```suggestion
```

editor/src/utils/feature-switches.ts
```diff
@@ -33,6 +35,7 @@ let FeatureSwitches: { [feature in FeatureName]: boolean } = {
   'Debug – Performance Marks (Slow)': false,
   'Debug – Performance Marks (Fast)': false,
   'Debug – Measure Selectors': false,
+  'Selectors Split': true,
```
Rheeseyb, line 38: Is this used? I don't recall seeing it, so I'm guessing it can now be deleted

Rheeseyb (approved):

editor/src/utils/feature-switches.ts
balazsbajorics: ahhhh good find!

balazsbajorics: fixed

PR #3082, opened 2023-01-11
Render inline expressions in text editable nodes

### LINKED ISSUE #3085: Render inline expressions inside text editable nodes

ruggi:
Fixes #3085 

**Problem:**

Inline interpolated variables/js expressions inside text editable nodes should be evaluated.

**Fix:**

This PR implements the following behavior:

Kapture 2023-01-11 at 16 23 39

1. valid expressions inlined inside a text editable are rendered when outside text edit mode (e.g. typing `hello {"wor" + "ld"}` will be rendered as `hello world`)
2. curly braces that don't wrap a valid expression are escaped to their HTML entity
3. curly braces can be escaped by prepending a `\`, making the wrapped expression not be evaluated (and the braces get escaped in the generated code)

Note: the "validity" of an expression is not actually validated outside a simple regex. We _should_ do such evaluations later on in order to do more precise rendering and error management, but for now it should be fine enough.


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -345,8 +346,20 @@ export function renderCoreElement(
...
+      return c.text
+    case 'JSX_ELEMENT':
+      return c.name.baseVariable === 'br' ? '\n' : null
+    case 'JSX_ARBITRARY_BLOCK':
+      if (c.transpiledJavascript === `return ${c.javascript}`) {
+        return `{${c.originalJavascript}}`
+      }
+      return null
+    default:
+      return null
```
bkrmendy, line 361: minor, this should be `assertNeven(c)` instead so it'll be easier to extend the type (when it comes to that)

bkrmendy (approved):

seanparsons (approved):

seanparsons (approved):

gbalint (approved):

Rheeseyb (approved):

PR #3086, opened 2023-01-11
Canvas TextEditor should use the Main editor store

balazsbajorics:
**Problem:**
The TextEditorWrapper lives inside the canvas. we did not realize that this means that it uses the CanvasStore, but also relies on metadata which is often empty in the CanvasStore. 

**Fix:**
Create a `MainEditorStoreProvider` similar to `LowPriorityStoreProvider`, and wrap the TextEditorWrapper in this new `MainEditorStoreProvider`.

**Commit Details:**
- Renamed `low-priority-store.tsx` to `store-context-providers.tsx`
- Added `MainEditorStoreProvider` to `store-context-providers.tsx`
- Wrapped the entire editor and the test editors in `OriginalMainEditorStateContext.Provider` so `MainEditorStoreProvider` can retrieve the original main editor even in contexts where a `LowPriorityStoreProvider` or a canvas store provider is in effect


editor/src/components/text-editor/text-editor.tsx
```diff
@@ -125,7 +126,15 @@ const handleSetFontWeightShortcut = (
   ]
 }
 
-export const TextEditorWrapper = React.memo((props: TextEditorProps) => {
+export const TextEditorWrapperWrapper = React.memo((props: TextEditorProps) => {
```
Rheeseyb, line 129: Until we wrap it again...

Rheeseyb (approved):

bkrmendy (approved):

gbalint (approved):

PR #3088, opened 2023-01-11
Keyboard opacity strategy: stricter check of pressed keys

bkrmendy:
## Problem
Pressing option + cmd + 1 (which toggles the navigator) also triggered the keyboard opacity strategy, since it didn't check which keys were pressed strictly enough.

## Solution
Check both modifiers and keys pressed when determining the fithness of `keyboardSetOpacityStrategy`


enidemi (approved):

maltenuhn:
I've just tested it with all sorts of keyboard combos, both ones that get picked up by the editor as well as ones that get picked up by the OS. Opacity-via-keyboard looks to behave correctly now!

Rheeseyb (approved):

PR #3094, opened 2023-01-12
convert `width: 100%` to `flex-grow: 1` when adding `display: flex` to container

bkrmendy:
## Description
`flex-grow: 1` is the preferred way to make a child element expand to fill available space in a flex container, so when converting a container to flex, we convert `width: 100%` to `flex-grow: 1` on all children

### DIFF
--- a/editor/src/components/inspector/inspector-common.tsx
+++ b/editor/src/components/inspector/inspector-common.tsx
@@ -1,10 +1,15 @@
-import { MetadataUtils } from '../../core/model/element-metadata-utils'
+import { getSimpleAttributeAtPath, MetadataUtils } from '../../core/model/element-metadata-utils'
 import { isStoryboardChild } from '../../core/shared/element-path'
 import { mapDropNulls } from '../../core/shared/array-utils'
 import { ElementInstanceMetadataMap } from '../../core/shared/element-template'
 import { ElementPath } from '../../core/shared/project-file-types'
 import { FlexDirection } from './common/css-utils'
 import { assertNever } from '../../core/shared/utils'
+import { CanvasCommand } from '../canvas/commands/commands'
+import { deleteProperties } from '../canvas/commands/delete-properties-command'
+import { setProperty } from '../canvas/commands/set-property-command'
+import * as PP from '../../core/shared/property-path'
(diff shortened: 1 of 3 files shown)


editor/src/components/inspector/inspector-common.tsx
```diff
@@ -235,3 +240,26 @@ export function widthHeightFromAxis(axis: Axis): 'width' | 'height' {
...
+  if (element == null) {
+    return []
+  }
+  const matches =
+    defaultEither(
+      null,
+      getSimpleAttributeAtPath(right(element.props), PP.create(['style', 'width'])),
+    ) === '100%'
+
+  if (!matches) {
```
Rheeseyb, line 258: This should also be checking that the flex direction of the parent is horizontal, as we should doing the same thing with `height` if the direction is vertical

balazsbajorics (approved):

editor/src/components/inspector/inspector-common.tsx
```diff
@@ -246,6 +251,37 @@ export function widthHeightFromAxis(axis: Axis): 'width' | 'height' {
...
+): Array<CanvasCommand> {
+  const element = MetadataUtils.getJSXElementFromMetadata(metadata, elementPath)
+  if (element == null) {
+    return []
+  }
+
+  const parentFlexDirection =
+    MetadataUtils.findElementByElementPath(metadata, parentPath)?.computedStyle?.[
+      'flexDirection'
+    ] ?? 'row'
```
Rheeseyb, line 267: Minor point, but we have an existing function for this that you could alternatively use [`getSimpleFlexDirection`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/model/element-metadata-utils.ts#L388) (or `getFlexDirection` immediately above that).

editor/src/components/inspector/inspector-common.tsx
```diff
@@ -246,6 +251,37 @@ export function widthHeightFromAxis(axis: Axis): 'width' | 'height' {
   }
 }
 
+export function convertWidthToFlexGrow(
```
Rheeseyb, line 254: This should be renamed now since it's no longer specific to the width

Rheeseyb (approved):

PR #3096, opened 2023-01-12
Always render a span for the text editor

gbalint:
**Problem:**
Another try for https://github.com/concrete-utopia/utopia/pull/3093/
It seems you can not prevent an onClick with stopping the propagation of events, you have to preventDefault, but that disables text editing.
So I did not find any better solution then to remove the keyboard and mouse handler props from the parent element. There can be still components which react to these events, but that is a rarer occurence.
The real solution will be to extract the text editor to a control layer, but until then this PR makes the situation better.

### DIFF
--- a/editor/src/components/text-editor/text-editor.tsx
+++ b/editor/src/components/text-editor/text-editor.tsx
@@ -364,7 +364,7 @@ async function setSelectionToOffset(
   }
 }
 
-function stopPropagation(e: React.KeyboardEvent | React.ClipboardEvent | React.MouseEvent) {
+function stopPropagation(e: React.UIEvent | React.ClipboardEvent) {
   e.stopPropagation()
 }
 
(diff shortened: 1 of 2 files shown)


Rheeseyb (approved):

editor/src/components/text-editor/text-editor.tsx
```diff
@@ -362,11 +364,11 @@ async function setSelectionToOffset(
   }
 }
 
-function stopPropagation(e: React.KeyboardEvent | React.ClipboardEvent) {
+function stopPropagation(e: React.KeyboardEvent | React.ClipboardEvent | React.MouseEvent) {
```
Rheeseyb, line 367: FYI you could use `React.UIEvent` here

gbalint: ClipboardEvent is not part of that, but still a good tip, I replace the KeyboardEvent and the MouseEvent with that.

ruggi (approved):

PR #3098, opened 2023-01-13
modified navigator icons

lankaukk:
- Switched the navigator icons for flex row and flex icon from arrows, to new block-like icons. 
- Uploaded a new *clickable* icon to be used for items that are clickable elements (including, but not limited to buttons)
   - Began using this icon for buttons 
- Uploaded a new *pure-text* icon
   - Began using this icon for span, p, h1, h2, h3, h4, h5, and h6 elements
- Uploaded new *text-generated* icon

Screenshot 2023-01-12 at 9 46 36 PM

Screenshot 2023-01-13 at 10 21 16 AM


seanparsons (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -315,6 +315,50 @@ export const MetadataUtils = {
...
+      elementName != null &&
+      PP.depth(elementName.propertyPath) === 0 &&
+      (elementName.baseVariable === 'p' ||
+        elementName.baseVariable === 'span' ||
+        elementName.baseVariable === 'h1' ||
+        elementName.baseVariable === 'h2' ||
+        elementName.baseVariable === 'h3' ||
+        elementName.baseVariable === 'h4' ||
+        elementName.baseVariable === 'h5' ||
+        elementName.baseVariable === 'h6')
```
Rheeseyb, line 330: I'd suggest storing these strings in an array in `dom-utils.ts` like we do with e.g. [`intrinsicHTMLElementNamesThatSupportChildren `](https://github.com/concrete-utopia/utopia/blob/6f8dbc103f8e83dd311a7f6a4166ac91622e9659/editor/src/core/shared/dom-utils.ts#L179), then we could use a `foldEither` on the `element.element` along the lines of:
```ts
if (element == null) { return false }
const isTextElement = foldEither(
  (elementString) => textElements.includes(elementString),
  (elementInstance) => 
    // TS might be alright with this...
    isJSXElement(elementInstance) && 
    textElements.includes(elementInstance.name.baseVariable),
  element.element
)
```
(that's based on the far more complex than we need here [`targetElementSupportsChildrenAlsoText`](https://github.com/concrete-utopia/utopia/blob/6f8dbc103f8e83dd311a7f6a4166ac91622e9659/editor/src/core/model/element-metadata-utils.ts#L732))

Rheeseyb (approved):

editor/src/core/shared/dom-utils.ts
```diff
@@ -204,6 +204,8 @@ export const intrinsicHTMLElementNamesThatSupportChildren: Array<string> = [
   'span',
 ]
 
+export const textElements = ['text', 'p', 'span', 'h1', 'h2', 'h2', 'h3', 'h5', 'h6']
```
balazsbajorics, line 207: ultra minor nitpick, but we use PascalCase for global consts 
```suggestion
export const TextElements = ['text', 'p', 'span', 'h1', 'h2', 'h2', 'h3', 'h5', 'h6']
```

balazsbajorics (approved):
very nice, clean PR!

maltenuhn:
@lankaukk we also have versions of the cube icon without the connecting lines visible, and a couple of size variants - are any of those still in need of updating to make everything consistent?

lankaukk:
> 

good call, I'll remove those and make sure the existing cubes are consistent.

PR #3102, opened 2023-01-13
fix(canvas) text editing can scroll the canvas

enidemi:
**Problem:**
When writing in text edit mode without fixed width/height set on the span near the edge of the screen the text content can scroll the canvas away automatically. After finishing text editing canvas interactions are incorrectly positioned as our canvas offset is not following this change.

**Fix:**
I found that the 'canvas-root' div receives the scroll event triggered by text editing. When this happens I dispatch the action to update the canvas offset and also reset the scroll offset of the canvas-root div.

**Commit Details:**
- `EditorCanvas ` receives `onScroll` event handler

### DIFF
--- a/editor/src/templates/editor-canvas.tsx
+++ b/editor/src/templates/editor-canvas.tsx
@@ -758,6 +758,21 @@ export class EditorCanvas extends React.Component<EditorCanvasProps> {
     event.preventDefault()
   }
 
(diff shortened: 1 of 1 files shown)


bkrmendy (approved):

Rheeseyb (commented):
👍

PR #3106, opened 2023-01-13
Collapse adjacent compatible text segments.

seanparsons:
**Problem:**
Sometimes there are an unnecessary number of adjacent elements containing text which can be combined to provide a simpler editing experience.

**Fix:**
When updating the text in an element, the logic in `collapseTextElements` attempts to find a run of elements adjacent to the target which can be combined and if so combines them into a single element.

**Commit Details:**
- Added `collapseTextElements` and supporting functions which handle combining text elements adjacent to the targeted element.
- Separated out `getComponentFromCode` so that it can be reused elsewhere.
- Fixed an issue in a `useEffect` cleanup handler for `TextEditor` which would cause an additional `UPDATE_CHILD_TEXT` to fire, thereby instantly combining the elements when the text editor was activated.
- Hooked `collapseTextElements` into the `UPDATE_CHILD_TEXT` action.


ruggi (approved):

editor/src/components/text-editor/text-handling.ts
```diff
@@ -0,0 +1,253 @@
...
+    isJSXElement(elementToCheck) &&
+    // Element to check only has a single child...
+    elementToCheck.children.length === 1 &&
+    // ...which is text.
+    isJSXTextBlock(elementToCheck.children[0]) &&
+    // The two elements share a type.
+    getJSXElementNameAsString(elementToCheck.name) ===
+      getJSXElementNameAsString(targetElement.name) &&
+    // The two elements have compatible properties.
+    arePropsCompatible(targetElement.props, elementToCheck.props)
```
Rheeseyb, line 148: This is honestly a thing of beauty

Rheeseyb (approved):

PR #3115, opened 2023-01-16
Integrate the eye dropper into the colour picker

bkrmendy:
pipette

## Problem 
The eye dropper is very useful, but unfortunately very undiscoverable as well.

## Solution
Add a button in the colour picker that triggers the eye dropper.

### DIFF
--- a/editor/src/components/inspector/controls/color-picker.tsx
+++ b/editor/src/components/inspector/controls/color-picker.tsx
@@ -46,7 +46,7 @@ export const ColorPicker: React.FunctionComponent<React.PropsWithChildren<ColorP
   ...props
 }) => {
   const onSubmitValue = props.onSubmitValue
-  const dispatchEyeDropper = React.useCallback(() => {
+  const onClickEyeDropper = React.useCallback(() => {
     closePopup()
     void pickColorWithEyeDropper()
       .then(({ sRGBHex }) => {
@@ -78,7 +78,7 @@ export const ColorPicker: React.FunctionComponent<React.PropsWithChildren<ColorP
           color='secondary'
           width={18}
           height={18}
-          onClick={dispatchEyeDropper}
+          onClick={onClickEyeDropper}
           style={{ marginLeft: 8 }}
         />
         <ColorPickerInner {...props} />


editor/src/components/inspector/sections/style-section/background-subsection/background-picker.tsx
```diff
@@ -466,6 +476,31 @@ export const BackgroundPicker: React.FunctionComponent<
...
   )
 
+  const closePopup = props.closePopup
+
+  const dispatchEyeDropper = React.useCallback(() => {
+    const selectedViews = selectedViewsFromStore
+    if (selectedViews.length === 0) {
+      return
+    }
+    const EyeDropper = window.EyeDropper
```
seanparsons, line 486: This segment looks a bit like we could wrap it up in a little function as it's the same in both files changed in this PR.

seanparsons (approved):

editor/src/components/inspector/controls/color-picker.tsx
```diff
@@ -37,6 +44,17 @@ export const ColorPicker: React.FunctionComponent<React.PropsWithChildren<ColorP
   portalTarget,
   ...props
 }) => {
+  const onSubmitValue = props.onSubmitValue
+  const dispatchEyeDropper = React.useCallback(() => {
```
balazsbajorics, line 48: I think dispatch is misleading here, I think it should be something like onEyeDropper, onClickEyeDropper or handleEyeDropper

```suggestion
  const onClickEyeDropper = React.useCallback(() => {
```

balazsbajorics (approved):

editor/src/components/inspector/controls/color-picker.tsx
```diff
@@ -37,6 +44,17 @@ export const ColorPicker: React.FunctionComponent<React.PropsWithChildren<ColorP
...
+    const EyeDropper = window.EyeDropper
+    if (EyeDropper == null) {
+      return
+    }
+    closePopup()
+    void new EyeDropper().open().then((result: any) => {
```
Rheeseyb, line 54: I dislike this `any`. Assuming there is no way we can guarantee type safety here, can we instead fall back to using `unknown` and check that is is an object with a `sRGBHex` key of type string before trying to then use that value. Even if we fail silently or write a console error if not.

editor/src/components/inspector/sections/style-section/background-subsection/background-picker.tsx
Rheeseyb: It does look like it would work well as a `useEyeDropper`

bkrmendy: I created a utility function in https://github.com/concrete-utopia/utopia/pull/3115/commits/79100a9a4b198caa19167298c517d86bccf8c5df, we can put it into a use if this turns out to be a feature that's used in a lot of places

editor/src/components/inspector/controls/color-picker.tsx
bkrmendy: absolutely! fixed in https://github.com/concrete-utopia/utopia/pull/3115/commits/79100a9a4b198caa19167298c517d86bccf8c5df

editor/src/components/inspector/controls/color-picker.tsx
bkrmendy: I applied your change manually in https://github.com/concrete-utopia/utopia/pull/3115/commits/34d554fd5b886f09b40e6dda45f117673941acc9

PR #3117, opened 2023-01-16
Fix/text editor fix default cursor

enidemi:
**Problem:**
When in text edit mode the cursor switches back to default.

**Fix:**
Pass cursor prop to text edit overlay and use it. I also noticed that the default text edit mode cursor is 'Insert', but it's actually not in insert mode, so I changed it to 'Select' cursor.

Small fix during insertion: On hover on a text editable field in insert mode shows 'Select' cursor too.

**Commit Details:**
- update`TextEditCanvasOverlay` 
- when mouseover on 'textEditable' field during insert strategy the set cursor command is added

### DIFF
--- a/editor/src/templates/editor-canvas.tsx
+++ b/editor/src/templates/editor-canvas.tsx
@@ -150,7 +150,7 @@ function getDefaultCursorForMode(mode: Mode): CSSCursor {
     case 'live':
       return CSSCursor.BrowserAuto
     case 'textEdit':
-      return CSSCursor.Insert
+      return CSSCursor.Select
     default:
       const _exhaustiveCheck: never = mode
       throw `Unable to get default cursor for unsupported mode ${(mode as any).type}`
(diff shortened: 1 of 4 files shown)


editor/src/templates/editor-canvas.tsx
```diff
@@ -150,7 +150,7 @@ function getDefaultCursorForMode(mode: Mode): CSSCursor {
     case 'live':
       return CSSCursor.BrowserAuto
     case 'textEdit':
-      return CSSCursor.Insert
+      return CSSCursor.Select
```
ruggi, line 153: mmmh I'm not super sure about this one: having the cursor change from outside a text editable area (a crosshair) from inside (a pointer) feels weird to me 🤔

gbalint: This is exactly how figma works, I think it is correct.

Figma:
16-42-dtylt-pdnh1

Eni's change:
16-44-b9hv1-lfw6f

It shows you where can you edit the text.

gbalint (approved):

editor/src/templates/editor-canvas.tsx
ruggi: gotcha 👍

ruggi (approved):

editor/src/templates/editor-canvas.tsx
enidemi: With this fix inside the text area the chrome text cursor is used automatically.
I think the insert cursor doesn't fit here because clicking outside of the text area just finishes the text editing but doesn't allow inserting?

Rheeseyb (approved):

PR #3118, opened 2023-01-16
Fill container: respect flex direction of the parent container

bkrmendy:


editor/src/components/inspector/inspector-strategies/inspector-strategies.tsx
```diff
@@ -3,6 +3,8 @@ import { setProperty } from '../../canvas/commands/set-property-command'
 import {
   Axis,
   convertWidthToFlexGrow,
+  detectFlexDirection,
```
balazsbajorics, line 6: unused import?

editor/src/components/inspector/inspector-strategies/inspector-strategies.tsx
```diff
@@ -136,12 +138,29 @@ export const setPropFillStrategies = (axis: Axis): Array<InspectorStrategy> => [
...
+      }
+
+      const flexDirection = detectFlexDirectionOne(metadata, EP.parentPath(path)) ?? 'row'
+
+      if (
+        (flexDirection.startsWith('row') && axis === 'vertical') ||
+        (flexDirection.startsWith('column') && axis === 'horizontal')
+      ) {
+        return [
+          nukePropsCommand(path),
```
balazsbajorics, line 155: is there an importance to the order of nukePropsCommand and setProperty? here nukePropsCommand comes first, on line 162, nukePropsCommand comes last

balazsbajorics (approved):

editor/src/components/inspector/inspector-strategies/inspector-strategies.tsx
bkrmendy: removed it, thanks for spotting it!

editor/src/components/inspector/inspector-strategies/inspector-strategies.tsx
bkrmendy: the order does matter in the general case, because these commands are passed to [`foldAndApplyCommandsSimple`](https://github.com/concrete-utopia/utopia/blob/82c5361ce43233c4eda86b2b25d0dc551db5921c/editor/src/components/canvas/commands/commands.ts#L177), so they are applied sequentially.

However, in these two specific cases the order is OK, because
- in the place where you commented, the width/height prop is removed first and then it is readded as the last step,
- in the other branch, the width/height prop is removed, and flex-grow is added (which are orthogonal, so the order doesn't matter).

PR #3119, opened 2023-01-16
Text editor inherits font settings.

seanparsons:
**Problem:**
With an element that has custom font settings like `fontFamily`, when triggering the text editor, the font family appears to "reset" to that of some global default.

**Fix:**
When constructing the `TextEditor` component, all the individual `font*` prefixed properties are set to `'inherit'` so that the settings cascade down from the element that is the container for the text (and when editing the text editor itself).

**Commit Details:**
- Set a bunch of `font*` properties in the editor editor to `'inherit'`.


maltenuhn:
Do we also need to force-inherit other typographic properties (like `line-height` or `letter-spacing`)? Or are these ok because in any event they can't be edited while in text edit mode?

enidemi:
I think there is at least one more style property to add, `text-decoration`  is modified via `cmd + u` and `cmd + shift + x` shortcuts.

seanparsons:
> I think there is at least one more style property to add, `text-decoration` is modified via `cmd + u` and `cmd + shift + x` shortcuts.

There's now a much more comprehensive list of properties to inherit.

gbalint (approved):

enidemi (approved):

PR #3120, opened 2023-01-16
Support retaining UIDs when changing multiple elements

### LINKED ISSUE #3108: Parsing a code file which will have more than one UID change prevents us from retaining old UIDs and breaks selection
When parsing a file that was previously parsed we will attempt to retain old UIDs for the sake of retaining selection. However, we'll only do this if only one UID has changed. If more than one UID has changed, we will bail on that process, which means selection can be lost. This is particularly common when inserting new elements via the canvas (where we should be selecting the newly inserted element). The cause is [this check in `fixParseSuccessUIDs`](https://github.com/concrete-utopia/utopia/blob/f06f16d549035e33afcd9b2994c18e45c582d6eb/editor/src/core/workers/parser-printer/uid-fix.ts#L73-L74).

Rheeseyb:
Fixes #3108

**Problem:**
The uid-fix behaviour is a bit too strict, and fails if there are changes to multiple elements at the same time.

**Fix:**
If we discover more than one UID in the file has changed, we then check that the number of elements has stayed the same, in which case we assume that we're still fine to retain the old UIDs. This also requires us to sort the results in descending order of the length of `pathsToModify`, since those are based on the new UIDs and not the old ones (meaning that correcting a parent's UID would then prevent us from being able to also correct a child's UID).

I've also added a test to check that element re-ordering is still compatible with this change.

### DIFF
--- a/editor/src/core/workers/parser-printer/uid-fix.ts
+++ b/editor/src/core/workers/parser-printer/uid-fix.ts
@@ -47,7 +47,7 @@ export function fixParseSuccessUIDs(
     }
   } = {}
 
-  zipTopLevelElements(
+  const numberOfElementsMatch = zipTopLevelElements(
     oldParsed.topLevelElements,
     newParsed.topLevelElements,
     (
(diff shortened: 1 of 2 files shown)


editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -144,61 +146,90 @@ function zipTopLevelElements(
   firstTopLevelElements: Array<TopLevelElement>,
   secondTopLevelElements: Array<TopLevelElement>,
   onElement: OnElement,
-): void {
+): boolean {
   const firstComponents = getComponentsFromTopLevelElements(firstTopLevelElements)
   const secondComponents = getComponentsFromTopLevelElements(secondTopLevelElements)
 
+  const lengthsMatchAtThisLevel = firstComponents.length === secondComponents.length
+  let numberOfElementsMatch = lengthsMatchAtThisLevel
```
seanparsons, line 154: For some reason when I see this I think it should hold a quantity.

seanparsons (approved):

ruggi (approved):
lgtm

PR #3123, opened 2023-01-17
Never escape curly braces

gbalint:
**Problem:**
When there are multiple `{}` expressions in a text content, it escapes some of them but not others, and the canvas shows the `Unexpected digit after hash token` exception

**Fix:**
Remove regex based brittle logic to decide which curly brace to escape and which not, and instead never escape curly braces, and leave escaping to the user.


Rheeseyb (approved):

balazsbajorics (approved):

seanparsons (approved):

ruggi (approved):
can you also please delete the `reValidInlineJSXExpression` regex?

note: if we want to support explicit escaping with  `\` we should remember to reintroduce that too

PR #3130, opened 2023-01-18
try to fail the Puppeteer tests instead of a silent error

balazsbajorics:
Instead of printing an error message in the comments, 
image
we should instead properly fail the PR in case the performance tests have a problem:
image

**Details:**
- In the catch block instead of logging the error, exit the node process with an error code


balazsbajorics:
awesome, the test "successfully" fails!! 
image

Rheeseyb (approved):

enidemi (approved):

PR #3143, opened 2023-01-20
Fire the nine-block control on mousedown

bkrmendy:
## Description
- the nine-block control is fired on mousedown for the "instant" effect
- fix a regression in the nine-block control so non-selected states are shown with 0.5 opacity on hover
- use `cursor: pointer` on the nine-block control cells

### DIFF
--- a/editor/src/components/inspector/nine-block-controls.tsx
+++ b/editor/src/components/inspector/nine-block-controls.tsx
@@ -5,7 +5,6 @@ import { cartesianProduct } from '../../core/shared/array-utils'
 import { size, Size } from '../../core/shared/math-utils'
 import { useColorTheme } from '../../uuiui'
 import { useDispatch } from '../editor/store/dispatch-context'
-import { EditorStorePatched } from '../editor/store/editor-state'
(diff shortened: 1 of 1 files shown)


editor/src/components/inspector/nine-block-controls.tsx
```diff
@@ -174,7 +174,7 @@ const NineBlockControlCell = React.memo<NineBlockControlCellProps>((props) => {
 
   return (
     <div
-      onClick={onClick}
+      onMouseDown={onClick}
```
ruggi, line 177: nit: maybe worth renaming the `onClick` prop as well? either `onMouseDown` or a more specific `onOpen` or something?

ruggi (approved):

editor/src/components/inspector/nine-block-controls.tsx
bkrmendy: Great idea! I ended up using `onSelect`

maltenuhn (approved):
So much better experience using this!!

PR #3148, opened 2023-01-20
`shift + a` to convert element to flex

bkrmendy:
## Description
This PR adds a shortcut (`shift + a`) which converts the selected element to flex containers.

Under the hood, this uses an existing inspector strategy set (`addFlexLayoutStrategies`)


editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -749,6 +759,24 @@ export function handleKeyDown(
...
+          editor.jsxMetadata,
+          editor.selectedViews,
+        )
+        const commands = commandsForFirstApplicableStrategy(
+          editor.jsxMetadata,
+          editor.selectedViews,
+          selectedElementsFlexContainers ? removeFlexLayoutStrategies : addFlexLayoutStrategies,
+        )
+        if (commands != null) {
+          return [EditorActions.applyCommandsAction(commands)]
```
ruggi, line 776: nit: just for consistency inside this function which keep the happy path to the left, I'd flip this condition and return `[]` here

ruggi (approved):

maltenuhn:
This is really nice to use, and the toggleability is (👨‍🍳, 💋 ). Once we pair this with more opinionated behaviour for stripping any position attributes for children...

seanparsons (approved):

PR #3152, opened 2023-01-20
Clear min/max fields when giving a flex element a fixed size dimension.

seanparsons:
**Problem:**
When setting the cross axis size of a flex element to a fixed size, the size may not actually appear to apply if there are `minX` and/or `maxX` properties associated with that axis.

**Fix:**
Now if the new value of the property is of a fixed size unit (like `px`), we clear the `minX` and `maxX` properties aassociated with it at the same time.

**Commit Details:**
- Made some values like `AbsoluteLengthUnits` be exported.
- Created `isFixedSize` utility function.
- Changed `setPropFixedStrategies` back to having `CSSNumber` as the type of `value, as well as adding the parameter `whenToRun`.
- `setPropFixedStrategies` now clears the `minX` and `maxX` properties relevant to the axis that is being updated.
- `PinsLayoutNumberControl` now takes an optional `fixedSizeHandler` property, which if exists is called when the property is a fixed size number.

### DIFF
--- a/editor/src/components/inspector/sections/layout-section/flex-element-subsection/flex-element-subsection.spec.browser2.tsx
+++ b/editor/src/components/inspector/sections/layout-section/flex-element-subsection/flex-element-subsection.spec.browser2.tsx
@@ -18,12 +18,6 @@ import {
 import { act, fireEvent } from '@testing-library/react'
 import { applyPrettier } from 'utopia-vscode-common'
 
-function downClickUp(targetElement: HTMLElement, mouseEventInit: MouseEventInit) {
-  fireEvent(targetElement, new MouseEvent('mousedown', mouseEventInit))
-  fireEvent(targetElement, new MouseEvent('click', mouseEventInit))
-  fireEvent(targetElement, new MouseEvent('mouseup', mouseEventInit))
-}
-
 function getCrossCapitalDimension(flexDirection: FlexDirection): string {
   return flexDirection.startsWith('row') ? 'Height' : 'Width'
 }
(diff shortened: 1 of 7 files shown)


editor/src/components/inspector/sections/layout-section/flex-element-subsection/flex-element-subsection.spec.browser2.tsx
```diff
@@ -0,0 +1,277 @@
...
+  LengthUnits,
+} from '../../../../../components/inspector/common/css-utils'
+import { act, fireEvent } from '@testing-library/react'
+import { applyPrettier } from 'utopia-vscode-common'
+
+function downClickUp(targetElement: HTMLElement, mouseEventInit: MouseEventInit) {
+  fireEvent(targetElement, new MouseEvent('mousedown', mouseEventInit))
+  fireEvent(targetElement, new MouseEvent('click', mouseEventInit))
+  fireEvent(targetElement, new MouseEvent('mouseup', mouseEventInit))
+}
```
Rheeseyb, line 25: `mouseClickAtPoint` from event-helpers.test-utils.ts should do the job here (it will also ensure those events are fired in the right order), unless you specifically didn't want to provide the mouse point here? If that's a suitable use case then maybe we should address that in event-helpers.test-utils.ts

Rheeseyb (approved):

editor/src/components/inspector/sections/layout-section/flex-element-subsection/flex-element-subsection.spec.browser2.tsx
seanparsons: I did have that in there originally, this is a victim of me having some grief with the input box itself. Will look into restoring that now.

seanparsons: Fixed in 5d8f5a6a5.

balazsbajorics (approved):

PR #3154, opened 2023-01-23
Resize to fit control & "Size" section

bkrmendy:
## Description
This PR adds a "Size to fit" button, which makes the selected element resize to fit its contents (if applicable). The control reuses the `setPropHugStrategies` strategy set (on both axes).

Since there are two new sizing controls with this PR (Resize to Fit and Fixed/Fill/Hug), a new Size section is created for these two controls.

### DIFF
--- a/editor/src/components/inspector/resize-to-fit-control.tsx
+++ b/editor/src/components/inspector/resize-to-fit-control.tsx
@@ -48,8 +48,12 @@ export const ResizeToFitControl = React.memo<ResizeToFitControlProps>(() => {
 
   return (
     <Tooltip title={'Resize to Fit'}>
-      <FlexRow data-testid={ResizeToFitControlTestId} onMouseDown={onMouseDown}>
-        <Icn type='warningtriangle' color='warning' width={16} height={16} />
+      <FlexRow
+        data-testid={ResizeToFitControlTestId}
+        onMouseDown={onMouseDown}
+        style={{ cursor: 'pointer' }}
+      >
+        <Icn type='fitToChildren' color='main' category='layout/commands' width={18} height={18} />
       </FlexRow>
     </Tooltip>
   )
--- a/editor/src/components/inspector/resize-to-fit-control.tsx
+++ b/editor/src/components/inspector/resize-to-fit-control.tsx
@@ -1,10 +1,14 @@
 import React from 'react'
 import { FlexRow, Icn, Tooltip } from '../../uuiui'
 import { EditorAction, EditorDispatch } from '../editor/action-types'
+import { applyCommandsAction } from '../editor/actions/action-creators'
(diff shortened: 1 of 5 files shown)


editor/src/components/inspector/resize-to-fit-control.tsx
```diff
@@ -0,0 +1,60 @@
...
+    )
+    executeFirstApplicableStrategy(
+      batchedDispatch,
+      metadata,
+      selectedViews,
+      setPropHugStrategies('vertical'),
+    )
+    if (batchedActions.length > 0) {
+      dispatch(batchedActions)
+    }
```
Rheeseyb, line 46: This is fine as it stands, but if we start using this pattern more often it might make more sense for the `executeFirstApplicableStrategy` to return an array of actions to dispatch instead.

Rheeseyb (approved):

editor/src/components/inspector/resize-to-fit-control.tsx
bkrmendy: we already have a similar function, `commandsForFirstApplicableStrategy`, I'll update the PR to use that

editor/src/components/inspector/resize-to-fit-control.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { setPropHugStrategies } from './inspector-strategies/inspector-strategies'
+import { executeFirstApplicableStrategy } from './inspector-strategies/inspector-strategy'
+
+export const ResizeToFitControlTestId = 'ResizeToFitControlTestId'
+
+interface ResizeToFitControlProps {}
+
+export const ResizeToFitControl = React.memo<ResizeToFitControlProps>(() => {
+  const dispatch = useDispatch()
+  const selectedViews = useEditorState(
```
balazsbajorics, line 15: you can make it a never-rerendering component if you make the selectedViews and metadata refEditorStates! since you only need it in the callback anyways.

balazsbajorics (approved):

editor/src/components/inspector/resize-to-fit-control.tsx
bkrmendy: Thanks for spotting it! Updated to use refs in https://github.com/concrete-utopia/utopia/pull/3154/commits/7aa95563d3493e736fdeaf1b1859da9a6b7bf59e

editor/src/components/inspector/resize-to-fit-control.tsx
bkrmendy: (updated in https://github.com/concrete-utopia/utopia/pull/3154/commits/7aa95563d3493e736fdeaf1b1859da9a6b7bf59e)

PR #3158, opened 2023-01-23
Picking "Fill Container" will remove positioning props

Rheeseyb:
**Problem:**
Setting "Fill Container" on an absolute positioned element would leave the offset, meaning it wouldn't actually fill the container.

**Fix:**
Remove the offset props in the given direction when setting an axis to "Fill Container". If the other axis is also set to fill, then we also remove `position: 'absolute'`.


bkrmendy (approved):

balazsbajorics (changes requested):
if there _was_ a position: absolute or position: relative, when nuking it we should add a `contain: 'layout'`

Rheeseyb:
> if there _was_ a position: absolute or position: relative, when nuking it we should add a `contain: 'layout'`

Good catch. Done 👍

PR #3159, opened 2023-01-23
Better whitespace trimming

gbalint:
**Problem:**
When you have a long line in the text editor, it is split into multiple lines in the code editor (by prettier).
These lines caused a problem next time you opened the text editor: the newline and the extra lines are removed so the words before and after the newline were merged.

**Fix:**
The whitespace sanitization logic become too complex and scattered, and moved it into a single place and simplified it.


bkrmendy (approved):

ruggi (approved):
nice

balazsbajorics (commented):
+1

PR #3160, opened 2023-01-23
Avoid container collapsing by setting children to fixed size

bkrmendy:
## Problem

When a child of a container that is not sized by its contents is set to `Fill container`, and the container is set to `Hug contents`, the element can become zero sized (if all children are set to `Fill container`, the container can become zero-sized as well)

## Fix 
Children can be in one of 4 states with respect to their sizing (according to https://github.com/concrete-utopia/utopia/blob/9bf46bb169ed167db19f7465c38b31276e02ff74/editor/src/components/inspector/inspector-common.ts#L368):
- Fixed size: this will size the parent container when it is set to hug contents, so no need for changing it
- Hug contents: we assume (hope) that an element cannot become zero sized as a result of setting it to "Hug contents". This means that if an element is set to "Hug contents", it is already sized by its children, so no need to change anything here either.
- Fill container: a child that is set to fill container **might** possibly be sized by its children. However, to follow the principle of least surprise, these element are converted to fixed size, so changes in the parent don't affect their visual appearance.
- Indeterminate: the function we use to determine the sizing of an element is partial (for now). In this case, we set the child to fixed size (so the parent will be in a valid "Hug contents" state).

Bottom line:
- if a child is set to Fill container or indeterminate sizing, set it to Fixed size
- otherwise leave it as it is


enidemi (approved):

editor/src/components/inspector/inspector-strategies/fixed-fill-hug.spec.browser2.tsx
```diff
@@ -131,6 +131,153 @@ describe('Fixed / Fill / Hug control', () => {
...
+        const button = (await editor.renderedDOM.findAllByText(HugContentsLabel))[0]
+        mouseClickAtPoint(button, { x: 5, y: 5 })
+
+        expect(parent.style.width).toEqual(MaxContent)
+        expect(child.style.width).toEqual('302px')
+        expect(parent.style.height).toEqual('759px')
+        expect(child.style.height).toEqual('141px')
+      })
+
+      it('child is set to fixed size on the vertical axis', async () => {
```
balazsbajorics, line 209: 'child is set to fixed size on the horizontal axis, no conversion happens'

editor/src/components/inspector/inspector-strategies/fixed-fill-hug.spec.browser2.tsx
```diff
@@ -131,6 +131,153 @@ describe('Fixed / Fill / Hug control', () => {
...
+        const button = (await editor.renderedDOM.findAllByText(HugContentsLabel))[0]
+        mouseClickAtPoint(button, { x: 5, y: 5 })
+
+        expect(parent.style.width).toEqual('700px')
+        expect(child.style.width).toEqual('302px')
+        expect(parent.style.height).toEqual(MaxContent)
+        expect(child.style.height).toEqual('141px')
+      })
+
+      it('child is set to hug contents on the horizontal axis', async () => {
```
balazsbajorics, line 233: 'child is set to hug contents on the horizontal axis, no conversion happens'

editor/src/components/inspector/inspector-strategies/fixed-fill-hug.spec.browser2.tsx
```diff
@@ -131,6 +131,153 @@ describe('Fixed / Fill / Hug control', () => {
...
+        const button = (await editor.renderedDOM.findAllByText(HugContentsLabel))[0]
+        mouseClickAtPoint(button, { x: 5, y: 5 })
+
+        expect(parent.style.width).toEqual(MaxContent)
+        expect(child.style.width).toEqual(MaxContent)
+        expect(parent.style.height).toEqual('751px')
+        expect(child.style.height).toEqual(MaxContent)
+      })
+
+      it('child is set to hug contents on the vertical axis', async () => {
```
balazsbajorics, line 257: 'child is set to hug contents on the vertical axis, no conversion happens'

balazsbajorics (approved):

editor/src/components/inspector/inspector-strategies/fixed-fill-hug.spec.browser2.tsx
bkrmendy: Thanks! I added the clarifications in https://github.com/concrete-utopia/utopia/pull/3160/commits/fb5b8bdea4b1b29ffceb7a288182d1d697824255

editor/src/components/inspector/inspector-strategies/fixed-fill-hug.spec.browser2.tsx
bkrmendy: added in https://github.com/concrete-utopia/utopia/pull/3160/commits/fb5b8bdea4b1b29ffceb7a288182d1d697824255

PR #3163, opened 2023-01-23
Canvas Resize nukes min max too

balazsbajorics:
**Problem:**
With #3152 we gained a functionality to automatically remove conflicting width/height properties (min-width, max-width, min-height, max-height) when setting width or height from the inspector.
But this did not cover the canvas strategies.

**Fix:**
Move the automatic prop removal to AdjustCssLengthProperty and SetCssLengthProperty commands. This ensures that all places using these commands will share the same behavior.

**Commit Details:**
- Extended SetCssLengthProperty so it can take an explicit CssNumber value
- Make `setPropFixedStrategies` use `SetCssLengthProperty`
- Removed prop removal code from `setPropFixedStrategies`
- Add prop removal code to SetCssLengthProperty
- Also use prop removal code in AdjustCssLengthProperty

### DIFF
--- a/editor/src/components/canvas/commands/set-css-length-command.ts
+++ b/editor/src/components/canvas/commands/set-css-length-command.ts
@@ -12,35 +12,36 @@ import * as PP from '../../../core/shared/property-path'
 import { EditorState, withUnderlyingTargetFromEditorState } from '../../editor/store/editor-state'
 import {
   CSSNumber,
+  cssPixelLength,
   parseCSSPercent,
-  parseCSSPx,
   printCSSNumber,
 } from '../../inspector/common/css-utils'
 import { applyValuesAtPath } from './adjust-number-command'
 import { BaseCommand, CommandFunction, WhenToRun } from './commands'
 
+type LengthValue =
+  | { type: 'EXPLICIT_CSS_NUMBER'; value: CSSNumber }
+  | { type: 'KEEP_ORIGINAL_UNIT'; valuePx: number; parentDimensionPx: number | undefined }
+
 export interface SetCssLengthProperty extends BaseCommand {
   type: 'SET_CSS_LENGTH_PROPERTY'
   target: ElementPath
   property: PropertyPath
-  valuePx: number
-  parentDimensionPx: number | undefined
+  value: LengthValue
 }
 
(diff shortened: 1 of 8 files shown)


editor/src/components/canvas/commands/set-css-length-command.ts
```diff
@@ -112,6 +132,10 @@ export const runSetCssLengthProperty: CommandFunction<SetCssLengthProperty> = (
     editorStatePatches: [propertyUpdatePatch],
     commandDescription: `Set Css Length Prop: ${EP.toUid(command.target)}/${PP.toString(
       command.property,
-    )} by ${command.valuePx}`,
+    )} by ${
+      command.value.type === 'EXPLICIT_CSS_NUMBER'
+        ? command.value.value.value
```
enidemi, line 137: beautiful

enidemi (approved):

editor/src/components/canvas/commands/set-css-length-command.ts
```diff
@@ -12,42 +12,51 @@ import * as PP from '../../../core/shared/property-path'
...
-  parseCSSPx,
   printCSSNumber,
 } from '../../inspector/common/css-utils'
+import { deleteConflictingPropsForWidthHeight } from './adjust-css-length-command'
 import { applyValuesAtPath } from './adjust-number-command'
 import { BaseCommand, CommandFunction, WhenToRun } from './commands'
 
+type CssNumberOrKeepOriginalUnit =
+  | { type: 'EXPLICIT_CSS_NUMBER'; value: CSSNumber }
+  | { type: 'KEEP_ORIGINAL_UNIT'; valuePx: number; parentDimensionPx: number | undefined }
```
Rheeseyb, line 25:
```ts
export function explicitCSSNumber()
export function keepOriginalUnit()
```
Please!

balazsbajorics: sigh ok

balazsbajorics: fixed in 1af442f7c

Rheeseyb (approved):

PR #3167, opened 2023-01-24
4/2/1 split chained number input controls for padding

### LINKED ISSUE #3168: Split controls (4/2/1) for chained number input

ruggi:
Fixes #3168 

This PR allows to enter chained number controls with a 4/2/1 split. Specifically, this is first used for the padding controls but a reusable component is introduced so we can extend this to other controls as well (border size, radius, margin, etc.).

There are 3 possible modes:
- a single value that applies to all sides
- two values that apply to the horizontal and vertical sides
- four distinct values, one per side (the existing implementation)

Kapture 2023-01-26 at 10 42 35

The default mode is loaded with this logic:
1. if all sides are unset, display 4 values, or the default mode override prop **(2 values for padding)**
2. if all sides have the same value, display 1 value
3. if the horizontal and vertical sides have the same value, display 2 values
4. otherwise, display 4 values


bkrmendy (approved):

maltenuhn:
Non-voting LGTM for the functionality from me :)

editor/src/core/shared/math-utils.ts
```diff
@@ -925,6 +925,16 @@ export function clampValue(value: number, minimum: number, maximum: number): num
   return Math.max(Math.min(value, maximum), minimum)
 }
 
+export function wrapValue(value: number, minimum: number, maximum: number): number {
+  if (value > maximum) {
+    return minimum
```
Rheeseyb, line 930: If this is wrapping, should this be doing something like `minimum + (value - maximum)` (with a little recursion sprinkled in)?

Rheeseyb (approved):

editor/src/core/shared/math-utils.ts
ruggi: true, it was a very quick implementation that does not count for the overflow value. will fix :)

ruggi: https://github.com/concrete-utopia/utopia/pull/3167/commits/16e855dd3122a67f29bbf0264ee3f34072ca654e

PR #3170, opened 2023-01-24
when setting Fixed size, prune all props that might interfere

bkrmendy:
## Problem
When setting an element to Fixed size, it it not enough to trim `min` and `max` props, but `flex*` also (since in a flex container these overrule `width`/`height`).

## Fix
When setting an element to Fixed size, prune the following props:
- minWidth/minHeight
- maxWidth/maxHeight
- flex
- flexBasis
- flexGrow
- flexShrink

The pruning takes the parent's `flex-direction` into account.


Rheeseyb (approved):

bkrmendy:
Great suggestion! `alignItems: stretch` is fortunately not a problem, since [accoring to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items) "Flex items are stretched such that the cross-size of the item's margin box is the same as the line **while respecting width and height constraints**".

bkrmendy:
> Does setting Fixed size behave as expected even if the flex parent sets `alignItems: stretch`, or do we also need to set a defensive `alignSelf` to override that?

bkrmendy:
already being implemented on a separate branch with a more principled solution

PR #3176, opened 2023-01-25
only store parentFlexDirection if the parent is display flex

balazsbajorics:
**Problem:**
If an element is display: block or otherwise not Flex, computedStyle.flexDirection will still default to `row`. 
In the specialSizeMeasurements we already store parentFlexDirection as FlexDirection | null. 

I'm working on a new feature that conditionally removes properties based on the parent being flex or not.

To simplify the logic, it'd be better if parentFlexDirection would be `null` for non-flex parents.

**Fix:**
Store parentFlexDirection as null if the parent is not flex


bkrmendy (approved):

balazsbajorics:
closing because the failing test reveals this would regress a new feature that is more important

PR #3177, opened 2023-01-25
Feature/fragment support

gbalint:
**Description:**
This PR is a first step towards having proper fragment support in utopia. The biggest change is that fragments have uids and element paths, so they are addressable, selectable, transformable, etc.

However, this change causes a lot of change in the features, notably:
- fragments are returned as parents/siblings/children of elements
- fragments can be selected, and then they can be a subject of transformations (which typically do not make sense for fragments, because most of our transformation manipulates style props, what fragments don't have)
- fragments have metadata

We tried to avoid regressions by
- introducing a feature switch and hide fragments from navigator when the feature switch is off
- preventing selection of fragments when the feature switch is off
- filtering out fragments from reparenting targets
- filtering out fragment from snapping targets

When the feature switch is on, we have a really incomplete implementation for selected fragments, but some basic preparations are still done so at least not everything throws an exception:
- `withUnderlyingTarget` and its variations accept a callback where the element is not `JSXElement`
- In all interaction/command code where we just assumed the target element is a JSXElement so it can have props, we switched to a more defensive code
- the navigator shows fragments and can be selected there
- fragment metadata contains globalFrame and localFrame and it is calculated from the frame of its children
- on the canvas fragments behave as first class citizens, e.g. the bounding box of them is shown on hover, you can select them, etc.

And we had to update a lot of tests, especially those which used fragments in their test code.


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -2964,7 +2967,7 @@ export function getValidElementPathsFromElement(
       ),
     )
 
-    const name = getJSXElementNameAsString(element.name)
+    const name = isJSXElement(element) ? getJSXElementNameAsString(element.name) : 'Fragment'
```
seanparsons, line 2970: This logic feels like it should probably be wrapped up in a function, but that can be done subsequently.

editor/src/components/canvas/commands/set-css-length-command.ts
```diff
@@ -81,7 +85,11 @@ export const runSetCssLengthProperty: CommandFunction<SetCssLengthProperty> = (
     editorState,
     left(`no target element was found at path ${EP.toString(command.target)}`),
     (_, element) => {
-      return getModifiableJSXAttributeAtPath(element.props, command.property)
+      if (isJSXElement(element)) {
```
seanparsons, line 88: This also feels like there's a case for a function that encapsulates this logic.

editor/src/components/canvas/canvas-utils.ts
Rheeseyb: It might be worth doing that now, because there's quite a significant implicit assumption being made here, since right now this is applying to an element that `isJSXElementLikeWithChildren`, which could easily be made explicit by converting this to a function.

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -386,7 +388,44 @@ describe('Select Mode Double Clicking', () => {
...
   })
 
-  it('Six double clicks will focus a generated Card and select its root element', async () => {
+  it('Five double clicks will focus a generated Card and select its root element', async () => {
+    // prettier-ignore
+    const desiredPath = EP.fromString(
+      'sb' +                 // Skipped as it's the storyboard
+      '/scene-CardList' +    // Skipped because we skip over Scenes
+      '/CardList-instance' + // <- First double click
+      ":38e" +               // <- Third double click skips fragment
```
Rheeseyb, line 397: Is it possible to create a hard coded uid for the fragments rather than relying on this generated one staying the same?

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -32,8 +33,20 @@ interface ComputedLook {
 
 export const BasePaddingUnit = 20
 
-export function getElementPadding(elementPath: ElementPath): number {
-  return EP.navigatorDepth(elementPath) * BasePaddingUnit
+export function getElementPadding(
+  elementPath: ElementPath,
+  visibleNavigatorTargets: Array<ElementPath>, // TODO: we only need this to filter out fragments, can be delete after 'Fragment support' FS is deleted
```
Rheeseyb, line 38: Can we make sure this is being tracked in the paper doc please

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
seanparsons: Ah yes, there's a test function for this now IIRC.

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1695,24 +1718,47 @@ function fillSpyOnlyMetadataWithFramesFromChildren(
...
+
+    childrenInDomCache[pathStr] = childrenAndUnfurledComponents
 
-    return children
+    return childrenAndUnfurledComponents
   }
 
   const elementsWithoutDomMetadata = Object.keys(fromSpy).filter((p) => fromDOM[p] == null)
+  elementsWithoutDomMetadata.sort()
+  elementsWithoutDomMetadata.reverse()
```
Rheeseyb, line 1761: Why the sort and reverse here? Please can you add a comment to this

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1450,10 +1452,8 @@ function parseElementProps(
 }
 
 type TSTextOrExpression = TS.JsxText | TS.JsxExpression
-type TSJSXElement = TS.JsxElement | TS.JsxSelfClosingElement
-type ElementsToParse = Array<
-  TSJSXElement | TSTextOrExpression | TS.JsxOpeningFragment | TS.JsxClosingFragment | TS.JsxFragment
->
+type TSJSXElement = TS.JsxElement | TS.JsxSelfClosingElement | TS.JsxFragment
```
Rheeseyb, line 1455: Interesting... Do we no longer need to cater for opening and closing fragments?

Rheeseyb (approved):
Nice work!

editor/src/components/navigator/navigator-item/navigator-item.tsx
gbalint: Ok, done!
image

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
gbalint: Yes, originally we  treated the opening and closing fragments as separate tokens (siblings of their children), but now we handle the full fragment with its content as a level in the hierarchy.

ruggi (approved):

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
seanparsons: Putting this in the backlog, as it's turning into a non-trivial set of changes.

editor/src/components/canvas/canvas-strategies/strategies/flex-reorder-strategy.spec.browser2.tsx
```diff
@@ -324,4 +418,255 @@ describe('Flex Reorder Strategy', () => {
       </div>`),
     )
   })
+
+  xdescribe('projects with fragments, with fragments support enabled', () => {
```
balazsbajorics, line 422: is this test disabled for a reason? or just leftover?

editor/src/core/model/element-metadata-utils.ts
seanparsons: Fixed in 810722e1d.

editor/src/components/canvas/canvas-strategies/strategies/flex-reorder-strategy.spec.browser2.tsx
seanparsons: This was done so that we don't have to get the fragments work fully functional (with the feature switch on) before merging it as these tests do not all pass currently.

editor/src/components/canvas/commands/set-css-length-command.ts
seanparsons: Added to backlog.

editor/src/components/canvas/canvas-utils.ts
seanparsons: Added to backlog.

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -645,8 +695,10 @@ describe('Select Mode Double Clicking With Fragments', () => {
     await doubleClick()
     await doubleClick()
     await doubleClick()
+    await doubleClick()
 
     expect(renderResult.getEditorState().editor.selectedViews).toEqual([desiredPath])
+    setFeatureEnabled('Fragment support', false)
```
balazsbajorics, line 701: these should store the original value for the 'Fragment support' feature switch, and restore that. other tests use `before`/`after` to achieve this.

see https://github.com/concrete-utopia/utopia/blob/6af6cdaea3803fbafcdaea16d726bcf2072328f6/editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.spec.browser2.tsx#L785-L792

balazsbajorics: btw we use this pattern in a bunch of places, it could be extracted into a single helper function that takes a feature switch name and the desired temporary value for the duration of the test.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -409,9 +401,13 @@ function trimWhitespaces(
   return trimmedText
 }
 
+export var Fragment: React.FunctionComponent<React.PropsWithChildren<any>> = ({ children }) => {
```
balazsbajorics, line 404: why not just 
`const Fragment = React.Fragment` ?

gbalint: Good idea, then we don't even have to create a local Fragment const, just use the one imported from React

gbalint: Fix in https://github.com/concrete-utopia/utopia/pull/3273

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1695,24 +1718,50 @@ function fillSpyOnlyMetadataWithFramesFromChildren(
...
 
-    return children
+    return childrenAndUnfurledComponents
   }
 
   const elementsWithoutDomMetadata = Object.keys(fromSpy).filter((p) => fromDOM[p] == null)
+  // Sort and then reverse these, so that lower level elements are handled ahead of their parents
+  // and ancestors. This means that if there are a grandparent and parent which both lack global frames
+  // then the parent is fixed ahead of the grandparent, which will be based on the parent.
+  elementsWithoutDomMetadata.sort()
```
balazsbajorics, line 1763: I _think_ if there is a UID in the dataset that is more than 3 letters long, the sort will not be correct anymore! (as far as I see this is sorting an array of stringified elementPaths, implicitly assuming that shorter string = higher level element)

balazsbajorics (approved):

PR #3178, opened 2023-01-25
Test utility to assert that a single UI action only generates a single undo step

bkrmendy:
## Description
This PR adds a test utility which enables us to assert that a given UI action only generates a single undo.

This implemented by adding a helper, `expectSingleUndoStep`, which checks the size of `EditorStorePatched.history.previous` before and after a given UI action, and expects the size difference to be 1 (meaning that 1 new undo frame was added).

### DIFF
--- a/editor/src/components/canvas/event-helpers.test-utils.tsx
+++ b/editor/src/components/canvas/event-helpers.test-utils.tsx
@@ -2,8 +2,7 @@ import { act, createEvent, fireEvent } from '@testing-library/react'
 import { emptyModifiers, Modifiers } from '../../utils/modifiers'
 import { resetMouseStatus } from '../mouse-move'
 import keycode from 'keycode'
-import { assertNever, NO_OP } from '../../core/shared/utils'
-import { EditorRenderResult } from './ui-jsx.test-utils'
+import { NO_OP } from '../../core/shared/utils'
 
 // TODO Should the mouse move and mouse up events actually be fired at the parent of the event source?
 // Or document.body?
--- a/editor/src/components/canvas/event-helpers.test-utils.tsx
+++ b/editor/src/components/canvas/event-helpers.test-utils.tsx
@@ -695,13 +695,3 @@ export function switchDragAndDropElementTargets(
     fireEvent(targetElement, makeDragEvent('dragover', targetElement, endPoint, fileList))
   })
 }
-
-export async function expectSingleUndoStep<T>(
-  editor: EditorRenderResult,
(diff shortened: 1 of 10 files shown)


editor/src/components/canvas/event-helpers.test-utils.tsx
```diff
@@ -694,3 +695,13 @@ export function switchDragAndDropElementTargets(
...
+
+export async function expectSingleUndoStep<T>(
+  editor: EditorRenderResult,
+  action: () => Promise<T>,
+): Promise<T> {
+  const undoCount = editor.getUndoCount()
+  const result = await action()
+  expect(editor.getUndoCount()).toEqual(undoCount + 1)
+  return result
+}
```
Rheeseyb, line 707: The function itself is fine, but this file is the wrong place for it. I _think_ `ui-jsx.test-utils.tsx` would be better 🤔 (unless you can find a better place)

bkrmendy: I moved it to `utils.test-utils.ts` in [e461839](https://github.com/concrete-utopia/utopia/pull/3178/commits/e461839dff9fa75dd25902d49ef5e623b60e7f1d)

Rheeseyb (approved):

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -199,6 +200,7 @@ export async function renderTestEditorWithModel(
   loginState: LoginState = notLoggedIn,
 ): Promise<EditorRenderResult> {
   const renderCountBaseline = renderCount
+  let undoCount = 0
```
balazsbajorics, line 203: seems to be leftover

editor/src/components/canvas/event-helpers.test-utils.tsx
```diff
@@ -3,6 +3,7 @@ import { emptyModifiers, Modifiers } from '../../utils/modifiers'
 import { resetMouseStatus } from '../mouse-move'
 import keycode from 'keycode'
 import { assertNever, NO_OP } from '../../core/shared/utils'
+import { EditorRenderResult } from './ui-jsx.test-utils'
```
balazsbajorics, line 6: leftover

balazsbajorics (approved):
very nice! easy to use API and easy to read code

gbalint (approved):

enidemi (approved):

PR #3182, opened 2023-01-26
Chore: smarter property path

balazsbajorics:
**Problem:**
I came across a problem: in our API I want to be able to say "property path where the last element is either width or height" or "property path where the last element is **not** one of these 6 string literals". 

**Fix:**
To be able to do this, the PropertyPath has to become a generic type which contains the _actual_ path elements. This allows  us to still make a _default_ property path type that matches with the current one, but also allows, if I need, to be more precise: every current place can use `function (myParameter: PropertyPath)` just like today, but I can now write `function newFunction(myParameter: PropertyPath<[string, 'width' | 'height']>)` and typescript will help me enforce new API contracts!

The only struggle was that our factory function `PP.create` takes an array of path elements. Which typescript types as an array instead of an enum. Instead of adding `as const` to 144 places, I changed the function signature from `PP.create(['style', 'left'])` to `PP.create('style', 'left')`. Places that need to pass in an array change from `PP.create(myArray)` to `PP.create(...myArray)`. This allows for excellent automatic type inference!

**Details:**
- New backwards-compatible PropertyPath signature
- PP.create loses the array-wrapping
- I regex fixed a few hundred callsites
- Profit!

### DIFF
--- a/editor/src/core/shared/property-path.ts
+++ b/editor/src/core/shared/property-path.ts
@@ -5,7 +5,7 @@ import { arrayEquals, fastForEach, longestCommonArray } from './utils'
 export function fromString(value: string): PropertyPath {
   let fromPathStringCache: PropertyPath | null = globalPathStringToPathCache[value]
   if (fromPathStringCache == null) {
-    const result = create(value.split('.'))
+    const result = create(...value.split('.'))
     globalPathStringToPathCache[value] = result
     return result
   } else {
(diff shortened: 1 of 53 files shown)


bkrmendy (approved):

editor/src/core/shared/property-path.ts
```diff
@@ -52,7 +52,20 @@ function getPathCache(elements: Array<PropertyPathPart>): PropertyPathCache {
   return workingPathCache
 }
 
-export function create(elements: Array<PropertyPathPart>): PropertyPath {
+export function create<T1 extends PropertyPathPart>(element1: T1): PropertyPath<[T1]>
```
seanparsons, line 55: Does it need to be generic?

balazsbajorics: yeah, this is what makes Typescript infer the narrowest type possible for the individual path elements

image

otherwise it would just infer this, which is not sufficient:
image

seanparsons (approved):

PR #3185, opened 2023-01-26
setProp -> setCssLength for 'left', 'top', 'right', 'bottom', 'width', 'height'

balazsbajorics:
**Problem:**
We have the somewhat overlapping setProperty and setCssLength commands. I wanted to make sure that for css length properties 'left', 'top', 'right', 'bottom', 'width', 'height' we use the setCssLength, and never setProperty.

**Fix:**
1. Type setProperty in a way to disallow PropertyPaths ending in 'left', 'top', 'right', 'bottom', 'width', 'height'.
2. Fix the problems by migrating them over to setCssLength
3. Profit


editor/src/utils/common-constants.ts
```diff
@@ -1 +1 @@
-export const styleStringInArray = ['style']
+export const styleStringInArray = ['style'] as ['style']
```
balazsbajorics, line 1: this looks a bit wonky but it makes sure that the type is the array literal with a string literal inside it. this means then typescript can pick up this type where we use it to build longer paths!

editor/src/components/inspector/inspector-strategies/fill-container-basic-strategy.tsx
```diff
@@ -29,21 +33,22 @@ export const fillContainerStrategyBasic = (
     }
 
     return elements.flatMap((path) => {
-      const parentInstance = MetadataUtils.findElementByElementPath(metadata, EP.parentPath(path))
-      if (!MetadataUtils.isFlexLayoutedContainer(parentInstance)) {
+      const instance = MetadataUtils.findElementByElementPath(metadata, path)
+      // TODO BEFORE MERGE!! this is an unrelated change I just fixed it
```
balazsbajorics, line 37: @bkrmendy I think this one is for you. I noticed that this place was using EP.parentPath to check something and I _believe_ I replaced it with something equivalent but less error-prone. could you double check please?

bkrmendy: the parent instance is needed to check whether it's a flex container, what you added is in fact a safer and better way of doing that so big +1 from me

editor/src/utils/common-constants.ts
bkrmendy: is this the same as suing `as const`?

balazsbajorics: `as const` makes it a `ReadonlyArray<'style'>` and quite a few places complained that they take Array and not ReadonlyArray, and I didn't want to inflate the scope of this PR

bkrmendy (approved):

Rheeseyb (approved):

PR #3189, opened 2023-01-27
parallelize karma tests

balazsbajorics:
**Problem:**
We have a _ton_ of new Karma tests, and we really like them. But they run in a single track in a single browser. They take upwards of 6 minutes locally! They take 9-10 minutes on the CI!!!

**Fix:**
Use [karma-parallel](https://github.com/joeljeske/karma-parallel) to split the tests and run them among multiple browser instances. For local dev envs, I set it to run CPU_CORES/2 number of browsers as our experience with Jest is that gives us the fastest execution while still allowing the machine to run everything else full-speed. For the CI, I am setting it to CPU_CORES, which in our case means... 2. I'm currently running the tests with `time` to see how fast it runs with two browsers, results in 15 minutes.

### DIFF
--- a/editor/karma.conf.js
+++ b/editor/karma.conf.js
@@ -1,5 +1,7 @@
 process.env.CHROME_BIN = require('puppeteer').executablePath() // Puppeteer v19.6.0 uses Chromium 110.0.5479.0
 
+const isGithubActionsEnvironment = process.env.CI === 'true'
+
 const os = require('os')
 const cpuCores = os.cpus().length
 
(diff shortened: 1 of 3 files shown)


editor/karma.conf.js
```diff
@@ -70,5 +77,21 @@ module.exports = function (config) {
         timeout: config.debug ? 1000000 : 10000,
       },
     },
+    parallelOptions: {
+      executors: isGithubActionsEnvironment ? cpuCores : cpuCores / 2,
+      shardStrategy: 'round-robin',
+      // shardStrategy: 'description-length'
```
enidemi, line 83: are these commented lines helpful for some future debugging?

enidemi (approved):

editor/karma.conf.js
balazsbajorics: good question! we can keep them as "user manual" if we don't like the default sharding strategy. or I can remove it and just link to the README.md

bkrmendy (approved):

PR #3202, opened 2023-01-30
Unset style props if possible when toggling them

gbalint:
**Problem:**
Toggling bold/italic/etc.... pollutes the code with `fontStyle`/`fontWeight` `normal` lines, which are in most cases unnecessary.

**Fix:**
The difficulty in this is that sometimes the explicit normal rule is necessary to override an inherited property.
But it is hard to know beforehand if that is necessary or not. The solution is to unset the property, check the updated style prop value in specialSizeMeasurements, and when it is incorrect dispatch another action to set it to te default value.

Missing:
- make it work outside of the text editor in select mode
- prevent double undo

### DIFF
--- a/editor/src/components/text-editor/text-editor-shortcut-helpers.ts
+++ b/editor/src/components/text-editor/text-editor-shortcut-helpers.ts
@@ -1,8 +1,14 @@
 import * as EditorActions from '../editor/actions/action-creators'
-import { ComputedStyle, emptyComments, jsxAttributeValue } from '../../core/shared/element-template'
+import {
+  ComputedStyle,
+  ElementInstanceMetadataMap,
+  emptyComments,
+  jsxAttributeValue,
+} from '../../core/shared/element-template'
 import { ElementPath } from '../../core/shared/project-file-types'
 import * as PP from '../../core/shared/property-path'
-import { EditorAction } from '../editor/action-types'
+import { EditorAction, EditorDispatch } from '../editor/action-types'
+import { MetadataUtils } from '../../core/model/element-metadata-utils'
 
 export function toggleTextBold(target: ElementPath, fontWeight: string | null): EditorAction {
   const toggledFontWeight = 'bold'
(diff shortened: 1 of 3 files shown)


editor/src/components/text-editor/text-editor-shortcut-helpers.ts
```diff
@@ -68,3 +190,36 @@ const toggleStyleProp = (
...
+    const computedValue = getComputedValue()
+    if (computedValue !== newValue) {
+      dispatch([
+        EditorActions.setProperty(
+          elementPath,
+          PP.create('style', prop),
+          jsxAttributeValue(newValue, emptyComments),
+        ),
+      ])
+    }
```
bkrmendy, line 215: can these two dispatches be combined into a single one? will this cause double undo?

editor/src/components/text-editor/text-editor.spec.browser2.tsx
```diff
@@ -781,8 +806,12 @@ async function pressShortcut(editor: EditorRenderResult, mod: Modifiers, key: st
...
-  const editor = await renderTestEditorWithCode(projectWithoutText, 'await-first-dom-report')
+async function testModifier(
+  mod: Modifiers,
+  key: string,
+  startingProject: string = projectWithoutText,
+) {
+  const editor = await renderTestEditorWithCode(startingProject, 'await-first-dom-report')
 
   await prepareTestModifierEditor(editor)
   await pressShortcut(editor, mod, key)
```
bkrmendy, line 817: would it be worth it o use `expectSingleUndoStep` here?

balazsbajorics: absolutely!!

editor/src/components/text-editor/text-editor-shortcut-helpers.ts
balazsbajorics: they cannot be combined, however, the second action dispatched should be wrapped in a `transientActions()` – until we have the proposed "appendToPreviousAction" behavior

editor/src/components/text-editor/text-editor.spec.browser2.tsx
gbalint: I'm happy to do it - in a separate PR after I fixed text editor undo

editor/src/components/text-editor/text-editor-shortcut-helpers.ts
gbalint: done

bkrmendy (approved):

balazsbajorics (approved):

PR #3204, opened 2023-01-30
paste image into flex container

bkrmendy:
## Problem
Images are always inserted as absolute positioned elements when pasted, which can be a frustrating surprise when the image is pasted into a flex container

## Fix
Check if the target parent is a flex container, and don't add any positioning props if it is.

## TODO
- [ ] tests


balazsbajorics (approved):
commit 0ea1d1004fbef80be7064db1cb5eabe16b33c016 looks good, the rest of the diff _seems like_ contains your other PR

bkrmendy:
> commit [0ea1d10](https://github.com/concrete-utopia/utopia/commit/0ea1d1004fbef80be7064db1cb5eabe16b33c016) looks good, the rest of the diff _seems like_ contains your other PR

thanks! 😬

PR #3205, opened 2023-01-30
paste image into flex container

bkrmendy:
Depends on https://github.com/concrete-utopia/utopia/pull/3214

## Problem
Images are always inserted as absolute positioned elements when pasted, which can be a frustrating surprise when the image is pasted into a flex container

## Fix
Check if the target parent is a flex container, and don't add any positioning props if it is.

## TODO
- [ ] tests


balazsbajorics (approved):

Rheeseyb (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -3490,16 +3490,26 @@ export const UPDATE_FNS = {
...
-                  left: relativeFrame.x,
-                  top: relativeFrame.y,
-                  width: relativeFrame.width,
-                  height: relativeFrame.height,
-                },
-                emptyComments,
-              ),
+              style: MetadataUtils.isFlexLayoutedContainer(
+                MetadataUtils.findElementByElementPath(editor.jsxMetadata, parent),
+              )
```
Rheeseyb, line 3495: We should probably extend this to check if we should be inserting with static positioning too (ideally using the same check as we do during reparenting)

PR #3206, opened 2023-01-30
Cmd + click to cycle the padding control backwards

bkrmendy:
## Description
The padding control can be cycled between the 4 / 2 / 1 segment setting by clicking the `[ - ]` icon next to the segments. This PR adds an improvement related to this, namely that when the icon is clicked while `cmd` is held, the segments are cycled in the reverse direction (so instead of going `1 -> 2 -> 4`, it goes `1 -> 4 -> 2`.


ruggi (approved):
> +1

balazsbajorics (approved):

PR #3209, opened 2023-01-30
Group flex children in flex container

bkrmendy:


editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -954,3 +975,55 @@ function addCreateHoverInteractionActionToSwitchModeAction(
...
+  }
+
+  if (element.specialSizeMeasurements.parentFlexDirection.startsWith('col')) {
+    return {
+      element: jsxElement('div', uid, props, []),
+      importsToAdd: {},
+    }
+  }
+
+  return 'default-empty-div'
```
Rheeseyb, line 1028: You've already checked that there is a `parentFlexDirection`, meaning one of these two must be true, so I'd say go with an `if else` and bin off this redundant fallback case

editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -954,3 +975,55 @@ function addCreateHoverInteractionActionToSwitchModeAction(
...
+    element == null ||
+    element.specialSizeMeasurements.parentFlexDirection == null ||
+    element.specialSizeMeasurements.parentLayoutSystem !== 'flex'
+  ) {
+    return 'default-empty-div'
+  }
+
+  const uid = makeUid()
+
+  // pragmatismus maximus, REVIEW: is there a better way to do this?
```
Rheeseyb, line 995: Not really. You could create the `gap` upfront as either a value or `undefined` and then shove it straight into the `style` inline. Other than that I'd say we've never really used `Record`, but rather stuck with standard objects for things like this where it doesn't really matter. I have no strong preference either way though.

Rheeseyb: Ok turns out I'm a liar and we have used `Record` in some places

Rheeseyb (approved):

editor/src/components/editor/global-shortcuts.tsx
balazsbajorics: > You could create the gap upfront as either a value or undefined and then shove it straight into the style inline.
sounds easy enough to avoid mutation!

also, I think the style object could by typed as 
`const style: CSSProperties = {` !

Rheeseyb: I totally forgot that that type exists! Good catch!

balazsbajorics (approved):

editor/src/components/editor/global-shortcuts.tsx
bkrmendy: thanks! I changed the type to `CSSProperties` in https://github.com/concrete-utopia/utopia/pull/3209/commits/a841cca84ff9c790ce1113f96f855652b345e36f

editor/src/components/editor/global-shortcuts.tsx
bkrmendy: it turns out that both ifs can be removed (since `element.specialSizeMeasurements.parentFlexDirection` is already incorporated into `props` a few lines above this snippet)

bkrmendy: implemented in https://github.com/concrete-utopia/utopia/pull/3209/commits/c2eb0c2ee18d7b8a7b95370b2daa4fe077c31625

PR #3210, opened 2023-01-30
adjust padding edges together

bkrmendy:
# Try it [here](https://utopia.pizza/p/39947443-celestial-recorder/?branch_name=feature/padding-adjust-together)

padd_improved

## Description
It is a common use case to adjust both paddings on a given cross-axis (i.e., both left and right or both top and bottom). It's also common to adjust all 4 paddings at once (to inset contents).

This PR adds support for these use cases in the following way:
- by default, only the padding edge under the cursor is adjusted.
- if `alt` is held, both paddings are adjusted along the affected cross-axis
- if `shift + alt` is held, all 4 padding edges are adjusted


editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
```diff
@@ -398,3 +411,49 @@ function indicatorPosition(
...
+      paddingRight: newPaddingValue,
+    }
+  }
+
+  if (adjustMode === 'cross-axis') {
+    newPadding[opposite(paddingPropInteractedWith)] = newPaddingValue
+    return newPadding
+  }
+
+  assertNever(adjustMode)
```
Rheeseyb, line 458: Can you change this to use a `switch` statement please, so that we can clearly see all cases laid out

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/set-padding-strategy.tsx
bkrmendy: refactored in https://github.com/concrete-utopia/utopia/pull/3210/commits/bee357129367cbbf95dd4ed87b940765c780d51a

ruggi (approved):

PR #3213, opened 2023-01-31
Fix Github overlay timing and commit+push flow

### LINKED ISSUE #3211: Fix Github commit+push operation and overlay

ruggi:
Fixes #3211

**Problem:**

1. The locked overlay not always stays in place until a GH operation is completed
2. The `Commit and push` (and `Push to branch`) buttons don't trigger the overlay correctly and are not disabled while the operation is pending

**Fix:**

This PR fixes the two issues above in a couple ways (the diff files count is huge but it's mostly due to import updates):
1. Wrap every Github operation in the `runGithubOperation` helper which makes sure the operation is added to and removed from `githubOperations` before/after the logic of the operation itself is completed
2. Get rid of the indirection of the `SAVE_TO_GITHUB` action which was messing with the editor state and effectively trimming the gh operations list; instead, call `saveProjecToGithub` directly.
3. Additionally, the gigantic `github.ts` file is split into multiple files, a `helper.ts` one and one file per operation under `github/operations` so to help out with navigating the code. This is definitely not the final form but it's a lot better already
4. Additionally, the `commish` action is renamed to `commit-and-push`. It's been fun but since the action is referenced everywhere as `commit and push` anyways, I figured it was due a renaming :)

Notes for the reviewers:
- the only important logic change is in `editor/src/components/navigator/left-pane/github-pane/index.tsx` which calls the operation instead of dispatching `saveToGithub`.
- all other github actions are not changed, aside from the wrapping with `runGithubOperation`


maltenuhn:
RIP commish, but it died for a worthy cause. Nice!

editor/src/components/navigator/left-pane/github-pane/index.tsx
```diff
@@ -610,14 +612,48 @@ const LocalChangesBlock = () => {
...
+        dispatch,
+        {
+          branchName: cleanedCommitBranchName,
+          commitMessage: commitMessage,
+          assetChecksums: assetChecksums ?? {},
+        },
       )
     }
-  }, [dispatch, repo, commitMessage, cleanedCommitBranchName])
+  }, [
```
bkrmendy, line 648: should we add a log here for debugging (which one of repo, cleanedCommitBranchName etc is null)

editor/src/core/shared/github/helpers.ts
```diff
@@ -178,370 +153,31 @@ export function repositoryEntry(
...
-      }
-    } else {
-      return [
-        showToast(
-          notice(`Github: Unexpected status returned from endpoint: ${response.status}`, 'ERROR'),
-        ),
-      ]
-    }
+    dispatch([updateGithubOperations(operation, 'add')], 'everyone')
+    await new Promise((res) => setTimeout(res, 1000))
```
bkrmendy, line 175: what are we really awaiting here?

ruggi: debug remnants I was sure I did not push 🙈

bkrmendy (approved):

editor/src/core/shared/github/helpers.ts
bkrmendy: oh :D

editor/src/components/navigator/left-pane/github-pane/index.tsx
ruggi: sounds like a good idea

ruggi: https://github.com/concrete-utopia/utopia/pull/3213/commits/1eb3b76d74180a620fe871c77967b05c89c4a733

seanparsons (approved):
Really nice cleanup!

editor/src/components/editor/store/editor-state.ts
```diff
@@ -262,8 +262,8 @@ export interface UserState extends UserConfiguration {
   githubState: GithubState
 }
 
-export interface GithubCommish {
-  name: 'commish'
+export interface GithubCommitAndPush {
+  name: 'commit-and-push'
```
Rheeseyb, line 266: 🙌 I'm never suggesting a joke name to you again!

ruggi: I am truly sorry 💔

Rheeseyb (approved):

PR #3214, opened 2023-01-31
Typed handling of infinitely sized frames

### LINKED ISSUE #3042: Infinity bugs sometimes when inserting onto the canvas
To repro, from outside of Utopia copy an image to the clipboard, and then paste into Utopia with nothing selected (so that the target parent will be the storyboard). Doing so will use infinity in the calculations used to insert the image.

The root cause of this was this PR https://github.com/concrete-utopia/utopia/pull/2765 which calls out the potential riskiness:

> Fix to 2: storyboard frame should cover the full infinite canvas. Maybe this solution is a little bit risky and controversial, but the tests are fine, and I did not find any bugs.

So it looks like we can’t use infinity here, but we also can’t use a zero sized frame in the storyboard’s metadata. Perhaps the solution here is to explicitly handle the storyboard frame...

Rheeseyb:
Fixes #3042 

**Problem:**
We give the storyboard element an infinitely sized frame so that we can target it for insertion and re-parenting. However, that leads to all sorts of problems later down the line when we try to calculate other frames relative to it, leading to other frames being updated with infinite dimensions or positions. This is easy to repro by copying an image from Finder and pasting it onto the canvas with no target parent selected.

**Fix:**
The core fix here was to introduce some new types, one called `InfinityRectangle`, and another `MaybeInfinityRectangle`. These are used in the element metadata when referring to a rectangle that may be infinitely sized, so that we can ensure we are correctly handling that case in all places where it might surface.

This is a bit ( 🙃 ) painful, since there are some places where we want an infinitely sized frame to behave the same as a `null` frame, and other places where we want it to actually behave like an infinitely sized one. We were previously implicitly (or even accidentally) handling that correctly, and of course some where we weren't, so now all of those places have been made explicit.

I've also added a test for pasting an image onto the canvas in `clipboard.spec.browser2.tsx`, and verified that it did indeed fail before making these changes.


bkrmendy (approved):

editor/src/core/shared/math-utils.ts
```diff
@@ -75,27 +75,48 @@ export function size(width: number, height: number): Size {
...
+export type MaybeInfinityCanvasRectangle = MaybeInfinityRectangle<CanvasModifier>
+export type MaybeInfinityLocalRectangle = MaybeInfinityRectangle<LocalModifier>
+
+export function isInfinityRectangle<C extends CoordinateMarker>(
+  value: MaybeInfinityRectangle<C>,
+): value is InfinityRectangle<C> {
+  return 'type' in value && value.type === 'INFINITY_RECTANGLE'
+}
+
+export function isNonInfinityRectangle<C extends CoordinateMarker>(
```
seanparsons, line 107:
```suggestion
export function isFiniteRectangle<C extends CoordinateMarker>(
```

Rheeseyb: Well yeah, sure, I guess in hindsight that probably makes way more sense 😁

PR #3215, opened 2023-01-31
The 'Resize to Fit' button toggles between fixed and hug

bkrmendy:
## Description

This PR adds an inverse action to Resize to fit, which converts the selected element(s) to Fixed size on both cross-axes. The main rationale for this is that in our experience, fixed size is basically the `position: absolute` of sizing, the setting that offers the most flexibility and the fewest number of implicit constraints, so it would be worth it to provide a quick way to get there.

### Heuristic behaviour
When neither cross-axis is set to fixed, the Resize to fit button converts to fixed size. Otherwise, it sets sizing along both cross axes to Hug contents.


editor/src/components/inspector/inspector-common.ts
```diff
@@ -557,3 +557,27 @@ export function addPositionAbsoluteTopLeft(
     setProperty('always', elementPath, styleP('position'), 'absolute'),
   ]
 }
+
+export function notFixedSizeOnEitherAxis(
```
Rheeseyb, line 561: Minor, but please can you rename this to `notFixedSizeOnBothAxes`. The current name implies that we're checking `horizontalState !== 'fixed' || verticalState !== 'fixed'`

editor/src/components/inspector/resize-to-fit-control.spec.browser2.tsx
```diff
@@ -49,6 +49,44 @@ describe('Resize to fit control', () => {
...
+    expect(view.style.minWidth).toEqual('')
+    expect(view.style.maxWidth).toEqual('')
+    expect(view.style.height).toEqual('154px')
+    expect(view.style.minHeight).toEqual('')
+    expect(view.style.maxHeight).toEqual('')
+    expect(view.style.flex).toEqual('')
+    expect(view.style.flexShrink).toEqual('')
+    expect(view.style.flexGrow).toEqual('')
+    expect(view.style.flexBasis).toEqual('')
+  })
```
Rheeseyb, line 89: Can you add a test for the case where one axis is set to hug and the other set to fixed please

Rheeseyb (approved):

balazsbajorics (approved):

editor/src/components/inspector/inspector-common.ts
bkrmendy: Absolutely! fixed in https://github.com/concrete-utopia/utopia/pull/3215/commits/8a22e1289b1d3054f2c8d3f847c0e50f1a5ca6ca

editor/src/components/inspector/resize-to-fit-control.spec.browser2.tsx
bkrmendy: added in https://github.com/concrete-utopia/utopia/pull/3215/commits/07724d722dcb0054f281e807e757008095ca61a2

PR #3217, opened 2023-01-31
Smart flex conversion, pass 1

balazsbajorics:
When converting an element to flex, we shouldn't just append a display: flex property, instead we should guess padding, gap, alignItems etc based on the visual/measured positions of the children, so the resulting conversion tries to approximate the original layout visually.

**Details:**
- New `convertLayoutToFlexCommands` function that uses heuristics to come up with flex direction, gap, padding. Used by the 9-box "Layout System plus button" and the shift-A shortcut
- `setPropertyOmitNullProp` helper to avoid inserting default values
- tests

**What it cannot do yet:**
- align-items and justify-content detection
- changing the flex parent size to hug, but it doesn't yet use the defensive code to avoid collapsing
- reorder children to actually match their visual layout to avoid children jumping around on the canvas.
- probably missing a ton of defensive code against edge cases I did not consider


bkrmendy (approved):

editor/src/components/common/shared-strategies/convert-to-flex-strategy.spec.browser2.tsx
```diff
@@ -0,0 +1,290 @@
...
+    makeTestProjectCodeWithSnippet(`
+    <div style={{ ...props.style }} data-uid='a'>
+      <div
+        style={{ backgroundColor: '#aaaaaa33', position: 'absolute', left: ${parentL}, top: ${parentT}, width: ${parentW}, height: ${parentH}  }}
+        data-uid='parent'
+      >
+        ${input.children
+          .map((child, i) => {
+            const [childL, childT, childW, childH] = child
+            return `<div data-uid='child-${i}' style={{ backgroundColor: '#aaaaaa33', position: 'absolute', left: ${childL}, top: ${childT}, width: ${childW}, height: ${childH} }} />`
```
Rheeseyb, line 245: Linebreaks 😩

editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
```diff
@@ -0,0 +1,192 @@
...
+  } else {
+    const paddingTop = firstChild.frame.y - parentRect.y
+    const paddingBottom =
+      parentRect.y + parentRect.height - (lastChild?.frame.y + lastChild?.frame.height)
+    const verticalPadding = Math.max(0, Math.min(paddingTop, paddingBottom))
+    if (verticalPadding === 0) {
+      return null
+    }
+    return `${verticalPadding}px 0`
+  }
```
Rheeseyb, line 132: I'd argue we don't need to handle these separately, and should instead include both horizontal and vertical padding

Rheeseyb (approved):

editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
balazsbajorics: hah, I'll try

editor/src/components/common/shared-strategies/convert-to-flex-strategy.spec.browser2.tsx
balazsbajorics: fixed

PR #3222, opened 2023-02-01
Better handling of Github errors

### LINKED ISSUE #3221: Improve handling of Github errors

ruggi:
Fixes #3221

**Problem:**

When Github API calls return errors, it's hard to understand what went wrong and why.

**Fix:**

This PR does two things:
1. consolidate the way Github operation errors are handled
2. handle the errors happening during the `runGithubOperation` calls, so that the errors are logged in a more structured way to the console
3. make the error toast a bit more informative and consistent about which operation failed, as well as pointing to the console for more info

Screenshot 2023-02-01 at 5 12 39 PM

The error format is as follows:
- for the toasts: `<name of the operation> failed. See the console for more information`
- for the console: `[GitHub] operation "<name>" failed: <JSON object>`

The `JSON object` is:
```typescript
{
  operation: GithubOperation // the operation which failed
  status: string // a short description of the error (e.g. status text + status code)
  data?: string // a string containing more failure data (e.g. the body of the error response)
}
```

As a followup to this PR we should also touch the server logic which performs the GH API calls so that the actual error messages coming from GH are not shadowed and are returned (at least in their meaningful parts) to the client.

### DIFF
--- a/editor/src/core/shared/github/operations/update-against-branch.ts
+++ b/editor/src/core/shared/github/operations/update-against-branch.ts
@@ -8,10 +8,12 @@ import {
   updateGithubChecksums,
   updateGithubData,
 } from '../../../../components/editor/actions/action-creators'
-import { GithubRepo } from '../../../../components/editor/store/editor-state'
+import { GithubOperation, GithubRepo } from '../../../../components/editor/store/editor-state'
 import {
   getBranchContentFromServer,
   GetBranchContentResponse,
+  githubAPIError,
+  githubAPIErrorFromResponse,
   runGithubOperation,
 } from '../helpers'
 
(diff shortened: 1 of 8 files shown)


bkrmendy (approved):

balazsbajorics (approved):

editor/src/core/shared/github/operations/update-against-branch.ts
```diff
@@ -21,79 +23,67 @@ export async function updateProjectAgainstGithub(
...
+        null,
+      )
+
+      const branchLatestResponse = await branchLatestRequest
+      const specificCommitResponse = await specificCommitRequest
 
-    const branchLatestResponse = await branchLatestRequest
-    const specificCommitResponse = await specificCommitRequest
+      if (!branchLatestResponse.ok) {
+        throw await githubAPIErrorFromResponse(operation, branchLatestResponse)
```
seanparsons, line 42: `throw await` looks so weird, but it works!

seanparsons (approved):

PR #3223, opened 2023-02-01
Extended undo API: merge with previous undo step

gbalint:
**Problem:**
It is not possible to dispatch an action without making a new undo step. But sometimes it is not possible, or really uncomfortable to collect all the actions and dispatch them together.

One example is text format toggling: when you toggle back a bold text to normal, we don't want to include `fontWeight: normal` in the style props if it is not necessary. But we can only decide whether it is necessary or not if we actually try to unset and check the format of the text. If it is not normal, we have to put `fontWeight: normal` back to style props, but we should avoid creating two undo steps.

**Fix:**
Created a new a wrapper action called `MergeWithPrevUndo`. When the actions are dispatched in this wrapper, the last undo snapshot will be replaced with the result of these actions.

Added mergeWithPrevUndo into the text format toggling logic, and extended the tests to check for the single undo action.

### DIFF
--- a/editor/src/components/text-editor/text-editor.spec.browser2.tsx
+++ b/editor/src/components/text-editor/text-editor.spec.browser2.tsx
@@ -1,4 +1,4 @@
-import { wait } from '../../utils/utils.test-utils'
+import { expectSingleUndoStep, wait } from '../../utils/utils.test-utils'
 import { altCmdModifier, cmdModifier, Modifiers, shiftCmdModifier } from '../../utils/modifiers'
 import { CanvasControlsContainerID } from '../canvas/controls/new-canvas-controls'
 import {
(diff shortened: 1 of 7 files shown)


bkrmendy (approved):

editor/src/components/text-editor/text-editor.spec.browser2.tsx
```diff
@@ -857,10 +857,12 @@ async function prepareTestModifierEditor(editor: EditorRenderResult) {
 }
 
 async function pressShortcut(editor: EditorRenderResult, mod: Modifiers, key: string) {
-  pressKey(key, {
-    modifiers: mod,
-    targetElement: document.getElementById(TextEditorSpanId) ?? undefined,
-  })
+  await expectSingleUndoStep(editor, async () =>
```
balazsbajorics, line 860: yay!

balazsbajorics (approved):

PR #3227, opened 2023-02-02
Make text format global shortcuts work well with property unset

gbalint:
**Problem:**
This is a followup to https://github.com/concrete-utopia/utopia/pull/3202 : the same feature is implemented for select mode too.

**Fix:**
The obstacle to do this was that the toggle helpers (e.f. `toggleTextBoldWithUnset`) only accept a single target element path, and they dispatch their own actions themselves. In select mode we have to support multiselection too, and I wanted to avoid creating separate undo steps for each selected element.

So my first plan was to refactor the toggle helpers to support multiple elements. Which means we need some bookkeeping, because we also have to check whether we can unset the prop or not one by one for each element, and collect all the actions to be dispatched for that too.

However, I realized there is a much simpler solution: I just use the new undo api from https://github.com/concrete-utopia/utopia/pull/3223 , call the toggle helpers separately for each element, and take care of not generating extra undo step by wrapping all actions into `MergeWithPrevUndo` actions, except the first one.

The downside of this solution that it does not fix the "juggling multiselect toggling" bug, but that was not the goal now, and we can always come back and improve this solution.

### DIFF
--- a/editor/src/components/editor/global-shortcuts.tsx
+++ b/editor/src/components/editor/global-shortcuts.tsx
@@ -104,7 +104,11 @@ import {
   boundingArea,
   createHoverInteractionViaMouse,
 } from '../canvas/canvas-strategies/interaction-state'
-import { emptyComments, jsxAttributeValue } from '../../core/shared/element-template'
+import {
+  ElementInstanceMetadataMap,
+  emptyComments,
+  jsxAttributeValue,
+} from '../../core/shared/element-template'
 import {
   toggleTextBold,
   toggleTextItalic,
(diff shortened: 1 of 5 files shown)


editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -733,42 +737,72 @@ export function handleKeyDown(
...
       [TOGGLE_TEXT_BOLD]: () => {
-        return isSelectMode(editor.mode)
-          ? editor.selectedViews.map((target) => {
-              const element = MetadataUtils.findElementByElementPath(editor.jsxMetadata, target)
-              return toggleTextBold(target, element?.specialSizeMeasurements.fontWeight ?? null)
-            })
-          : []
+        if (isSelectMode(editor.mode)) {
+          let first = true
+          editor.selectedViews.forEach((target) => {
```
Rheeseyb, line 742: You could get rid of the `let first` here by using `.forEach((target, index) => {`

balazsbajorics (approved):

Rheeseyb (approved):

editor/src/components/editor/global-shortcuts.tsx
gbalint: good idea, done

balazsbajorics (approved):

PR #3156, opened 2023-01-23
Reopen the code editor setting it at least 100px wide

ruggi:
Fixes #3155 

**Problem:**

When the code editor is closed, reopening it is almost impossible due to its width being just a couple pixels.

**Fix:**

When reopening the code editor, if the width is less than 100px, max it to 100px.

`.pizza` vs fix:
Kapture 2023-01-23 at 11 42 54

PR #3229, opened 2023-02-02
Fix/code pane toggle min width

ruggi:
This PR adds the `MIN_CODE_PANE_REOPEN_WIDTH` safeguard check to the `TOGGLE_PANEL` action as well, like we did for https://github.com/concrete-utopia/utopia/pull/3156


bkrmendy (approved):

maltenuhn (approved):
Feels correct in use too, including starting from various code editor sizes and both resizing-to-close and toggling the editor on.

balazsbajorics (approved):

PR #3235, opened 2023-02-02
Trigger canvas controls via the inspector padding controls

Rheeseyb:
**Problem:**
We want to make it extra clear what you are changing on the canvas when hovering or focusing the padding controls in the Inspector

**Fix:**
By using a pair of `jotai` atoms we can specify which canvas controls to render when a given inspector field is either hovered or focused. This is tied to the `mouseenter`, `mouseleave`, `focus` and `blur` events of the inspector control (in this case the `SplitChainedNumberInput`), as we can rely on the ordering of the firing of those events to set the array of controls directly in both cases. The atoms are then read in `new-canvas-controls.tsx` and used to render the controls with the provided props (similarly to how we do with controls tied to canvas strategies).

padding-inspector-canvas-controls

### DIFF
--- a/editor/package.json
+++ b/editor/package.json
@@ -142,7 +142,7 @@
     "immutability-helper": "3.1.1",
     "istextorbinary": "5.11.0",
     "jest-fetch-mock": "3.0.1",
-    "jotai": "^1.4.2",
+    "jotai": "2.0.0",
     "json5": "0.5.1",
     "jszip": "3.5.0",
     "keycode": "2.2.1",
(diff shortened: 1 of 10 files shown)


maltenuhn:
I'm already so curious to try this!!

bkrmendy (approved):

balazsbajorics (approved):

enidemi (approved):

ruggi (approved):

PR #3241, opened 2023-02-03
Fix flicker when exiting text editor

gbalint:
**Problem:**
Text flickers back to original state for a frame when exiting the text editor.

**Fix:**
The reason is twofold:
1) We save the text when the text editor unmounts
2) We do the save in a setTimeout (since this: https://github.com/concrete-utopia/utopia/pull/3166 )

I brought the save dispatch to an earlier phase with two methods:
1) I save text in the `onBlur` handler - that happens earlier than the unmount, because we go to select mode in the blur handler, and then the text editor unmounts, because in select mode it is not rendered
2) I use `requestAnimationFrame` instead of `setTimeout` in unmount

I thought about completely removing the save on unmount, but that sounds a bit unsafe. If the editor goes to select mode from a different source, it could happen that the blur handler is never called (the blur handler is not called when a focused element is unmounted without losing focus).

This solution could result in a double save, which we should avoid (we don't want to have 2 undo steps). So I added an extra check before save whether there is real change in the text content. This is advantageous anyway, because it avoids adding an unnecessary undo step when you leave the text editor with unmodified content. Because of this change I needed to update one test to contain an editing operation.

### DIFF
--- a/editor/src/components/text-editor/text-editor.spec.browser2.tsx
+++ b/editor/src/components/text-editor/text-editor.spec.browser2.tsx
@@ -430,6 +430,8 @@ describe('Use the text editor', () => {
 
         await wait(50) // give it time to adjust the caret position
 
+        typeText('l')
+
         await closeTextEditor()
         await editor.getDispatchFollowUpActionsFinished()
 
@@ -446,7 +448,7 @@ describe('Use the text editor', () => {
               data-testid='first-div'
               data-uid='first-div'
             >
-              Hello this
+              Helllo this
             </div>
             <div
               style={{ backgroundColor: 'red' }}
--- a/editor/src/components/text-editor/text-editor.spec.browser2.tsx
+++ b/editor/src/components/text-editor/text-editor.spec.browser2.tsx
@@ -22,7 +22,7 @@ describe('Use the text editor', () => {
 
     await enterTextEditMode(editor)
     typeText(' Utopia')
-    await closeTextEditor()
+    await expectSingleUndoStep(editor, async () => closeTextEditor())
     await editor.getDispatchFollowUpActionsFinished()
 
     expect(editor.getEditorState().editor.mode.type).toEqual('select')
@@ -55,7 +55,7 @@ describe('Use the text editor', () => {
 
     await enterTextEditMode(editor)
     typeText('Utopia')
-    await closeTextEditor()
+    await expectSingleUndoStep(editor, async () => closeTextEditor())
     await editor.getDispatchFollowUpActionsFinished()
 
     expect(editor.getEditorState().editor.mode.type).toEqual('select')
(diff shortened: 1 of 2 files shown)


editor/src/components/text-editor/text-editor.spec.browser2.tsx
```diff
@@ -446,7 +448,7 @@ describe('Use the text editor', () => {
               data-testid='first-div'
               data-uid='first-div'
             >
-              Hello this
+              Helllo this
```
balazsbajorics, line 451: does this test break on current master?

editor/src/components/text-editor/text-editor.spec.browser2.tsx
```diff
@@ -430,6 +430,8 @@ describe('Use the text editor', () => {
 
         await wait(50) // give it time to adjust the caret position
 
+        typeText('l')
+
         await closeTextEditor()
```
bkrmendy, line 435: can we wrap `await closeTextEditor()` in `expectSingleUndoStep`?

bkrmendy (approved):

editor/src/components/text-editor/text-editor.tsx
```diff
@@ -337,8 +343,17 @@ const TextEditor = React.memo((props: TextEditorProps) => {
   )
 
   const onBlur = React.useCallback(() => {
-    dispatch([updateEditorMode(EditorModes.selectMode())])
-  }, [dispatch])
+    const content = myElement.current?.textContent
+    if (content != null && elementState != null && savedContentRef.current !== content) {
+      savedContentRef.current = content
+      dispatch([
+        updateChildText(elementPath, escapeHTML(content)),
```
balazsbajorics, line 350: I think this updateChildText(elementPath, escapeHTML(content) is worth pulling into a function – just in case we need to dispatch it from a third place, to avoid someone not copy-pasting it

editor/src/components/text-editor/text-editor.spec.browser2.tsx
gbalint: No, but it breaks with my modification, because I don't save when there is no change to the text, that is why I had to type something

editor/src/components/text-editor/text-editor.tsx
gbalint: done

editor/src/components/text-editor/text-editor.spec.browser2.tsx
gbalint: I wrapped closeTextEditor-s in expectSingleUndoStep in multiple text editing tests in f48683ce471dad92b2ab71d0e52bda57a9bf5ef2 . Unfortunately, collapsing similar divs still seems to generate more undo steps, but that is not really connected to this PR (text editor still saves only once), so I would not solve it here.

balazsbajorics (approved):

PR #3243, opened 2023-02-03
Ensure border width is nonzero when enabled in the inspector

bkrmendy:
## Problem
When enabling `border` in the inspector, it is possible that a border of `borderWidth: 0` is applied. While technically this is still a valid border, this goes against the principle of least surprise (since nothing happens on screen).

## Fix
Check if the border width to be applied is not null and nonzero, and if it isn't, apply a `borderWidth` of `1px`


ruggi (approved):

maltenuhn (approved):
Ok, I feel adequately informed to approve this :) LGTM!

PR #3247, opened 2023-02-06
Strategy for changing flex direction in the presence of children set to Fill Container

bkrmendy:
## Description
This PR adds an inspector strategy for changing flex direction on a flex container. This strategy kicks in when the selected container has at least one child element set to "Fill container" on one axis and "Fixed size" on the other. These settings are swapped along the cross-axes so that they are the other way around (e.g., if the child was Fixed Size horizontally and Fill Container vertically, after the strategy as run, the child will be set to Fixed Size vertically and Fill Container horizontally).

### Details
- The new strategy is added
- The FIll Container strategy is taken apart into two separate strategies: one that applies to flex containers (and uses `flexGrow: 1` to implement Fill Container), and a catchall one (which uses `width/height: 100%` to implement Fill Container). This makes it easier to reuse the either of the specific variants of the Fill Container strategies.

### DIFF
--- a/editor/src/components/inspector/inspector-strategies/change-flex-direction-swap-axes.ts
+++ b/editor/src/components/inspector/inspector-strategies/change-flex-direction-swap-axes.ts
@@ -1,0 +1,91 @@
+import * as PP from '../../../core/shared/property-path'
+import { MetadataUtils } from '../../../core/model/element-metadata-utils'
+import { elementPath } from '../../../core/shared/element-path'
+import { ElementInstanceMetadataMap } from '../../../core/shared/element-template'
+import { ElementPath } from '../../../core/shared/project-file-types'
+import { CanvasCommand } from '../../canvas/commands/commands'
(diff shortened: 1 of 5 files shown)


editor/src/components/inspector/inspector-strategies/change-flex-direction-swap-axes.ts
```diff
@@ -0,0 +1,90 @@
...
+      return [
+        ...(fixedSizeBasicStrategy('always', 'vertical', horizontalSizing.value).strategy(
+          metadata,
+          [child],
+        ) ?? []),
+        ...(fillContainerStrategyFlexParent('horizontal', 'default', {
+          forceFlexDirectionForParent: direction,
+        }).strategy(metadata, [child]) ?? []),
+      ]
+    }
```
Rheeseyb, line 67: You could roll this into one to prevent any copy paste errors (or changes made to one branch but not the other) occurring later down the line

Rheeseyb (approved):

PR #3249, opened 2023-02-06
Smart Flex Conversion fix: minimum padding is 0

balazsbajorics:
**Problem:**
When converting an element to flex and its children overflow, we may end up setting a negative padding value on the parent.

**Fix:**
Make the minimum padding size 0, also a new test

### DIFF
--- a/editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
+++ b/editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
@@ -154,7 +154,7 @@ function guessPadding(
   const paddingLeft = firstChild.frame.x - parentRect.x
   const paddingRight =
     parentRect.x + parentRect.width - (lastChild?.frame.x + lastChild?.frame.width)
-  const horizontalPadding = Math.min(paddingLeft, paddingRight)
+  const horizontalPadding = Math.max(0, Math.min(paddingLeft, paddingRight))
   const paddingTop = firstChild.frame.y - parentRect.y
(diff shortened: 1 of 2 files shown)


editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
```diff
@@ -154,7 +154,7 @@ function guessPadding(
   const paddingLeft = firstChild.frame.x - parentRect.x
   const paddingRight =
     parentRect.x + parentRect.width - (lastChild?.frame.x + lastChild?.frame.width)
-  const horizontalPadding = Math.min(paddingLeft, paddingRight)
+  const horizontalPadding = Math.max(0, Math.min(paddingLeft, paddingRight))
```
bkrmendy, line 157: I think `clamp` from `math-utils` would be more readable here

bkrmendy (approved):

editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
balazsbajorics: fair, fair

bkrmendy: retracted, the original intention was to pick the smaller of `paddingLeft` and `paddingRight`, and ensure that the resulting value is larger or equal to 0, clamp is obviously not applicable here

ruggi (approved):

PR #3256, opened 2023-02-07
More feedback during navigator reparent

bkrmendy:
# Try it [here](https://utopia.pizza/p/3551bda9-before-i-go-basic/?branch_name=feature/better-navigator-reparent)
(in the Before I Go project where there's a lot of stuff to reparent)

## Description
This PR enhances the navigator reparent UX in the following ways:
1. The preview of the dragged item is hidden, so it's easier to see where the drop target line is
2 When reparenting in the navigator, if the mouse is moved sideways by a given delta (20px), the dragged element is reparented to a level higher in the hierarchy.
3. The would-be parent of the dragged element is highlighted during reparenting. 
4. Up to now, each element in the navigator had a drop target line above the element and a drop target line below the element. This led to a situation where by slightly moving the mouse, drag could alternate between "after the first" and "before the second" element, which is the same in practical terms (this also showed up as a visual glitch). This PR rewamps drop lines in a way so that
    - in a given level, only the bottom drop line is rendered for all elements but the first child (the first child gets a top drop line as well)
    - The bottom and top drop lines are positioned above each other, so only one drop line can be hovered at a time
    - Separate drop handlers are created for each of the top and bottom drop lines, and the reparent area.
    The advantage of this approach is that it gets rid of the fragile offset-based calculation that is present on master as of now (at the expense of more code).

### Implementation details
- A new field was added to `DropTargetHint`, `displayAtElementPath` since when reparenting to a grandparent (2. above), the drop target line is not actually displayed at the target element
- Up to now, the "reparent to grandparent" was activated when the dragged element was dragged to the left of another element in the navigator. This regressed, and this PR refactors it so that "reparent to grandparent" activates when the dragged element is moved to the left by a certain delta (similar to how Figma does it).


maltenuhn:
Awesome!!!

editor/src/components/navigator/navigator.spec.browser2.tsx
```diff
@@ -9,14 +9,10 @@ import { offsetPoint, windowPoint, WindowPoint } from '../../core/shared/math-ut
...
 import { BakedInStoryboardVariableName, BakedInStoryboardUID } from '../../core/model/scene-utils'
-import { CSSCursor } from '../canvas/canvas-types'
 import { getDomRectCenter } from '../../core/shared/dom-utils'
 import { selectComponents } from '../editor/actions/action-creators'
 import * as EP from '../../core/shared/element-path'
-
-interface CheckCursor {
-  cursor: CSSCursor | null
-}
+import { wait } from '../../core/model/performance-scripts'
```
ruggi, line 15: unused import

editor/src/components/navigator/navigator.tsx
```diff
@@ -41,14 +41,6 @@ const Item = React.memo(({ index, style }: ItemProps) => {
       dragSelections: dragSelections,
     }
   })
-  const getDistanceFromAncestorWhereImTheLastLeaf = React.useCallback(
-    (componentId: string, distance: number): number => {
-      // TODO FIXME HOLY SHIT THIS IS STUCK IN OLDE WORLDE
```
Rheeseyb, line 46: FIXED!

editor/src/core/shared/element-path.ts
```diff
@@ -653,6 +653,13 @@ function dropFromPath(path: ElementPath, n: number): ElementPath {
 
 export const getNthParent = dropFromPath
 
+export function dropNPathParts(path: ElementPath, n: number): ElementPath {
```
Rheeseyb, line 656: I think this is doing the same as `dropFromPath` above

Rheeseyb (approved):

editor/src/components/navigator/navigator.spec.browser2.tsx
bkrmendy: removed in https://github.com/concrete-utopia/utopia/pull/3256/commits/3b61a73535dec9b44a4223aaf271fa3775d9deba

ruggi (approved):

PR #3257, opened 2023-02-07
Fix Karma capturing async errors.

seanparsons:
**Problem:**
Our browser testing library Karma has an unfixed bug which means that unhandled exceptions from promises just get ejected into the abyss and are never logged: https://github.com/karma-runner/karma/issues/3798

**Fix:**
By adding an `unhandledrejection` exception handler we end up pulling those async errors somehow back into the tracking that Karma already does.

In testing the above change it was discovered that only way to ensure the handler would fire before the test completed was to make the triggering event be called itself in a Promise. That bled through a lot of the testing code, given how much they've been reused and also to ensure that the React testing library would not fire warnings of overlapping `act()` calls.

**Commit Details:**
- Added an `unhandledrejection` handler to Karma, so that unhandled exceptions
  from promises will fail a test.
- Added a bunch of `await`'s to some calls which need to be asynchronous.


balazsbajorics (approved):

editor/src/components/text-editor/text-editor.spec.browser2.tsx
```diff
@@ -293,6 +293,7 @@ describe('Use the text editor', () => {
 
       await pressShortcut(editor, cmdModifier, 'b')
 
+      await wait(1000)
```
Rheeseyb, line 296: Please can we remove this or replace it with a much shorter timeout?

seanparsons: Fixed in bec60f538.

Rheeseyb (approved):

enidemi (approved):

PR #3261, opened 2023-02-07
jotai useAtomsDevtools for the editor

balazsbajorics:
image

**Problem:**
If we ever end up in a situation where an Atom-controlled behavior needs debugging, it'd be handy to see the values in the Redux Devtool.

This PR adds a useAtomsDevtools to the editor root. it only runs when process.env.NODE_ENV !== 'production'

**Not nice things:**
- The atoms are anonymous by default, and the "actions" are just an incrementing index. So the debug experience is inferior to classic redux which has named actions.

**Nice things:**
- The Time Traveling Debugger works for these! should we ever use it...
- It's in a separate store so it's not making the store debugging noisier. 
- it can be improved upon by wrapping individual atoms in debug hooks


editor/package.json
```diff
@@ -313,6 +313,7 @@
     "jest-environment-jsdom": "^27.2.2",
     "jest-matcher-deep-close-to": "2.0.1",
     "jest-transform-stub": "2.0.0",
+    "jotai-devtools": "^0.2.0",
```
Rheeseyb, line 316:
```suggestion
    "jotai-devtools": "0.2.0",
```

Rheeseyb (approved):

gbalint (approved):

seanparsons (approved):

editor/package.json

PR #3269, opened 2023-02-10
Double click on resize corner to resize to fit

bkrmendy:
## Description
This PR upgrades the resize corner controls so that when they are double clicked on, the component they are rendered on resizes to fit its contents (if this behaviour is applicable).

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/absolute-resize-control.tsx
+++ b/editor/src/components/canvas/controls/select-mode/absolute-resize-control.tsx
@@ -16,10 +16,16 @@ import { Modifier } from '../../../../utils/modifiers'
 import { when } from '../../../../utils/react-conditionals'
 import { useColorTheme } from '../../../../uuiui'
 import { EditorDispatch } from '../../../editor/action-types'
+import { applyCommandsAction } from '../../../editor/actions/action-creators'
 import { useDispatch } from '../../../editor/store/dispatch-context'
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/controls/select-mode/absolute-resize-control.tsx
```diff
@@ -200,6 +225,7 @@ const ResizePoint = React.memo(
           }}
         />
         <div
+          onClick={onEdgeDblClick}
```
balazsbajorics, line 228: I think we should always use `onDoubleClick` when listening to double click events.

listening to click plus event.detail===2 may trigger under this _wrong_ circumstance:
1. click anywhere in the browser, on the canvas etc. for exmaple: click on the element to select it!
2. then click on the resize control within the doubleclick timeout.  (happens surprisingly more frequently than you might assume. especially if someone is in the flow, and knows what they are doing)

Detail will be 2, you will dispatch the dbclick behavior.

balazsbajorics (approved):

editor/src/components/canvas/controls/select-mode/absolute-resize-control.tsx
bkrmendy: great idea! implemented in https://github.com/concrete-utopia/utopia/pull/3269/commits/e4731a5865e9fa45de452eac7a7609ca7dc8aad6

enidemi (approved):

PR #3270, opened 2023-02-10
fill/fixed/hug control does not always show the right options

bkrmendy:
## Repro
- [Before project](https://utopia.pizza/p/bd9748f3-good-block-(forked))
- [After project](https://utopia.pizza/p/bd9748f3-good-block-(forked)/?branch_name=fix/fill-fixed-hug-options)

### Steps to repro:
1. select the upper div (note that the element inside is `position: absolute`)
2. check the options in either of the fill/fixed/hug dropdowns (only Fixed is available)
3. select the lower div
4. check the options in either of the fill/fixed/hug dropdowns again (only Fixed is available, but Hug should be available as well! This is the bug)

## Problem
The available options in the dropdown menus of the fill/fixed/hug control are obtained via `useRefEditorState`, which means that when another element is selected, and the available options change, the fill/fixed/hug control is not re-rendered, and so the new options don't show up in the dropdown.

## Solution
`optionsRef` was refactored into value (instead of a ref), and was wrapped into `createSelector`.


editor/src/components/inspector/fill-hug-fixed-control.tsx
```diff
@@ -104,20 +91,38 @@ function elementComputedDimension(
 
 interface FillHugFixedControlProps {}
 
+function fixedFillHugModeForElement(
```
balazsbajorics, line 94:
```suggestion
function getFixedFillHugOptionsForElement(
```

editor/src/components/inspector/fill-hug-fixed-control.tsx
```diff
@@ -298,7 +303,7 @@ export const FillHugFixedControl = React.memo<FillHugFixedControlProps>((props)
     >
       <PopupList
         value={optionalMap(selectOption, widthCurrentValue?.type) ?? undefined}
-        options={optionsRef.current}
+        options={options}
         onSubmitValue={onSubmitWidth}
         controlStyles={controlStylesRef.current}
```
balazsbajorics, line 308: while you are here, I think this is also wrong!

editor/src/components/inspector/fill-hug-fixed-control.tsx
```diff
@@ -279,7 +284,7 @@ export const FillHugFixedControl = React.memo<FillHugFixedControlProps>((props)
 
   const controlStylesRef = React.useRef(getControlStyles('simple'))
```
balazsbajorics, line 285: actually, this is all sorts of wrong!

balazsbajorics: If I read this correctly, and this is just a constant simple control style, then we should make it a simple `const` at the file root

bkrmendy: fixed in https://github.com/concrete-utopia/utopia/pull/3270/commits/6ed5ddb5b563de88bbacd03af1996e5d02ceed4f

balazsbajorics (approved):

editor/src/components/inspector/fill-hug-fixed-control.tsx
bkrmendy: fixed in [6ed5ddb](https://github.com/concrete-utopia/utopia/pull/3270/commits/6ed5ddb5b563de88bbacd03af1996e5d02ceed4f)

editor/src/core/shared/math-utils.ts
```diff
@@ -1109,3 +1109,12 @@ export function resize(
...
 }
+
+export function intersection<T>(sets: Array<Set<T>>): Set<T> {
+  const [first, ...rest] = sets
+  if (first == null) {
+    return new Set()
+  }
+
+  return rest.reduce((acc, v) => new Set([...acc].filter((x) => v.has(x))), first)
+}
```
Rheeseyb, line 1120: We have a `set-utils.ts` that would be better suited for this function

Rheeseyb (approved):

editor/src/core/shared/math-utils.ts
bkrmendy: moved it there! (the function in https://github.com/concrete-utopia/utopia/pull/3270/commits/1d3501c17e198f945b8bc53a0ce541efab703bf9, the tests in https://github.com/concrete-utopia/utopia/pull/3270/commits/035a7ac98b157c4b1a4a8e4d1140719e9a8d05f9)

PR #3274, opened 2023-02-13
Padding/Border radius controls update the selected elements

bkrmendy:
## Problem
Due to refs leaking into JSX markup, the padding/border radius controls didn't update the right element after selection has been changed, and the control wasn't re-rendered.

### Repro
- open https://utopia.pizza/p/644e2d16-cute-phosphorus/
- select the div on the left
- select the div on the right
- update padding (or border radius)
- the div on the left is updated instead of the one on the right

## Solution
Factor out the problematic refs and use `React.useMemo` to memoize curried callback functions.

### DIFF
--- a/editor/src/components/inspector/sections/layout-section/layout-system-subsection/layout-system-controls.tsx
+++ b/editor/src/components/inspector/sections/layout-section/layout-system-subsection/layout-system-controls.tsx
@@ -298,7 +298,7 @@ export const PaddingControl = React.memo(() => {
           B: 'paddingBottom',
           L: 'paddingLeft',
         },
-        selectedViewsRef.current[0],
+        selectedViewsRef,
         dispatch,
       )}
     />
--- a/editor/src/components/inspector/sections/layout-section/layout-system-subsection/layout-system-controls.tsx
+++ b/editor/src/components/inspector/sections/layout-section/layout-system-subsection/layout-system-controls.tsx
@@ -18,7 +18,6 @@ import { EdgePieces } from '../../../../canvas/padding-utils'
 import { InspectorContextMenuWrapper } from '../../../../context-menu-wrapper'
 import { switchLayoutSystem } from '../../../../editor/actions/action-creators'
 import { useDispatch } from '../../../../editor/store/dispatch-context'
-import { Substores, useEditorState } from '../../../../editor/store/store-hook'
 import { optionalAddOnUnsetValues } from '../../../common/context-menu-items'
 import {
   ControlStatus,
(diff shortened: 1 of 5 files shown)


balazsbajorics (approved):

editor/src/components/inspector/sections/layout-section/layout-system-subsection/layout-system-controls.tsx
```diff
@@ -273,6 +273,28 @@ export const PaddingControl = React.memo(() => {
...
+    () =>
+      longhandShorthandEventHandler(
+        'padding',
+        {
+          T: 'paddingTop',
+          R: 'paddingRight',
+          B: 'paddingBottom',
+          L: 'paddingLeft',
+        },
+        selectedViews[0],
```
balazsbajorics, line 292: if you make longhandShorthandEventHandler take a selectedViewRef you can restore using the selectedViewsRef here! – since you removed the line `selectedViews={selectedViewsRef.current}` anyways

editor/src/components/inspector/sections/style-section/container-subsection/radius-row.tsx
```diff
@@ -120,6 +117,28 @@ export const BorderRadiusControl = React.memo(() => {
...
+    () =>
+      longhandShorthandEventHandler(
+        'borderRadius',
+        {
+          T: 'borderTopLeftRadius',
+          R: 'borderTopRightRadius',
+          B: 'borderBottomRightRadius',
+          L: 'borderBottomLeftRadius',
+        },
+        selectedViews[0],
```
balazsbajorics, line 136: same here! this could now be turned into a ref again and passed in as-is

editor/src/components/inspector/sections/layout-section/layout-system-subsection/split-chained-number-input.tsx
```diff
@@ -320,6 +320,12 @@ export const SplitChainedNumberInput = React.memo((props: SplitChainedNumberInpu
 
   const isCmdPressedRef = useRefEditorState((store) => store.editor.keysPressed.cmd === true)
 
+  const selectedViews = useEditorState(
+    Substores.selectedViews,
+    selectedViewsSelector,
+    'PaddingControl selectedViews',
+  )
+
```
balazsbajorics, line 328: hmm this is a tougher nut to crack – as far as I understand, if selectedViews change, you want to call `updateMode()` and `setMode(null)` – without a major refactor, i think the most efficient way to do this is to use a `useSelectorWithCallback` – it's the same as a useEditorState, but instead of forcing a re-render, it merely executes a callback function.

pseudo code:
```typescript
 useSelectorWithCallback(
    Substores.selectedViews,
    selectedViewsSelector,
    () => {
      updateMode();
      setMode(null);
    },
    'PaddingControl selectedViews',
  )
```

bkrmendy: During discussions we decided against implementing this suggestion in this PR

ruggi (approved):

PR #3275, opened 2023-02-14
Feature switch setter helper for tests

gbalint:
**Problem:**
We repeat the same boilerplate everywhere to set feature switches to a value for tests (and setting it back after them)

**Fix:**
Create a test helper and use that instead of repeating code.


bkrmendy (approved):

balazsbajorics (approved):
nice!

PR #3276, opened 2023-02-14
Make `x` play nice with spans

bkrmendy:
## Problem(s)
The `x` shortcut either removes absolute positioning from and element, or sizes it to its visual dimension (its `clientWidth` / `clientHeight`), and adds `position: absolute` with a `top` and `left` offset.

This however break down in the case of `<span />`s for the following reasons:
- when converting from `position: absolute` to participating in the layout, spans collapse to their smallest possible size, since inline elements disregard their width/height set from CSS
- when converting to `position: absolute`, we cannot detemine the `clientWidth` / `clientHeight`of inline elements, because they are considered to be 0 by the browser [1]

### Repro
- create a text element by pressing `t`, drawing a text box and typing in some text
- select the new text element
- press `x`, observe the text box collapsing to hug the text inside
- press `x` again, observe the text box collapsing to a zero-sized element

#### `defaultSpanElement`
Our `defaultSpanElement` has `display: 'inline-block'` set. This way, all out helpers that expect elements to behave as block elements will work as expected.

#### Special casing
When converting from `x` special cases inline elements in the following way:
- when converting from `position: absolute` to participating in the layout, inline elements are sized to an arbitrary size and have `display: 'inline-block'` set, so they can be dragged around the canvas.
- when converting to `position: absolute`, if an inline element isn't set to block display, the same conversion as above is performed. If an inline element already has block sizing set (for example, as a result of converting it via `x`), it is simply sized to its visual dimensions.

[1] https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth

### DIFF
--- a/editor/src/components/editor/defaults.ts
+++ b/editor/src/components/editor/defaults.ts
@@ -158,6 +158,7 @@ export function defaultSpanElement(uid: string): JSXElement {
       style: jsxAttributeValue(
         {
           position: 'absolute',
+          display: 'inline-block',
           wordBreak: 'break-word',
         },
         emptyComments,
--- a/editor/src/components/editor/defaults.ts
+++ b/editor/src/components/editor/defaults.ts
@@ -158,7 +158,6 @@ export function defaultSpanElement(uid: string): JSXElement {
       style: jsxAttributeValue(
         {
           position: 'absolute',
-          display: 'inline-block',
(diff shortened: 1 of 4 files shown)


Rheeseyb (approved):

ruggi (approved):
\o\ \o\

PR #3279, opened 2023-02-14
delete element values

### LINKED ISSUE #3278: Remove element style props when deleting split control values

ruggi:
Fixes #3278 

**Problem:**

Deleting values from the 4/2/1 split controls does not remove the related props from controlled element.

**Fix:**

This PR updates the event handler so that when a value is deleted the element is updated consequently.
Tests are added for all possible combinations - in a nutshell:
- if a shorthand is used, if the shorthand is not entirely made of `0`s, a `0` is set for the deleted value; otherwise, the entire shorthand is deleted
- if a longhand is used, it's removed on delete

### DIFF
--- a/editor/src/components/inspector/sections/layout-section/layout-system-subsection/split-chained-number-input.tsx
+++ b/editor/src/components/inspector/sections/layout-section/layout-system-subsection/split-chained-number-input.tsx
@@ -21,6 +21,7 @@ import {
   CSSNumber,
   CSSNumberType,
   isCSSNumber,
+  isEmptyInputValue,
   printCSSNumber,
   UnknownOrEmptyInput,
 } from '../../../common/css-utils'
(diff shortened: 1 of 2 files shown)


bkrmendy (approved):

editor/src/components/inspector/sections/layout-section/layout-system-subsection/split-chained-number-input.tsx
```diff
@@ -214,6 +222,70 @@ const handleSplitChainedEvent =
...
+        case 'one-value':
+          return e.value == null
+        case 'two-value':
+          const otherDirections = useShorthand
+            ? e.value.type === 'V'
+              ? [aggregates.horizontal]
+              : [aggregates.vertical]
+            : e.value.type === 'V'
+            ? [aggregates.top, aggregates.bottom]
+            : [aggregates.right, aggregates.left]
```
Rheeseyb, line 247: This nested ternary threw me off 😬 But do we even need the nesting here? Wouldn't `[aggregates.top, aggregates.bottom]` both be zero if `[aggregates.vertical]` is zero?

editor/src/components/inspector/sections/layout-section/layout-system-subsection/split-chained-number-input.tsx
```diff
@@ -214,6 +222,70 @@ const handleSplitChainedEvent =
...
+      switch (e.type) {
+        case 'one-value':
+          return e.value == null
+        case 'two-value':
+          const otherDirections = useShorthand
+            ? e.value.type === 'V'
+              ? [aggregates.horizontal]
+              : [aggregates.vertical]
+            : e.value.type === 'V'
+            ? [aggregates.top, aggregates.bottom]
```
Rheeseyb, line 246: I believe this part of the ternary has been incorrectly flipped. If `e.value.type === 'V'` then the other directions would be `[aggregates.right, aggregates.left]`

ruggi: nope, this is correct - so that it deletes values on the same direction (probably worth adding a comment)

Rheeseyb: In that case, why is the above flipped?
```
e.value.type === 'V'
  ? [aggregates.horizontal]
  : [aggregates.vertical]
```

ruggi: this branch doesn't exist anymore as per https://github.com/concrete-utopia/utopia/pull/3279/commits/5893cd4fcdded5c0e457019b68c9e2ed2324e27f

editor/src/components/inspector/sections/layout-section/layout-system-subsection/split-chained-number-input.tsx
```diff
@@ -214,6 +222,95 @@ const handleSplitChainedEvent =
...
+          let shorthandOtherDirections: CSSNumberOrNull[] = []
+          if (useShorthand) {
+            if (e.value.type === 'V') {
+              shorthandOtherDirections.push(aggregates.horizontal)
+            } else {
+              shorthandOtherDirections.push(aggregates.vertical)
+            }
+          }
+
+          return !useShorthand || shorthandOtherDirections.every((o) => o === null || o.value === 0)
```
Rheeseyb, line 255: You can bin off the use of an array here completely
```suggestion
          if (useShorthand) {
            if (e.value.type === 'V') {
              return aggregates.horizontal === null || aggregates.horizontal.value === 0
            } else {
              return aggregates.vertical === null || aggregates.vertical.value === 0
            }
          } else {
            return true
          }
```

ruggi: oh nice :)

Rheeseyb (approved):

PR #3281, opened 2023-02-14
Use the correctly ordered children when determining if a layout is 1D

Rheeseyb:
**Problem:**
The Reorder canvas strategy relies on the existing children order to determine whether or not the current layout is in a single dimension, but was using the order returned by calling `Object.keys` on the metadata, meaning a duplication action could trip the logic.

**Fix:**
Use the correctly ordered children as part of the calculation. As a follow up to this I'm going to rename some of the existing functions for fetching children or siblings to reflect the fact that we can't rely on their order


bkrmendy (approved):

enidemi (approved):

editor/src/components/canvas/canvas-strategies/strategies/reorder-slider-strategy.spec.browser2.tsx
```diff
@@ -613,6 +617,26 @@ describe('Reorder Slider Strategy Control', () => {
 
     await selectAndHover('utopia-storyboard-uid/scene-aaa/app-entity:container/ccc', renderResult)
 
+    const targetControl = renderResult.renderedDOM.queryByTestId('reorder-slider-control')
+    expect(targetControl).toBeNull()
+  })
+  it('the reorder control is not visible on simple, 1d flow layouts for metadata with out of order children', async () => {
```
balazsbajorics, line 623: > for metadata with out of order children'

I don't understand, are these children out of order? not in the code and DOM right? only in the metadata, right?

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/reorder-slider-strategy.spec.browser2.tsx
Rheeseyb: Sorry this is a good point, I'll clarify the naming of the test. It's that the order returned by calling `Object.keys` on the metadata doesn't match the code or DOM order.

PR #3283, opened 2023-02-14
Support generated UID mocking in the parser.

seanparsons:
**Problem:**
When writing tests that covered fragments, we'd previously have to rely on the generated UIDs for shorthand fragments which is very brittle to minor changes.

While fixing the above issue, it was discovered that the handling of the collation of arbitrary blocks would cause the UIDs to be regenerated, as they would be identified as a duplicate from the previous parse done of the parts individually.

**Fix:**
The previously written functionality for mocking the generation of UIDs was adapted for inclusion during a parse.

For the arbitrary blocks issue, an imperfect mitigation was made against it by using the UIDs collected from before the arbitrary blocks were parsed originally. Which at least for the tests as most things have fixed UIDs works for those. However, in regular use there's still the possibility for a UID clash which (for future functionality) could result in a difference in UIDs between the individual arbitrary elements and the collated block.

**Commit Details:**
- Moved the `MOCK_NEXT_GENERATED_UIDS` prefixed variables into `uid-utils.ts`.
- Logic for generating the mock UIDs is now centralised into `generateMockNextGeneratedUID`.
- `generateConsistentUID` and `generateUID` now invoke `generateMockNextGeneratedUID`.
- Fixed `parseCode` to not regenerate UIDs for root level arbitrary JS blocks.
- Added some `data-uid` properties to some test data so that those would not consume the mock UIDs added to some tests.
- Fixed some tests that were for fragments testing which weren't using a project with fragments in them.


ruggi (approved):

editor/src/core/shared/uid-utils.ts
```diff
@@ -29,6 +29,21 @@ import { UTOPIA_PATH_KEY } from '../model/utopia-constants'
...
+export const MOCK_NEXT_GENERATED_UIDS: { current: Array<string> } = { current: [] }
+export const MOCK_NEXT_GENERATED_UIDS_IDX = { current: 0 }
+
+export function generateMockNextGeneratedUID(): string | null {
+  if (
+    MOCK_NEXT_GENERATED_UIDS.current.length > 0 &&
+    MOCK_NEXT_GENERATED_UIDS_IDX.current < MOCK_NEXT_GENERATED_UIDS.current.length
+  ) {
+    MOCK_NEXT_GENERATED_UIDS_IDX.current += 1
+    return MOCK_NEXT_GENERATED_UIDS.current[MOCK_NEXT_GENERATED_UIDS_IDX.current - 1]
```
ruggi, line 41: small nit: I think this would be more readable as e.g. this

```ts
const nextID = MOCK_NEXT_GENERATED_UIDS.current[MOCK_NEXT_GENERATED_ID.current]
MOCK_NEXT_GENERATED_ID.current += 1
return nextID
```

(of course there's also the option of `++`, but imho it'd not be an improvement as far as readability is concernet)

seanparsons: I just transplanted the original code with no changes.

ruggi: oh yeah for sure, just noting the nit :)

Rheeseyb (approved):

editor/src/core/shared/uid-utils.ts
seanparsons: Fixed in c8df6e8c8.

PR #3290, opened 2023-02-16
Feature/conditional support

gbalint:
**Conditionals:**
This PR introduces conditional expressions to the utopia model.
It is very similar to the earlier fragment PR: https://github.com/concrete-utopia/utopia/pull/3177

Conditionals have uid and elementpath, so they are addressable and selectable.

This update causes a lot of change in the features, notably:
- Conditionals are returned as parents/siblings/children of elements
- Conditionals can be selected, and then they can be a subject of transformations (which typically do not make sense for fragments, because most of our transformation manipulates style props, what fragments don't have)
- Conditionals are not found by the spy or the dom walker, so we need to construct metadata for them relying on their active branches. As of today we only update their globalFrame and localFrame from them.

We tried to avoid regressions by
- introducing a feature switch and hide conditionals from navigator when the feature switch is off
- preventing selection of conditionals when the feature switch is off

When the feature switch is on, conditionals are selectable, but there can be all kinds of bugs and errors if you try to work with them. These will be fixed in subsequent PRs.


editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1847,15 +1929,17 @@ export function parseOutJSXElements(
...
         })
 
-        // empty element fix: if the element contains only fragments, remove any empty text blocks
+        // empty element fix: if the element contains only fragments or conditionals, remove any empty text blocks
         // that may have been introduced in between due to the code formatting
         if (isRight(children)) {
           const nonEmptyTextBlockChildren = children.value.filter(
             (c) => !(isJSXTextBlock(c) && c.text.trim().length === 0),
           )
           const onlyFragments =
```
ruggi, line 1938: this should then become `shouldRemoveEmptyTextBlocks` or something like that, so we also catch future cases

ruggi (approved):
imo this is looking good 👌

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
gbalint: true, fixed in ef257d9da3d439074e222a80c7a3fdfad2180a74

bkrmendy (approved):

editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
```diff
@@ -64,7 +65,10 @@ const labelSelector = createSelector(
   (store: MetadataSubstate) => store.editor.allElementProps,
   (elementMetadata, allElementProps) => {
     if (elementMetadata == null) {
-      return 'Element 👻'
+      if (!isFeatureEnabled('Conditional support')) {
+        return 'Element 👻'
+      }
+      return 'Conditional'
```
Rheeseyb, line 71: Do we know it's a conditional at this point? Don't we need to actually check that?

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1764,11 +1812,45 @@ export function parseOutJSXElements(
...
-            const possibleExpression = produceArbitraryBlockFromJsxExpression(elem)
-            if (isLeft(possibleExpression)) {
-              return possibleExpression
+            let parseResult: Either<string, SuccessfullyParsedElement> =
+              left('Expression fallback.')
+            // Handle ternaries.
+            if (elem.expression != null && TS.isConditionalExpression(elem.expression)) {
+              const childrenOfExpression = elem.getChildren(sourceFile)
+              // Trailing comments of the entire expression appear to be attached to the
+              // closing brace of the expression.
```
Rheeseyb, line 1821: Yikes!

Rheeseyb (approved):

editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
ruggi: This was put in place before having conditionals metadata - but now that we have that, this is still the case (for now?) for nested conditionals AFAICT

gbalint: Fixed in 17e7060c229026586bfbcea6304e4408b77a5f97

PR #3305, opened 2023-02-20
Fix navigator item click and hover

bkrmendy:
## Problem
In the navigator, the hidden drop target lines intercept mouse events. This regressed clicking on a navigator item to select it, and hovering to see the lock and hide buttons. NB, this only happens when the click/hover happens over where the drop target line is, when clicking on the center of the navigator item, the corresponding element is still selected.

## Solution
On the drop target lines, set `pointer-events` to `none` when there's no drag in progress.

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item.tsx
@@ -21,6 +21,9 @@ import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import { ThemeObject } from '../../../uuiui/styles/theme/theme-helpers'
 import { isFeatureEnabled } from '../../../utils/feature-switches'
 
+export const NavigatorItemTestId = (pathString: string): string =>
+  `NavigatorItemTestId-${pathString}`
+
 interface ComputedLook {
   style: React.CSSProperties
   iconColor: IcnProps['color']
@@ -389,6 +392,7 @@ export const NavigatorItem: React.FunctionComponent<
       }}
     >
       <FlexRow
+        data-testid={NavigatorItemTestId(EP.toString(props.elementPath))}
         style={rowStyle}
         onMouseDown={select}
         onMouseMove={highlight}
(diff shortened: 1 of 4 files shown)


editor/src/components/navigator/navigator.spec.browser2.tsx
```diff
@@ -939,372 +945,477 @@ export var ${BakedInStoryboardVariableName} = (
...
-      y: notDraggableDivElementRect.y + notDraggableDivElementRect.height - 3,
-    }
-
-    const dragDelta = windowPoint({
-      x: dragTo.x - dragMeElementCenter.x,
-      y: dragTo.y - dragMeElementCenter.y,
-    })
+      await mouseClickAtPoint(dragMeElement, {
+        x: dragMeElementRect.x + dragMeElementRect.width / 2,
+        y: dragMeElementRect.y + 1,
```
balazsbajorics, line 994: isn't this the top and the other test the bottom?

balazsbajorics (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -21,6 +21,9 @@ import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import { ThemeObject } from '../../../uuiui/styles/theme/theme-helpers'
 import { isFeatureEnabled } from '../../../utils/feature-switches'
 
+export const NavigatorItemTestId = (pathString: string): string =>
+  `NavigatorItemTestId-${pathString}`
```
ruggi, line 25: this could have been the trailing part of an `ElementPath` so that it mimics the `uid` and `testid` props with a prefix

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -21,6 +21,9 @@ import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import { ThemeObject } from '../../../uuiui/styles/theme/theme-helpers'
 import { isFeatureEnabled } from '../../../utils/feature-switches'
 
+export const NavigatorItemTestId = (pathString: string): string =>
```
ruggi, line 24: why not just using the other one?

https://github.com/concrete-utopia/utopia/blob/35a60ea87ef8826598a365eb6a0ca62eb3dee85c/editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx#L42

If there's a specific reason not to, can we rename this? the double export for two things that are veeeeery similar is a bit clunky imho

ruggi (approved):
lgtm, left just a couple of observations

PR #3316, opened 2023-02-22
Toggle conditionals

### LINKED ISSUE #3321: Toggle conditionals in the inspector

ruggi:
Fixes #3321 

This PR adds an inspector section to override the rendered branch of a conditional.

Kapture 2023-02-22 at 17 17 45

Notes:
- only single conditionals are supported, on purpose (I don't think it'd make much sense to toggle multiple conditionals at the same time).
- the icons used for the conditional elements (`arc`) are pure placeholders, I just picked something that was not being used elsewhere
- the section only appears if the FS is enabled
- the conditional overrides are controlled via "flag comments" in the form `@utopia/[flag]=[value]`. For now they're placed in the trailing portion of the comments, because the leading portion is not being parsed for conditionals (yet).

### DIFF
--- a/editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
+++ b/editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
@@ -55,18 +55,6 @@ async function getControlValue(controlTestId: string, renderedDOM: RenderResult)
   return control.value
 }
 
-function elementDoesNotExist(renderResult: EditorRenderResult, testId: string) {
-  try {
-    const trueBranch = renderResult.renderedDOM.getByTestId(testId)
-    if (trueBranch != null) {
-      return false
-    }
-  } catch {
-    // nothing to do
-  }
-  return true
-}
-
 async function setControlValue(
   controlTestId: string,
   newValue: string,
@@ -2205,7 +2193,7 @@ describe('inspector tests with real metadata', () => {
         await renderResult.getDispatchFollowUpActionsFinished()
 
         expect(renderResult.renderedDOM.getByTestId('ccc')).not.toBeNull()
-        expect(elementDoesNotExist(renderResult, 'bbb')).toBe(true)
+        expect(renderResult.renderedDOM.queryByTestId('bbb')).toBeNull()
 
         expect(getPrintedUiJsCode(renderResult.getEditorState())).toEqual(
           makeTestProjectCodeWithSnippet(`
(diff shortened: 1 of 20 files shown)


bkrmendy (approved):

editor/src/core/shared/comment-flags.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+    return null
+  }
+
+  const comment = commentString(c)
+  const prefix = utopiaFlagKey(type) + '='
+
+  if (comment.startsWith(prefix)) {
+    const value = comment
+      .replace(new RegExp(`^${prefix}`), '')
```
seanparsons, line 55: I think this will do the same right?
```suggestion
      .slice(prefix.length)
```

(only suggesting it because regexes are quite slow)

seanparsons (approved):

editor/src/core/shared/comment-flags.ts
ruggi: oh that's a very nice alternative

balazsbajorics:
> only single conditionals are supported, on purpose (I don't think it'd make much sense to toggle multiple conditionals at the same time).

I'm not sure about this statement! – especially with nested conditionals, I _think_ I'd want to be able to "dive in" through multiple conditionals to reach a UI element.

that being said, I don't mind if we start from here with this PR, but this should be not a hard technical constraint IMO

gbalint (approved):

editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
```diff
@@ -54,6 +55,18 @@ async function getControlValue(controlTestId: string, renderedDOM: RenderResult)
   return control.value
 }
 
+function elementDoesNotExist(renderResult: EditorRenderResult, testId: string) {
```
balazsbajorics, line 58: https://testing-library.com/docs/queries/about

there's a cheat sheet table on this page, there's queryBy and queryAllBy which don't throw error on missing elements

ruggi: 😱

ruggi: https://github.com/concrete-utopia/utopia/pull/3316/commits/6e4f7a44736a3038158a0779f52335c1d79278fc

balazsbajorics (approved):

PR #3322, opened 2023-02-22
Don't show the size label for fragments and conditionals

bkrmendy:
## Repro
Select the fragment in the navigator in the project:
```

const foo = true

export var storyboard = (
  <div>
    <>
      
    </>
  </div>
)
```

## Description
We show the absolute resize control for fragments, which leads to the editor crashing.

## Fix
The `!`s in the size label on the absolute resize control crash the editor. The direct fix for this is to do the null check before generating the label text and render the label with the generated text.

The root cause of this problem was that `MetadataUtils.targetHonoursPropsSize` assumed that any element that doesn't have a corresponding underlying component, as determined by `findUnderlyingTargetComponentImplementationFromImportInfo`, so the absolute resize control was shown. This PR adds a check for JSX fragments and conditionals and classifies them as elements that don't honor size props.


ruggi (commented):
can we add a test for this? :)

balazsbajorics (approved):

ruggi (approved):

PR #3327, opened 2023-02-23
Handle nested ternaries.

seanparsons:
**Problem:**
Currently if ternaries are nested, the parsing fails and defaults back down to treating the code as an arbitrary block.

**Fix:**
The primary change here is to slightly tweak the parser to handle conditionals which are not wrapped with a JSX expression. When one is a child of a JSX element, that will be the case, but when it is a child of another conditional or is the root element of a component there will not be a wrapping JSX expression around it.

Otherwise there was a fix to how spy metadata is backfilled for conditionals, as it did not handle the case that a child of a conditional could be another conditional and assumed that there would be an element which it could find in the metadata.

**Commit Details:**
- `fillSpyOnlyMetadata` now looks up the children for a spy element from the combined metadata of the spy and missing filled in metadata.
- `clearJSXElementUniqueIDs` now supports conditional expressions.
- The JSX parser supports parsing a conditional expression both within a JSX expression and without one surrounding it, which is necessary to capture child conditionals of another.
- Added `jsxElementToJSXExpression` as a wrapper around `jsxElementToExpression` so that conditionals get wrapped in a JSX expression which is necessary for the case where a conditional is a child of another JSX element.
- Fixed some tests as these changes made more code parseable than was previously possible.

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer-conditionals.test-utils.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer-conditionals.test-utils.ts
@@ -19,3 +19,27 @@ export var ${BakedInStoryboardVariableName} = (
   </Storyboard>
 )
 `
+
+export const NestedTernariesExample = `
+import * as React from 'react'
+import { Scene, Storyboard, View } from 'utopia-api'
+export var App = (props) => {
+  return <div data-uid={'div'}>
+    {[0, 1].length > 1 ? (
+      [0, 1].length === 0 ? (
+        <div data-uid='middle'/>
+      ) : null
+    ) : null}
+  </div>
+}
+export var ${BakedInStoryboardVariableName} = (
+  <Storyboard data-uid={'eee'}>
+    <Scene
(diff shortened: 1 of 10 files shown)


gbalint (approved):

editor/src/core/workers/parser-printer/parser-printer-conditionals.test-utils.ts
```diff
@@ -19,3 +19,27 @@ export var ${BakedInStoryboardVariableName} = (
...
+  return <div data-uid={'div'}>
+    {[0, 1].length > 1 ? (
+      [0, 1].length === 0 ? (
+        <div data-uid='middle'/>
+      ) : null
+    ) : null}
+  </div>
+}
+export var ${BakedInStoryboardVariableName} = (
+  <Storyboard data-uid={'eee'}>
```
Rheeseyb, line 36: 🐬

Rheeseyb (approved):

PR #3328, opened 2023-02-23
Inspector clean up

lankaukk:
**Problem:**
The inspector as it is today has many inconsistencies, making it difficult to introduce space for more features (like conditionals, interactions, etc.), or redesign the right pane.

**Fix:**
This is the beginning of cleaning up what we already have, but there is more to be done! 

Before:
image

After:
image


editor/src/components/inspector/sections/layout-section/self-layout-subsection/self-layout-controls.tsx
```diff
@@ -27,7 +27,6 @@ export const LayoutTypePicker = React.memo(({ value, setActiveTab }: LayoutTypeP
           break
         }
         default: {
-          const _exhaustiveCheck: never = newValue
```
Rheeseyb, line 30: We need this please. This line is to ensure we catch changes made to the type `SelfLayoutTab` so that we can update the switch statement. It looks like both of the new types should be grouped with "absolute" and "sticky", calling `onSubmitValue`

enidemi (approved):

PR #3330, opened 2023-02-23
Fix keyboard move distance guidelines

Rheeseyb:
**Problem:**
Distance guidelines (shown by holding alt) are supposed to be measured against the element under the mouse, but when moving the selected element via the keyboard the distances would switch to using the parent element.

**Fix:**
The problem is that we are clearing the `highlightedViews` during the keyboard move interaction, so I've removed that line. This does mean that we'll still be showing the highlight control on the canvas, which feels wrong, but it feels even more wrong to hide the highlight control but still show the distance guidelines. I think the optimal setup here would be to introduce a new type of border for the sake of the distance guidelines, but I suggest that doesn't happen as part of this bug fix.


maltenuhn (approved):
Tested this and it's so much better (and still v performant, holding down arrow key will reliably show the line each time and not skip).

bkrmendy (approved):

balazsbajorics (approved):

PR #3333, opened 2023-02-24
Move conditionals toggle flag comments to the leading comments

### LINKED ISSUE #3332: put conditional flag comments to the top

ruggi:
Fixes #3332 

**Problem:**

The `@utopia` flag comments in https://github.com/concrete-utopia/utopia/pull/3316 were temporarily put in the trailing comments for the expression, because the expression parsing was not taking the leading comments into account. However this is suboptimal since those annotations should intuitively be at the top of the expression block.

**Fix:**

This PR adjusts the comments parsing and printing so that the flag comments are inserted at the top of the expression when using the inspector controls:
Kapture 2023-02-24 at 11 29 04

Additionally, the changes make sure that when there are other comments mixed in with the flag comment(s), they are arranged correctly so that leading and trailing comments stay where they were.

Kapture 2023-02-24 at 11 29 44

Finally, the changes make it so that the comment flag can be put in either leading or trailing comments when done via the code, and the canvas renders the right conditional consequently.

To recap, the final form of the comments will follow this structure:

```
{
  [leading comments]
  [toggle flag]
  [condition] ? [true branch] : [false branch]
  [trailing comments]
}
```

### DIFF
--- a/editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
+++ b/editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
@@ -2262,7 +2262,7 @@ describe('inspector tests with real metadata', () => {
         <div data-uid='aaa'>
         {
           // hello
-          [].length === 0 ? (
+          [].length === 0 /*inside*/ ? (
           <div data-uid='bbb' data-testid='bbb'>foo</div>
         ) : (
           <div data-uid='ccc' data-testid='ccc'>bar</div>
@@ -2294,7 +2294,7 @@ describe('inspector tests with real metadata', () => {
               {
                 // hello
                 // @utopia/conditional=false
-                [].length === 0 ? (
+                [].length === 0 /*inside*/ ? (
                   <div data-uid='bbb' data-testid='bbb'>foo</div>
                 ) : (
                   <div data-uid='ccc' data-testid='ccc'>bar</div>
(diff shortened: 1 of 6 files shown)


editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
```diff
@@ -91,6 +91,25 @@ function actionsForUpdatedCode(updatedCodeSnippet: string) {
   return [updateAction, requestLintAction]
 }
 
+async function toggleConditional(
+  renderResult: EditorRenderResult,
+  buttonTestId: string,
+  targetPath: ElementPath,
+): Promise<void> {
+  await act(async () => {
```
bkrmendy, line 99: could we use the single undo step checker function here?

bkrmendy (approved):

editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
ruggi: we can!

ruggi: done in https://github.com/concrete-utopia/utopia/pull/3333/commits/3e8e45faf79cd9480673889952720c7795fa88b8

Rheeseyb (approved):

editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
bkrmendy: thanks!

balazsbajorics (approved):

PR #3334, opened 2023-02-24
Detect layout systems based on children

bkrmendy:
## Problem
Now that we have fragments with children, but without `specialSizeMeasurements` filled out, checking the `layoutSystem` of a parent doesn't always yield the correct results, since for example children in a fragment in a flex context are laid out by flex, but their parent (according to EP.parentPath) doesn't reflect this fact.

## Fix
Factor out usages of `specialSizeMeasurements.layoutSystem`, and replace it with a helper call that determines the layout system based on the children's `specialSizeMeasurements`, which reflects the layout system of the actual DOM parent.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
@@ -1,7 +1,6 @@
 import { styleStringInArray } from '../../../../utils/common-constants'
 import { MetadataUtils } from '../../../../core/model/element-metadata-utils'
-import { mapDropNulls } from '../../../../core/shared/array-utils'
-import { defaultDisplayTypeForHTMLElement } from '../../../../core/shared/dom-utils'
+import { allElemsEqual, mapDropNulls } from '../../../../core/shared/array-utils'
 import * as EP from '../../../../core/shared/element-path'
 import {
   DetectedLayoutSystem,
@@ -66,16 +65,20 @@ export function singleAxisAutoLayoutContainerDirections(
   container: ElementPath,
   metadata: ElementInstanceMetadataMap,
 ): SingleAxisAutolayoutContainerDirections | 'non-single-axis-autolayout' {
-  const containerElement = MetadataUtils.findElementByElementPath(metadata, container)
   const children = MetadataUtils.getOrderedChildrenParticipatingInAutoLayout(metadata, container)
-  if (containerElement == null) {
+  if (children.length === 0) {
     return 'non-single-axis-autolayout'
   }
 
-  const layoutSystem = containerElement.specialSizeMeasurements.layoutSystemForChildren
-  const flexDirection = MetadataUtils.getSimpleFlexDirection(containerElement)
+  const layoutSystem = MetadataUtils.findLayoutSystemForChildren(metadata, container)
+  const flexDirection = MetadataUtils.findFlexDirectionForChildren(metadata, container) ?? 'row'
 
-  return singleAxisAutoLayoutDirections(children, metadata, layoutSystem, flexDirection)
+  return singleAxisAutoLayoutDirections(
+    children,
+    metadata,
+    layoutSystem,
+    MetadataUtils.flexDirectionToSimpleFlexDirection(flexDirection),
+  )
 }
 
 export function singleAxisAutoLayoutChildrenDirections(
--- a/editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
@@ -26,6 +26,7 @@ import { stylePropPathMappingFn } from '../../../inspector/common/property-path-
 import { DeleteProperties } from '../../commands/delete-properties-command'
 import { SetProperty, setProperty } from '../../commands/set-property-command'
 import { getTargetPathsFromInteractionTarget, InteractionTarget } from '../canvas-strategy-types'
+import { AllElementProps } from '../../../editor/store/editor-state'
 
 export function isValidFlowReorderTarget(
   path: ElementPath,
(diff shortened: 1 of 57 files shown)


seanparsons (approved):
Worth confirming if the de-selection performance test is flagging up an actual regression as it appears to have changed quite dramatically.

seanparsons:
> Worth confirming if the de-selection performance test is flagging up an actual regression as it appears to have changed quite dramatically.

Ignore this, was related to something else which has been fixed in master.

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
```diff
@@ -66,16 +65,20 @@ export function singleAxisAutoLayoutContainerDirections(
   container: ElementPath,
   metadata: ElementInstanceMetadataMap,
 ): SingleAxisAutolayoutContainerDirections | 'non-single-axis-autolayout' {
-  const containerElement = MetadataUtils.findElementByElementPath(metadata, container)
   const children = MetadataUtils.getOrderedChildrenParticipatingInAutoLayout(metadata, container)
-  if (containerElement == null) {
+  if (children.length === 0) {
```
balazsbajorics, line 69: if it has 0 children we should fall back to layoutSystemForChildren

editor/src/components/canvas/canvas-strategies/strategies/reparent-metastrategy.tsx
```diff
@@ -87,13 +87,25 @@ export function getApplicableReparentFactories(
...
+          ),
+        )
+
+        const parentLayouSystems = mapDropNulls(
+          (path) => MetadataUtils.findElementByElementPath(canvasState.startingMetadata, path),
+          childrenPaths,
+        ).map((instance) => instance.specialSizeMeasurements.parentLayoutSystem)
+
+        if (!allElemsEqual(parentLayouSystems)) {
+          throw new Error('All children should have the same parent layout system')
```
balazsbajorics, line 105: I think this is not an invariant, if the target element in question is a component that wraps children in a weird way we can end up with children that have different actual DOM parents

editor/src/components/canvas/canvas-strategies/strategies/reparent-metastrategy.tsx
```diff
@@ -87,13 +87,25 @@ export function getApplicableReparentFactories(
       case 'REPARENT_AS_STATIC': {
         const fitness = result.isFallback ? 2 : 3
 
-        const targetParentDisplayType =
-          MetadataUtils.findElementByElementPath(
+        const childrenPaths = replaceContentAffectingPathsWithTheirChildrenRecursive(
```
balazsbajorics, line 90: what if it has no children at this point? I think we should fall back to layoutSystemForChildren here too if it's available

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1756,6 +1757,44 @@ export const MetadataUtils = {
...
+    if (parentLayouts.length === 0) {
+      // fallback to parent instance
+      return (
+        MetadataUtils.findElementByElementPath(metadata, parentPath)?.specialSizeMeasurements
+          .layoutSystemForChildren ?? 'none'
+      )
+    }
+    const allEqual = parentLayouts.slice(1).every((x) => x === parentLayouts[0])
+    if (!allEqual) {
+      throw new Error('All children should have the same `parentLayoutSystem`')
```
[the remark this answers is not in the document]
balazsbajorics: ah I didn't see this comment, I also wrote about this here https://github.com/concrete-utopia/utopia/pull/3334/files#r1119051346

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1788,6 +1789,44 @@ export const MetadataUtils = {
       isJSXConditionalExpression(element.element.value)
     )
   },
+  findLayoutSystemForChildren(
+    metadata: ElementInstanceMetadataMap,
+    parentPath: ElementPath,
+  ): DetectedLayoutSystem {
+    const children = MetadataUtils.getOrderedChildrenParticipatingInAutoLayout(metadata, parentPath)
```
balazsbajorics, line 1796: I _think_ we don't need to filter for children that are participating in the layout

editor/src/components/canvas/canvas-strategies/strategies/reparent-metastrategy.tsx
balazsbajorics: actually, why can't this use findLayoutSystemForChildren ?

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-helpers.ts
bkrmendy: I removed this check in https://github.com/concrete-utopia/utopia/pull/3334/commits/c19346a02a9a4afc365760a1da4b9fcdb49e1817 since `findLayoutSystemForChildren` already does the fallback you mentioned

editor/src/core/model/element-metadata-utils.ts
bkrmendy: fixed as part of https://github.com/concrete-utopia/utopia/pull/3334/commits/9061d33830fd6e528efad4f611ca641d1a7ec414

editor/src/components/canvas/canvas-strategies/strategies/reparent-metastrategy.tsx
bkrmendy: great idea, fixed in https://github.com/concrete-utopia/utopia/pull/3334/commits/9061d33830fd6e528efad4f611ca641d1a7ec414

editor/src/components/canvas/canvas-strategies/strategies/reparent-metastrategy.tsx
bkrmendy: we can use it, implemented in https://github.com/concrete-utopia/utopia/pull/3334/commits/9061d33830fd6e528efad4f611ca641d1a7ec414

editor/src/core/model/element-metadata-utils.ts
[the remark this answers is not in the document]
bkrmendy: How so though? All elements have only one parent in the DOM, and that parent provides only one layout system. In the PR we check `specialSizeMeasurements.parentLayoutSystem`, which is provided by the DOM parent, and we only check one level deep (so only the children of a given parent, modulo `replaceContentAffectingPathsWithTheirChildrenRecursive`).

balazsbajorics (approved):

PR #3340, opened 2023-02-24
Feature/canvas resize flex grow insertion

enidemi:
**Problem:**
Inserting into a flex layout should use the new flex grow snapping resize strategy.

**Fix:**
My original solution didn't work because it relied on the target element position, for that the fake metadata is not a good solution as the top/left position is unknown, we only know where the mousedown started not where the element is placed in the layout.
I updated the snapping logic to find the snap point based on siblings sizes, gaps and paddings (this was needed only for center alignment previusly). 

**Commit Details:**
- removed filtered strategies `RegisteredCanvasStrategiesWithoutFlexResize`
- extended the insertion fake metadata with more parent specialSizeMeasurements
- tests for insertion
- much more tests for resize


seanparsons (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2080,6 +2080,12 @@ export function createFakeMetadataForElement(
   specialSizeMeasurements.parentLayoutSystem = isFlex ? 'flex' : 'none'
   specialSizeMeasurements.parentFlexDirection =
     parentElement?.specialSizeMeasurements.flexDirection ?? 'row'
+  specialSizeMeasurements.immediateParentBounds =
+    parentBounds == null || isInfinityRectangle(parentBounds) ? null : parentBounds
```
Rheeseyb, line 2084: We have `nullIfInfinity` in `math-utils.ts` for this

Rheeseyb (approved):

PR #3348, opened 2023-02-24
Inspector clean up

lankaukk:
**Problem:**
The inspector as it is today has many inconsistencies, making it difficult to introduce space for more features (like conditionals, interactions, etc.), or redesign the right pane.

**Fix:**
This is the beginning of cleaning up what we already have, but there is more to be done! 

Before:
image

After:
image


editor/src/components/inspector/sections/layout-section/flex-element-subsection/flex-element-subsection.spec.browser2.tsx
```diff
@@ -202,7 +202,8 @@ async function changeDimensionValue(
   await editor.getDispatchFollowUpActionsFinished()
 }
 
-describe('flex dimension controls', () => {
+// These flex dimenson controls no longer exist
```
Rheeseyb, line 205: Should the tests be binned off?

Rheeseyb (approved):

bkrmendy (approved):

PR #3349, opened 2023-02-27
Outline view for conditional expressions.

seanparsons:
**Problem:**
Currently conditionals only show the currently enabled branch, which means that it's impossible to modify the structure without flipping the conditional back and forth.

**Fix:**
Now when rendering the canvas, upon reaching a conditional expression the alternative case of the conditional is partly explored to provide an entry so that it can be shown in the navigator. As opposed to all the other elements which rely on insertion order to provide the ordering in the navigator for a conditional we fix the ordering to have the `whenTrue` case before the `whenFalse` case.

A fix was also included for `fillSpyOnlyMetadata` which was only including recursive children, not the immediate children in a listing of children.

**Screenshot:**
image

**Limitations:**
Currently this is a view only rendering, so reparenting into the alternative case will not work.

**Commit Details:**
- Added `addConditionalAlternative` which builds spy entries for conditional clauses which don't ordinarily generate them.
- Created `addFakeSpyEntry` to create a fake spy entry.
- Added `getThenOrElsePath` and `getConditionalClausePath` utility functions.
- Implemented `reorderConditionalChildPathTrees` to create a consistent ordering of `ElementPathTree` children if an element is a conditional expression.
- Fixed an issue in `fillSpyOnlyMetadata` where `childrenAndUnfurledComponentsNotInDom` was not included in `childrenAndUnfurledComponents`.
- Added `forEachOf` optic utility function.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-conditionals.spec.browser2.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-conditionals.spec.browser2.tsx
@@ -3,6 +3,7 @@ import { StoryboardFilePath } from '../../editor/store/editor-state'
 import { renderTestEditorWithModel } from '../ui-jsx.test-utils'
 import * as EP from '../../../core/shared/element-path'
 import { FOR_TESTS_setNextGeneratedUids } from '../../../core/model/element-template-utils.test-utils'
+import { setFeatureForBrowserTests } from '../../../utils/utils.test-utils'
 
 const appFilePath = '/src/app.js'
 const appFileContent = `
@@ -62,6 +63,7 @@ async function createAndRenderProject() {
 }
 
 describe('a project with conditionals', () => {
+  setFeatureForBrowserTests('Conditional support', true)
   it('fills the content of the navigator', async () => {
     FOR_TESTS_setNextGeneratedUids([
       'mock1',
@@ -74,7 +76,7 @@ describe('a project with conditionals', () => {
       'conditional1',
     ])
     const renderedProject = await createAndRenderProject()
-    const navigatorTargets = renderedProject.getEditorState().derived.navigatorTargets
+    const navigatorTargets = renderedProject.getEditorState().derived.visibleNavigatorTargets
     const pathStrings = navigatorTargets.map(EP.toString)
     expect(pathStrings).toEqual([
       'storyboard/scene',
@@ -83,6 +85,8 @@ describe('a project with conditionals', () => {
       'storyboard/scene/app:app-root/conditional1',
       'storyboard/scene/app:app-root/conditional1/conditional2',
       'storyboard/scene/app:app-root/conditional1/conditional2/div-inside-conditionals',
+      'storyboard/scene/app:app-root/conditional1/conditional2/else-case',
+      'storyboard/scene/app:app-root/conditional1/else-case',
     ])
   })
 })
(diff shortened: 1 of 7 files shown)


Rheeseyb (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -340,17 +344,22 @@ export function renderCoreElement(
     case 'JSX_CONDITIONAL_EXPRESSION': {
       const commentFlag = findUtopiaCommentFlag(element.comments, 'conditional')
       const override = isUtopiaCommentFlagConditional(commentFlag) ? commentFlag.value : null
-      const conditionValue: boolean =
+      const conditionValue =
```
Rheeseyb, line 347: I think this explicit typing was required because the return type of `jsxAttributeToValue` is `any`. Maybe it's worth adding a runtime check that the value is of type `boolean`?

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-conditionals.spec.browser2.tsx
```diff
@@ -62,6 +63,7 @@ async function createAndRenderProject() {
 }
 
 describe('a project with conditionals', () => {
+  setFeatureForBrowserTests('Conditional support', true)
```
ruggi, line 66: it'd be good to do a cleanup pass as well to reset the flag to the default value
```suggestion
  before(() => setFeatureEnabled('Conditional support', true))
  after(() => setFeatureEnabled('Conditional support', false))
```

ruggi (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-conditionals.spec.browser2.tsx
balazsbajorics: that's what setFeatureForBrowserTests does – check the implementation! :)

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
seanparsons: I removed it because it would be valid in JavaScript to write something like `"definitelytrue" ? 1 : 2`, but treating that string as a boolean would definitely be wrong and conceivably it could end up being passed into some of our logic at some point and then all kinds of havoc could be wrought.

seanparsons: To make our truthiness check happy I think I'll check if it is truthy then use that resulting boolean in the logic after that.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-conditionals.spec.browser2.tsx
ruggi: oooh sorry, I should have checked better. ignore me!

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
seanparsons: Fixed in 60dbbfb90.

PR #3352, opened 2023-02-27
Set conditional branch value to null when deleting

### LINKED ISSUE #3351: when deleting a conditional branch, set its code to `null`

ruggi:
Fixes #3351 

**Problem:**

Deleting a conditional branch from the canvas or the navigator does nothing.

**Fix:**

This PR makes it so that when deleting a conditional branch its code is replaced with `null` in the project.

Note: this does not solve that the _conditional_ parent element will disappear from the navigator, but that's a problem for another ~~day~~ PR.

### DIFF
--- a/editor/src/core/shared/element-template.ts
+++ b/editor/src/core/shared/element-template.ts
@@ -1060,15 +1060,15 @@ export interface JSXConditionalExpression extends WithComments {
   type: 'JSX_CONDITIONAL_EXPRESSION'
   uniqueID: string
   condition: JSXAttribute
-  whenTrue: ChildOrAttributeWithBranch
-  whenFalse: ChildOrAttributeWithBranch
+  whenTrue: ChildOrAttribute
+  whenFalse: ChildOrAttribute
 }
 
 export function jsxConditionalExpression(
   uid: string,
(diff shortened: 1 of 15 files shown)


editor/src/core/shared/element-template.ts
```diff
@@ -1929,6 +1929,29 @@ export function getElementsByUIDFromTopLevelElements(
 
 export type ChildOrAttribute = JSXElementChild | JSXAttribute
 
+export type ChildOrAttributeWithBranch = ChildOrAttribute & { branch: boolean }
```
seanparsons, line 1932: I think we can avoid needing this boolean by using some code from `conditionals.ts` (with a tiny bit of tweaking) to get the last part of the path for both clauses of the conditional. We should be able to take the logic from `getConditionalClausePath` for that.

ruggi: this should do it! https://github.com/concrete-utopia/utopia/pull/3352/commits/2322fb32339dc44f84032f481c17ad9272ce18ed

seanparsons (approved):

gbalint (approved):

PR #3354, opened 2023-02-27
Fix keyboard reorder tests

balazsbajorics:
**Problem:**
The keyboard reorder tests for fragments were wonky. In certain directions they asserted pressing the keyboard _twice_ to get one unit of reorder!

**Why?**
Because on master every fragment was parsed with leading and trailing whitespace elements, and the keyboard reorder apparently takes them into account.

**But then what happened?**
This PR #3338 removed empty white spaces from being parsed.

**But that's good, right?**
Yeah! And these tests broke because they were asserting a buggy behavior that got fixed.

**Yay?**
Yay!


Rheeseyb (approved):
Yay

ruggi (approved):

PR #3355, opened 2023-02-27
Make flow dragging more permissive

Rheeseyb:
**Test it [here](https://utopia.pizza/p/67e0642d-glory-handstand/?branch_name=fix-brittle-flow-reordering)**

**Problem:**
We resort to showing the slider re-order control far too easily for flow elements

**Fix:**
There are a number of tweaks in this PR to try to make the re-ordering behaviour feel less brittle:

1. Rather than requiring a strict single direction for the elements (preventing dragging to re-order any kind of mixture of `inline` and `block` elements even if they are visually aligned along a single axis), we now determine the predominant dimension and check if the elements are ordered along that axis. We do this by simply counting the number of `inline` vs `block` elements, which IMO works better than comparing the total width and height for this specific use case.
2. Filter out any zero sized elements (based on the `globalFrame`, which takes into account the frames of descendants too)
3. Filter non-autolayout elements (this was previously the case but recently regressed)
4. Remove the conversion between `inline-block` and `block`

Tests incoming. I just wanted to open the PR so that others could play with it.


editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-strategy.spec.browser2.tsx
```diff
@@ -228,9 +351,10 @@ async function startDraggingAnElement(
   const canvasControlsLayer = renderResult.renderedDOM.getByTestId(CanvasControlsContainerID)
 
   const startPoint = windowPoint({ x: targetElementBounds.x + 5, y: targetElementBounds.y + 5 })
+  const endPoint = offsetPoint(startPoint, dragDelta)
 
-  await mouseDownAtPoint(canvasControlsLayer, startPoint, {})
-  await mouseMoveToPoint(canvasControlsLayer, dragDelta, {})
+  await mouseDownAtPoint(canvasControlsLayer, startPoint, { modifiers: cmdModifier })
+  await mouseMoveToPoint(canvasControlsLayer, endPoint, {})
```
Rheeseyb, line 357: We weren't selecting the actual element here, or using the `dragDelta` as a delta, so the tests using this function were only working by pure coincidence (since they were checking that a strategy wasn't applicable, which it wasn't!)

editor/src/components/canvas/canvas-strategies/strategies/flow-reorder-strategy.spec.browser2.tsx
```diff
@@ -334,11 +458,38 @@ describe('Flow Reorder Strategy (Mixed Display Type)', () => {
     const strategies = renderResult.getEditorState().strategyState.sortedApplicableStrategies
 
     const flowReorderNotInStrategies = Array.isArray(strategies)
-      ? strategies.findIndex((strategy) => strategy.name === 'FLOW_REORDER')
+      ? strategies.findIndex((strategy) => strategy.strategy.id === 'FLOW_REORDER')
```
Rheeseyb, line 461: Also here we were filtering the applicable strategies by comparing our expected ID against the name.

ruggi (approved):
🏅

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/drag-to-move-metastrategy.spec.browser2.tsx
```diff
@@ -12,7 +12,7 @@ const TestProject1 = `
     data-testid='child-1'
   />
   <div
-    style={{ backgroundColor: '#aaaaaa33', position: 'absolute', left: 100, top: 150, width: 200, height: 100 }}
```
balazsbajorics, line 15: why this change in the test?

balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/strategies/drag-to-move-metastrategy.spec.browser2.tsx
Rheeseyb: This test was used to check that the `DO_NOTHING` strategy would be applicable, and it no longer was as a result of this PR.

PR #3359, opened 2023-02-28
Make absolute duplicate work with fragments

bkrmendy:
02-38-tc9nz-py3cc

## Problem
The absolute duplicate strategy doesn't work when trying to duplicate a fragment.

The root cause of this is that the absolute duplicate strategy consists of two distinct parts:
1. Duplicating the selected element(s)
2. Moving the selected elements using the absolute move strategy.
The duplication was performed via a command (`duplicateElement`) and by creating a fake metadata map with a duplicate of the selected element(s) metadata. The absolute move strategy from step 2 used this fake metadata.

However, when trying to absolute duplicate fragments, this breaks down:
1. Duplicating the fragment works
2. Moving the fragment doesn't work, because only the fragment itself is duplicated, and the absolute move strategy expects the fragment to have children, and since it doesn't have any, it deems the fragment unmoveable and nopes out.

## Fix
Remove the code for manually duplicating the metadata, and instead move the original element and duplicate it based on the starting editor state. This way, the original element is moved, and the new element ends up in the place of the original element.


editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.spec.browser2.tsx
```diff
@@ -44,7 +52,7 @@ describe('Absolute Duplicate Strategy', () => {
       'await-first-dom-report',
     )
 
-    FOR_TESTS_setNextGeneratedUid('hello')
+    FOR_TESTS_setNextGeneratedUids(['hello', 'hello', 'hello', 'hello', 'hello', 'hello'])
```
balazsbajorics, line 55: hello! :D is it legal to generate the same UID? :D other tests usually do something like "dummy1", "dummy2" ... "hello"

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.spec.browser2.tsx
```diff
@@ -55,12 +63,12 @@ describe('Absolute Duplicate Strategy', () => {
...
-          />
-          <div
+            />
+            <div
             style={{ backgroundColor: '#aaaaaa33', position: 'absolute', left: 80, top: 25, width: 200, height: 120 }}
-            data-uid='hello'
+            data-uid='bbb'
             data-testid='bbb'
           />
         </div>
```
balazsbajorics, line 74: I think this test change reveals a problem that should be fixed:
if I drag to duplicate `bbb`, I'd expect the duplicated element to appear _below_ the original element in the code.

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.spec.browser2.tsx
```diff
@@ -93,16 +101,91 @@ describe('Absolute Duplicate Strategy', () => {
...
+      await mouseDragFromPointToPoint(canvasControlsLayer, startPoint, endPoint, {
+        modifiers: altModifier,
+      })
+
+      await renderResult.getDispatchFollowUpActionsFinished()
+
+      expect(Object.keys(renderResult.getEditorState().editor.jsxMetadata)).toEqual([
+        'sb',
+        'sb/fragment2',
+        'sb/fragment2/child',
```
balazsbajorics, line 158: hmmm, this is suspicious to me. isn't it a problem that the uid 'child' is duplicated?

balazsbajorics (commented):
looks promising!

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.spec.browser2.tsx
bkrmendy: great find! It turns out that I was a bit too heavy-handed on refactoring, and accidentally dropped the code that saved the newly generated UIDs in `customStrategyState`. Without this, every time `apply` was called, a fresh UID was created, which bubbled up in tests as a need to generate the same UID multiple times, so the UID in the expected DOM is the desired one. Looking on the bright side, `FOR_TESTS_setNextGeneratedUid` is also a way to assert that only one new UID is generated, and it worked very well in this case.

Fixed in https://github.com/concrete-utopia/utopia/pull/3359/commits/ebfff98ed56674009b6eb9416ee785554b007884

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.spec.browser2.tsx
bkrmendy: great catch! seems like the utility function used to ensure unique UIDs (`guaranteeUniqueUids`) didn't take fragments and their children into account. Fixed in https://github.com/concrete-utopia/utopia/pull/3359/commits/65a775ac7f4697b9be24a3769162b0cafad2e9a5

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.spec.browser2.tsx
```diff
@@ -105,4 +114,100 @@ describe('Absolute Duplicate Strategy', () => {
...
+          style={{
+            backgroundColor: '#d089cc',
+            width: 150,
+            height: 186,
+            contain: 'layout',
+            left: 7,
+            top: 186,
+            position: 'absolute',
+          }}
+          data-uid='child'
```
Rheeseyb, line 133: Very minor, but can you change the uid to 3 characters please in this test, since the duplicate will have a 3 character uid

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.spec.browser2.tsx
Rheeseyb: I thought about this, because I was expecting the same thing, but thinking about it it doesn't _actually_ matter which one comes first (the original or the duplicate), as long as the **moved** element comes second (because that gives the impression that the duplicate element has come second). If the only thing showing this to be the original element is the UID then that doesn't matter since that's internal anyway (and will change on the next parse).

The only part where this might become a problem is if it introduces some edge case that can cause selection to switch elements at some point after the duplication (due to the UIDs switching)

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.tsx
```diff
@@ -73,43 +77,46 @@ export function absoluteDuplicateStrategy(
...
+            duplicateElement('always', selectedElement, newUid, 'before'),
+            /**
+             * TODO: the UIDs are swapped to make it seem like that the element being dragged
+             * (the original element) is the new, duplicated element. While this works on the canvas,
+             * in the code editor, the wrong code snippet is highlighted until the affected files are saved.
+             *
+             * I added this as a fallback solution, definitely needs a better approach
+             */
+            updatePropIfExists('on-complete', selectedElement, create('data-uid'), newUid),
+            updatePropIfExists('on-complete', newPath, create('data-uid'), oldUid),
```
Rheeseyb, line 104: This issue still appears to be present (I was just able to reproduce it on the current version of this branch). I don't think we need to (or even should) swap the UIDs here. As I mentioned in another comment, the UIDs are internal only anyway, so as long as the moved element is the one that appears second in the Navigator / in code then I think we're good tbh

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.spec.browser2.tsx
bkrmendy: done!

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.tsx
bkrmendy: Removed the uid switching in https://github.com/concrete-utopia/utopia/pull/3359/commits/c7b7dd1730fe94b548303f68d595f0747b989934

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.spec.browser2.tsx
bkrmendy: Fixed in https://github.com/concrete-utopia/utopia/pull/3359/commits/c7b7dd1730fe94b548303f68d595f0747b989934

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.tsx
```diff
@@ -165,10 +143,27 @@ function isApplicable(
...
     )
 
+    const isElementContentAffecting = treatElementAsContentAffecting(
+      canvasState.startingMetadata,
+      canvasState.startingAllElementProps,
+      element,
+    )
+
+    const isElementAbsolute = isElementContentAffecting
+      ? MetadataUtils.getChildrenPathsUnordered(canvasState.startingMetadata, element).every(
```
balazsbajorics, line 153: this should be recursively unrolling the children instead! `replaceContentAffectingPathsWithTheirChildrenRecursive` to the rescue

balazsbajorics (approved):

PR #3361, opened 2023-02-28
Dim inactive conditional branches in navigator

### LINKED ISSUE #3360: Dim inactive conditional branches in navigator

ruggi:
Fixes #3360 

**Problem:**

Both conditional branches are now shown in the navigator, but it's impossible to discriminate there between the rendered and the not rendered ones.

**Fix:**

This PR dims the inactive branch in the navigator for conditional elements. It still plays nice with the visibility toggle, without stuttering with its functionality.

Kapture 2023-02-28 at 12 27 03

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item.tsx
@@ -276,14 +276,14 @@ function useIsProbablyScene(path: ElementPath): boolean {
 }
 
 const isHiddenConditionalBranchSelector = createSelector(
-  (elementPath: ElementPath) => elementPath,
-  (_elementPath: ElementPath, parentPath: ElementPath) => parentPath,
-  (_elementPath: ElementPath, parentPath: ElementPath, store: MetadataSubstate) =>
+  (store: MetadataSubstate, _elementPath: ElementPath, parentPath: ElementPath) =>
     MetadataUtils.findElementByElementPath(store.editor.jsxMetadata, parentPath),
+  (_store: MetadataSubstate, elementPath: ElementPath, _parentPath: ElementPath) => elementPath,
+  (_store: MetadataSubstate, _elementPath: ElementPath, parentPath: ElementPath) => parentPath,
   (
+    parent: ElementInstanceMetadata | null,
     elementPath: ElementPath,
     parentPath: ElementPath,
-    parent: ElementInstanceMetadata | null,
   ): boolean => {
     const conditional = asConditional(parent)
     if (conditional == null) {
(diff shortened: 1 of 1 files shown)


editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -372,14 +373,45 @@ export const NavigatorItem: React.FunctionComponent<
...
+          wantOverride: true,
+          parentOverride: true,
+        })
+      : matchesOverriddenBranch(elementPath, parentPath, {
+          clause: conditional.whenFalse,
+          branch: 'else',
+          wantOverride: false,
+          parentOverride: false,
+        })
+  }, [parent, elementPath, parentPath])
```
balazsbajorics, line 405: could you turn this into a selector please? that way we don't re-render this row if the ElementInstanceMetadata of parent changes

ruggi: https://github.com/concrete-utopia/utopia/pull/3361/commits/7edbd68e2270e434c8704a6926684d7109887bf1

balazsbajorics: @ruggi thank you! I'd only ask one more tweak: please make store the first parameter of the selectors – it's a convention that makes it easier to see store vs parameters when browsing the code for all our selectors

ruggi: 🫡 https://github.com/concrete-utopia/utopia/pull/3361/commits/b8d619d765c326eb855139a3cf32193f1d78e8bb

balazsbajorics (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -272,6 +275,69 @@ function useIsProbablyScene(path: ElementPath): boolean {
...
+    parent: ElementInstanceMetadata | null,
+    elementPath: ElementPath,
+    parentPath: ElementPath,
+  ): boolean => {
+    const conditional = asConditional(parent)
+    if (conditional == null) {
+      return false
+    }
+    const flag = getConditionalFlag(conditional)
+    return flag == false
```
Rheeseyb, line 293:
```suggestion
    return getConditionalFlag(conditional)
```

Rheeseyb (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
ruggi: 🤦 🤯 good spot!
doing a full commit instead of the suggestion to simplify the related code

ruggi: actually no, this won't work: the existing code is correct because you need to flip the check for the _other_ branch that is not active, while checking the current one.

PR #3362, opened 2023-02-28
Ancestor metastrategy behavior with children affecting elements

gbalint:
**Problem:**
Ancestor metastrategy doesn't work well with children-affecting targets, because it checks for the properties of the children-affecting target, and not its children (even though e.g. with absolute move, the children will be moved and not the original target).

**Fix:**
Make the children affecting target transparent for the ancestor metastrategy: when we are not checking a property which is in relation to the parent (e.g. is this a root element), then use the children of a children affecting element.

**Disclaimer:**
The ancestor metastrategy does not work for multiselection. Because we modify the target to all children, when there are multiple children, the ancestor metastrategy will never be active. I did not improve this in the PR, because ancestor metastrategy behavior for multiselection was out of scope. On the other hand this really limits the usecases for the PR - because e.g. why would you use fragments at all when there is only a single child. So let's improve this later.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.spec.browser2.tsx
@@ -530,7 +530,7 @@ describe('finds an applicable strategy for the nearest ancestor', () => {
 
 describe('Fragments are transparent for ancestor metastrategy', () => {
   setFeatureForBrowserTests('Fragment support', true)
-  it('dragging a fragment with absolute children trigger absolute move', () =>
+  it('dragging a fragment with multiple absolute children trigger absolute move', () =>
     runTest(
       TestProjectFragmentWithAbsoluteChildren,
       EP.fromString('storyboard/scene/container/outer/fragment'),
--- a/editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.spec.browser2.tsx
@@ -545,7 +545,7 @@ describe('Fragments are transparent for ancestor metastrategy', () => {
         expect(strategies[0].strategy.id).toEqual('ABSOLUTE_MOVE')
       },
     ))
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.spec.browser2.tsx
```diff
@@ -400,6 +528,40 @@ describe('finds an applicable strategy for the nearest ancestor', () => {
...
+
+        expect(strategies?.length).toBeGreaterThan(0)
+        if (strategies == null) {
+          // here for type assertion
+          throw new Error('`strategies` should not be null')
+        }
+        expect(strategies[0].strategy.id).toEqual('ABSOLUTE_MOVE')
+      },
+    ))
+  it('dragging a fragment with no siblings and children no siblings triggers ancestor metastrategy', () =>
```
balazsbajorics, line 548:
```suggestion
  it('dragging a fragment with no siblings and single child triggers ancestor metastrategy', () =>
```

editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.spec.browser2.tsx
```diff
@@ -400,6 +528,40 @@ describe('finds an applicable strategy for the nearest ancestor', () => {
     }))
 })
 
+describe('Fragments are transparent for ancestor metastrategy', () => {
+  setFeatureForBrowserTests('Fragment support', true)
+  it('dragging a fragment with absolute children trigger absolute move', () =>
```
balazsbajorics, line 533:
```suggestion
  it('dragging a fragment with multiple absolute children trigger absolute move', () =>
```

editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.tsx
```diff
@@ -39,16 +43,27 @@ export function ancestorMetaStrategy(
       return []
     }
 
+    const groupLikeChildren = retargetStrategyToChildrenOfContentAffectingElements(canvasState)
```
Rheeseyb, line 46: Minor, but I think we should rename this, as it will only be the children if the original target was group-like. Maybe something as simple as `correctedTargets`?

Rheeseyb (approved):

PR #3364, opened 2023-02-28
Clearer handling when no default strategy available

Rheeseyb:
**Problem:**
The "DO_NOTHING" strategy serves as a placeholder for when there are strategies missing, or we actually want to do nothing by default, but it is very unclear that nothing is happening **deliberately**.

**Fix:**
1. Change styling of the top bar to reflect that nothing is applicable
2. Change the "Do nothing" strategy's label from "(Move)" to "No Default Available"

do-nothing-strategy


seanparsons (approved):

balazsbajorics (approved):
nice

PR #3369, opened 2023-03-01
Rename uniqueID to uid in JSXConditionalExpression

gbalint:
**Description:**
This is a leftover task, this uniqueID is now a real persistent utopia uid, there is no way to call it otherwise.

### DIFF
--- a/editor/src/core/workers/parser-printer/__snapshots__/parser-printer.spec.ts.snap
+++ b/editor/src/core/workers/parser-printer/__snapshots__/parser-printer.spec.ts.snap
@@ -719,6 +719,7 @@ const b = (n) => n > 0 ? n : a(10)
           "uniqueID": "",
         },
         "type": "JSX_CONDITIONAL_EXPRESSION",
+        "uid": "7dd",
         "uniqueID": "",
         "whenFalse": Object {
           "definedElsewhere": Array [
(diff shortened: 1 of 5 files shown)


editor/src/core/workers/parser-printer/__snapshots__/parser-printer.spec.ts.snap
```diff
@@ -719,6 +719,7 @@ const b = (n) => n > 0 ? n : a(10)
           "uniqueID": "",
         },
         "type": "JSX_CONDITIONAL_EXPRESSION",
+        "uid": "7dd",
         "uniqueID": "",
```
Rheeseyb, line 723: Why is there still an empty `uniqueID` showing up?

Rheeseyb (approved):

ruggi (approved):

seanparsons (approved):

editor/src/core/workers/parser-printer/__snapshots__/parser-printer.spec.ts.snap
gbalint: Nice catch, removed in e258b7323b2f1788cbedb053cf22611a3e3badb1

PR #3376, opened 2023-03-01
re-reselect

balazsbajorics:
This PR removes a small fixed overhead (0.3 - 0.5ms for each frame in sufficiently large projects) that happened because I made a mistake with selector memoization.

I have two Jest tests that explain the problem and the solution:

```typescript
it('if the selectors have arguments and it is not the same for consecutive calls, for example multiple instances of a component are calling with different target paths', () => {
  let numberOfTimesCalled = 0
  const selector = createCachedSelector(
    (store: ExampleStore) => store.exNumber,
    (store: ExampleStore) => store.exString,
    (store: ExampleStore, argument: { greeting: string }) => argument.greeting,
    (n, s, argument) => {
      // imagine I am the expensive selector that needs memoization
      numberOfTimesCalled += 1
    },
  )(() => 'defeat') // deliberately crippling createCachedSelector to simulate classic reselect for demo purposes

  const store: ExampleStore = {
    exNumber: 5,
    exString: 'hello',
  }

  selector(store, { greeting: 'hi there!' })
  expect(numberOfTimesCalled).toBe(1)

  // calling it two more times will TOTALLY increase numberOfTimesCalled!
  selector(store, { greeting: 'hello there!' })
  selector(store, { greeting: 'howdy there!' })

  expect(numberOfTimesCalled).toBe(3)

  // but the HUGE problem is that going back to a previous value will increase numberOfTimesCalled because the memo size is 1
  // notice how we are calling these with the exact same store and yet they keep recalculating!
  selector(store, { greeting: 'hi there!' })
  expect(numberOfTimesCalled).toBe(4)
  selector(store, { greeting: 'hello there!' })
  expect(numberOfTimesCalled).toBe(5)
  selector(store, { greeting: 'howdy there!' })
  expect(numberOfTimesCalled).toBe(6)
  // this is the same as multiple component instances sharing the same selector but calling it with a different target path

  // how do we solve this?
})

it('the solution is a re-reselect memoized selector', () => {
  let numberOfTimesCalled = 0
  const selector = createCachedSelector(
    (store: ExampleStore) => store.exNumber,
    (store: ExampleStore) => store.exString,
    (store: ExampleStore, argument: { greeting: string }) => argument.greeting,
    (n, s, argument) => {
      numberOfTimesCalled += 1
    },
  )((_, argument) => argument.greeting) // <- we use argument.greeting as the memo key

  const store: ExampleStore = {
    exNumber: 5,
    exString: 'hello',
  }

  selector(store, { greeting: 'hi there!' })
  expect(numberOfTimesCalled).toBe(1)

  // calling it two more times will TOTALLY increase numberOfTimesCalled! – this is expected
  selector(store, { greeting: 'hello there!' })
  selector(store, { greeting: 'howdy there!' })

  expect(numberOfTimesCalled).toBe(3)

  // but LOOK! problem is that going back to a previous value will NOT increase the numberOfTimesCalled, because greeting is the memo key that retrieves that selector
  selector(store, { greeting: 'hi there!' })
  expect(numberOfTimesCalled).toBe(3)
  selector(store, { greeting: 'hello there!' })
  expect(numberOfTimesCalled).toBe(3)
  selector(store, { greeting: 'howdy there!' })
  expect(numberOfTimesCalled).toBe(3)
  // aha! so each selector ran once, that gives us the number 3. but re-calling the selectors didn't re-run them! re-reselect properly memoized them
})
```

TLDR: Selectors have a memo size of 1, so selectors with parameters are not memoized correctly. And we were using them in quite a few places! 
Solution: Re-reselect selectors can be memoized with an extra key.

Note: I was afraid that someone will accidentally not learn about re-reselect and use createSelector for parametrized selectors, so I created a patch-package patch that changes reselect types and removes the option for multiple parameters. This had the nice side effect that I could rely on typescript to find all palces where I needed to replace reselect with re-reselect. Nice!


enidemi (approved):

bkrmendy (approved):

Rheeseyb (commented):
:+1:

PR #3377, opened 2023-03-01
Then and Else clauses show in the navigator.

seanparsons:
**Problem:**
Previously we created fake entries in the metadata for the clauses of a conditional as well as some synthetic entries for values that were within unrendered clauses so that it could be seen what (at the most basic level) was there. However this caused issues with the expectation that metadata only represents those things which are rendered, as those would appear like children to any logic handling them as such.

**Fix:**
This PR moves the concept of those clauses and the fake entries within them to the navigator itself. This means that rather than each entry in the navigator representing a single (and unique) `ElementPath`, they now represent a single (and unique) `NavigatorEntry`, which forms three possible cases:

- A regular entry for what we currently treat as an `ElementPath`, which is to say something that we are rendering.
- A conditional clause, a "Then" or "Else" to distinguish the structure within a ternary.
- A synthetic entry, which can represent an unrendered value (currently) within a conditional.

Anything relying on `navigatorTargets` or `visibleNavigatorTargets` unnecessarily was switched to using the logic that was there previously for those.

**Notes:**
This includes the very start of the work to support reparenting directly into the clauses which will be finished off once this has been merged, as otherwise this would turn into an even bigger PR.

**Commit Details:**
- Refactored `navigatorDepth` and switched other functions to use `fullDepth` from `element-path.ts` when they only care about relative depth.
- Added `NavigatorEntry`, along with its constituent subtypes and support functions.
- Switched `navigatorTargets` and `visibleNavigatorTargets` to be `Array<NavigatorEntry>`.
- Removed `addConditionalAlternative` completely.
- Created new `getNavigatorTargets`.
- Added `childOrBlockIsAttribute`.
- Removed unnecessary `DropTarget` type `ReparentToIndex`.
- Refactored `NAVIGATOR_REORDER` to work with `NavigatorEntry`, abandoning a partial effort to handle reparenting into conditional clauses.
- `DropTargetHint` now makes use of `NavigatorEntry`.
- Moved some deep equality instances to prevent issues with circular references.


editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
```diff
@@ -0,0 +1,303 @@
...
+            conditional-clause-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2/then-then-div-then
+              regular-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2/then-then-div
+            conditional-clause-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2/else-case-else
+              synthetic-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2/else-case-attribute
+        conditional-clause-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/else-div-else
+          synthetic-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/else-div-element-else-div
+      regular-utopia-storyboard-uid/scene-aaa/containing-div/sibling-div`)
+  })
+  it('dragging into child of an active clause, works as it would without the conditional', () => {
+    // TODO: Fill this out.
```
bkrmendy, line 301: Is this a TODO for this PR or a subsequent one?

seanparsons: A subsequent one.

editor/src/components/editor/store/editor-state.ts
```diff
@@ -2076,9 +2084,158 @@ export const defaultElementWarnings: ElementWarnings = {
...
+      } else if (first.type === 'CONDITIONAL_CLAUSE' && second.type === 'CONDITIONAL_CLAUSE') {
+        return conditionalClauseNavigatorEntriesEqual(first, second)
+      } else if (first.type === 'SYNTHETIC' && second.type === 'SYNTHETIC') {
+        return syntheticNavigatorEntriesEqual(first, second)
+      } else {
+        return false
+      }
+    }
+  }
+}
```
ruggi, line 2184: I think that this would be worth simplifying and made more readable:
```ts
  if (first?.type === 'REGULAR' && second?.type === 'REGULAR') {
    return regularNavigatorEntriesEqual(first, second)
  } else if (first?.type === 'CONDITIONAL_CLAUSE' && second?.type === 'CONDITIONAL_CLAUSE') {
    return conditionalClauseNavigatorEntriesEqual(first, second)
  } else if (first?.type === 'SYNTHETIC' && second?.type === 'SYNTHETIC') {
    return syntheticNavigatorEntriesEqual(first, second)
  } else {
    return false
  }
```
(or a variation without the `else`s).

balazsbajorics (approved):
Very very nice PR!

The only thing I'd change is the names True False in the UI instead of Then Else.

gbalint (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -44,27 +49,16 @@ interface ComputedLook {
...
-    return EP.navigatorDepth(elementPath, visibleNavigatorTargets) * BasePaddingUnit
-  }
-
-  const ancestors = EP.getAncestors(elementPath)
-  const ancestorsNotInNavigator = ancestors.filter(
-    (path) => !visibleNavigatorTargets.some((navigatorPath) => EP.pathsEqual(path, navigatorPath)),
-  )
+export function getElementPadding(withNavigatorDepth: number): number {
   // an empty path and the storyboard is always part of the ancestorsNotInNavigator list and that doesn't matter,
   // so we can add 2 to the offset. NOTE: this 2 is also a constant in EP.navigatorDepth for the same reason
```
ruggi, line 54: this comment should probably be updated

ruggi (commented):
just did a first pass

editor/src/components/navigator/navigator-item/navigator-item.tsx
seanparsons: Good spot!

editor/src/components/navigator/navigator.tsx
```diff
@@ -47,16 +48,26 @@ const Item = React.memo(({ index, style }: ItemProps) => {
     return editorSliceRef.current.dragSelections
   }, [editorSliceRef])
 
+  // Used to determine the views that will be selected by starting with the last selected item
+  // and selecting everything from there to `targetIndex`.
```
Rheeseyb, line 52: Thank you for adding this!

editor/src/components/editor/store/editor-state.ts
Rheeseyb: Alternatively you could replace this with 2 functions - one that performs the null checks, and then calls the second if both `first` and `second` are non-null values

editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
```diff
@@ -72,72 +84,145 @@ const labelSelector = createCachedSelector(
...
+        default:
+          throw assertNever(navigatorEntry.clause)
+      }
+    case 'SYNTHETIC': {
+      if (childOrBlockIsChild(navigatorEntry.childOrAttribute)) {
+        switch (navigatorEntry.childOrAttribute.type) {
+          case 'JSX_ELEMENT':
+            return getJSXElementNameLastPart(navigatorEntry.childOrAttribute.name)
+          case 'JSX_ARBITRARY_BLOCK':
+            return 'Unknown'
```
Rheeseyb, line 143: I _think_ this should either be `Code` or the actual code itself (which is what we do right now)

editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
```diff
@@ -72,72 +84,145 @@ const labelSelector = createCachedSelector(
...
+          default:
+            throw assertNever(navigatorEntry.childOrAttribute)
+        }
+      } else {
+        const simpleAttributeValue = jsxSimpleAttributeToValue(navigatorEntry.childOrAttribute)
+        return foldEither(
+          () => 'Unknown',
+          (value) => {
+            if (value === null) {
+              return 'null'
```
Rheeseyb, line 159: Should this and `'undefined'` use an upper case first letter? I feel like these are using the wrong casing, but equally 'Null' and 'Undefined' feel strange to me...

editor/src/components/navigator/navigator-utils.ts
```diff
@@ -0,0 +1,204 @@
...
+  visibleNavigatorTargets: Array<NavigatorEntry>
+}
+
+export function getNavigatorTargets(
+  metadata: ElementInstanceMetadataMap,
+  collapsedViews: Array<ElementPath>,
+  hiddenInNavigator: Array<ElementPath>,
+): GetNavigatorTargetsResults {
+  // Note: This will not necessarily be representative of the structured ordering in
+  // the code that produced these elements.
```
Rheeseyb, line 95: Is this still true? I know this comment came from the original source of this code, but I think that was from before we started using a tree here 🤔

Rheeseyb: If it is still true, but only applies to the `const projectTree`, please can you mention that in the comment

editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
```diff
@@ -0,0 +1,303 @@
...
+  navigatorEntryToKey,
+  regularNavigatorEntry,
+  varSafeNavigatorEntryToKey,
+} from '../editor/store/editor-state'
+import { getDomRectCenter } from '../../core/shared/dom-utils'
+import { FOR_TESTS_setNextGeneratedUids } from '../../core/model/element-template-utils.test-utils'
+import { setFeatureForBrowserTests } from '../../utils/utils.test-utils'
+import { navigatorDepth } from './navigator-utils'
+
+function dragElement(
```
Rheeseyb, line 26: Should this function be moved into `event-helpers.test-utils.tsx`?

Rheeseyb: Since this is prep work for tests from a follow up PR, feel free to disregard this question until then

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1468,12 +1485,16 @@ function updateSelectedComponentsFromEditorPosition(
     // Looks like the canvas has errored out, so leave it alone for now.
     return editor
   } else {
-    const allElementPaths = derived.navigatorTargets
     const highlightBoundsForUids = getHighlightBoundsForFile(editor, filePath)
+    const allElementPathsOptic: Optic<Array<NavigatorEntry>, ElementPath> = compose3Optics(
```
Rheeseyb, line 1489: Nice

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -387,33 +382,8 @@ export function renderCoreElement(
...
-          addConditionalAlternative(
-            metadataContext,
-            elementPath,
-            filePath,
-            imports,
-            actualElement,
-            conditionValue ? 'then' : 'else',
-          )
-        }
+        return jsxAttributeToValue(filePath, inScope, requireResult, actualElement)
```
Rheeseyb, line 385: Yay!

ruggi (approved):

Rheeseyb (approved):
chef kiss

editor/src/components/editor/store/editor-state.ts
seanparsons: Addressed in this PR: https://github.com/concrete-utopia/utopia/pull/3399

editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
seanparsons: Addressed in this PR: https://github.com/concrete-utopia/utopia/pull/3399

editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
seanparsons: I'm going to sidestep this for now as there's other design tweaks happening.

editor/src/components/navigator/navigator-utils.ts
seanparsons: Addressed in this PR: https://github.com/concrete-utopia/utopia/pull/3399

editor/src/components/navigator/navigator-item/navigator-item.tsx
seanparsons: Addressed in this PR: https://github.com/concrete-utopia/utopia/pull/3399

PR #3378, opened 2023-03-01
Support flex grow in shorthand prop in the Inspector

Rheeseyb:
**Problem:**
The Fill / Fixed / Hug controls aren't correctly recognising flex-grow when using the shorthand `flex` style prop.

**Fix:**
Parse the shorthand prop if no longhand prop is found, and include a test.

**Note:**
The shorthand prop also includes flex-basis, which we currently don't support. Also, there is a pretty important caveat to this - we will favour the longhand prop over the shorthand, regardless of the order they are set on the element.


balazsbajorics (approved):

balazsbajorics:
I pushed some test changes from #3348 here so we don't forget that there is no longer a need to `xit` that shorthand test :)

bkrmendy (approved):

enidemi (approved):

editor/src/components/inspector/inspector-common.ts
```diff
@@ -526,12 +527,26 @@ export function detectFillHugFixedState(
...
     (value) => defaultEither(null, parseCSSNumber(value, 'Unitless')),
     getSimpleAttributeAtPath(right(element.element.value.props), PP.create('style', 'flexGrow')),
   )
 
+  const flexGrow =
+    flexGrowLonghand ??
+    foldEither(
+      () => null,
+      (value) => {
+        return foldEither(
```
seanparsons, line 541: Might make more sense to use `flatMapEither` here.

seanparsons (approved):

editor/src/components/inspector/inspector-common.ts
Rheeseyb: I did start with that but ended up finding this to be slightly easier to read

editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
```diff
@@ -1920,34 +1920,34 @@ describe('inspector tests with real metadata', () => {
...
-    )) as HTMLInputElement
+    // const flexBasis = (await renderResult.renderedDOM.findByTestId(
+    //   'position-flexBasis-number-input',
+    // )) as HTMLInputElement
     const flexGrow = (await renderResult.renderedDOM.findByTestId(
-      'position-flexGrow-number-input',
-    )) as HTMLInputElement
-    const flexShrink = (await renderResult.renderedDOM.findByTestId(
-      'position-flexShrink-number-input',
+      'hug-fixed-fill-width',
```
balazsbajorics, line 1927: 🫂

PR #3380, opened 2023-03-02
Drag to trigger the escape hatch

Rheeseyb:
**Try it [here](https://utopia.pizza/p/4ad22474-saber-mist/?branch_name=spike-drag-to-escape-hatch)**

**Problem:**
Often the user will expect to be able to drag an element to move that element, but in the case where that would require the escape hatch we will either never do anything until a reparent can be triggered (bad), or trigger the ancestor strategy (worse!).

**Fix:**
Display the auto-layout bounding area, when dragging a flow or flex element. Then, when the user drags the element beyond that area, trigger the escape hatch. In the case of an element with no siblings, always trigger the escape hatch _unless_ the element is the same size as (or larger than) its parent element.

**Note:**
There were some clashes with the weighting of the various strategies here. I've tweaked them a bit here, but we need to review those in general since there are some bugs with the existing weights (unrelated to this PR). This also meant I've had to disable a test as fragments are triggering a flow re-order that is causing the drag-to-escape-hatch.

**Examples:**

**Flex:** (1D flow is the same)
drag-to-escape-hatch-flex

**2D Flow:**
drag-to-escape-hatch-do-nothing

**Single, Non-Contiguous, Flow Child:**
drag-to-escape-hatch

**Single, Contiguous, Flow Child:** (triggers the ancestor meta-strategy as before)
drag-to-escape-hatch-ancestor


editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.spec.browser2.tsx
```diff
@@ -528,7 +528,8 @@ describe('finds an applicable strategy for the nearest ancestor', () => {
     }))
 })
 
-describe('Fragments are transparent for ancestor metastrategy', () => {
+// FIXME Fragments are triggering regular strategies but probably never should be
+xdescribe('Fragments are transparent for ancestor metastrategy', () => {
```
balazsbajorics, line 532: I'm nervous about merging and forgetting this xdescribe. I'm pretty sure this is a meaningful test, it may be worth it to block this PR until we figure out how to resolve the issue

Rheeseyb: I disagree about blocking the PR because the issue is already there, it just doesn't happen to show in the specific sample project used in the test.

Rheeseyb: I've adjusted the escape hatch strategy to target the children of child-affective elements, therefore skipping over this problem, so the test has been re-enabled. As mentioned though, this problem still exists, and should be tackled separately to this PR.

enidemi (approved):

seanparsons (approved):

PR #3382, opened 2023-03-03
`x` shortcut on children-affecting elements

bkrmendy:
## Description
This PR improves the `x` shortcut so that it can handle children-affecting elements.

When `x` is pressed and a children-affecting element is selected
- if the element is a fragment, the shortcut does nothing
- if it is an unstyled div, it sizes that div to the dimension of the bounding box of its children


editor/src/utils/utils.test-utils.ts
```diff
@@ -423,14 +423,29 @@ export function slightlyOffsetPointBecauseVeryWeirdIssue(point: { x: number; y:
   return { x: point.x - 0.001, y: point.y - 0.001 }
 }
 
-export async function expectSingleUndoStep(
+async function expectNUndoSteps(
```
ruggi, line 426: Probably worth exporting this one as well

ruggi (approved):

editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -862,41 +872,78 @@ export function handleKeyDown(
...
+            editor.allElementProps,
+            elementPath,
+          )
+
+          if (contentAffectingType === 'fragment') {
+            return []
+          }
+
+          if (contentAffectingType === 'simple-div') {
+            const childrenBoundingFrame = getChildrenBoundingRectangle(
```
Rheeseyb, line 893: You should just be able to use `MetadataUtils.getFrameInCanvasCoords` here instead. That would be safer too since it would take into account any text children

editor/src/utils/utils.test-utils.ts
bkrmendy: I didn't export it on purpose because I think expecting 0 and 1 undo steps are the only sensible things to assert on, because
- you need to assert that 0 undo steps are generated if you want to assert that a disabled UI element doesn't generate an undo step even if it's clicked on
- you need to assert that 1 undo step is generated using a control (this was the original helper).
I think asserting that multiple undo steps are generated is an antipattern. Should the need for this arise while writing tests, either the test should be refactored so that a single UI interaction is wrapped in a single `expectSingleUndoStep` (or `expectNoAction`), or it the assert shouldn't be there at all because it asserts something that we consider a bug (having multiple undo steps for a single UI action)

ruggi: I guess the argument for premature optimization can go both ways - to me this sounds like it and I don't see testing for `N` steps as an anti-pattern per se, depending on the use cases; hence my suggestion since we'd be exporting the general purpose method _and_ the special cases with "nice names" (of course the counter argument stands, which is that only the "relevant" functions are exported and the implementation is hidden, if there's really no point in testing `> 1` steps). But no point in bikeshedding on this I guess, my review approval stands :)

editor/src/components/editor/global-shortcuts.tsx
bkrmendy: 🤦‍♂️ Sounds obvious in retrospect, thanks for pointing it out! https://github.com/concrete-utopia/utopia/pull/3382/commits/c37e140feccf4d03b6758940a5ec7140800c8892

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -165,7 +166,7 @@ import { fastForEach } from '../../core/shared/utils'
 import { getContentsTreeFileFromString, ProjectContentTreeRoot } from '../assets'
 import { getAllTargetsAtPointAABB } from './dom-lookup'
 import { CSSNumber, parseCSSLengthPercent, printCSSNumber } from '../inspector/common/css-utils'
-import { uniqBy } from '../../core/shared/array-utils'
+import { mapDropNulls, uniqBy } from '../../core/shared/array-utils'
```
Rheeseyb, line 169: The new imports in this file are now redundant

Rheeseyb (approved):

seanparsons (approved):

PR #3385, opened 2023-03-03
Replace non-JSXElementChild conditional branches with `null` upon deletion

### LINKED ISSUE #3383: Make conditional branch deletion work with non-child branches

ruggi:
Fixes #3383 

**Problem:**

Deleting conditional branches works only if the branch itself is a `JSXElementChild`. This is not necessarily the case, for example for conditional branches being text strings (`{ true ? 'foo' : 'bar' }`).

**Fix:**

This PR adjusts the `findJSXElementChildAtPath` function so that it handles correctly the cases where the branches are `JSXAttribute`.

Kapture 2023-03-03 at 12 14 55

### DIFF
--- a/editor/src/components/editor/conditionals.spec.browser2.tsx
+++ b/editor/src/components/editor/conditionals.spec.browser2.tsx
@@ -14,20 +14,21 @@ import { deleteSelected, selectComponents } from '../editor/actions/action-creat
 describe('conditionals', () => {
   before(() => setFeatureEnabled('Conditional support', true))
   after(() => setFeatureEnabled('Conditional support', false))
-  it('deleting a conditional branch replaces it with null', async () => {
-    FOR_TESTS_setNextGeneratedUids([
-      'skip1',
-      'skip2',
-      'skip3',
-      'skip4',
-      'skip5',
-      'skip6',
-      'conditional',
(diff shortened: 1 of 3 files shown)


bkrmendy (approved):

editor/src/components/editor/conditionals.spec.browser2.tsx
```diff
@@ -14,51 +14,86 @@ import { deleteSelected, selectComponents } from '../editor/actions/action-creat
...
          `),
-    )
+      )
+    })
+    it('replaces a text string branch with null', async () => {
+      FOR_TESTS_setNextGeneratedUids(['skip1', 'skip2', 'conditional'])
+      const startSnippet = `
+        <div data-uid='aaa'>
+        {
+          true ? 'hello' : 'there'
```
Rheeseyb, line 70: image

Rheeseyb (approved):

balazsbajorics (approved):

editor/src/components/editor/conditionals.spec.browser2.tsx
ruggi: actually lol'ed

PR #3388, opened 2023-03-07
Store conditional value in spy metadata

### LINKED ISSUE #3387: Store conditional value in spy metadata

ruggi:
Fixes #3387 

This PR adds the `conditionValue` field to `ElementInstanceMetadata`, and sets it to `true | false | null` in the spy metadata map when parsing a conditional. It also adds a helper method in `MetadataUtils` to retrieve that value.

This can be later reused e.g. in the navigator with something like

```ts
const elementSpy = useEditorState(
  Substores.metadata,
  (store) => MetadataUtils.findElementByElementPath(store.editor.spyMetadata, elementPath),
  '...',
)

// ...

const conditionValue = MetadataUtils.getConditionalValueFromMetadata(elementSpy))
```

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
@@ -352,7 +352,7 @@ export function renderCoreElement(
       const actualElement = conditionValue ? element.whenTrue : element.whenFalse
 
       if (elementPath != null) {
-        addFakeSpyEntry(metadataContext, elementPath, element, filePath, imports)
+        addFakeSpyEntry(metadataContext, elementPath, element, filePath, imports, conditionValue)
       }
 
       let result: any
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.tsx
@@ -33,7 +33,7 @@ export function addFakeSpyEntry(
   elementOrAttribute: ChildOrAttribute,
   filePath: string,
   imports: Imports,
-  conditionValue: boolean | null,
+  conditionalValue: boolean | null,
 ): void {
   let element: Either<string, JSXElementChild>
   if (childOrBlockIsChild(elementOrAttribute)) {
@@ -80,7 +80,7 @@ export function addFakeSpyEntry(
       },
       element,
     ),
-    conditionValue: conditionValue,
+    conditionalValue: conditionalValue,
   }
   const elementPathString = EP.toComponentId(elementPath)
   metadataContext.current.spyValues.metadata[elementPathString] = instanceMetadata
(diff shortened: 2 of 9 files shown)


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.tsx
```diff
@@ -79,6 +80,7 @@ export function addFakeSpyEntry(
       },
       element,
     ),
+    conditionalValue: conditionalValue,
```
balazsbajorics, line 83: shouldn't this be null? the fake entry is the -child- of the ternary, right?

balazsbajorics: oh sorry disregard, the fake entry is the one true entry for the ternary :D we should probably not call it fake. I confused it with `addConditionalAlternative` which _is_ the fake entry for the path-non-taken

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -352,7 +352,7 @@ export function renderCoreElement(
       const actualElement = conditionValue ? element.whenTrue : element.whenFalse
 
       if (elementPath != null) {
-        addFakeSpyEntry(metadataContext, elementPath, element, filePath, imports)
+        addFakeSpyEntry(metadataContext, elementPath, element, filePath, imports, conditionValue)
```
seanparsons, line 355: Hmmm, I've removed this function from my branch completely...

ruggi: is it not used anymore or it just moved?

balazsbajorics: so for the conditional itself, we should add the metadata entry. for the non-active-branch or contents of the non-active branch, we should _not_ have ElementInstanceMetadata

editor/src/core/shared/element-template.ts
```diff
@@ -1641,6 +1641,7 @@ export interface ElementInstanceMetadata {
   attributeMetadatada: StyleAttributeMetadata | null
   label: string | null
   importInfo: ImportInfo | null
+  conditionalValue: boolean | null
```
balazsbajorics, line 1644: oh I had an idea! this type could be `boolean | 'not-a-conditional'`!

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
seanparsons: Righto!

editor/src/core/shared/element-template.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/3388/commits/b76bf99eb2a5e23a3eccb599828d6d8389ddc2eb

Rheeseyb (approved):

seanparsons (approved):

PR #3389, opened 2023-03-07
Retarget keyboard strategies when grouplike element is selected

bkrmendy:
## Description
This PR upgrades our keyboard strategies so they work with grouplike/content-affecting elements. The way to accomplish this is that when a grouplike element is selected (such as a fragment), the keyboard strategies are retargeted to the children of the grouplike element using `retargetStrategyToChildrenOfContentAffectingElements`.

The strategies modified by this PR are:
- `keyboard-set-font-size`
- `keyboard-set-font-weight`
- `keyboard-set-font-opacity`
- `keyboard-absolute-move`
- `keyboard-absolute-resize`

The keyboard reorder strategy is tweaked as well, since it clashed with the keyboard resize strategy when a fragment was selected.


editor/src/components/canvas/canvas-strategies/strategies/keyboard-reorder-strategy.ts
```diff
@@ -108,10 +109,11 @@ function fitness(interactionSession: InteractionSession | null): number {
...
   const accumulatedPresses = accumulatePresses(interactionSession.interactionData.keyStates)
-  const matches = accumulatedPresses.some((accumulatedPress) =>
-    Array.from(accumulatedPress.keysPressed).some(
-      (key) => key === 'left' || key === 'right' || key === 'up' || key === 'down',
-    ),
+  const matches = accumulatedPresses.some(
+    (accumulatedPress) =>
+      Array.from(accumulatedPress.keysPressed).some(
+        (key) => key === 'left' || key === 'right' || key === 'up' || key === 'down',
+      ) && shallowEqual(accumulatedPress.modifiers, emptyModifiers),
```
Rheeseyb, line 116: We have an equality function for modifiers here: <https://github.com/concrete-utopia/utopia/blob/master/editor/src/utils/modifiers.ts#L89-L100>

Rheeseyb:
I'm very unsure why this PR would have affected the render count tests...

editor/src/components/canvas/canvas-strategies/strategies/keyboard-reorder-strategy.ts
bkrmendy: thanks! fixed in [2c7b067](https://github.com/concrete-utopia/utopia/pull/3389/commits/2c7b067e2b06fe17a5d2d77624360f0e59a1e241)

bkrmendy:
> I'm very unsure why this PR would have affected the render count tests...

It's the keyboard absolute resize strategy (found out through trial and error 😅), I assume the reason is that with this change, multiple elements are resized (the children of the grouplike container, not just the parent element), and that is causing the re-renders

seanparsons (approved):

Rheeseyb (approved):

PR #3394, opened 2023-03-08
Handle asset renames when undoing or redoing.

seanparsons:
**Problem:**
If an asset like an image is renamed, but then undo is triggered over the step which renamed the file, then the model is updated but the file in S3 is not renamed back to the original path.

This had the side problem of also causing the Github integration to fail when trying to push the project to Github, as it would attempt to retrieve assets that did not exist.

**Fix:**
Now when traversing backwards or forwards through the state history, asset renames are collected and either reversed or replayed as appropriate for the action triggered.

**Commit Details:**
- Added `AssetRename` interface for the history, to represent a rename.
- Implemented `triggerAssetRenames` for applying the asset renames in a consistent ordering.
- `undo` and `redo` now fire `triggerAssetRenames` but only if `run-side-effects` is passed as they are both called from two different places in the editor code.
- `editorDispatch` collects `UPDATE_FILE_PATH` actions and adds the entries for them to the history.

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -1634,7 +1634,7 @@ export const UPDATE_FNS = {
   },
   UNDO: (editor: EditorModel, stateHistory: StateHistory): EditorModel => {
     if (History.canUndo(stateHistory)) {
-      const history = History.undo(stateHistory)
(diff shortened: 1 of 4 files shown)


ruggi (approved):
looks good!

Rheeseyb (approved):

PR #3396, opened 2023-03-08
Only save when there are project model changes

Rheeseyb:
**Problem:**
The saving logic for the editor's model is far too sensitive to changes, meaning we are currently saving the project on `mouseup` events even if there was no change to the actual project.

**Fix:**
The actual change of interest here is in `dispatch.tsx`, and is the introduction of the function `editorChangesShouldTriggerSave` which will check if the project contents, branch contents, or github settings have changed as a result of a given action.

The rest of the changes are around testing that. Without a way to test all possible actions in a way that wouldn't require duplicating the logic in the test for every combination, the solution I've gone with here is to hijack the existing `expectNUndoSteps` (and its variants) to create `expectNUndoStepsNSaves`. We now track the total number of saves that have occurred during a given session in the `EditorState` by replacing the previously existing `alreadySaved` boolean property with a `saveCountThisSession` number.


editor/src/utils/utils.test-utils.ts
```diff
@@ -425,29 +425,46 @@ export function slightlyOffsetPointBecauseVeryWeirdIssue(point: { x: number; y:
...
 export async function expectNoAction(
   editor: EditorRenderResult,
   action: () => Promise<void>,
 ): Promise<void> {
-  return expectNUndoSteps(editor, 0, action)
+  return expectNUndoStepsNSaves(editor, 0, 0, action)
+}
+
+// FIXME This appears to be the default in a lot of cases,
+// but really we should only be experiencing 1 save in most cases
```
Rheeseyb, line 454: I can update this comment actually because I've just realised the reason

bkrmendy (approved):

maltenuhn:
As a user, what should my expectation be when this PR is merged? And should I manually test some things like the "are you sure you want to leave this page" dialogue still firing correctly  (I'm assuming we don't have automated test coverage for that)?

ruggi (approved):

PR #3400, opened 2023-03-09
Don't show ⚠️ icon for group children

bkrmendy:
## Problem
Absolute children of sizeless divs, which we consider group-like elements, displayed a ⚠️ indicating that they aren't positioned by their parent.

## Fix
Since we consider sizeless divs group-like elements, they aren't supposed to position their children anyway. To prevent misleading messaging on the UI, to reduce clutter, and to make the behaviour consistent with fragments, sizeless divs are now special-cased not to trigger the ⚠️ warning.

### DIFF
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -189,6 +189,7 @@ import { ThenOrElse } from '../../../core/model/conditionals'
 import { Optic } from '../../../core/shared/optics/optics'
 import { fromTypeGuard } from '../../../core/shared/optics/optic-creators'
 import { getNavigatorTargets } from '../../../components/navigator/navigator-utils'
+import { treatElementAsContentAffecting } from '../../canvas/canvas-strategies/strategies/group-like-helpers'
 
 const ObjectPathImmutable: any = OPI
 
@@ -2533,6 +2534,7 @@ export interface OriginalCanvasAndLocalFrame {
 
 function getElementWarningsInner(
   rootMetadata: ElementInstanceMetadataMap,
+  allElementProps: AllElementProps,
 ): ComplexMap<ElementPath, ElementWarnings> {
   let result: ComplexMap<ElementPath, ElementWarnings> = emptyComplexMap()
   Object.values(rootMetadata).forEach((elementMetadata) => {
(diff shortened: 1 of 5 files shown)


editor/src/components/editor/store/editor-state.ts
```diff
@@ -2545,26 +2547,23 @@ function getElementWarningsInner(
...
-    let absoluteWithUnpositionedParent: boolean = false
-    if (
-      elementMetadata.specialSizeMeasurements.position === 'absolute' &&
+    const isParentGroupLike = treatElementAsContentAffecting(
+      rootMetadata,
+      allElementProps,
+      EP.parentPath(elementMetadata.elementPath),
+    )
+
+    const isParentConfiguredForPins =
```
Rheeseyb, line 2556: This name is incorrect, because you're checking if the parent _isn't_ configured for pins

Rheeseyb (commented):
A rename is required, but otherwise looks good to me 👍

ruggi (approved):

editor/src/components/editor/store/editor-state.ts
bkrmendy: 🤦‍♂️ (fixed in https://github.com/concrete-utopia/utopia/pull/3400/commits/dfbf64598aa967c58e9d53b828db3440a28578ca)

editor/src/components/editor/store/editor-state.ts
```diff
@@ -2545,26 +2547,23 @@ function getElementWarningsInner(
...
+      MetadataUtils.isPositionAbsolute(elementMetadata) &&
       !elementMetadata.specialSizeMeasurements.immediateParentProvidesLayout
-    ) {
-      absoluteWithUnpositionedParent = true
-    }
+    const absoluteWithUnpositionedParent = isParentConfiguredForPins && isParentGroupLike === false
```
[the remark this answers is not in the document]
bkrmendy: fixed in https://github.com/concrete-utopia/utopia/pull/3400/commits/dfbf64598aa967c58e9d53b828db3440a28578ca

Though just for the note, I think === false is way easier to read than `!` because it's simply "bigger" on screen, and avoids having to mentally emulate the `!` operator

PR #3410, opened 2023-03-10
Fix calculation of which conditional branch is hidden

gbalint:
**Problem:**
When we count which branch should be hidden, we only take the condition override into account and not the current real  condition value.

See e.g. this simple project where the true branch should be hidden:
image

**Fix:**
In `isHiddenConditionalBranchSelector` the current condition value is not used at all, only the override from the comment flag.

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item.tsx
@@ -284,24 +284,32 @@ const isHiddenConditionalBranchSelector = createCachedSelector(
     elementPath: ElementPath,
     parentPath: ElementPath,
   ): boolean => {
+    if (parent == null) {
+      return false
+    }
+    const originalConditionValue = parent.conditionalValue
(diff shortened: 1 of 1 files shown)


editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -284,24 +284,32 @@ const isHiddenConditionalBranchSelector = createCachedSelector(
...
-        })
-      : matchesOverriddenBranch(elementPath, parentPath, {
-          clause: conditional.whenFalse,
-          branch: 'else',
-          wantOverride: false,
-          parentOverride: false,
-        })
+
+    // the final condition value, either from the original or from the override
+    const overriddenConditionValue = flag ?? originalConditionValue
```
seanparsons, line 303: Might be worth constraining this to `boolean` just in case anything here changes type.

seanparsons (approved):

bkrmendy (approved):

PR #3417, opened 2023-03-10
Conditional inspector: Condition expression

gbalint:
**Description:**
Add the condition expression of the selected conditional to the conditional inspector section.

image

**Notes**
- No design, just POC implementation
- We did not have a function to convert any kind of attributes to string, I implemented a simple one just for `JSXAttributeValue` and `JSXAttrbibuteOtherJavascript`, this can be improved later if we need it
- No multiselection support, we need a more complex ui for that
- Modified the condition override implementation so that the `True` and `False` buttons are toggles. Before that you could only remove an override by closing the whole section, but now we need the section for the condition expression even when there is no override. So you can just press on any of the active buttons again to remove the override.

### DIFF
--- a/editor/src/components/inspector/sections/layout-section/conditional-section.tsx
+++ b/editor/src/components/inspector/sections/layout-section/conditional-section.tsx
@@ -13,6 +13,7 @@ import {
 } from '../../../../core/shared/element-template'
 import { ElementPath } from '../../../../core/shared/project-file-types'
 import { jsxAttributeToString } from '../../../../core/workers/parser-printer/parser-printer'
+import { unless } from '../../../../utils/react-conditionals'
 import {
   Button,
   FlexRow,
@@ -201,7 +202,8 @@ export const ConditionalSection = React.memo(({ paths }: { paths: ElementPath[] 
           </SquareButton>
         )}
       </InspectorSubsectionHeader>
-      {conditionExpression !== 'multiselect' ? (
+      {unless(
+        conditionExpression === 'multiselect',
         <UIGridRow
           padded={true}
           variant='<---1fr--->|------172px-------|'
@@ -213,8 +215,8 @@ export const ConditionalSection = React.memo(({ paths }: { paths: ElementPath[] 
           <FlexRow style={{ flexGrow: 1, gap: 4 }}>
             <span style={{ flex: 1, textAlign: 'center' }}>{conditionExpression}</span>
           </FlexRow>
-        </UIGridRow>
-      ) : null}
+        </UIGridRow>,
+      )}
       <UIGridRow
         padded={true}
         variant='<---1fr--->|------172px-------|'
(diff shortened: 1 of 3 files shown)


editor/src/components/inspector/sections/layout-section/conditional-section.tsx
```diff
@@ -122,54 +183,65 @@ export const ConditionalSection = React.memo(({ paths }: { paths: ElementPath[]
...
+            onClick={setConditionOverride(true)}
             data-testid={ConditionalsControlSectionOpenTestId}
           >
             <Icons.Plus style={{ opacity: 'var(--buttonContentOpacity)' }} />
           </SquareButton>
         )}
       </InspectorSubsectionHeader>
-      {when(
-        condition != 'not-overridden',
+      {conditionExpression !== 'multiselect' ? (
```
ruggi, line 204: probably good to use `when` as well (or `unless` if you want to keep the comparison with `"multiselect"`)

editor/src/components/inspector/sections/layout-section/conditional-section.tsx
```diff
@@ -76,27 +76,88 @@ const conditionSelector = createCachedSelector(
...
 
+const conditionExpressionSelector = createCachedSelector(
+  (store: MetadataSubstate) => store.editor.jsxMetadata,
+  (_store: MetadataSubstate, paths: ElementPath[]) => paths,
+  (jsxMetadata: ElementInstanceMetadataMap, paths: ElementPath[]): ConditionExpression => {
+    const elements = mapDropNulls((path) => {
+      const elementMetadata = MetadataUtils.findElementByElementPath(jsxMetadata, path)
+      if (
+        elementMetadata == null ||
+        !isRight(elementMetadata.element) ||
```
ruggi, line 87: just for consistency:
```suggestion
        isLeft(elementMetadata.element) ||
```

editor/src/components/inspector/sections/layout-section/conditional-section.tsx
```diff
@@ -76,27 +76,88 @@ const conditionSelector = createCachedSelector(
...
+            return null
+          }
+          return true
+        }
+
+        if (conditionOverride === false) {
+          return null
+        }
+        return false
+      })()
```
ruggi, line 154: imho something like this would be more readable:
```ts
const newCond = whichButton === conditionalOverride ? null : whichButton
```

ruggi (approved):
just a couple nits but this all makes sense to me

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -2027,3 +2027,18 @@ export function lintAndParse(
...
   }
 }
+
+export function jsxAttributeToString(attribute: JSXAttribute): string {
+  switch (attribute.type) {
+    case 'ATTRIBUTE_VALUE':
+      if (typeof attribute.value === 'string') {
+        return attribute.value
+      } else {
+        return attribute.value.toString()
```
Rheeseyb, line 2037: `'[object Object]'`
Though I suspect if we wanted to use `JSON.stringify` here we'd have to be careful to handle all of the possible edge cases (type errors, deep objects).

Rheeseyb (approved):

seanparsons (approved):

PR #3419, opened 2023-03-13
handle conditionals when detecting content affecting type

### LINKED ISSUE #3418: Handle conditionals when detecting ContentAffectingType

ruggi:
Fixes #3418 

**Problem:**

Conditionals are treated as elements with no w/h inside `getElementContentAffectingType`. One side effect of this is not displaying correctly the conditionals icon in the navigator.

**Fix:**

Handle conditionals in `getElementContentAffectingType` so they can be handled specifically.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/group-like-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/group-like-helpers.ts
@@ -134,12 +134,5 @@ export function treatElementAsContentAffecting(
   allElementProps: AllElementProps,
   path: ElementPath,
 ): boolean {
-  const type = getElementContentAffectingType(metadata, allElementProps, path)
-  switch (type) {
-    case null:
-    case 'conditional':
-      return false
-    default:
-      return true
-  }
+  return getElementContentAffectingType(metadata, allElementProps, path) != null
 }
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/canvas-strategies/strategies/group-like-helpers.ts
```diff
@@ -119,5 +134,12 @@ export function treatElementAsContentAffecting(
   allElementProps: AllElementProps,
   path: ElementPath,
 ): boolean {
-  return getElementContentAffectingType(metadata, allElementProps, path) != null
+  const type = getElementContentAffectingType(metadata, allElementProps, path)
+  switch (type) {
+    case null:
+    case 'conditional':
```
balazsbajorics, line 140: why return false for the conditional? isn't this changing the behavior? (we are going to write test coverage for the conditional-as-group-parent this week, but we don't have those yet, so tests would not break because of this change – but nevertheless, it feels like this is changing behavior here)

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/group-like-helpers.ts
ruggi: returning `true` does change the behavior of, i.e., the navigator not displaying the right icon for the conditionals (here: https://github.com/concrete-utopia/utopia/blob/e7b9ec5bccadd4c690dded3da333a5f03c92b3e6/editor/src/components/navigator/layout-element-icons.ts#L80-L92). It can be fixed differently by not using the function directly and evaluating the content type ofc

ruggi: fixed/reverted https://github.com/concrete-utopia/utopia/pull/3419/commits/f5690029fc5522dc13b8f5bf8fb55ff8fc41b68b

bkrmendy (approved):

balazsbajorics (approved):

PR #3420, opened 2023-03-13
Tweaking canvas selection logic

Rheeseyb:
**Problem:**
It is too easy to accidentally select elements on the canvas that you do not wish to make changes to, leading to accidentally making changes to those elements rather than the ones you _actually_ wanted to update.

**Fix:**
Tweak the canvas selection logic to follow these rules:
1) Skip over the immediate `Scene` children of the `Storyboard` (but no other element types)
2) Skip over component children of `Scenes` (replacing with their root paths and children paths)
3) Skip over root element paths

Where a path cannot be skipped over (because it has no children) we will permit selection of that path (so e.g. a `Scene` at the very top level will become selectable if it has no children).

As a "bonus" I have also invested some time in improving the selection tests, as fixing them was an absolute nightmare, especially when it came to finding where the tests were actually breaking. Part of that was the inclusion of these functions that will likely come in handy for other browser tests:

```ts
function checkFocusedPath(renderResult: EditorRenderResult, expected: ElementPath | null) {
  checkWithKey('focusedPath', renderResult.getEditorState().editor.focusedElementPath, expected)
}

function checkSelectedPaths(renderResult: EditorRenderResult, expected: Array<ElementPath>) {
  checkWithKey('selectedPaths', renderResult.getEditorState().editor.selectedViews, expected)
}

function checkWithKey<T>(key: string, actual: T, expected: T) {
  expect({
    check: key,
    value: actual,
  }).toEqual({
    check: key,
    value: expected,
  })
}
```
This way we can see in the diff the value of that `check` string, giving us a clue which check is actually failing (which in this case was particularly helpful)

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
@@ -256,7 +256,12 @@ export function getSelectableViews(
   let candidateViews: Array<ElementPath>
 
   if (allElementsDirectlySelectable) {
-    candidateViews = MetadataUtils.getAllPathsIncludingUnfurledFocusedComponents(componentMetadata)
+    const unfilteredCandidates =
+      MetadataUtils.getAllPathsIncludingUnfurledFocusedComponents(componentMetadata)
+    candidateViews = unfilteredCandidates.filter(
+      (path) =>
+        path.parts.length > 2 || (path.parts.length > 1 && !EP.isRootElementOfInstance(path)),
+    )
   } else {
     const allRoots = MetadataUtils.getAllCanvasRootPathsUnordered(componentMetadata)
     const siblings = collectSelectableSiblings(
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
@@ -178,36 +178,6 @@ function filterNonSelectableElements(
   )
 }
 
-function collectSelectableSiblings(
-  componentMetadata: ElementInstanceMetadataMap,
-  selectedViews: Array<ElementPath>,
-  childrenSelectable: boolean,
-  lockedElements: LockedElements,
-): Array<ElementPath> {
-  let siblingsChildrenAndAncestors: Array<ElementPath> = []
-  Utils.fastForEach(selectedViews, (view) => {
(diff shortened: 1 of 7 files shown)


editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -184,41 +184,17 @@ function collectSelectableSiblings(
   childrenSelectable: boolean,
   lockedElements: LockedElements,
 ): Array<ElementPath> {
-  let siblings: Array<ElementPath> = []
+  let siblingsChildrenAndAncestors: Array<ElementPath> = []
   Utils.fastForEach(selectedViews, (view) => {
     function addChildrenAndUnfurledFocusedComponents(paths: Array<ElementPath>) {
```
balazsbajorics, line 189: now that the diff is _already_ unreadable, we may as well improve the code quality here! I think we could kill this function, and we could also kill the `let siblingsChildrenAndAncestors` and make this part functional

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -228,7 +204,60 @@ function collectSelectableSiblings(
...
+    const isConditional = MetadataUtils.isElementPathConditionalFromMetadata(
+      componentMetadata,
+      selectablePath,
+    )
+    const isRootPath = EP.isRootElementOfInstance(selectablePath)
+
+    const mustReplaceWithChildren =
+      isLocked ||
+      (!isFeatureEnabled('Fragment support') && isFragment) ||
+      (!isFeatureEnabled('Conditional support') && isConditional)
```
balazsbajorics, line 234: I think this should be removed. fragments and conditionals should be selectable on the canvas: they are groups now!

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
Rheeseyb: Yeah good point, I have tidied this all up now so the code should be a lot clearer 👍

seanparsons (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
balazsbajorics: ahhhh ok, I missed the `!`. ok now it makes sense!

PR #3421, opened 2023-03-13
adding test cases for findJSXElementChildAtPath

balazsbajorics:
This PR adds jest test for `findJSXElementChildAtPath`.

They are capturing expected behavior which I believe should be true, there were two tests that failed:

**1 The conditional handling missed a UID check. this meant that if the conditional was present as a sibling, it would "win" the logic and we'd walk into the conditional, even if the current visited UID was not matching the conditional's UID at all.**
Berci already fixed this bug in #3412 and I copied it over. To make it easier to avoid this mistake in the future, I moved the UID check to the same IF statements where we check the visited element's type. So hopefully if someone adds further code here, they will see that they also have to include the uid check too.

**2 When asking for aaa/parent/ca0/then-case, `findJSXElementChildAtPath` would return the conditional itself at `aaa/parent/ca0`.** 
After consulting in Discord, I changed this so it returns the actual JSXElementChild at the correct branch, or returns null if the branch's content is a JSXAttribute. 

**3 Fix No 2. broke `deleteElements` if the selected element was a JSXAttribute.** 
I fixed that by removing the actually not necessary `findJSXElementChildAtPath` check in the `deleteElements` code.

**Note:**
The ONLY case `findJSXElementChildAtPath` supports the special-cased `then-branch` or `else-branch` syntax if those are the last path elements. You can not address something with a path like `aaa/parent/conditional/then-branch/some-child-uid`. This is expected, but now it's explicit.

**Note 2:**
I think the consensus is that we probably don't want to keep the then-branch / else-branch code alive for too long, but I fixed it up for now. However I am very much looking for it to be replaced with real UIDs because now ElementPath is an array of UIDs plus this special-cased magic string, increasing the mental load, maintenance cost etc. One example: `ElementPath.toUID()` is simply broken for these special-cased paths, as it will return a string like `"then-branch"` which is certainly not unique.

### DIFF
--- a/editor/src/core/model/element-template-utils.spec.tsx
+++ b/editor/src/core/model/element-template-utils.spec.tsx
@@ -1,3 +1,5 @@
+/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "expectElementAtPathHasMatchingUID*", "expectElementFoundNull"] }] */
+
 import {
   jsxAttributeValue,
   isJSXAttributeValue,
@@ -16,6 +18,7 @@ import {
   componentHonoursPropsPosition,
   componentHonoursPropsSize,
   componentUsesProperty,
+  findJSXElementChildAtPath,
   getUtopiaID,
   guaranteeUniqueUids,
   rearrangeJsxChildren,
@@ -38,8 +41,15 @@ import {
   StoryboardFilePath,
 } from '../../components/editor/store/editor-state'
 import { ProjectContentFile } from '../../components/assets'
-import { ParseSuccess, StaticElementPath } from '../shared/project-file-types'
-import { dynamicPathToStaticPath, fromString, fromStringStatic } from '../shared/element-path'
+import { ElementPath, ParseSuccess, StaticElementPath } from '../shared/project-file-types'
+import {
+  dynamicPathToStaticPath,
+  fromString,
+  fromStringStatic,
+  toUid,
+} from '../shared/element-path'
+import { getComponentsFromTopLevelElements } from './project-file-utils'
+import { setFeatureForUnitTests } from '../../utils/utils.test-utils'
 
 describe('guaranteeUniqueUids', () => {
   it('if two siblings have the same ID, one will be replaced', () => {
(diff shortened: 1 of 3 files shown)


editor/src/core/model/element-template-utils.spec.tsx
```diff
@@ -739,3 +749,245 @@ describe('rearrangeJsxChildren', () => {
...
+    expectElementAtPathHasMatchingUIDForPaths(projectFile, [
+      'utopia-storyboard-uid/scene-aaa/app-entity:aaa',
+      'utopia-storyboard-uid/scene-aaa/app-entity:aaa/parent',
+      'utopia-storyboard-uid/scene-aaa/app-entity:aaa/parent/ca0',
+    ])
+
+    // !!! what to do with the then-case and else-case "uid" behaviors? how do we proceed from here?
+
+    expectElementFoundNull(projectFile, [
+      'utopia-storyboard-uid/scene-aaa/app-entity:aaa/parent/ca0/then-case',
```
balazsbajorics, line 989: ⚠️ When asking for `aaa/parent/ca0/then-case`, findJSXElementChildAtPath will return the conditional itself at `aaa/parent/ca0`!!

I believe this is not the way it should be. I think if we target either branch of the conditional _and_ the branch doesn't have a JSXElementChild to return, we should return null. 

@ruggi do you have any thoughts about this? 

Zooming out, the then-case and else-case code has been growing in the last week with new navigator features and lookup functions. Is the behavior internally consistent? What would happen if findJSXElementChildAtPath starts returning (I believe correctly) null for the cases when the element at the then-case or else-case is, in fact, NOT a jsxElementChild?

editor/src/core/model/element-template-utils.spec.tsx
```diff
@@ -739,3 +749,245 @@ describe('rearrangeJsxChildren', () => {
...
+      `),
+    )
+
+    expectElementAtPathHasMatchingUIDForPaths(projectFile, [
+      'utopia-storyboard-uid/scene-aaa/app-entity:aaa',
+      'utopia-storyboard-uid/scene-aaa/app-entity:aaa/parent',
+      'utopia-storyboard-uid/scene-aaa/app-entity:aaa/parent/child-a',
+    ])
+
+    // !!!! This is a deliberately failing test to demonstrate the issue Berci found
```
balazsbajorics, line 839: I wanted to commit the breaking test first, but I have a one-liner fix to this problem.

editor/src/core/model/element-template-utils.spec.tsx
ruggi: > When asking for aaa/parent/ca0/then-case, findJSXElementChildAtPath will return the conditional itself at aaa/parent/ca0
this should definitely not be the case :)

bkrmendy (approved):

editor/src/core/model/element-template-utils.spec.tsx
```diff
@@ -739,3 +749,301 @@ describe('rearrangeJsxChildren', () => {
...
+            <div data-uid='child-a' />
+          </div>
+        </div>
+      `),
+    )
+
+    expectElementAtPathHasMatchingUIDForPaths(projectFile, [
+      'utopia-storyboard-uid/scene-aaa/app-entity:aaa',
+      'utopia-storyboard-uid/scene-aaa/app-entity:aaa/parent',
+      'utopia-storyboard-uid/scene-aaa/app-entity:aaa/parent/ca0',
```
seanparsons, line 951: I'd recommend using `FOR_TESTS_setNextGeneratedUids` or similar to make the conditional expression have a fixed UID.

seanparsons (approved):

editor/src/core/model/element-template-utils.spec.tsx
balazsbajorics: fixed!

PR #3426, opened 2023-03-13
Drag in and out of conditional clauses.

seanparsons:
**Problem:**
Users will want to drag in and out of conditional clauses as if they were just regular children within the hierarchy.

**Fix:**
There are two main tranches to the changes to support this kind of change:
- Changing some of our reparenting logic to handle the case where the target location to be reparented into can be a clause in a conditional as well as a regular element path. This has required some small changes in the strategies support code as that currently handles all reparenting.
- Tweaks to the navigator drag-and-drop handling which drives the actual reparenting, expanding the existing code to cater for the conditional clause entries.

**Limitations:**
This does not handle the case of dragging into a clause which already has an element within it, primarily to keep the scope of this work to the minimal amount possible.

**Commit Details:**
- Added `ReparentTargetParent` to cover reparenting to an `ElementPath` or into a conditional clause.
- Implemented some support functions for `ReparentTargetParent`, like `getElementPathFromReparentTargetParent` to help integrate it into existing code.
- Altered `getReparentOutcome` to use a `targetParent` of `ReparentTargetElement`.
- `ReparentElement` command now uses `ReparentTargetParent`.
- `editorMoveMultiSelectedTemplates` uses `ReparentTargetParent` for the `targetParent`.
- Created `ConditionalClause` type which can be used to "target" a specific clause in a non-navigator specific way.
- Renamed `displayAtElementPath` to `displayAtEntry` and `moveToElementPath` to `moveToEntry`.
- Renamed `getDragSelections` to `getCurrentlySelectedEntries` as that more accurately describes what it now does.
- Renamed `canDrop` function to `notDescendant`, so that a new function called `canDrop` can use the name.
- `canDrag` in `NavigatorItemContainer` allows for synthetic navigator entries to be dragged.
- `canBeReparentedIntoSelector` renamed to `elementSupportsChildrenSelector` and a new value of `canReparentInto` uses the result of that and a check for conditional clauses into account.
- Incorporated fix from PR #3412 for `findJSXElementChildAtPath`, because it was necessary for this PR to work correctly.
- `insertJSXElementChild` now caters for inserting into conditional clauses.

### DIFF
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -603,8 +603,8 @@ export const DefaultTheme: ThemeSetting = 'system'
 export type DropTargetType = 'before' | 'after' | 'reparent' | null
 
 export interface DropTargetHint {
-  displayAtElementPath: NavigatorEntry | null
-  moveToElementPath: NavigatorEntry | null
+  displayAtEntry: NavigatorEntry | null
+  moveToEntry: NavigatorEntry | null
   type: DropTargetType
 }
 
@@ -1965,12 +1965,15 @@ export function removeElementAtPath(
 export function insertElementAtPath(
   projectContents: ProjectContentTreeRoot,
   openFile: string | null,
-  targetParent: ElementPath | null,
+  targetParent: ReparentTargetParent<ElementPath> | null,
   elementToInsert: JSXElementChild,
   components: Array<UtopiaJSXComponent>,
   indexPosition: IndexPosition | null,
 ): Array<UtopiaJSXComponent> {
-  const staticTarget = targetParent == null ? null : EP.dynamicPathToStaticPath(targetParent)
+  const staticTarget =
+    targetParent == null
+      ? null
+      : dynamicReparentTargetParentToStaticReparentTargetParent(targetParent)
   return insertJSXElementChild(
     projectContents,
     openFile,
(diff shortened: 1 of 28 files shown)


editor/src/components/editor/store/editor-state.ts
```diff
@@ -2104,12 +2107,62 @@ export function regularNavigatorEntriesEqual(
   return EP.pathsEqual(first.elementPath, second.elementPath)
 }
 
-export interface ConditionalClauseNavigatorEntry {
-  type: 'CONDITIONAL_CLAUSE'
-  elementPath: ElementPath
+export interface ConditionalClause<P extends ElementPath> {
```
Rheeseyb, line 2110: Why do we need `P extends ElementPath` here?

Rheeseyb: Oh is this needed because of dynamic and static element paths?

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -77,8 +80,34 @@ export interface NavigatorItemDragAndDropWrapperProps {
...
+): boolean {
+  return !EP.isDescendantOfOrEqualTo(draggedOnto.navigatorEntry.elementPath, draggedItem)
+}
+
+function canDrop(
+  editorState: EditorState,
+  draggedItem: NavigatorItemDragAndDropWrapperProps,
+  draggedOnto: NavigatorItemDragAndDropWrapperProps,
+): boolean {
+  // TODO: What is this checking?
```
Rheeseyb, line 95: This is checking that the drop target can accept children if the drag will result in a reparent onto that target

editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
```diff
@@ -173,6 +182,66 @@ export var ${BakedInStoryboardVariableName} = (
...
+          [].length === 0 ? null : null
+        ) : (
+          <div
+            style={{
+              height: 150,
+              position: 'absolute',
+              left: 154,
+              top: 134,
+            }}
+            data-uid='else-div'
```
Rheeseyb, line 222: Can you rename the uses of "then" and "else" to match up with #3428 please

editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
```diff
@@ -186,14 +255,135 @@ function navigatorStructure(editorState: EditorState, deriveState: DerivedState)
...
   setFeatureForBrowserTests('Conditional support', true)
   setFeatureForBrowserTests('Fragment support', true)
-  it('dragging into a non-empty clause, creates a fragment wrapper', async () => {
+  xit('dragging into a non-empty clause, creates a fragment wrapper', async () => {
     // TODO: Fill this out.
   })
-  it('dragging into an empty clause, creates a fragment wrapper', async () => {
+  it('dragging into an empty active clause, takes the place of the empty value', async () => {
     FOR_TESTS_setNextGeneratedUids([
+      'conditional2',
```
Rheeseyb, line 266: What's the deal with this huge list of alternating UIDs? This strikes me as something that's going to be a nightmare to maintain, because it's not clear which interactions below will be triggering new UID generation (and how many new UIDs are needed each time).

editor/src/components/editor/store/editor-state.ts
seanparsons: So that we can have static ones as well as regular ones.

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
seanparsons: Doh, this was a reminder to myself to check this!

editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
seanparsons: Seems pretty inconsequential given this is test data surely?

editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
seanparsons: Without spending a bunch of time getting them down to exactly the number of entries needed I mostly did this for expedience, especially as I was using the tests from the start. It's just a bit awkward because it's not super intuitive how many you do need without doing it by trial and error.

seanparsons: Improved these in 154326c26.

gbalint (approved):

editor/src/components/navigator/navigator-utils.ts
```diff
@@ -75,7 +75,7 @@ export function navigatorDepth(
     isFeatureEnabled('Conditional support') &&
     isConditionalClauseNavigatorEntry(navigatorEntry)
   ) {
-    result = result - 1
+    result = result + 1
```
Rheeseyb, line 78: ???

Rheeseyb (approved):

editor/src/components/navigator/navigator-utils.ts
seanparsons: Ah, because there's a tweak I made related to the conditional clause entry not having the path down into the clause anymore:
https://github.com/concrete-utopia/utopia/pull/3426/files#diff-9927081e43e2bc863911820ed9808e422cbd44f73d7ef5229db0b2999d073d77R152

PR #3428, opened 2023-03-14
Rename conditional cases

gbalint:
**Problem:**
We started to use two distinct terminology for conditional cases: then/else and true/false.

**Fix:**
Unify the terminology by only using `true-case` and `false-case`.

### DIFF
--- a/editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
+++ b/editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
@@ -396,13 +396,13 @@ describe('conditionals in the navigator', () => {
     ).toEqual(`  regular-utopia-storyboard-uid/scene-aaa
     regular-utopia-storyboard-uid/scene-aaa/containing-div
       regular-utopia-storyboard-uid/scene-aaa/containing-div/conditional1
-        conditional-clause-utopia-storyboard-uid/scene-aaa/containing-div/conditional1-then
+        conditional-clause-utopia-storyboard-uid/scene-aaa/containing-div/conditional1-true-case
           regular-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2
-            conditional-clause-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2-then
+            conditional-clause-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2-true-case
               regular-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2/then-then-div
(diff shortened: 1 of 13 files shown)


editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
```diff
@@ -220,8 +220,8 @@ xdescribe('conditionals in the navigator', () => {
           regular-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2
             conditional-clause-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2/then-then-div-then
               regular-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2/then-then-div
-            conditional-clause-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2/else-case-else
-              synthetic-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2/else-case-attribute
+            conditional-clause-utopia-storyboard-uid/scene-aaa/containing-div/conditional1/conditional2/false-case-else
```
Rheeseyb, line 223: Hmmm why has this ended up as `false-case-else`, whereas the tests in `ui-jsx-canvas-conditionals.spec.browser2.tsx` have been changed to `false-case-false-case`?

editor/src/core/model/element-template-utils.spec.tsx
```diff
@@ -957,17 +957,17 @@ describe('findJSXElementChildAtPath', () => {
       'utopia-storyboard-uid/scene-aaa/app-entity:aaa/parent/conditional-1/ternary-false-root/ternary-false-child',
     ])
 
-    // !!!! Do I misunderstand something?? shouldn't the then-case and else-case return the true-root and false-root here?? these tests fail now...
+    // !!!! Do I misunderstand something?? shouldn't the true-case and false-case return the true-root and false-root here?? these tests fail now...
```
Rheeseyb, line 960: This comment seems to be invalid now...

editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
gbalint: I just missed it and the test is disabled, so it did not break the build.

gbalint: I reverted my changes and added a separate task to fix and enable this test

ruggi (approved):

bkrmendy (approved):

seanparsons (approved):

PR #3429, opened 2023-03-14
Fix: insert into conditional's parent

balazsbajorics:
**Problem:**
Inserting into the parent element of a conditional expression would often wrongly default to Flow insertion.

**Fix:**
Instead of extending the special-cased support for fragments and conditionals in `findValidTargetsUnderPoint` and other reparent utility functions, I added `specialSizeMeasurements.position` to the list of things a non-dom element "inherits" from its children.

**Commit Details:**
- Tests!
- Removing special cased code!
- Adding position to be one of those specialSizeMeasurements that comes from the children!

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -2020,6 +2020,9 @@ function fillSpyOnlyMetadata(
     return parentLayoutSystem == null
   })
 
+  elementsWithoutParentData.sort()
+  elementsWithoutParentData.reverse()
+
   fastForEach(elementsWithoutParentData, (pathStr) => {
     const spyElem = fromSpy[pathStr]
     const sameThingFromWorkingElems = workingElements[pathStr]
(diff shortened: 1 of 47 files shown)


editor/src/core/shared/element-template.ts
```diff
@@ -1114,6 +1114,12 @@ export function isJSXElementLike(element: JSXElementChild): element is JSXElemen
   return isJSXElement(element) || isJSXFragment(element)
 }
 
+export type JSXNonDomElement = JSXFragment | JSXConditionalExpression
+
+export function isNonDomElement(element: JSXElementChild): element is JSXNonDomElement {
```
balazsbajorics: as part of the original fix I added this, but then we ended up not using this. The same thing happened two weeks ago when Berci's PR added a function similar to this that we ended up not using and removing. 

I now wonder if this should be considered a code smell if in the future we end up needing this helper function.

bkrmendy (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2020,6 +2020,9 @@ function fillSpyOnlyMetadata(
     return parentLayoutSystem == null
   })
 
+  elementsWithoutParentData.sort()
+  elementsWithoutParentData.reverse()
```
Rheeseyb, line 2024: But why?!

Rheeseyb (approved):
Approved on the conditional (😎) that you add a comment to the sort and reverse lines please

PR #3431, opened 2023-03-14
fix(inspector) keyboard strategies should update the inspector

enidemi:
**Fix:**
Keyboard absolute strategies should update the inspector during interaction.

**Commit Details:**
- update the low-priority store when a keyboard absolute strategy is active

### DIFF
--- a/editor/src/components/inspector/inspector.tsx
+++ b/editor/src/components/inspector/inspector.tsx
@@ -72,7 +72,10 @@ import { ElementPath, PropertyPath } from '../../core/shared/project-file-types'
 import { when } from '../../utils/react-conditionals'
 import { createSelector } from 'reselect'
 import { isTwindEnabled } from '../../core/tailwind/tailwind'
-import { isStrategyActive } from '../canvas/canvas-strategies/canvas-strategies'
+import {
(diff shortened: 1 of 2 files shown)


seanparsons (approved):

bkrmendy (approved):

editor/src/components/inspector/inspector.tsx
```diff
@@ -231,7 +234,10 @@ export function shouldInspectorUpdate(
   strategyState: StrategyState,
   elementsToRerender: ElementsToRerender,
 ): boolean {
-  return !isStrategyActive(strategyState) && elementsToRerender === 'rerender-all-elements'
+  return (
+    (!isStrategyActive(strategyState) && elementsToRerender === 'rerender-all-elements') ||
+    isKeyboardAbsoluteStrategy(strategyState.currentStrategy)
```
Rheeseyb, line 239: I suspect we should be checking for any keyboard strategy here, rather than just absolute move / resize 🤔

enidemi: I skipped keyboard reorder strategy now because it doesn't change anything in the inspector sections

PR #3433, opened 2023-03-14
Extend group-like support to conditionals

balazsbajorics:
**Problem:**
We had a half-official strategy support for conditionals. 

**Fix:**
Now they are included in the `ContentAffectingType` enumeration!
This meant that I had to extend the two existing content-affecting test suites, the absolute reparent and the flex-> absolute reparent. As part of fixing these tests, I actually found two bugs in the code which I also amended.

**Commit Details:**
- Added 'conditional' to ContentAffectingType
- Fixed 'children-affecting reparent tests'
- Fixed 'Flex Reparent to Absolute – children affecting elements' tests
- Fixed setUtopiaID to work with conditional expressions
- Fixed renderJSXElement to not put a style tag on display none / hidden fragments to avoid throwing a big noisy warning

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.spec.browser2.tsx
@@ -29,6 +29,7 @@ import { selectComponents } from '../../../editor/actions/meta-actions'
 import * as EP from '../../../../core/shared/element-path'
 import { MetadataUtils } from '../../../../core/model/element-metadata-utils'
 import { ContentAffectingType, AllContentAffectingTypes } from './group-like-helpers'
+import { FOR_TESTS_setNextGeneratedUids } from '../../../../core/model/element-template-utils.test-utils'
 
 interface CheckCursor {
   cursor: CSSCursor | null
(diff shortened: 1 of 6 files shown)


ruggi (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -548,6 +539,18 @@ function renderJSXElement(
 
   const FinalElement = elementIsIntrinsic ? jsx.name.baseVariable : elementOrScene
   const FinalElementOrFragment = elementIsFragment ? React.Fragment : FinalElement
+
+  let elementProps = { key: key, ...passthroughProps }
+  if (!elementIsFragment) {
```
Rheeseyb, line 544: Nice

editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.spec.browser2.tsx
```diff
@@ -1175,9 +1199,11 @@ describe('children-affecting reparent tests', () => {
 function getOpeningTag(type: ContentAffectingType): string {
   switch (type) {
     case 'sizeless-div':
-      return `<div data-uid='children-affecting' data-testid='children-affecting'>`
+      return `<div data-uid='children-affecting' data-testid='children-affecting'><>`
     case 'fragment':
-      return `<React.Fragment data-uid='children-affecting' data-testid='children-affecting'>`
+      return `<React.Fragment data-uid='children-affecting' data-testid='children-affecting'><>`
+    case 'conditional':
+      return `{ true ? ( <>`
```
Rheeseyb, line 1206: Why are these now also including a (shorthand) fragment?

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.spec.browser2.tsx
balazsbajorics: because the test code has two siblings as children of this element, which is possible for divs and fragments, but not for conditional root elements. 

actually the nested fragment is not a bad thing anyways because we uncovered some bugs with nested groups in the last week

PR #3436, opened 2023-03-15
Edit conditional expressions in the inspector

### LINKED ISSUE #3435: Edit conditional expressions in the inspector pane

ruggi:
Fixes #3435 

This PR adds the ability to inline-edit conditional expressions via the inspector.

https://user-images.githubusercontent.com/1081051/225343429-e9640374-f81e-44d2-840c-2d8f7bc4aaa0.mp4

The `condition` of the `JSXConditionalElement` is set by creating a `JSXAttributeOtherJavascript` with the verbatim string entered in the text input, which then triggers a reparse.

### DIFF
--- a/editor/src/components/inspector/sections/layout-section/conditional-section.tsx
+++ b/editor/src/components/inspector/sections/layout-section/conditional-section.tsx
@@ -34,6 +34,7 @@ import {
 import { useDispatch } from '../../../editor/store/dispatch-context'
 import { Substores, useEditorState } from '../../../editor/store/store-hook'
 import { MetadataSubstate } from '../../../editor/store/store-hook-substore-types'
+import { usePropControlledStateV2 } from '../../common/inspector-utils'
 import { UIGridRow } from '../../widgets/ui-grid-row'
 
 export const ConditionalsControlSectionOpenTestId = 'conditionals-control-section-open'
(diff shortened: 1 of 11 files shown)


editor/src/third-party/react-error-overlay/components/CodeBlock.tsx
```diff
@@ -30,7 +30,7 @@ const secondaryPreStyle = {
   backgroundColor: yellowTransparent,
 }
 
-const codeStyle = {
+export const codeStyle = {
```
balazsbajorics, line 33: I think this should not be exported as this is source code of react-error-overlay, and instead, if there's no relevant entry, add one to `UtopiaStyles`

editor/src/components/inspector/sections/layout-section/conditional-section.tsx
```diff
@@ -126,12 +128,18 @@ export const ConditionalSection = React.memo(({ paths }: { paths: ElementPath[]
...
     (store) => conditionExpressionSelector(store, paths),
     'ConditionalSection condition expression',
   )
 
+  const [conditionExpression, setConditionExpression] = React.useState(originalConditionExpression)
+
+  React.useEffect(() => {
+    setConditionExpression(originalConditionExpression)
+  }, [originalConditionExpression])
+
```
balazsbajorics, line 142:
```suggestion
  const [conditionExpression, setConditionExpression] = usePropControlledStateV2(originalConditionExpression)
```

I believe usePropControlledStateV2 achieves exactly this behavior, but without an extra re-render penalty, and no useEffect

editor/src/third-party/react-error-overlay/components/CodeBlock.tsx
ruggi: 👍

editor/src/components/inspector/sections/layout-section/conditional-section.tsx
ruggi: oh that's very nice, thank you for the heads up 👌

ruggi: https://github.com/concrete-utopia/utopia/pull/3436/commits/d8d4b5ad806be19b40c4f403e1d23bb2d51a466a

seanparsons (approved):

PR #3438, opened 2023-03-15
Handle dragging into conditional clauses that have content.

seanparsons:
**Problem:**
We currently handle adding to an empty conditional clause when dragging to them in the navigator, but not those that have existing content.

**Fix:**
When dragging to a conditional clause that has content:
- If it is not a fragment, wrap the existing element and the to be inserted content in a fragment.
- If it is a fragment, insert the new value within that fragment.

**Commit Details:**
- Added logic to `insertJSXElementChild` to handle wrapping with or inserting into existing fragments at the root of a conditional clause.

### DIFF
--- a/editor/src/core/model/element-template-utils.ts
+++ b/editor/src/core/model/element-template-utils.ts
@@ -565,19 +565,46 @@ export function insertJSXElementChild(
                 const simpleValue = jsxSimpleAttributeToValue(clauseValue)
                 return foldEither(
                   () => {
+                    // Not a simple value, so replacing it would be a bad thing.
                     return clauseValue
                   },
                   (value) => {
+                    // Simple value of some kind.
                     if (value == null) {
+                      // Simple value is null, so replace it with the new content.
                       return elementToInsert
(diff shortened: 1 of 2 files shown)


editor/src/core/model/element-template-utils.ts
```diff
@@ -565,19 +565,46 @@ export function insertJSXElementChild(
...
+                      // Recent conversations have been about unifying attributes and arbitrary JSX elements,
+                      // which could result in this content being a valid child of any element.
                       return clauseValue
                     }
                   },
                   simpleValue,
                 )
               } else {
-                return clauseValue
+                if (isJSXFragment(clauseValue)) {
```
gbalint, line 586: Shouldn't we accept other children affecting elements too and not just fragments?

seanparsons: That would be reparenting into those elements though, which is possible to do by dragging into those within the conditional.

gbalint (approved):

PR #3443, opened 2023-03-16
UID comment flag

### LINKED ISSUE #3444: Add a comment flag to set uids of conditional elements

ruggi:
Fixes #3444 

**Problem:**

To set the generated UID of conditional elements (or any other non-dom-elements) during tests we rely on `FOR_TESTS_setNextGeneratedUids` but it has the ugly downside of requiring to set a number of throwaway UIDs.

E.g., this monstrosity:

```ts
FOR_TESTS_setNextGeneratedUids([
  'skip1',
  'skip2',
  'skip3',
  'skip4',
  'skip5',
  'skip6',
  'skip7',
  'skip8',
  'skip9',
  'skip10',
  'skip11',
  'skip12',
  'skip13',
  'skip14',
  'skip15',
  'conditional1',
  'conditional2',
])
```

This is super brittle and _very_ hard to debug if something does not work.

**Fix:**

This PR adds a new comment flag, `@utopia/uid` that allows setting the UID of an element, so that we can get rid of `FOR_TESTS_setNextGeneratedUids` when dealing with conditionals and use instead inline comments.

E.g.:
```ts
{
  // @utopia/uid=conditional1
  [].length === 0 ? // ...
```

Notes:
- this approach would work fine with other non-dom elements as well, but for scope's sake I just hooked it to the conditional elements parsing.
- the new comment flag is just used for tests for now, but it's looking like it can have nice additional uses as well, since it plays along with the current UIDs generation.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-conditionals.spec.browser2.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-conditionals.spec.browser2.tsx
@@ -1,8 +1,6 @@
 import { createModifiedProject } from '../../../sample-projects/sample-project-utils.test-utils'
 import { navigatorEntryToKey, StoryboardFilePath } from '../../editor/store/editor-state'
 import { renderTestEditorWithModel } from '../ui-jsx.test-utils'
-import * as EP from '../../../core/shared/element-path'
-import { FOR_TESTS_setNextGeneratedUids } from '../../../core/model/element-template-utils.test-utils'
 import { setFeatureForBrowserTests } from '../../../utils/utils.test-utils'
 
 const appFilePath = '/src/app.js'
@@ -21,13 +19,16 @@ export var App = (props) => {
         height: '100%'
       }}
     >
-      {[0, 1].length > 1 ? (
-        [0, 1].length === 2 ? (
-          <div data-uid='div-inside-conditionals' />
-        ) : null
-      ) : (
-        5
-      )}
+      {
+        // @utopia/uid=conditional1
+        [0, 1].length > 1 ? (
+            // @utopia/uid=conditional2
+            [0, 1].length === 2 ? (
+            <div data-uid='div-inside-conditionals' />
+          ) : null
+        ) : (
+          5
+        )}
     </div>
   )
 }
@@ -65,20 +66,6 @@ async function createAndRenderProject() {
 describe('a project with conditionals', () => {
   setFeatureForBrowserTests('Conditional support', true)
   it('fills the content of the navigator', async () => {
-    FOR_TESTS_setNextGeneratedUids([
-      'mock1',
-      'mock2',
-      'mock3',
-      'mock4',
-      'conditional1',
-      'conditional2',
-      'mock1',
-      'mock2',
-      'mock3',
-      'mock4',
-      'conditional1',
-      'conditional2',
-    ])
     const renderedProject = await createAndRenderProject()
     const navigatorTargets = renderedProject.getEditorState().derived.visibleNavigatorTargets
     const pathStrings = navigatorTargets.map(navigatorEntryToKey)
--- a/editor/src/components/inspector/sections/layout-section/conditional-section.tsx
+++ b/editor/src/components/inspector/sections/layout-section/conditional-section.tsx
@@ -75,7 +75,7 @@ const conditionOverrideSelector = createCachedSelector(
     elements.forEach((element) => {
       const flag = findUtopiaCommentFlag(element.comments, 'conditional')
       if (isUtopiaCommentFlagConditional(flag)) {
-        conditions.add(flag?.value ?? null)
+        conditions.add(flag.value)
       }
     })
 
--- a/editor/src/components/inspector/sections/layout-section/conditional-section.tsx
+++ b/editor/src/components/inspector/sections/layout-section/conditional-section.tsx
@@ -4,7 +4,10 @@ import createCachedSelector from 're-reselect'
 import React from 'react'
 import { MetadataUtils } from '../../../../core/model/element-metadata-utils'
 import { mapDropNulls } from '../../../../core/shared/array-utils'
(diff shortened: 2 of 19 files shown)


bkrmendy (approved):

editor/src/core/shared/uid-utils.ts
```diff
@@ -91,7 +94,18 @@ const atoz = [
...
+  const commentFlag = deepFindUtopiaCommentFlag(comments ?? null, 'uid')
+  if (commentFlag != null && isUtopiaCommentFlagUid(commentFlag)) {
+    const { value } = commentFlag
+    COMMENT_FLAG_UIDS.current.add(value)
+    return value
+  }
+  if (COMMENT_FLAG_UIDS.current.has(possibleStartingValue)) {
+    return possibleStartingValue
+  }
+
```
balazsbajorics, line 108: is this the right place for this? how do we parse and use the data-uid attribute for elements? I _think_ if an element has a data-uid attribute, we don't even visit the `generateConsistentUID` code path

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-conditionals.spec.browser2.tsx
```diff
@@ -21,13 +19,16 @@ export var App = (props) => {
...
     >
-      {[0, 1].length > 1 ? (
-        [0, 1].length === 2 ? (
-          <div data-uid='div-inside-conditionals' />
-        ) : null
-      ) : (
-        5
-      )}
+      {
+        // @utopia/uid=conditional1
```
Rheeseyb, line 23: So good 😍

seanparsons (approved):

editor/src/core/shared/uid-utils.ts
ruggi: I think that's not the case - that function is used by `getUIDBasedOnElement` which is what we call when generating the jsx conditional expressions

editor/src/components/inspector/sections/layout-section/conditional-section.tsx
```diff
@@ -71,7 +74,9 @@ const conditionOverrideSelector = createCachedSelector(
     let conditions = new Set<boolean | null>()
     elements.forEach((element) => {
       const flag = findUtopiaCommentFlag(element.comments, 'conditional')
-      conditions.add(flag?.value ?? null)
+      if (isUtopiaCommentFlagConditional(flag)) {
```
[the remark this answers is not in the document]
Rheeseyb: Oh so really this is just narrowing the type of `flag.value` (in which case the `flag?.` can become `flag.`)?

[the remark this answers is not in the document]
ruggi: ah yes that's true!

balazsbajorics (approved):

editor/src/components/inspector/sections/layout-section/conditional-section.tsx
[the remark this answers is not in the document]
ruggi: https://github.com/concrete-utopia/utopia/pull/3443/commits/39b332d762293329817f1ce39119d2f51d54bab8

editor/src/core/shared/uid-utils.ts
```diff
@@ -92,6 +95,10 @@ export function generateConsistentUID(
   existingIDs: Set<string>,
   possibleStartingValue: string,
 ): string {
+  if (COMMENT_FLAG_UIDS.current.has(possibleStartingValue)) {
+    return possibleStartingValue
+  }
```
Rheeseyb, line 100: Is this code needed? I don't think it is if we're returning the value in `parseUID` below?

ruggi: prooobably the whole `COMMENT_FLAG_UIDS` is not needed anymore 🧐 running the tests to see if it's true

editor/src/core/shared/element-template.ts
```diff
@@ -1077,7 +1119,7 @@ export function jsxConditionalExpression(
     type: 'JSX_CONDITIONAL_EXPRESSION',
     uid: uid,
     originalConditionString: originalConditionString,
-    condition: condition,
+    condition: fixJSXConditionalExpressionCondition(condition, comments), // remove any duplicate comments shared with the element
```
Rheeseyb, line 1122: Is this optimisation necessary? It seems like this runs the risk of deleting user comments which might have been deliberately duplicated (e.g. `eslint-disable-next-line`)

ruggi: this is only removing duplicate `@utopia/` flags that were passed over between the condition and the conditional element, because of the way comments are parsed. It was an existing issue btw, but it was not surfacing. Fixing this lazily is the best approach I could come up with compared with the alternative of fixing the parsing, making it _much_ more complex

Rheeseyb: Ok in that case I'm fine with it, as long as it's not going to cause issues with a chain of overrides of the form ```// @utopia/conditional=true```

editor/src/core/shared/uid-utils.ts
Rheeseyb: Oh I think the problem might stem from this function (that looks to be very incorrect): <https://github.com/concrete-utopia/utopia/blob/b00117d11559bac9fc4c4301bac95b63700257fc/editor/src/core/shared/uid-utils.ts#L257-L262>

I'm not sure how that `includes` line came to be, but that will surely always return `true` 😀

ruggi: found the problem, sec

ruggi: https://github.com/concrete-utopia/utopia/pull/3443/commits/5244efc5a360189ed4d625db87ef1459d3c5eefd

Rheeseyb (approved):

editor/src/core/shared/uid-utils.ts
Rheeseyb: Thanks for fixing this!

PR #3445, opened 2023-03-16
fix(canvas) keyboard resize shouldn't result in negative width height

enidemi:
**Problem:**
When using keyboard resize it's possible to end up with negative width and height values. It's very visible since the keyboard strategy updates the inspector on all keypress.

**Fix:**
Compare the original size to the accumulated keypress vectors to avoid negative width/height results. Minimum width/height is limited to 1px.

**Commit Details:**
- pass width/height to `pressesToVectorAndEdges`
- added 1 test that uses a tiny element 
- update tests

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-resize.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-resize.spec.browser2.tsx
@@ -214,6 +214,38 @@ describe('Keyboard Absolute Resize E2E', () => {
       height: 101,
     })
   })
+  it('Pressing Shift + Cmd + ArrowLeft 3 times, then pressing Shift + Cmd + ArrowRight once', async () => {
+    const width = 5
+    const {
+      expectElementWidthOnScreen,
+      expectElementPropertiesInPrintedCode,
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-resize.spec.browser2.tsx
```diff
@@ -214,6 +214,38 @@ describe('Keyboard Absolute Resize E2E', () => {
...
+      top: 100,
+      width: width,
+      height: 101,
+    })
+
+    await pressKey('ArrowLeft', { modifiers: shiftCmdModifier })
+    await pressKey('ArrowLeft', { modifiers: shiftCmdModifier })
+    await pressKey('ArrowLeft', { modifiers: shiftCmdModifier })
+
+    expectElementWidthOnScreen(-width)
```
balazsbajorics, line 234: why does pressing left 3 times turn the width into -width?

enidemi: it has a bit of a confusing name, the param is an offset, not the expected value

bkrmendy (approved):

PR #3448, opened 2023-03-16
More Pins and a Flip Icon

lankaukk:
Uploading variations of pin icons, and a new flip icon in anticipation of needing them in the inspector for overrides and switching conditional branches (or other things maybe!)

Note: the former *only* pin icon we had was renamed to a more specific `pinrightoutline`

Screenshot 2023-03-16 at 3 39 21 PM

Screenshot 2023-03-16 at 3 40 06 PM


maltenuhn:
nice! Can you double-check that renaming the `pin` to `pinrightoutline` doesn't cause anything in the `Icons.` object to now be unavailable? I don't remember if we eg had `Icons.Pin` or something like that. 

Also,  whether we referenced the `pin` icon somewhere in our code with without using the `Icon.` object?

Rheeseyb (approved):

PR #3451, opened 2023-03-17
Test for keyboard startegies

bkrmendy:
## Description
Fixes a reorder bug in `keyboard-reorder-strategy` that manifested itself by not updating the navigator order of the siblings being reordered when using the strategy.

Also, this PR adds tests using `ContentAffectingTypes` for all the keyboard strategies.


seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/keyboard-reorder-strategy.ts
```diff
@@ -35,7 +35,7 @@ export function keyboardReorderStrategy(
     return null
   }
   const target = selectedElements[0]
-  const siblings = MetadataUtils.getSiblingsUnordered(canvasState.startingMetadata, target).map(
+  const siblings = MetadataUtils.getSiblingsOrdered(canvasState.startingMetadata, target).map(
```
Rheeseyb, line 38: Aha! So this PR actually contained a fix! Please call that out in the PR description and title, since this is actually now the more important part of the PR wrt actual logic changes

Rheeseyb (approved):

PR #3453, opened 2023-03-17
Feat/paste into conditionals

### LINKED ISSUE #3452: Allow pasting into conditionals

ruggi:
Fixes #3452 

This PR allows to paste into conditional elements.

The paste will happen into the active branch of the conditional (whether it's overridden or not).

The paste behavior is as follows:
- If the target branch is `null`, its replaced with the pasted element
- If the target branch is a `JSXElement` or an attribute, the existing branch is wrapped in a new fragment and the pasted element is appended to it
- If the target branch is a fragment, the pasted element is appended to it

https://user-images.githubusercontent.com/1081051/225950056-411487a2-77d1-498c-b96d-116bc2f48361.mp4

### DIFF
--- a/editor/src/core/model/element-template-utils.ts
+++ b/editor/src/core/model/element-template-utils.ts
@@ -628,67 +628,65 @@ export function insertJSXElementChild(
           children: updatedChildren,
         }
       } else if (isJSXConditionalExpression(parentElement)) {
-        const trueCase = getConditionalCasePath(parentPath, 'true-case')
-        const parentMetadata = spyMetadata[EP.toString(parentPath)] ?? null
-        const isTrueCase =
-          getConditionalCase(trueCase, parentElement, parentMetadata, parentPath) === 'true-case'
-        let branch = isTrueCase ? { ...parentElement.whenTrue } : { ...parentElement.whenFalse }
-        switch (branch.type) {
-          case 'ATTRIBUTE_VALUE':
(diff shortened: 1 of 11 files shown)


seanparsons (approved):

editor/src/core/model/element-template-utils.ts
```diff
@@ -543,92 +548,151 @@ export function insertJSXElementChild(
...
+          ...parentElement,
+          children: updatedChildren,
+        }
+      } else if (isJSXConditionalExpression(parentElement)) {
+        const trueCase = getConditionalCasePath(parentPath, 'true-case')
+        const parentMetadata = spyMetadata[EP.toString(parentPath)] ?? null
+        const isTrueCase =
+          getConditionalCase(trueCase, parentElement, parentMetadata, parentPath) === 'true-case'
+        let branch = isTrueCase ? { ...parentElement.whenTrue } : { ...parentElement.whenFalse }
+        switch (branch.type) {
```
Rheeseyb, line 636: Rather than mutating `branch`, please can you extract this into a function

gbalint (approved):

editor/src/core/model/element-template-utils.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/3453/commits/248e4598c3e068fc007ede6be6553839830571f0

PR #3454, opened 2023-03-17
Merge two big arbitrary representations.

seanparsons:
**Problem:**
Currently there are two types which represent very similar concepts of arbitrary JavaScript producing a value, one for attributes and another for JSX content.

**Fix:**
The very beginning of this change revolves around changing:
```typescript
export interface JSXArbitraryBlock {
  type: 'JSX_ARBITRARY_BLOCK'
  originalJavascript: string
  javascript: string
  transpiledJavascript: string
  definedElsewhere: Array<string>
  sourceMap: RawSourceMap | null
  uniqueID: string
  elementsWithin: ElementsWithin
}
```
To:
```typescript
export type JSXArbitraryBlock = JSExpression
```
Where `JSExpression` is just `JSXAttribute` renamed.

With this every case of `JSExpression` now has a `uniqueID` property so that they can potentially be addressed.

**Commit Details:**
- Switched some test data back to the consistent and generated UIDs as the mock generated UIDs are even more difficult to wrangle now that the properties are getting unique IDs as well.
- Refactored `getNavigatorEntryLabel` a little as a result of the type changes.
- Added `clearInsertableComponentUniqueIDs` and `clearInsertableComponentGroupUniqueIDs`.
- Removed `getConditionalCasePath` as it is unnecessary now.
- Fixed up `transformAtPathOptionally`.
- Reworked `findJSXElementChildAtPath` to handle the cases for conditional clauses by treating them more like children.
- `JSXAttributeValue` renamed to `JSExpressionValue`.
- `JSXAttributeOtherJavaScript` renamed to `JSExpressionOtherJavaScript`.
- `JSXAttributeNestedObject` renamed to `JSExpressionNestedObject`.
- `JSXAttributeNestedArray` renamed to `JSExpressionNestedArray`.
- `JSXAttributeFunctionCall` renamed to `JSExpressionFunctionCall`.
- `JSXAttribute` renamed to `JSExpression`.
- Renamed the functions that check what case of `ModifiableAttribute` a value is to be prefixed with `modifiableAttributeIs`.
- Added `clearJSXElementWithoutUIDUniqueIDs`.
- `JSXArbitraryBlock` is now a type alias to `JSExpression`.
- `isJSXArbitraryBlock` now refers to the 5 possible cases related to what it used to be.
- Removed `ChildOrAttribute`, `childOrBlockIsChild` and `childOrBlockIsAttribute`.
- Added `clearModifiableAttributeUniqueIDs`.
- Updated `fixUtopiaElement` and `findElementWithUID`.
- Added `isUtopiaJSExpressionOtherJavaScript`, `isUtopiaJSExpressionValue`, `isUtopiaJSExpressionNestedArray`, `isUtopiaJSExpressionNestedObject` and `isUtopiaJSExpressionFunctionCall`.
- Added some utility functions for creating expression values and their associated UIDs during a parse.
- Fixed a bug in `parseClause` which was causing the code to think UIDs had been duplicated, as it was running `innerParse` and `parseAttribute` but only using the result from one.
- Quite a few minor changes to switch/case statements to handle the change in types of `JSXElementChild`.

### DIFF
--- a/editor/src/core/shared/element-template.ts
+++ b/editor/src/core/shared/element-template.ts
@@ -17,7 +17,7 @@ import { v4 as UUID } from 'uuid'
 import { RawSourceMap } from '../workers/ts/ts-typings/RawSourceMap'
 import * as PP from './property-path'
 import { Sides, sides, LayoutSystem } from 'utopia-api/core'
-import { fastForEach, unknownObjectProperty } from './utils'
+import { assertNever, fastForEach, unknownObjectProperty } from './utils'
 import { addAllUniquely, mapDropNulls, reverse } from './array-utils'
 import { objectMap } from './object-utils'
 import { CSSPosition, FlexDirection } from '../../components/inspector/common/css-utils'
@@ -1230,13 +1230,21 @@ export function isJSExpressionOtherJavaScript(
 }
 
 export function isJSXArbitraryBlock(element: JSXElementChild): element is JSXArbitraryBlock {
-  return (
-    isJSExpressionValue(element) ||
-    isJSExpressionNestedArray(element) ||
-    isJSExpressionNestedObject(element) ||
-    isJSExpressionFunctionCall(element) ||
-    isJSExpressionOtherJavaScript(element)
-  )
+  switch (element.type) {
+    case 'JSX_ELEMENT':
+    case 'JSX_TEXT_BLOCK':
+    case 'JSX_FRAGMENT':
+    case 'JSX_CONDITIONAL_EXPRESSION':
+      return false
+    case 'ATTRIBUTE_VALUE':
+    case 'ATTRIBUTE_OTHER_JAVASCRIPT':
+    case 'ATTRIBUTE_NESTED_ARRAY':
+    case 'ATTRIBUTE_NESTED_OBJECT':
+    case 'ATTRIBUTE_FUNCTION_CALL':
+      return true
+    default:
+      assertNever(element)
+  }
 }
 
 export function isJSXTextBlock(element: JSXElementChild): element is JSXTextBlock {


editor/src/core/shared/element-template.ts
```diff
@@ -1994,32 +2039,16 @@ export function getElementsByUIDFromTopLevelElements(
...
-    case 'ATTRIBUTE_NESTED_ARRAY':
-    case 'ATTRIBUTE_NESTED_OBJECT':
-    case 'ATTRIBUTE_FUNCTION_CALL':
-    case 'ATTRIBUTE_OTHER_JAVASCRIPT':
-      return false
-    default:
-      const _exhaustiveCheck: never = childOrBlock
-      throw new Error(`Unhandled type ${JSON.stringify(childOrBlock)}`)
-  }
+  return true
```
Rheeseyb, line 2047: 😀 I guess this is a `TODO`

seanparsons: It can be bulldozed, I just kept it in so that I didn't have to delete it from a bunch of places for now.

editor/src/core/shared/element-template.ts
```diff
@@ -992,19 +1035,40 @@ export type JSXConditionalExpressionWithoutUID = Omit<JSXConditionalExpression,
...
+): JSXFragmentWithoutUID {
+  return {
+    ...fragment,
+    children: fragment.children.map(clearJSXElementUniqueIDs),
+  }
 }
 
+export type ElementsWithin = { [uid: string]: JSXElement }
+
+export type JSXArbitraryBlock = JSExpression
```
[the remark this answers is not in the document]
seanparsons: Mostly convenience, I can remove it, it just means the PR will grow a bit more.

gbalint (approved):

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -553,20 +553,14 @@ describe('Select Mode Double Clicking With Fragments', () => {
...
-      'cardlistfragment',
-      'manuallistfragment',
-    ])
     // prettier-ignore
     const desiredPaths = createConsecutivePaths(
       'sb' +                  // Skipped as it's the storyboard
       '/scene-CardList' +     // Skipped because we skip over Scenes
       '/CardList-instance' +  // Skipped because we skip component children of Scenes
-      ':manuallistfragment' + // Skipped because we skip over root elements
+      ':38e' +                // Skipped because we skip over root elements
```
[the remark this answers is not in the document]
balazsbajorics: sorry, I missed this! right. makes sense. I think the long term fix is to make the `// @utopia/uid=` comment work for fragments too, and then we can explicitly annotate the test code with fragment uids

editor/src/core/shared/project-file-types.ts
```diff
@@ -589,18 +589,20 @@ export function codeFile(
   )
 }
 
-export function isTextFile(projectFile: ProjectFile | null): projectFile is TextFile {
-  return projectFile != null && projectFile.type === 'TEXT_FILE'
+export function isTextFile(projectFile: ProjectFile): projectFile is TextFile {
```
balazsbajorics, line 592: is this just making the type nicer? or was this necessitated somehow?

seanparsons: Ah, it was causing grief with the optics composition.

editor/src/components/inspector/common/longhand-shorthand-hooks.spec.tsx
```diff
@@ -299,7 +299,7 @@ describe('useInspectorInfo: updating padding shorthand and longhands', () => {
           setProp_UNSAFE(
             TestSelectedComponent,
             PP.create('style', 'paddingLeft'),
-            jsxAttributeValue(100, emptyComments),
+            jsExpressionValue(100, emptyComments, expect.stringContaining('')),
```
balazsbajorics, line 302: nice!

editor/src/core/model/conditionals.ts
```diff
@@ -16,24 +14,13 @@ import { findUtopiaCommentFlag, isUtopiaCommentFlagConditional } from '../shared
...
-): ElementPath {
-  return EP.appendToPath(elementPath, conditionalCase)
-}
-
 // Get the path for the clause (true case or false case) of a conditional.
 export function getConditionalClausePath(
   conditionalPath: ElementPath,
-  conditionalClause: ChildOrAttribute,
+  conditionalClause: JSXElementChild,
   conditionalCase: ConditionalCase,
```
balazsbajorics, line 21: I think with this PR merged we should make a quick follow-up PR that erases the concept of the ConditionalClausePath

editor/src/core/model/element-template-utils.ts
```diff
@@ -765,6 +744,18 @@ export function elementOnlyHasTextChildren(element: JSXElementChild): boolean {
...
+      return element.content.every((contentElement) => {
+        return elementOnlyHasTextChildren(contentElement.value)
+      })
+    case 'ATTRIBUTE_NESTED_OBJECT':
+      return element.content.every((contentElement) => {
+        return elementOnlyHasTextChildren(contentElement.value)
+      })
+    case 'ATTRIBUTE_FUNCTION_CALL':
+      return allElementsAndChildrenAreText(element.parameters)
     default:
```
balazsbajorics, line 759: intriguing! does this have an effect on the text editing capabilites?

editor/src/core/model/element-template-utils.ts
```diff
@@ -765,6 +744,18 @@ export function elementOnlyHasTextChildren(element: JSXElementChild): boolean {
...
+    case 'ATTRIBUTE_NESTED_ARRAY':
+      return element.content.every((contentElement) => {
+        return elementOnlyHasTextChildren(contentElement.value)
+      })
+    case 'ATTRIBUTE_NESTED_OBJECT':
+      return element.content.every((contentElement) => {
+        return elementOnlyHasTextChildren(contentElement.value)
+      })
+    case 'ATTRIBUTE_FUNCTION_CALL':
+      return allElementsAndChildrenAreText(element.parameters)
```
balazsbajorics, line 758: hmmm this sounds like a stretch to check if all parameters are text here. shouldn't this just return false?

seanparsons: Yeah, you're probably right there actually.

editor/src/core/shared/element-template.ts
```diff
@@ -1137,8 +1201,42 @@ export function isJSXElement(element: JSXElementChild): element is JSXElement {
...
+
 export function isJSXArbitraryBlock(element: JSXElementChild): element is JSXArbitraryBlock {
-  return element.type === 'JSX_ARBITRARY_BLOCK'
+  return (
+    isJSExpressionValue(element) ||
+    isJSExpressionNestedArray(element) ||
+    isJSExpressionNestedObject(element) ||
+    isJSExpressionFunctionCall(element) ||
+    isJSExpressionOtherJavaScript(element)
+  )
```
balazsbajorics, line 1239: I'm afraid without an exhaustive check here we will forget to extend this function. the only way I can imagine the exhaustive check though if we go with a helper function like `isTransientAction()` which enumerates _all_ possible JSXElementChild types and returns an explicit true or false.

seanparsons: Yeah, this made sense at the time as I had those guards to hand, but it will definitely cause trip ups in the future.

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1115,6 +1122,92 @@ function parseJSXArbitraryBlock(
...
+  sourceFile: TS.SourceFile,
+  value: any,
+  comments: ParsedComments,
+  alreadyExistingUIDs: Set<string>,
+): JSExpressionValue<any> {
+  const uid = generateUIDAndAddToExistingUIDs(sourceFile, value, alreadyExistingUIDs)
+  return jsExpressionValue(value, comments, uid)
+}
+
+function createExpressionOtherJavaScript(
```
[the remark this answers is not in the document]
seanparsons: New code, we were invoking the functions like `jsExpressionValue` individually all over the parser, but we want these to have consistent short uniqueIDs like the elements have. So I made these utility functions that handle the UID part and replaced the calls to the original functions with these.

editor/src/core/workers/parser-printer/parser-printer-utils.spec.ts
```diff
@@ -52,18 +52,18 @@ describe('guaranteeUniqueUidsFromTopLevel', () => {
       jsxElement(
         'View',
         'root',
-        jsxAttributesFromMap({ 'data-uid': jsxAttributeValue('root', emptyComments) }),
+        jsxAttributesFromMap({ 'data-uid': jsExpressionValue('root', emptyComments, 'root') }),
```
balazsbajorics, line 55: wait a minute! if the unique ID of root is root. the unique ID of the string setting data-uid to root cannot be root!! because these two uids share the same global uid space, right?

editor/src/core/model/element-template-utils.ts
seanparsons: Fixed in d80aab8e9.

editor/src/core/model/element-template-utils.ts
seanparsons: Now removed with d80aab8e9.

editor/src/core/shared/element-template.ts
seanparsons: Fixed in 8ec7d8e.

editor/src/core/workers/parser-printer/parser-printer-utils.spec.ts
seanparsons: Very true, so I would expect it to be update I think but the test in this case doesn't validate that, just the siblings.

balazsbajorics (approved):

PR #3458, opened 2023-03-20
Review what target elements we are passing down as strategy control props

bkrmendy:
## Problem
Strategy controls were originally designed to work with elements that are present in the rendered DOM. With the introduction of fragments, conditionals, and sizeless divs (content-affecting elements), and the retargeting of some strategies to the children of content-affecting elements, we have to review if strategy controls are rendered based on the originally selected elements, or the retargeted targets. 

In this PR, retargeted strategies are checked (i.e., the strategies that use `replaceContentAffectingPathsWithTheirChildrenRecursive`), and strategies that use a control that is also used by one of the retargeted strategies.

#### `ImmediateParentOutlines` and `ImmediateParentBounds`

`ImmediateParentOutlines` and `ImmediateParentBounds` only render based on their props if

1. the editor is in select mode
2. all target paths from props have the same parent path according to `EP.parentPath`
3. This common parent is not the storyboard

The assumption from 2 is broken when targets are provided by `replaceContentAffectingPathsWithTheirChildrenRecursive` in a multiselection involving content-affecting elements since the parent element will be the content-affecting elements, which are not the same.

Based on this `ImmediateParentBounds` and `ImmediateParentOutlines` should always be passed the original, non-retargeted targets.

#### `AbsoluteResizeControl`
`AbsoluteResizeControl` renders based on the bounding boxes of the elements passed down through props. Since content-affecting elements are sized to the bounding box of their children's BBs, it doesn't matter if the control is retargeted or not, so for the sake of consistency, it should be passed the original targets.

#### `ZeroSizeResizeControlWrapper`
`ZeroSizeResizeControlWrapper` only renders for elements zero-sized with non-null global frames. The situation in the case of content-affecting elements is the following:
- if a content-affecting element has no children, its `globalFrame` will be set to `infinityCanvasRectangle`, so the zero-size control won't render
- if a content-affecting element has children, its `globalFrame` will be set to the bounding box of its children's global frames, which can be zero-sized, which is a correct bounding box.

Based on this, the control will render correctly either way if the original (potentially content-affecting) targets are passed to it, so for the sake of consistency, it should be passed the original targets.

## Commit details
- clarified original targets vs selected targets naming
- upgraded test code as needed
- added tests for the following strategies (both with DOM elements and content-affecting elements):
    - absolute-duplicate-strategy
    - absolute-move-strategy
    - absolute-resize-bounding-box-strategy


editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
```diff
@@ -606,6 +615,8 @@ export var storyboard = (
 )
 `
 
+/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "expectElementWithTestIdNotToBeRendered", "expectElementWithTestIdToBeRendered"] }] */
```
ruggi, line 618: is this needed? 🧐

ruggi (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
bkrmendy: yes, a lot of tests that check whether controls are rendered for the correct targets don't use `expect` directly, only the custom expect functions

PR #3461, opened 2023-03-20
Fix/update wrap with conditional

### LINKED ISSUE #3460: Update wrap with conditionals to work with the new reparenting

ruggi:
Fixes #3460 

This PR updates the ability to wrap elements with conditionals to play nicely with the new reparenting/jsx element insertion.

- it uses `WRAP_IN_ELEMENT` instead of `WRAP_IN_VIEW` for the reparenting as conditionals don't need the extras offered by the latter
- it simplifies the code for inserting the conditional element, thanks to the updated reparenting/insertion logic
- it updates the `getConditionalCase` function so that if the metadata is empty (e.g. the spy metadata is not up-to-date), the `true-branch` is returned (which makes sense)
- It _also_ enables wrapping for Fragments


editor/src/components/editor/actions/action-creators.ts
```diff
@@ -775,7 +772,7 @@ export function wrapInView(
 
 export function wrapInElement(
   targets: Array<ElementPath>,
-  whatToWrapWith: { element: JSXElement; importsToAdd: Imports },
+  whatToWrapWith: { element: JSXElement; importsToAdd: Imports } | 'conditional',
```
gbalint, line 775: This is a genuine question and not criticism: have you thought about just using `JSXElementLike | JSXConditionalExpression` instead of treating conditionals as a special case?
Maybe even fragments could mostly work out of the box that way

ruggi: I think it's a good point, the original reasoning was that conditionals (or fragments, since they're also a good mention) don't need any "element" per se as it'll never change, but indeed with the current situation where reparenting takes care of everything I think it makes sense to do what you're suggesting 👍

Rheeseyb (approved):

gbalint (approved):

PR #3447, opened 2023-03-16
Fix/fix flex reorder conditionals

balazsbajorics:
Fixes #[ticket_number]

**Problem:**
Describe the problem being addressed as succinctly as possible.

**Fix:**
Describe the fix you have made as succinctly as possible.

**Commit Details:**
- Renamed `thing` to `otherThing`
- Removed `cake` from `fridge-contents.ts`
- Did [other things]

PR #3465, opened 2023-03-21
Fix 'escape hatch' Convert to absolute and move strategy for multiselect and groups

balazsbajorics:
**Problem:**
My previous PR #3447 disabled the "Convert to absolute and move" strategy for multiselected elements, and for group-like elements with more than 1 child. The multiselection behavior was never 100% defined for this strategy, which is signified not the least by the fact that my change broke 0 tests.
The reason I disabled it in #3447 was that for group-like elements the calculation for autolayout sibling AABB ( introduced in  @Rheeseyb's #3380) was broken, which made this strategy activate way too agressively, beaking autolayout reorder. I also realized there was an assumption in the code that a multiselection can only ever be of siblings. This PR fixes the two problems I've identifed last week.

**Fix:**
* When calculating autolayout siblings, first check if the multiselected elements (or group targets) are actually siblings in the first place. If they are not proper siblings, fall back to the no-siblings behavior.
* In the no-siblings behavior, change the fitness weight so you have to press space to activate the strategy. This may need fine tuning in the future, and insert a selection length check, but it feels like an improvement to me.
* A Whole Ton of New Tests

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx
@@ -12,7 +12,13 @@ import * as Prettier from 'prettier/standalone'
 import { PrettierConfig } from 'utopia-vscode-common'
 import { MetadataUtils } from '../../../../core/model/element-metadata-utils'
 import * as EP from '../../../../core/shared/element-path'
-import { localRectangle, LocalRectangle } from '../../../../core/shared/math-utils'
+import {
+  canvasPoint,
+  localRectangle,
+  LocalRectangle,
+  nullIfInfinity,
+  offsetRect,
+} from '../../../../core/shared/math-utils'
 import { fastForEach } from '../../../../core/shared/utils'
 import {
   runEscapeHatch,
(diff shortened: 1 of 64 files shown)


editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx
```diff
@@ -708,86 +735,224 @@ describe('Convert to absolute/escape hatch', () => {
     const currentStrategy = renderResult.getEditorState().strategyState.currentStrategy
     expect(currentStrategy).toEqual(ConvertToAbsoluteAndMoveStrategyID)
   })
+  ;(['flex', 'flow'] as const).forEach((parentLayoutSystem) =>
```
balazsbajorics, line 738: I recommend reading the tests zero-based, ignore the diff and just look at the tests because I significantly rewrote them.

bkrmendy (approved):

editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx
Rheeseyb: Nah

PR #3472, opened 2023-03-22
Show override pin all the time

gbalint:
**Problem:**
Make the override pin visible all the time.
When there is no override the clicking on the pin will enable an override to the current condition value.
When there is an existing override, clicking on the pin will disable the override.

22-27-tb9i2-lrooz


bkrmendy (approved):

ruggi (approved):
🎉

PR #3482, opened 2023-03-27
Keep selection on empty conditional branch upon deletion

### LINKED ISSUE #3481: When deleting a conditional branch, keep the selection on it

ruggi:
Fixes #3481 

**Problem:**

When deleting a branch, the selection will jump to the parent conditional.

**Fix:**

Make the selection stay in place on the empty branch. Note: this also works for multiple selected targets.

Kapture 2023-03-27 at 14 57 39

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -14,6 +14,7 @@ import {
   generateUidWithExistingComponents,
   getAllUniqueUids,
   getZIndexOfElement,
+  JSXElementChildByElementPath,
   transformJSXComponentAtElementPath,
   transformJSXComponentAtPath,
 } from '../../../core/model/element-template-utils'
@@ -1065,11 +1066,15 @@ export function restoreDerivedState(history: StateHistory): DerivedState {
   }
 }
 
-function deleteElements(targets: ElementPath[], editor: EditorModel): EditorModel {
+function deleteElements(
+  targets: ElementPath[],
+  editor: EditorModel,
+): { editor: EditorModel; affectedElements: JSXElementChildByElementPath } {
   const openUIJSFilePath = getOpenUIJSFileKey(editor)
+  let affectedElements: JSXElementChildByElementPath = {}
   if (openUIJSFilePath == null) {
     console.error(`Attempted to delete element(s) with no UI file open.`)
-    return editor
+    return { editor, affectedElements }
   } else {
     const updatedEditor = targets.reduce((working, targetPath) => {
       const underlyingTarget = normalisePathToUnderlyingTarget(
@@ -1082,10 +1087,12 @@ function deleteElements(targets: ElementPath[], editor: EditorModel): EditorMode
 
       function deleteElementFromParseSuccess(parseSuccess: ParseSuccess): ParseSuccess {
         const utopiaComponents = getUtopiaJSXComponentsFromSuccess(parseSuccess)
-        const withTargetRemoved: Array<UtopiaJSXComponent> = removeElementAtPath(
-          targetPath,
-          utopiaComponents,
-        )
+        const { components: withTargetRemoved, affectedElements: newAffectedElements } =
+          removeElementAtPath(targetPath, utopiaComponents)
+        affectedElements = {
+          ...affectedElements,
+          ...newAffectedElements,
+        }
         return modifyParseSuccessWithSimple((success: SimpleParseSuccess) => {
           return {
             ...success,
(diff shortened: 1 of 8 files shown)


bkrmendy (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1082,10 +1087,12 @@ function deleteElements(targets: ElementPath[], editor: EditorModel): EditorMode
 
       function deleteElementFromParseSuccess(parseSuccess: ParseSuccess): ParseSuccess {
         const utopiaComponents = getUtopiaJSXComponentsFromSuccess(parseSuccess)
-        const withTargetRemoved: Array<UtopiaJSXComponent> = removeElementAtPath(
-          targetPath,
-          utopiaComponents,
-        )
+        const { components: withTargetRemoved, affectedElements: newAffectedElements } =
+          removeElementAtPath(targetPath, utopiaComponents)
+        affectedElements = {
```
gbalint, line 1092: Maybe it is just my preference, and feel free to ignore it, but I dislike using reduce with a function which has an important side effect. In this case I would just use a for loop or a forEach
(My philosophy is that a reduce should continue to work well even if you memoize the function inside it).

ruggi: something like this? https://github.com/concrete-utopia/utopia/pull/3482/commits/b264e393bbf76db93a22bf1c4bc5ad988f440f5e

Rheeseyb: Absolutely agree, reduce shouldn't be triggering side effects. The function could be modified to return an object containing both the `updatedEditor` and the `affectedElements` instead if we wanted to do that.

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1873,10 +1885,28 @@ export const UPDATE_FNS = {
...
           mapDropNulls((view) => {
             const parentPath = EP.parentPath(view)
+            const oldParent = MetadataUtils.findElementByElementPath(editor.jsxMetadata, parentPath)
+            if (
+              oldParent != null &&
+              isRight(oldParent.element) &&
+              isJSXConditionalExpression(oldParent.element.value)
+            ) {
+              const isTrueBranch = EP.toUid(view) === getUtopiaID(oldParent.element.value.whenTrue)
+              const newParent = affectedElements[EP.toString(parentPath)]
```
gbalint, line 1902: This is the original parent of the deleted element, isn't this in the original `editor.jsxMetadata`?

gbalint: ok, i get it!
But then we could also just store the original element path to the branch, am I right?

ruggi: yeah, for sure, it just seemed cleaner this way to me as when deleting you're always dealing with a parent element. normally deleting will remove "something" from a parent, the `null` case for conditionals is a special case. with the current setup you just have a list of the parent elements that were affected by the deletion, which you can then inspect as you wish (in this case, to get the new branch's ID)

gbalint: Moreover, we can just generated the new element path on the fly, can't we? We just have to append the uid of the new null expression to the original parent path, don't we?

ruggi: refactored to actually walk the project contents instead of returning the elements, it's definitely a better solution but I have no idea why it did not work before, I probably did the mistake of using the original project contents and never retry when I tested using the `withElementDeleted` state ¯\_(ツ)_/¯

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1877,6 +1878,32 @@ export const UPDATE_FNS = {
...
+                  }
+                  if (isJSXConditionalExpression(element) && element.uid === EP.toUid(parentPath)) {
+                    newUID = getUtopiaID(isTrueBranch ? element.whenTrue : element.whenFalse)
+                  }
+                })
+              })
+              if (newUID != null) {
+                return EP.appendToPath(parentPath, newUID)
+              }
+            }
```
gbalint, line 1906: This could be even simpler using `withUnderlyingTarget`, with which you could directly access the element for an element path and modify it.

Something like 
```typescript
withUnderlyingTarget(parentPath, editor.projectContents, editor.nodeModules, editor.canvas.openFile, null, (success, element) => {
  if (isJSXConditionalExpression(element) && element.uid === EP.toUid(parentPath)) {
    newUID = getUtopiaID(isTrueBranch ? element.whenTrue : element.whenFalse)
  }
  if (newUID != null) {
    return EP.appendToPath(parentPath, newUID)
  }
})
```

PR #3483, opened 2023-03-27
Patch `globalContentBoxForChildren`

bkrmendy:
## Description
Fragments and conditionals are not rendered in the DOM, so `globalContentBoxForChildren` needs to be faked for them. This PR adds their parent's `globalContentBoxForChildren` to elements that don't have it defined in `fillSpyOnlyMetadata`.

Also, the way `globalContentBoxForChildren` is calculated in `getSpecialSizeMeasurements` is improved, since some elements don't have an offsetParent (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent) for details). To work around this, this PR adds a functions that finds the closest non-null offset parent of an element.


editor/src/components/canvas/dom-walker.ts
```diff
@@ -1365,3 +1365,19 @@ function walkElements(
     return { childPaths: [], rootMetadata: {}, cachedPaths: [] }
   }
 }
+
+function getClosestOffsetParent(element: HTMLElement): Element | null {
+  let currentElement: HTMLElement | null = element
+
+  while (currentElement != null) {
+    const currentElementStyle = window.getComputedStyle(currentElement)
+    if (isElementAContainingBlockForAbsolute(currentElementStyle)) {
```
balazsbajorics, line 1374: IIUC this means `getClosestOffsetParent` may sometimes return the `element` and not any parent. that means the function's name is a lie!

editor/src/core/shared/element-template.ts
```diff
@@ -1941,7 +1941,7 @@ export function specialSizeMeasurements(
   alignItems: FlexAlignment | null,
   htmlElementName: string,
   renderedChildrenCount: number,
-  globalContentBox: CanvasRectangle | null,
+  globalContentBox: MaybeInfinityCanvasRectangle | null,
```
balazsbajorics, line 1944: missing rename

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1997,10 +1991,6 @@ function fillSpyOnlyMetadata(
     return parentLayoutSystem == null
   })
 
-  // Sort and then reverse these, so that lower level elements (with longer paths) are handled ahead of their parents
-  // Sort and then reverse these, so that lower level elements (with longer paths) are handled ahead of their parents
-  // and ancestors. This means that if there are a grandparent and parent which both lack global frames
-  // then the parent is fixed ahead of the grandparent, which will be based on the parent.
```
balazsbajorics, line 2003: why remove these comments?

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2080,9 +2073,11 @@ function fillSpyOnlyMetadata(
 
     workingElements[pathStr] = {
       ...spyElem,
+      ...domElem,
```
balazsbajorics, line 2076: this feels like a massive unrelated, change to me that has barely anything to do with globalContentBoxForChildren, and _more important_ than the content box. if we need this change, I think this deserves a full explanation of the original behavior, how it was bad, the new behavior, how it is good, and also tests that demonstrate both

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1260,39 +1260,33 @@ export const MetadataUtils = {
...
-    // Potentially here the parent was something like a fragment that itself doesn't have
-    // the above properties set, so we should move up the hierarchy to find one where they are set.
-    const nextAncestor =
-      targetParent == null || EP.isEmptyPath(targetParent) ? null : EP.parentPath(targetParent)
-    if (nextAncestor == null) {
-      return zeroCanvasRect
-    } else {
-      return MetadataUtils.getParentCoordinateSystemBounds(nextAncestor, metadata)
-    }
+    return zeroCanvasRect
```
balazsbajorics, line 1273: why not return null here? I'm afraid the zero rect would help hide bugs with the behavior. if we don't have the data, we shouldn't default to not offsetting, we should loudly fail

editor/src/core/model/element-metadata-utils.ts
bkrmendy: remnant of a reverted refactor, readded the comments in https://github.com/concrete-utopia/utopia/pull/3483/commits/db7e55e812e57e8613e8adf2818540f02f099bdc

editor/src/core/shared/element-template.ts
```diff
@@ -1900,7 +1900,7 @@ export interface SpecialSizeMeasurements {
   alignItems: FlexAlignment | null
   htmlElementName: string
   renderedChildrenCount: number
-  globalContentBox: CanvasRectangle | null
+  globalContentBoxForChildren: MaybeInfinityCanvasRectangle | null
```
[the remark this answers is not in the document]
bkrmendy: in case of the storyboard, yes

editor/src/core/model/element-metadata-utils.ts
bkrmendy: reverted in https://github.com/concrete-utopia/utopia/pull/3483/commits/9040fa0cd45ae9501182281f97398f6c3b373c68

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2099,6 +2096,41 @@ function fillSpyOnlyMetadata(
...
+    const parentPathStr = EP.toString(EP.parentPath(EP.fromString(pathStr)))
+
+    const parentGlobalContentBoxForChildren =
+      fromDOM[parentPathStr]?.specialSizeMeasurements.globalContentBoxForChildren ??
+      workingElements[parentPathStr]?.specialSizeMeasurements.globalContentBoxForChildren ??
+      infinityCanvasRectangle
+
+    workingElements[pathStr] = {
+      ...spyElem,
+      ...sameThingFromWorkingElems,
```
balazsbajorics, line 2120: can't you just merge it on the `...sameThingFromWorkingElems` here, leaving out the spyElem completely?

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2099,6 +2096,41 @@ function fillSpyOnlyMetadata(
...
+      globalFrame:
+        sameThingFromWorkingElems?.globalFrame ?? domElem?.globalFrame ?? spyElem.globalFrame,
+      localFrame:
+        sameThingFromWorkingElems?.localFrame ?? domElem?.localFrame ?? spyElem.localFrame,
+      specialSizeMeasurements: {
+        ...emptySpecialSizeMeasurements,
+        ...spyElem.specialSizeMeasurements,
```
balazsbajorics, line 2127: do we need these two lines?

balazsbajorics (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1535,10 +1533,16 @@ export const MetadataUtils = {
       openFile,
     )
 
-    return {
+    workingElements = {
       ...workingElements,
       ...spyOnlyElements,
     }
+
+    const elementsInheritingFromAncestors = fillMissingDataFromAncestors(workingElements) // fontos hogy a workingElements-et adjuk neki
```
seanparsons, line 1541: Errr...

seanparsons (approved):

editor/src/core/model/element-metadata-utils.ts
bkrmendy: Blame Balazs!

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2155,6 +2159,36 @@ function fillSpyOnlyMetadata(
   return workingElements
 }
 
+function fillMissingDataFromAncestors(mergedMetadata: ElementInstanceMetadataMap) {
+  let workingElements: ElementInstanceMetadataMap = { ...mergedMetadata }
+
+  const elementsWithoutGlobalContentBox = Object.keys(workingElements).filter((p) => {
```
Rheeseyb, line 2165: Minor point, but you could use [`Object.entries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) here instead

Rheeseyb (approved):

editor/src/core/model/element-metadata-utils.ts
balazsbajorics: no no no, this is your PR Berci :D the blame deflects from me :D

PR #3487, opened 2023-03-28
Remove condition case labels from ids and test fixes

gbalint:
**Description:**
This is a followup PR to https://github.com/concrete-utopia/utopia/pull/3480
Since conditional clause navigator entries contain the element path of their clause (with the element or expression inside), there is no need to postfix them with `true-case` or `false-case` conditional case label anymore.

I also fixed the broken tests (which I did not realise earlier because the test runner did not run the tests)

### DIFF
--- a/editor/src/core/shared/project-file-types.ts
+++ b/editor/src/core/shared/project-file-types.ts
@@ -1,5 +1,4 @@
 import { ImportType, NormalisedFrame } from 'utopia-api/core'
-import { pathsEqual } from './element-path'
 import {
   ArbitraryJSBlock,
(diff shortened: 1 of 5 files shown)


editor/src/core/shared/project-file-types.ts
```diff
@@ -25,6 +26,10 @@ export interface ElementPath {
   parts: Array<ElementPathPart>
 }
 
+function eq(p: StaticElementPath, p1: StaticElementPath) {
```
ruggi, line 29: This is not used, is it?

PR #3488, opened 2023-03-28
Fix/inspector size section unset

enidemi:
**Problem:**
New inspector size section looks the same when using fallback values as set values. 

**Fix:**
Show fallback values with 'unset' controlstatus. This change is in the new Size section (fill, hug, fixed) and the Text size section.
When using the 'unset' style the hover effect was missing, I wrapped the row divs together to use the same css as other rows.

**Commit Details:**
- update fill-hug-fixed control and text-autosizing-control
- export css for mouseover effects in context-menu-wrapper

It looks like this when the width is not set.
Screenshot 2023-03-28 at 16 42 42

### DIFF
--- a/editor/src/components/context-menu-wrapper.tsx
+++ b/editor/src/components/context-menu-wrapper.tsx
@@ -187,6 +187,24 @@ export class ContextMenuWrapper<T> extends ReactComponent<
   }
 }
 
+export const InspectorRowHoverCSS = {
+  '--control-styles-interactive-unset-main-color': colorTheme.fg7.value,
+  '--control-styles-interactive-unset-secondary-color': colorTheme.fg7.value,
+  '--control-styles-interactive-unset-track-color': colorTheme.bg5.value,
+  '--control-styles-interactive-unset-rail-color': colorTheme.bg3.value,
+  '&:hover': {
+    '--control-styles-interactive-unset-main-color': getControlStyles('simple').mainColor,
+    '--control-styles-interactive-unset-secondary-color': getControlStyles('simple').secondaryColor,
+    '--control-styles-interactive-unset-track-color': getControlStyles('simple').trackColor,
+    '--control-styles-interactive-unset-rail-color': getControlStyles('simple').railColor,
+  },
+  '&:focus-within': {
+    '--control-styles-interactive-unset-main-color': getControlStyles('simple').mainColor,
+    '--control-styles-interactive-unset-secondary-color': getControlStyles('simple').secondaryColor,
+    '--control-styles-interactive-unset-track-color': getControlStyles('simple').trackColor,
+    '--control-styles-interactive-unset-rail-color': getControlStyles('simple').railColor,
+  },
+}
 export class InspectorContextMenuWrapper<T> extends ReactComponent<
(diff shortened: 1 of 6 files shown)


seanparsons (approved):

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -200,24 +218,7 @@ export class InspectorContextMenuWrapper<T> extends ReactComponent<
         css={{
           width: '100%',
           ...(this.props.style as any), // TODO Emotion and React 18 types don't like each other
```
Rheeseyb, line 220: I'm sure we had this issue before 🤔

editor/src/components/inspector/inspector-common.ts
```diff
@@ -515,9 +516,9 @@ export const nukeAllAbsolutePositioningPropsCommands = (
 }
 
 export type FixedHugFill =
-  | { type: 'fixed'; value: CSSNumber }
-  | { type: 'fill'; value: CSSNumber }
-  | { type: 'hug' }
+  | { type: 'fixed'; value: CSSNumber; source: 'parsed' | 'fallback' }
```
Rheeseyb, line 519: Putting the source in here feels very wrong. It might be minor but I think it's worth us creating a separate type that includes the `FixedHugFill` along with the `source`

editor/src/components/inspector/inspector-common.ts
```diff
@@ -551,51 +553,81 @@ export function detectFillHugFixedState(
...
+      parseCSSNumber(element.computedStyle?.flexGrow, 'Unitless'),
+    )
+
+    const isAxisMatchingFlexDirection =
+      (axis === 'horizontal' && isFlexDirectionHorizontal) ||
+      (axis === 'vertical' && isFlexDirectionVertical)
+
+    if (isAxisMatchingFlexDirection) {
+      if (flexGrow != null) {
+        const valueWithType = { type: 'fill' as const, value: flexGrow }
```
Rheeseyb, line 583: I'd say it's worth creating factory functions for these

editor/src/components/inspector/inspector-common.ts
```diff
@@ -812,3 +844,32 @@ export function removeExtraPinsWhenSettingSize(
     deleteProperties('always', elementMetadata.elementPath, [styleP(frameProp)]),
   )
 }
+
+export function isFixedHugFillEqual(
```
Rheeseyb, line 848: I appreciate this was moved from elsewhere and then updated, but this equality function desperately needs improving. It's not only tricky to follow all of the cases where we consider them to be equal, but it also means that if we add anything to the `FixedHugFill` type we are going to default to returning `true`. Maybe it would be better to use an exhaustive switch statement after determining that `a.fixedHugFill` is not `null`?

Rheeseyb (approved):
I'm not too fussed about the factory functions comment, but that equality function needs tidying up. Otherwise this looks good to me 👍

editor/src/components/inspector/inspector-common.ts
enidemi: good idea! fixed :)

PR #3490, opened 2023-03-28
Disable test which ruins the test runner

gbalint:
**Description:**
Disabling this test restores our test runner, so it executes all the other tests. 
Naturally we should fix this test too, but it is still better to skip this single test until then.


ruggi (approved):
(the other broken test was probably a ghost _due_ to this one)

Rheeseyb (approved):

seanparsons:
Are the failing browser tests on this branch another issue?

PR #3491, opened 2023-03-29
Paste into inactive clauses.

seanparsons:
**Problem:**
Users will expect that they can paste into inactive clauses of a conditional in the navigator.

**Fix:**
The main changes to support this are:
- Using `ReparentTargetParent` in more of the logic for reparenting, extending into some utility functions so that it can target the clauses specifically, rather than an element within them.
- Permitting the handling of elements without metadata in the reparent strategies so that an element which is not rendered can be the target of a reparent operation.
- Detail of a potentially important change can be surfaced out of the insertion logic, which is then used to produce a toast informing the user that a value has been replaced instead of preserving it.

**Commit Details:**
- Refactored out the contents of the `ADD_TOAST` and `REMOVE_TOAST` editor actions into utility functions.
- Added `includeToast` and `includeToastPatch` functions.
- `flowParentAbsoluteOrStatic` now permits the `parentMetadata` to be null.
- `AddElement.parentPath` converted to using the `ReparentTargetParent` type.
- `PasteJSXElements.pasteInto` converted to using the `ReparentTargetParent` type.
- `SaveImageInsertWith.parentPath` converted to using the `ReparentTargetParent` type.
- Removed `dispatch` parameter from `ADD_TOAST` handler.
- Added `InsertChildAndDetails` type which holds the updated components and possibly a string which describes a change which may need raising to the user.
- `MetadataUtils.targetElementSupportsChildrenAlsoText` now calls `elementChildSupportsChildrenAlsoText` which contains the logic previously held in the former for checking a `JSXElementChild`.
- Added `MetadataUtils.resolveReparentTargetParentToPath` which handles finding the path of a conditional clause.
- `insertJSXElementChild` now returns `InsertChildAndDetails`, so that it can raise a message about a change which obliterated something.
- `getTargetParentForPaste` now has some special case handling to make dragging into an attribute at the root of a conditional clause target the clause instead.

### DIFF
--- a/editor/src/components/canvas/commands/add-element-command.ts
+++ b/editor/src/components/canvas/commands/add-element-command.ts
@@ -1,16 +1,15 @@
 import {
   getElementPathFromReparentTargetParent,
   ReparentTargetParent,
+  reparentTargetToString,
 } from '../../../components/editor/store/reparent-target'
 import {
   EditorState,
   EditorStatePatch,
   forUnderlyingTargetFromEditorState,
   insertElementAtPath,
-  removeElementAtPath,
 } from '../../../components/editor/store/editor-state'
 import { getUtopiaJSXComponentsFromSuccess } from '../../../core/model/project-file-utils'
-import * as EP from '../../../core/shared/element-path'
 import { JSXElementChild } from '../../../core/shared/element-template'
 import { ElementPath } from '../../../core/shared/project-file-types'
 import { BaseCommand, CommandFunction, getPatchForComponentChange, WhenToRun } from './commands'
@@ -78,6 +77,6 @@ export const runAddElement: CommandFunction<AddElement> = (
 
   return {
     editorStatePatches: editorStatePatches,
-    commandDescription: `Add Element to ${JSON.stringify(command.parentPath)}`,
+    commandDescription: `Add Element to ${reparentTargetToString(command.parentPath)}`,
   }
 }
--- a/editor/src/components/canvas/commands/add-element-command.ts
+++ b/editor/src/components/canvas/commands/add-element-command.ts
@@ -1,4 +1,8 @@
 import {
+  getElementPathFromReparentTargetParent,
+  ReparentTargetParent,
+} from '../../../components/editor/store/reparent-target'
+import {
   EditorState,
   EditorStatePatch,
   forUnderlyingTargetFromEditorState,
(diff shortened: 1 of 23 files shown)


editor/src/components/canvas/commands/add-element-command.ts
```diff
@@ -63,12 +69,15 @@ export const runAddElement: CommandFunction<AddElement> = (
...
+        editorStatePatchNewParentFile,
+        includeToastPatch(insertionResult.insertionDetails, editorState),
+      ]
     },
   )
 
   return {
     editorStatePatches: editorStatePatches,
-    commandDescription: `Add Element to ${EP.toString(command.parentPath)}`,
+    commandDescription: `Add Element to ${JSON.stringify(command.parentPath)}`,
```
ruggi, line 81: maybe it'd be worth to add a utility function to convert `ReparentTargetParent` to string

ruggi (approved):
looks good to me!

editor/src/components/canvas/commands/add-element-command.ts
seanparsons: Fixed in e1cd2a6b4.

editor/src/components/editor/store/reparent-target.ts
```diff
@@ -53,3 +53,13 @@ export function dynamicReparentTargetParentToStaticReparentTargetParent(
     return EP.dynamicPathToStaticPath(reparentTargetParent)
   }
 }
+
+export function reparentTargetToString<P extends ElementPath>(
```
ruggi, line 57: thank you!

editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
```diff
@@ -1406,4 +1494,251 @@ describe('conditionals in the navigator', () => {
...
+          ),
+        ),
+      )
+      const navigatorEntryToPasteIntoRect = navigatorEntryToPasteInto.getBoundingClientRect()
+      const navigatorEntryToPasteIntoCenter = getDomRectCenter(navigatorEntryToPasteIntoRect)
+
+      // Select the target paste into in the navigator.
+      await act(async () => {
+        await mouseClickAtPoint(navigatorEntryToPasteInto, navigatorEntryToPasteIntoCenter)
+      })
```
Rheeseyb, line 1692: You don't need to wrap this in an `await act`, because it's already doing that

Rheeseyb (approved):

PR #3492, opened 2023-03-29
Remove reliance on cross iframe IndexedDB use for VS Code communication

Rheeseyb:
**Problem:**
We recently discovered that an upcoming security feature will break the code editor on Utopia, due to the way we have been using IndexedDB as a communication channel, combined with an iframe switcheroo that would have an inner iframe on the same origin as the editor (so that both could access the same IndexedDB store), but inside an outer iframe that would be use a different origin (so that Chrome would give the code editor a dedicated process).

The security feature in question is Storage Partitioning: https://developer.chrome.com/en/docs/privacy-sandbox/storage-partitioning, and Chrome is planning to switch this on in “early 2023”, with other browsers following suit. Storage partitioning means that the inner iframe will be unable to access the same IndexedDB store as the main Utopia editor, even though it is running on the same TLD, because the outer iframe’s change of origin introduces a partition that is inherited by the inner iframe.

**Fix:**
To save us running a full investigation here to find a possibly better solution (if one is even possible), and to save us having to patch any code deep in the core of the VS Code codebase, we have instead moved the IndexedDB communication channel completely inside VS Code, where it now creates channel from the outermost layer (in a newly created file `communication.ts` with the exported function `setupEventListenersForProject` called in `workbench.ts`, which is the main entrypoint) to our VS Code extension. We then complete the communication with Utopia via `postMessage` calls.

This effectively means that channel becomes split into two parts:

1. `window.postMessage` to communicate file changes and messages between Utopia and the outermost layer of VS Code
2. IndexedDB used as before, but to now communicate those file changes and messages between the outermost layer of VS Code and our extension

Since we've now moved a lot of the previous contents of `vscode-bridge.ts` inside VS Code, it meant adding a dependency on `utopia-vscode-common` to VS Code. Due to the way in which VS Code imports dependencies, I had to add a webpack build for that module, update the VS Code build file to copy across the result of that build, and provide the location for it in the iframe's HTML file itself.

Finally, since we no longer need an inner and outer iframe, I have removed the previous outer iframe, and renamed the inner one.


vscode-build/vscode.patch
```diff
@@ -176,11 +188,195 @@ index 6148eb30092..6948ededa4a 100644
...
++        alreadyExistingFile.unsavedContent !== unsavedContent
++      )
++      if (fileDiffers) {
++        // Avoid pushing a file to the file system if the content hasn't changed.
++        return writeFileAsUTF8(filePath,
++          savedContent,
++          unsavedContent)
++      }
++    }
++    // FIXME Default never
```
seanparsons, line 279: You put a FIXME into the patch? :)

seanparsons (approved):

vscode-build/vscode.patch
Rheeseyb: I should probably fix that

PR #3493, opened 2023-03-29
Rename uniqueID to uid.

seanparsons:
**Problem:**
As a result of historic design and recent refactors, there was an inconsistency between the various subtypes of `JSXElementChild`, where some had a property `uniqueID` and others a property `uid`, which represented the same value.

**Fix:**
Mostly this was done with a rename refactor in my editor. But `clearJSXElementChildUniqueIDs` needed some corrections to prevent it from creating fields that didn't exist but still sort of matched the return type of the function.

**Commit Details:**
- Renamed `clearJSXElementUniqueIDs` to `clearJSXElementChildUniqueIDs` and removed the type parameter.
- Fixed `clearJSXElementChildUniqueIDs` to use a comprehensive switch/case.
- Created `clearJSXElementUniqueIDs` purely for `JSXElement`.
- Updated the various cases where `uniqueID` was used as a property to `uid`.


editor/src/components/custom-code/code-file.ts
```diff
@@ -39,7 +39,7 @@ import {
   JSXElementWithoutUID,
   JSXFragmentWithoutUID,
   UtopiaJSXComponent,
-  clearJSXElementUniqueIDs,
+  clearJSXElementChildUniqueIDs,
```
ruggi, line 42: this can be removed

ruggi (approved):

editor/src/components/canvas/__snapshots__/ui-jsx-canvas-bugs.spec.tsx.snap
```diff
@@ -72,13 +72,13 @@ Object {
...
                   "type": "ATTRIBUTE_VALUE",
-                  "uniqueID": "",
+                  "uid": "",
                   "value": "app",
                 },
               },
             ],
             "type": "JSX_ELEMENT",
-            "uid": "app",
+            "uid": "",
```
Rheeseyb, line 81: Should this value have been cleared?

gbalint (approved):

PR #3500, opened 2023-03-31
wrap in a fragment if target does not support children

### LINKED ISSUE #3499: fix inserting into elements which don't support children

ruggi:
Fixes #3499 

**Problem:**

Reparenting _directly_ into a conditional clause with an image inserts the element _into_ the image.

**Fix:**

This is due to a more broad issue, caused by reparenting into elements which don't support children. This PR fixes it by creating a fragment for the insertion target and the inserted element when the parent `doesNotSupportChildren`.

Before:
Kapture 2023-03-31 at 10 32 10

After:
Kapture 2023-03-31 at 10 33 29


gbalint (approved):

maltenuhn:
Does this mean that I can now reparent into an image anywhere, and it'll create a fragment? If so, that's definitely not the desired behaviour; we should continue to _prevent_ a reparent (as we already do in a number of situations)

PR #3501, opened 2023-03-31
Make expression UIDs persist across parses.

seanparsons:
**Problem:**
Elements have historically had their UIDs be made persistent across parses so that selection can be maintained while changes are made to the code or at least a best effort thereof. However the expressions/attributes now have UIDs but this persistence has not been extended to them so selection will likely not be maintained with those.

**Fix:**
`fixParseSuccessUIDs` which performs the work of carrying over the old UIDs from the previous parse has been heavily rewritten to support the new cases of `JSXElementChild`. Making use of exhaustive switch statements to ensure that should any new cases arise they will have to be handled.

Oddly a circular reference issue arose as a result of this change which was fixed by just removing a constant from `project-file-utils.ts`.

So as to check that the code was covering as many cases as possible FastCheck was used, which necessitated updating some arbitrary generators as they were out of date. As a result of that a minor issue with the code printing was discovered, whereby `ATTRIBUTE_OTHER_JAVASCRIPT` entries needed to be handled differently depending on if they were contained within a JSX element or not.

**Commit Details:**
- Reworked `fixParseSuccessUIDs` to support persisting the UIDs of expressions as well as elements.
- Lots of support functions for `fixParseSuccessUIDs` were implemented, handling specific cases.
- Removed a constant from `project-file-utils.ts` due to it creating a cycle that became broken.
- Updated the arbitrary generator for `JSXElementChild`, adding new generators for fragments and conditional expressions.
- Fixed a bug in `jsxElementToExpression` by adding a `parentIsJSX` parameter so that `ATTRIBUTE_OTHER_JAVASCRIPT` cases can have braces added around them if they are directly within a JSX element (as opposed to another expression).

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer.ts
@@ -288,7 +288,7 @@ function updateJSXElementsWithin(
   ): TS.Node {
     return withUID(undefined, attributes, originalNode, (uid) => {
       if (uid in elementsWithin) {
-        return jsxElementToExpression(elementsWithin[uid], imports, stripUIDs)
+        return jsxElementToExpression(elementsWithin[uid], imports, stripUIDs, false)
       } else {
         return originalNode
       }
@@ -376,6 +376,7 @@ function jsxElementToExpression(
   element: JSXElementChild,
   imports: Imports,
   stripUIDs: boolean,
+  parentIsJSX: boolean,
 ):
   | TS.JsxElement
   | TS.JsxSelfClosingElement
@@ -440,42 +441,48 @@ function jsxElementToExpression(
         const closing = TS.createJsxClosingElement(tagName)
         return TS.createJsxElement(
           opening,
-          element.children.map((child) => jsxElementToJSXExpression(child, imports, stripUIDs)),
+          element.children.map((child) =>
+            jsxElementToJSXExpression(child, imports, stripUIDs, true),
+          ),
           closing,
         )
       }
     }
     case 'ATTRIBUTE_OTHER_JAVASCRIPT': {
-      const maybeExpressionStatement = rawCodeToExpressionStatement(element.javascript)
-      let rawCode: string = element.javascript // Fallback for the case where the code is simply a comment
-      if (maybeExpressionStatement != null) {
-        const { statement, sourceFile } = maybeExpressionStatement
-        const lastToken = statement.getLastToken(sourceFile)
-        const finalComments =
-          lastToken == null
-            ? emptyComments
-            : parsedComments([], getLeadingComments(element.javascript, lastToken))
+      if (parentIsJSX) {
+        const maybeExpressionStatement = rawCodeToExpressionStatement(element.javascript)
+        let rawCode: string = element.javascript // Fallback for the case where the code is simply a comment
+        if (maybeExpressionStatement != null) {
+          const { statement, sourceFile } = maybeExpressionStatement
+          const lastToken = statement.getLastToken(sourceFile)
+          const finalComments =
+            lastToken == null
+              ? emptyComments
+              : parsedComments([], getLeadingComments(element.javascript, lastToken))
 
-        const updatedStatement = updateJSXElementsWithin(
-          statement.expression,
-          element.elementsWithin,
-          imports,
-          stripUIDs,
-        )
-        addCommentsToNode(updatedStatement, finalComments)
-        rawCode = TS.createPrinter({ omitTrailingSemicolon: true }).printNode(
-          TS.EmitHint.Unspecified,
-          updatedStatement,
-          sourceFile,
-        )
+          const updatedStatement = updateJSXElementsWithin(
+            statement.expression,
+            element.elementsWithin,
+            imports,
+            stripUIDs,
+          )
+          addCommentsToNode(updatedStatement, finalComments)
+          rawCode = TS.createPrinter({ omitTrailingSemicolon: true }).printNode(
+            TS.EmitHint.Unspecified,
+            updatedStatement,
+            sourceFile,
+          )
+        }
+
+        // By creating a `JsxText` element containing the raw code surrounded by braces, we can print the code directly
+        return TS.createJsxText(`{${rawCode}}`)
+      } else {
+        return jsxAttributeToExpression(element)
       }
-
-      // By creating a `JsxText` element containing the raw code surrounded by braces, we can print the code directly
-      return TS.createJsxText(`{${rawCode}}`)
     }
     case 'JSX_FRAGMENT': {
       const children = element.children.map((child) => {
-        return jsxElementToJSXExpression(child, imports, stripUIDs)
+        return jsxElementToJSXExpression(child, imports, stripUIDs, true)
       })
       if (element.longForm) {
         const tagName = jsxElementNameToExpression(getFragmentElementNameFromImports(imports))
@@ -499,8 +506,8 @@ function jsxElementToExpression(
     }
     case 'JSX_CONDITIONAL_EXPRESSION': {
       const condition = jsxAttributeToExpression(element.condition)
-      const whenTrue = jsxElementToExpression(element.whenTrue, imports, stripUIDs)
-      const whenFalse = jsxElementToExpression(element.whenFalse, imports, stripUIDs)
+      const whenTrue = jsxElementToExpression(element.whenTrue, imports, stripUIDs, false)
+      const whenFalse = jsxElementToExpression(element.whenFalse, imports, stripUIDs, false)
 
       const node = TS.createConditional(
         condition,
@@ -527,8 +534,9 @@ function jsxElementToJSXExpression(
   element: JSXElementChild,
   imports: Imports,
   stripUIDs: boolean,
+  parentIsJSX: boolean,
 ): TS.JsxElement | TS.JsxSelfClosingElement | TS.JsxText | TS.JsxExpression | TS.JsxFragment {
-  const expression = jsxElementToExpression(element, imports, stripUIDs)
+  const expression = jsxElementToExpression(element, imports, stripUIDs, parentIsJSX)
   if (
     TS.isJsxElement(expression) ||
     TS.isJsxSelfClosingElement(expression) ||
@@ -627,7 +635,7 @@ function printUtopiaJSXComponent(
   element: UtopiaJSXComponent,
   detailOfExports: ExportsDetail,
 ): TS.Node {
-  const asJSX = jsxElementToExpression(element.rootElement, imports, printOptions.stripUIDs)
+  const asJSX = jsxElementToExpression(element.rootElement, imports, printOptions.stripUIDs, true)
   if (
     TS.isJsxElement(asJSX) ||
     TS.isJsxSelfClosingElement(asJSX) ||
--- a/editor/src/core/workers/parser-printer/parser-printer.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer.ts
@@ -441,9 +441,7 @@ function jsxElementToExpression(
         const closing = TS.createJsxClosingElement(tagName)
         return TS.createJsxElement(
           opening,
-          element.children.map((child) =>
-            jsxElementToJSXExpression(child, imports, stripUIDs, true),
-          ),
+          element.children.map((child) => jsxElementToJSXExpression(child, imports, stripUIDs)),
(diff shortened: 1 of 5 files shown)


editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,569 @@
...
-        workingComponents = transformJSXComponentAtElementPath(
-          workingComponents,
-          mapping.pathToModify,
-          (element) => {
-            return setUtopiaID(element, mapping.oldUID)
-          },
+  const fixedCombinedTopLevelArbitraryBlock =
+    oldParsed.combinedTopLevelArbitraryBlock == null ||
+    newParsed.combinedTopLevelArbitraryBlock == null
+      ? newParsed.combinedTopLevelArbitraryBlock
```
ruggi, line 81: I think that for clarity this could be explicitly be `null`

```ts
const fixedCombinedTopLevelArbitraryBlock =
  oldParsed.combinedTopLevelArbitraryBlock == null ||
  newParsed.combinedTopLevelArbitraryBlock == null
    ? null
    : fixArbitraryJSBlockUIDs(
        oldParsed.combinedTopLevelArbitraryBlock,
        newParsed.combinedTopLevelArbitraryBlock,
        fixUIDsState,
      )
```

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,569 @@
...
-      )
-      numberOfElementsMatch = numberOfElementsMatch && lengthsOfChildrenMatch
+  fixUIDsState: FixUIDsState,
+  baseValue: T,
+): T {
+  const newUID = unsafeGet(uidOptic, baseValue)
+  let uidToUse: string = oldUID
+  if (oldUID === newUID) {
+    // Old one is the same as the new one, so everything is great.
+    uidToUse = newUID
```
ruggi, line 117: this case can be removed, as it's already covered by the assignment on line `114`

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,569 @@
...
+  mappings: Array<{ oldUID: string; newUID: string }>
+}
 
 export function fixParseSuccessUIDs(
   oldParsed: ParseSuccess | null,
   newParsed: ParsedTextFile,
   alreadyExistingUIDs: Set<string>,
 ): ParsedTextFile {
-  if (oldParsed == null || !isParseSuccess(newParsed)) {
+  if (oldParsed == null || newParsed.type !== 'PARSE_SUCCESS') {
```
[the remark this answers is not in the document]
seanparsons: This was done in the hunt for the circular dependency.

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,569 @@
...
+  } else {
+    // The uids have changed:
+    // - Remove the new one from the set, as now it will be unused.
+    // - Add the old one to the set, as it will become used.
+    // - Return the old one.
+    // One assumption here is that there is no duplicate uids in the new elements.
+    fixUIDsState.mutableAllNewUIDs.delete(newUID)
+    fixUIDsState.mutableAllNewUIDs.add(oldUID)
+    fixUIDsState.mappings.push({ oldUID: oldUID, newUID: newUID })
+    uidToUse = oldUID
```
ruggi, line 130: this can also be removed

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,569 @@
...
+    uidToUse = newUID
+  } else if (fixUIDsState.mutableAllNewUIDs.has(oldUID)) {
+    // The old uid is already present somewhere, so using it will introduce a duplicate.
+    uidToUse = newUID
+  } else {
+    // The uids have changed:
+    // - Remove the new one from the set, as now it will be unused.
+    // - Add the old one to the set, as it will become used.
+    // - Return the old one.
+    // One assumption here is that there is no duplicate uids in the new elements.
```
Rheeseyb, line 126: Hmmm can we do anything about this assumption so that we don't have to consider it an assumption? This sounds like something that would happen after a `cmd`+`d` duplication.

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,569 @@
...
+}
+
+export function fixUtopiaJSXComponentUIDs(
+  oldElement: UtopiaJSXComponent,
+  newElement: UtopiaJSXComponent,
+  fixUIDsState: FixUIDsState,
+): UtopiaJSXComponent {
+  const fixedArbitraryJSBlock =
+    oldElement.arbitraryJSBlock == null || newElement.arbitraryJSBlock == null
+      ? newElement.arbitraryJSBlock
```
ruggi, line 213: also here I think it'd be clearer to explicitly use `null`

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,569 @@
...
+        return {
+          ...newExpression,
+          spreadValue: fixedSpreadValue,
+        }
+      }
+      break
+    }
+    default:
+      const _exhaustiveCheck: never = newExpression
+      throw new Error(`Unhandled case ${JSON.stringify(newExpression)}.`)
```
ruggi, line 343: and this with `assertNever`?

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,569 @@
...
+    case 'ATTRIBUTE_FUNCTION_CALL':
+    case 'ATTRIBUTE_OTHER_JAVASCRIPT': {
+      if (oldElement.type === newElement.type) {
+        return fixExpressionUIDs(oldElement, newElement, fixUIDsState)
+      }
+      break
+    }
+    default:
+      const _exhaustiveCheck: never = newElement
+      throw new Error(`Unhandled case ${JSON.stringify(newElement)}.`)
```
ruggi, line 467: can this also be replaced with `assertNever`?

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,569 @@
...
+        return updateUID(expressionOtherJavaScriptUIDOptic, oldExpression.uid, fixUIDsState, {
+          ...newExpression,
+          elementsWithin: fixedElementsWithin,
+        })
+      }
+      break
     }
+    default:
+      const _exhaustiveCheck: never = newExpression
+      throw new Error(`Unhandled case ${JSON.stringify(newExpression)}.`)
```
ruggi, line 565: and here `assertNever` :)

editor/src/core/workers/parser-printer/uid-fix.ts
seanparsons: Currently we only utilise this during a parse and the parse should have resulted in there being no duplicates. It seemed worth mentioning just in case some future changes meant that this assumption resulted in breakage.

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -527,8 +534,9 @@ function jsxElementToJSXExpression(
   element: JSXElementChild,
   imports: Imports,
   stripUIDs: boolean,
+  parentIsJSX: boolean,
```
ruggi, line 537: as far as I can see, this function is used where all parents are JSX elements, so unless there's a clear future need for the opposite, maybe this could be removed here? (keeping it in `jsxElementToExpression` instead, of course)

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,569 @@
...
+    }
+    case 'IMPORT_STATEMENT': {
+      return newElement
+    }
+    case 'UNPARSED_CODE': {
+      return newElement
+    }
+    default:
+      const _exhaustiveCheck: never = newElement
+      throw new Error(`Unhandled case ${JSON.stringify(newElement)}.`)
```
ruggi, line 200: can this be replaced with `assertNever`?

editor/src/core/workers/parser-printer/uid-fix.ts
ruggi: uh, that's true 🤦

Rheeseyb (approved):

editor/src/core/workers/parser-printer/uid-fix.ts
seanparsons: The code on line 114 sets it to `oldUID`...

ruggi: but they are checked to be the same, right?

seanparsons: I've removed the assignment from line 114, so that it's mandatory to assign it in the various conditionals, in 6a3d04e93.

editor/src/core/workers/parser-printer/uid-fix.ts
seanparsons: Fixed in 6a3d04e93.

editor/src/core/workers/parser-printer/uid-fix.ts
seanparsons: Fixed in 6a3d04e93.

editor/src/core/workers/parser-printer/uid-fix.ts
seanparsons: Fixed in 6a3d04e93.

editor/src/core/workers/parser-printer/uid-fix.ts
seanparsons: Fixed in 6a3d04e93.

editor/src/core/workers/parser-printer/uid-fix.ts
seanparsons: Fixed in 6a3d04e93.

editor/src/core/workers/parser-printer/uid-fix.ts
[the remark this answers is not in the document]
seanparsons: Fixed in 6a3d04e93.

editor/src/core/workers/parser-printer/parser-printer.ts
ruggi: my comment is related to `jsxElementToJSXExpression`, isn't this it?

seanparsons: Gah, these functions are too closely named, this is `jsxElementToJSXExpression`.

seanparsons: Fixed in bfa4f21ee.

ruggi (approved):

gbalint (approved):

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -446,32 +447,36 @@ function jsxElementToExpression(
...
-        addCommentsToNode(updatedStatement, finalComments)
-        rawCode = TS.createPrinter({ omitTrailingSemicolon: true }).printNode(
-          TS.EmitHint.Unspecified,
-          updatedStatement,
-          sourceFile,
-        )
-      }
+      if (parentIsJSX) {
+        const maybeExpressionStatement = rawCodeToExpressionStatement(element.javascript)
+        let rawCode: string = element.javascript // Fallback for the case where the code is simply a comment
```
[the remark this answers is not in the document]
seanparsons: That's the pre-existing code however, I don't think I should be refactoring that in this.

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -1,284 +1,570 @@
...
+      break
+    }
+    case 'ARBITRARY_JS_BLOCK': {
+      if (oldElement.type === newElement.type) {
+        return fixArbitraryJSBlockUIDs(oldElement, newElement, fixUIDsState)
+      }
+      break
+    }
+    case 'IMPORT_STATEMENT': {
+      return newElement
```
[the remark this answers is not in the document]
seanparsons: That's incredibly subjective though, I definitely have a preference for the case to be as near to the result for that case as possible.

PR #3504, opened 2023-04-03
group flag

bkrmendy:
## Description
This PR adds a new comment flag and a special property, so that we can designate elements as "groups". This way, we can decouple the "implementation" of a group from group behaviours (i.e., the specific code construct implementing the group behaviours in a certain context from the group behaviours themselves). Additionally, once an element is marked as a group, the editor can apply the group behaviours proactively.


editor/src/components/canvas/canvas-strategies/strategies/group-like-helpers.ts
```diff
@@ -162,3 +167,33 @@ export function treatElementAsContentAffecting(
 ): boolean {
   return getElementContentAffectingType(metadata, allElementProps, path) != null
 }
+
+export const GroupFlagKey = 'data-group'
+
+export function isGroupPragmaApplied(
```
balazsbajorics, line 173: I think pragma has a ton of baggage, I'd pick a name which is a little more on the safe side about using big words  

```suggestion
export function isElementMarkedAsGroup(
```

bkrmendy: makes perfect sense, renamed in https://github.com/concrete-utopia/utopia/pull/3504/commits/c79e4f708be85ebe7f46c7c022cd725b33eb186e

editor/src/components/canvas/canvas-strategies/strategies/group-like-helpers.ts
```diff
@@ -162,3 +167,33 @@ export function treatElementAsContentAffecting(
...
+
+  if (isJSXElement(instance.element.value)) {
+    return (
+      instance.element.value.props.find(
+        (prop) =>
+          prop.type === 'JSX_ATTRIBUTES_ENTRY' &&
+          prop.key === GroupFlagKey &&
+          prop.value.type === 'ATTRIBUTE_VALUE' &&
+          prop.value.value === true,
+      ) != null
```
Rheeseyb, line 194: I'm sure we have a function for checking the value of a prop

balazsbajorics: oh crap I totally missed this! yes we have very good helpers for this :)

bkrmendy: I wish we had chatgpt or something to pick this up automatically 😅 any pointers about what this function might be?

Rheeseyb: Ok I think the best function to use here is [`getSimpleAttributeAtPath`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/model/element-metadata-utils.ts#L2304-L2323>)

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/group-like-helpers.ts
bkrmendy: thanks! added in https://github.com/concrete-utopia/utopia/pull/3504/commits/e8296ed05698f78e61c9aaece04d7eea74524218

ruggi (approved):

PR #3505, opened 2023-04-03
Navigator: no reparent/paste into a conditional

gbalint:
**Problem:**
Currently, reparenting/pasting into a conditional means we reparent/paste into the active branch of the conditional.
However, according the slot-like plans we would like to eliminate this behavior, and only allow reparenting/pasting directly into a clause (and even them in a different way, but that is not in the scope of this PR)

**Fix:**
- `targetElementSupportsChildren` should return false for conditionals - this guarantees you can not drag into a conditional in the navigator
- I removed the whole insertion logic into conditional logic from `insertJSXElementChild` - this allowed me to remove the metadata parameter (which was only needed to check the actual condition value)
- I added a new test to make sure you can not drag into a conditional in the navigator
- I removed the paste tests for conditionals (because we do not want to allow that neither)

**Missing:**
When you drag `after` or `before` elements in the navigator, we don't check if the future parent supports children or not. See this line https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx#L97 (When the target is not a reparent target, we don't check whether the element supports children or not.) 
The reparent will not happen because I removed the code from `insertJSXElementChild`, but we still see the drop target highlight.
This should be fixed in a subsequent PR, but the hover behavior is so different in the slot-like world, that I would only fix this when that design is implemented.


seanparsons:
There are some paste tests in `navigator-conditionals.spec.browser2.tsx` as well, which I assume should also be removed?

gbalint:
> There are some paste tests in `navigator-conditionals.spec.browser2.tsx` as well, which I assume should also be removed?

@seanparsons Those tests check pasting into clauses, and not directly into the conditional expression itself. So those are still valid and important (otherwise they would be broken)

ruggi (approved):

Rheeseyb (approved):

PR #3507, opened 2023-04-03
Clear down conditional inactive branch metadata.

seanparsons:
**Problem:**
When toggling the condition of a conditional expression, the navigator still shows the content of the now unrendered branch as if it was still being rendered.

**Fix:**
Nothing was performing any cleanup on the metadata for the entries being created when those entries ceased to exist, which resulted in the navigator logic finding them and populating additional navigator entries from those.

Now if any metadata exists for the conditional clause that is not rendered, we run a small process that sweeps up those spy entries for any descendants of that clause.

**Commit Details:**
- Implemented `clearOpposingConditionalSpyValues` to remove any metadata which is a descendant of the unrendered branch of a conditional.
- Added call to `clearOpposingConditionalSpyValues` to `renderCoreElement`.


ruggi (approved):
\o\ \o\

Rheeseyb (approved):

PR #3510, opened 2023-04-04
`x` shortcut on sizeless div

bkrmendy:
# Try it [here](https://utopia.pizza/p/256cf461-pie-vermicelli/?branch_name=fix-sizeless-div-x-shortcut)

## Problem
When pressing `x` on a sizeless div, we only add a `width` and `height` prop. We should adding an additional `position: absolute` and appropriate `top, left` props, and the sizeless div's children should have their positions adjusted as well (so that they stay in the same visual position).

## Fix
Implement the above in the shortcut handler.

Additionally, a heuristic rule is added to handle converting into a group (the reverse action of the above). If an element
- is absolutely positioned,
- has more than 0 children,
- all these children are absolutely positioned,
the element is converted into a sizeless div, and the children offsets are adjusted so that they remain visually intact at the same absolute positions.


editor/src/components/editor/shortcuts.spec.browser2.tsx
```diff
@@ -28,6 +29,8 @@ const StoryBoardId = 'StoryBoardId'
 const ParentId = 'ParentId'
 const backgroundColor = '#384C5CAB'
 
+/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "expectBoundsToEqual", "expectNoAction"] }] */
```
Rheeseyb, line 32: Nice!

editor/src/components/inspector/inspector-common.ts
```diff
@@ -516,6 +516,12 @@ export const nukeAllAbsolutePositioningPropsCommands = (
   ]
 }
 
+export const nukeSizingProps = (path: ElementPath): Array<CanvasCommand> => {
```
Rheeseyb, line 519: We have `nukeSizingPropsForAxisCommand` which also includes the `min*` and `max*` props, so maybe it's worth either calling that for both axes, or including those props here

editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -903,7 +917,48 @@ export function handleKeyDown(
               return []
             }
 
+            const parentBounds =
```
Rheeseyb, line 920: I think this entire piece of code (everything under `REMOVE_ABSOLUTE_POSITIONING`) needs to be extracted into its own file plus helper functions

editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -921,6 +976,46 @@ export function handleKeyDown(
...
+              (path) => MetadataUtils.findElementByElementPath(editor.jsxMetadata, path),
+              replaceContentAffectingPathsWithTheirChildrenRecursive(
+                editor.jsxMetadata,
+                editor.allElementProps,
+                MetadataUtils.getChildrenPathsUnordered(editor.jsxMetadata, elementPath),
+              ),
+            )
+
+            return [
+              deleteProperties('always', elementPath, [PP.create('style')]),
```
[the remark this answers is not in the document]
bkrmendy: 😬 fixed in https://github.com/concrete-utopia/utopia/pull/3510/commits/08051de7ece2a962fb4cb23735391b9009e75363

editor/src/components/inspector/inspector-common.ts
bkrmendy: fixed in https://github.com/concrete-utopia/utopia/pull/3510/commits/08051de7ece2a962fb4cb23735391b9009e75363

Rheeseyb (approved):

enidemi (approved):

PR #3512, opened 2023-04-04
(WIP) slots for conditional branches

ruggi:
WIP

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item.tsx
@@ -27,10 +27,11 @@ import { LayoutIcon } from './layout-icon'
 import { Substores, useEditorState } from '../../editor/store/store-hook'
 import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import { ThemeObject } from '../../../uuiui/styles/theme/theme-helpers'
-import { when } from '../../../utils/react-conditionals'
-import { isLeft } from '../../../core/shared/either'
+import { unless, when } from '../../../utils/react-conditionals'
+import { isLeft, isRight } from '../../../core/shared/either'
 import {
   ElementInstanceMetadata,
+  isJSXAttributeValue,
   isJSXConditionalExpression,
   JSXConditionalExpression,
 } from '../../../core/shared/element-template'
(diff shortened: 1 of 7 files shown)


editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -536,26 +533,58 @@ export const NavigatorItem: React.FunctionComponent<
...
+  )
+
+  const isSlot = React.useMemo(() => {
+    const isParentConditional =
+      parentElement != null &&
+      isRight(parentElement.element) &&
+      isJSXConditionalExpression(parentElement.element.value)
+    const isNullValue =
+      isSyntheticNavigatorEntry(props.navigatorEntry) &&
+      isJSXAttributeValue(props.navigatorEntry.childOrAttribute) &&
```
seanparsons, line 552: Feels a bit like this logic should be wrapped up in a function, probably using `jsxSimpleAttributeToValue` to evaluate the value itself after checking it's a `JSExpression`.

editor/src/core/shared/element-template.ts
```diff
@@ -617,6 +617,10 @@ export function isJSXAttributeValue(element: JSXElementChild): element is JSExpr
   return element.type === 'ATTRIBUTE_VALUE'
 }
 
+export function isNullJSXAttributeValue(element: JSXElementChild): boolean {
```
seanparsons, line 620: This function is what I was thinking of, but it should probably use `jsxSimpleAttributeToValue` to cater for more cases.

seanparsons (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
ruggi: ah yeah makes sense, I forgot this one - nice catch!

editor/src/core/shared/element-template.ts
ruggi: would this work? https://github.com/concrete-utopia/utopia/pull/3512/commits/1504c93a00ae02a09442cf1bfc831db2fada9872

bkrmendy (approved):

balazsbajorics (approved):

PR #3514, opened 2023-04-04
Wrap in div inside conditionals.

seanparsons:
**Problem:**
Wrapping in a div (or other elements) is not handled correctly when targeting the element at the root of a conditional clause, with it replacing the item to be wrapped entirely.

**Fix:**
So that it is possible to target a conditional clause specifically, the `ReparentTargetParent` type has been utilised in place of `ElementPath` for these cases, with the parent to be manipulated being identified using that type and some new functions implemented in this PR.

The reason why the original elements were being lost was due to the new parent being inserted and then the target being moved in a separate step afterwards. But in the case of a conditional clause there can't be multiple elements in that position, so when the new parent was inserted the target was eliminated by default.

**Notes:**
`ReparentTargetParent` is increasingly being used as an alternative to `ElementPath` with some special case logic that handles it being implemented on a case by case basis. Which implies that `ElementPath` should really be extended to cover the cases that `ReparentTargetParent` covers.

**Commit Details:**
- Added `commonReparentTarget` to attempt to find a common reparent between two instances of `ReparentTargetParent`, akin to `closestSharedAncestor` for `ElementPath`.
- Implemented `commonReparentTargetFromArray` which utilises `commonReparentTarget` across multiple entries as well as null ones.
- `getClauseOptic` added to cover a recurring case of needing one or the other optic.
- `getReparentTargetOfTarget` implemented to identify a `ReparentTargetParent` for a particular `ElementPath`, which handles finding the appropriate clause of a conditional should the element be an immediate "child" of the condtional.
- For both `WRAP_IN_VIEW` and `WRAP_IN_ELEMENT` actions, use functions operating on `ReparentTargetParent` so as to cater for conditional clauses correctly. With a check on the type of where the reparenting is being applied.

### DIFF
--- a/editor/src/components/editor/store/reparent-target.ts
+++ b/editor/src/components/editor/store/reparent-target.ts
@@ -1,6 +1,8 @@
 import type { ElementPath, StaticElementPath } from '../../../core/shared/project-file-types'
 import * as EP from '../../../core/shared/element-path'
 import { ConditionalCase } from '../../../core/model/conditionals'
+import { getUtopiaID } from '../../../core/shared/uid-utils'
+import { drop } from '../../../core/shared/array-utils'
 
 export interface ConditionalClause<P extends ElementPath> {
   elementPath: P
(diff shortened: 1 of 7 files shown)


editor/src/components/editor/store/reparent-target.ts
```diff
@@ -63,3 +65,57 @@ export function reparentTargetToString<P extends ElementPath>(
...
+  } else {
+    if (reparentTargetParentIsElementPath(second)) {
+      return EP.closestSharedAncestor(first.elementPath, second, true)
+    } else {
+      if (EP.pathsEqual(first.elementPath, second.elementPath)) {
+        if (first.clause === second.clause) {
+          return first
+        } else {
+          return first.elementPath
+        }
```
Rheeseyb, line 88: Can you add a comment in here please because although it's clear what is being returned, it's not clear why that is being returned

Rheeseyb (approved):

ruggi (approved):

PR #3518, opened 2023-04-05
Convert to group/frame dropdown

bkrmendy:


editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -46,14 +63,88 @@ function isAbsolutePositionedFrame(
...
 }
 
-function convertFrameToGroupCommands(
+// when removing the parent from between the children and the grandparent
+function offsetChildrenByVectorCommands(
+  childInstances: Array<ElementInstanceMetadata>,
+  offset: CanvasPoint | LocalPoint,
+) {
+  return childInstances.flatMap((child) =>
+    setElementTopLeft(child, {
```
balazsbajorics, line 72: this should set whichever exists from left, top, right bottom. and make sure there's at least one in each dimension. 
there's something like this in [createMoveCommandsForElement](https://github.com/concrete-utopia/utopia/blob/caa448ae6aa3e7583fda4d7c3281f4788b6e2e93/editor/src/components/canvas/canvas-strategies/strategies/shared-move-strategies-helpers.ts#L217) 

using this helper https://github.com/concrete-utopia/utopia/blob/caa448ae6aa3e7583fda4d7c3281f4788b6e2e93/editor/src/components/canvas/canvas-strategies/strategies/shared-move-strategies-helpers.ts#L350-L377

balazsbajorics (approved):

PR #3524, opened 2023-04-05
Always reparse after a print

Rheeseyb:
**Problem:**
We are gradually enabling the addition and manipulation of JS Expressions directly in the parsed model, which will always require a re-parsing of the printed code result (to capture the required scope for function caching and evaluation, as well as triggering some other behaviour that is tied to the parsing such as capturing the highlight bounds). However, so far we've been dealing with this in a very ad-hoc way (instructing the editor to print and then re-parse is very specific cases), meaning we are by default missing any cases that aren't explicitly marked as requiring a re-parse (one specific example of this - changes that affect a conditional will lose the highlight bounds for it).

**Fix:**
Always reparse after a print. This isn't as expensive as we imagined, as we're already triggering 2 minor parses of the printed code for the sake of attempting to update the highlight bounds, which we now no longer need to do.

This fix removes the changes from #3121 and #3082 around `RevisionsState.ParsedAheadNeedsReparsing`, as that is now obsolete.


gbalint:
Nice, lots of deleted code :)

gbalint (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -124,7 +124,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`464`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`488`)
```
balazsbajorics, line 127: hmm, is this expected? does it worth understanding what rerenders on the UI?

balazsbajorics (approved):

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -165,6 +165,7 @@ export interface EditorRenderResult {
   getRecordedActions: () => ReadonlyArray<EditorAction>
 }
 
+// FIXME This should be calling formatTestProjectCode
```
seanparsons, line 168: Is there a reason to not do that?

seanparsons (approved):

editor/src/components/canvas/ui-jsx.test-utils.tsx
Rheeseyb: This is only because I want to make that change in a separate PR so that it doesn't pollute this diff

editor/src/components/canvas/dom-walker-caching.spec.browser2.tsx
```diff
@@ -124,17 +124,12 @@ describe('Dom-walker Caching', () => {
...
       'storyboard-entity/scene-1-entity/app-entity:app-outer-div',
       'storyboard-entity/scene-1-entity/app-entity:app-outer-div/card-instance:card-outer-div',
       'storyboard-entity/scene-2-entity/same-file-app-entity:same-file-app-div',
     ])
-    expect(saveDomReportActions[2].cachedPaths).toEqual([
-      EP.fromString('storyboard-entity/scene-2-entity/same-file-app-entity:same-file-app-div'),
-      EP.fromString('storyboard-entity/scene-2-entity/same-file-app-entity'),
-      EP.fromString('storyboard-entity/scene-2-entity'),
-    ])
+    expect(saveDomReportActions[2].cachedPaths).toEqual([])
```
Rheeseyb, line 132: This is because we invalidate the DOM walker cache as a whole when we detect a parsed model update coming from the parser-printer, so this is also the reason for the render count tests bump

editor/src/core/performance/performance-regression-tests.spec.tsx
Rheeseyb: I've investigated this and answered in a comment above about the dom walker caching tests, as that is the cause

PR #3545, opened 2023-04-14
Fix/unwrap shortcut elements conditionals

enidemi:
**Problem:**
Unwrap shortcut/context menu item is not doing anything even on simple elements.

**Fix:**
The main problem is that the original unwrapping worked only on Views. Now it is replaced with checking if the element support children or not, and of course it is enabled on Conditionals keeping the active branch.
It's possible to unwrap conditionals like this too: `{true ? 'hello' : 'hi'}`

**Commit Details:**
- rename 'UNWRAP_GROUP_OR_VIEW' to 'UNWRAP_ELEMENT'
- remove flag 'onlyForGroups' from action types, since unwrapping should work on all kind of elements
- instead of the number of children check if the target can contain children
- children were reversed after unwrapping, this PR fixes that and covers it with tests too
- tests for conditionals
- tests for different elements, including fragments


editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -271,4 +271,155 @@ describe('actions', () => {
       })
     })
   })
+  describe('UNWRAP_GROUP_OR_VIEW', () => {
+    it(`Unwraps a group`, async () => {
+      const testCode = `
+				<div data-uid='aaa' style={{contain: 'layout', width: 300, height: 300}}>
```
balazsbajorics, line 277: the indentation police is here, they are asking about these tabs in the sample code 👮‍♀️
image

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -271,4 +271,155 @@ describe('actions', () => {
       })
     })
   })
+  describe('UNWRAP_GROUP_OR_VIEW', () => {
+    it(`Unwraps a group`, async () => {
```
balazsbajorics, line 275: I think it'd be safer to rename this to "Unwraps a content-affecting element" or "Unwraps an unstyled div" because what a group is is under heavy development.

editor/src/components/editor/actions/actions.spec.browser2.tsx
enidemi: I don't know what they are talking about 🙃

editor/src/components/editor/actions/actions.spec.browser2.tsx
enidemi: fixed

bkrmendy (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2809,70 +2815,115 @@ export const UPDATE_FNS = {
...
+        if (parentPath != null && reparentTargetParentIsConditionalClause(parentPath)) {
+          return unwrapConditionalClause(editor, action.target, parentPath)
+        }
+
+        if (elementIsContentAffecting) {
+          if (isTextContainingConditional(action.target, editor.jsxMetadata)) {
+            return unwrapTextContainingConditional(editor, action.target, dispatch)
+          }
+
+          // this movetemplate uses commands!
```
Rheeseyb, line 2871: This is fantastic news but I don't think we need a comment about it 😀 unless I'm missing something?

PR #3546, opened 2023-04-14
is this thing even needed?

balazsbajorics:
will this break any tests?


balazsbajorics:
ok at least _some_ tests broke, thank goodness. still, very very bad test coverage, must be fixed

PR #3549, opened 2023-04-17
Update Nix and Node JS versions

Rheeseyb:
**Problem:**
The version of node.js we have been using is old and needed updating in order to build the latest version of VS Code.

**Fix:**
Update the node.js version to v16.18.1, and update nix to the latest stable version (22.11).
The reason for the nix update is because the previous version of node we were using was the latest v16 version available on the older nix release.

Updating nix revealed that a package we were using on the server for the AWS API ([amazonka](https://hackage.haskell.org/package/amazonka)) had been effectively abandoned and marked as broken, so that has been replaced with a new package ([aws](https://hackage.haskell.org/package/aws)). Whilst we were there we also updated the version of the haskell compiler so that we could continue to pull cached builds from [hydra](https://hydra.nixos.org/) as part of the server build (without that we'd have to build every single package in the dependency chain, making the build very long and painful).

Updating Node introduced / flagged the following issues that are resolved in this PR:
- An error `Error: error:0308010C:digital envelope routines::unsupported` is thrown by webpack builds due to its use of a crypto function now considered to be legacy. The workaround for this is to use the node flag `--openssl-legacy-provider` pretty much everywhere. Further details [here](https://www.newline.co/@kchan/how-to-fix-the-error-errorerror0308010cdigital-envelope-routinesunsupported--0f8d3f17).
- The updated version of pnpm seems to have introduced some change that was causing ts-loader resolution issues in webpack builds, so we had to replace an `exclude` rule with an explicit `include` rule in the editor's webpack config file for that loader. I'm not entirely sure why this only affects the dependency on `utopia-api` and not `utopia-vscode-common`.
- It seems that by using `pnpx tsc` in places we were using a different version of `tsc` to the one coming from the `typescript` package specified in the `package.json`, so those cases have been replaced with an explicit node script instead.
- `patch-package` was causing issues with pnpm, and is no longer necessary as [pnpm has native support for patching](https://pnpm.io/cli/patch), so the patches and `package.json` have been updated to reflect that.
- The custom install steps we had for `utopia-api` were not only unnecessary, but were actively causing issues during the build, so those have been removed now and replaced with an install and build step in the `install-editor` script used everywhere.


editor/patches/@emotion__react@11.1.2.patch
```diff
@@ -1,7 +1,7 @@
-diff --git a/node_modules/@emotion/react/dist/emotion-react.browser.cjs.js b/node_modules/@emotion/react/dist/emotion-react.browser.cjs.js
+diff --git a/dist/emotion-react.browser.cjs.js b/dist/emotion-react.browser.cjs.js
```
Rheeseyb, line 1: All of the patches are file renames (to match the style used by pnpm's native patching support to make updating the patches later easier), and the removal of `/node_modules/<package_name>` from the file paths used in the diffs (like has happened here), so you can ignore the patch diffs

editor/jest.config.js
```diff
@@ -116,7 +116,7 @@ module.exports = {
         '\\.[jt]sx?$': 'babel-jest',
       },
       roots: ['src', 'node_modules', '<rootDir>/node_modules'],
-      transformIgnorePatterns: ['/node_modules/(?!utopia-api)'], // this lets ts-jest work on `/node_modules/utopia-api` which is a simlink to `../utopia-api`.
+      transformIgnorePatterns: ['<rootDir>/node_modules/.pnpm/(?!file+..+)'],
```
Rheeseyb, line 119: I'm honestly not sure how this hadn't broken before, but it seems like we were on the edge of this breaking since we first switched to pnpm

editor/package.json
```diff
@@ -14,37 +14,36 @@
     "clean": "rm -rf lib/",
     "check": "tsc --noEmit && pnpm run lint && pnpm run check-dependency-cruiser && pnpm run prettier-check && pnpm run test && pnpm run test-browser",
     "check-ci": "tsc --noEmit && pnpm run lint && pnpm run check-dependency-cruiser && pnpm run prettier-check && pnpm run test-ci && pnpm run test-browser",
-    "check-code": "tsc --noEmit && pnpm run lint && pnpm run check-dependency-cruiser && pnpm run prettier-check",
-    "lint": "pnpx eslint \"src/**/*.{js,ts,tsx}\" --quiet",
-    "lint-loud": "pnpx eslint \"src/**/*.{js,ts,tsx}\"",
+    "check-code": "tsc --version && tsc --noEmit && pnpm run lint && pnpm run check-dependency-cruiser && pnpm run prettier-check",
+    "lint": "eslint \"src/**/*.{js,ts,tsx}\" --quiet",
```
Rheeseyb, line 18: `pnpx` had been changed [over a year ago](https://github.com/pnpm/pnpm/pull/4267) to become an alias for [`pnpm dlx`](https://pnpm.io/cli/dlx), which downloads and runs the latest version of a package, rather than using the locally installed version

shell.nix
```diff
@@ -45,51 +30,60 @@ let
   ]);
 
   baseEditorScripts = [
+    (pkgs.writeScriptBin "install-utopia-api" ''
+      #!/usr/bin/env bash
+      set -e
+      cd $(${pkgs.git}/bin/git rev-parse --show-toplevel)/utopia-api
+      pnpm install
+      pnpm run build
+    '')
```
Rheeseyb, line 39: This was added to replace the script previously living in the editor's `package.json`, which was causing build issues

shell.nix
```diff
@@ -14,29 +14,14 @@ let
...
-  };
-  pnpm = "${pnpmPkg}/lib/node_modules/pnpm/bin/pnpm.cjs";
-  pnpx = "${pnpmPkg}/lib/node_modules/pnpm/bin/pnpx.cjs";
-
-  yarn = "${pkgs.nodePackages_latest.yarn}/lib/node_modules/yarn/bin/yarn.js";
+  nodePackages = [
+    node
+    pnpm
+    (yarn.override { nodejs = node; })
+  ];
```
Rheeseyb, line 24: This is synching up the version of node used by pnpm and yarn

maltenuhn:
FWIW
* I've been able to build the project with this change, even on my cursed machine
* I've had this branch open all day and making theme-related changes, including occasional full reloads, and everything worked as expected

bkrmendy (approved):

editor/package.json
balazsbajorics: wow, good to know!

editor/package.json
```diff
@@ -14,37 +14,36 @@
...
     "parse-print-analysis": "npx -s sh ts-node --require ignore-styles --transpile-only ./src/scripts/parse-print-analysis.ts",
     "parse-debug": "npx -s sh ts-node --pretty --require ignore-styles --transpile-only ./src/scripts/parse-debug.ts",
     "fetch-tailwind-classes": "npx -s sh ts-node --pretty --require ignore-styles --transpile-only ./src/scripts/fetch-and-parse-tailwind-classes.ts && npx prettier --write ./src/core/third-party/tailwind-defaults.ts",
     "lint-editor": "lint-staged",
     "dev-fast": "APP_ENV=development REACT_APP_ENVIRONMENT_CONFIG=development REACT_APP_AUTH0_CLIENT_ID=$AUTH0_CLIENT_ID REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=`git rev-parse HEAD` vite --force",
-    "vite-build": "node --max_old_space_size=16384 ./node_modules/vite/bin/vite build -- -DEBUG"
+    "vite-build": "node --max_old_space_size=16384 ./node_modules/vite/bin/vite build -- -DEBUG",
+    "watch-tsc": "tsc --watch && NODE_OPTIONS='--max_old_space_size=4096 --openssl-legacy-provider'"
```
balazsbajorics, line 46: I think this will also make iTerm 2's cmd + click to open VSCode work properly!

editor/package.json
```diff
@@ -14,37 +14,36 @@
...
     "production-print-json": "pnpm run production -- --json=lib/production-stats.json",
-    "staging": "REACT_APP_ENVIRONMENT_CONFIG=staging REACT_APP_AUTH0_CLIENT_ID=$AUTH0_CLIENT_ID REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=$UTOPIA_SHA WEBPACK_MODE='staging' NODE_OPTIONS=--max_old_space_size=6656 webpack",
+    "staging": "REACT_APP_ENVIRONMENT_CONFIG=staging REACT_APP_AUTH0_CLIENT_ID=$AUTH0_CLIENT_ID REACT_APP_AUTH0_ENDPOINT=$AUTH0_ENDPOINT REACT_APP_AUTH0_REDIRECT_URI=$AUTH0_REDIRECT_URI REACT_APP_COMMIT_HASH=$UTOPIA_SHA WEBPACK_MODE='staging' NODE_OPTIONS='--max_old_space_size=6656 --openssl-legacy-provider' webpack",
     "staging-print-json": "pnpm run staging --  --json=lib/staging-stats.json",
-    "install-utopia-api": "cd ../utopia-api && pnpm install && pnpm run build && pnpm run dts-bundle",
-    "preinstall": "npx only-allow pnpm && pnpm run install-utopia-api",
-    "postinstall": "patch-package",
```
balazsbajorics, line 39: nice!

editor/webpack.config.js
```diff
@@ -241,7 +241,15 @@ const config = {
       { test: /graceful-fs/, use: 'null-loader' },
       // Match typescript
       {
-        exclude: /node_modules(?!\/utopia-api)/,
+        include: (filePath) => {
+          if (filePath.includes('node_modules')) {
+            // We need to use the ts-loader to load the utopia-api module,
+            // but nothing else from node_modules
+            return filePath.includes('/utopia-api/')
```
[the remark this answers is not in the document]
balazsbajorics: I didn't know you can write a function here instead of a regex. this makes it much easier to maintain! bonus points for the comments!

balazsbajorics (approved):

PR #3552, opened 2023-04-18
Always add longhand fragments with React imported

bkrmendy:
## Problem
When inserting a longhand fragment, it's necessary to ensure that `React` is imported, otherwise the affected components cannot be rendered (since React.Fragment is not defined). Also, inserting a shorthand fragment (`<></>`) can be problematic in projects that use a framework that has a custom implementation for it, which might lead to unintended behaviour.

## Solution
When programmatically inserting a fragment, always insert `<React.Fragment>` with `React` imported. This PR refactors all occurrences of inserting fragments to follow the above principle.


editor/src/components/shared/project-components.ts
```diff
@@ -352,8 +352,8 @@ const conditionalElementsDescriptors: ComponentDescriptorsForFile = {
         elementToInsert: jsxConditionalExpressionWithoutUID(
           jsExpressionValue(true, emptyComments),
           'true',
-          jsExpressionValue(null, emptyComments),
-          jsExpressionValue(null, emptyComments),
+          jsExpressionValue('True branch', emptyComments),
+          jsExpressionValue('False branch', emptyComments),
```
Rheeseyb, line 356: Was this intended to be part of this PR, or is this PR branching off another piece of work?

Rheeseyb:
The code to add missing imports looks good to me, but the default text in conditionals needs to be separated out of this PR. Once that is done this looks good to me 👍

balazsbajorics (approved):
As Rheese said.

Rheeseyb (approved):

PR #3555, opened 2023-04-18
Fix misplaced conditionals in the navigator root

### LINKED ISSUE #3554: Fix misplaced conditionals in navigator root

ruggi:
Fixes #3554 

**Problem:**

Conditional elements at the root of the navigator would be displayed always at the top regardless of their position in the actual tree.

For example, this tree
```
div1
conditional
div2
```

would result in the navigator displaying
```
conditional
div1
div2
```

**Fix:**

Use the (ordered) project tree as a base when populating the navigator items.

### DIFF
--- a/editor/src/components/editor/store/store-deep-equality-instances.ts
+++ b/editor/src/components/editor/store/store-deep-equality-instances.ts
@@ -171,6 +171,9 @@ import {
   importedOrigin,
   ConditionValue,
   JSXConditionalExpressionWithoutUID,
+  isJSXConditionalExpression,
+  JSXConditionalExpression,
+  jsxConditionalExpression,
 } from '../../../core/shared/element-template'
 import {
   CanvasRectangle,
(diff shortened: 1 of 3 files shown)


bkrmendy (approved):

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -1035,6 +1038,8 @@ export function JSXElementChildKeepDeepEquality(): KeepDeepEqualityCall<JSXEleme
       return JSXTextBlockKeepDeepEquality(oldElement, newElement)
     } else if (isJSXFragment(oldElement) && isJSXFragment(newElement)) {
       return JSXFragmentKeepDeepEquality(oldElement, newElement)
+    } else if (isJSXConditionalExpression(oldElement) && isJSXConditionalExpression(newElement)) {
```
seanparsons, line 1041: Oooh, that's upsetting. Not the fix, but that this issue was in there.

ruggi: Yeah it's... puzzling, it'd be nice to have the compiler yell at us there

seanparsons: We could change this to a switch/case statement, which would make this slightly better, just not as nice as we still have to check both types.

editor/src/components/navigator/navigator-utils.ts
```diff
@@ -179,7 +179,20 @@ export function getNavigatorTargets(
     }
   }
 
-  const canvasRoots = MetadataUtils.getAllStoryboardChildrenPathsUnordered(metadata)
+  function getCanvasRoots(trees: ElementPathTree[]): ElementPath[] {
```
seanparsons, line 182: I wonder if this code should become an implementation of `getAllStoryboardChildrenPathsOrdered`.

seanparsons (approved):

PR #3567, opened 2023-04-20
only show conditional inspector when a conditional is selected

bkrmendy:
# Try it [here](https://utopia.pizza/p/ac9485df-indecisive-peridot/?branch_name=feature-hide-inspector-sections-when-conditional-selected)

## Description
This PR takes the first step towards separating control structures from DOM elements in the inspector, by hiding all inspector sections other than conditional when a conditional is selected.

### DIFF
--- a/editor/src/components/inspector/inspector.tsx
+++ b/editor/src/components/inspector/inspector.tsx
@@ -248,12 +248,13 @@ export const Inspector = React.memo<InspectorProps>((props: InspectorProps) => {
   const onlyConditionalsSelected = useEditorState(
     Substores.metadata,
     (store) =>
+      store.editor.selectedViews.length > 0 &&
       store.editor.selectedViews.every((path) =>
         MetadataUtils.isConditionalFromMetadata(
           MetadataUtils.findElementByElementPath(store.editor.jsxMetadata, path),
         ),
       ),
-    'Inspector isConditionalSelected',
+    'Inspector onlyConditionalsSelected',
   )
 
   React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


ruggi (approved):

editor/src/components/inspector/inspector.tsx
```diff
@@ -348,29 +360,37 @@ export const Inspector = React.memo<InspectorProps>((props: InspectorProps) => {
...
           }}
         >
-          <AlignmentButtons numberOfTargets={selectedViews.length} />
+          {unless(
+            isConditionalSelected,
+            <AlignmentButtons numberOfTargets={selectedViews.length} />,
+          )}
           {when(isTwindEnabled(), <ClassNameSubsection />)}
           {anyComponents ? <ComponentSection isScene={false} /> : null}
           <ConditionalSection paths={selectedViews} />
```
Rheeseyb, line 369: Hmmm should this only be rendered when a conditional is selected? This feels like the fork in the Inspector logic should be higher up unless there's a reason for it to live here.

This also raises questions about multiselection - e.g. right now if I multiselect a conditional and a DOM element we end up with a very broken looking Inspector

bkrmendy: > should this only be rendered when a conditional is selected
as far as I know, the Inspector function is basically the outermost place where this logic can be placed.

Regarding the broken looking conditional control, thanks for spotting it! I refactored the code so that all inspector sections are shown when both a DOM element and a conditional is selected (since the conditional control already handles multiselection, and the gist of the PR is preserved when only a single conditional is selected)

gbalint (approved):

PR #3570, opened 2023-04-21
Better toasts

maltenuhn:
**Problem:**
Our toast visibility was both too low (from afar) and too visually distracting (from up close)

**Fix:**
Redesigned toasts. 

- Lower levels (info, warning) now look v similar to each other
- Higher levels (primary, notice, error) look more pronounced and get icons
- Shadows blend in more, and x-to-dismiss is substantially de-emphasized

NB this does not address (prior) z-index issues (eg toasts should be below toolbar), pending the other changes to editor layout.

image

### DIFF
--- a/editor/src/components/common/notices.tsx
+++ b/editor/src/components/common/notices.tsx
@@ -27,6 +27,18 @@ export const getStylesForLevel = (level: NoticeLevel): React.CSSProperties => {
   return resultingStyle
 }
 
+export const getPrefixForLevel = (level: NoticeLevel): string => {
+  let resultingPrefix = ''
+  if (level === 'WARNING') {
+    resultingPrefix = '﹗'
+  } else if (level === 'ERROR') {
+    resultingPrefix = '⚠️'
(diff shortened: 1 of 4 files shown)


editor/src/components/common/notices.tsx
```diff
@@ -27,6 +27,18 @@ export const getStylesForLevel = (level: NoticeLevel): React.CSSProperties => {
...
+  if (level === 'WARNING') {
+    resultingPrefix = '﹗'
+  } else if (level === 'ERROR') {
+    resultingPrefix = '⚠️'
+  } else if (level === 'SUCCESS') {
+    resultingPrefix = '✓'
+  }
+  return resultingPrefix
+}
+
```
bkrmendy, line 41: a switch/case with returns and an `assertNever` in the default case would be more ideal (since it automatically breaks when the `NoticeLevel` type changes), but this is more of a nice-to-have

bkrmendy (approved):

enidemi (approved):

Issue #1580, opened 2021-07-23
When wrapping in element in 'preserve position' mode, we should understand the layout of the fresh wrapper element


balazsbajorics:
the floating-insert-menu 's wrapping functionality now uses two disgusting functions named `getIsFlexBasedOnName_KILLME_EXPERIMENTAL` and `getIsFlexDirectionBasedOnName_KILLME_SERIOUSLY_EXPERIMENTAL` 

we need an _approach_ to fix this. 

the challenge is that we should somehow know upfront before insertion what kind of layout this mystic wrapper element provide for us

maltenuhn:
@balazsbajorics is this still current? Should we close it?

balazsbajorics:
@maltenuhn I believe it is still current! 

Wrap In... is kind of like a reparenting. and when you reparent across layout systems, now on master we have automatic layout prop conversion. We will probably build something more sophisticated for Strategies.


but I assume even the more sophisticated layout system converter will need to know what is the layout system of the new parent – hence this ticket.

the idea is that somehow during insertion we should figure out the layout of the newly inserted (wrapper) parent, ideally before we draw on the screen.

PR #3571, opened 2023-04-21
Fix/wrap in actions

enidemi:
**Problem:**
There are some problems with the wrap-in actions. First is there are 2 of them, mostly the same code but one of them is not updated that much for example only one handles fragments. Inside the wrap-in update where the insertion happens it is hard to follow, the nested if-elses makes it quite error prone. The result is often not good, for example wrapping in elements with flexGrow just makes them zero size, or wrapping even just absolute element causes shifts.

**Fix:**
Remove `wrapInView`, only keep `wrapInElement`. Updated `wrapInElement` to use insertion commands for simple jsx elements, and replaced the if-elses with switch cases, it's easier to see if all insertion target cases are handled, next time we add a new type of InsertionPath this will break.

Next step is the wrapper element frame/position etc. based on where the insertion happens. The current PR keeps the default divs and elements as wrappers, so it's still not good.

**Commit Details:**
- added basic tests for wrap-in simple elements
- found tests for `wrapInView` context-menu item too, I grouped them into one `describe`
- renamed `unwrap-conditional-helpers` to `wrap-unwrap-helpers`, and moved the wrapping insertion here.
- removed `getIsFlexBasedOnName_KILLME_EXPERIMENTAL` and its friends from the floating-menu


bkrmendy (approved):

editor/src/components/canvas/ui/floating-insert-menu.tsx
```diff
@@ -160,26 +159,6 @@ function useGetInsertableComponents(): InsertableComponentFlatList {
...
 }
 
-function getIsFlexBasedOnName_KILLME_EXPERIMENTAL(name: JSXElementName): boolean {
-  return (
-    name.propertyPath.propertyElements.length === 0 &&
-    (name.baseVariable === 'FlexRow' || name.baseVariable === 'FlexCol')
-  )
-}
-
-function getIsFlexDirectionBasedOnName_KILLME_SERIOUSLY_EXPERIMENTAL(
```
Rheeseyb, line 170: https://github.com/concrete-utopia/utopia/issues/1580!

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -1121,4 +1127,111 @@ describe('actions', () => {
       )
     })
   })
+  describe('WRAP_IN_ELEMENT', () => {
+    it(`Wraps in 2 elements`, async () => {
```
Rheeseyb, line 1131:
```suggestion
    it(`Wraps 2 elements`, async () => {
```

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -1121,4 +1127,111 @@ describe('actions', () => {
...
+          `<div data-uid='aaa' style={{contain: 'layout', width: 300, height: 300}}>
+            <div style={{backgroundColor: '#aaaaaa33', position: 'absolute'}} data-uid='${testUID}'>
+              <div data-uid='ccc' style={{position: 'absolute', left: 20, top: 50, bottom: 150, width: 100}} />
+              <div data-uid='ddd' style={{width: 60, height: 60}} />
+            </div>
+          </div>`,
+        ),
+      )
+    })
+    it(`Wraps in 2 elements inside a flex layout`, async () => {
```
Rheeseyb, line 1164:
```suggestion
    it(`Wraps 2 elements inside a flex layout`, async () => {
```

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -1121,4 +1127,111 @@ describe('actions', () => {
...
+              <div style={{backgroundColor: '#aaaaaa33', position: 'absolute'}} data-uid='${testUID}'>
+                <div data-uid='ddd' style={{flexGrow: 1, height: '100%'}} />
+                <div data-uid='eee' style={{width: 100, height: 60}} />
+              </div>
+            </div>
+          </div>`,
+        ),
+      )
+    })
+    it(`Wrap in 2 elements with a fragment`, async () => {
```
Rheeseyb, line 1203:
```suggestion
    it(`Wrap 2 elements with a fragment`, async () => {
```

editor/src/components/editor/actions/wrap-unwrap-helpers.tsx
```diff
@@ -0,0 +1,417 @@
...
+            editor,
+            staticTarget,
+            elementToInsert,
+            importsToAdd,
+          )
+          return { updatedEditor: withTargetAdded, newPath: newPath }
+        default:
+          const _exhaustiveCheck: never = staticTarget
+          return { updatedEditor: editor, newPath: null }
+      }
```
Rheeseyb, line 289: It looks like this block of code is identical for the 3 cases?

Rheeseyb (approved):

editor/src/components/canvas/ui/floating-insert-menu.tsx
enidemi: yes, it's time to fix that :)

PR #3572, opened 2023-04-21
Prevent elements from snapping to their content-affecting containers

bkrmendy:
## Problem
When dragging an element, it snaps to its parent even if it's a content-affecting element

## Fix
Make snap target selection more explicit, which makes it possible to special-case content-affecting parents in a maintainable way.

Also, fake metadata tests from `absolute-move-strategy.spec.tsx` are ported to `absolute-move-strategy.spec.browser2.tsx`. `it(works with a TL pinned absolute element with px values and snapping` and `it(works with TL pinned absolute elements in multiselection`  are omitted, since the same use cases are already tested in `absolute-move-strategy.spec.browser2.tsx`.


enidemi (approved):

ruggi (approved):

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.spec.browser2.tsx
```diff
@@ -163,6 +184,420 @@ function positioningFromCss(css: CSSStyleDeclaration) {
...
+                style={{ backgroundColor: '#0091FFAB', position: 'absolute', left: 100, top: 100, width: 250, height: 300 }}
+                data-uid='ccc'
+              />
+            </View>
+          </View>
+`,
+        ),
+      )
+    })
+  })
```
ruggi, line 600: I would suggest to refactor these to be table tests, but not necessarily for now :)

ruggi (approved):

enidemi (approved):

PR #3575, opened 2023-04-21
Highlight bounds for expressions

seanparsons:
**Problem:**
As we now support expressions in places like the clauses of conditional expressions, those do not usefully highlight in the code editor.

**Fix:**
There have been several fixes done in the process of supporting this feature, with everything other than the first entry being a consequence of what has been needed to change:
- Highlight bounds were included in the parser for expressions with the  `ATTRIBUTE_*` prefixed types.
- It was necessitated to have some logic then ignore `data-uid` attributes, because those are added within the parse at times and we don't want the logic that compares the UIDs of every value to all the highlight bounds as the newly added attributes will not have highlight bounds.
- So that we have the maximal highlight bounds, `simplifyAttributeIfPossible` was removed from the parser logic, which then resulted in a lot of test failures because countless tests relied on the simplified structure. `simplifyParsedTextFileAttributes` is used to handle these cases by applying that simplification much later.
- It was discovered that a long standing and never noticed bug in the parser was inadvertently fixed by these changes. The bug being that in a second pass parse, the UIDs found during the first pass were being carried over, causing UIDs to be regenerated for some complicated cases like the `elementsWithin` of an arbitrary block which was an arbitrary block as it contained a class component. The Babel transpilation in master was failing as a result, which oddly caused preferable code to be emitted for these cases. An inelegant fix implemented for this was to pass an empty array to `elementsWithin` parameter of `transpileJavascript` if the `rootLevel` flag has been passed as true.
- With maximal highlight bounds came the result that highlight bounds were being identified that could not be selected, as a result `trimHighlightBounds` was implemented to remove those which could not be selected from being included in the parse result.

**Commit Details:**
- Added `simplifyJSXElementAttributes`, `simplifyJSXElementChildAttributes`, `simplifyElementsWithinAttributes`, `simplifyArbitraryJSBlockAttributes`, `simplifyTopLeveElementAttributes` and `simplifyParsedTextFileAttributes` to help simplifying the attribute representation which is no longer done by the parser. These are used to allow us to keep the old test values in the place without needing to update them.
- In `updateSelectedComponentsFromEditorPosition` remove `regularNavigatorEntryOptic` from the optic to lookup paths so that we can include the synthetic navigator entries and anything that might be in them.
- `simplifyAttributeIfPossible` maintains the `uid` values from the previous instance where possible.
- `simplifyAttributeIfPossible` is now no longer called during the parse, so that we end up with the maximal level of highlight bounds.
- `jsxArbitraryBlock` and `jsxTextBlock` can now be supplied with a `uid` directly.
- Changed parameter of `getUtopiaIDFromJSXElement` to support `JSXElementChild`.
- In lots of the parsing, the highlight bounds are merged with `mergeHighlightBounds` so as not to lose entries.
- `parseOtherJavaScript` now includes the highlight bounds for the entire value.
- Renamed `createExpressionValue` to `createRawExpressionValue`, so that `createExpressionValue` could be created in line with similar functions already there for other types.
- `createExpressionNestedArray`, `createExpressionNestedObject` and `createExpressionFunctionCall` now return `WithParserMetadata<X>` as they build the highlight bounds for those entities.
- Added `createArbitraryJSBlock`, `createJSXArbitraryBlock` and `createJSXTextBlock` utility functions to be used in place of their regular factory functions.
- `parseAttributeExpression` now correctly handles the various parts of `WithParserMetadata`.
- Added `buildHighlightBoundsForUids`, `mergeHighlightBounds`, `addBoundsIntoWithParser` and `addToHighlightBounds` utility functions to help avoid random custom snippets of similar code in the parser.
- `parseOutJSXElements` handles the case where some spacing elements are removed by also removing their highlight bounds.
- Implemented `getChildrenWithoutWhitespaceOnlyText` to handle removing the highlight bounds for the whitespace text being excluded and to also extract that logic from two near identical bits of code in the parser.
- Reworked the handling of `WithParserMetadata` for conditional expressions, as it wasn't carrying over the highlight bounds correctly.
- Prevented `parsedElementsWithin` from being passed to `transpileJavascript` in `parseArbitraryNodes`, which corrects a case which has been introduced as elements are now no longer getting their UIDs regenerated with something new.
- Made `getBoundsOfNodes` return a null bounds in the case of an empty array of nodes instead of throwing an exception.
- Upgraded `walkElements` to cover some cases that weren't being covered, as well as having it support ignoring the `data-uid` attribute, along with supporting that in functions like `ensureAribtraryBlocksHaveUID`.
- Added `trimHighlightBounds` to give us more practical highlight bounds, as it's undesirable to drill down into properties as we need those things which are selectable to be included only.
- `lintAndParse` now optionally calls `trimHighlightBounds`, where in most cases we want to trim them for practical reasons but for testing that we obtain them all appropriately we do not trim them.


editor/src/components/canvas/ui-jsx-canvas.spec.tsx
```diff
@@ -1405,8 +1405,12 @@ export var ${BakedInStoryboardVariableName} = (props) => {
...
             data-uid=\\"scene-aaa\\"
           >
-            <div id=\\"nasty-div\\" data-uid=\\"ce9~~~1\\" data-path=\\"d1b~~~2/ce9~~~1\\">
-              huhahuha
+            <div
+              id=\\"nasty-div\\"
+              data-uid=\\"aaa\\"
+              data-path=\\"utopia-storyboard-uid/scene-aaa/app-entity:aaa\\"
+            >
+              huha huha
```
Rheeseyb, line 1413: Oh my, why was this previously getting rendered without a space?!

seanparsons: Good question.

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1165,47 +1214,154 @@ function createExpressionOtherJavaScript(
...
+  sourceFile: TS.SourceFile,
+  originalJavascript: string,
+  javascript: string,
+  transpiledJavascript: string,
+  definedElsewhere: Array<string>,
+  sourceMap: RawSourceMap | null,
+  elementsWithin: ElementsWithin,
+  alreadyExistingUIDs: Set<string>,
+): JSXArbitraryBlock {
+  // Ideally the value we hash is stable regardless of location, so exclude the SourceMap value from here and provide an empty UID.
```
Rheeseyb, line 1334: Should this also be true for `createArbitraryJSBlock` above?

seanparsons: Yes, yes I think it should.

Rheeseyb (approved):

ruggi (approved):

balazsbajorics (approved):

PR #3581, opened 2023-04-24
Fix/reparent wiggle 2

bkrmendy:
## Problem
The reparent outline in the navigator is shown in a confusing way.

## Fix 

The drop target lines/parent outline should work according to the following:
- The parent outline should always be active on the new parent of a dragged item. This implies that it should always be visible during the drag and drop interaction (either when dropping the element being dragged "into" another element, or when inserting the dragged element between two siblings).
- The drop target lines should be active if
    - the dragged item will be dropped between two siblings
    - the dragged item is positioned between two siblings, but the pointer is dragged to the side so that the dragged item is reparented into higher up the hierarchy.

We also support moving the pointer sideways to reparent up the hierarchy (the "reparent wiggle"). This should work like so:
- we only support moving the cursor left to reparent up the hierarchy. The opposite direction, reparenting down the hierarchy is not supported.
- when a reparent line is shown, the wiggle can be activated by dragging the pointer to the left by a preset amount (20px currently)
- when the wiggle is active, the target line is shown in its original place, and the parent outline is shown on the parent-to-be, further up the hierarchy
- this will reparent the dragged element so that it becomes the first sibling among the children of its new parent

## TODO: tests

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item-components.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item-components.tsx
@@ -41,35 +41,41 @@ export const NavigatorHintTop = React.forwardRef<HTMLDivElement, NavigatorHintPr
         <div
           style={{
             opacity: props.shouldBeShown ? 1 : 0,
-            marginLeft: props.margin,
             position: 'absolute',
             top: -6,
             width: '100%',
-            display: 'flex',
-            flexDirection: 'row',
-            alignItems: 'center',
-            justifyContent: 'flex-start',
             height: 16,
           }}
         >
           <div
             style={{
-              backgroundColor: colorTheme.navigatorResizeHintBorder.value,
-              height: 2,
-              flexGrow: 1,
+              marginLeft: props.margin,
+              height: '100%',
+              display: 'flex',
+              flexDirection: 'row',
+              alignItems: 'center',
+              justifyContent: 'flex-start',
             }}
-          />
-          <div
-            style={{
-              position: 'absolute',
-              backgroundColor: colorTheme.bg0.value,
-              width: NavigatorHintCircleDiameter,
-              height: NavigatorHintCircleDiameter,
-              contain: 'layout',
-              border: `2px solid ${colorTheme.navigatorResizeHintBorder.value}`,
-              borderRadius: '50%',
-            }}
-          />
+          >
+            <div
+              style={{
+                backgroundColor: colorTheme.navigatorResizeHintBorder.value,
+                height: 2,
+                flexGrow: 1,
+              }}
+            />
+            <div
+              style={{
+                position: 'absolute',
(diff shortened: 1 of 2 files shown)


editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
```diff
@@ -945,7 +985,7 @@ describe('conditionals in the navigator', () => {
       await dispatchDone
     })
 
-    act(() =>
+    await act(() =>
```
balazsbajorics, line 988: does this need to be wrapped in an act at all? other places dragElement is not wrapped

editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
```diff
@@ -113,6 +118,9 @@ function dragElement(
       }),
     )
 
+    await wait(1) // FIXME: this might be necessary for react-dnd internal delays
+    await midDragCallback()
```
balazsbajorics, line 122: isn't the problem here that the code above is not using using an async/await act for wrapping fireEvent? see for example https://github.com/concrete-utopia/utopia/blob/03dec098e333408a5f80ac410b38fe64d2345458/editor/src/components/canvas/event-helpers.test-utils.tsx#L67-L80

bkrmendy: thanks, that worked, added it in https://github.com/concrete-utopia/utopia/pull/3581/commits/7b9ee78eb61f30e683189edfd6011ac7ebc5abe0! (Good to know this little gotcha 🙃)

However I still needed to add a `wait` in the test for sideways drag (to reparent up the hierarchy)

editor/src/components/navigator/navigator.spec.browser2.tsx
```diff
@@ -765,31 +858,50 @@ describe('Navigator', () => {
...
       ).toEqual([
         'regular-utopia-storyboard-uid/scene-aaa',
+        'regular-utopia-storyboard-uid/scene-aaa/dragme', // <- moved to under the grandparent
         'regular-utopia-storyboard-uid/scene-aaa/sceneroot',
         'regular-utopia-storyboard-uid/scene-aaa/sceneroot/firstdiv',
         'regular-utopia-storyboard-uid/scene-aaa/sceneroot/seconddiv',
         'regular-utopia-storyboard-uid/scene-aaa/sceneroot/thirddiv',
         'regular-utopia-storyboard-uid/scene-aaa/sceneroot/notdrag',
         'regular-utopia-storyboard-uid/scene-aaa/parentsibling',
-        'regular-utopia-storyboard-uid/dragme', // <- moved to under the grandparent
```
balazsbajorics, line 792: hah! a lying comment!

balazsbajorics (approved):

gbalint (approved):

PR #3582, opened 2023-04-24
Prettify code before tests

Rheeseyb:
**Problem:**
This is a follow up to #3524, where it was discovered that the code used in browser tests should be prettified before running them, as otherwise a re-print and re-parse could alter the model.

**Fix:**
I've added a function `formatAllCodeInModel` which will traverse the project contents and call `formatTestProjectCode` on any code files, which is then used at the start of `renderTestEditorWithModel`. Then I've fixed up some of the tests that check the before and after code in cases where re-printing and re-parsing aren't triggered.

### DIFF
--- a/editor/src/components/assets.ts
+++ b/editor/src/components/assets.ts
@@ -331,7 +331,7 @@ export function walkContentsTree(
   })
 }
 
-interface PathAndFileEntry {
+export interface PathAndFileEntry {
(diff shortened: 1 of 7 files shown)


bkrmendy (approved):
🎉

enidemi (approved):

PR #3586, opened 2023-04-25
`a` to insert in conditional slot

bkrmendy:
## Problem
When inserting an element into a conditional slot from the floating insert menu, the inserted element isn't selected after insertion.

## Fix
The action used to perform the insertion (`INSERT_INSERTABLE`) wasn't aware of the distinction between normal element paths and conditional slots, so it simply appended the UID of the new element to the target parent path, which resulted in an invalid element path. The insertion still succeeded because `insertJSXElementChild_DEPRECATED` checked the target parent path, and special cased conditional slots.

To fix the problem, `insertJSXElementChild_DEPRECATED` was swapped out for its new version, `insertJSXElementChild`. The behaviour of `insertJSXElementChild` is controlled by the `InsertionPath` passed in, so the logic to check if the target parent is a conditional slot was naturally propagated up to `INSERT_INSERTABLE`. The same logic can be used to create the element path for the newly inserted element, so the fix simply "fell out" of the refactor.

### DIFF
--- a/editor/src/core/model/conditionals.ts
+++ b/editor/src/core/model/conditionals.ts
@@ -4,6 +4,7 @@ import {
   ElementInstanceMetadata,
   ElementInstanceMetadataMap,
   isJSXConditionalExpression,
+  isNullJSXAttributeValue,
   JSXConditionalExpression,
   JSXElementChild,
 } from '../shared/element-template'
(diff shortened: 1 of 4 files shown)


editor/src/core/model/conditionals.ts
```diff
@@ -168,6 +168,18 @@ export function maybeBranchConditionalCase(
   }
 }
 
+export function getConditionalCaseFromMetadata(
```
gbalint, line 171: This looks very similar to `getConditionalCase`, maybe the two can be unified?

ruggi (approved):

gbalint (approved):

seanparsons (approved):

editor/src/core/model/conditionals.ts
bkrmendy: `getConditionalCase` seems to be unused, so I'm rolling it into `getConditionalCaseFromMetadata`

bkrmendy: I'd rather not touch the deprecated function (since that'll be refactored out anyway with time and I don't want to risk breaking unrelated functionality).

`getConditionalCase` is not used anywhere, and the problem it solves isn't relevent to `getConditionalCaseFromMetadata`: `getConditionalCase` was originally written to determine whether a given branch is the currently active branch of a conditional, and `getConditionalCaseFromMetadata` is to determine if a branch is in the true or the false branch. I deleted `getConditionalCase` in https://github.com/concrete-utopia/utopia/pull/3586/commits/8ef4a20560099baed38c631a1667698f286da3f7 and renamed `getConditionalCaseFromMetadata` to `getConditionalCaseCorrespondingToBranchPath`

balazsbajorics (approved):

Rheeseyb (approved):

PR #3587, opened 2023-04-25
render count test mega snapshots

balazsbajorics:
**Problem**
When the render count test goes up, it's really hard to tell why, as it is a single number.

**Fix**
Store a full list of the "path" of all re-rendered elements, and check them against a snapshot.

**Note**
The snapshot is massive and not super useful by itself. But when the render counts change, the Jest diff should be pretty useful, as it should only highlight the elements that are doing extra re-renders.

**Note 2**
I hope that this will not cause a ton of extra need to update the snapshots, but if it's too bad, we can always disable these checks.


seanparsons (approved):

editor/src/utils/canvas-react-utils.ts
```diff
@@ -26,9 +26,11 @@ export function applyUIDMonkeyPatch(): void {
   }
 }
 
-function getDisplayName(type: any): string {
+export function getDisplayName(type: any): string {
   // taken from https://github.com/facebook/react/blob/7e405d458d6481fb1c04dfca6afab0651e6f67cd/packages/react/src/ReactElement.js#L415
-  if (typeof type === 'function') {
+  if (type == null) {
+    return 'Unknown'
```
Rheeseyb, line 32:
```suggestion
    return 'null'
```

PR #3589, opened 2023-04-25
Require a system test before automatically triggering a deploy

Rheeseyb:
**Problem:**
The automated deploy runs every morning regardless of the state of the staging environment, unless we explicitly deactivate it.

**Fix:**
Separate out the automated triggering of the Tag and Deploy workflow so that it can run a system test against the Staging environment before actually tagging and deploying the project.


.github/workflows/tag-release.yml
```diff
@@ -8,12 +8,12 @@ on:
       commit-ref:
         description: 'Branch or SHA to tag (defaults to master)'
         required: false
-  schedule:
-    - cron: '0 5 * * 1-5' # Run every weekday 5:00 UTC
+  workflow_call:
```
ruggi, line 11: I may be wrong, but I'm not sure this is needed 🤔

ruggi (approved):

seanparsons (approved):

.github/workflows/tag-release.yml
Rheeseyb: You are wrong 😀 https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_call
The `workflow_dispatch` event is for manually triggering the workflow

ruggi: aha! TIL

Rheeseyb: Me too!

PR #3592, opened 2023-04-25
Clicking conditional clause in the Navigator toggles override

Rheeseyb:
## [Try it here](https://utopia.pizza/p/aa429bab-glimmer-midnight/?branch_name=spike-nav-condition-label-overrides)

**Problem:**
Often when selecting an element in the inactive clause of a conditional we want to actually view that element in the context of the running application. Not only that, but we want a quick and clear way to toggle overrides on conditionals without having the select the conditional itself

**Fix:**
Add override toggling behaviour to the Navigator. The following two very similar behaviours have been tweaked:

1) When clicking the label of a clause of a conditional in the Navigator, we will now either clear the override (if the clause is the default outcome), or toggle the override (if the clause is not the default outcome). This gives a quick and intuitive way to toggle the override without changing the current selection. As a side note, I've also changed the cursor when the mouse is over the label to indicate that it has a different behaviour (since it no longer updates the selection)
2) When clicking the _child_ of a clause in the Navigator, we will either clear the override (if the child is in the default branch), or set the override (if the child is not in the default branch). Note that in this case we don't toggle the override, we just either clear it or set it (or leave it unchanged if no action was necessary). This gives an intuitive way to select and view the inactive case in the context of the running application, in a way that wouldn't break with double clicks (or cause accidental toggling)

To support this I've had to add the default case to the metadata, which I've done by changing the `ConditionValue` type to be an object containing an `active` and a `default` field.

navigator-conditional-override-3


editor/src/components/navigator/navigator-conditionals.spec.browser2.tsx
```diff
@@ -1679,3 +1620,229 @@ describe('conditionals in the navigator', () => {
...
+      'await-first-dom-report',
+    )
+
+    // The true case is the default but not the active case
+    await clickChildOfClause('true-div', false, renderResult)
+
+    // No override should have been added
+    expect(getPrintedUiJsCode(renderResult.getEditorState())).toEqual(codeWithoutOverride)
+  })
+})
```
ruggi, line 1848: these tests are :chef-kiss:

ruggi (approved):
nice stuff :)

(one nit would be to unify `isActiveBranchOfConditional` and `isDefaultBranchOfConditional` into a single function since their logic is ~the same, but that's pure personal preference talking)

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -81,38 +86,66 @@ export interface NavigatorItemInnerProps {
...
+): Array<EditorAction> {
+  switch (conditionalOverrideUpdate) {
+    case 'no-update':
+      return []
+    case 'clear-override':
+      return [EditorActions.setConditionalOverriddenCondition(targetPath, null)]
+    case 'true-case':
+      return [EditorActions.setConditionalOverriddenCondition(targetPath, true)]
+    case 'false-case':
+      return [EditorActions.setConditionalOverriddenCondition(targetPath, false)]
```
seanparsons, line 124: This looks like it could do with a default with `assertNever` in it.

seanparsons (approved):

PR #3600, opened 2023-04-26
Navigator: separate components for entries

bkrmendy:
## Problem
Currently we have 3 different kinds of navigator entries (regular elements, conditional slots, and TRUE/FALSE labels for conditionals), and we only have 1 component to render these entries. An unintended consequence of this is that a lot drag-and-drop related code has become quite complex, since react-dnd works on a per-component basis, and the one-size-fits-all approach of having a single navigator entry component has become quite error-prone.

## Fix
This PR adds a separate navigator entry component for each of the navigator entry types. This allows the drag-and-drop code to be driven by the specific type of component it's in, so a lot of the defensive code in drag/drop helpers becomes unnecessary.

### Limitations
It could be possible to go even deeper and refactor the functions that are called from the navigator components and use `navigatorEntry`, since the navigator components already "know" what kind of entry they represent, so a lot of business logic can be pushed up from action handlers.


ruggi (approved):

seanparsons (approved):
Nice!

PR #3614, opened 2023-04-27
end-to-end paste test

bkrmendy:


editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -1171,6 +1176,105 @@ describe('actions', () => {
...
+        const setClipboardDataStub = sandbox.stub(Clipboard, 'setClipboardData')
+        setClipboardDataStub.callsFake(async (c) => {
+          mockClipBoard.data = encodeUtopiaDataToHtml(c!.data)
+        })
+      })
+
+      afterEach(() => {
+        sandbox.restore()
+        mockClipBoard.data = null
+      })
```
balazsbajorics, line 1206: if this works out, you could move the beforeEach and afterEach and the variables to a separate function (in a helper file) so multiple tests can share it

bkrmendy: extracted it in [491c2ab](https://github.com/concrete-utopia/utopia/pull/3614/commits/491c2ab567b0ef0cde8ea8b3ce4c9602ddcc67a7)

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -1171,6 +1176,87 @@ describe('actions', () => {
...
+        await selectComponentsForTest(renderResult, [makeTargetPath('aaa')])
+
+        const canvasRoot = renderResult.renderedDOM.getByTestId('canvas-root')
+
+        FOR_TESTS_setNextGeneratedUids(['child1', 'child2', 'parent'])
+
+        firePasteEvent(canvasRoot)
+
+        // Wait for the next frame
+        await clipboardMock.pasteDone
```
ruggi, line 1215: I think it'd be more ergonomic to make this the return value of `firePasteEvent` and use it as
```ts
await firePasteEvent(clipboardMock, canvasRoot)
```

(you could introduce another helper that ties the two together but imho it'd be redundant)

Even if there were cases where you don't want to `await` right after the call, you could always save the promise and use it independently

ruggi (approved):

editor/src/components/editor/actions/actions.spec.browser2.tsx
bkrmendy: The promise is here because Clipboard.parseClipboardData is originally an async function (since it might read files from the clipboard, which is an async operation), and the pasteDone promise basically exposes this fact. It's not ideal that an implementation detail (Clipboard.parseClipboardData being async) is exposed, but since one is already knowingly mocking the clipboard functions this isn't that big a deal

balazsbajorics (approved):

PR #3618, opened 2023-04-28
Maintain element UIDS, especially when stripUIDs is set.

seanparsons:
**Problem:**
When inserting a new element on production and staging, the newly created element isn't selected as expected and there appear to be no matching highlight bounds for it.

As we now print and parse on a change to the parsed model, when `stripUIDs` is set to true (as it is in production and staging), the editor has become a lot more reliant on the logic in `fixParseSuccessUIDs` than it was previously.

It was also discovered that in some cases the `data-uid` attribute didn't match up correctly with the `uid` property on instances of `JSXElement` after being processed by `fixParseSuccessUIDs`.

**Fix:**
Making the `data-uid` attribute and `uid` property consistent involved some refactoring to `fixJSXElementUIDs` and `updateUID`. `updateUID` will now generate a new UID in the case of a clash that couldn't otherwise be resolved, then once that has changed update the `data-uid` attribute.

Otherwise there are fixes which revolve around trying to keep UIDs where possible, like having `fixArrayElements` look for matching elements by UID first, then fallback to index lookups.

**Commit Details:**
- Removed `WorkerCodeAndParsedUpdate.highlightBounds` as it wasn't used.
- Removed `PrintAndReparseResult.highlightBounds` as it wasn't used.
- Slightly tweaked `elementsStructure` to have it produce more output.
- Added `validateParsedTextFileElementUIDs` and supporting functions to check if the `data-uid` attribute and `uid` property are consistent.
- Changed two property tests to have them permit UIDs to be copied from one type of `JSXElementChild` to another.
- Slightly refactored the type of `FixUIDsState.mappings` for clarity.
- `updateUID` now doesn't give up on the case where a UID is used elsewhere in the new structure.
- `fixArrayElements` is now more intelligent, handling shifting array elements by UID first before falling back to checking indexed entries.
- Functions like `fixTopLevelElementUIDs` now support the old element being null or undefined.
- `fixJSXElementUIDs` now better ensures that changes to the `uid` property are also mirrored with changes to the `data-uid` attribute.


editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -452,58 +507,87 @@ export function fixJSXElementChildUIDs(
...
-  oldExpression: JSExpression,
+  oldExpression: JSExpression | null | undefined,
   newExpression: JSExpression,
   fixUIDsState: FixUIDsState,
 ): JSExpression {
   switch (newExpression.type) {
     case 'ATTRIBUTE_VALUE': {
-      if (oldExpression.type === newExpression.type) {
+      if (oldExpression == null) {
+        return newExpression
```
bkrmendy, line 582: minor, but I think this could be extracted to the top of the function (since it's repeated in all the case branches)

seanparsons: Good shout, fixed in 44f39ba48.

bkrmendy (approved):

ruggi (approved):

PR #3620, opened 2023-04-28
Improvements/insert menu

maltenuhn:
**Problem:**
The insert menu in the right-hand column was painful to use, and far too long

**Fix:**
Condensed things, and also grouped fragments and conditionals in a 'Code' group.
image
image

### DIFF
--- a/editor/src/components/shared/project-components.ts
+++ b/editor/src/components/shared/project-components.ts
@@ -92,23 +92,13 @@ export function insertableComponentGroupHTML(): InsertableComponentGroupHTML {
   }
 }
 
-export interface InsertableComponentGroupConditionals {
-  type: 'CONDITIONALS_GROUP'
+export interface InsertableComponentGroupCodeConstructs {
+  type: 'CODE_CONSTRUCTS_GROUP'
 }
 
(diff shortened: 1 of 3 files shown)


editor/src/components/shared/project-components.ts
```diff
@@ -92,23 +92,13 @@ export function insertableComponentGroupHTML(): InsertableComponentGroupHTML {
   }
 }
 
-export interface InsertableComponentGroupConditionals {
```
maltenuhn, line 95: I grouped conditionals and fragments into one `CodeConstructs` bucket, hope that won't cause issues elsewhere?

seanparsons (approved):

editor/src/components/shared/project-components.ts
Rheeseyb: Looks fine to me 👍

Rheeseyb (approved):
Approved on the assumption the snapshot tests will be updated before merging (they look fine to me based on these changes)

maltenuhn:
Closing so the superior PR #3640 can replace it (once it has arrow navigation)

PR #3623, opened 2023-04-28
Chore: remove insertElementAtPath_DEPRECATED from runInsertElementInsertionSubject

### LINKED ISSUE #3622: Remove insertElementAtPath_DEPRECATED from runInsertElementInsertionSubject

ruggi:
Fixes #3622 

Removed `insertElementAtPath_DEPRECATED` from `runInsertElementInsertionSubject`.

Also, the PR refactors `INSERT_INSERTABLE` to receive an `InsertionPath` instead of building one internally.


bkrmendy (approved):

editor/src/components/canvas/commands/insert-element-insertion-subject.ts
```diff
@@ -59,10 +59,19 @@ export const runInsertElementInsertionSubject: CommandFunction<InsertElementInse
...
+        targetParent,
         editor.projectContents,
-        underlyingFilePath,
-        childInsertionPath(targetParent),
+        editor.nodeModules.files,
+        editor.canvas.openFile?.filename,
+        editor.jsxMetadata,
+      )
+      if (insertionPath == null) {
+        return // maybe this should throw instead?
```
gbalint, line 70: Definitely not. That would mean the editor throws an exception when you try to insert into a non-empty clause.

gbalint: I'm not sure where this action is used from, but let me note that this way we don't have automatic wrapping to a fragment when inserting into a non-empty clause, it will just silently do nothing.
I guess depending on the actual feature we should either
(1) show a toast
(2) implement fragment insertion with a new type of InsertionPath
(3) just leave it as it is

If it turns out that this function is used in all different contexts and this can not be decided, I would change the action object to store the InsertionPath directly, and then the call sites can decide how to call the action creator.

balazsbajorics: > I would change the action object to store the InsertionPath directly, and then the call sites can decide how to call the action creator.

in general I think this is very important to keep this in mind! 
this is a very fuzzily defined task, but we are basically looking for the "highest" place where we can put the slot-like behavior. the ideal is to put it in code that "talks" about behavior / ux. so in this case, that should be the strategies, and not the commands. the commands should be low-level executors, not place where behavior lives.

reality is way fuzzier than this, and there's sometimes not a good solution. but something to keep in mind

ruggi: I think that yeah, providing the insertion path in the strategies (which use this one) would be the preferable solution. Will adjust

editor/src/components/canvas/canvas-strategies/strategies/drag-to-insert-metastrategy.spec.browser2.tsx
```diff
@@ -100,8 +100,9 @@ async function dragFromInsertMenuDivButtonToPoint(
   await renderResult.getDispatchFollowUpActionsFinished()
 }
 
-describe('Dragging from the insert menu into an absolute layout', () => {
-  const inputCode = makeTestProjectCodeWithSnippet(`
+describe('drag-to-insert', () => {
```
Rheeseyb, line 103: Why the nesting here?

editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-metastrategy.tsx
```diff
@@ -656,3 +667,11 @@ function runTargetStrategiesForFreshlyInsertedElementToResize(
 
   return foldAndApplyCommandsInner(editorState, [], resizeCommands, commandLifecycle).statePatches
 }
+
+export function getRootPath(startingMetadata: ElementInstanceMetadataMap): ElementPath | null {
```
Rheeseyb, line 671: Please can you rename this function to `getStoryboardPath` and move it into `element-metadata-utils.ts`

editor/src/components/canvas/canvas-strategies/strategies/drag-to-insert-metastrategy.tsx
```diff
@@ -343,8 +347,7 @@ function runTargetStrategiesForFreshlyInsertedElement(
   }>,
   strategyLifeCycle: InteractionLifecycle,
 ): Array<EditorStatePatch> {
-  const storyboard = MetadataUtils.getStoryboardMetadata(editorState.jsxMetadata)
-  const rootPath = storyboard != null ? storyboard.elementPath : elementPath([])
+  const rootPath = getRootPath(editorState.jsxMetadata) ?? EP.elementPath([])
```
Rheeseyb, line 350: I know this is based on the previous code, but I'm not sure we should protect against there being no storyboard path here, because the empty element path fallback would be completely wrong. Maybe it's worth having this throw like you've done above?

editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-strategy.spec.browser2.tsx
```diff
@@ -168,10 +164,133 @@ function isIndicatorBetweenSiblingsBBBCCC(
   }
 }
 
+const testDragToInsertImageAspectRatio = async (inputCode: string, expectedCode: string) => {
```
Rheeseyb, line 167:
```suggestion
const testDrawToInsertImageAspectRatio = async (inputCode: string, expectedCode: string) => {
```

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -2889,6 +2896,55 @@ export const SizeKeepDeepEquality: KeepDeepEqualityCall<Size> = combine2Equality
...
   createCallWithTripleEquals()
 
+export const ChildInsertionPathKeepDeepEquality: KeepDeepEqualityCall<ChildInsertionPath> =
+  combine1EqualityCall(
+    (c) => c.intendedParentPath,
+    StaticElementPathKeepDeepEquality,
+    childInsertionPath,
+  )
+
+export function ConditionalCaseKeepDeepEquality(): KeepDeepEqualityCall<ConditionalCase> {
```
seanparsons, line 2906:
```suggestion
export const ConditionalCaseKeepDeepEquality: KeepDeepEqualityCall<ConditionalCase> = createCallWithTripleEquals<ConditionalCase>()
```

editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-metastrategy.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/3623/commits/53a9ce32850d54b16b0b2310e7e40895caac1bdb

editor/src/components/canvas/canvas-strategies/strategies/drag-to-insert-metastrategy.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/3623/commits/90f446e0e1a2b21313327ee2c6cb0dfb339585f9

editor/src/components/editor/store/store-deep-equality-instances.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/3623/commits/68e28a72e1a9a590834d3cb497e75bddaa9a1706

gbalint (approved):

PR #3624, opened 2023-04-28
Design: better toolbar and shadows

maltenuhn:
**Problem:**
- Our toolbar has a weird ordering and info hierarchy (w/ zoom at the top). It does, however, have good shadows
- Our strategy indicator looks ok, but had simple shadows that did not convey its true depth (of character, and otherwise)

**Fix:**
- Move zoom back into tools
- Use the standard drop shadow everywhere

(some design debt here @lankaukk: we do not have a good, layered shadow system yet. Our `theme` shadows are poor and now hardly used anywhere. Nothing to fix right now, but something to be aware of)

image

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/strategy-indicator.tsx
+++ b/editor/src/components/canvas/controls/select-mode/strategy-indicator.tsx
@@ -27,6 +27,7 @@ export const StrategyIndicator = React.memo(() => {
   return (
     <FlexRow
       style={{
+        ...UtopiaStyles.popup,
         pointerEvents: 'none',
         position: 'absolute',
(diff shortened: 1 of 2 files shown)


bkrmendy (approved):

lankaukk:
I might consider removing all of the `paddingBottom: 6` from the individual `FlexCol`s and instead add a `gap: 6` to their parent

lankaukk (approved):

maltenuhn:
> I might consider removing all of the `paddingBottom: 6` from the individual `FlexCol`s and instead add a `gap: 6` to their parent

Indeed, nice spot!

PR #3626, opened 2023-05-02
fix(wrap) when wrapping in root elements show toast instead of throwing errors

enidemi:
**Problem:**
When wrapping in root elements nothing happens on the canvas. There are errors in the console.

**Fix:**
Show toast when wrapping in a root div

**Commit Details:**
- update `WRAP_IN_ELEMENT`

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -2355,8 +2355,12 @@ export const UPDATE_FNS = {
             EP.isRootElementOfInstance(elementPath) &&
             EP.isParentOf(getElementPathFromInsertionPath(parentPath), elementPath),
         )
-        if (anyTargetIsARootElement && targetThatIsRootElementOfCommonParent == null) {
-          return editor
+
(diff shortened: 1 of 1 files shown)


editor/src/components/editor/actions/actions.tsx
```diff
@@ -2355,8 +2355,12 @@ export const UPDATE_FNS = {
             EP.isRootElementOfInstance(elementPath) &&
             EP.isParentOf(getElementPathFromInsertionPath(parentPath), elementPath),
         )
-        if (anyTargetIsARootElement && targetThatIsRootElementOfCommonParent == null) {
```
bkrmendy, line 2358: `targetThatIsRootElementOfCommonParent == null` is this not necessary anymore?

bkrmendy (approved):

editor/src/components/editor/actions/actions.tsx
enidemi: yes, this needs a real fix soon

PR #3631, opened 2023-05-02
Fix/convert elements to fragments

enidemi:
**Problem:**
Trying to convert an element to fragment or convert a fragment to element results in no changes. Not just in conditional clauses but for non special elements too.

**Fix:**
Update the floating menu to dispatch actions when trying to convert an element to fragment, this was missing before. Change action to update from fragments and to fragments too. 

**Commit Details:**
- update floating-insert-menu
- added tests
- disable 'convert' context menu and 'c' shortcut on conditionals


editor/src/components/editor/action-types.ts
```diff
@@ -786,7 +786,7 @@ export interface InsertImageIntoUI {
 export interface UpdateJSXElementName {
   action: 'UPDATE_JSX_ELEMENT_NAME'
   target: ElementPath
-  elementName: JSXElementName
+  elementName: { type: 'JSX_ELEMENT'; name: JSXElementName } | { type: 'JSX_FRAGMENT' }
```
seanparsons, line 789: This makes me think the whole type needs a better name.

seanparsons (approved):

bkrmendy (approved):

PR #3633, opened 2023-05-03
Fix/bring-to-front-twice

bkrmendy:
## Problem
When clicking "Bring Forward" on an element that is already the last child, we throw an exception, and bring to front/send to back are totally broken within conditionals.

## Problem
Refactor the related code so that
- if bring to front/send to back is clicked on the root of a conditional branch, nothing happens (and no undo frame is generated)
- if the selected element is already the last element and it's moved forward, or it's the first element and it's moved backward, nothing happens (and no undo frame is generated)
- the code involved in moving elements forward/backward is refactored so that it uses `insertJSXElementChild` (so we work down some tech debt)
- test coverage is added for the relevant context menu items and shortcuts


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -2292,9 +2293,8 @@ export function moveTemplate(
                     updatedUtopiaComponents,
                   )
 
-                  const insertResult = insertElementAtPath_DEPRECATED(
+                  const insertResult = insertJSXElementChild(
                     workingEditorState.projectContents,
-                    workingEditorState.canvas.openFile?.filename ?? null,
                     childInsertionPath(underlyingNewParentPath),
```
ruggi, line 2298: I guess this should be updated as well if the parent is a conditional (I suppose the `getInsertionPathWithSlotBehavior` helper could be used here)

Rheeseyb (approved):
Very nice tests!

editor/src/components/canvas/canvas-utils.ts
bkrmendy: great idea, done in https://github.com/concrete-utopia/utopia/pull/3633#:~:text=use%20getInsertionPathWithSlotBehavior-,1bc4bf4,-Merge%20state

PR #3641, opened 2023-05-04
Fix reparent of children of sizeless div

gbalint:
**Problem:**
1. Create a new default project
2. Select the utopia logo
3. Press x to convert to absolute
4. Drag it outside of its original boundary.

The image will be reparented outside of its parent, and it is going to jump around, its offset will be incorrect (note: the same thing happens when the image is wrapped into a conditional)

**Root cause:**
`getReparentPropertyChanges` calculates the offset using the `specialSizeMeasurements.globalContentBoxForChildren` frames.
`globalContentBoxForChildren` is a solution developed to solve the sizing issue of group-like elements which do not have intrinsic size.
However, these frames are incorrect for the parent div of the image, because the div is sizeless and gets it `globalContentBoxForChildren` from its parent (which is the root div of the `Playground` component)
Because of this the original parent (the parent div of the image), and the new parent (the root div of Playground) have the same `globalContentBoxForChildren`, so there will be no offsetting during reparent. But the (0, 0) point of the new parent is in a different position, so the image will jump.

There are multiple problems with our approach:
1. There is no guarantee that the `globalContentBoxForChildren` has the same (0,0) point as the coordinate system of the element (because we treat even absolute positioned elements sizeless divs if they don't have explicit width _or_ height, see https://github.com/concrete-utopia/utopia/blob/cff6b8b616340c2b6f52ac6ccdfb1bc7813ef160/editor/src/core/model/element-metadata-utils.ts#L2080 )
2. In `fillSpyOnlyMetadata` we calculate the `globalFrame` and `localFrame` from children, but we do not do that for ``globalContentBoxForChildren`, see: https://github.com/concrete-utopia/utopia/blob/cff6b8b616340c2b6f52ac6ccdfb1bc7813ef160/editor/src/core/model/element-metadata-utils.ts#L2138 )

I point fixed this in this PR by getting `globalContentBoxForChildren` from the DOM metadata when available.


bkrmendy (approved):

ruggi (approved):

editor/src/core/model/element-metadata.spec.browser2.tsx
```diff
@@ -266,6 +266,101 @@ describe('globalContentBoxForChildren calculation', () => {
...
+
+      const childInstance = MetadataUtils.findElementByElementPath(
+        editor.getEditorState().editor.jsxMetadata,
+        elementPathInInnards(`a7b/b15/b0e`),
+      )
+      if (childInstance == null) {
+        throw new Error('childInstance should not be null')
+      }
+
+      const globalContentBoxForChild = MetadataUtils.getGlobalContentBoxForChildren(childInstance)
```
balazsbajorics, line 353: may be nitpick, but img cannot have children, so globalContentBoxForChild on the img is probably not the best example use case! for example in the future if we find a better dom API instead of [isElementAContainingBlockForAbsolute](https://github.com/concrete-utopia/utopia/blob/e3cd990b4915af11f736cb3ab6f4ef898f4c2a73/editor/src/components/canvas/dom-walker.ts#L119) then this may become (rightfully!) the globalFrame of the containingBlock parent

PR #3642, opened 2023-05-04
Navigator reparent to flex

bkrmendy:
## Problem
When reparenting an absolute positioned element into a flex container via the navigator, the element does not become a flex sibling. Instead, the element is absolutely positioned in a weird way above the flex siblings.

## Solution
The problem above is already fixed with copy/paste. This PR extracts code behind copy/paste, and makes the navigator reorder action call the same code.

This PR also adds tests that check layout property changes after reparenting via the navigator. Since a lot of functionality is still far from ideal (e.g., when reparenting from flex to absolute, no absolute positioning props are applied), these tests are primarily there as documentation.

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -2920,6 +2920,7 @@ export const UPDATE_FNS = {
           action.pasteInto,
           'always', // TODO Before merge make sure this is the right pick here
           'use-new-insertJSXElementChild',
+          null,
         )
 
         if (outcomeResult == null) {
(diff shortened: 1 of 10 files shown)


editor/src/components/editor/actions/actions.tsx
```diff
@@ -5604,31 +5572,66 @@ export function isSendPreviewModel(action: any): action is SendPreviewModel {
...
+    indexPosition,
+  )
+
+  if (outcomeResult == null) {
+    return null
   }
+
+  const { commands: reparentCommands, newPath } = outcomeResult
+
+  const reparentStrategy = reparentStrategyForPaste(
```
Rheeseyb, line 5611: I guess `reparentStrategyForPaste` should probably be renamed now. I'm not sure what to. Perhaps something like `reparentStrategyForStaticReparent`? Naming is hard...

Rheeseyb (approved):

editor/src/components/editor/actions/actions.tsx
bkrmendy: https://github.com/concrete-utopia/utopia/pull/3642/commits/3f275e2492b58a8d448f576334869917ef09b791 
image

enidemi (approved):

PR #3646, opened 2023-05-05
Improve duplicate uid handling.

seanparsons:
**Problem:**
Duplicate UIDs were still arising from certain actions, which was then causing the editor to end up effectively in a broken state, but in a way that wasn't necessarily obvious to the user.

**Cause:**
When performing actions like duplication, the attributes (which now have uids of their own internally) were not being updated and ended up clashing with the original, even though the uids of the elements had been updated.

Also, the handling of existing uids wasn't multi-element safe. The collation of existing uids was done once, but then multiple elements were inserted, where the generation of uids ended up clashing between those newly inserted elements.

**Fix:**
The four main parts to this PR are as follows:
- Changing how the duplicate UIDs are surfaced, so that instead of breaking the editor behind the scenes with the user mostly none the wiser, a big error toast is displayed indicating the user should restart the editor.
- Identifying duplicate UIDs is improved where previously attributes weren't being included and those are very likely to cause clashes because entities often have common values like `'absolute'`, which end up hashing to the same UID.
- Fixing of duplicates mostly centering around `fixUtopiaElement` is now multi-insert safe, as well as including the UIDs of attribute values in it's handling of the duplicate checks. Also as highlight bounds were recently trimmed to only include those things which can be readily selected currently, the full highlight bounds are now stored so that we can make use of the keys as a shortcut for checking what already exists.
- More handling of the mappings from when UIDs changed have been added, for example with `guaranteeUniqueUidsFromTopLevel`.

**Commit Details:**
- `duplicate` simplified and as a result now more comprehensively handles the various types it may encounter.
- `PASTE_JSX_ELEMENTS` is now multi-insert safe for the generation of uids.
- `editorDispatchInner` when running in a browser displays a toast indicating an internal issue should duplicates occur, otherwise in tests it will throw an exception so that errors can be correctly attributed to the test that caused them.
- `fullHighlightBounds` was added to `ParseSuccess`, so that when checking for existing uids the keys from that can be used instead of having to walk all the elements and their attributes.
- `getAllUniqueUids` now returns the currently unique uids, those which are duplicates and a collection of both.
- `getAllUniqueUids` was switched to using a switch-case statement so that changes to the model will be flagged up as breaking changes.
- `getAllUniqueUids` now correctly includes attributes and their uids.
- `getAllUniqueUids` also includes the arbitrary block attached to a component as well as `ARBITRARY_JS_BLOCK` cases.
- `isJSExpression` added as a convenience because it's easier to find by name.
- `generateConsistentUID` now always falls back to generating UIDs from the letters a-z, instead of only when `possibleStartingValue` had a length greater than or equal to 3.
- `fixUtopiaExpression` implemented as a small wrapper around `fixUtopiaElement`.
- `fixUtopiaElement` now includes attributes in the fixing pass that it performs.
- `fixUtopiaElement` assigns a uid to the `data-uid` property entry that is also generated to be consistent and unique.
- `fixUtopiaElement` now also fixes the `condition` property of a conditional expression.
- `fixParseSuccessUIDs` also now updates the newly added `fullHighlightBounds` property.
- `updateUID` now always adds the `uidToUse` to `mutableAllNewUIDs`.
- Added some small utility functions for Sets.
- `getAllUniqueUids` now has a "path" that allows for debugging, to help identify where for example a duplicate UID comes from.
- `fixUtopiaElement` returns mappings for UIDs that have changed.
- Added `fromArrayIndex` and `identityOptic` optics.
- Extracted out `updateHighlightBounds` so that it can be used in multiple places.
- `fixUtopiaElement` ensures to add a mapping when the uid for the `data-uid` prop changes.
- `guaranteeUniqueUidsFromTopLevel` now returns mappings for changes that have been made.
- `parseCode` maintains the existing UIDs from the changes that happen when it invokes `guaranteeUniqueUidsFromTopLevel`.


editor/karma-base.conf.js
```diff
@@ -1,4 +1,6 @@
-process.env.CHROME_BIN = require('puppeteer').executablePath() // Puppeteer v19.6.0 uses Chromium 110.0.5479.0
+if (!('CHROME_BIN' in process.env)) {
+  process.env.CHROME_BIN = require('puppeteer').executablePath() // Puppeteer v19.6.0 uses Chromium 110.0.5479.0
+}
```
bkrmendy, line 3: Does this mean we will be able to run the tests with an arbitrary Chrome executable? Why should we have this rather than a fixed Chrome version?

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -742,6 +745,34 @@ function editorDispatchInner(
...
+          notice(
+            `Utopia has suffered from an irrecoverable error, please reload the editor.`,
+            'ERROR',
+            true,
+          ),
+        )
+        result = {
+          ...result,
+          unpatchedEditor: UPDATE_FNS.ADD_TOAST(errorToast, result.unpatchedEditor),
+        }
```
bkrmendy, line 772: I think there're better ways to go about this. We should either keep only the toast, and write a pair of tests that assert that a toast is shown when duplicate uids are present, and that the toast is not shown otherwise. If we want to keep the exception, we should but this code into an object export and stub it with Sinon to throw an exception. I think interleaving production code with test code muddles up the implementation, and introduces a big side effect in a function that could otherwise be pure

bkrmendy (approved):

editor/karma-base.conf.js
seanparsons: Yes, with the default behaviour being to use the fixed Chrome version supplied by Puppeteer.

The reason why this is beneficial is a long standing issue of npm packages that assume the particular binary they've bundled will work on every variant of the platform. In the case of Linux that's a big assumption as it usually means the bundle expects things like glibc to live in one particular directory, which is not always the case and results in Chrome insta-crashing.

editor/src/components/editor/store/dispatch.tsx
seanparsons: Previously the logic was always throwing the exception, which was not necessarily obvious from the call being made.

If we can stub it to throw an exception like that for both unit and browser tests, that sounds like it would be an improvement however.

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -742,6 +750,35 @@ function editorDispatchInner(
...
+        throw new Error(errorMessage)
+      } else {
+        // When running in the browser log the error and tell the user to restart the editor.
+        console.error(errorMessage)
+        const errorToast = EditorActions.addToast(
+          notice(
+            `Utopia has suffered from an irrecoverable error, please reload the editor.`,
+            'ERROR',
+            true,
+            'reload-editor',
```
bkrmendy, line 772: can we make this error message contain the cause of the error? image trying to report or debug this based on this description

seanparsons: Well we do `console.error` the detail, I'm just somewhat trying to avoid this becoming a big pile of static to a regular user.

editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.spec.browser2.tsx
```diff
@@ -166,7 +176,184 @@ export var ${BakedInStoryboardVariableName} = (
 )`
 }
 
+function getElementByDataUID(renderResult: EditorRenderResult, dataUID: string): HTMLElement {
+  const queryByDataUID = queryHelpers.queryByAttribute.bind(null, 'data-uid')
+  return forceNotNull(
+    `Could not find element with ${dataUID}`,
+    queryByDataUID(renderResult.renderedDOM.container, dataUID),
+  )
+}
```
Rheeseyb, line 185: Very nice! I can see this being pulled out into one of our `.test-utils` files in the very near future

editor/src/components/editor/store/dispatch.tsx
Rheeseyb: I don't want to block this PR since it needs to be merged, but what I'm thinking I might do after it has been merged is update this error message to say something like "Check the console for more details"

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -65,7 +65,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`456`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`454`)
```
Rheeseyb, line 68: Ooooh!

ruggi (approved):

Rheeseyb (approved):

PR #3660, opened 2023-05-09
Static reparent for pinned elements

bkrmendy:
## Problem
When pasting/reparenting in the navigator, we want to preserve the sizes of the reparented elements. This principle doesn't hold up in (at least) two places:
- If an element has both pins set on a cross-axis, without a size set in that cross-axis, it won't preserve its size along that cross axis (as we remove the pins when pasting into a flex container, and when pasting into a container without a layout system, the pins might size the element to a totally different size).
- If an element has a relative size set along a cross-axis, the pasted element will be sized by its new parent, which can lead to surprising changes in sizing too.

## Fix
The code for these fixes is added with a "mini plugin architecture" where each case is handled by a separate function (`stripPinsConvertToVisualSize` and `convertRelativeSizingToVisualSize`). These functions can bail out early (by returning the reason for the bailout in `left`), or return a set of commands (in `right`). The functions are run by an orchestrator function (`runReparentPropertyEdgeCases`), which concatenates the command arrays.

### Limitations
`runReparentPropertyEdgeCases` doesn't deal with situations when the functions passed to it issue conflicting commands. Since this problem doesn't arise in the concrete fix, solving it is deferred until later.

### DIFF
--- a/.github/workflows/master-pushes.yml
+++ b/.github/workflows/master-pushes.yml
@@ -154,8 +154,8 @@ jobs:
       [
         test-editor-code,
         test-editor-jest,
-        test-editor-karma-shard-1,
-        test-editor-karma-shard-2,
+        call-test-editor-karma-shard-1,
+        call-test-editor-karma-shard-2,
         test-server,
       ]
     runs-on: ubuntu-latest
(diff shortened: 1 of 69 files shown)


balazsbajorics (approved):
I have a nitpick that these shouldn't be called edgeCases, but otherwise looks good!

enidemi (approved):

PR #3661, opened 2023-05-10
disallow dropping next to root element of component

bkrmendy:
## Problem
Navigator entries can be reordered next to the root element of a component. Inserting an element there would be something like inserting an element next to the root component of a conditional branch, since it'd have to be wrapped into a fragment along with the original root.

## Solution
Instead of fixing it properly, we simply disallow this for now (and will fix it along with other component editing related problems).


editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -147,7 +140,7 @@ function onDrop(
 ): void {
   const dragSelections = propsOfDraggedItem.getCurrentlySelectedEntries()
   const filteredSelections = dragSelections.filter((selection) =>
-    notDescendant(propsOfDropTargetItem, selection.elementPath),
+    notDescendant(propsOfDropTargetItem.elementPath, selection.elementPath),
```
ruggi, line 143: this is used extensively now, it would make sense to extract it to a single function and reuse it

ruggi: e.g. line 115, 193, 292

ruggi (approved):

enidemi (approved):

PR #3666, opened 2023-05-11
Fix/navigator select after reparent

enidemi:
**Problem:**
There is a regression in the editor that the navigator reparent and paste actions are not updating the selected elements.

**Commit Details:**
- fix `NAVIGATOR_REORDER` and `PASTE_JSX_ELEMENTS`
- extended test to check selection


bkrmendy (approved):
💯

PR #3667, opened 2023-05-11
Text editing of string literals and expressions in conditional clauses

gbalint:
**Problem:**
When both branches of the conditional has a text value, we just allow editing the whole conditional a single text expression.
However, when one branch of a conditional is a text and the other is an element, there is no way to edit the text today.

**Scope:**
This PR makes it possible to edit the text expression inside a conditional.
 Disclaimer: since expressions do not have metadata and bounds, there is no way to select the expression on the canvas. This PR does not change that, so there is also no way to edit the text by double clicking on the canvas!

The way to edit an expression is to select it in the navigator and press Enter. Everything else can come in later PRs.

**String literals vs js expressions**
There was an interesting challenge to decide on how to separate between `{true ? 'hello' : <div/>}` (where `'hello'` is a string literal) and the `{true ? hello : <div/>}` (where `hello` is a variable name) cases when editing the branch content in the text editor.
Our earlier approach in text editing (to show the curly braces around code) doesn't work that well here, because there are no curly braces when you are already in a conditional expression.
My simple solution was to show curly braces anyway just to makes sure you know this is code. If the user deletes any of the curly braces, we just save the content as a string literal.

**New lines**
In normal elements we replace newlines with `<br />` tags, but that doesn't make sense in string literals or js expressions, so I just leave the newlines in place in these cases (even though they are not rendered).


bkrmendy (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -262,6 +262,35 @@ export function renderCoreElement(
...
+        const textEditorProps: TextEditorProps = {
+          elementPath: elementPath,
+          filePath: filePath,
+          text: textContent,
+          component: React.Fragment,
+          passthroughProps: {},
+          editingItselfOrChild: 'itself',
+        }
+
+        return buildSpyWrappedElement(
```
seanparsons, line 279: I assume this is being used to get an entry for the text into the metadata?

seanparsons (approved):

PR #3668, opened 2023-05-11
Reparent strategies: reparent into absolute

bkrmendy:
## Description
This PR adds the following improvements for pasting/navigator reparenting elements into absolute containers:
- The element is always converted to its visual size on paste (with `width` and `height`), flex props and `min...`/`max...` props are removed
- The element is always pasted near the center of the parent, but with tweaks similar to what Figma has (https://screenshot.click/11-36-bo6tf-nncpz.mp4)
- When element is pasted into a container that has children with z-index set, the pasted element will get the highest z-index present among the new parent's children.


editor/src/components/inspector/common/css-utils.ts
```diff
@@ -4427,6 +4429,7 @@ export const cssParsers: CSSParsers = {
   width: parseCSSLengthPercent,
   height: parseCSSLengthPercent,
   flexBasis: parseCSSLengthPercent,
+  zIndex: parseCSSLengthPercent,
```
enidemi, line 4432: I think this could be `parseCSSUnitless`

enidemi (approved):

editor/src/components/inspector/common/css-utils.ts
bkrmendy: great idea! updated in https://github.com/concrete-utopia/utopia/pull/3668/commits/67d6378b9252e628a029c27ccb943c62e458ad41

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-property-strategies.ts
```diff
@@ -187,6 +186,105 @@ export const convertSizingToVisualSizeWhenPastingFromFlexToFlex =
...
+
+export const setZIndexOnPastedElement =
+  (
+    elementToReparent: { oldPath: ElementPath; newPath: ElementPath },
+    targetParent: ElementPath,
+    metadata: ElementInstanceMetadataMap,
+  ): ReparentPropertyStrategy =>
+  () => {
+    const siblings = MetadataUtils.getChildrenUnordered(metadata, targetParent)
+    const asd = mapDropNulls((sibling) => {
```
Rheeseyb, line 263: `asd`?

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-property-strategies.ts
```diff
@@ -187,6 +186,105 @@ export const convertSizingToVisualSizeWhenPastingFromFlexToFlex =
...
+    if (maximumZIndexOfOverlappingElements == null) {
+      return left('No siblings have z-index applied')
+    }
+
+    return right([
+      setCssLengthProperty(
+        'always',
+        elementToReparent.newPath,
+        PP.create('style', 'zIndex'),
+        setExplicitCssValue(cssNumber(maximumZIndexOfOverlappingElements, null)),
```
Rheeseyb, line 282: Should this be +1 to ensure it's on top?

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -4500,6 +4503,7 @@ const cssPrinters: CSSPrinters = {
   height: printCSSNumberOrUndefinedAsAttributeValue('px'),
   flexBasis: printCSSNumberOrUndefinedAsAttributeValue('px'),
   gap: printCSSNumberAsAttributeValue('px'),
+  zIndex: printCSSNumberOrUndefinedAsAttributeValue('px'), // TODO
```
Rheeseyb, line 4506: This should definitely not have a unit!

editor/src/components/navigator/navigator.spec.browser2.tsx
```diff
@@ -1584,391 +1584,513 @@ describe('Navigator', () => {
...
+        expect(element.style.height).toEqual('543px')
+      })
     })
 
-    it('reparenting between flex containers hardcodes width/height', async () => {
-      const editor = await renderTestEditorWithCode(
+    describe('reparenting to absolute', () => {
+      const absoluteSnippet = (insides: string) =>
         makeTestProjectCodeWithSnippet(`
       <div
```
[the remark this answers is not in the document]
Rheeseyb: Oh my bad, I think this was just message up because I had "hide whitespace" on 😕

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-property-strategies.ts
bkrmendy: I thought about it and I don't this so because of the following:
- When reparenting via copy/paste, we place the pasted element on top, so even if the same z-index is applied, it'll appear on top due to the order of the siblings.
- When reparenting through the navigator, there's a very precise indication of where the element will end up, and even after it's very easy to adjust the position of the element (either via the navigator or the bring forward/send backward shortcuts).

I think this strikes a good 80/20 balance in terms of time spent vs polished behaviour

editor/src/components/inspector/common/css-utils.ts
bkrmendy: Thanks! fixed in [82a032a](https://github.com/concrete-utopia/utopia/pull/3668/commits/82a032a9dabe0b48fcaa35f17a2c65400d28ad5d)

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-property-strategies.ts
bkrmendy: 😬 Thanks for noticing it! Must have been left behind from debugging, I removed it in [6eda2ee](https://github.com/concrete-utopia/utopia/pull/3668/commits/6eda2eec9e05e8ee365393ac2448f2ad0e2b78e4)

PR #3670, opened 2023-05-12
Improve duplicate uid handling...Redux.

seanparsons:
This resurrects and improves on https://github.com/concrete-utopia/utopia/pull/3646

**Problem:**
Duplicate UIDs were still arising from certain actions, which was then causing the editor to end up effectively in a broken state, but in a way that wasn't necessarily obvious to the user.

**Cause:**
When performing actions like duplication, the attributes (which now have uids of their own internally) were not being updated and ended up clashing with the original, even though the uids of the elements had been updated.

Also, the handling of existing uids wasn't multi-element safe. The collation of existing uids was done once, but then multiple elements were inserted, where the generation of uids ended up clashing between those newly inserted elements.

NEW: It appears that `UPDATE_FROM_WORKER` isn't necessarily safe for multiple rapid changes to the model, as some changes being introduced back in would result in duplicate UIDs across the whole of the project contents. But individually in the parser printer worker were determined to not be resulting in duplicate UIDs.

**Fix:**
The five main parts to this PR are as follows:
- NEW: Preventing partially stale updates from the workers being applied to the model, as newly added elements in one file might clash with new/updated attributes/elements fed through the worker in another file.
- Changing how the duplicate UIDs are surfaced, so that instead of breaking the editor behind the scenes with the user mostly none the wiser, a big error toast is displayed indicating the user should restart the editor.
- Identifying duplicate UIDs is improved where previously attributes weren't being included and those are very likely to cause clashes because entities often have common values like `'absolute'`, which end up hashing to the same UID.
- Fixing of duplicates mostly centering around `fixUtopiaElement` is now multi-insert safe, as well as including the UIDs of attribute values in it's handling of the duplicate checks. Also as highlight bounds were recently trimmed to only include those things which can be readily selected currently, the full highlight bounds are now stored so that we can make use of the keys as a shortcut for checking what already exists.
- More handling of the mappings from when UIDs changed have been added, for example with `guaranteeUniqueUidsFromTopLevel`.

**Commit Details:**
- NEW: `UPDATE_FROM_WORKER` now skips over partially stale updates to the model.
- NEW: Disabled the exception throwing logic that caused some issues with tests.
- NEW: Fixed some cases in the `fixXXXXXUIDs` functions that wouldn't necessarily descend
  into the new element to fix the UIDs of those as well.
- NEW: In `fixJSXElementUIDs` correctly update the `uid` field of the `data-uid` attribute
  if one is present.
- `duplicate` simplified and as a result now more comprehensively handles the various types it may encounter.
- `PASTE_JSX_ELEMENTS` is now multi-insert safe for the generation of uids.
- `editorDispatchInner` when running in a browser displays a toast indicating an internal issue should duplicates occur, otherwise in tests it will throw an exception so that errors can be correctly attributed to the test that caused them.
- `fullHighlightBounds` was added to `ParseSuccess`, so that when checking for existing uids the keys from that can be used instead of having to walk all the elements and their attributes.
- `getAllUniqueUids` now returns the currently unique uids, those which are duplicates and a collection of both.
- `getAllUniqueUids` was switched to using a switch-case statement so that changes to the model will be flagged up as breaking changes.
- `getAllUniqueUids` now correctly includes attributes and their uids.
- `getAllUniqueUids` also includes the arbitrary block attached to a component as well as `ARBITRARY_JS_BLOCK` cases.
- `isJSExpression` added as a convenience because it's easier to find by name.
- `generateConsistentUID` now always falls back to generating UIDs from the letters a-z, instead of only when `possibleStartingValue` had a length greater than or equal to 3.
- `fixUtopiaExpression` implemented as a small wrapper around `fixUtopiaElement`.
- `fixUtopiaElement` now includes attributes in the fixing pass that it performs.
- `fixUtopiaElement` assigns a uid to the `data-uid` property entry that is also generated to be consistent and unique.
- `fixUtopiaElement` now also fixes the `condition` property of a conditional expression.
- `fixParseSuccessUIDs` also now updates the newly added `fullHighlightBounds` property.
- `updateUID` now always adds the `uidToUse` to `mutableAllNewUIDs`.
- Added some small utility functions for Sets.
- `getAllUniqueUids` now has a "path" that allows for debugging, to help identify where for example a duplicate UID comes from.
- `fixUtopiaElement` returns mappings for UIDs that have changed.
- Added `fromArrayIndex` and `identityOptic` optics.
- Extracted out `updateHighlightBounds` so that it can be used in multiple places.
- `fixUtopiaElement` ensures to add a mapping when the uid for the `data-uid` prop changes.
- `guaranteeUniqueUidsFromTopLevel` now returns mappings for changes that have been made.
- `parseCode` maintains the existing UIDs from the changes that happen when it invokes `guaranteeUniqueUidsFromTopLevel`.


editor/src/core/shared/set-utils.ts
```diff
@@ -29,3 +29,19 @@ export function intersection<T>(sets: Array<Set<T>>): Set<T> {
...
+  return result
+}
+
+export function getSingleValueOnly<T>(set: Set<T>): T {
+  if (set.size === 1) {
+    for (const value of set) {
+      return value
+    }
+  }
+  throw new Error(`Set had ${set.size} when it was expected to have just 1.`)
```
bkrmendy, line 46: Why is this an error and not `null`? As far as I can see, none of the call sites actually checks whether `set` is a singleton, and this way the inner workings of this function leak out in a really unpredictable way

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -742,6 +750,35 @@ function editorDispatchInner(
...
+    const uniqueIDsResult = getAllUniqueUids(result.unpatchedEditor.projectContents)
+    if (uniqueIDsResult.duplicateIDs.length > 0) {
+      const errorMessage = `Running ${actionNames} resulted in duplicate UIDs ${JSON.stringify(
+        uniqueIDsResult.duplicateIDs,
+      )}.`
+      //if (IS_TEST_ENVIRONMENT) {
+      // In tests blow out with an exception so that the error is correctly attributed.
+      //  throw new Error(errorMessage)
+      //} else {
+      // When running in the browser log the error and tell the user to restart the editor.
```
bkrmendy, line 765: I think we can delete this comment

editor/src/core/shared/set-utils.ts
seanparsons: I disagree, this is being used to ensure that the sets at the call sites _are_ singletons and then retrieve the single value within. If It were called `getFirst` I would be more inclined to agree.

editor/src/components/editor/store/dispatch.tsx
seanparsons: I was thinking that it would still be preferable to have this behaviour, albeit only if we can fix the browser tests to not fail silently. As it requires us to scan the big pile of logging when running the browser tests to identify any cases that now fail.

But I'm happy for the wisdom of the crowd to determine if this should just be obliterated until such a time as our tests do capture these exceptions.

Rheeseyb (approved):

PR #3671, opened 2023-05-15
Spike: Show UID mocking race conditions in tests

bkrmendy:
This PR changes `FOR_TESTS_setNextGeneratedUid(s)` so that an exception is thrown when `FOR_TESTS_setNextGeneratedUid(s)` is called with mocked UIDs still present (i.e., when `MOCK_NEXT_GENERATED_UIDS.current` is a nonempty array).

The hypothesis is that `FOR_TESTS_setNextGeneratedUid(s)` should only be called when `MOCK_NEXT_GENERATED_UIDS.current` is an empty array (either because uids haven't been mocked yet by anyone, or `FOR_TESTS_CLEAR_MOCK_NEXT_GENERATED_UIDS` has been called). It seems like that with sharded tests, this assumption doesn't hold true.


balazsbajorics (approved):

bkrmendy:
doesn't seem to be a problem anymore

PR #3680, opened 2023-05-15
Support for benchmarking editor code.

seanparsons:
**Problem:**
Sometimes it's hard to get concrete numbers for the performance of some code we've written especially for the internals of the editor which aren't directly tied to the rendering.

**Fix:**
Created a benchmarking "application" that can call into editor code and run it to see what kind of performance we get, which uses the `benny` library to do the heavy lifting and reporting. Now we can create alternative implementations of important functions and then compare and contrast their performance.

**Screenshot:**
image

**Commit Details:**
- Added `benny` dev dependency.
- Added `build-benchmark` and `benchmark` scripts to `editor/package.json`.
- Moved some functions and a const from `uid-utils.ts` to `element-path.ts` because they caused a circular reference.
- Created `fake-localforage.ts` so that it can be used to shim in inplace of the real `localforage` because that does not like being run by Node.
- Implemented `benchmarkBuildTree` to test this and because it will likely be changed imminently so we can use this as a comparison point.
- Added `benchmarks.benchmark.ts` as the root of all the benchmarking functions.
- Created new webpack config to launch the benchmarking.


editor/src/core/shared/element-path.ts
```diff
@@ -8,7 +8,6 @@ import {
 import { arrayEquals, longestCommonArray, identity, fastForEach } from './utils'
 import { replaceAll } from './string-utils'
 import { last, dropLastN, drop, splitAt, flattenArray, dropLast } from './array-utils'
-import { extractOriginalUidFromIndexedUid } from './uid-utils'
```
balazsbajorics, line 11: was this change needed to make it compile? or just an improvement related to your example test case?

seanparsons: It was a change needed to avoid a circular reference.

balazsbajorics: oh boy

ruggi (approved):

balazsbajorics (approved):
nice!!

editor/src/core/shared/element-path.ts
Rheeseyb: So this would be the circular reference that has bitten us a number of times in tests!

editor/src/utils/fake-localforage.ts
```diff
@@ -0,0 +1,19 @@
+class FakeLocalForage {
```
Rheeseyb, line 1: We do something similar to this via a jest mock, except that backs it with an object to actually store and retrieve the results. Maybe it's worth doing that to prevent confusing issues later down the line? <https://github.com/concrete-utopia/utopia/blob/2a66237225cf36e7f6cf259010e13d03316dacfa/editor/src/components/editor/persistence/persistence.spec.ts#L154-L167>

Rheeseyb (approved):
Very nice!

editor/src/utils/fake-localforage.ts
seanparsons: That code was the inspiration for mine, for benchmarking I think we don't want it to do anything because it could skew results in unexpected ways.

PR #3686, opened 2023-05-16
Testing: Try using the cache-install Github action to cache the nix store.

seanparsons:
**Problem:**
Our Github actions pull from the Hydra cache over and over, which takes a fair amount of time.

**Fix:**
Switch to a different Github action, which uses the Github caching to cache the contents of the nix store.

**Commit Details:**
- Pull requests config updated to use new Github action.


seanparsons:
Possibly worth revisiting in the future.

PR #3690, opened 2023-05-17
Limit Github concurrency on the server.

### LINKED ISSUE #3689: Failure when attempting to import a Github repository.
**Describe the bug**
Importing this repo resulted in an error related to loading the branch: [https://github.com/Onnson/langflow](https://github.com/Onnson/langflow)

Result:
Error in console: `[GitHub] operation "Loading branch" failed.`

**Expected behaviour**
Branch should load with no problems.

**Screenshots**
image

seanparsons:
Fixes #3689

**Problem:**
It was observed that when importing certain Github repositories that memory usage on the server would spike quite high triggering memory limits on our provider.

**Cause:**
As some Github calls take a surprising amount of time to respond, we had some quite aggressive concurrency which it turns out caused lots of requests to be made simultaneously consuming lots of memory.

**Fix:**
Now all of the calls to Github are limited server wide with a semaphore so that only a limited number of service calls are being made to Github at any one time. In tests this has resulted in 1/2 to 1/3rd of the memory being used while importing a project and shortly after that.

**Commit Details:**
- Added "-O2" optimisation option to GHC options because we really should just have that anyway.
- Added `_githubSemaphore` to the production and development server resources.
- Added `limitWithSemaphore` utility function.
- Added a `githubSemaphore` parameter to `callGithub` and threaded that through numerous other functions to reach `callGithub` and then wrapped the entire function with a call to `limitWithSemaphore`.
- Refactored some NPM code to use the new `limitWithSemaphore` function.
- Deleted some unused functions.

### DIFF
--- a/server/package.yaml
+++ b/server/package.yaml
@@ -14,6 +14,7 @@ ghc-options:
   - -Wno-unused-imports
   - -Wno-deprecations
   - -rtsopts
+  - -O2
 
 default-extensions:
   - Strict
--- a/server/src/Utopia/Web/Executors/Common.hs
+++ b/server/src/Utopia/Web/Executors/Common.hs
@@ -375,21 +375,21 @@ useAccessToken githubResources logger metrics pool userID action = do
         Nothing    -> throwE "User not authenticated with Github."
         Just token -> action token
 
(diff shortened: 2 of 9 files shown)


bkrmendy (approved):

Rheeseyb (commented):
:+1: nice catch!

PR #3693, opened 2023-05-17
paste image in the middle of the viewport

bkrmendy:
## Problem
When pasting images from the system clipboard, with no parent selected, the image ends up at a "random" position, potentially way off screen.

## Fix
Always position pasted images in the middle of the viewport.

### DIFF
--- a/editor/src/templates/editor-canvas.tsx
+++ b/editor/src/templates/editor-canvas.tsx
@@ -1594,7 +1594,7 @@ export class EditorCanvas extends React.Component<EditorCanvasProps> {
         // needs testing if it's any help for other platforms
       } else {
         const canvasWrapperRect = this.canvasWrapperRef?.getBoundingClientRect() ?? zeroCanvasRect
-        const offset = canvasPoint({
+        const canvasViewportCenter = canvasPoint({
           x:
             -editor.canvas.roundedCanvasOffset.x +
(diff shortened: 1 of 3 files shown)


enidemi (approved):

editor/src/templates/editor-canvas.tsx
```diff
@@ -1592,11 +1593,21 @@ export class EditorCanvas extends React.Component<EditorCanvasProps> {
         // on macOS it seems like alt prevents the 'paste' event from being ever fired, so this is dead code here
         // needs testing if it's any help for other platforms
       } else {
+        const canvasWrapperRect = this.canvasWrapperRef?.getBoundingClientRect() ?? zeroCanvasRect
+        const offset = canvasPoint({
```
ruggi, line 1597: for readability would it be better to call this `canvasViewportCenter[offset]`?

ruggi (approved):

gbalint (approved):

editor/src/templates/editor-canvas.tsx
bkrmendy: great idea! Renamed it to `canvasViewportCenter` in https://github.com/concrete-utopia/utopia/pull/3693/commits/a4e26a0d131a8f33f34cd74a7c170f77e090c244

Rheeseyb (commented):
👍

PR #3700, opened 2023-05-18
Nulls in conditional clauses should have metadata

gbalint:
**Problem:**
We would like to be able to reparent into an empty conditional slot on the canvas. 
The bounds shown should be the bounds of the parent of the conditional.

**Fix:**
This PR only implements the necessary changes in the metadata for the feature.
This is a minimal solution which does not affect any other expressions, just nulls inside conditional branches.

- I added nulls inside conditional clauses into validPaths
- I made sure a fake spy entry is created for expressions (it checks if the element path is in validPaths)
- I gathered globalFrame and localFrame from the parent of the conditional from these elements.

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -41,6 +41,7 @@ import {
   isJSXConditionalExpression,
   isNullJSXAttributeValue,
   isJSXArbitraryBlock,
+  isJSXAttributeValue,
 } from '../../core/shared/element-template'
 import {
   getAllUniqueUids,
@@ -127,6 +128,7 @@ import {
   NavigatorEntry,
   isSyntheticNavigatorEntry,
   insertElementAtPath,
+  withUnderlyingTarget,
 } from '../editor/store/editor-state'
 import * as Frame from '../frame'
 import { getImageSizeFromMetadata, MultipliersForImages, scaleImageDimensions } from '../images'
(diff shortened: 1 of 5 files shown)


balazsbajorics (approved):

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -3054,6 +3062,25 @@ export function getValidElementPathsFromElement(
       )
     })
     return paths
+  } else if (isJSXAttributeValue(element)) {
```
ruggi, line 3065: you could use `isNullJSXAttributeValue` here and drop the `isNull` variable

ruggi (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2274,6 +2275,50 @@ function fillMissingDataFromAncestors(mergedMetadata: ElementInstanceMetadataMap
...
+  const nullsInConditional = Object.keys(workingElements).filter((p) => {
+    const element = workingElements[p]
+    const isNull =
+      element?.element != null && isLeft(element.element) && element.element.value == 'null'
+    if (!isNull) {
+      return false
+    }
+    const parentElement = workingElements[EP.toString(EP.parentPath(EP.fromString(p)))]
+    return MetadataUtils.isConditionalFromMetadata(parentElement)
+  })
```
bkrmendy, line 2287: Down the line it would be nice to take these "passes" out into separate functions, so that each new addition is forced into its own context, rather than interleaved with the existing stuff

bkrmendy (approved):

Rheeseyb (approved):

PR #3701, opened 2023-05-19
fix(reparent) reparent changes element to flow too often, switch fallback to absolute

enidemi:
**Problem:**
Reparent in the 'Before I Go' project 'Moodboard' component converts element to flow.
Pasting an absolute element sometimes converts them flow.

**Fix:**
Update `flowParentAbsoluteOrStatic` to prefer `absolute` in more cases. When checking the sibling layouts only reparent as static if all siblings are `static`. Switch fallback to `absolute`.

**Commit Details:**
- update `flowParentAbsoluteOrStatic`


seanparsons (approved):
Makes me nervous that this hasn't affected any tests...

bkrmendy (approved):

PR #3708, opened 2023-05-22
Conditionals receive specialSizeMeasurements.display from parent

gbalint:
**Description:**
Conditionals should receive their specialSizeMeasurement.display property from their parent.
We can use this to decide which reparent strategy should we use when we reparent into an empty conditional slot.

**Note:**: This PR just fixes the metadata, and not the reparent feature, that can come in a separate PR

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -2267,6 +2267,7 @@ function fillSpyOnlyMetadata(
 function fillMissingDataFromAncestors(mergedMetadata: ElementInstanceMetadataMap) {
   let workingElements: ElementInstanceMetadataMap = { ...mergedMetadata }
 
+  // elements without globalContentBoxForChildren should receive globalContentBoxForChildren from their parent
   const elementsWithoutGlobalContentBox = Object.keys(workingElements).filter((p) => {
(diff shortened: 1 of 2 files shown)


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2291,6 +2292,30 @@ function fillMissingDataFromAncestors(mergedMetadata: ElementInstanceMetadataMap
...
+    if (parentSpecialSizeMeasurements != null) {
+      workingElements[pathStr] = {
+        ...elem,
+        specialSizeMeasurements: {
+          ...elem.specialSizeMeasurements,
+          display: parentSpecialSizeMeasurements.display,
+        },
+      }
+    }
+  })
```
bkrmendy, line 2316: I think the structure would be way more apparent here if each of these passes had their separate functions, which take `workingElements`, do their thing, and return an updated `workingElements`

bkrmendy (approved):

editor/src/core/model/element-metadata-utils.ts
gbalint: I absolutely agree, I just did not want to give up on the mutation of `workingElements` for performance reasons, maybe it is an overoptimization though.

PR #3714, opened 2023-05-23
Editor Contract Dropdown

balazsbajorics:
the EditorContractDropdown
_The `EditorContractDropdown`_

This is a first of a string of PRs that eleveate the Frame / Fragment / Group dropdown into a major indicator telling us what Contract the utopia editor (canvas + inspector) is applying to the selected elements.

More detail can be found in the `Groups and Fragments and Frames` paper doc.

**Goals:**
* The indicator tells us how the currently selected is going to behave when you try to edit it. 
  * Frame is the "default" mode for elements that can be pinned (when they are `position: absolute`) and resized/reordered when they are participating in the document flow (which I sometimes will call "static", short for `position: static / position: relative (sic)`).
  * Fragment means the element is either a Fragment, or a fragment-like elefant (currently the only one we support is the conditional expression, coming soon are maps, React Higher Order Components, div with `display: contents`)
  * Group means the element is a Group – this is **disabled for now**, because we currently don't have Group behavior (coming right up this week)
* The indicator can also be used for conversion between the types. **This PR disables a lot of the conversion behavior**, and only allows the conversion to do something if it leads to a predictable outcome.
  * The conversion behavior is secondary to the dropdown as an indicator. Conversions should only be allowed if we are certain about their outcome. A bad conversion is worse than doing nothing.
* The logic that makes the decision if something is a Frame or Group or Fragment should be a central helper function shared by the inspector, canvas strategies, and navigator.

**As a first step,** the EditorContractDropdown always shows Frame unless we encounter something that can be considered Fragment -like. 

**Details:**
- Rename GroupDropdown to EditorContractDropdown
- Introduce the type `EditorContract =  'fragment' | 'frame'` (group coming later)
- Refactor the EditorContractDropdown to use EditorContract
- Temporarily remove references to the group option from EditorContractDropdown
- Remove tests for the group option from the dropdown. I've only kept the tests converting Fragment -> Frame and Frame -> Fragment
- The Frame -> Fragment and Fragment -> Frame conversions were broken / not working intuitively if the selected element has children which is not position: absolute. I changed the conversion logic so it bails out if not every children is position: absolute. 
- The EditorContractDropdown was disabled grey a lot of the time, I made it so it's always black, even if no conversion can be applied.


gbalint:
> The Frame -> Fragment and Fragment -> Frame conversions were broken / not working intuitively if the selected element has children which is not position: absolute. I changed the conversion logic so it bails out if not every children is position: absolute.

I think it would make sense to write a test for this case.

gbalint (approved):

bkrmendy (approved):

ruggi (approved):

PR #3715, opened 2023-05-23
Use the real element of expressions in metadata

gbalint:
**Problem:**
When we add a fake spy entry for elements, we add a `left(string)` element to the metadata, even though the `ElementInstanceMetadata.element` has the type `JSXElementChild`, which allows `JSExpression`s.

This means we can not access the original element object from projectContents from the metadata this way. It also breaks our convention that the left side of the Either should be an error case. And it makes it impossible to differentiate between a `null` value and `"null"` string.

This PR is a preliminary technical task which will help me implement text editing for expressions in conditionals. (And in my opinion it is a valid technical task anyway).

**Fix:**
Let's just insert the real element to the metadata for JSExpressions too. I had to make sure we don't rely on expressions being left-s, which is not trivial, because the compiler won't complain about it.
A bunch of tests were broken, that helped my identify some cases which were needed to be modified.


editor/src/core/shared/element-template.ts
```diff
@@ -2067,7 +2067,7 @@ export const emptySpecialSizeMeasurements = specialSizeMeasurements(
 )
 
 export const emptyComputedStyle: ComputedStyle = {}
-export const emptyAttributeMetadatada: StyleAttributeMetadata = {}
```
balazsbajorics, line 2070: this is a fun typo

enidemi (approved):

PR #3716, opened 2023-05-23
revamp navigator drop line calculation

bkrmendy:
# Try it [here](https://utopia.pizza/p/d6ddab55-before-i-go-basic-(forked)/?branch_name=fix-navigator-drop-line-indices)

## Problem
Drop lines are really unintuitive and unreliable in the navigator. In a related way, when reparenting up the hierarchy by dragging the mouse to the left, the dragged element can be moved to a location in the navigator that is far away from where the drop line was originally.

## Fix
Instead of applying point fixes for the problems above, the drop line mechanism was rebuilt on the following principle: with the exception of reparenting into an element via a slot or the parent outline, we **always** reparent into a parent, to a given index. This structure is now encoded in the types related to the drop line, the drop handler, and the navigator reorder action.

Another problematic aspect of the drop lines was that they only accepted mouse events when an item was dragged (so that otherwise, they can be clicked through to select navigator entries). This proved to be flaky because the hover handler of elements that cannot be reparented into (such as images) messed with the way we determined if a drag session was in progress. This was fixed by using an atom to store whether a drag is underway and using the react-dnd API to update the value of this atom.

Regarding moving the mouse to the left, we learned that this can only be done around navigator entries that can provide a valid index in the parent, grandparent, etc (valid drop line means that if the parent was collapsed, the drop line would still remain visible). Extra logic is added for this `targetEntryWithReparentWiggle`. Concretely, dragging the mouse to the left to reparent somewhere up the hierarchy is only available at the last sibling in the children array, and you can only reparent up to the highest common ancestor of the entry you're dragging at, and the navigator entry below that one. To make it possible to reparent elements up the hierarchy in place, this PR allows the drop lines to be shown right next to the dragged element. To avoid the drop line and the selected item blurring together, the selected element color is updated to a lighter shade.


editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -354,14 +379,19 @@ function isHintDisallowed(elementPath: ElementPath | null, metadata: ElementInst
 export const NavigatorItemContainer = React.memo((props: NavigatorItemDragAndDropWrapperProps) => {
   const editorStateRef = useRefEditorState((store) => store.editor)
 
+  const [isDragSessionInProgress, updateDragSessionInProgress] = useAtom(DragSessionInProgressAtom)
```
enidemi, line 382: have you tried using `collected.isDragging` from the useDrag hook? https://react-dnd.github.io/react-dnd/docs/api/use-drag#usedrag

ruggi (approved):

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -193,13 +212,9 @@ function onHoverDropTargetLine(
...
         notDescendant(propsOfDropTargetItem.elementPath, selection.elementPath),
       ) ||
-    EP.pathsEqual(propsOfDraggedItem.elementPath, propsOfDropTargetItem.elementPath) ||
     isHintDisallowed(propsOfDropTargetItem.elementPath, metadata)
   ) {
-    return propsOfDraggedItem.editorDispatch(
-      [showNavigatorDropTargetHint(null, null, null)],
-      'leftpane',
-    )
+    return propsOfDraggedItem.editorDispatch([hideNavigatorDropTargetHint()], 'leftpane')
```
Rheeseyb, line 217: Much better!

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -209,47 +224,61 @@ function onHoverDropTargetLine(
...
-    (() => {
-      if (cursorDelta.x >= -BasePaddingUnit) {
-        return null
-      }
+  const targetEntryWithReparentWiggle: {
+    type: DropTargetType
+    targetParent: NavigatorEntry
+    indexPosition: IndexPosition
+  } | null = (() => {
+    const WiggleUnit = BasePaddingUnit * 1.5
```
Rheeseyb, line 235: Very minor - we tend to only use PascalCase when a constant is global (or at least at the file level)

bkrmendy: ah OK, I'll move this constant to the file level

bkrmendy: https://github.com/concrete-utopia/utopia/pull/3716/commits/9943626c7ab22b69b23cf1d2a89e521b04ff203f

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -386,18 +416,38 @@ export const NavigatorItemContainer = React.memo((props: NavigatorItemDragAndDro
...
+    if (firstSibling == null) {
+      return false
+    }
+
+    return EP.pathsEqual(firstSibling.elementPath, props.elementPath)
+  }, [metadata, elementPathTree, props.elementPath])
+
+  const isLastSibling = React.useMemo(() => {
+    const siblings = MetadataUtils.getSiblingsOrdered(metadata, elementPathTree, props.elementPath)
+    const firstSibling = siblings.at(-1)
```
Rheeseyb, line 444: `lastSibling` surely

bkrmendy: whoops https://github.com/concrete-utopia/utopia/pull/3716/commits/84a1c3f3a621131d5eac6b3cff510b819394a6cf

Rheeseyb (approved):

PR #3718, opened 2023-05-24
fix(navigator) droptarget area

enidemi:
**Problem:**
While dragging in the navigator droptarget line between sibling shows up slowly.

**Fix:**
Change droptarget areas for elements where reparent is not available: dropline mouse catcher cover 50% of the navigator height, where reparent is available the droptarget mouse catcher is sized ~30%.

**Commit Details:**
- needs tests


enidemi:
closing this as this PR doesn't solve the problem I noticed

PR #3722, opened 2023-05-24
Drop metadata for nulls, use conditionals as retargeting parent when target has viable conditional slots

### LINKED ISSUE #3726: Drop metadata for nulls, use conditionals as retargeting parent when target has viable conditional slots

ruggi:
Fixes #3726 

**Problem:**

The nulls metadata in #3700 causes more harm than good, and is not strictly required for reparenting into conditionals.

**Fix:**

This PR does two things:
1. reverts #3700, removing the fake metadata frame bounds for (conditional) null elements
2. tweaks `findValidTargetsUnderPoint` so that it can still target conditional elements, but the logic is entirely self-contained in the `possibleTargetParentsUnderPoint` calculations (if the target children contain a conditional with an active empty branch, use that conditional as the target)


bkrmendy (approved):

gbalint (approved):

balazsbajorics (commented):
👌

PR #3728, opened 2023-05-25
Support reparenting into nested conditionals

### LINKED ISSUE #3727: Support reparenting into nested conditionals

ruggi:
Fixes #3727 

This is a followup to #3722 that allows reparenting into nested conditionals.

Kapture 2023-05-25 at 12 17 32

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
@@ -34,8 +34,11 @@ import {
 } from '../group-like-helpers'
 import { ReparentStrategy, ReparentSubjects, ReparentTarget } from './reparent-strategy-helpers'
 import { drawTargetRectanglesForChildrenOfElement } from './reparent-strategy-sibling-position-helpers'
-import { ElementPathTreeRoot } from '../../../../../core/shared/element-path-tree'
-import { isConditionalWithEmptyActiveBranch } from '../../../../../core/model/conditionals'
+import { ElementPathTreeRoot, getSubTree } from '../../../../../core/shared/element-path-tree'
+import {
+  getActualReparentParentForConditional,
+  isConditionalWithEmptyActiveBranch,
+} from '../../../../../core/model/conditionals'
 import { getInsertionPathForReparentTarget } from './reparent-helpers'
 
 export type FindReparentStrategyResult = {
@@ -105,6 +108,22 @@ export function getReparentTargetUnified(
   return targetParentUnderPoint
 }
 
+function findConditionalWithEmptyBranch(
+  tree: ElementPathTreeRoot,
+  metadata: ElementInstanceMetadataMap,
+): ElementPath | null {
+  for (const child of Object.values(tree)) {
+    if (isConditionalWithEmptyActiveBranch(child.path, metadata)) {
+      return child.path
+    }
+    const sub = findConditionalWithEmptyBranch(child.children, metadata)
+    if (sub != null) {
+      return sub
+    }
+  }
+  return null
+}
+
 function findValidTargetsUnderPoint(
   reparentSubjects: ReparentSubjects,
   pointOnCanvas: CanvasPoint,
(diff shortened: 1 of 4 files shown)


editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
```diff
@@ -147,15 +166,20 @@ function findValidTargetsUnderPoint(
...
-      metadata,
-      elementPathTree,
-      target,
-    )
-      .map((c) => c.elementPath)
-      .find((path) => isConditionalWithEmptyActiveBranch(path, metadata))
-    if (conditionalChildWithEmptyBranch != null) {
-      return conditionalChildWithEmptyBranch
+    if (!EP.isStoryboardPath(target)) {
+      const subtree = getSubTree(elementPathTree, target)
```
balazsbajorics, line 170: what's the cost of getSubTree? because we're going to call this for every element under the mouse for every frame during every Strategy interaction

ruggi: at a very quick glance, it's an O(n^2), maybe it would be nice to cache it, traversing the tree is gonna be needed one way or the other 🧐

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
```diff
@@ -105,6 +108,22 @@ export function getReparentTargetUnified(
...
 
+function findConditionalWithEmptyBranch(
+  tree: ElementPathTreeRoot,
+  metadata: ElementInstanceMetadataMap,
+): ElementPath | null {
+  for (const child of Object.values(tree)) {
+    if (isConditionalWithEmptyActiveBranch(child.path, metadata)) {
+      return child.path
+    }
+    const nestedConditional = findConditionalWithEmptyBranch(child.children, metadata)
```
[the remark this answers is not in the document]
ruggi: this wouldn't work if you have something like
```
div
  cond
    div
      cond
```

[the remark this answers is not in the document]
balazsbajorics: sorry I missed this comment. as I wrote in chat, in this case the inner div should be under the mouse, meaning it will be walked independently because it's listed in elementsUnderPoint

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
gbalint: I don't think you need to get the subtree here, you can just traverse conditionals until the active branch is either another conditional or empty. And then it will be linear compared to the number of directly embedded conditionals

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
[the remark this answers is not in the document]
gbalint: I don't think the 

div
  cond
    div
      cond

should be supported. That is going to work anyway for the inner div.

editor/src/components/canvas/controls/parent-outlines.tsx
```diff
@@ -82,7 +86,9 @@ export const ParentOutlines = controlForStrategyMemoized(
         }
         const isSlotTarget =
           findMaybeConditionalExpression(targetParent, store.editor.jsxMetadata) != null
-        const target = isSlotTarget ? EP.parentPath(targetParent) : targetParent
+        const target = isSlotTarget
+          ? getActualReparentParentForConditional(targetParent, store.editor.jsxMetadata)
```
balazsbajorics, line 90: I didn't catch this before! what is this isSlotTarget behavior here? is this walking up to find the first non-conditional parent? so it can get the size of it for the outline? in that case I think the function's name is misleading because the _actual_ reparent target _is_ the conditional expression

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
```diff
@@ -105,6 +105,31 @@ export function getReparentTargetUnified(
   return targetParentUnderPoint
 }
 
+function findConditionalWithEmptyBranch(
```
gbalint, line 108: Minor, but maybe the function name should reflect that this is a recursive search.

editor/src/components/canvas/controls/parent-outlines.tsx
gbalint: I would just call it `findFirstNonConditionalAncestor`

ruggi: ah, good point :D

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts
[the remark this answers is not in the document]
ruggi: yeah you're both right, my mistake

gbalint (approved):

balazsbajorics (approved):

PR #3734, opened 2023-05-26
Set conditional override on navigator reparent

### LINKED ISSUE #3733: Set conditional override on navigator reparent

ruggi:
Fixes #3733 

**Problem:**

When reparenting into an inactive branch via the navigator, the override should be set to that branch, similarly to what happens by clicking the conditional labels.

**Fix:**

If the target branch does not match the active clause, set the override.

Kapture 2023-05-26 at 10 50 10


bkrmendy (approved):
🧑‍🍳👌

balazsbajorics (approved):

PR #3737, opened 2023-05-26
Spike: paste sessions

bkrmendy:
# Try it [here](https://utopia.pizza/p/eb0e98b9-wise-zinnia/?branch_name=feature-replace-props-with-runtime-values)

## Basic idea
When pasting elements/components with some property values coming from the surrounding code, we should find a way to allow toggling between pasting them with these properties replaced (might screw up existing data flow), and pasting them with the unresolved references (might crash the canvas).

## Spike goals
- Create a POC implementation of paste sessions on top of the interaction session framework
- Check what falls out of this approach
- Check what would have to be modified in the interaction session framework to make this approach work, and how big those changes would be

### Questions
- When does a paste session end? On selection change? When the canvas loses focus? When the navigator loses focus? When a non-paste action is dispatched?

## Shortcuts taken
(incomplete list)
- [ ] multiple pasted elements are not wrapped in a fragment
- [ ] repeated paste is broken
- [ ] the property replacing function doesn't recurse into children


bkrmendy:
implemented in a more full-featured way on https://github.com/concrete-utopia/utopia/pull/3798

PR #3738, opened 2023-05-26
Fix remaining duplicate UIDs issues in the parser.

seanparsons:
**Problem:**
Duplicate UID errors are still originating from the `UPDATE_FROM_WORKER` action.

Also, while the error would raise the UIDs that were causing problems it's difficult to then identify _where_ those originate from without slowly trawling the entire model.

**Cause:**
There were a few smaller cases left that triggered this:
- Not including the highlight bounds of top level arbitrary blocks.
- When updating the `data-uid` property, the uid of the old expression would be left behind in the highlight bounds.
- For the non-full highlight bounds, entries other than elements, fragments and conditionals would be included when they shouldn't.

**Fix:**
The logic of `getAllUniqueUids` now reports on the paths in the project contents to where the duplicates originate. This means that should the error arise anew from changes to the editor in the future the values causing the error should be much more easily identified.

Previously there was a test that tried to check that the uids that could be found in a successful parse result would be consistent with the highlight bounds that are also calculated. This has been extended to cover a series of axioms:
- Check that the UIDs for the elements are the same as the keys of the regular highlight bounds.
- Check that the UIDs for everything are the same as the keys of the full highlight bounds.
- Check that the parse does not produce any duplicates within itself.
This test is now also run with `stripUIDs` both set to true and to false in two test cases.

The arbitrary generators that are used to generate test cases now also generate uids that are fairly likely to cause a collision instead of allowing those values to be UUIDs in some cases, which makes the failure cases more likely to arise.

With these done it was mostly just a case of covering the cases identified (listed above) from running the tests.

**Notes:**
The highlight bounds entry for the expression holding the value of the `data-uid` attribute is given the bounds of the element, probably this should be tweaked to instead hold a value like `'non-existant'`, as it's effectively not a real value that exists in the code being parsed.

**Commit Details:**
- Have `getElementFromRenderResult` provide more information about what is wrong.
- Move `getAllUniqueUids` code into `get-unique-ids.ts`
- Break apart `getAllUniqueUids` into separate functions, so that multiple `getAllUniqueUids` functions could be implemented.
- Call `checkUID` in `extractUIDFromArbitraryBlock`, so that the uid of the block is included.
- `GetAllUniqueUIDsWorkingResult.duplicateIDs` now holds the paths to the values that hold the uid, so that any duplicates raised can be identified immediately from the editor state.
- In the parser, permit attributes to be null in a few places which makes sense for shorthand fragments as they cannot support them at all.
- In `forciblyUpdateDataUID` extract out multiple steps for building/modifying the highlight bounds.
- `forciblyUpdateDataUID` checks the change in the attributes and trims from the highlight bounds any uids which now don't exist anymore.
- In `forciblyUpdateDataUID` if the element isn't a fragment (which can't support props), include the UID in the highlight bounds.
- Refactored `createJSXElementOrFragmentAllocatingUID` to handle the lack of props for shorthand fragments, as longhand fragments may have `data-uid` in the code, but that isn't preserved as in properties.
- `produceElementOrFragmentFromTSElement` now assigns highlight bounds from the response to `createJSXElementOrFragmentAllocatingUID`, because it needs to be able to remove highlight bounds.
- `parseOutFunctionContents` ensures that top level arbitrary blocks have their highlight bounds included.
- Changed the UIDs alignment with highlight bounds test to be based on multiple axioms.
- Updated the arbitrary instances for elements and related types, to have collision triggering uids and some other better representations.
- Extracted out `walkWantedElementsOnly`.
- For `walkJSXElementChild` only call `includeElement` for elements, fragments and conditionals.


bkrmendy (approved):

bkrmendy (approved):

editor/src/core/workers/parser-printer/parser-printer.spec.ts
```diff
@@ -4822,61 +4827,109 @@ export var whatever2 = (props) => <View data-uid='aaa'>
...
       )
+      FastCheck.assert(dataUIDProperty, { verbose: false, numRuns: 100 })
     }
-    const printableArbitrary = printableProjectContentArbitrary()
-    const dataUIDProperty = FastCheck.property(
-      printableArbitrary,
-      checkElementUIDSMatchHighlightBounds,
-    )
-    FastCheck.assert(dataUIDProperty, { verbose: true })
+    /* eslint-disable-next-line jest/expect-expect */
```
Rheeseyb, line 4925: Rather than do this we can provide the names of the functions that will include the `expect` like so https://github.com/concrete-utopia/utopia/blob/f6ab2ccf0ea6bc78e4c82988f8f86be95d80b170/editor/src/components/canvas/canvas-context-menu.spec.browser2.tsx#L77

PR #3740, opened 2023-05-26
Allow turning on conditional override on default branch via the navigator

### LINKED ISSUE #3739: Allow setting the override for the default branch via the navigator

ruggi:
Fixes #3739 

**Problem:**

Clicking the `TRUE` label in the navigator does not allow setting the override (it can only be done via the inspector), which feels in contradiction with the false branch.

**Fix:**

Toggle the conditional overrides via the label in a consistent way.
Note: this affects the labels _only_: clicking on the branch contents still keeps the old behavior.

Kapture 2023-05-26 at 15 51 29

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item.tsx
@@ -108,6 +108,22 @@ export function getSelectionActions(
   }
 }
 
+function selectItem(
+  dispatch: EditorDispatch,
+  getSelectedViewsInRange: (i: number) => Array<ElementPath>,
+  navigatorEntry: NavigatorEntry,
+  index: number,
+  selected: boolean,
+  event: React.MouseEvent<HTMLDivElement>,
+) {
+  const elementPath = navigatorEntry.elementPath
+
+  dispatch(
+    getSelectionActions(getSelectedViewsInRange, index, elementPath, selected, event),
+    'leftpane',
+  )
+}
+
 const highlightItem = (
   dispatch: EditorDispatch,
   elementPath: ElementPath,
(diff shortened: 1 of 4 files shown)


balazsbajorics (approved):

Rheeseyb (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -140,7 +140,7 @@ function selectItem(
   index: number,
   selected: boolean,
   event: React.MouseEvent<HTMLDivElement>,
-  conditionalOverrideUpdate: ConditionalCase | 'clear-override' | 'no-update',
+  conditionalOverrideUpdate: ConditionalOverrideUpdate,
```
bkrmendy, line 143: this functionality should be moved to a dedicated click handler in `ConditionalClauseNavigatorItemContainer`

bkrmendy (changes requested):

editor/src/components/navigator/navigator-item/navigator-item.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/3740/commits/b548e36b8595134b86e87f91789fa16e557c7b39

ruggi: Note @bkrmendy : I reverted the refactor you requested (https://github.com/concrete-utopia/utopia/pull/3740/commits/930dbee776c53dae7893b6759cba93c07ed73435) because it would still miss the handling of clicking the individual branch items. I still think you're right and the logic should live in the `ConditionalsClauseNavigatorItemContainer` but it's a bigger refactor and I think it should live in its own PR.

bkrmendy: Sounds good

PR #3741, opened 2023-05-26
Allow editing text of the active branch when the conditional is selected

gbalint:
**Description:**
This PR is a step towards making double-click to text edit work for expressions in conditionals.
The goal of this specific PR is to make text edit work when the conditional itself is selected. Why is this useful? This will make it possible to edit the text content without adding metadata for js expressions (the conditional already have metadata, and that will make it easier to activate editing on double click).

The result of the PR is already usable, you can select the conditional in the navigator, and press enter to edit the text content of the active branch (if the active branch is text-only)

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
@@ -263,8 +263,7 @@ export function renderCoreElement(
       const elementIsTextEdited = elementPath != null && EP.pathsEqual(elementPath, editedText)
 
       if (elementIsTextEdited) {
-        const text = trimAndJoinTextFromJSXElements([element])
-        const textContent = unescapeHTML(text ?? '')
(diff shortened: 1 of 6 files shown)


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -401,6 +401,36 @@ export function renderCoreElement(
...
+          metadataContext,
+          updateInvalidatedPaths,
+          [],
+          TextEditorWrapper,
+          inScope,
+          jsxFactoryFunctionName,
+          shouldIncludeCanvasRootInTheSpy,
+          imports,
+          filePath,
+        )
```
balazsbajorics, line 431: is this a copy-paste of the other `if (elementIsTextEdited)` block? if so, could we turn this into a shared function? I worry that we'll fix an unescape bug in one path but not the other

balazsbajorics (approved):
looks good, except the (possibly) duplicated code in ui-jsx-canvas-element-renderer-utils

seanparsons (approved):

PR #3747, opened 2023-05-30
Absolute Group Component

balazsbajorics:
**From the paper doc:**

1. Absolute Groups

`<Group style={{position: 'absolute'}}>`
This can also be achieved without using the Group component, with pure divs
This concept is a bit of a breakaway from Figma, and more similar to Sketch’s Groups. Basically groups in Figma that are put into Absolute Frames are their “Absolute Groups”. You can make a Group inside an Autolayout absolute by clicking on the absolute button. It “frees” the group from participating in the autolayout context, similarly to how it would work with Frames. 
But for Figma, “Absolute” Groups work 100% like Fragments: they don’t provide offset bounds to their children, and the groups themselves cannot be pinned.
However, in Utopia, we have the opportunity to make the Groups pinnable! Malte and I believe it’s worth trying to match Sketch instead of Figma, for the following reasons:

- We already have Fragments, which 100% match Figma’s “Absolute” Groups
- Sketch-like “Absolute” Groups would match Static Groups (`flex: 0 0` variant), unlike for Figma, where Absolute Groups are very very different from Static Groups!

 
Here’s some the pseudo code for better understanding:
Usage in Utopia:
Let’s start with the simplest example

    <Group style={{ position: 'absolute', left: 0, top: 0 }}>
      <div style={{ position: 'absolute', left: 0, top: 50, width: 50, height: 50 />
      <div style={{ position: 'absolute', left: 50, top: 0, width: 50, height: 50 />
    </Group>

!!!Note how the Group does not require you to give it a width and height!! 
So far, so Fragment! But what if we try to give the Group pins? 

    <Group style={{ position: 'absolute', left: 50, top: 50 }}>
      <div style={{ position: 'absolute', left: 0, top: 50, width: 50, height: 50 />
      <div style={{ position: 'absolute', left: 50, top: 0, width: 50, height: 50 />
    </Group>

Now we’re cooking with gasoline!

But wait, it gets crazier!!

    <Group style={{ position: 'absolute', bottom: 50, right: 50 }}>
      <div style={{ position: 'absolute', left: 0, top: 50, width: 50, height: 50 />
      <div style={{ position: 'absolute', left: 50, top: 0, width: 50, height: 50 />
    </Group>

You can of course pin it on the bottom right too! Still no explicit size given.

Okay, so far all the children have been pinned with left/top/width/height. What about other types of pins?

    <Group style={{ position: 'absolute' }}>
      <div style={{ position: 'absolute', left: 0, top: 50, width: 50, height: 50 />
      <div style={{ position: 'absolute', left: 50, top: 0, width: 50, height: 50 />
      <div style={{ position: 'absolute', top: 0, right: 0, width: 25, height: 25 />
    </Group>

That top right rectangle is top/right/width/height pinned! 

caveat: For elements that are pinned in a certain combination, it’s unavoidable to give the Group an actual size. For example this group couldn’t look like this without an explicit size set on it:

    <Group style={{ position: 'absolute', width: 150, height: 100 }}>
      <div style={{ position: 'absolute', left: 0, top: 50, width: 50, height: 50 />
      <div style={{ position: 'absolute', left: 50, top: 0, width: 50, height: 50 />
      <div style={{ position: 'absolute', top: 0, right: 0, width: 25, height: 25 />
    </Group>

This brings us to Groups that have explicit size!
Actually, it’s a user convenience that they don’t need to define an explicit size on the group element. But if during canvas interactions they get to a configuration that requires the group to have an explicit size, that’s no problem: the editor detects it and takes care of it, by automatically adding width/height to the Group. This width and height is then properly maintained during Canvas interactions!

Implementation details:
The Group component uses some lightweight runtime javascript (in the form of a MutationObserver) to set its own width and height to match the width and height of the AABB of its children. The group does not try to shift the position of the children, so its the children’s own duty to not have an improper offset that makes them stick out of their own canvas frame. 
Simply put, it’s the editor’s job to keep truing up the children offset during editing, but it’s the Group component’s job to keep updating its own width and height to match the children AABB. 

The children of the group must be position: absolute elements. This means that it’s the editor’s job to prevent the user from making an element position: static inside a group. For example, pressing X to make static needs to be either disabled, or it needs to behave differently.

**Implementation Details:**
- a Placeholder empty group component has been added to utopia-api
- For the sake of easy iteration / quick turnaround times, I've put a group-component.tsx in the editor codebase and I replace the utopia-api placeholder Group with the editor's Group component. once the group component is stable, we can move it to utopia-api
- I've co-opted the `PushIntendedBounds` command, and renamed it to `PushIntendedBoundsAndUpdateGroups`
- Render tests have been added for various configurations
- Interaction tests have been added for moving the child elements of a Group
- Reparenting into groups on the canvas is disabled similar to reparenting into Fragments. This is intentional and good
- Fun fact: utopia groups are permitted to have background color!

**Not in the scope of this PR:**
- The only way to add a Group to a project is manually in code
- CMD + G to group is still wrapping in Fragment
- Tests for resizing child elements coming in follow-up PR (the functionality works but doesn't have test coverage)
- Tests for keyboard interactions missing
- Resizing the selected Group element is temporarily disabled, coming in follow up PR
- Groups which are _not_ position: absolute are coming in a follow-up PR
- The Inspector's Frame / Fragment / Group dropdown doesn't handle groups yet


editor/package.json
```diff
@@ -28,7 +28,7 @@
     "test-update-snapshot": "pnpm run test -- --updateSnapshot",
     "test-browser-base": "NODE_OPTIONS=$NODE_OPENSSL_OPTION REACT_APP_AUTH0_CLIENT_ID='test' REACT_APP_AUTH0_ENDPOINT='test' REACT_APP_AUTH0_REDIRECT_URI='test' REACT_APP_ENVIRONMENT_CONFIG=development karma start",
     "test-browser": "pnpm run test-browser-base karma-no-shard.conf.js --single-run",
-    "test-browser-debug": "pnpm run test-browser-base karma-no-shard.conf.js --debug",
+    "test-browser-debug": "REACT_APP_BROWSER_TEST_DEBUG='true' pnpm run test-browser-base karma-no-shard.conf.js --debug",
```
balazsbajorics, line 31: this is the only unrelated change, it restores the feature that test-browser-debug doesn't mess with the URL so you can reload the browser to re-run the tests

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
```diff
@@ -71,28 +71,7 @@ import { FOR_TESTS_setNextGeneratedUids } from '../../../../core/model/element-t
 import { isRight } from '../../../../core/shared/either'
 import { ImmediateParentOutlinesTestId } from '../../controls/parent-outlines'
 import { ImmediateParentBoundsTestId } from '../../controls/parent-bounds'
-
-async function resizeElement(
```
balazsbajorics, line 75: this has been moved to absolute-resize.test-utils.tsx

bkrmendy (approved):

editor/src/core/es-modules/package-manager/group-component.tsx
```diff
@@ -0,0 +1,181 @@
...
+  let maxNegativeBottom = 0
+
+  children.forEach((child) => {
+    if (!(child instanceof HTMLElement)) {
+      return
+    }
+
+    const sanitizedPins = getAllPinsSanitized(child)
+
+    const width = sanitizedPins.left + sanitizedPins.width + sanitizedPins.right
```
seanparsons, line 97: I think this would be clearer if it was renamed to `groupWidth`, because I can't help but think it relates to this particular child.

editor/src/core/shared/array-utils.ts
```diff
@@ -433,3 +433,7 @@ export function strictEvery<T>(
 ): boolean {
   return ts.length > 0 && ts.every(predicate)
 }
+
+export function isNotNull<T>(value: T | null | undefined): value is T {
```
seanparsons, line 437: Why is this in `array-utils.ts`?

seanparsons (approved):

editor/src/core/shared/array-utils.ts
balazsbajorics: This is the only use case of isNotNull that is important:

``` typescript
const exampleArray = [1, 2, null, 4, 5] // this is type Array<number | null>

const filteredArrayBad = exampleArray.filter(e => e != null) // this is type Array<number | null>

const filteredArrayGood = exampleArray.filter(isNotNull) // this is type Array<number> !! 
```

I couldn't really find a better util file for this, so I put it in the array utils

balazsbajorics: yeah that might work! :)

editor/src/components/canvas/canvas-strategies/strategies/groups.spec.browser2.tsx
```diff
@@ -0,0 +1,1367 @@
...
+
+async function dragByPixels(
+  editor: EditorRenderResult,
+  delta: WindowPoint,
+  testid: string,
+  modifiers: Modifiers = emptyModifiers,
+) {
+  const targetElement = editor.renderedDOM.getByTestId(testid)
+  const targetElementBounds = targetElement.getBoundingClientRect()
+  const targetElementCenter = windowPoint(getDomRectCenter(targetElementBounds))
```
Rheeseyb, line 52: We don't need to use `WindowPoint`s with these event helper functions. I think the cases where we do are just leftover from when we were hand rolling the code each time.

editor/src/components/canvas/canvas-strategies/strategies/groups.spec.browser2.tsx
```diff
@@ -0,0 +1,1367 @@
...
+      NO_OP()
+    },
+  })
+}
+
+function assertStylePropsSet(
+  editor: EditorRenderResult,
+  elementPathStr: string,
+  stylePropsObjToMatch: Record<string, any>,
+) {
```
Rheeseyb, line 67: This is such a nice way to check the result without printing or string interpolating the expected code. The only thing I'm trying to wrap my head around is if the `stylePropsObjToMatch` should be exhaustive (so that we can also ensure no rogue props have been added along the way)

editor/src/components/canvas/canvas-strategies/strategies/groups.spec.browser2.tsx
```diff
@@ -0,0 +1,1367 @@
...
+                bottom: 100,
+                width: 50,
+                height: 50,
+              }}
+            />
+          </Group>
+        `)
+        const groupDiv = editor.renderedDOM.getByTestId('group')
+
+        // notice that the child ends up with zero width and height because it was set to auto
```
Rheeseyb, line 400: This comment is more of a lie than the cake!

editor/src/components/canvas/canvas-strategies/strategies/groups.spec.browser2.tsx
```diff
@@ -0,0 +1,1367 @@
...
+    if (actualValue !== expectedValue) {
+      throw new Error(
+        `assertStylePropsSet equality
+
+props.style.${key} on element ${elementPathStr}:
+
+Expected: ${JSON.stringify(expectedValue)}
+Received: ${JSON.stringify(actualValue)}`,
+      )
+    }
```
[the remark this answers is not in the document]
balazsbajorics: then the printed error will be along the lines of

`expected 230 to be 140`

that's not super helpful in a test that calls this function 10 times! 

instead, this error message matches the style of the expect error message and gives you key information which helps you fix the bug!

editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
```diff
@@ -0,0 +1,336 @@
...
+  targets.forEach((frameAndTarget) => {
+    const parentPath = EP.parentPath(frameAndTarget.target)
+    const parentIsGroup = treatElementAsGroupLike(editor.jsxMetadata, parentPath)
+
+    if (!parentIsGroup || parentPath == null) {
+      // bail out
+      return
+    }
+
+    // I assume that affectedAncestors are ordered bottom-up
```
Rheeseyb, line 122: This doesn't sound like a safe assumption to make

editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
```diff
@@ -0,0 +1,336 @@
...
+        frame: newGlobalFrame,
+        target: parentPath,
+      }
+    }
+  })
+
+  // we REALLY need a lens for this!
+  let commandsToRun: Array<CanvasCommand> = []
+
+  // okay so now we have a bunch of new globalFrames. what do we do with them?
```
Rheeseyb, line 143: Yes, what _do_ we do with them?

editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
```diff
@@ -0,0 +1,336 @@
...
+  Object.keys(updatedGlobalFrames).forEach((pathStr) => {
```
Rheeseyb, line 144: [`Object.entries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) is your friend

editor/src/core/es-modules/package-manager/group-component.tsx
```diff
@@ -0,0 +1,181 @@
...
+  }, [])
+
+  return (
+    <div
+      {...props}
+      ref={groupRef}
+      style={{
+        ...props.style,
+        flex: '0 0 auto',
+        contain: 'layout',
```
Rheeseyb, line 175: What's the deal with these coming after the object spread?

editor/src/core/shared/array-utils.ts
Rheeseyb: Huh, I could have sworn we already had something like this for exactly this use case, but I can't find it...

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/groups.spec.browser2.tsx
balazsbajorics: pushed

editor/src/components/canvas/canvas-strategies/strategies/groups.spec.browser2.tsx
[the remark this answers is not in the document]
balazsbajorics: I forgot to write that the main pain point with the expect in the helper function is that the context you get from the error is the helper function, not the caller of the helper function

actually how does Jest / Mocha do this? they must be messing with the stack trace to cut out the library internals

editor/src/components/canvas/canvas-strategies/strategies/groups.spec.browser2.tsx
balazsbajorics: that's why I put the undefined checks in. so you can assert a pin is definitely not defined.

I would not make this exhaustive, because then we introduce the noise of background colors, position: absolute etc

PR #3749, opened 2023-05-30
Fix/paste allow pasting textlike elements to spans

enidemi:
**Problem:**
It's possible to paste a Rectangle into a span.

**Fix:**
Restrict pasting based on element type, text elements should accept only other text elements, with this PR it's still allowed to paste a `p` or `h1` into a span, but not a `div` or `Rectangle`.

This fix is only for paste, insertion and reparenting behaviours are not changed.

**Commit Details:**
- a helper named `canInsertElementsToTargetText`, but I'm open to suggestions about the naming
- update `getTargetParentForPaste`
- tests


bkrmendy (approved):

seanparsons (approved):
Not a reason to prevent this from being merged, but I wonder if this should pop a little toast indicating why the element didn't end up where the user wanted it to go.

PR #3750, opened 2023-05-30
Area select

### LINKED ISSUE #3758: Allow to area-select on the canvas

ruggi:
Fixes #3758 

**Problem:**

We currently don't support area-select on the canvas.

**Fix:**

This PR adds support for area selection on the canvas.
Caviats:
1. it only works with direct storyboard children, not nested elements. _Or..._
2. it also works with scene children, if their parent is the root of the scene. BUT! If the selection includes non-scene-children, only those will be included in the selection. _But..._
3. it also works with the scenes themselves, if they are fully contained in the selection area.

Kapture 2023-06-01 at 13 06 35


editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -156,6 +473,27 @@ export const CanvasWrapperComponent = React.memo(() => {
...
+        background: colorTheme.primary10.value,
+        position: 'absolute',
+        width: rectangle.width,
+        height: rectangle.height,
+        left: rectangle.x,
+        top: rectangle.y,
+      }}
+    />
+  )
+})
```
bkrmendy, line 495: if you set `.displayName` on this, it'll show up with that name further down in re-rendering screenshot

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -93,11 +119,296 @@ export const CanvasWrapperComponent = React.memo(() => {
...
+  const mode = useEditorState(
+    Substores.restOfEditor,
+    (store) => store.editor.mode,
+    'CanvasWrapperComponent mode',
+  )
+  const metadata = useEditorState(
+    Substores.metadata,
+    (store) => store.editor.jsxMetadata,
+    'CanvasWrapperComponent metadata',
+  )
```
bkrmendy, line 152: I'm not sure how big a problem this is, but this will cause a re-render on each metadata change, I'd check the impact with the re-rendering highlights on just to make sure

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -93,11 +119,296 @@ export const CanvasWrapperComponent = React.memo(() => {
...
+      x: selectionArea.x - refRect.x - actualNavigatorWidth,
+      y: selectionArea.y - refRect.y,
+      width: selectionArea.width,
+      height: selectionArea.height,
+    })
+  }, [selectionArea, actualNavigatorWidth])
+
+  function onMouseMove(e: React.MouseEvent) {
+    setMousePoint(canvasPoint({ x: e.clientX, y: e.clientY }))
+  }
```
bkrmendy, line 247: could this be wrapped with a `useCallback`?

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -93,11 +119,296 @@ export const CanvasWrapperComponent = React.memo(() => {
...
+  }, [mode.type])
+
+  React.useEffect(() => {
+    if (elementsUnderSelectionArea == null) {
+      return
+    }
+    dispatch([
+      setHighlightedViews(elementsUnderSelectionArea),
+      setHoveredViews(elementsUnderSelectionArea),
+    ])
```
bkrmendy, line 410: I'd be wary of dispatching actions in `useEffect` call, because I'm afraid this might be hard to debug later on. Can this be rephrased so that the action dispatch is tied to a more "discrete" action, such as a mouse move handler firing?

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -93,11 +119,296 @@ export const CanvasWrapperComponent = React.memo(() => {
...
+      dispatch,
+      selectionAreaStart,
+      highlightedViews,
+      clearAndGetActions,
+      isValidMouseEventForSelectionArea,
+    ],
+  )
+
+  React.useEffect(() => {
+    if (selectionArea == null || mousePoint == null || selectionAreaStart == null) {
```
bkrmendy, line 305: Can this be moved into a standalone function? This is a lot of logic, and probably it could be broken up even further. Also, if the logic here only run during an area select session, I think a lot of things can be precomputed here

editor/src/components/canvas/canvas-wrapper-component.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/3750/commits/ae5f541ea826bd44ded86ec29c14312c4519656c

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -93,11 +119,296 @@ export const CanvasWrapperComponent = React.memo(() => {
...
+  )
+  const canvasOffset = useEditorState(
+    Substores.canvasOffset,
+    (store) => store.editor.canvas.realCanvasOffset,
+    'CanvasWrapperComponent canvasOffset',
+  )
+  const highlightedViews = useEditorState(
+    Substores.highlightedHoveredViews,
+    (store) => store.editor.highlightedViews,
+    'CanvasWrapperComponent highlightedViews',
```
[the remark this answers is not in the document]
ruggi: This selector https://github.com/concrete-utopia/utopia/pull/3750/commits/db20b5ded4e3ae04f768a32dea15bf81081490f1 aims at fixing this, probably it would need to become a `cachedSelector` but that's on the list

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1660,26 +1660,15 @@ export const UPDATE_FNS = {
...
-    ) {
-      return editor
-    } else {
-      return {
-        ...editor,
-        highlightedViews: [action.target],
-      }
+    return {
+      ...editor,
+      highlightedViews: action.targets,
```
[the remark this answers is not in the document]
ruggi: https://github.com/concrete-utopia/utopia/pull/3750/commits/9a7fc44aa222d1901d15e8dcda2103f57330eddc

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -93,11 +119,296 @@ export const CanvasWrapperComponent = React.memo(() => {
...
+        res.push({
+          path: element.elementPath,
+          type: isChildOfSceneRoot
+            ? 'scene-child'
+            : MetadataUtils.isProbablyScene(metadata, element.elementPath)
+            ? 'scene'
+            : 'storyboard-child',
+          fullyCovered: rectangleContainsRectangle(rect, frame),
+        })
+      }
```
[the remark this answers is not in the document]
ruggi: yeah, see the new selectors 🙏

editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -1,43 +1,116 @@
...
 import { StrategyIndicator } from './controls/select-mode/strategy-indicator'
-import { CanvasToolbar } from '../editor/canvas-toolbar'
-import { useDispatch } from '../editor/store/dispatch-context'
+import { windowToCanvasCoordinates } from './dom-lookup'
+import { metadataSelector, selectedViewsSelector } from '../inspector/inpector-selectors'
+import { createSelector } from 'reselect'
+
+export const CanvasWrapperTestId = 'canvas-wrapper'
+
+const possibleElementsUnderMouseSelector = createSelector(
```
seanparsons, line 70: This looks like it could be potentially quite costly, have you tried it on the Before I Go project?

editor/src/core/shared/math-utils.ts
```diff
@@ -263,6 +263,24 @@ export function sizeFitsInTarget(sizeToCheck: Size, target: Size): boolean {
...
+  const yOverlap = a.y < b.y + b.height && a.y + a.height > b.y
+  return xOverlap && yOverlap
+}
+
+export function rectangleContainsRectangle(
+  outer: CanvasRectangle,
+  inner: CanvasRectangle,
+): boolean {
+  return (
+    rectContainsPoint(outer, canvasPoint({ x: inner.x, y: inner.y })) &&
```
seanparsons, line 277: More a note for myself, I wonder if there's a way of implementing this that doesn't necessitate all these allocations.

editor/src/components/canvas/canvas-wrapper-component.tsx
bkrmendy: totally valid point, I'd actually turn this into a function so that it would have to explicitly called before an areas select session starts

seanparsons (approved):
I'm a little wary of the number of things that are doing a "for every element in the project" type operation, but otherwise looks great!

editor/src/components/canvas/canvas-wrapper-component.tsx
Rheeseyb: Yeah, a bunch of these selectors have been added that only really need to be run as part of an area selection, meaning the canvas will be doing a whole bunch of extra computation for no reason 99% of the time.

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -269,15 +309,173 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
...
   const { maybeHighlightOnHover, maybeClearHighlightsOnHoverEnd } = useMaybeHighlightElement()
 
-  const { onMouseMove, onMouseDown, onMouseUp } = useSelectAndHover(
-    cmdKeyPressed,
-    setLocalSelectedViews,
+  const selectModeHooks = useSelectAndHover(cmdKeyPressed, setLocalSelectedViews)
+
+  const ref = React.useRef<HTMLDivElement | null>(null)
+
+  const [mousePoint, setMousePoint] = React.useState<CanvasPoint | null>(null)
```
Rheeseyb, line 328: It seems like you've confused `WindowPoint` with `CanvasPoint` when dealing with the `mousePoint`, since it is being stored as a `CanvasPoint` but is actually a `WindowPoint`. But I'd say we don't need to (or want to) track the `mousePoint` in a `useState` here. We only need to use it inside `onMouseDown` (where we have access to the data on the event) and when updating the `selectionArea` (which should probably be happening inside `onMouseMove` rather than as a `useMemo`).

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -269,15 +309,173 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
...
+  )
+
+  const mousePointOnCanvas = React.useMemo(() => {
+    if (mousePoint == null) {
+      return null
+    }
+    return getCanvasPoint(mousePoint.x, mousePoint.y)
+  }, [mousePoint, getCanvasPoint])
+
+  const selectionArea = React.useMemo((): CanvasRectangle | null => {
```
Rheeseyb, line 350: I think `selectionArea` would probably work better as a `useState` (set via the mouse event handlers) than a `useMemo`

editor/src/components/canvas/controls/new-canvas-controls.tsx
maltenuhn: >  confused `WindowPoint` with `CanvasPoint`

Is that a root cause of the various hard-to-repro bugs I experienced last week?

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -269,15 +309,173 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
...
+  }, [mousePoint, selectionAreaStart])
+
+  const canSelectArea = React.useMemo((): boolean => {
+    if (selectionArea != null || isSelectModeWithArea(editorMode)) {
+      return true
+    }
+    if (editorMode.type !== 'select') {
+      return false
+    }
+    return true
```
Rheeseyb, line 364: This can all be simplified to `return selectionArea != null || isSelectModeWithArea(editorMode)`, and shouldn't be a `useMemo` (it could either be inlined or become `function canSelectArea(...)`)

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -269,15 +309,173 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
...
+
+  const isValidMouseEventForSelectionArea = React.useCallback(
+    (e: React.MouseEvent): boolean => {
+      if (mousePoint == null) {
+        return false
+      }
+      return e.button === 0 && !(e.shiftKey || e.metaKey || e.ctrlKey || e.altKey)
+    },
+    [mousePoint],
+  )
```
Rheeseyb, line 375: This `useCallback` should be a top level `function` that doesn't require the `mousePoint` check (since it is only used in event handlers that provide the mouse point)

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -269,15 +309,173 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
...
+
+  const selectionAreaCanvasRect: CanvasRectangle | null = React.useMemo(() => {
+    if (selectionArea == null || !canSelectArea) {
+      return null
+    }
+    return rectangleFromTLBR(
+      getCanvasPoint(selectionArea.x, selectionArea.y),
+      getCanvasPoint(selectionArea.x + selectionArea.width, selectionArea.y + selectionArea.height),
+    )
+  }, [selectionArea, getCanvasPoint, canSelectArea])
```
Rheeseyb, line 385: I'd suggest inlining this inside the event handler rather than having an extra `useMemo` here

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -269,15 +309,173 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
...
+  )
+
+  const onMouseDown = React.useCallback(
+    (e: React.MouseEvent<HTMLDivElement>) => {
+      const possibleElementsUnderMouse = getPossibleElementsUnderMouse(
+        storeRef.current.jsxMetadata,
+        localSelectedViews,
+      )
+      if (
+        !possibleElementsUnderMouse.some(elementIsUnderMouse(mousePointOnCanvas)) &&
```
Rheeseyb, line 457: I think these two are essentially checking if there are any `highlightedViews`, since a mousedown event on one of those would select it. If that's the case we should just be checking that, meaning these two functions could be deleted.

editor/src/components/canvas/controls/selection-area-helpers.ts
```diff
@@ -0,0 +1,138 @@
...
+  (mousePointOnCanvas: CanvasPoint | null) =>
+  (element: ElementInstanceMetadata | null): boolean => {
+    return mousePointOnCanvas == null || element == null
+      ? false
+      : element.globalFrame != null &&
+          isFiniteRectangle(element.globalFrame) &&
+          rectContainsPoint(element.globalFrame, mousePointOnCanvas)
+  }
+
+export function getSelectionAreaRenderedRect(
```
Rheeseyb, line 116: I'm a bit confused by this function. Is this just applying the `canvasOffset` to the `selectionArea`?

editor/src/components/canvas/controls/selection-area-helpers.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+  return canvasRectangle({
+    x: selectionArea.x - boundingRect.x,
+    y: selectionArea.y - boundingRect.y,
+    width: selectionArea.width,
+    height: selectionArea.height,
+  })
+}
+
+export function makeSelectionArea(from: CanvasPoint, to: CanvasPoint): CanvasRectangle {
```
Rheeseyb, line 131: We have an existing function `rectFromTwoPoints` which I think should be doing exactly this...

editor/src/core/shared/math-utils.ts
Rheeseyb: This is a good point. I think given how many times this calculation would be run whilst drawing a selection area, we should avoid these extra allocations by hand rolling the calculation (i.e. `outer.x < inner.x && outer.y < inner.y && ...`)

ruggi: good point - fixed here! https://github.com/concrete-utopia/utopia/pull/3750/commits/497ce98368a5b5f21529ffc55175d5aa09288a3d

editor/src/components/canvas/controls/new-canvas-controls.tsx
ruggi: makes total sense, I got rid of it and made it local to `onMouseMove` here https://github.com/concrete-utopia/utopia/pull/3750/commits/b342869dcca26246b84df8bb7caca117f9082fcc

editor/src/components/canvas/controls/new-canvas-controls.tsx
ruggi: yeah indeed, moved to a helper https://github.com/concrete-utopia/utopia/pull/3750/commits/b342869dcca26246b84df8bb7caca117f9082fcc

editor/src/components/canvas/controls/new-canvas-controls.tsx
ruggi: moved to a helper https://github.com/concrete-utopia/utopia/pull/3750/commits/b342869dcca26246b84df8bb7caca117f9082fcc

editor/src/components/canvas/controls/new-canvas-controls.tsx
ruggi: removed basically all memoized vars and made them local to the `onMouseDown` handler (https://github.com/concrete-utopia/utopia/pull/3750/commits/b342869dcca26246b84df8bb7caca117f9082fcc)

editor/src/components/canvas/controls/new-canvas-controls.tsx
ruggi: yes and no - highlighted views are updated when there is no selection, but if there is a selection, moving the mouse over a selected element won't highlight it again (and I believe this is the correct behaviour), so this would fail and prevent the user from being able to move the selected elements (it would restart a new selection area)

editor/src/components/canvas/controls/selection-area-helpers.ts
ruggi: ah - great shout, used it here https://github.com/concrete-utopia/utopia/pull/3750/commits/b342869dcca26246b84df8bb7caca117f9082fcc

editor/src/components/canvas/controls/new-canvas-controls.tsx
ruggi: my bad, indeed I got confused

editor/src/components/canvas/controls/selection-area-helpers.ts
ruggi: kinda, but not just the canvas offset - by using the bounding rect it also takes into account both the navigator _and_ the code editor (but maybe there's a better way?)

editor/src/components/canvas/controls/selection-area-helpers.ts
```diff
@@ -0,0 +1,145 @@
...
+      (path) => MetadataUtils.findElementByElementPath(metadata, path),
+      selectedViews,
+    ),
+  ]
+}
+
+export const getElementsUnderSelectionArea = (
+  metadata: ElementInstanceMetadataMap,
+  selectionAreaCanvasRect: CanvasRectangle,
+): ElementPath[] => {
```
gbalint, line 47: We already have functions to select elements under a point in `dom-lookup.ts`, e.g. `getAllTargetsAtPoint` which uses dom lookup or `getAllTargetsAtPointAABB` which uses the global frames of elements.

I think those could be reused, partially to avoid duplication, partially because they include all kinds of logic which is necessary to select the correct elements. E.g. as far as I see with this PR you can select locked and even hidden elements, also elements which are behind other elements in z-order (or is that deliberate?)

editor/src/components/canvas/controls/new-canvas-controls.tsx
Rheeseyb: In that case we could check if there is a selected element under the cursor too?

ruggi: like https://github.com/concrete-utopia/utopia/pull/3750/commits/b5153ca3c31dfa14e5b708a8a0195d0b50bbfbee ?

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -266,22 +307,194 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
...
+      selectionAreaStart,
+      getCanvasPoint,
+      pathTrees,
+      allElementProps,
+      localSelectedViews,
+      hiddenInstances,
+    ],
+  )
+
+  const onMouseMove = React.useCallback(
```
balazsbajorics, line 439: I think this can be written back to how it was: just directly pass in selectModeHooks.onMouseMove to the element

editor/src/components/canvas/controls/selection-area-helpers.ts
Rheeseyb: I believe this change has made it possible to highlight and select and element whilst 1px to the left of it

editor/src/components/canvas/canvas.ts
```diff
@@ -287,11 +281,21 @@ const Canvas = {
...
       }
     })
   },
+  getMousePositionCanvasArea(canvasPosition: CanvasPoint | null): CanvasRectangle | null {
+    if (canvasPosition == null) {
+      return null
+    }
+    return canvasRectangle({
+      x: canvasPosition.x - 1,
+      y: canvasPosition.y - 1,
```
balazsbajorics, line 294: isn't this shifting the click by a single pixel to the left? shouldn't this just be x;y here? or are these changes matching what was on master?

editor/src/components/canvas/canvas.ts
```diff
@@ -327,13 +361,13 @@ const Canvas = {
               height: frameWithPath.frame.height === 0 ? 1 : frameWithPath.frame.height,
             }),
             canvasRectangle({
-              x: canvasPosition.x - looseReparentThreshold,
-              y: canvasPosition.y - looseReparentThreshold,
-              width: 2 * looseReparentThreshold,
-              height: 2 * looseReparentThreshold,
+              x: canvasArea.x - looseReparentThreshold,
+              y: canvasArea.y - looseReparentThreshold,
+              width: canvasArea.width * 2 * looseReparentThreshold,
```
balazsbajorics, line 366: this looks wrong. originally the ` 2 * looseReparentThreshold` is a fixed value, 10 pixels. this will now be area width times 10, which is MASSIVE for a non-zero selection area! 

```suggestion
              width: canvasArea.width + 2 * looseReparentThreshold,
```

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -470,3 +532,30 @@ const RenderControlMemoized = React.memo(
...
+        style={{
+          border: `1px solid ${colorTheme.primary.value}`,
+          background: colorTheme.primary10.value,
+          position: 'absolute',
+          pointerEvents: 'none',
+          width: rectangle.width,
+          height: rectangle.height,
+          left: rectangle.x,
+          top: rectangle.y,
+          // zIndex: 100,
```
Rheeseyb, line 554:
```suggestion
```

editor/src/components/canvas/controls/select-mode/scene-label.tsx
```diff
@@ -92,6 +93,12 @@ const SceneLabel = React.memo<SceneLabelProps>((props) => {
   const offsetX = scaledFontSize
   const borderRadius = 3 / scale
 
+  const mode = useEditorState(
```
balazsbajorics, line 96: general tip that since this value is _only_ used in a callback, if you make this a useRefEditorState, then you can use isSelectModeWithArea(mode.current) in the callback, and then you don't need to incur a re-render of scene label on mode change, and you don't need to blow the reference equality of onMouseMove either!

editor/src/components/canvas/controls/selection-area-helpers.ts
```diff
@@ -0,0 +1,132 @@
...
+        elements.some((other) => {
+          return other.type !== 'scene' && EP.isDescendantOf(element.path, other.path)
+        })
+      ) {
+        return false
+      }
+
+      const parentScene = elements.find(isSceneChild(element))
+
+      // if the element is a schene child and there are storyboard children, skip it
```
Rheeseyb, line 89:
```suggestion
      // if the element is a scene child and there are storyboard children, skip it
```
:D

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -126,20 +133,34 @@ export function useMaybeHighlightElement(): {
...
+        !resizing &&
+        !inserting &&
+        !alreadyHighlighted &&
+        !isSelectModeWithArea(mode)
+      ) {
         dispatch([setHighlightedView(target)], 'canvas')
       }
     },
-    [dispatch, stateRef],
+    [dispatch, stateRef, mode],
```
balazsbajorics, line 147: same tip about turning mode into a useRefEditorState here and in maybeClearHighlightsOnHoverEnd too

editor/src/components/canvas/controls/selection-area-helpers.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+  return canvasRectangle({
+    x: selectionArea.x - boundingRect.x,
+    y: selectionArea.y - boundingRect.y,
+    width: selectionArea.width,
+    height: selectionArea.height,
+  })
+}
+
+export function makeSelectionArea(from: WindowPoint, to: WindowPoint): CanvasRectangle {
```
balazsbajorics, line 121: I think this should be returning a WindowRectangle!

editor/src/components/canvas/controls/selection-area-helpers.ts
balazsbajorics: if makeSelectionArea returns a WindowRectangle, and this here takes a WindowRectangle, and then this thing returns a WindowRectangle, that sounds more correct to me

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -169,48 +182,53 @@ export const NewCanvasControls = React.memo((props: NewCanvasControlsProps) => {
...
+              transform: canvasControlProps.scale < 1 ? `scale(${canvasControlProps.scale}) ` : '',
+            }}
+          >
+            <NewCanvasControlsInner
+              windowToCanvasPosition={props.windowToCanvasPosition}
+              localSelectedViews={localSelectedViews}
+              localHighlightedViews={localHighlightedViews}
+              setLocalSelectedViews={setSelectedViewsLocally}
+              setLocalHighlightedViews={setLocalHighlightedViews}
+              setSelectionAreaRectangle={setSelectionAreaRectangle}
```
balazsbajorics, line 225: it _looks like_ if you move SelectionAreaRectangle into NewCanvasControls inner, you can push the selectionAreaRectangle state down, so you don't need to re-render this thing. 

actually, extrapolating from that, it _looks like_ only SelectionAreaRectangle cares about selectionAreaRectangle, so if you'd move it in there, I _think_ you could skip re-rendering all the other canvas controls on selectionAreaRectangle change

editor/src/components/canvas/controls/selection-area-helpers.ts
Rheeseyb, line 121: This function is converting from a `WindowPoint` to a `CanvasRectangle` without taking into account the canvas offset, meaning there are situations where the resultant rectangle will be at the incorrect position. From reading the code I'm not sure why this isn't causing issues with the area select when the canvas has been scrolled

editor/src/components/canvas/controls/selection-area-helpers.ts
```diff
@@ -0,0 +1,132 @@
...
+    (element: ElementUnderSelectionArea) => (other: ElementUnderSelectionArea) => {
+      return other.type === 'scene' && EP.isDescendantOf(element.path, other.path)
+    }
+
+  const thereAreStoryboardChildren = elements.some(
+    (element) => element.type === 'regular' && !elements.some(isSceneChild(element)),
+  )
+
+  return elements
+    .filter((element) => {
```
Rheeseyb, line 76: There is something still not quite right about this algorithm. In the default project I can single click to select the `div` containing the Utopia logo when nothing is selected, but when using the area select I can only select the `Playground` component (which shouldn't be selectable) 
image

editor/src/components/canvas/canvas.ts
ruggi: this mimics what's on master without breaking the tests - the skew of 1 px to the northwest is compensated by the 1x1 area :)

editor/src/components/canvas/controls/selection-area-helpers.ts
ruggi: 🙈

editor/src/components/canvas/controls/new-canvas-controls.tsx
Rheeseyb: Yeah this is a good point, as this is constantly re-rendering `NewCanvasControlsInner` when area selecting with nothing under the rectangle

editor/src/components/canvas/controls/selection-area-helpers.ts
ruggi: this https://github.com/concrete-utopia/utopia/pull/3750/commits/b5d9eebf4a6499412c1fe996c0a23fbe84ea42dc should solve this, I included the same selectable views used in the regular select mode in the calculations (before the selection area-specific filters)

editor/src/components/canvas/controls/new-canvas-controls.tsx
ruggi: the reason for keeping that outside of `NewCanvasControlsInner` was to avoid having to deal with the canvas zoom / positioning, which affect all the canvas controls 🤔

editor/src/components/canvas/controls/selection-area-helpers.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/3750/commits/c5a9fea57b067302c5f2c212aa9047a2175afc6d

Rheeseyb (approved):
Thank you for bearing with us with this PR!

editor/src/components/canvas/controls/selection-area-helpers.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+  return {
+    x: selectionArea.x - boundingRect.x,
+    y: selectionArea.y - boundingRect.y,
+    width: selectionArea.width,
+    height: selectionArea.height,
+  } as WindowRectangle
+}
+
+export function makeSelectionArea(from: WindowPoint, to: WindowPoint): WindowRectangle {
```
gbalint, line 131: This seems to be the same function as `rectFromTwoPoints`

gbalint (approved):

editor/src/components/canvas/controls/selection-area-helpers.ts
ruggi: it does yes! gonna get rid of it

ruggi: https://github.com/concrete-utopia/utopia/pull/3750/commits/3c01789db22df57059c21634eb41aaf8101f985c

PR #3751, opened 2023-05-30
Dropping an element next to/onto itself

bkrmendy:
## Problem
Dropping a navigator entry right before or after itself should be a NOOP, and shouldn't generate undo frames. Neither of these hold up as of this PR.

## Fix
Disable dropping an element into itself via the parent outline, and check the boundary conditions of dropping an elements on a drop line.

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
@@ -111,6 +111,43 @@ export interface ConditionalClauseNavigatorItemContainerProps
   navigatorEntry: ConditionalClauseNavigatorEntry
 }
 
+function isDroppingToOriginalPosition(
+  metadata: ElementInstanceMetadataMap,
+  elementPathTree: ElementPathTreeRoot,
+  dropTargetHint: DropTargetHint,
+  elementPath: ElementPath,
+): boolean {
+  const parentMatches = EP.pathsEqual(
+    EP.parentPath(elementPath),
+    dropTargetHint.targetParent.elementPath,
+  )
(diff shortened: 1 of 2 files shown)


enidemi (approved):

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -111,6 +111,65 @@ export interface ConditionalClauseNavigatorItemContainerProps
   navigatorEntry: ConditionalClauseNavigatorEntry
 }
 
+function isDroppingToOriginalPosition(
```
ruggi, line 114: this function is very nice 👏

ruggi (approved):

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -111,6 +111,65 @@ export interface ConditionalClauseNavigatorItemContainerProps
...
+  if (
+    (dropTargetHint.targetIndexPosition.type === 'back' &&
+      EP.pathsEqual(siblings.at(0)?.elementPath ?? null, draggedElementPath)) ||
+    (dropTargetHint.targetIndexPosition.type === 'front' &&
+      EP.pathsEqual(siblings.at(-1)?.elementPath ?? null, draggedElementPath))
+  ) {
+    return true
+  }
+
+  // absolute, for the sake of completeness
```
balazsbajorics, line 162: you mention completeness, maybe we should assertNever on the indexPosition.type? or is that very _very_ unlikely that we introduce a new indexposition type?

balazsbajorics (approved):

PR #3752, opened 2023-05-30
Conditionals without siblings and only expression in them receive globalFrame from parent

gbalint:
**Description:**
This PR is part of a series of PRs to implement text editing for expressions in conditional branches.

To make it possible to select a conditional on the canvas, conditionals need a `globalFrame`.

However, normally conditionals receive their `globalFrame` from their active branch. And js expressions do not have `globalFrame` (or metadata), so when a conditional only has an expression in its active branch, the conditional itself doesn't have `globalFrame` neither. But this is the exact case when we want to allow text editing, and double click to text edit too!

In the simplest case, when the conditional doesn't have siblings in its parent, and only has an expression in its active branch, it makes sense to say that the globalFrame of the conditional should be the same as its parent. Because the real parent of the expression in the dom will be the parent element of the conditional.

So this PR handles this very simple special case and fills the `globalFrame` for these conditionals from their ancestor.

Note: `targetHonoursPropsSize` now returns false for non-jsxelements. I _think_ it doesn't make sense to set the size of any other type of elephants. This prevents the resize controls to appear for conditionals.

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -2280,7 +2280,17 @@ function fillSpyOnlyMetadata(
 }
 
 function fillMissingDataFromAncestors(mergedMetadata: ElementInstanceMetadataMap) {
-  let workingElements: ElementInstanceMetadataMap = { ...mergedMetadata }
+  const metadataWithGlobalContentBox = fillGlobalContentBoxFromAncestors(mergedMetadata)
+  const metadataWithConditionaGlobalFrames = fillConditionalGlobalFrameFromAncestors(
+    metadataWithGlobalContentBox,
+  )
+  return metadataWithConditionaGlobalFrames
+}
+
+function fillGlobalContentBoxFromAncestors(
+  metadata: ElementInstanceMetadataMap,
+): ElementInstanceMetadataMap {
+  const workingElements = { ...metadata }
 
   const elementsWithoutGlobalContentBox = Object.keys(workingElements).filter((p) => {
     return workingElements[p]?.specialSizeMeasurements.globalContentBoxForChildren == null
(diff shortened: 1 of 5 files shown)


bkrmendy (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2299,6 +2306,76 @@ function fillMissingDataFromAncestors(mergedMetadata: ElementInstanceMetadataMap
...
+
+      const activeBranch = element.conditionValue.active
+        ? condElement.whenTrue
+        : condElement.whenFalse
+
+      if (!isJSExpression(activeBranch)) {
+        return false
+      }
+
+      const parentOfConditionalPath = EP.parentPath(EP.fromString(p))
```
seanparsons, line 2334:
```suggestion
      const parentOfConditionalPath = EP.parentPath(element.elementPath)
```
I think that should work right?

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2299,6 +2306,76 @@ function fillMissingDataFromAncestors(mergedMetadata: ElementInstanceMetadataMap
...
+    },
+  )
+
+  // sorted, so that parents are fixed first
+  conditionalsWithNoSiblingsAndExpressionInActiveBranch.sort()
+
+  fastForEach(conditionalsWithNoSiblingsAndExpressionInActiveBranch, (pathStr) => {
+    const elem = workingElements[pathStr]
+
+    const condParentPathStr = EP.toString(EP.parentPath(EP.fromString(pathStr)))
```
seanparsons, line 2351: I think similarly this works here too:
```suggestion
    const condParentPathStr = EP.toString(EP.parentPath(elem.elementPath))
```

seanparsons (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2299,6 +2306,76 @@ function fillMissingDataFromAncestors(mergedMetadata: ElementInstanceMetadataMap
...
+
+    const condParentGlobalFrame = workingElements[condParentPathStr]?.globalFrame
+    const condParentGlobalContentBoxForChildren =
+      workingElements[condParentPathStr]?.specialSizeMeasurements.globalContentBoxForChildren
+    const localFrameFromCondParent = (() => {
```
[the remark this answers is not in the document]
gbalint: You mean like this? e06af624165d6a67b873b7ddb4afe7ab364dbecb

PR #3753, opened 2023-05-30
File patch commands update the timestamp

Rheeseyb:
**Problem:**
None of the strategy commands that update a file will also update the `lastRevisedTime`, which is required when checking if the file can be overwritten by the result of an `UPDATE_FROM_WORKER` action, which can under the right circumstances result in the current file being overwritten by a stale version.

**Fix:**
This fix introduces some helper functions to ensure that the `lastRevisedTime` field is updated whenever a command updates a file. This PR also introduces a check inside `renderTestEditorWithModel` which ensures any file updates will also increase the value of `lastRevisedTime`, to help us catch any regressions here.

**Todo:**
- [x] Update tests of the existing strategies
- [x] Fix a keyboard strategy test that appears to be victim of a race condition which is caught out by this

### DIFF
--- a/editor/src/components/canvas/commands/add-imports-to-file-command.ts
+++ b/editor/src/components/canvas/commands/add-imports-to-file-command.ts
@@ -14,7 +14,7 @@ import {
 import { BaseCommand, WhenToRun, CommandFunction, CommandFunctionResult } from './commands'
 import { Spec } from 'immutability-helper'
 import { ProjectContentTreeRoot } from '../../../components/assets'
-import { createProjectContentsPatch } from './patch-utils'
+import { patchParseSuccessAtFilePath } from './patch-utils'
 
 export interface AddImportsToFile extends BaseCommand {
   type: 'ADD_IMPORTS_TO_FILE'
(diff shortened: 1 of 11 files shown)


ruggi (approved):

editor/src/components/canvas/commands/add-imports-to-file-command.ts
```diff
@@ -39,52 +39,19 @@ export const runAddImportsToFile: CommandFunction<AddImportsToFile> = (
   editorState: EditorState,
   command: AddImportsToFile,
 ): CommandFunctionResult => {
-  const updatedEditorState = modifyParseSuccessAtPath(
+  const editorStatePatch = patchParseSuccessAtFilePath(
     command.targetFile,
     editorState,
     (parseSuccess) => {
+      const updatedImports = mergeImports(command.targetFile, parseSuccess.imports, command.imports)
```
seanparsons, line 46: That's cleaner!

seanparsons (approved):

bkrmendy (approved):

PR #3754, opened 2023-05-31
Rename GroupLike to FragmentLike

balazsbajorics:
Mini PR, massive diff: I am about to introduce helper functions about _real_ groups, and to avoid the naming confusion, I am renaming all variations of GroupLike, ContentAffecting and ChildrenAffecting to FragmentLike.

I decided that the unified FragmentLike name is better for all existing cases, because everywhere we used GroupLike, ContentAffecting and ChildrenAffecting, we actually describe a singular shared behavior, that of the Fragment-like elements. 

Fragment-like elements today are: Fragment, Conditional expression, and for some cases the sizeless div.


editor/src/components/canvas/canvas-strategies/strategies/group-like-helpers.ts
```diff
@@ -175,19 +164,19 @@ export function getElementContentAffectingType(
   }
 
   if (MetadataUtils.isFlexLayoutedContainer(elementMetadata)) {
-    // for now, do not treat flex parents ever as content-affecting / group-like
+    // for now, do not treat flex parents ever as fragment-like / group-like
```
enidemi, line 167: group-like can be deleted from this comment

editor/src/components/canvas/canvas-strategies/strategies/group-like-helpers.ts
enidemi, line 1: I would rename this to file to `fragment-like-helpers`

balazsbajorics: yeah good shout

PR #3757, opened 2023-05-31
Feature/paste-same-size-absolute

bkrmendy:
31-12-ga2jl-gzbgw

## Problem
When pasting multiple elements in an absolute layout, many times it's super convenient to paste the same element many times, so that later on they can be combined into a list or gallery view for example. Right now, the elements are pasted on top of each other, which slows down the UI building flow. We already have this implemented for Flex layouts, but this feature is missing in absolute layouts.

## Fix
Implement a fix for absolute layout, and at the same time, centralise how an element is placed to a certain absolute position on paste.


enidemi (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2781,35 +2819,26 @@ export const UPDATE_FNS = {
...
-        return true
-      }
-      const parentPath = EP.parentPath(action.pasteInto.intendedParentPath)
-      if (findMaybeConditionalExpression(parentPath, editor.jsxMetadata) != null) {
-        // TODO invariant violation!
-        return true
-      }
-      return false
+    if (target == null) {
+      // TODO: toast
```
Rheeseyb, line 2836: Don't forget me before merging :)

Rheeseyb (approved):

PR #3763, opened 2023-06-01
Replace lastRevisedTime with versionNumber

Rheeseyb:
**Problem:**
In #3753 we added a check to ensure that all file updates were updating `lastRevisedTime`, but that makes any tests using fake timers very fragile, as it can result in the same timestamp being used.

**Fix:**
Replaced `lastRevisedTime` (which was always set using `Date.now()`) with `versionNumber` (which is now incremented each time).

There is one caveat that `lastRevisedTime` meant we didn't need access to the previous value when just overwriting a file, but it turns out that in all cases where we were doing that we could actually access the file, so some tests have been updated accordingly.


editor/src/core/shared/project-file-types.ts
```diff
@@ -556,36 +556,36 @@ export interface TextFile {
   fileContents: TextFileContents
   lastSavedContents: TextFileContents | null // it is null when the file is saved
   lastParseSuccess: ParseSuccess | null
-  lastRevisedTime: number
+  versionNumber: number
```
seanparsons, line 559: The similar type in `ClientModel.hs` needs to be updated too.

seanparsons (approved):

seanparsons (approved):

balazsbajorics (approved):
you made this look easy!

editor/src/core/shared/project-file-types.ts
Rheeseyb: Good catch! This was fixed in https://github.com/concrete-utopia/utopia/pull/3763/commits/c03047d5c6c9fbac89d53128e59eb63176606642

PR #3764, opened 2023-06-01
Conditionals with expressions should be selectable on the canvas

gbalint:
**Description:**
Conditionals with an expression in the active branch should be selectable on the canvas. This is necessary, because we want to be able to double click to text edit the expression in them.

Since https://github.com/concrete-utopia/utopia/pull/3752 , conditionals with expressions inherit their globalFrame from their parent (when they do not have siblings).

But this alone is not enough to make `findValidTarget` return a conditional like that.

Why is that?

In `findValidTarget` we find all the dom elements under the mouse, and then search upwards in the elementpath hierarchy to find the first ancestor which is a valid target. This makes it possible to find elements which do not appear in the dom (e.g. fragments), and return them as valid targets, but only when the element has a descendant which appear in the dom!

In the current situation this is not the case. A string expression is a leaf, and its parent (the conditional) is one level above the leaf. Neither of them appear on the dom, so the existing algorithm will never find them.

I extended the search algorithm in the following way:
- when there is a valid target in the dom, check if it has a child which is under the mouse (according to its globalFrame)
- if that child has a valid element path to choose from, return that child instead of the original target (we always prefer the deepest possible target).


editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -651,6 +651,99 @@ describe('Select Mode Double Clicking With Fragments', () => {
...
+    checkSelectedPaths(renderResult, [desiredPaths[0]])
+
+    await doubleClick()
+    checkFocusedPath(renderResult, null)
+    checkSelectedPaths(renderResult, [desiredPaths[1]])
+
+    await doubleClick()
+    checkFocusedPath(renderResult, null)
+    // can not dive deeper into the conditional because it has a sibling
+    checkSelectedPaths(renderResult, [desiredPaths[1]])
```
balazsbajorics, line 698: something is fishy here. the desiredPaths in these tests should not contain paths you cannot reach!

balazsbajorics: actually all these tests are missing an assertion that the desiredPaths were eventually exhausted

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -1096,7 +1189,16 @@ export const Card = (props) => (
     <Row style={{ minHeight: 40, gap: 12 }} data-uid="Card-Row-Buttons">
       <Button data-uid="Card-Button-1">Hello</Button>
       <Button data-uid="Card-Button-2">Button</Button>
-      <Button data-uid="Card-Button-3">Button</Button>
+      <Button data-uid="Card-Button-3">${
+        conditional
+          ? `{
+        // @utopia/uid=cond
```
balazsbajorics, line 1195: should I see a test case in this file where clicking eventually reaches `.../cond/<expressionUID>`?

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -121,12 +131,22 @@ export function findFirstParentWithValidElementPathInner(
       for (const validPath of validStaticElementPaths) {
```
balazsbajorics, line 131: I _think_ I just realized something! the reason this function needs to be called with both `allow-descendants-from-path` and `search-dom-only` is because the `allow-descendants-from-path` case cannot return with the dom element! If that is true, then I think this code could be vastly simplified here, for future ease of maintenance

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
gbalint: No, we only select the conditional and not the expression. But with the conditional selected you will be able to edit the content of the active branch.

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -165,32 +186,103 @@ export function findFirstParentWithValidElementPathInner(
...
+    staticPath: ElementPath
+    dynamic: ElementPath
+  },
+  nonDomPath: ElementPath,
+  point: CanvasPoint,
+  dynamicElementPaths: ElementPath[],
+  metadata: ElementInstanceMetadataMap,
+): ElementPath | null {
+  // check if there is a non-dom path which has a frame which contains the mouse point
+  // and which is a direct child of a dom path
```
balazsbajorics, line 225: actually, do we need the direct child restriction here? shouldn't it be enough that the non-dom path is a descendant of the last dom element without any dom elements in the middle? actually does it matter if we even filter for that? we always want to find the deepest valid path anyways.

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
gbalint: I doesn't contain paths we can not reach. I just explicitly made another double click with the same assertion to show that another double click doesn't change the selected element.

gbalint: I'm not sure what you mean, I think this test proves that the desiredPaths is exhausted.

balazsbajorics: ah ok, so desiredPaths.length in this case is 2?

editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
balazsbajorics: aha! so should I see a test where the path eventually reaches `../cond`?

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -165,32 +186,103 @@ export function findFirstParentWithValidElementPathInner(
...
         parentSceneValidPathsCache,
+        metadata,
+        point,
       )
     }
   }
 }
 
+function extendValidPathsWithNonDomParent(
+  domPath: {
```
seanparsons, line 197: Could this be extracted into a type?

seanparsons: Looks like the definition is used more than once as well.

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -165,32 +186,103 @@ export function findFirstParentWithValidElementPathInner(
...
+    static: string
+    staticPath: ElementPath
+    dynamic: ElementPath
+  },
+  nonDomPath: ElementPath,
+): ElementPath | null {
+  // check if there is a dom path which is a descendant of a non-dom path: then the
+  // non-dom path should be a valid target too
```
seanparsons, line 205: This comment says that the "...non-dom path should be a valid target too", but it doesn't return `nonDomPath`, which seems odd.

seanparsons (approved):

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -165,32 +186,103 @@ export function findFirstParentWithValidElementPathInner(
...
+  if (EP.isDescendantOfOrEqualTo(domPath.staticPath, nonDomPath)) {
+    const depthDiff = EP.fullDepth(domPath.staticPath) - EP.fullDepth(nonDomPath)
+    return EP.nthParentPath(domPath.dynamic, depthDiff)
```
Rheeseyb, line 208: This is a bit confusing... We've determined that A is a descendant of B, then calculated the difference in the depths of those to be N, and then are moving up the hierarchy N steps from A... So they why is the result not just B? Or is it B?

Rheeseyb: Oh, because it is potentially a dynamic path to an instance of B! Ok can you add a comment about that please

Rheeseyb (approved):

Rheeseyb (approved):

PR #3767, opened 2023-06-02
Improve selection change (amongst others) performance.

seanparsons:
**Problem:**
Currently selection changes are still quite an expensive operation mostly because there is a change to the metadata and the inspector needs to be largely completely re-rendered.

As previously mentioned in https://github.com/concrete-utopia/utopia/pull/3755, it was observed that once `MetadataUtils.getChildrenOrdered` is made a lot faster, then it could replace `MetadataUtils.getChildrenUnordered` as that is currently performing a O(n) operation every time it is called.

**Fix:**
The primary part of this fix has just been to replace `getChildrenUnordered` with `getChildrenOrdered` and `getChildrenPathsUnordered` with `getChildrenPathsOrdered`, which necessitated a lot of threading instances of `ElementPathTrees` to the numerous locations where each of those were called.

A secondary change was made to remove the use of `ComplexMap` from `getElementWarningsInner` as that was causing a surprising 4-5ms of wastage during each call to the latter.

**Results:**
The top level result is that the mousedown handler goes from an average of 140ms to 120ms to process, which was measured by ignoring the first result and then taking the average of 3 subsequent measurements.

The cost change can be observed here comparing the before and then after profiles:
before
after

The entire `editorDispatch` call highlighted goes from around 19ms to around 10ms, because the `deriveState` call has massively shrunk as `getElementWarningsInner` almost disappears off the profile.

On the right hand side of the first profile the `getNameAndIconResult` can also be seen dominating that side, in the after case it has dropped to 1/10th of the profile time.

**Commit Details:**
- Removed `ComplexMap` from the return type of `getElementWarningsInner`.
- Replaced calls to `getChildrenUnordered` with `getChildrenOrdered`.
- Replaced calls to `getChildrenPathsUnordered` with `getChildrenPathsOrdered`.
- Added `reorderSubTreeChildren` which extracts out some special case logic so we can run it on a chunk of children, not necessarily on the entire listing of elements.
- Lots of functions now have 1 or more instances of `ElementPathTrees` added to their parameters, often to pass through into other functions where they've similarly had them added.


editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.tsx
```diff
@@ -57,7 +57,11 @@ export function ancestorMetaStrategy(
     const unrolledChild = unrolledChildren[0]
 
     // Is the selected element an only child?
-    const siblings = MetadataUtils.getSiblingsUnordered(canvasState.startingMetadata, unrolledChild)
+    const siblings = MetadataUtils.getSiblingsUnordered(
```
Rheeseyb, line 60: At this point is it not just using `getSiblingsOrdered`? In which case we can bin this off too

editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.tsx
```diff
@@ -248,13 +255,18 @@ const getAutoLayoutSiblings = memoize(getAutoLayoutSiblingsInner, { maxSize: 1 }
...
   jsxMetadata: ElementInstanceMetadataMap,
+  pathTrees: ElementPathTrees,
   targets: Array<ElementPath>,
 ): Array<ElementInstanceMetadata> {
   if (!targets.every((t) => EP.isSiblingOf(targets[0], t))) {
     // this function only makes sense if the targets are siblings
     return []
   }
-  return MetadataUtils.getSiblingsParticipatingInAutolayoutUnordered(jsxMetadata, targets[0])
+  return MetadataUtils.getSiblingsParticipatingInAutolayoutUnordered(
```
Rheeseyb, line 265: Likewise with this, we already have an `Ordered` version of this function

editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.tsx
seanparsons: Oooh, yes, I was meaning to look back at this after creating the PR.

editor/src/components/canvas/canvas-strategies/strategies/fragment-like-helpers.ts
```diff
@@ -56,23 +59,34 @@ export const replaceFragmentLikePathsWithTheirChildrenRecursive = memoize(
...
+    const elementIsFragmentLike = treatElementAsFragmentLike(
+      metadata,
+      allElementProps,
+      pathTrees,
+      path,
+    )
 
     if (elementIsFragmentLike) {
-      const children = MetadataUtils.getChildrenPathsUnordered(metadata, path) // I think it's fine to get the unordered children here?
+      const children = MetadataUtils.getChildrenPathsOrdered(metadata, pathTrees, path) // I think it's fine to get the unordered children here?
```
Rheeseyb, line 76: Haha this comment can go in the bin now!

editor/src/components/canvas/canvas-strategies/strategies/fragment-like-helpers.ts
```diff
@@ -89,15 +103,21 @@ export const replaceNonDOMElementPathsWithTheirChildrenRecursive = memoize(
...
+    const elementIsNonDOMElement = isElementNonDOMElement(
+      metadata,
+      allElementProps,
+      pathTrees,
+      path,
+    )
 
     if (elementIsNonDOMElement) {
-      const children = MetadataUtils.getChildrenPathsUnordered(metadata, path) // I think it's fine to get the unordered children here?
+      const children = MetadataUtils.getChildrenPathsOrdered(metadata, pathTrees, path) // I think it's fine to get the unordered children here?
```
Rheeseyb, line 120: And here

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -680,23 +692,25 @@ export const MetadataUtils = {
...
   getImmediateChildrenPathsUnordered(
     elements: ElementInstanceMetadataMap,
+    elementPathTree: ElementPathTrees,
     target: ElementPath,
   ): Array<ElementPath> {
     const element = MetadataUtils.findElementByElementPath(elements, target)
     if (element == null) {
       return []
     } else {
       const rootPaths = MetadataUtils.getRootViewPathsUnordered(elements, target)
```
Rheeseyb, line 702: Looks like this should be replaced with the Ordered version

editor/src/components/canvas/canvas-strategies/strategies/ancestor-metastrategy.tsx
seanparsons: Fixed in d1565dc9d.

editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.tsx
seanparsons: Fixed in d1565dc9d.

editor/src/components/canvas/canvas-strategies/strategies/fragment-like-helpers.ts
seanparsons: Fixed in d1565dc9d.

editor/src/core/model/element-metadata-utils.ts
seanparsons: Fixed in d1565dc9d.

bkrmendy (approved):

Rheeseyb (approved):

PR #3768, opened 2023-06-05
Refactoring findFirstParentWithValidElementPath

gbalint:
**Problem:**
`findFirstParentWithValidElementPath` us very hard to read and reason about.

**Why is it hard to read:**
I identified the following problems in the code:

We have two separate algorithms to find the best parent with a valid element path, and how they are mixed together is quite confusing. The `allowDescendantsFromPath` parameter describes the two cases, and the function contains a long and complex recursive search, which runs for both cases.
In both searches we collect a list of potential targets (`filteredValidPathsMappedToDynamic`), and then we choose the deepest one from these. Then in the final step we choose the deeper result from the two `allowDescendantsFromPath` runs.

These two cases can be merged quite seamlessly. Both do an iteration over the valid paths, and both do a maximum search for depth after collecting all results. It seems like both do resursive search, but in real life only the `'search-dom-only'` case is really recursive.

**My solution to clean up the code**

1. There is no reason to run the two `allowDescendantsFromPath` algorithms separately. We can just do a single iteration and for all valid path items do both searches. This means we can also omit choosing the deeper path result from the two runs, moreover, we can remove the `inner` version (`findFirstParentWithValidElementPathInner`) of the function.
2. There is no reason to collect all potential valid targets and then iterate again to choose the deepest one among them:  we can just do the maximum search simultaneously with the search itself and always store the currently deepest result in a variable.
3. There is no reason to have a recursion for the whole search, when the recursion in only needed for the `'search-dom-only'` version of the search, and only when searching for a single element path (inside the loop).
4. Even that recursion can easily be converted to an iteration.

### DIFF
--- a/editor/src/components/canvas/dom-lookup.ts
+++ b/editor/src/components/canvas/dom-lookup.ts
@@ -50,51 +50,22 @@ export function findParentSceneValidPaths(
   }
 }
 
-// Take a DOM element, and try to find the nearest selectable path for it
-export function findFirstParentWithValidElementPath(
-  validDynamicElementPathsForLookup: Set<string> | 'no-filter',
-  target: Element,
-): ElementPath | null {
-  const parentSceneValidPathsCache = new Map()
-
-  const firstParentFromDom = findFirstParentWithValidElementPathInner(
-    validDynamicElementPathsForLookup,
-    target,
-    'search-dom-only',
-    parentSceneValidPathsCache,
-  )
-
-  const firstParentFromPath = findFirstParentWithValidElementPathInner(
-    validDynamicElementPathsForLookup,
-    target,
-    'allow-descendants-from-path',
-    parentSceneValidPathsCache,
-  )
-
-  if (firstParentFromDom == null || firstParentFromPath == null) {
-    return firstParentFromDom ?? firstParentFromPath
-  }
-
-  return EP.fullDepth(firstParentFromDom) < EP.fullDepth(firstParentFromPath)
-    ? firstParentFromPath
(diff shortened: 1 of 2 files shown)


seanparsons (approved):
+1 for your notes from this PR to end up as comments in some form in the code.

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -114,60 +85,72 @@ export function findFirstParentWithValidElementPathInner(
       }
     }
   }
+  return validStaticElementPaths
+}
+
+// Take a DOM element, and try to find the nearest selectable path for it
+export function findFirstParentWithValidElementPath(
```
balazsbajorics, line 92: this also returns the element itself if that has the selectable path, right? so both the comment and the function name are a bit wrong, because it's nearest parent OR the element itself

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -114,60 +93,66 @@ export function findFirstParentWithValidElementPathInner(
...
       }
-      break
-    default:
-      const _exhaustiveCheck: never = allowDescendantsFromPath
-  }
+    }
+
+    // 2. Start to traverse the DOM elements upwards in the hierarchy.
+    // When we find an element which is attached to a static path which is also in the valid path list,
+    // the dynamic version of that path is a valid target
```
balazsbajorics, line 137: possibly minor but I think we can skip this walk up the tree if we know that the depth will not be able to be greater than maxDepth

PR #3774, opened 2023-06-06
Disable copying when an element has a child that refers to a value "elsewhere"

bkrmendy:
## Problem
The `elementReferencesElsewhere` check doesn't include the `children` array of JSXElement. This can lead to crashing the canvas on paste.

In a related way, we don't check whether a a conditional's condition refers to a variable elsewhere.

## Solution
Extend `elementReferencesElsewhere` so that it checks the `children` array of JSXElements, and checks the condition of conditionals.

Also, this PR refactors `elementReferencesElsewhere` so that it reuses `attributeReferencesElsewhere` for JSExpressions, so that no duplicated code is present among the two functions.

### DIFF
--- a/editor/src/core/shared/element-template.ts
+++ b/editor/src/core/shared/element-template.ts
@@ -866,6 +866,7 @@ export function elementReferencesElsewhere(element: JSXElementChild): boolean {
       return element.children.some(elementReferencesElsewhere)
     case 'JSX_CONDITIONAL_EXPRESSION':
       return (
+        elementReferencesElsewhere(element.condition) ||
         elementReferencesElsewhere(element.whenTrue) ||
         elementReferencesElsewhere(element.whenFalse)
       )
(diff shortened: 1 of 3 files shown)


editor/src/core/shared/element-template.ts
```diff
@@ -854,7 +854,10 @@ export function jsxAttributesPartReferencesElsewhere(attrPart: JSXAttributesPart
 export function elementReferencesElsewhere(element: JSXElementChild): boolean {
   switch (element.type) {
     case 'JSX_ELEMENT':
-      return element.props.some(jsxAttributesPartReferencesElsewhere)
+      return (
+        element.props.some(jsxAttributesPartReferencesElsewhere) ||
+        element.children.some(elementReferencesElsewhere)
```
balazsbajorics, line 859: is this recursive? does it work for children deeply nested inside the element-to-copy?

balazsbajorics (approved):

editor/src/core/shared/element-template.ts
bkrmendy: It's recursive (since `elementReferencesElsewhere` checks JSXElements, and through the `children` prop, the all child elements/expressions are checked)

balazsbajorics (approved):

Rheeseyb (approved):

seanparsons (approved):

PR #3776, opened 2023-06-06
Forking uses passed model rather than stale context

Rheeseyb:
**Problem:**
The forking process in our persistence state machine uses the version of the project that is stored in the machine's context rather than the model that was passed in when asking the machine to save the model, meaning there is an edge case where auto-forking an old project could fail if migration steps were required.

**Fix:**
Use the model that was passed in as part of the save event when performing the fork. I've also added a test to ensure that the fork uses the latest version of the project.

### DIFF
--- a/editor/src/components/editor/persistence/generic/persistence-machine.ts
+++ b/editor/src/components/editor/persistence/generic/persistence-machine.ts
@@ -129,13 +129,15 @@ function innerSave<ModelType>(projectModel: ProjectModel<ModelType>): InnerSaveE
   }
 }
 
-interface ForkEvent {
+interface ForkEvent<ModelType> {
   type: 'FORK'
+  projectModel: ProjectModel<ModelType>
 }
 
-export function forkEvent(): ForkEvent {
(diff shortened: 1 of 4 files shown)


gbalint (approved):

editor/src/components/editor/persistence/generic/persistence-machine.ts
```diff
@@ -743,6 +748,7 @@ export function createPersistenceMachine<ModelType, FileType>(
                         },
                       }),
                     },
+                    // FIXME Error handling
```
seanparsons, line 751: Is this what I saw you (elsewhere) comment on that some further fix was needed?

seanparsons (approved):

editor/src/components/editor/persistence/generic/persistence-machine.ts
Rheeseyb: Yeah, I have a follow up PR I'll be opening in a sec that deals with this

PR #3778, opened 2023-06-07
Replace properties coming from code

bkrmendy:
## Description
This PR adds the code necessary to replace property values coming from code in `JSXElement`s and `JSXFragment`s.
Concretely,
- Component property values are inlined (such as `onClick` or `data-label`)
- Property values in nested objects are inlined (such as properties in `style`)
- Property spread assignments (such as `style={{ top: 10, left: 10, ...position }}`) are inlined.
- The inlining function also recurses into `children` of JSXElements and JSXFragments, and the true/false branches of conditionals.
The PR also adds a feature flag to gate these new features, as they are heavily under development.

### Out of scope
- The `ATTRIBUTE_OTHER_JAVASCRIPT` branch of `JSExpression` is not handled at all. This will be implemented on a follow-up PR (will be necessary to inline generated text)
- Conditional conditions are not inlined


editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-helpers.ts
```diff
@@ -56,6 +76,74 @@ export function isAllowedToReparent(
...
+      return right(target)
+    }
+    return left('Cannot find element metadata')
+  }
+
+  return foldEither(
+    (_) => right(target),
+    () => {
+      if (!MetadataUtils.targetHonoursPropsPosition(editor.projectContents, metadata)) {
+        return left('target does not honour positioning props')
```
enidemi, line 105: I think we should allow copying an element and on paste maybe we can't position it but it's still pasted

enidemi (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-helpers.ts
bkrmendy: great idea! removed it in [9503dfd](https://github.com/concrete-utopia/utopia/pull/3778/commits/9503dfd781b11a8a348357b305e752aad836deb0)

editor/src/components/editor/actions/actions.tsx
```diff
@@ -3094,64 +3098,125 @@ export const UPDATE_FNS = {
...
+              styleClipboard: [],
+              elements: copyData?.data ?? [],
+            },
+          }
+        },
+        dispatch,
+      )
+    }
+
+    const canReparent = sequenceEither(
```
seanparsons, line 3132: You probably want `traverseEither` here, it's effectively the map and sequence combined.

editor/src/components/editor/actions/actions.tsx
```diff
@@ -3094,64 +3098,125 @@ export const UPDATE_FNS = {
...
-          dispatch,
-        )
+            dispatch,
+          )
+        },
+        dispatch,
+      )
+    }
+
+    const canReparent = sequenceEither(
```
seanparsons, line 3194: Is this 15-ish lines the same as the block above?

editor/src/core/shared/element-template.ts
```diff
@@ -871,6 +871,53 @@ export function elementReferencesElsewhere(element: JSXElementChild): boolean {
...
+          : getElementReferencesElsewherePathsFromProps(
+              prop.value,
+              PP.append(pathSoFar, PP.create(prop.key)),
+            ),
+      )
+    case 'ATTRIBUTE_NESTED_OBJECT':
+      const spreads: JSXSpreadAssignment[] = []
+      const assigments: JSXPropertyAssignment[] = []
+      element.content.forEach((c) =>
+        c.type === 'PROPERTY_ASSIGNMENT'
```
seanparsons, line 892: I'd argue this would be clearer with regular if..else statements, as we're discarding the result.

seanparsons (approved):
Real nice!

ruggi (approved):

editor/src/components/editor/actions/actions.tsx
bkrmendy: great idea! added in https://github.com/concrete-utopia/utopia/pull/3778/commits/7908d7a67e7de467aa4278d5f0c2aa940653db39

editor/src/components/editor/actions/actions.tsx
bkrmendy: I was very much on the fence about this too, the reason I left in this duplication is that there isn't that much going on here:
- checking if an element can be copied + early return
- updating the clipboard
- returning the editor state or deleting the cut elements
The third step already can't be generalised among copy and cut, so I decided to leave the code as it is, and when the FS is cleaned up, clean up this part too.

seanparsons: Yeah, we can totally refactor this later on as an atomic thing, as opposed to it getting threaded into the work you're doing.

bkrmendy: OK, looking at @Rheeseyb 's comment, this isn't negotiable 😅

editor/src/core/shared/element-template.ts
bkrmendy: implemented in https://github.com/concrete-utopia/utopia/pull/3778/commits/92647e40a27037686556d8b42f8277f4e66eb548

Rheeseyb (changes requested):
As mentioned elsewhere, we can't use the `UidToPathLookup` since that would break for a number of cases, but we have access to the original path and original path tree, so that should be pretty easy to replace.

editor/src/components/editor/actions/actions.tsx
Rheeseyb: Yeah, the duplication shouldn't block this PR, since to tidy it up requires refactoring both action handlers. That can be tackled separately.

editor/src/components/editor/actions/actions.tsx
```diff
@@ -3094,64 +3099,127 @@ export const UPDATE_FNS = {
...
+    return {
+      ...editor,
+      pasteTargetsToIgnore: editor.selectedViews,
+      internalClipboard: {
+        styleClipboard: [],
+        elements: copyData?.data ?? [],
+      },
+    }
   },
   CUT_SELECTION_TO_CLIPBOARD: (
```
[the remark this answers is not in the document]
bkrmendy: reduced the duplication here [b71d842](https://github.com/concrete-utopia/utopia/pull/3778/commits/b71d8427d91cb1928b984f06c0456716f0ad5323)

editor/src/components/editor/actions/actions.tsx
bkrmendy: I generalized some of the logic here [b71d842](https://github.com/concrete-utopia/utopia/pull/3778/commits/b71d8427d91cb1928b984f06c0456716f0ad5323)

bkrmendy:
> As mentioned elsewhere, we can't use the `UidToPathLookup` since that would break for a number of cases, but we have access to the original path and original path tree, so that should be pretty easy to replace.

Implemented in [f2148ba](https://github.com/concrete-utopia/utopia/pull/3778/commits/f2148ba12a63fea24a2eadeae887308e8db0b474)

Rheeseyb (approved):

PR #3783, opened 2023-06-08
Fix: label for conditional branches with js expressions

### LINKED ISSUE #3782: Fix navigator label for conditional branches with js expressions

ruggi:
Fixes #3782 

**Problem:**

Conditional branches with js expressions show `(code)` in the navigator.

Screenshot 2023-06-08 at 11 58 17 AM

**Fix:**

Calculate the right label from the generated element's metadata, which need to be extracted from its dynamic path.

Screenshot 2023-06-08 at 11 58 30 AM

**Notes**
- This does not fix inline values like `{true ? (()=>'foo)() : <div />}` as they don't have metadata. There are some workarounds that can be done (e.g. looking up the children of `allElementProps` for the conditional parent, but none of the things I tried are actually nice. I think it should be addressed separately.
- This does not fix the fact that conditionals with both branches being js expressions look like inlined text values, imo it should be addressed separately as well.


editor/src/core/shared/element-template.ts
```diff
@@ -1066,6 +1066,14 @@ export function clearJSXFragmentWithoutUIDUniqueIDs(
 
 export type ElementsWithin = { [uid: string]: JSXElement }
 
+export interface WithElementsWithin {
```
seanparsons, line 1069: Could have the types that have this extend this interface.

seanparsons (approved):

editor/src/core/shared/element-template.ts
ruggi: aha! yeah good idea https://github.com/concrete-utopia/utopia/pull/3783/commits/d6f69f0101b8e160720fc2f9d2216496d6d32911

gbalint (approved):

PR #3785, opened 2023-06-08
Prevent the persistence state machine from committing failed changes

Rheeseyb:
**Problem:**
This is a follow up to #3776. If a save fails for whatever reason, the persistence machine would return to the Ready state, but still dispatch any queued up actions, meaning it would be committing any changes along the way. This meant that a failed fork would still dispatch an action updating the project name, ID and model (which in turn would enter an infinite loop of constantly failed forking)

**Fix:**
I've introduced a rollback state to the persistence machine's context, which we now use to rollback the project's ID and model stored in the context when reverting to the Ready state after a backend error.

Whilst the machine is transitioning between states, the interpreter will listen for certain events and use those to queue up actions to dispatch when it next enters the Ready state. Since we don't want to dispatch those after an error, I've updated the machine's interpreter to clear the queued actions on a backend error, and instead dispatch a toast informing the user of the error.

Finally, I've added a test that deliberately fails to fork a project the first time, and then successfully forks the second, so that we can check the dispatched actions and the context at each point.

### DIFF
--- a/editor/src/components/editor/persistence/persistence.spec.ts
+++ b/editor/src/components/editor/persistence/persistence.spec.ts
@@ -49,11 +49,11 @@ const SecondRevision = updateModel(FirstRevision)
 const ThirdRevision = updateModel(SecondRevision)
 const FourthRevision = updateModel(ThirdRevision)
 
-let forceNextProjectId: string | null = null
+let forcedNextProjectId: string | null = null
 
 function mockRandomProjectID(): string {
-  const newId = forceNextProjectId == null ? generateUID(allProjectIds) : forceNextProjectId
-  forceNextProjectId = null
+  const newId = forcedNextProjectId == null ? generateUID(allProjectIds) : forcedNextProjectId
+  forcedNextProjectId = null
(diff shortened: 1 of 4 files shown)


editor/src/components/editor/persistence/persistence.spec.ts
```diff
@@ -45,8 +49,11 @@ const SecondRevision = updateModel(FirstRevision)
 const ThirdRevision = updateModel(SecondRevision)
 const FourthRevision = updateModel(ThirdRevision)
 
+let forceNextProjectId: string | null = null
```
seanparsons, line 52: Astonishingly minor.
```suggestion
let forcedNextProjectId: string | null = null
```

Rheeseyb: Sorted in https://github.com/concrete-utopia/utopia/pull/3785/commits/d35df866409d3d39f11e6adcdd59aecb4acb7baf

gbalint (approved):

PR #3786, opened 2023-06-08
Interaction session for static reparenting

bkrmendy:
## Description 

This PR adds the necessary scaffolding for paste strategies. This PR itself doesn't introduce any new features, because
1. the paste strategies depend on https://github.com/concrete-utopia/utopia/pull/3778
2. the new interaction session can be tested the best when the actual paste strategies are in place.

The PR adds a placeholder strategy and a debug button to trigger it, and a way to set event listeners to commit the interaction session on `mousedown` and `keydown`.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
@@ -514,6 +514,7 @@ export function interactionInProgress(interactionSession: InteractionSession | n
         )
       case 'KEYBOARD':
       case 'HOVER':
+      case 'STATIC_REPARENT':
         return true
       default:
         const _exhaustiveCheck: never = interactionSession.interactionData
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
@@ -55,6 +55,7 @@ import { flexResizeStrategy } from './strategies/flex-resize-strategy'
 import { basicResizeStrategy } from './strategies/basic-resize-strategy'
 import { InsertionSubject, InsertionSubjectWrapper } from '../../editor/editor-modes'
 import { generateUidWithExistingComponents } from '../../../core/model/element-template-utils'
+import { testStaticReparentStrategy } from './strategies/test-static-reparent-strategy'
 
 export type CanvasStrategyFactory = (
   canvasState: InteractionCanvasState,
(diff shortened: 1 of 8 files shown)


editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -152,6 +153,16 @@ const keyboardShortcutStrategies: MetaCanvasStrategy = (
   )
 }
 
+const staticReparentStrategies: MetaCanvasStrategy = (
+  canvasState: InteractionCanvasState,
+  interactionSession: InteractionSession | null,
+): Array<CanvasStrategy> => {
+  return mapDropNulls(
+    (factory) => factory(canvasState, interactionSession),
+    [testStaticReparentStrategy],
```
seanparsons, line 162: Is it likely for there to be more values here? If not I'd just call `testStaticReparentStrategy` and default the result.

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -954,6 +954,45 @@ export function useClearKeyboardInteraction(editorStoreRef: {
...
+      window.removeEventListener('mousedown', clearStaticReparentInteractionOnMouseDown)
+      if (
+        editorStoreRef.current.editor.canvas.interactionSession?.interactionData.type ===
+        'STATIC_REPARENT'
+      ) {
+        dispatch([CanvasActions.clearInteractionSession(true)], 'everyone')
+      }
+    }
+
+    window.addEventListener('mousedown', clearStaticReparentInteractionOnMouseDown, {
```
seanparsons, line 972: I think it's possible for these event listeners to leak and then potentially duplicate, we often use `useEffect` for these as we can return a cleanup function that calls `removeEventListener`.

seanparsons (changes requested):
Other than the event listener handling this looks good!

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
bkrmendy: I'll remove this in the next PR anyway so I don't think there's much use in polishing this

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
bkrmendy: added a useEffect to clean the handlers up in https://github.com/concrete-utopia/utopia/pull/3786/commits/667daff4af8da84ad113f9f721d8a9035d31420e

enidemi (approved):

seanparsons (approved):
All of the thumbs up!

PR #3788, opened 2023-06-08
Stricter text editing requirements for conditionals

gbalint:
**Problem:**
We allow text editing of an expression inside the active conditional branch.
However, we should _not_ allow text editing when that expressions returns real elements/elefants, which appear in the metadata. Text editing only makes sense for leaf elements (which basically mean they just return text content).

**Fix:**
I created a new function called `isTextEditableConditional` which clearly lists all the requirement for a conditional to be text editable, and use that in the necessary helper functions.

### DIFF
--- a/editor/src/core/model/conditionals.ts
+++ b/editor/src/core/model/conditionals.ts
@@ -353,7 +353,7 @@ export function isTextEditableConditional(
     elementPathTree,
     nonConditionalAncestor,
   )
-  if (siblings.length > 1) {
+  if (siblings.length !== 1) {
     // we don't allow text editing of conditional branches when the conditional has siblings
     // (or if the topmost nested conditional has siblings)
     return false
--- a/editor/src/core/model/conditionals.ts
+++ b/editor/src/core/model/conditionals.ts
@@ -3,12 +3,13 @@ import * as EP from '../shared/element-path'
 import {
   ElementInstanceMetadata,
   ElementInstanceMetadataMap,
+  isJSExpression,
   isJSXConditionalExpression,
   isNullJSXAttributeValue,
   JSXConditionalExpression,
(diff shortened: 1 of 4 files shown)


ruggi (approved):

editor/src/core/model/conditionals.ts
```diff
@@ -330,3 +331,43 @@ export function findFirstNonConditionalAncestor(
...
+    // element is not a conditional
+    return false
+  }
+  const nonConditionalAncestor = findFirstNonConditionalAncestor(path, metadata)
+  const siblings = MetadataUtils.getChildrenOrdered(
+    metadata,
+    elementPathTree,
+    nonConditionalAncestor,
+  )
+  if (siblings.length > 1) {
```
ruggi, line 356: should this be a safer `!== 1` ?

gbalint: For me this is more readable, because it expresses that the problem is when there are real siblings (one "sibling" is the conditional itself, so 0 is not possible).

editor/src/core/model/conditionals.ts
```diff
@@ -330,3 +331,43 @@ export function findFirstNonConditionalAncestor(
...
+  }
+
+  const childrenInMetadata = MetadataUtils.getChildrenUnordered(metadata, path)
+  if (childrenInMetadata.length > 0) {
+    // we don't allow text editing of conditional branches when the conditional have children in metadata
+    // that would mean there are elefants in the active branch, which should not be text editable
+    return false
+  }
+
+  // Finally we check if the active conditional branch is a js expression. Maybe this is an overkill,
```
seanparsons, line 369: Love the comments!

seanparsons (approved):

PR #3793, opened 2023-06-09
Handle arrow functions that just return an identifier

Rheeseyb:
**Problem:**
There were a pair of very specific edge cases in the parser where a JS expression contained a function (arrow or regular) that just returned an identifier that was defined outside of that function, e.g.:

```ts
const hello = 'hi'
function sayHi() { return hello }
const sayHiAgain = () => hello 
```

Neither of these two would add the identifier `hello` to their array of `definedElsewhere`.

**Fix:**
Add those specific cases to the parser-printer.


ruggi (approved):
🏅

enidemi (approved):

bkrmendy (approved):

PR #3794, opened 2023-06-09
Restrict GITHUB_TOKEN permissions on all Discord workflows

Rheeseyb:
Lock down the GITHUB_TOKEN access for all workflows that don't need the token, so that we can re-enable the Discord integration for issues and discussions.


Rheeseyb:
Just double checking comments still work

PR #3796, opened 2023-06-09
Embedded conditionals can inherit globalFrame from their parent too

gbalint:
**Description:**
This is a followup fix to https://github.com/concrete-utopia/utopia/pull/3752
1. It allows conditionals to be embedded and still inherit their frames from the parent when the necessary conditions are met.
2. It only allows from inheritance from parent when the expression in the active branch do not return real elements (so `elementsWithin` is empty)

I added a long comment to `fillConditionalGlobalFrameFromAncestors` , because this has become way too complex, and there is no way anybody would understand why it works this way without the explicit explanation.

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -2330,7 +2330,7 @@ function fillGlobalContentBoxFromAncestors(
 // none of the conditionals have siblings.
 // Why is this useful? Because these expressions don't have globalFrame, so their conditional parent do not have
 // a globalFrame neither. But when these conditionals only contain text expressions, and they don't have siblings,
-// it makes sense to treat the whole parent element of the conditional as a container for the text expression itself.
+// it make sense to treat the whole parent element of the conditional as a container for the text expression itself.
 // Example: In the case `<div>Hello</div>`, we treat the whole div as text, we do not differentiate between the size of
 // Hello and the size of the div.
 // When we have the element `<div>{true ? 'Hello' : <div />}</div>`, the conditional appears between the <div> parent and
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -2330,7 +2330,7 @@ function fillGlobalContentBoxFromAncestors(
 // none of the conditionals have siblings.
 // Why is this useful? Because these expressions don't have globalFrame, so their conditional parent do not have
 // a globalFrame neither. But when these conditionals only contain text expressions, and they don't have siblings,
-// it make sense to treat the whole parent element of the conditional as a container for the text expression itself.
+// it makes sense to treat the whole parent element of the conditional as a container for the text expression itself.
 // Example: In the case `<div>Hello</div>`, we treat the whole div as text, we do not differentiate between the size of
 // Hello and the size of the div.
 // When we have the element `<div>{true ? 'Hello' : <div />}</div>`, the conditional appears between the <div> parent and
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -2342,9 +2342,9 @@ function fillConditionalGlobalFrameFromAncestors(
 ): ElementInstanceMetadataMap {
   const workingElements = { ...metadata }
 
-  const conditionalsWithNoSiblingsAndExpressionInActiveBranch = Object.keys(workingElements).filter(
+  const conditionalsWithNoSiblingsAndExpressionActiveBranch = Object.keys(workingElements).filter(
     (p) => {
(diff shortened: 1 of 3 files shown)


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2321,44 +2322,80 @@ function fillGlobalContentBoxFromAncestors(
...
+// none of the conditionals have siblings.
+// Why is this useful? Because these expressions don't have globalFrame, so their conditional parent do not have
+// a globalFrame neither. But when these conditionals only contain text expressions, and they don't have siblings,
+// it make sense to treat the whole parent element of the conditional as a container for the text expression itself.
+// Example: In the case `<div>Hello</div>`, we treat the whole div as text, we do not differentiate between the size of
+// Hello and the size of the div.
+// When we have the element `<div>{true ? 'Hello' : <div />}</div>`, the conditional appears between the <div> parent and
+// the Hello text, but we still want to be able to select and text edit the text content.
+// Maybe in the future we will treat text expressions as elefants, and they will have a measured size, but until then this is
+// a good compromise to give these conditionals/expressions selectability and editability on the canvas.
```
ruggi, line 2339: God's work 👏

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2321,44 +2322,80 @@ function fillGlobalContentBoxFromAncestors(
...
     (p) => {
-      const element = workingElements[p]
-      if (element.conditionValue === 'not-a-conditional') {
-        return false
-      }
-      const condElement =
-        isRight(element.element) && isJSXConditionalExpression(element.element.value)
-          ? element.element.value
-          : null
+      const isConditionalsWithNoSiblingsAndExpressionInActiveBranch = (
```
ruggi, line 2347: can I suggest to rename this to avoid the potential misunderstanding `InActive === Inactive`?

ruggi (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2321,44 +2322,80 @@ function fillGlobalContentBoxFromAncestors(
...
-// should be inherited from its parent
+// There is a case when conditionals should inherit their globalFrame from their parent:
+// - when a conditional has no siblings
+// - when the active branch is a js expression
+// - when that expression has no elementsWithin (so it is a leaf element in the hierarchy)
+// We also allow embedded conditionals where the deepest one has the correct active branch and
+// none of the conditionals have siblings.
+// Why is this useful? Because these expressions don't have globalFrame, so their conditional parent do not have
+// a globalFrame neither. But when these conditionals only contain text expressions, and they don't have siblings,
+// it make sense to treat the whole parent element of the conditional as a container for the text expression itself.
```
ruggi, line 2333:
```suggestion
// it makes sense to treat the whole parent element of the conditional as a container for the text expression itself.
```

editor/src/core/model/element-metadata-utils.ts
gbalint: oh, I did not think about this 🤦

bkrmendy (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2321,51 +2322,87 @@ function fillGlobalContentBoxFromAncestors(
...
+// the Hello text, but we still want to be able to select and text edit the text content.
+// Maybe in the future we will treat text expressions as elefants, and they will have a measured size, but until then this is
+// a good compromise to give these conditionals/expressions selectability and editability on the canvas.
 function fillConditionalGlobalFrameFromAncestors(
   metadata: ElementInstanceMetadataMap,
 ): ElementInstanceMetadataMap {
   const workingElements = { ...metadata }
 
-  const conditionalsWithNoSiblingsAndExpressionInActiveBranch = Object.keys(workingElements).filter(
+  const conditionalsWithNoSiblingsAndExpressionActiveBranch = Object.keys(workingElements).filter(
```
Rheeseyb, line 2345: It doesn't look like you're using the key anywhere here, so you might as well use `Object.values`, and do away with the creation of and immediate calling of the internal function

Rheeseyb (approved):

seanparsons (approved):

editor/src/core/model/element-metadata-utils.ts
gbalint: I would like to filter the keys not the values, even if i don't use them inside. And the internal function is recursive, that is why I need it.

Rheeseyb: > I would like to filter the keys not the values

I totally spaced on the fact that this was filtering the keys. Ok, as you were!

PR #3798, opened 2023-06-12
Paste strategies

bkrmendy:
# Try it [here](https://utopia.pizza/p/f5909611-before-i-go-lite-(forked)/?branch_name=feature-paste-strategies)

## Description
Building on https://github.com/concrete-utopia/utopia/pull/3786 and https://github.com/concrete-utopia/utopia/pull/3778, this PR adds two actual static reparent strategies: one for pasting with props replaced with their runtime values, and another that pastes all elements with no modifications.

This PR also removes the `Paste with props replaced` FS, since all the code is in place to make paste strategies work.

### Overview of how paste works with this PR
- When a paste event is fired, instead of directly inserting the elements from the clipboard into the project, a static reparent interaction session is started. If there was another ongoing interaction session when the paste event fired (such as an earlier static reparent session), it is committed into the project.
- One of the two existing paste strategies runs. It's possible to choose between the strategies via the strategy picker (see Limitations). Even if the canvas crashes (due to a `ReferenceError`), the strategy picker still remains visible, and users can toggle back to another strategy.
- The paste session ends on `mousedown`, or `keydown` if the key in question isn't a digit or `Tab` (which are needed to toggle between strategies)

## Limitations
- Paste to Replace still uses an action, and doesn't replace any props
- Navigator reparent still uses the original action
- The canvas strategy picker might not be the best UI to present the static reparent options. This will need more consideration, so it's left to a follow up PR


editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -1548,19 +1535,12 @@ export var storyboard = (
           await selectComponentsForTest(editor, [targetPath])
           await pressKey('c', { modifiers: cmdModifier })
 
-          const canvasRoot = editor.renderedDOM.getByTestId('canvas-root')
-
           FOR_TESTS_setNextGeneratedUids(['aaa', 'bbb', 'ccc', 'ddd'])
```
enidemi, line 1538: do you know why `FOR_TESTS_setNextGeneratedUids` is unreliable here?

enidemi (approved):

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -1447,55 +1464,25 @@ export var storyboard = (
...
-          )
-          `,
-          'await-first-dom-report',
-        )
+      describe('repeated paste', () => {
+        async function pasteNTimes(editor: EditorRenderResult, n: number) {
+          const canvasRoot = editor.renderedDOM.getByTestId('canvas-root')
 
-        await selectComponentsForTest(editor, [EP.fromString('sb/container')])
+          for (const _ in Array(n).fill(0)) {
```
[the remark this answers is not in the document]
bkrmendy: Of course, on the other hand I prefer the more "wholemeal" semantics of for/of loops compared to the index twiddling of traditional for loops. I have no strong opinions here, I can change this if it's too much.

editor/src/components/editor/actions/actions.spec.browser2.tsx
bkrmendy: I couldn't find out, so I removed it in https://github.com/concrete-utopia/utopia/pull/3798/commits/ab9ca56df93b493103d81604f97fc36fddaec1cf so that it doesn't cause any confusion

editor/src/components/editor/actions/actions.spec.browser2.tsx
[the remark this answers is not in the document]
gbalint: In normal cases I would agree with Berci, but here you don't use the loop variable, so I really don't think this is more expressive, but it absolutely unnecessarily allocates an array with 0-s.

editor/src/components/canvas/canvas-strategies/strategies/paste-metastrategy.tsx
```diff
@@ -0,0 +1,247 @@
...
+  (mode: PasteMode) =>
+  (
+    canvasState: InteractionCanvasState,
+    interactionSession: InteractionSession | null,
+  ): CanvasStrategy | null => {
+    if (interactionSession?.interactionData.type !== 'STATIC_REPARENT') {
+      return null
+    }
+
+    return mode === 'replace'
```
gbalint, line 189: I suggest to use inner or even anonymous functions and not embedded ternaries, they are so hard to parse.

editor/src/components/canvas/canvas-strategies/strategies/test-static-reparent-strategy.tsx
```diff
@@ -1,23 +0,0 @@
-import {
```
gbalint, line 1: I think this file can be deleted (it is not used in canvas-strategies anymore)

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -955,38 +955,47 @@ export function useClearKeyboardInteraction(editorStoreRef: {
...
+    mouseDownHandlerRef.current = (e) => {
       removeEventListeners()
-
       if (
         editorStoreRef.current.editor.canvas.interactionSession?.interactionData.type ===
         'STATIC_REPARENT'
       ) {
         dispatch([CanvasActions.clearInteractionSession(true)], 'everyone')
+        e.returnValue = 'Unsaved changes'
+        return 'Unsaved changes'
```
gbalint, line 984: Return value in a mouse event handler function? Is there a reason for this?

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -389,8 +393,8 @@ describe('actions', () => {
             <div data-uid='bbb'>foo</div>
             <div data-uid='ccc'>bar</div>
             <div data-uid='ddd'>baz</div>
-            <div data-uid='aad'>foo</div>
-            <div data-uid='aag'>bar</div>
+            <div data-uid='aac'>foo</div>
+            <div data-uid='aaf'>bar</div>
```
gbalint, line 397: Do we know what caused these uid changes?

editor/src/components/canvas/canvas-strategies/strategies/paste-metastrategy.tsx
```diff
@@ -0,0 +1,247 @@
...
+      if (commands.length === 0) {
+        return emptyStrategyApplicationResult
+      }
+
+      return strategyApplicationResult([updateSelectedViews('always', []), ...commands])
+    },
+  }
+}
+
+export const pasteMetaStrategy =
```
Rheeseyb, line 179: In our model we use `MetaStrategy` to represent something that will return an array of strategies, so rather than this taking the `mode` and deciding which one to return, this should just return the array of both strategies (after of course first checking the interaction type)

editor/src/components/canvas/canvas-strategies/strategies/paste-metastrategy.tsx
```diff
@@ -0,0 +1,247 @@
...
+              },
+              indexPosition,
+              canvasState.builtInDependencies,
+            )
+
+            if (result == null) {
+              return []
+            }
+
+            return foldAndApplyCommandsInner(
```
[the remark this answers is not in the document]
Rheeseyb: I think the day that function was `export`ed was the day it should have been renamed!

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
Rheeseyb: This is because the event handler is being reused for the `beforeunload` event, which is very cheeky. The string return value there instructs the browser to show a dialog with that text (https://developer.mozilla.org/en-US/docs/Web/API/BeforeUnloadEvent)

editor/src/components/editor/actions/actions.spec.browser2.tsx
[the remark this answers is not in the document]
Rheeseyb: > so I really don't think this is more expressive

I'd argue that in this case it's far less clear than a regular for loop

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -1447,55 +1464,25 @@ export var storyboard = (
...
-          `,
-          'await-first-dom-report',
-        )
+      describe('repeated paste', () => {
+        async function pasteNTimes(editor: EditorRenderResult, n: number) {
+          const canvasRoot = editor.renderedDOM.getByTestId('canvas-root')
 
-        await selectComponentsForTest(editor, [EP.fromString('sb/container')])
+          for (const _ in Array(n).fill(0)) {
+            // paste 4 times
```
Rheeseyb, line 1472: This comment is a lie now :)

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -1447,55 +1464,25 @@ export var storyboard = (
...
+            firePasteEvent(canvasRoot)
 
-        await expectNoAction(editor, () => pressKey('c', { modifiers: cmdModifier }))
-        await editor.getDispatchFollowUpActionsFinished()
+            // Wait for the next frame
+            await clipboardMock.pasteDone
+            await editor.getDispatchFollowUpActionsFinished()
+
+            await pressKey('l')
```
Rheeseyb, line 1479: `'l'`?

editor/src/components/editor/actions/actions.spec.browser2.tsx
bkrmendy: fixed in `[7442e2f](https://github.com/concrete-utopia/utopia/pull/3798/commits/7442e2f0fd888dff214981287d11f3de11de9d28)`

Rheeseyb: Hahaha ok technically it's not because you use 4 every time, but still!

editor/src/components/editor/actions/actions.spec.browser2.tsx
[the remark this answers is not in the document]
bkrmendy: fixed in [7442e2f](https://github.com/concrete-utopia/utopia/pull/3798/commits/7442e2f0fd888dff214981287d11f3de11de9d28)

editor/src/components/editor/actions/actions.spec.browser2.tsx
bkrmendy: I chose a key that wasn't tied to an existing shortcut, actually is indicates the problem that we agreed to fix before the FS is removed from this feature, so this should be changed to Esc or any other non-digit or tab key in the long run

editor/src/components/canvas/canvas-strategies/strategies/paste-metastrategy.tsx
bkrmendy: fixed in https://github.com/concrete-utopia/utopia/pull/3798/commits/3e21d9a526d1af2cc2d467f72aff0a38811d5366

editor/src/components/canvas/canvas-strategies/strategies/test-static-reparent-strategy.tsx
bkrmendy: fixed in https://github.com/concrete-utopia/utopia/pull/3798/commits/0db863999a814ca1e76a95bcd52206264c386256

editor/src/components/canvas/canvas-strategies/strategies/paste-metastrategy.tsx
bkrmendy: fixed in https://github.com/concrete-utopia/utopia/pull/3798/commits/0db863999a814ca1e76a95bcd52206264c386256

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
bkrmendy: I renamed the handler ref itself in https://github.com/concrete-utopia/utopia/pull/3798/commits/a4647e861bc28eeddd54640e51cea63347c4ca72, this is necessary because the same handler is subscribed to `beforeunload`, and the way to show the built-in `Changes that you made may not be saved` popup is to return a non-null or undefined value (https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#compatibility_notes)

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -955,38 +955,47 @@ export function useClearKeyboardInteraction(editorStoreRef: {
...
       }
+      return undefined
```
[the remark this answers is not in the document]
bkrmendy: this is purely here to make a linter warning go away (since a string might be returned, the function explicitly has to return something on all paths)

editor/src/components/canvas/canvas-strategies/strategies/test-static-reparent-strategy.tsx
bkrmendy: it is deleted :D

editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -62,6 +63,11 @@ export interface KeyboardInteractionData {
 
 export type StaticReparentInteractionData = {
```
balazsbajorics, line 64: shouldn't this be `PasteInteractionData` ?

editor/src/components/canvas/canvas-strategies/strategies/test-static-reparent-strategy.tsx
gbalint: 🤦

editor/src/components/canvas/canvas-strategies/strategies/paste-metastrategy.tsx
```diff
@@ -0,0 +1,254 @@
...
+        canvasState.startingMetadata,
+        canvasState.startingAllElementProps,
+        canvasState.startingElementPathTree,
+        target.parentPath.intendedParentPath,
+      )
+
+      const commands = elements.flatMap((currentValue) => {
+        return [
+          updateFunctionCommand('always', (editor, commandLifecycle) => {
+            if (interactionSession.interactionData.type !== 'STATIC_REPARENT') {
```
balazsbajorics, line 101: nit: if you put interactionSession.interactionData in a const and then type check that, TS won't bother you with later property accesses.

gbalint (approved):

editor/src/templates/editor.tsx
```diff
@@ -402,12 +402,16 @@ export class Editor {
...
+    const shouldCommitInteraction =
+      this.canvasStore.getState().editor.canvas.interactionSession?.interactionData.type ===
+      'STATIC_REPARENT'
+
     this.boundDispatch(
       [
         EditorActions.clearHighlightedViews(),
         CanvasActions.clearDragState(false),
-        CanvasActions.clearInteractionSession(false),
+        CanvasActions.clearInteractionSession(shouldCommitInteraction),
```
balazsbajorics, line 414: TBH I think this could just be always true. I wouldn't want my keyboard interaction to be reset once I'm done keyboarding and blur the editor!

balazsbajorics (approved):

gbalint (approved):

PR #3808, opened 2023-06-12
Feature/group resize

balazsbajorics:
**Problem:**
Groups are not resizable on master.

**Fix:**
This PR adds group resize abilities to the editor. As with the group size true-up PR before, I only made it work for the canvas interactions first. A follow up PR should make it work correctly with a group-aware inspector.

**Commit Details:**
- The PushIntendedBoundsAndUpdateGroups command now does two passes of size true-ups:
  - the first pass, called `getUpdateResizedGroupChildrenCommands` goes downwards in the tree, and changes the size of any children of resized groups.
  - the second pass, called `getResizeAncestorGroupsCommands` travels the tree up, and trues up any group ancestor that may have children (newly) sticking out from them.
- The Absolute Resize strategy has been changed so it doesn't enforce width/height props on groups if they did not have one before.
- `componentHonoursPropsSize` no longer returns false for groups
- So many new tests


bkrmendy (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize.test-utils.tsx
```diff
@@ -4,9 +4,14 @@ import { EdgePosition } from '../../canvas-types'
 import { mouseDragFromPointWithDelta } from '../../event-helpers.test-utils'
 import { EditorRenderResult } from '../../ui-jsx.test-utils'
 
+interface Point {
```
seanparsons, line 7: Surely we already have a type like this?

balazsbajorics: yes, but these are interfaces we don't want to expose to non-test code. in the editor we want to enforce the developer to pick CanvasPoint / WindowPoint etc. 

As far as I understand, it doesn't matter to Typescript (compile time or otherwise) if an interface is duplicated (but identical duck-typing wise) or if you are pointing to the same "reference" imported from the same declaration.

seanparsons: Thinking out loud: The parameter is a delta and we could just make a `Delta` type as that is scope-agnostic.

balazsbajorics: that's a good idea. I'm doing it!

editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
```diff
@@ -90,11 +111,110 @@ function pushCommandStatePatch(
...
+  const targets: Array<{
+    target: ElementPath
+    size: Size
+  }> = command.value.map((ft) => ({
+    target: ft.target,
+    size: sizeFromRectangle(ft.frame),
+  }))
+
+  // we are going to mutate this as we iterate over targets
+  let updatedLocalFrames: { [path: string]: LocalFrameAndTarget | undefined } = {}
```
seanparsons, line 133:
```suggestion
  let updatedLocalFrames: { [path: string]: LocalFrameAndTarget } = {}
```
Surely the `undefined` would mean you could create the slightly confusing case of a key present with the value `undefined`?

seanparsons (approved):

editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
balazsbajorics: maybe there is a better way I couldn't think of:

the original problem was that typescript said the type of  `updatedLocalFrames['garbageKey']` is `LocalFrameAndTarget` and did not enforce an existence check. 
It was combined with the other typescript behavior we observed a few weeks ago when it believes that the type is never null, and nullchecks stop being properly flow typed.

seanparsons: Yeah, I wish indexing was type checked, that's why I wrote `safeIndex` for arrays.

seanparsons: Could add a similar function, using the `fromField` optic feels a bit overkill.

PR #3811, opened 2023-06-13
Fix: ordering for dynamic children of js expressions

### LINKED ISSUE #3810: Fix ordering for dynamic children of js expressions

ruggi:
Fixes #3810 

**Problem:**

Multiple dynamic children coming from js expressions don't keep the right ordering in the tree and, subsenquenly, navigator.

**Fix:**
Add a reordering step in the `reorderTree` function so that dynamic children of JSX elements/fragments are correctly reordered according to their path/metadata.

Before:
Screenshot 2023-06-13 at 3 53 36 PM

After:
Screenshot 2023-06-13 at 3 54 01 PM


editor/src/core/shared/element-path-tree.ts
```diff
@@ -133,6 +134,69 @@ export function reorderTree(trees: ElementPathTrees, metadata: ElementInstanceMe
   }
 }
 
+function maybeReorderDynamicChildren(
+  children: ElementPathTree[],
+  metadata: ElementInstanceMetadataMap,
+): ElementPathTree[] {
+  function keepOnlyUnique(path: ElementPath, index: number, array: ElementPath[]) {
+    return array.indexOf(path) === index
```
seanparsons, line 142: Surely it needs to check more than the first index it finds?
```suggestion
    return array.indexOf(path) === index && array.lastIndexOf(path) === index
```

ruggi: 🤔  I don't think it needs to, it needs to trim uniques from the head of the array

seanparsons: Oh yes, my mistake, this will filter out subsequent instances that match earlier ones!

editor/src/core/shared/element-path-tree.ts
```diff
@@ -133,6 +134,69 @@ export function reorderTree(trees: ElementPathTrees, metadata: ElementInstanceMe
...
+  // E.g.:
+  //  +--------------+     +-----------+
+  //  | foo/bar      |     | foo/bar   |
+  //  | foo/baz~~~1  |  →  | foo/baz   |
+  //  | foo/baz~~~2  |     | foo/qux   |
+  //  | foo/baz~~~3  |     | foo/corge |
+  //  | foo/qux      |     +-----------+
+  //  | foo/corge~~1 |
+  //  +--------------+
+  const groupedPaths = children
```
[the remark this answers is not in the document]
ruggi: https://github.com/concrete-utopia/utopia/pull/3811/commits/49a38d1381382e799a3c9ffbb3bd6b2363cafcd9

editor/src/core/shared/element-path-tree.ts
```diff
@@ -133,6 +134,69 @@ export function reorderTree(trees: ElementPathTrees, metadata: ElementInstanceMe
...
+  // If there are no dynamic children, do nothing.
+  if (dynamicChildren.length === 0) {
+    return children
+  }
+
+  // Build a stack to reorder dynamic children
+  const dynamicChildrenStack = dynamicChildren
+    .sort((a, b) => {
+      return (
+        Object.keys(metadata).indexOf(a.pathString) - Object.keys(metadata).indexOf(b.pathString)
```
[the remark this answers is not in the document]
ruggi: https://github.com/concrete-utopia/utopia/pull/3811/commits/dcd1fd3bd25579e80460283e00ceac08fda7575d

editor/src/core/shared/element-path-tree.ts
```diff
@@ -133,6 +134,73 @@ export function reorderTree(trees: ElementPathTrees, metadata: ElementInstanceMe
   }
 }
 
+function maybeReorderDynamicChildren(
+  children: ElementPathTree[],
+  metadata: ElementInstanceMetadataMap,
+): ElementPathTree[] {
+  function keepOnlyUnique(path: ElementPath, index: number, array: ElementPath[]) {
+    return array.indexOf(path) === index
```
gbalint, line 142: array.indexOf uses reference equality, shouldn't this use `find` and `EP.pathsEqual`?

editor/src/core/shared/element-path-tree.ts
```diff
@@ -133,6 +134,73 @@ export function reorderTree(trees: ElementPathTrees, metadata: ElementInstanceMe
...
+      groupedPaths.push(child.path)
+    }
+  }
+
+  // Now calculate the complete full children array,
+  // expanding the grouped paths and replacing them with the
+  // list of sorted dynamic paths calculated earlier.
+  const expandedChildren = groupedPaths.flatMap((path): ElementPathTree[] => {
+    const matchingDynamicChildren = dynamicChildren.filter(
+      (other) => EP.dynamicPathToStaticPath(other.path) === path,
```
gbalint, line 193: I think this should also use `EP.pathsEqual` we should not rely on reference equality for element paths.

ruggi: https://github.com/concrete-utopia/utopia/pull/3811/commits/96eb805a1f7aa28c783f6ba30d1e50a05b4bf9cc

editor/src/core/shared/element-path-tree.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/3811/commits/96eb805a1f7aa28c783f6ba30d1e50a05b4bf9cc

editor/src/core/shared/element-path-tree.ts
```diff
@@ -133,6 +134,73 @@ export function reorderTree(trees: ElementPathTrees, metadata: ElementInstanceMe
   }
 }
 
+function maybeReorderDynamicChildren(
+  children: ElementPathTree[],
+  metadata: ElementInstanceMetadataMap,
+): ElementPathTree[] {
+  function keepOnlyUnique(path: ElementPath, index: number, array: ElementPath[]) {
+    return array.findIndex((e) => EP.pathsEqual(path, e)) === index
```
gbalint, line 142: Even better idea: you can just `Utils.uniqBy`

editor/src/core/shared/element-path-tree.ts
```diff
@@ -133,6 +134,73 @@ export function reorderTree(trees: ElementPathTrees, metadata: ElementInstanceMe
...
+    return children
+  }
+
+  const metadataKeys = Object.keys(metadata)
+
+  // Build a stack to reorder dynamic children
+  const dynamicChildrenStack = dynamicChildren
+    .sort((a, b) => {
+      return metadataKeys.indexOf(a.pathString) - metadataKeys.indexOf(b.pathString)
+    })
```
bkrmendy, line 162: What I don't like about this part is that this is essentially `O(metadataKeys.length * dynamicChildren.length)`, and ideally we'd like to avoid depending on project size for performance. This can be easily fixed by calculating a reverse lookup for indices up front and using it here:
```typescript
const metadataKeysToIndices = Object.keys(metadata).reduce(
    (acc: { [key: string]: number }, val: string, idx: number) => {
      acc[val] = idx
      return acc
    },
    {},
  )
  ```
  
```typescript
  // Build a stack to reorder dynamic children
  const dynamicChildrenStack = dynamicChildren
    .sort((a, b) => {
      return metadataKeys[a.pathString] - metadataKeys[b.pathString]
    })
    .map((c) => EP.dynamicPathToStaticPath(c.path))
    .filter(keepOnlyUnique)
 ```

File: editor/src/core/shared/element-path-tree.ts [hunk 8]
```diff
@@ -133,6 +134,73 @@ export function reorderTree(trees: ElementPathTrees, metadata: ElementInstanceMe
...
+function maybeReorderDynamicChildren(
+  children: ElementPathTree[],
+  metadata: ElementInstanceMetadataMap,
+): ElementPathTree[] {
+  function keepOnlyUnique(path: ElementPath, index: number, array: ElementPath[]) {
+    return array.findIndex((e) => EP.pathsEqual(path, e)) === index
+  }
+
+  // Get all the dynamic children and sort them
+  // alphabetically, since their paths will have incremental indexes (~~~N).
```
balazsbajorics, line 146: won't this fail if there's more than 10 dynamic children? ie if we are comparing `aaa~~~2` with `aaa~~~12`?

Rheeseyb: I just quickly checked and it looks like that's handled by `localeCompare`:

```
> '~~~12'.localeCompare('~~~1')
< 1
> '~~~2'.localeCompare('~~~1')
< 1
```

editor/src/core/shared/element-path-tree.ts
Rheeseyb: We really need to get rid of `Utils`. We separated all of the important parts into separate utils files, so the function you're talking about is actually in `array-utils.ts`: https://github.com/concrete-utopia/utopia/blob/c64919e1e2de7856b801dd8ba083ca48acd7d6d4/editor/src/core/shared/array-utils.ts#L149

bkrmendy (approved):

editor/src/core/shared/element-path-tree.ts
```diff
@@ -25,57 +31,160 @@ export function elementPathTree(
...
+  for (const path of paths) {
+    let parent = EP.parentPath(path)
+    while (parent.parts.length > 0) {
+      if (metadata[EP.toString(parent)] == null) {
+        missingParentPaths.add(parent)
+      }
+      parent = EP.parentPath(parent)
+    }
+  }
+  return [...missingParentPaths.values()]
```
seanparsons, line 137: I _think_ you can do
```suggestion
  return Array.from(missingParentPaths)
```
to avoid an intermediary value.

seanparsons (approved):

editor/src/core/shared/element-path-tree.ts
```diff
@@ -25,57 +31,160 @@ export function elementPathTree(
...
+  const elementPaths = Object.values(metadata).map((m) => m.elementPath)
+  if (
+    elementPaths.length === 0 ||
+    elementPaths[0].parts.length < 1 ||
+    elementPaths[0].parts[0].length < 1
+  ) {
+    return {}
+  }
+
+  const root = EP.fromString(elementPaths[0].parts[0][0])
```
Rheeseyb, line 46: This makes me very uneasy because it's leaking the internal abstraction of the element path very badly. I think simply moving this into `element-path.ts` as a function that was something like `getStoryboardPathFromPath(elementPath: ElementPath): ElementPath` then that would be enough.

editor/src/core/shared/element-path-tree.ts
```diff
@@ -25,57 +31,160 @@ export function elementPathTree(
...
-function emptyElementPathTree(path: ElementPath, pathString: string): ElementPathTree {
-  return elementPathTree(path, pathString, [])
+export type ElementPathTrees = { [key: string]: ElementPathTree }
+
+export function buildTree(metadata: ElementInstanceMetadataMap): ElementPathTrees {
+  const elementPaths = Object.values(metadata).map((m) => m.elementPath)
+  if (
+    elementPaths.length === 0 ||
+    elementPaths[0].parts.length < 1 ||
+    elementPaths[0].parts[0].length < 1
```
Rheeseyb, line 41: `isEmptyPath` should do the trick here https://github.com/concrete-utopia/utopia/blob/4130f42930a17f44249d00f2a1b93d83dbdb2f00/editor/src/core/shared/element-path.ts#L218

editor/src/core/shared/element-path-tree.ts
```diff
@@ -25,57 +31,160 @@ export function elementPathTree(
...
+    (path) => findMaybeConditionalExpression(path, metadata) != null,
+  )
+  let paths = original.filter(
+    (path) =>
+      !elementsToReorderStack.some((other) => EP.pathsEqual(other, path)) && // omit conditionals, that will be reordered
+      !missingParents.some((parentPath) => EP.isDescendantOf(path, parentPath)), // omit elements that have a missing parent
+  )
+
+  while (elementsToReorderStack.length > 0) {
+    const conditional = elementsToReorderStack.shift()
```
Rheeseyb, line 155: Can this be rewritten as an array reduce please to keep the amount of mutation to a minimum.

editor/src/core/shared/element-path-tree.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/3811/commits/9d3d3de881172c81bf5cc0b3cae596f3171ac393

editor/src/core/shared/element-path-tree.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/3811/commits/5337e065eee25e8558f5c59be1df873d9a751333

editor/src/core/shared/element-path-tree.ts
```diff
@@ -25,57 +31,160 @@ export function elementPathTree(
...
+  let tree: ElementPathTrees = {}
+  buildTreeRecursive(root, tree, paths, metadata)
+
+  return tree
 }
 
-export type ElementPathTrees = { [key: string]: ElementPathTree }
+function buildTreeRecursive(
+  rootPath: ElementPath,
+  trees: ElementPathTrees,
```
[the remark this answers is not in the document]
ruggi: makes sense!

Rheeseyb (approved):

editor/src/core/shared/element-path-tree.ts
ruggi: https://github.com/concrete-utopia/utopia/pull/3811/commits/2ca47625827f7782014f6519a357661a50862c40

PR #3812, opened 2023-06-13
Don't invalidate SVG root elements triggering the DOM walker.

seanparsons:
**Problem:**
It was observed that any and every action were triggering the DOM walker, even some that make no relevant changes like the mouseover handler for highlighting.

**Cause:**
An entry for a `svg` element at the root of a component ended up in `invalidatedPaths` but the DOM walker ignores elements which aren't instances of `HTMLElement`. Instead as they are `SVGElement` instances, the entry in `invalidatedPaths` never gets cleared by the DOM walker.

**Fix:**
At the point in which the component is rendered and the root element path is added to `invalidatedPaths` an extra check was inserted which ensures that no `svg` elements are added.

**Commit Details:**
- Added a check in `createComponentRendererComponent` which ensures that `svg` component root elements are not added to the invalidated paths.
- Added `getDomWalkerState` to `EditorRenderResult`.
- Added `isSVGElement` utility function.

### DIFF
--- a/editor/src/components/canvas/dom-walker.spec.browser2.tsx
+++ b/editor/src/components/canvas/dom-walker.spec.browser2.tsx
@@ -18,7 +18,11 @@ import {
   zeroRectIfNullOrInfinity,
 } from '../../core/shared/math-utils'
 import { MapLike } from 'typescript'
-import { duplicateSelected, setProp_UNSAFE } from '../editor/actions/action-creators'
+import {
+  clearSelection,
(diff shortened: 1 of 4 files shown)


bkrmendy (approved):

Rheeseyb (approved):

balazsbajorics (approved):
we should probably mark it down somewhere that a follow up task should be to make the dom walker be able to do process SVG roots

PR #3814, opened 2023-06-14
Fix/paste insert elements

enidemi:
**Problem:**
There are multiple bugs when pasting more than one element into conditionals. The classic paste action just accidentally skips adjusting the layout because it manually wraps into fragment first, and it's easy to end up with nested fragments when there is a fragment originally in the conditional clause. The new "paste to replace" is in a bit worse shape because it misses the manual wrapping too, but it would have the same problems.

**Fix:**
First step of the fixes is to introduce `insertJSXElementChildren` which accepts an array of child elements. With that it's easier to figure out if a wrapper fragment is required.
It is not used anywhere now, because #3798 is changing the action to a strategy. I temporarly switched it in `insertElementAtPath` to try it locally and run the tests.

**Commit Details:**
- made a new version of `insertJSXElementChild` called `insertJSXElementChildren`
- added tests, based on the original one

### DIFF
--- a/editor/src/core/model/element-template-utils.ts
+++ b/editor/src/core/model/element-template-utils.ts
@@ -37,6 +37,7 @@ import {
   JSXConditionalExpression,
   isJSExpression,
   ArbitraryJSBlock,
+  JSXFragment,
 } from '../shared/element-template'
 import {
   isParseSuccess,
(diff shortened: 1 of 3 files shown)


editor/src/core/model/element-template-utils.ts
```diff
@@ -543,6 +544,92 @@ export function insertJSXElementChild(
...
+      if (!isJSXElementLike(parentElement)) {
+        throw new Error("Target parent for child element insertion doesn't support children")
+      }
+      let updatedChildren: Array<JSXElementChild>
+      if (indexPosition == null) {
+        updatedChildren = [...parentElement.children, ...elementsToInsert]
+      } else {
+        updatedChildren = elementsToInsert.reduce((working, elementToInsert) => {
+          return addToArrayAtIndexPosition(elementToInsert, working, indexPosition)
+        }, parentElement.children)
```
bkrmendy, line 567: I'm not sure this will handle all kinds of IndexPositions well

editor/src/core/model/element-template-utils.ts
```diff
@@ -543,6 +544,92 @@ export function insertJSXElementChild(
...
+    } else if (isConditionalClauseInsertionPath(targetParent)) {
+      if (!isJSXConditionalExpression(parentElement)) {
+        throw new Error('Target parent for conditional insertion is not conditional expression')
+      }
+      // Determine which clause of the conditional we want to modify.
+      const conditionalCase = targetParent.clause
+      const toClauseOptic =
+        conditionalCase === 'true-case' ? conditionalWhenTrueOptic : conditionalWhenFalseOptic
+
+      function wrapIntoFragment(clauseValue: JSXElementChild | null): JSXFragment {
```
bkrmendy, line 582: I think the null could be removed here if this function was passed the new children array (so clauseValue + the array to be inserted if there's an existing clause value, or just the array to be inserted if the clause value is null)

editor/src/core/model/element-template-utils.ts
enidemi: fixed

bkrmendy (approved):

Rheeseyb:
> It is not used anywhere now, because https://github.com/concrete-utopia/utopia/pull/3798 is changing the action to a strategy.

We've agreed that the strategy should remain behind a feature switch, so it's worth actually using this in the action (and then using it in the strategy too once #3798 has been merged)

Rheeseyb (approved):

editor/src/core/model/element-template-utils.ts
```diff
@@ -543,6 +549,94 @@ export function insertJSXElementChild(
...
+          }
+          return wrapIntoFragment(clauseValue)
+        },
+        parentElement,
+      )
+    } else {
+      assertNever(targetParent)
+    }
+  })
+  return insertChildAndDetails(updatedComponents, null, importsToAdd) // TODO is this wrapper type needed?
```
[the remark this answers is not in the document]
enidemi: I actually plan to change this to return the freshly added element paths (with the new fragment)

PR #3816, opened 2023-06-14
Committing the paste session on keydown

bkrmendy:
## Problem
Whan a paste session is in progress, and one wants to end it by pressing a key, if the key is bound to a shortcut, the shortcut is executed (such as Esc jumping to the parent, or Enter entering text edit mode)

## Fix
Call preventDefault and stopPropagation on the keydown event.

### Trivia
It turns out that the `removeEventListener` needs to be passed the same options object that the respective `addEventListener` received. The following example illustrated "correctly" removing an event listener registered with `capture: true`:

```typescript
window.addEventListener('mousedown', this.onMouseDown, {
  once: true,
  capture: true,
})

// ...
 
window.removeEventListener('mousedown', this.everythingElse, { capture: true })
```

Docs: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

Also,I refactored `useClearDiscreteReparentInteraction` to use an old-school class, so that the memoization is handled in a centralised place, and we don't have to juggle refs throughout the whole function.

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
@@ -47,7 +47,7 @@ import {
 } from '../../canvas-strategies/interaction-state'
 import { Modifier } from '../../../../utils/modifiers'
 import { pathsEqual } from '../../../../core/shared/element-path'
-import { EditorAction } from '../../../../components/editor/action-types'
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):

enidemi (approved):

Rheeseyb (commented):
👍

PR #3822, opened 2023-06-15
Setting size in the inspector handles Groups

balazsbajorics:
**Problem:**
Setting the size of a Group from the inspector would just change the size prop of a group without stretching-resizing its children.
Setting the size of the child of a group would just change the size of the child without updating the group to continue hugging the children.

**Fix:**
The size change strategy used by the inspector now uses the `pushIntendedBoundsAndUpdateGroups` command, allowing the automatic group true-ups to run.

Furthermore, the inspector only allows setting pixel sizes on both the group and children of group. This is done for two reasons: 
- setting % on the group size means it will have a runtime flexible size, but the group cannot resize its children during runtime, leading to inconsistency.
- setting % on a child of a group creates a problem because the group is sized based on the children, but the child is based on the group

Minor note: setting non-% non-px value such as 1em should be possible, but I didn't want to include this in this PR because it requires a better understanding of the pixel value of those units for the selected element, which involves possible dom-walker changes.
 

**Commit Details:**
- Added new `predictElementSize` function which takes the original globalFrame of the element and the new dimension value (px or %) that the user is setting in the inspector and returns a new predicted globalFrame
- this predicted frame is then used to feed `pushIntendedBoundsAndUpdateGroups`
- `FillHugFixedControl` explicitly handles groups and children of groups to only allow 'px' input form users
- Nice test coverage


bkrmendy (approved):

enidemi (approved):

gbalint (approved):

editor/src/components/inspector/inspector-common.ts
```diff
@@ -965,3 +972,37 @@ export function isFixedHugFillEqual(
...
+  newValue: CSSNumber,
+): CanvasRectangle {
+  const elementMetadata = forceNotNull(
+    `found no metadata for element at path ${EP.toString(path)}`,
+    MetadataUtils.findElementByElementPath(metadata, path),
+  )
+  const boundingParentSizeForPercentCalc = forceNotNull(
+    'coordinateSystemBounds was null in metadata',
+    elementMetadata.specialSizeMeasurements.coordinateSystemBounds,
+  )
```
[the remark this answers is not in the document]
balazsbajorics: maybe we should have a sibling to force not null called invariant. the `coordinateSystemBounds` should never be null. it's actually something we should fix in the type of specialSizeMeasurements. it has a bunch of nullable things that are actually only null for stub / spy metadata. but the _final_ SpecialSizeMeasurements should have a non-nullable coordinateSystemBounds.

Regarding 
> it can silently fail the test suite
we should fix the test suite if this can still happen. I know easier said than done.

> Can't this be just a simple return null or an Either with the reason the function failed?
if I turn this into a null or Either with a failure reason, then I just moved up the burden of throwing the error to the caller. 
1. But for the user it doesn't really matter who throws the error
2. the caller can then opt to make this a _silent_ error, without even an error message. that's not only -worse- for the user, but super painful to debug.

I guess you meant that this should be a proper user facing error message, like showing a toast.. I think in general there's something to that for actions that can sometimes fail for non-bug reasons. _But_ in this specific case (invariant violation) I think the right approach is throwing error (and making sure the tests fail on uncaught top level exceptions!!)

editor/src/components/inspector/inspector-common.ts
```diff
@@ -965,3 +972,37 @@ export function isFixedHugFillEqual(
...
+    return currentBounds
+  }
+
+  const newSizePx =
+    newValue.unit === '%'
+      ? newValue.value * 0.01 * boundingParentSizeForPercentCalc[changedProp]
+      : newValue.value
+
+  const newBounds = canvasRectangle({
+    ...(currentBounds as SimpleRectangle), // this conversion to SimpleRectangle needed for TS to allow the spread
```
seanparsons, line 1004: Doesn't this mean that if `currentBounds` has an incompatible type that this will just blindly cast it?

seanparsons (approved):

editor/src/components/inspector/inspector-common.ts
seanparsons: Ignore me, this doesn't appear to be the case.

balazsbajorics: AFAIK to cast to an incompatible type, you have to do `as any as SomeIncompatibleType`

Rheeseyb (approved):

PR #3825, opened 2023-06-16
Selecting elements on overflown text content

gbalint:
**Problem:**
Make elements selectable by clicking on their overflown text content.
(even if the element is sizeless)

**Fix:**
I created a new `specialSizeMeasurements` field called `globalFrameWithTextContent`. This contains the bounding box of the element with all the text inside (or inside one of its descendants).

In dom-lookup earlier I implemented a new algorithm to find the element under a certain point ( https://github.com/concrete-utopia/utopia/pull/3768 ). The goal was to be able to select a leaf element which is not in the dom, so not returned by `document.elementsUnderPoint`.
However, that algorithm can be reused for clicking on overflown text, because (even though it is in the dom) the problem is really similar: the element is not returned by `document.elementsUnderPoint`. But it does not work directly, because the algorithm check if the point is in the globalFrame of the element, and that is not true for a point over overflown content. So I also allow elements for which only their `globalFrameWithTextContent` is under the mouse.


editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
```diff
@@ -1528,6 +1528,37 @@ describe('mouseup selection', () => {
...
   })
+  describe('overflown text', () => {
+    it('Clicking on overflown text selects the parent element', async () => {
+      // prettier-ignore
+      const desiredPath = EP.fromString(
+      'sb' +      // Skipped as it's the storyboard
+      '/sc' +     // Skipped because we skip over Scenes
+      '/pg' +     // Skipped because we skip component children of Scenes
+      ':pg-root' + // Skipped because we skip root element of root component of Scenes
+      '/pg-div'  // Single click
```
bkrmendy, line 1539: 🧑‍🍳👌

bkrmendy (approved):

ruggi (approved):

PR #3829, opened 2023-06-19
disallow pasting a component into its own definition

bkrmendy:
## Problem

Cosnider the following project:
```tsx
import { Storyboard, Scene } from 'utopia-api'

const App = (props) => (
  <div style={{ ...props.style }} data-uid='app-root'>
    <ThisComponent data-uid='component-1' />
    <ThisComponent data-uid='component-2' />
  </div>
)

const ThisComponent = (props) => (
  <div style={{ ...props.style }} data-uid='custom-root'>
    <div data-uid='hello-1' data-testid='target'>
      Hello there 1!
    </div>
    <div data-uid='aap'>Hello there 2!</div>
    <div data-uid='aat'>Hello there 3!</div>
  </div>
)

export var storyboard = (
  <Storyboard data-uid='0cd'>
    <Scene
      data-uid='b58'
      style={{ width: 432, height: 356, left: 98, top: 36 }}
    >
      <App
        style={{
          backgroundColor: '#aaaaaa33',
          position: 'absolute',
          left: 34,
          top: 31,
          width: 346,
          height: 223,
        }}
        data-uid='app'
      />
    </Scene>
  </Storyboard>
)
```

If I copy `component-1` and paste it inside `ThisComponent`, next to `hello-1`, `ThisComponent` would infinitely re-render itself, breaking the canvas. The same thing can happen if in the above project `app` is pasted into `custom-root`, since that way `App` and `ThisComponent` would be mutually recursive.

## Fix
Even though recursively rendering components can have legit use cases (such as a tree structure visualizer), for our tool, at this time, this is too advanced a use case (since it's non-trivial to decide whether a recursive function will stop recursing if called). In order to avoid crashing the canvas, reparenting a component instance into its own definition is disallowed in the navigator and via cut/copy/paste.

If we disallow pasting to prevent recursive rendering, we show a toast that tries to explain the situation.

Creating a recursive structure is still possible via the code editor.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-helpers.ts
@@ -240,7 +240,7 @@ export function getInsertionPathForReparentTarget(
   return conditionalClauseInsertionPath(newParent, clause, 'replace')
 }
 
-function instancesOfSameComponent(
+function areElementsInstancesOfTheSameComponent(
   firstInstance: ElementInstanceMetadata | null,
   secondInstance: ElementInstanceMetadata | null,
 ): boolean {
@@ -258,7 +258,7 @@ function instancesOfSameComponent(
   return jsxElementNameEquals(firstInstance.element.value.name, secondInstance.element.value.name)
 }
 
-export function instanceInSubtreeOfSameComponent(
+export function isElementRenderedBySameComponent(
   metadata: ElementInstanceMetadataMap,
   targetPath: ElementPath,
   instance: ElementInstanceMetadata | null,
@@ -273,7 +273,7 @@ export function instanceInSubtreeOfSameComponent(
   )
 
   return (
-    instancesOfSameComponent(currentInstance, instance) ||
-    instanceInSubtreeOfSameComponent(metadata, EP.renderedByParentPath(targetPath), instance)
+    areElementsInstancesOfTheSameComponent(currentInstance, instance) ||
+    isElementRenderedBySameComponent(metadata, EP.renderedByParentPath(targetPath), instance)
   )
 }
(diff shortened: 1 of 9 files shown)


editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-helpers.ts
```diff
@@ -238,3 +239,41 @@ export function getInsertionPathForReparentTarget(
...
+  firstInstance: ElementInstanceMetadata | null,
+  secondInstance: ElementInstanceMetadata | null,
+): boolean {
+  if (
+    firstInstance == null ||
+    secondInstance == null ||
+    isLeft(firstInstance.element) ||
+    isLeft(secondInstance.element) ||
+    firstInstance.element.value.type !== 'JSX_ELEMENT' ||
+    secondInstance.element.value.type !== 'JSX_ELEMENT'
```
ruggi, line 253:
```suggestion
    !isJSXElement(firstInstance.element.value) ||
    !isJSXElement(secondInstance.element.value)
```

bkrmendy: added it! [688b0dc](https://github.com/concrete-utopia/utopia/pull/3829/commits/688b0dcb43afe66e17d613faa358cadb0b594b13)

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-helpers.ts
```diff
@@ -238,3 +239,41 @@ export function getInsertionPathForReparentTarget(
...
+  }
+
+  const currentInstance = MetadataUtils.findElementByElementPath(
+    metadata,
+    EP.renderedByParentPath(targetPath),
+  )
+
+  return (
+    areElementsInstancesOfTheSameComponent(currentInstance, instance) ||
+    isElementRenderedBySameComponent(metadata, EP.renderedByParentPath(targetPath), instance)
```
enidemi, line 277: I think you can use `EP.getContainingComponent` for the closest component path

enidemi (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-helpers.ts
bkrmendy: added in https://github.com/concrete-utopia/utopia/pull/3829/commits/9786303b57ec5eeb33629cf81980a3621313c023

PR #3831, opened 2023-06-20
Post-action menu for paste

bkrmendy:
# Try it [here](https://utopia.pizza/p/f5909611-before-i-go-lite-(forked)/?branch_name=spike-post-action-menu)

Depends on https://github.com/concrete-utopia/utopia/pull/3861

## Post-action menu for paste

We need a mechanism to tweak an action after it has been dispatched, the same way drag interactions can be tweaked while in progress. We want this post-action window of opportunity for tweaking to end seamlessly, such as when selection is changed or a new non-transient action is dispatched.

Also, we want the new mechanism to be "transparent" to the rest of the editor, allowing strategy controls to render, hover controls to render, keyboard shortcuts to work, etc.

This PR presents one possible implementation of this mechanism with the following architecture:

### UX
When the post-action menu is turned on (with the `Paste strategies` FS), pasting should work like the pre-strategies version, i.e., once an element is pasted, all strategy controls should be visible and immediately usable, repeated paste should work as expected, keyboard shortcuts (such as deleting via backspace) should work immediately as well. The post-action menu looks precisely like the strategy selector with a blue border added (the look and feel of the menu can be tweaked down the line), and alternate post-action choices can be chosen with the digit keys and tab (directly clicking on them could be supported, but was omitted to cut the scope of the spike).

### Architecture
- A new `EditorState` property, called `postActionInteractionSession`, is added. This property stores a reference to the history and the unpatched editor state at the time of dispatching the command, and a property that is specific to the type of post-action interaction (currently only `PastePostActionMenuData`). This is basically the same as `InteractionSession`, `postActionInteractionSession` stores data that subsequent post-action menu choices and the post-action menu can read.
- A post-action session can be started by dispatching `startPostActionSession`, which takes a `PostActionMenuData` parameter as payload. This action creates a new `PostActionMenuSession`, based on the current editor state and the `PostActionMenuData` passed with the action. `startPostActionSession` is idempotent in the sense that it always creates a new `PostActionMenuSession` with the current editor state, so for example, repeated paste doesn't need special handling.
- The post-action session is reset to `null` either by explicitly dispatching `clearPostActionData`, or by dispatching a non-transient, non-post action session-related command (as implemented by `updatePostActionState`). This way, the post-action session ends "automatically".
- A post-action choice (the equivalent of canvas strategies) can be run with `executeCommandsWithPostActionMenu`. `executeCommandsWithPostActionMenu` takes the editor state and history snapshots from the post-action session, applies the commands generated by the post-action choice to the editor state snapshot, and returns an updated `EditorStoreUnpatched`, where the history is swapped out to the history snapshot coming from the post-action session. This way, no explicit undo action is needed to undo the effect of any previous post-action choices.


editor/src/components/editor/store/dispatch-strategies.tsx
```diff
@@ -794,3 +799,33 @@ function patchCustomStrategyState(
     ...patch,
   }
 }
+
+export function updatePostActionState(
+  editorState: EditorState,
+  actions: readonly EditorAction[],
+): EditorState {
+  const nonTransientActions = actions.filter((a) => !isTransientAction(a))
+  const ExecutePostActionMenuChoiceAction = actions.find(
```
seanparsons, line 808: Tiny thing:
```suggestion
  const executePostActionMenuChoiceAction = actions.find(
```

bkrmendy: added it, thanks!

editor/src/components/editor/store/dispatch-strategies.tsx
```diff
@@ -794,3 +799,33 @@ function patchCustomStrategyState(
...
+    (a) => a.action === 'EXECUTE_POST_ACTION_MENU_CHOICE',
+  ) as ExecutePostActionMenuChoice | null
+
+  const startPostActionSessionAction = actions.find(
+    (a) => a.action === 'START_POST_ACTION_SESSION',
```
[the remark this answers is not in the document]
bkrmendy: added it!

editor/src/templates/editor.tsx
```diff
@@ -249,6 +249,7 @@ export class Editor {
       unpatchedDerived: derivedState,
       patchedDerived: derivedState,
       strategyState: strategyState,
+      postActionInteractionSession: null,
```
enidemi, line 252: the name feels too similar to the canvas interaction session but this one works very differently, I think we should name it something like `postActionMenuData` what do you think ?

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -4054,97 +4067,22 @@ export var storyboard = (
...
+          'regular-utopia-storyboard-uid/scene-aaa/app-entity:aaa',
+          'regular-utopia-storyboard-uid/scene-aaa/app-entity:aaa/bbb',
+          'regular-utopia-storyboard-uid/scene-aaa/app-entity:aaa/bbb/ccc',
+          'regular-utopia-storyboard-uid/scene-aaa/app-entity:aaa/bbb/ddd',
+          'regular-utopia-storyboard-uid/scene-aaa/app-entity:aaa/aaf', // <- the pasted element
+          'regular-utopia-storyboard-uid/scene-aaa/app-entity:aaa/aaf/aab',
+          'regular-utopia-storyboard-uid/scene-aaa/app-entity:aaa/aaf/aad',
+          'regular-utopia-storyboard-uid/scene-aaa/app-entity:aaa/aaf/aal',
+          'regular-utopia-storyboard-uid/scene-aaa/app-entity:aaa/aaf/aal/aah',
+          'regular-utopia-storyboard-uid/scene-aaa/app-entity:aaa/aaf/aal/aaj',
```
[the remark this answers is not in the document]
bkrmendy: Great catch, the problem was that repeated paste is whitelisted only to work either when pasting into a flex layout or when both the element and its would-be sibling are `position: absolute` (we can improve on this in the future). I fixed the project (https://github.com/concrete-utopia/utopia/pull/3831/commits/561d1c95b4462cff9a7aa514cf42780042fe4f25) in the test so that all elements that affect the outcome of paste are `position: absolute`.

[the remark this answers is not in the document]
bkrmendy: Great catch, the problem was that repeated paste is whitelisted only to work either when pasting into a flex layout or when both the element and its would-be sibling are `position: absolute` (we can improve on this in the future). I fixed the project (https://github.com/concrete-utopia/utopia/pull/3831/commits/561d1c95b4462cff9a7aa514cf42780042fe4f25) in the test so that all elements that affect the outcome of paste are `position: absolute`.

editor/src/templates/editor.tsx
Rheeseyb: Maybe `postActionMenuSession` would be better? I think it's the `Interaction` part that is a bit confusing

Rheeseyb (approved):
🥳

enidemi (approved):

PR #3836, opened 2023-06-20
Fix/paste to replace conditional fragments

enidemi:
**Problem:**
Paste to replace into a conditional clause with multiple elements on clipboard inserts only one element. With the current paste code when a fragment is inserted to the conditional clause the layout properties are not updated because with the fragment the target element paths are changing after the reparent commands are applied.

**Fix:**
Using the new `insertJSXElementChildren` helper in the Paste to Replace action. I added a new helper `insertWithReparentStrategiesMultiSelect` that I intend to use in normal paste and navigator reparent too in the next PR, and it should completely replace `insertWithReparentStrategies`.
It reparents the elements first, then with the results that may contain new fragments too the layout properties are updated.

**Commit Details:**
- new helper `insertWithReparentStrategiesMultiSelect`
- new command to add an array of elements
- `insertJSXElementChildren` is extended to return the freshly added element paths
- new tests for paste to replace with conditionals


editor/src/components/editor/store/editor-state.ts
```diff
@@ -1941,6 +1942,21 @@ export function insertElementAtPath(
     indexPosition,
   )
 }
+export function insertElementsAtPath(
```
seanparsons, line 1945: Isn't this redundant? I only checked against my local code, but the type definitions of the two functions match.

seanparsons (approved):

editor/src/components/editor/store/editor-state.ts
bkrmendy: I think master only has a singular `insertElementAtPath`, which accepts a single `JSXElementChild`, while this version expects an array of `JSXElementChild`(ren) (but maybe I'm misunderstanding the problem)

bkrmendy (approved):

editor/src/components/editor/store/editor-state.ts
enidemi: that's correct, this one receives an array of elements

seanparsons: I meant in relation to the function it calls, `insertElementsAtPath` and `insertJSXElementChildren` have identical type signatures.

enidemi: okay I see the redundant problem now, just fixed it :)

PR #3837, opened 2023-06-20
Restore selected views on undo

bkrmendy:
## Problem
Undo does not restore selected views, which can be percieved as a bug

## Fix
Restore `selectedViews` and `focusedElementPath` from the popped editor


gbalint:
I would add a test somewhere (maybe history.spec.ts ?)

enidemi (approved):

ruggi (approved):

Rheeseyb (commented):
👍

PR #3838, opened 2023-06-20
Remove unnecessary UID changing during the parse.

seanparsons:
**Problem:**
It was observed that within `parseCode` the call to `fixParseSuccessUIDs` can effectively duplicate the work also done by `guaranteeUniqueUIDsFromTopLevel`. In some cases this could even cause UIDs on an entity to be fixed twice.

**Cause:**
The division of responsibility between `guaranteeUniqueUIDsFromTopLevel` and `fixParseSuccessUIDs` was ill defined and not necessarily distinct:
- `guaranteeUniqueUIDsFromTopLevel`: Would attempt to ensure that there was no duplication of UIDs from the previously parsed files and within the same parse.
- `fixParseSuccessUIDs`: Attempts to re-use the UIDs from the previous parse success, but also has to ensure this doesn't cause duplication should a UID have been previously generated for something else.

**Fix:**
Having `fixParseSuccessUIDs` be more comprehensive in its handling of duplicate UIDs and then completely remove `guaranteeUniqueUIDsFromTopLevel` from the parser was the starting point.

This necessitated a few more changes:
- Have `updateUID` be configurable for two cases: Update UIDs based on the mappings, which is used for the combined arbitrary block as we don't want to generate new UIDs for the elements in there and a forced update for the UIDs of `data-uid` property values.
- Generating new UIDs excludes those generated during the parse, as we may not have reached those values yet, so they will not be in the already existing UIDs set.

**Commit Details:**
- `generateConsistentUID` now supports multiple sets of IDs to avoid.
- Deleted `guaranteeUniqueUIDsFromTopLevel` as it's now completely unused.
- `parseCode` now does not invoke `guaranteeUniqueUIDsFromTopLevel` anymore.
- In `parseCode`, after invoking `fixParseSuccessUIDs` any newly created UIDs are added to `alreadyExistingUIDs_MUTABLE`.
- `FixUIDsState` now includes `uidsExpectedToBeSeen`, which are the UIDs created in the process of parsing the current file.
- `FixUIDsState` now includes `uidUpdateMethod` which determines how `updateUID` functions.
- `fix*UIDs` functions now always operate on values to ensure deduplication of UIDs.
- `updateUID` now has some configurable behaviour via `uidUpdateMethod`.
- `fixCombinedArbitraryJSBlockUIDs` now updates the UIDs by purely operating on the mappings created so far, as the values within should end up with same UIDs as the entries they duplicate.


editor/src/core/shared/uid-utils.ts
```diff
@@ -106,17 +106,25 @@ const atoz = [
...
+  ...existingIDSets: Array<Set<string>>
 ): string {
+  function alreadyExistingID(idToCheck: string): boolean {
+    for (const existingIDs of existingIDSets) {
+      if (existingIDs.has(idToCheck)) {
+        return true
+      }
+    }
+    return false
+  }
```
Rheeseyb, line 119: You could simplify this to:
```suggestion
  function alreadyExistingID(idToCheck: string): boolean {
    return existingIDSets.some(s => s.has(idToCheck))
  }
```
I don't think this is a pattern we've used before, otherwise I'd have suggested adding a generic function to `set-utils.ts`

seanparsons: That's a lot better, I wrote it longer hand at the time because I was trying to think through what I actually wanted, but that's not necessary now.

Rheeseyb (approved):

bkrmendy (approved):

ruggi (approved):

PR #3839, opened 2023-06-20
Only auto-focus single children of a scene

Rheeseyb:
**Problem:**
Scenes are only meant to automatically focus a child component when it is the only child of the scene, but we are instead auto focusing all children of scenes.

**Fix:**
Include a check that the child of a scene is the only child inside the logic for determining the `validPaths` (which is in turn used to determine which elements are selectable and should have metadata recorded).

I've added some tests to ensure that we're not incorrectly auto-focusing multiple children of a scene or children of the storyboard by checking which element is selected when holding `cmd` whilst clicking.

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -3038,6 +3038,9 @@ export function getValidElementPathsFromElement(
     // }
     let paths: Array<ElementPath> = []
     fastForEach(Object.values(element.elementsWithin), (e) =>
+      // We explicitly prevent auto-focusing generated elements here, because to support it would
+      // require using the elementPathTree to determine how many children of a scene were actually
+      // generated, creating a chicken and egg situation.
       paths.push(
         ...getValidElementPathsFromElement(
           focusedElementPath,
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -2967,13 +2967,14 @@ export function getValidElementPathsFromElement(
...
   uiFilePath: string,
-  parentIsScene: boolean,
+  isOnlyChildOfScene: boolean,
   parentIsInstance: boolean,
   transientFilesState: TransientFilesState | null,
   resolve: (importOrigin: string, toImport: string) => Either<string, string>,
 ): Array<ElementPath> {
   if (isJSXElementLike(element)) {
     const isScene = isSceneElement(element, filePath, projectContents)
+    const isSceneWithOneChild = isScene && element.children.length === 1
```
gbalint, line 2977: This just means that the element has a single static child - it can still have multiple dynamic children if e.g. that child is a map expression returning elements.
And I think regarding selection the number of rendered children what matters.

bkrmendy (approved):

gbalint (approved):

PR #3843, opened 2023-06-21
Generate new uid for each property set by the inspector

Rheeseyb:
**Problem:**
Since we have added UIDs to all attributes, along with a check for duplicate UIDs, we are now triggering a duplicate UIDs error when multiselecting elements and setting a value via the Inspector.

**Fix:**
The problem here is that we create a single `JSExpressionValue` upfront once (at which point we generate the random UID) via a printer function (printers are the internal term used in the Inspector code for constructing the parsed model value from a css value), and then use that to set the attribute on all of the selected elements. To fix this, I've changed the function `collectActionsToSubmitValue` used by the Inspector's property path hooks to take a value printer rather than the final value itself, meaning that each generated action will be creating a new `JSExpressionValue`, and therefore generating a new UID for each.

### DIFF
--- a/editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
+++ b/editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
@@ -52,6 +52,7 @@ import {
   ConditionalsControlSectionOpenTestId,
   ConditionalsControlSwitchBranchesTestId,
 } from '../sections/layout-section/conditional-section'
+import { getAllUniqueUids } from '../../../core/model/get-unique-ids'
 async function getControl(
   controlTestId: string,
   renderedDOM: RenderResult,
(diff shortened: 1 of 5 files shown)


enidemi (approved):

editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
```diff
@@ -52,6 +52,7 @@ import {
   ConditionalsControlSectionOpenTestId,
   ConditionalsControlSwitchBranchesTestId,
 } from '../sections/layout-section/conditional-section'
+import { getAllUniqueUids } from '../../../core/model/get-unique-ids'
```
seanparsons, line 55: Not necessary?

seanparsons (approved):

editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
Rheeseyb: Ah this is left over from a previous attempt

PR #3844, opened 2023-06-21
Text only conditionals parsed as conditionals

gbalint:
**Problem:**
We wanted text only conditionals to be fully text editable, so when you have a conditional like `{ true ? 'hello' : 'utopia }`, we wanted the text editor to show the full conditional code instead of just the active branch.
A solution for this was to not parse these text only conditionals as conditionals, but as ATTRIBUTE_OTHER_JAVASCRIPT expressions. These expressions are text editable anyway, so that solved the issue.

However, it became really clear that it is unpredictable for the users that some conditionals appear as conditionals in the navigator, others as expressions, and the whole thing is just really confusing.

**Fix:**
I wanted to keep the advantage that text only conditionals are text editable, but at the same time I wanted all conditionals to be parsed as conditionals, so you can see them in the navigator (and even text edit their branch separately).

The solution was to add a new kind of `textProp` to the text editor, the `fullConditional`. In this mode the text editor can edit the full conditional code and save back in the project. The element renderer decides whether this conditional should only provide text editing for its active branch (when the inactive branch is not just an expression, but e.g. a jsx element), or the full conditional code.

After save we reparse the full project, so it is not a problem if the user modified the conditional code in a way that it is not a conditional anymore.

For the note, we can also decide the provide full conditional text editing for all conditionals - I was just afraid of potentially huge jsx code inside the text editor, so I did not want to go this far. But later we can make this decision more explicit, we could even provide UI to the user to switch between the different text edit modes.

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer-parsing.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer-parsing.ts
@@ -2519,51 +2519,6 @@ export function parseOutJSXElements(
       : expression.whenFalse
     const whenFalseBlock = parseClause(innerWhenFalse)
 
-    const parseAsFullConditionalExpression = (() => {
-      const trueBlockJsxElementLike =
-        isRight(whenTrueBlock) && isJSXElementLike(whenTrueBlock.value.value)
-      const trueBlockConditionalExpression =
-        isRight(whenTrueBlock) && isJSXConditionalExpression(whenTrueBlock.value.value)
-      const falseBlockJsxElementLike =
-        isRight(whenFalseBlock) && isJSXElementLike(whenFalseBlock.value.value)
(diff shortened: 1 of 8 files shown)


editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -2519,51 +2519,6 @@ export function parseOutJSXElements(
       : expression.whenFalse
     const whenFalseBlock = parseClause(innerWhenFalse)
 
-    const parseAsFullConditionalExpression = (() => {
```
seanparsons, line 2522: Nice.

seanparsons (approved):

bkrmendy (approved):

PR #3848, opened 2023-06-22
Easier un-focusing of focused components

Rheeseyb:
**Problem:**
When a component has been explicitly focused, the only way to un-focus it is via the "Exit editing component" in the Navigator's context menu.

**Fix:**
We used to un-focus a component when deselecting it, but that was far too eager, as often you'd want to keep the component focused whilst making other changes on the Canvas. There was also the idea to use the esc key to un-focus components, but that key is also used for jumping the current selection to the parent, which makes sense if the focused component isn't currently the selected element. So I've introduced 2 ways to un-focus a component:

1. If nothing is selected, clicking the empty canvas (which triggers a `CLEAR_SELECTION` action) will then clear the `focusedPath`. This means you can completely deselect the element without immediately un-focusing it.
2. If nothing is selected, hitting the esc key will clear the `focusedPath`.
3. If the focused element itself is selected, hitting the esc key will clear the `focusedPath`. This is so that we can still use esc to jump to the parent selection if any other element is selected. IMO this feels very intuitive, since if we always cleared the `focusedPath` when something else was selected the action would be invisible in a lot of cases (seeming like the editor was unresponsive), or jarring if it were noticed.

**Examples:**

1. Hitting esc repeatedly to move up through the hierarchy. Here the first `App` component has been focused, so as soon as we hit esc when it is selected we un-focus it:
easier-unfocusing-1

2. Hitting esc when nothing has been selected. Again the `App` component has been focused, so hitting esc here will un-focus it:
easier-unfocusing-2

3. Selecting nothing, then clicking the empty canvas again. The first canvas click will just clear the selection, whereas the second will un-focus the `App` component (please ignore the styling of the `App` component in the Navigator - it should be styled to show that it is focused, but that is a separate known bug):
easier-unfocusing-3

### DIFF
--- a/editor/src/components/canvas/canvas.ts
+++ b/editor/src/components/canvas/canvas.ts
@@ -144,7 +144,7 @@ const Canvas = {
       case 1:
         // Only a single element is selected...
         const parentPath = EP.parentPath(selectedViews[0])
-        if (parentPath == null) {
+        if (EP.isEmptyPath(parentPath)) {
           // ...the selected element is a top level one, so deselect.
           return 'CLEAR'
         }
--- a/editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
+++ b/editor/src/components/canvas/controls/select-mode/select-mode.spec.browser2.tsx
@@ -1106,6 +1106,258 @@ describe('Storyboard auto-focusing', () => {
   })
 })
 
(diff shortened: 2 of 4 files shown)


ruggi (approved):
This is awesome

enidemi (approved):

PR #3850, opened 2023-06-22
Fix `flushSync` warnings

bkrmendy:
## Problem
Lately we've been getting a lot more `flushSync` warnings

## Fix
Comb through the `React.useEffect` calls, and check if all `dispatch` calls are wrapped into a setTimeout.

Also, `data-testId` -> `data-testid` in the strategy picker


seanparsons (approved):

gbalint (approved):
yes!

PR #3854, opened 2023-06-22
Fixing some noUncheckedIndexedAccess errors.

seanparsons:
**Problem:**
We really want to turn on the `noUncheckedIndexedAccess` TypeScript option but then we'd get a huge pile of errors immediately breaking the build.

**Fix:**
I turned the option on and worked my way through some of these files, which I hope to do periodically to eventually allow us to turn `noUncheckedIndexedAccess` on permanently. I tried to keep the style of the code as much as reasonably possible.

**Commit Details:**
- Fixing some noUncheckedIndexedAccess errors.

### DIFF
--- a/editor/src/utils/utils.ts
+++ b/editor/src/utils/utils.ts
@@ -393,7 +393,7 @@ function resolveRef($ref: string, completeSchema: JsonSchema): JsonSchema {
     // this ONLY works with top level definitions
     // TODO make it work with deep refs
     const definitionName = $ref.slice(18)
-    return completeSchema.definitions![definitionName]
+    return completeSchema.definitions![definitionName]!
   } else {
     throw new Error(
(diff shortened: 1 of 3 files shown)


ruggi (approved):

editor/src/utils/utils.ts
```diff
@@ -837,36 +837,36 @@ function immutableUpdate(
...
-        )
-        return result
-      } else {
-        // Objects.
-        const defaultedObject: { [key: string]: any } = defaultIfNull({}, valueToUpdate)
-        return {
-          ...defaultedObject,
-          [first]: immutableUpdate(defaultedObject[first], remainder, valueToSet),
-        }
+  const first = objPath[0]
```
Rheeseyb, line 840: It looks like we're not using `immutableUpdate` anyway, so this whole function (as well as `immutableUpdateField`) can go in the bin

Rheeseyb (approved):

PR #3855, opened 2023-06-23
Always use the component orange colour for all focused elements in the Navigator

Rheeseyb:
## [Try me](https://utopia.pizza/p/31541d66-rainbow-archaeology/?branch_name=fix-consistent-nav-focus-styling)

**Problem:**
The Navigator has inconsistent styling for focused elements, sometimes using orange, sometimes using whatever other default colour a row would have.

**Fix:**
Fix up the styling to keep it consistent in all cases where a component is focused. This includes automatically focused components, i.e. those that are a single child of a `Scene`.

### DIFF
--- a/editor/src/core/shared/element-path.ts
+++ b/editor/src/core/shared/element-path.ts
@@ -328,8 +328,15 @@ export function fullDepth(path: ElementPath): number {
   return path.parts.reduce((working, part) => working + part.length, 0)
 }
 
-export function isInsideFocusedComponent(path: ElementPath): boolean {
-  return path.parts.length > 1
+export function isInsideFocusedComponent(
+  path: ElementPath,
+  autoFocusedPaths: Array<ElementPath>,
+): boolean {
+  return (
+    path.parts.length > 2 ||
+    (path.parts.length > 1 &&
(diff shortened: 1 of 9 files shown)


bkrmendy (approved):

ruggi (approved):

seanparsons (approved):

ruggi (approved):

editor/src/core/shared/element-path.ts
```diff
@@ -328,8 +328,15 @@ export function fullDepth(path: ElementPath): number {
   return path.parts.reduce((working, part) => working + part.length, 0)
 }
 
-export function isInsideFocusedComponent(path: ElementPath): boolean {
-  return path.parts.length > 2
+export function isInsideFocusedComponent(
```
seanparsons, line 331: I think I'd be tempted to put these in a different file as it feels a bit like it's diluting the purpose of this one, but I certainly wouldn't protest if you didn't.

seanparsons (approved):

editor/src/core/shared/element-path.ts
Rheeseyb: Yeah I felt the same way tbh but couldn't think of a better place to put them.

PR #3862, opened 2023-06-27
Overflowing elements drag handling more consistent.

seanparsons:
**Problem:**
When the children overflow out of an ancestor, the behaviour is inconsistent between these two cases:
- Click on the child and immediately drag.
- Click on the child, then start a drag from a subsequent click.

**Example:**
image
Clicking on the bottom element which is a descendant of the bottom element.

**Cause:**
In the second case above, selection is triggered without the interaction session starting, which then wipes the highlighted views value. Some of the logic for the mouse down handling is reliant on that highlighted views value being present to determine the behaviour.

**Fix:**
Where currently we wipe the highlighted views value when changing the selected views, that change has now been removed which means that the handling stays more consistent regardless of how many times the element gets clicked.

**Commit Details:**
- Removed the `setLocalHighlightedViews` call from `setSelectedViewsLocally`.
- Removed update for `highlightedViews` update from `SELECT_COMPONENTS` action handler.


ruggi (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -692,7 +692,7 @@ function useSelectOrLiveModeSelectAndHover(
   const { onMouseMove: innerOnMouseMove } = useHighlightCallbacks(
     active,
     cmdPressed,
-    false,
+    true,
```
Rheeseyb, line 695: Looks like this is now `true` everywhere, so we should just remove that param

Rheeseyb (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
seanparsons: Made that change in 51d530ebf.

PR #3863, opened 2023-06-27
Testing: Try Magic Nix Cache.

seanparsons:
**Problem:**
Nix can be quite slow to populate its cache.

**Fix:**
Use Determinate Systems "Magic Nix Cache" to cache the nix store values directly via an API.

**Commit Details:**
- Added Magic Nix Cache to the pull requests workflow.


seanparsons:
Possibly worth revisiting in the future.

PR #3866, opened 2023-06-27
Code items in the navigator

gbalint:
**Description:**
The goal of the PR is to make code (ATTRIBUTE_OTHER_JAVASCRIPT) items first class citizens in the navigator.

**Changes:**
- introduced a new feature switch `'Code in navigator'` so the PR is mergeable early
- include the uid of ATTRIBUTE_OTHER_JAVASCRIPT in the element path, make sure they are in the metadata
- metadata for the code elements: using the fake spy to generate that, and existing code for fragments/conditionals already calculates the frame for them from their children
- new navigator icon for the code elements
- made sure code which does not generate children elements (e.g. just returns with a string) do not appear in the navigator
- changed the label from `(code)` to the actual code of the element (can be too long, but that is true for text too)
- added a few tests (not enough, but the feature switch is off by default, so that can come later).


editor/src/core/shared/element-path-tree.ts
```diff
@@ -141,9 +141,16 @@ function getReorderedPaths(
...
+  const elementsToReorder = original.filter((path) => {
+    const element = MetadataUtils.findElementByElementPath(metadata, path)
+    if (element == null) {
+      return false
+    }
+    return (
+      MetadataUtils.isConditionalFromMetadata(element) ||
+      MetadataUtils.isExpressionOtherJavascriptFromMetadata(element)
+    )
+  })
```
Rheeseyb, line 153: (I think the answer is "no", but just to be sure) Does this block need to also check the feature switch?

ruggi: no I don't think so, that reordering for conditionals should just apply transparently in this case as well

gbalint: Not really, because there are expressions are not in the metadata without the feature switch

ruggi: I mean, even if they are not in the metadata this won't be affecting the editors where the FS is off 🤔

Rheeseyb (approved):

PR #3842, opened 2023-06-21
Elements outside visible area indicators

ruggi:
Fixes #3859 

**Problem:**

If a selected/highlighted element is outside the visible canvas area it can be difficult to figure out "where" it actually is.

**Fix:**

This PR introduces indicator arrows that appear at the edges of the visible canvas, pointing to selected/highlighted elements that are outside of it.

https://github.com/concrete-utopia/utopia/assets/1081051/89b4e899-c3b1-40cf-9db3-94af72df5d63

- The arrows are oriented relative to the center of the canvas.
- If there are 2+ overlapping arrows, they get clustered and only one is displayed, with a counter at its side
- If the arrow is clicked, the canvas scrolls to the target element. Note: the scroll will bring the element to the top-left corner of the canvas instead than center, but that's besides the scope of this PR (the `scrollToElement` action is used)
- To avoid having arrows disappearing/hidden by the canvas toolbar, the canvas toolbar is considered a safe area and arrows on the left side of the canvas will be shown on the right _of the toolbar_ (if applicable)
- If there are 10+ clustered elements, the cluster label will be forced to `"10+"`

Note: to avoid having to do a very complicated set of zoom/scale/positioning adjustments, I positioned the component rendering the arrows so that it is _not_ part of the zoomed canvas (and I don't think it should be, since canvas zooming should not affect this kind of indicators) but instead relative to its container.

PR #3870, opened 2023-06-28
Scroll to element and bring it to the center of the canvas

### LINKED ISSUE #3869: Add the ability to scroll to an element bringing it to the center of the canvas

ruggi:
Fixes #3869 

**Problem:**

The `SCROLL_TO_ELEMENT` function does not support scrolling so that the target element ends up in the center of the canvas. This came up clearly in #3842, and this PR is a follow up to that.

**Fix:**

https://github.com/concrete-utopia/utopia/assets/1081051/1fc9944b-8f31-4648-a36a-ae96c368240d

Extend the "behaviour" of the `SCROLL_TO_ELEMENT` action so that it can do three things depending on its invocation:
1. scroll to origin (top left) of the canvas
2. scroll to the center of the canvas **`(← added in this PR)`**
3. do nothing if the element is already visible

The new scroll-to-center behaviour is used only in the arrow indicators, as I kept the other two for compatibility; they can be swapped out though by changing the behaviour parameter.

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -424,6 +424,7 @@ import {
   reparentTargetFromNavigatorEntry,
   modifyOpenJsxChildAtPath,
   isConditionalClauseNavigatorEntry,
+  DefaultNavigatorWidth,
 } from '../store/editor-state'
 import { loadStoredState } from '../stored-state'
 import { applyMigrations } from './migrations/migrations'
@@ -4875,7 +4876,7 @@ export const UPDATE_FNS = {
     if (targetElementCoords != null && isFiniteRectangle(targetElementCoords)) {
       const isNavigatorOnTop = !editor.navigator.minimised
       const containerRootDiv = document.getElementById('canvas-root')
-      const navigatorOffset = isNavigatorOnTop ? LeftPaneDefaultWidth : 0
+      const navigatorOffset = isNavigatorOnTop ? DefaultNavigatorWidth : 0
 
       // This returns the offset for 'to-origin' scroll behaviours, or used as the default
       // for the other behaviours.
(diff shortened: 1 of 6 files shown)


editor/src/components/editor/actions/actions.tsx
```diff
@@ -4877,36 +4878,79 @@ export const UPDATE_FNS = {
...
+
+      function canvasOffsetToCenter(
+        frame: CanvasRectangle,
+        bounds: CanvasRectangle | null,
+      ): CanvasVector {
+        if (bounds == null) {
+          return canvasOffsetToOrigin(frame) // fallback default
+        }
+        const xOffset = navigatorOffset > 0 ? navigatorOffset + 20 : 0
+        return canvasPoint({
```
enidemi, line 4898: I think here the returned point needs to be offset with the scroll position too, now this only works correctly on elements placed with top, left 0,0 on the canvas.

https://utopia.pizza/p/a3003332-eastern-ferry/?branch_name=feat-scroll-to-center

ruggi: should be fixed with https://github.com/concrete-utopia/utopia/pull/3870/commits/a630a155d511f4c79f9378ac9d70e2737216df70

editor/src/components/editor/actions/actions.tsx
```diff
@@ -4873,36 +4875,88 @@ export const UPDATE_FNS = {
...
+        const canvasCenter = getRectCenter(
+          canvasRectangle({
+            x: bounds.x,
+            y: bounds.y,
+            width: bounds.width * scale,
+            height: bounds.height * scale,
+          }),
+        )
+        const topLeftTarget = canvasPoint({
+          x: canvasCenter.x - frame.width / 2 - bounds.x + (navigatorOffset / 2 + 10) * scale,
```
Rheeseyb, line 4904: What's with the `+ 10`?

Rheeseyb (approved):

editor/src/components/editor/actions/actions.tsx
ruggi: ah it's an ugly remnant of the navigator padding (20 / 2), I updated to use `DefaultNavigatorWidth` instead of `LeftPaneDefaultWidth` which correctly is 280 instead of 260 https://github.com/concrete-utopia/utopia/pull/3870/commits/f5d3088ea55fdc0a74197b571c2aaa5783721126

bkrmendy (approved):

PR #3872, opened 2023-06-28
Handle elements that subsequently become expressions.

seanparsons:
**Problem:**
If an element with a given `data-uid` was wrapped in the code editor with an expression, then a duplicate UIDs error would trigger.

**Cause:**
In `fixJSXElementChildUIDs` if the new element was an attribute like value and the old element is not, then the new element is not walked and only its UID is updated. In the case above, this meant that `elementsWithin` would not be walked, but the UID would be copied over but would already exist in the unchanged `elementsWithin`.

**Fix:**
`fixJSXElementChildUIDs` has now been changed to always walk the attribute like values, regardless of the type.

**Commit Details:**
- `fixJSXElementChildUIDs` now calls `fixExpressionUIDs` for every possible case of `oldElement`.
- `fixExpressionUIDs` now supports a wider type for `oldExpression`.

### DIFF
--- a/editor/src/core/workers/parser-printer/uid-fix.spec.ts
+++ b/editor/src/core/workers/parser-printer/uid-fix.spec.ts
@@ -8,6 +8,8 @@ import {
   isJSXElementLike,
   isJSXFragment,
   JSExpression,
+  jsExpressionOtherJavaScript,
+  jsxElement,
   JSXElement,
   JSXElementChild,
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):

enidemi (approved):

Rheeseyb (commented):
👍

PR #3874, opened 2023-06-28
remove Feature Switch "Nine block control"

balazsbajorics:
The 'Nine block control' feature switch has been enabled for a long time now, and I believe nothing relies on its off state. So lets remove it!


editor/src/components/inspector/sections/layout-section/self-layout-subsection/gigantic-size-pins-subsection.tsx
```diff
@@ -587,18 +587,6 @@ export const GiganticSizePinsSubsection = React.memo((props: GiganticSizePinsSub
 
   return (
     <>
-      {unless(
-        isFeatureEnabled('Nine block control'),
-        <WidthHeightRow
```
seanparsons, line 592: I think this now makes this component unused right?

balazsbajorics: yeah I shall remove it probably, altough it _is_ containing the missing code that allows width and height pinning... maybe I should resurrect just that part and remove the others

balazsbajorics: image

@maltenuhn wdyt how wonky this looks? on master we are missing the two pin buttons for width and height, with this change I could put them back in an empty row. it looks unprofessional, but today IMO is even worse because we've simply lost these buttons and setting ie right, bottom, width, height pins is a pain today

Rheeseyb (commented):
👍

PR #3876, opened 2023-06-28
Remove DragState and canvasTransientState and related unused functions

balazsbajorics:
**Situation**
The DragState / canvasTransientState ha been completely superseded by the InteractionSession / Canvas Strategy code.
For a while we've had a line in dispatch.tsx that threw an Error if it encountered a non-null EditorState.dragState value. So it was guaranteed that no drag-state related code is still alive. 
But the code was left around and never removed... until today! 

**Fix**
This PR is removing EditorState.dragState, the DragState type and related types, the canvas transient state and `produceCanvasTransientState` family of functions. I went on a little killing spree and removed all unused functions from canvas-utils.ts too. So much dead code removed!


bkrmendy (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -67,30 +61,12 @@ const DRAG_START_THRESHOLD = 2
 
 export function isResizing(editorState: EditorState): boolean {
   // TODO retire isResizing and replace with isInteractionActive once we have the strategies turned on, and the old controls removed
```
Rheeseyb, line 63: Given this function and the one below are both returning true in exactly the same conditions, it looks like it's time for this `TODO` (maybe as a follow up PR if it causes too much fallout in this one?)

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2131,26 +2122,18 @@ export const UPDATE_FNS = {
       // FIXME: this is a bit unfortunate as this action should just do what its name suggests, without additional flags.
       // For now there's not much more that we can do since the action here can be (and is) evaluated also for transient states
       // (e.g. a `textEdit` mode after an `insertMode`) created with wildcard patches.
-      return clearDragState(editor, derived, false)
+      return editor
```
Rheeseyb, line 2125: I wonder if this action and `UPDATE_EDITOR_MODE` should be merged now 😬

editor/src/templates/editor-canvas.tsx
```diff
@@ -328,7 +293,7 @@ function on(
       return []
     }
   } else if (canvas.keysPressed['z']) {
-    if (event.event === 'MOUSE_UP' && !isDragging(canvas.editorState)) {
+    if (event.event === 'MOUSE_UP') {
```
Rheeseyb, line 296: Oh interesting, it seems we've had a bug here for a while then, because this means you can zoom in and out accidentally whilst dragging to move or resize an element. I'd say it's an extremely minor bug tbh, and such an edge case that I'm not surprised we hadn't caught it before.

editor/src/templates/editor-canvas.tsx
```diff
@@ -357,31 +322,6 @@ function on(
     } else {
       return [CanvasActions.scrollCanvas(event.delta as any as CanvasVector)]
     }
-  } else if (
```
Rheeseyb, line 360: Is it maybe worth seeing if we can adapt this to use the canvas strategies work so that we can restore the canvas scrolling when dragging past the bounds of the canvas?

Rheeseyb: (I guess as a follow up PR?)

editor/src/templates/editor-canvas.tsx
```diff
@@ -888,11 +822,9 @@ export class EditorCanvas extends React.Component<EditorCanvasProps> {
     )
 
     const modeOverrideCursor = this.getModeSpecificCursor()
-    const dragStateCursor = null // FIXME: dragState == null ? null : dragState.cursor
```
Rheeseyb, line 891: Fixed!

editor/src/templates/editor-canvas.tsx
```diff
@@ -1104,12 +1036,16 @@ export class EditorCanvas extends React.Component<EditorCanvasProps> {
       <EditorCommon
         mouseDown={this.handleMouseDown}
         mouseUp={this.handleMouseUp}
-        keyDown={this.handleKeyDown}
-        keyUp={this.handleKeyUp}
+        keyDown={this.handleKeyNoOp}
+        keyUp={this.handleKeyNoOp}
```
Rheeseyb, line 1040: Do we need these?

Rheeseyb (approved):

editor/src/templates/editor-canvas.tsx
balazsbajorics: good catch, this was meant to be temporary

seanparsons (approved):

editor/src/components/editor/actions/actions.tsx
balazsbajorics: added a follow-up comment

editor/src/templates/editor-canvas.tsx
balazsbajorics: fixed

editor/src/templates/editor-canvas.tsx
balazsbajorics: added a TODO in the paper doc

PR #3881, opened 2023-06-30
Correct paths using tilde in workflows

Rheeseyb:
**Problem:**
We are incorrectly caching the wrong path for the pnpm store. This meant we couldn't fall back to relying on the offline installed versions of packages when the `console-feed` package disappeared from the public npm registry.

**Fix:**
Introduce a shared workflow that sets up nix, captures the actual pnpm-store path, runs all of the pnpm install steps, caches the result, and then gives that store path to all other jobs so that they can use that cache

### DIFF
--- a/.github/workflows/master-pushes.yml
+++ b/.github/workflows/master-pushes.yml
@@ -4,9 +4,15 @@ on:
     branches: [master]
 
 jobs:
+  cache-pnpm-store:
+    name: Install everything PNPM and cache it
+    secrets: inherit
+    uses: ./.github/workflows/cache-pnpm-install.yml
+
   test-editor-code:
     name: Test Editor - TypeScript, ESLint, dependency-cruiser
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


ruggi (approved):

balazsbajorics (approved):

.github/workflows/master-pushes.yml
```diff
@@ -4,9 +4,15 @@ on:
     branches: [master]
 
 jobs:
+  cache-pnpm-store:
+    name: Install everything PNPM and cache it
```
seanparsons, line 8: I think this makes more sense right?
```suggestion
    name: Install everything with PNPM and cache it
```

seanparsons (approved):

.github/workflows/master-pushes.yml
Rheeseyb: Bah, I missed this! Will fix and push

PR #3882, opened 2023-06-30
Chore/remove modify open parse success

balazsbajorics:
It was discovered that `modifyOpenParseSuccess` is dangerous and may not even work properly.

**Details:**
I removed the following dormant functions / actions:
- removed `SET_PROP_WITH_ELEMENT_PATH`
- removed `modifyOpenParseSuccess` / `modifyOpenJSXElements`
- removed unrelated but unused `modifyOpenJSXElementsAndMetadata`
- `modifyUnderlyingTargetElement` no longer accepts a null as target path for the element
- `modifyUnderlyingJsxElementChild` no longer accepts a null target path

The only place using `SET_PROP_WITH_ELEMENT_PATH` was `SAVE_ASSET` where I commented out the line and left a comment.

### DIFF
--- a/editor/src/components/editor/action-types.ts
+++ b/editor/src/components/editor/action-types.ts
@@ -734,13 +734,6 @@ export interface SetProp {
   value: JSExpression
 }
 
-export interface SetPropWithElementPath {
-  action: 'SET_PROP_WITH_ELEMENT_PATH'
-  target: StaticElementPathPart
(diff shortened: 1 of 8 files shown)


enidemi (approved):

seanparsons (approved):

Rheeseyb (commented):
🎉

PR #3884, opened 2023-06-30
feature(navigator) show warning icon for elements with z-index

enidemi:
**Problem:**
Elements with z-index can make it difficult to work on the canvas and we display no information if an element has z-index set.

**Fix:**
Navigator warning triangle is shown for elements with z-index.

**Commit Details:**
- added z-index to `specialSizeMeasurements`
- added a new warning to `elementWarnings`


balazsbajorics (approved):

seanparsons (approved):

bkrmendy (changes requested):
Looks good but let's use a different icon, since detecting whether a z-index has an effect or not is out of scope for this PR, and just the simple fact of having z-index set on a component is not a problem users need to be warned about (which is implied by the warning icon)

PR #3886, opened 2023-06-30
Apply edits to the root element or instance appropriately.

seanparsons:
**Problem:**
There are a few cases around editing root elements or component instances where the changes should actually go to the component instance properties or the root element of the component directly, depending on a few circumstances:
- If the property is set in the component instance properties.
- If the property is set in the root element properties.
- If `props.style` are spread into the style prop of the root element of the component.


seanparsons:
Not the approach that we want to go with.

PR #3891, opened 2023-07-04
Handle duplicate data-uid prop uids.

seanparsons:
**Problem:**
When inserting certain components into the code at the start of a file, it was possible for a duplicate UID error to trigger.

**Cause:**
This was triggered as a result of a prior fix for how to update the `uid` for the attribute that assigns the `data-uid` prop, where that `uid` ends up being allocated earlier in the model but then gets reassigned from the prior parse result.

**Fix:**
Now we check to see if the `uid` for that attribute changed during the pass that attempts to line up the UIDs for the attributes. If there was a change, we apply the regular `updateUID` call to it, otherwise the forced update is applied (the regular case) which does not attempt any generation of UIDs.

This also necessitated extracting out the logic for adding the mapping, so that a mapping could be applied on top of an already existing mapping.

**Commit Details:**
- In `fixJSXElementUIDs` make the `data-uid` prop uid fixing be run whatever the case.
- Extract out `addMapping` and have it handle the case where a mapping is needed on top of an existing mapping.
- `fixJSXElementUIDs` now checks the status of the `data-uid` property uid after the properties are fixed to see how it should be handled.

### DIFF
--- a/editor/src/core/workers/parser-printer/uid-fix.ts
+++ b/editor/src/core/workers/parser-printer/uid-fix.ts
@@ -133,6 +133,17 @@ function updateUID<T>(
   fixUIDsState: FixUIDsState,
   baseValue: T,
 ): T {
+  function addMapping(originalUID: string, newUID: string): void {
+    const priorMapping = fixUIDsState.mappings.find((mapping) => {
+      return mapping.newUID === originalUID
+    })
+    if (priorMapping == null) {
+      fixUIDsState.mappings.push({ originalUID: originalUID, newUID: newUID })
+    } else {
+      priorMapping.newUID = newUID
+    }
+  }
+
   const newUID = unsafeGet(uidOptic, baseValue)
   let uidToUse: string
   switch (fixUIDsState.uidUpdateMethod) {
@@ -148,14 +159,14 @@ function updateUID<T>(
             fixUIDsState.mutableAllNewUIDs,
             fixUIDsState.uidsExpectedToBeSeen,
           )
-          fixUIDsState.mappings.push({ originalUID: newUID, newUID: uidToUse })
+          addMapping(newUID, uidToUse)
         } else if (oldUID === newUID) {
           // Old one is the same as the new one, so everything is great.
           uidToUse = newUID
         } else {
           // The UID has changed, add a mapping so the highlight bounds can be updated.
           uidToUse = oldUID
-          fixUIDsState.mappings.push({ originalUID: newUID, newUID: uidToUse })
+          addMapping(newUID, uidToUse)
         }
         fixUIDsState.mutableAllNewUIDs.add(uidToUse)
       }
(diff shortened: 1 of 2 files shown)


editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -133,6 +133,17 @@ function updateUID<T>(
   fixUIDsState: FixUIDsState,
   baseValue: T,
 ): T {
+  function addMapping(originalUID: string, newUID: string): void {
+    const priorMapping = fixUIDsState.mappings.find((mapping) => {
```
bkrmendy, line 137: Won't `find` here cause perf problems? I'm not sure this will scale with large projects

seanparsons: There doesn't tend to be huge numbers of mappings and this runs on the parser-printer worker so the performance demands are much lower than say during a canvas operation.

seanparsons: Also these mappings are per file, so the size of the project doesn't influence this particularly.

Rheeseyb: I think this is definitely worth checking for a large project. It seems we now use the `mappings` in two places: one which calls `find` based on the `mapping.originalUID`, and now this one which calls `fine` based on `mapping.newUID`. If this is a performance issue maybe it would be better two create a pair of maps (from originalUID to newUID and vice versa)?

Rheeseyb: Urgh, I've just seen your comments now that I've posted that. When I posted it they weren't showing because AJAX is still pretty new technology

Rheeseyb (approved):

editor/src/core/workers/parser-printer/uid-fix.ts
seanparsons: Oh yes and I tested this on the Before I Go and the 31x10 performance test projects and performance of the updates coming in appeared to be fine.

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -595,21 +617,31 @@ export function fixJSXElementUIDs(
...
-          identityOptic<string>(),
-          oldDataUIDProp.uid,
-          { ...fixUIDsState, uidUpdateMethod: 'forced-update' },
-          newDataUIDProp.uid,
-        )
-      }
+  let dataUIDPropUID: string
+  if (newDataUIDProp == null) {
+    // Backup case for where there is no `data-uid` prop.
+    dataUIDPropUID = generateConsistentUID(
```
bkrmendy, line 623: Shouldn't `updateUID` be called here? Is the mapping updated in this branch?

seanparsons: So it can't be there because we have no uids to update with, but it should add the UID to `mutableAllNewUIDs`, fixed in 67a36bd03.

PR #3893, opened 2023-07-04
Exit text edit mode if element does not support children

bkrmendy:
## Problem
When text edit mode is enabled on a component that doesn't use the `children` prop, the editor will enter text edit mode, but the actual contenteditable span doesn't make its way into the DOM (because its parent element doesn't render it). As a consequence, the event handlers that switch back to select mode (added to the contenteditable span) cannot fire, which makes the editor get stuck in text editor mode without an actual text editor.

## Fix
Add a check in `TextEditor` that checks whether the span is actually rendered. If it isn't, it switches the editor back to select mode and shows a toast that explains the situation.

### DIFF
--- a/editor/src/components/text-editor/text-editor.tsx
+++ b/editor/src/components/text-editor/text-editor.tsx
@@ -375,7 +375,7 @@ const TextEditor = React.memo((props: TextEditorProps) => {
         ])
       }, 0)
     }
-  })
+  }, [dispatch])
 
   React.useEffect(() => {
     if (myElement.current == null) {
--- a/editor/src/components/text-editor/text-editor.tsx
+++ b/editor/src/components/text-editor/text-editor.tsx
@@ -23,6 +23,7 @@ import {
   deleteView,
   updateText,
   updateEditorMode,
+  showToast,
 } from '../editor/actions/action-creators'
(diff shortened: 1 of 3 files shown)


editor/src/components/text-editor/text-editor.tsx
```diff
@@ -357,6 +359,24 @@ const TextEditor = React.memo((props: TextEditorProps) => {
     myElement.current.textContent = firstTextProp
   }, [firstTextProp])
 
+  React.useEffect(() => {
+    if (myElement.current == null) {
```
Rheeseyb, line 363: It seems this should live in the `useEffect` below, since that is also checking `myElement.current == null` and just early returning if true

Rheeseyb: Actually, on second thought I think you're right to separate this and provide no dependencies.

bkrmendy: Great suggestion, added it https://github.com/concrete-utopia/utopia/pull/3893/commits/c053a6cdf30cfbdd793e188a5c17d30e9412bc0c

Rheeseyb (approved):

PR #3894, opened 2023-07-04
fallback group true-up based on dom measurements 

balazsbajorics:


editor/src/templates/editor.tsx
```diff
@@ -471,36 +471,118 @@ export class Editor {
...
+                  this.canvasStore.setState(
+                    patchedStoreFromFullStore(this.storedState, 'canvas-store'),
+                  )
+                })
+              })
+            })
+
+            // re-run the dom-walker
+            Measure.taskTime(`Dom walker re-run because of groups ${updateId}`, () => {
+              const domWalkerResult = runDomWalker({
```
Rheeseyb, line 553: I suspect we should extract this block of code since it's near enough identical to the code above, so that if we need to make a change to the process we can ensure it's made in both places

bkrmendy (approved):

editor/src/templates/editor.tsx
balazsbajorics: good point, done

editor/src/components/canvas/canvas-strategies/strategies/groups.spec.browser2.tsx
```diff
@@ -59,6 +60,22 @@ async function dragByPixels(
   })
 }
 
+async function changeInspectorNumberControl(
```
seanparsons, line 63: Probably worth putting this in a test utils file somewhere.

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -352,32 +353,103 @@ export async function renderTestEditorWithModel(
...
+        )
+        workingEditorState = editorWithNewMetadata
+      }
+    }
+
+    // true-up groups if needed
+    if (workingEditorState.unpatchedEditor.trueUpGroupsForElementAfterDomWalkerRuns.length > 0) {
+      ;(() => {
+        // updated editor with trued up groups
+        {
```
seanparsons, line 390: Isn't this one scope too many?

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -352,32 +353,103 @@ export async function renderTestEditorWithModel(
...
+          workingEditorState,
+          spyCollector,
+        )
+        workingEditorState = editorWithNewMetadata
+      }
+    }
+
+    // true-up groups if needed
+    if (workingEditorState.unpatchedEditor.trueUpGroupsForElementAfterDomWalkerRuns.length > 0) {
+      ;(() => {
```
seanparsons, line 388: Is this in a lambda just so that `return` can be used at line 407?

balazsbajorics: yep. I tried to keep this code as similar to the one in editor.tsx as possible. I think eventually we should merge the two functions and the similar shape should help us match what's there vs what's here

editor/src/components/canvas/canvas-strategies/strategies/groups.spec.browser2.tsx
balazsbajorics: done!

editor/src/components/canvas/ui-jsx.test-utils.tsx
balazsbajorics: fixed!

enidemi (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -127,7 +127,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`494`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`537`)
```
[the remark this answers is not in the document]
balazsbajorics: all I know is that the re-render is caused by losing an editor state reference. but I couldn't pin down which selector is bad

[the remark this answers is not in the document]
balazsbajorics: OK I could nail it down to the ComponentSectionInner doing an extra re-render. I am 100% confident this only happens on SET_PROP interactions so it is not on a critical path and the actual performance regression caused by this is negligible. 
`ComponentSectionInner` uses some hooks that we wrote before we started using reselect and they look very hairy to me with a ton of FIXMEs sprinkled around. I think one of those hooks erroneously triggers a re-render when we lose reference to the store. React devtools helpfully tells me it's "Hook 19" that is causing the problem.

PR #3896, opened 2023-07-05
remove most ad-hoc performance measurements from editor.tsx

balazsbajorics:
I have an upcoming PR #3894 that touches the editor dispatch function in editor.tsx. the current ad-hoc performance measurements make the code hard to read and unsafe to refactor. 

in this PR I aim to make the code a little bit more palatable by using a single helper function to measure performance instead of relying on ad-hoc function calls and variables

### DIFF
--- a/editor/src/components/editor/store/editor-dispatch-performance-logging.tsx
+++ b/editor/src/components/editor/store/editor-dispatch-performance-logging.tsx
@@ -1,0 +1,63 @@
+import { PERFORMANCE_MARKS_ALLOWED } from '../../../common/env-vars'
+import { isFeatureEnabled } from '../../../utils/feature-switches'
+import { EditorAction } from '../action-types'
+import { simpleStringifyActions } from './dispatch'
+
+export function createPerformanceMeasure() {
+  const MeasureSelectorsEnabled = isFeatureEnabled('Debug – Measure Selectors')
+  const PerformanceMarks =
+    (isFeatureEnabled('Debug – Performance Marks (Slow)') ||
+      isFeatureEnabled('Debug – Performance Marks (Fast)')) &&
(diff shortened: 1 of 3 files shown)


editor/src/templates/editor.tsx
```diff
@@ -518,75 +503,44 @@ export class Editor {
...
 
-            // reset selector timings right before the end of flushSync means we'll capture the re-render related selector data with a clean slate
-            resetSelectorTimings()
+              // reset selector timings right before the end of flushSync means we'll capture the re-render related selector data with a clean slate
+              resetSelectorTimings()
+            })
           })
         })
         if (PerformanceMarks) {
           performance.measure(
```
balazsbajorics, line 540: I've kept this one as a manual measure because it measures is the amount of work React.flushSync does after the inner callback have returned.

enidemi (approved):

editor/src/components/editor/store/editor-dispatch-performance-logging.tsx
```diff
@@ -0,0 +1,63 @@
...
+  let stringifiedActions = ''
+
+  let tasks: {
+    [name: string]: { timingInfo: string; startMarkName: string; endMarkName: string }
+  } = {}
+  return {
+    logActions(actions: ReadonlyArray<EditorAction>): void {
+      stringifiedActions = simpleStringifyActions(actions)
+    },
+    taskTime<T>(name: string, task: () => T): T {
```
Rheeseyb, line 22: I think we should just early return in this function if `!MeasureSelectorsEnabled`, purely so that we don't later accidentally add extra computation that shouldn't be running when they're not enabled

Rheeseyb (approved):

PR #3902, opened 2023-07-06
Feature/paste here post action menu

enidemi:
**Problem:**
Paste Here is not using the post-action menu yet.

**Fix:**
This PR adds post-action menu commands and deletes the original action.

**Commit Details:**
- uses the common parts from paste post-action commands


editor/src/components/canvas/canvas-strategies/post-action-options/post-action-paste.ts
```diff
@@ -240,7 +298,188 @@ export const PasteWithPropsReplacedPostActionChoice = (
...
+    {
+      elementPaste: elementToPaste,
+      originalContextMetadata: originalMetadata,
+      originalContextElementPathTrees: originalPathTree,
+    },
+    editor.elementPathTree,
+  )
+  if (isLeft(target)) {
+    return target
+  }
```
Rheeseyb, line 461: If there's no target we should probably fall back to trying to paste onto the storyboard here

editor/src/components/editor/actions/actions.tsx
```diff
@@ -5608,7 +5471,7 @@ function saveFileInProjectContents(
   }
 }
 
-type ElementToInsert = {
+export type PasteElementToInsert = {
```
Rheeseyb, line 5474: It doesn't look like this (or the function below) need to be exported

Rheeseyb (approved):

editor/src/components/editor/actions/actions.tsx
enidemi: fixed

editor/src/components/canvas/canvas-strategies/post-action-options/post-action-paste.ts
enidemi: good idea, fixed

bkrmendy (approved):

PR #3907, opened 2023-07-07
Add tsconfig option to aid import autocompletion

Rheeseyb:
**Problem:**
#3900 introduced an eslint rule that forces us to replace type imports of the form `import {SomeType}` with `import Type SomeType`, but VS Code's import statement autocomplete was defaulting to the now disallowed style.

**Fix:**
Added the [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) option to the tsconfig files that I could. This provides enough of a hint to VS Code for it to use the `import type` style where it can (i.e. where the type is only being used as a type, which is the exact case that the eslint rule throws an error for).

### DIFF
--- a/utopia-vscode-extension/tsconfig.json
+++ b/utopia-vscode-extension/tsconfig.json
@@ -4,8 +4,9 @@
     "module": "commonjs",
     "moduleResolution": "node",
     "outDir": "out",
-    "lib": [ "es2019", "WebWorker" ],
-    "sourceMap": true
+    "lib": ["es2019", "WebWorker"],
+    "sourceMap": true,
+    "verbatimModuleSyntax": true
   },
-  "include": [ "src" ]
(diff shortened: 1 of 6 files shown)


bkrmendy (approved):

utopia-vscode-extension/tsconfig.json
```diff
@@ -4,8 +4,8 @@
     "module": "commonjs",
     "moduleResolution": "node",
     "outDir": "out",
-    "lib": [ "es2019", "WebWorker" ],
```
balazsbajorics, line 7: did you mean to add it to this tsconfig too?

Rheeseyb:
Closing for now as this requires updating TS and results in hundreds of errors 😞

PR #3908, opened 2023-07-07
Fix/inspector size section group true up

balazsbajorics:
**Problem:**
I found a bug
https://screenshot.click/07-04-gub1v-og4s8.mp4
When changing the width / height in the inspector section for an element that had a right / bottom pin, the `predictElementSize` function used in the `fixedSizeBasicStrategy` inspector strategy would predict a wrong frame, resulting in a wrong group.

**Fix:**
Actually just use the measurement based group true-up system for these controls too. 

**Commit Details:**
- Added `QueueGroupTrueUp` command which queues the target element in `trueUpGroupsForElementAfterDomWalkerRuns`
- Fixed a bug in `getUpdateResizedGroupChildrenCommands` that assumed we are working with stale metadata
- Use QueueGroupTrueUp in fixedSizeBasicStrategy
- New test case demonstrating a scenario that fails on master

### DIFF
--- a/editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
+++ b/editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
@@ -143,15 +143,19 @@ function getUpdateResizedGroupChildrenCommands(
         frameAndTarget.target,
       )
 
+      // the original size of the group before the interaction ran
       const originalSize: Size =
         command.isMetadataStale === 'metadata-is-stale'
-          ? sizeFromRectangle(
+          ? // if we have the starting metadata, we can simply get the original measured bounds of the element and we know it's the originalSize
+            sizeFromRectangle(
               MetadataUtils.getLocalFrameFromSpecialSizeMeasurements(
                 frameAndTarget.target,
                 editor.jsxMetadata,
               ),
             )
-          : sizeFromRectangle(
+          : // if the metadata is fresh, the group is already resized. so we need to query the size of its children AABB
+            //(which was not yet updated, since this function is updating the children sizes) to get the originalSize
+            sizeFromRectangle(
               boundingRectangleArray(
                 children.map((c) =>
                   nullIfInfinity(
--- a/editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
+++ b/editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
@@ -49,15 +49,18 @@ import { wildcardPatch } from './wildcard-patch-command'
 export interface PushIntendedBoundsAndUpdateGroups extends BaseCommand {
   type: 'PUSH_INTENDED_BOUNDS_AND_UPDATE_GROUPS'
   value: Array<CanvasFrameAndTarget>
+  isMetadataStale: 'metadata-is-stale' | 'metadata-is-fresh'
 }
 
 export function pushIntendedBoundsAndUpdateGroups(
   value: Array<CanvasFrameAndTarget>,
+  isMetadataStale: 'metadata-is-stale' | 'metadata-is-fresh',
 ): PushIntendedBoundsAndUpdateGroups {
   return {
     type: 'PUSH_INTENDED_BOUNDS_AND_UPDATE_GROUPS',
     whenToRun: 'always',
     value: value,
+    isMetadataStale: isMetadataStale,
   }
 }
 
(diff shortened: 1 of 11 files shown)


bkrmendy (approved):

editor/src/components/inspector/inspector-strategies/fixed-size-basic-strategy.ts
```diff
@@ -43,7 +36,7 @@ export const fixedSizeBasicStrategy = (
           setExplicitCssValue(value),
           parentFlexDirection,
         ),
-        pushIntendedBoundsAndUpdateGroups([{ target: path, frame: predictedElementSize }]),
+        queueGroupTrueUp(path),
```
balazsbajorics, line 39: basically this line is the interesting change in the PR

editor/src/components/canvas/commands/queue-group-true-up-command.ts
```diff
@@ -0,0 +1,31 @@
...
+    element: element,
+  }
+}
+
+export const runQueueGroupTrueUp: CommandFunction<QueueGroupTrueUp> = (
+  editorState: EditorState,
+  command: QueueGroupTrueUp,
+) => {
+  return {
+    commandDescription: `Queue element for group true-up once the intearction finished: ${EP.toString(
```
Rheeseyb, line 24:
```suggestion
    commandDescription: `Queue element for group true-up once the interaction has finished: ${EP.toString(
```

editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
```diff
@@ -134,24 +137,38 @@ function getUpdateResizedGroupChildrenCommands(
...
+            )
+          : sizeFromRectangle(
+              boundingRectangleArray(
+                children.map((c) =>
+                  nullIfInfinity(
+                    MetadataUtils.findElementByElementPath(editor.jsxMetadata, c)?.globalFrame,
+                  ),
+                ),
+              ),
+            )
```
Rheeseyb, line 162: I really don't understand why if the metadata is stale we're using the special size measurements of the target, but if the metadata is fresh we're using the bounding rectangle of the children. Please can you add a comment here clarifying why

editor/src/components/inspector/inspector-strategies/fixed-fill-hug.spec.browser2.tsx
```diff
@@ -894,7 +985,7 @@ describe('Fixed / Fill / Hug control', () => {
 
       const control = editor.renderedDOM.getByTestId(FillFixedHugControlId('height'))
       await mouseClickAtPoint(control, { x: 5, y: 5 })
-      await expectSingleUndo2Saves(editor, async () => {
+      await expectSingleUndoNSaves(editor, 3, async () => {
```
Rheeseyb, line 988: 3 saves! 👀

balazsbajorics: single undo!

editor/src/components/inspector/inspector-strategies/fixed-size-basic-strategy.ts
Rheeseyb: Don't sell your work short, there are many interesting changes in this PR :)

editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
```diff
@@ -134,24 +137,38 @@ function getUpdateResizedGroupChildrenCommands(
...
-        ),
+      const children = MetadataUtils.getChildrenPathsOrdered(
+        editor.jsxMetadata,
+        editor.elementPathTree,
+        frameAndTarget.target,
       )
+
+      const originalSize: Size =
+        command.isMetadataStale === 'metadata-is-stale'
+          ? sizeFromRectangle(
```
seanparsons, line 148: Minor: Could probably make this a little less hectic by extracting out into the conditional what we pass to `sizeFromRectangle`.

seanparsons (approved):

editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
balazsbajorics: I've pushed comments. if you have any suggestions that are better than 'metadata-is-stale' I'm happy to go for it

Rheeseyb (approved):

PR #3909, opened 2023-07-07
Sizeless div flag in `getElementFragmentLikeType`

bkrmendy:
## Description

This PR adds a flag to `getElementFragmentLikeType`, which can be used to exclude sizeless divs from the element types that are considered fragment-like.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/fragment-like-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/fragment-like-helpers.ts
@@ -171,11 +171,20 @@ export const AllFragmentLikeNonDomElementTypes = ['fragment', 'conditional'] as 
 export const AllFragmentLikeTypes = [...AllFragmentLikeNonDomElementTypes, 'sizeless-div'] as const
 export type FragmentLikeType = typeof AllFragmentLikeTypes[number] // <- this gives us the union type of the Array's entries
 
+export interface GetElementFragmentLikeTypeOptions {
+  sizelessDivsConsideredFragmentLike: boolean
+}
+
+export const DefaultGetElementFragmentLikeTypeOptions: GetElementFragmentLikeTypeOptions = {
+  sizelessDivsConsideredFragmentLike: true,
+}
+
 export function getElementFragmentLikeType(
   metadata: ElementInstanceMetadataMap,
   allElementProps: AllElementProps,
   pathTrees: ElementPathTrees,
   path: ElementPath,
+  options: Partial<GetElementFragmentLikeTypeOptions> = DefaultGetElementFragmentLikeTypeOptions,
 ): FragmentLikeType | null {
   const elementMetadata = MetadataUtils.findElementByElementPath(metadata, path)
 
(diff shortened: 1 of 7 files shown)


balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/strategies/fragment-like-helpers.ts
```diff
@@ -171,11 +171,20 @@ export const AllFragmentLikeNonDomElementTypes = ['fragment', 'conditional'] as
 export const AllFragmentLikeTypes = [...AllFragmentLikeNonDomElementTypes, 'sizeless-div'] as const
 export type FragmentLikeType = typeof AllFragmentLikeTypes[number] // <- this gives us the union type of the Array's entries
 
+export interface GetElementFragmentLikeTypeOptions {
+  sizelessDivsConsideredFragmentLike: boolean
+}
```
Rheeseyb, line 176: Are we likely to add other options here? cc @balazsbajorics

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/fragment-like-helpers.ts
```diff
@@ -171,11 +171,20 @@ export const AllFragmentLikeNonDomElementTypes = ['fragment', 'conditional'] as
 export const AllFragmentLikeTypes = [...AllFragmentLikeNonDomElementTypes, 'sizeless-div'] as const
 export type FragmentLikeType = typeof AllFragmentLikeTypes[number] // <- this gives us the union type of the Array's entries
 
+export interface GetElementFragmentLikeTypeOptions {
+  sizelessDivsConsideredFragmentLike: boolean
```
seanparsons, line 175: We've been generally trying to avoid boolean fields, instead using unions to make them clearer.
```suggestion
  howToTreatSizelessDivs: 'like-fragments' | 'no-special-handling'
```
Example above likely terribly named because naming things is hard.

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/fragment-like-helpers.ts
Rheeseyb: This is an object for passing options through the chain of function calls, so it will appear like: `someFn({ sizelessDivsConsideredFragmentLike: true })`, which achieves the same goal as the string unions

editor/src/components/canvas/canvas-strategies/strategies/fragment-like-helpers.ts
balazsbajorics: we don't have anything planned but this work hasn't started in earnest, so no idea

Rheeseyb: In that case might I suggest we just use a string union param rather than construct a new object every time we call this function. If we do later want to add more options then fair enough, but right now it seems very wrong to introduce an options object for a single option.

bkrmendy: agreed, fixed in https://github.com/concrete-utopia/utopia/pull/3909/commits/e18e09620f46d7c95e8ff87e45d4596f764596be

PR #3917, opened 2023-07-10
Keep track of reparented element paths

bkrmendy:
## Problems
Duplicating elements that are in the root of a conditional branch is broken in two ways:
- The shortcuts don't select the duplicated element(s)
- Alt + drag doesn't work at all

The root cause of these problems is the same: if an element is inserted into a conditional branch next to its root element, the elements are wrapped in a fragment, which results in a different element path compared to what was calculated in the calling code, before the actual duplication was run.

## Fix
To avoid having to duplicate the logic that finds out the new element path, `reparentedToPaths` this PR converts `reparentedToPaths` into a lookup, which maps the "naively" calculated path of an element (which consists of the target parent path with the duplicated element's UID appended, not taking into account the fragment wrapping) to its actual new element path. When duplicating the element, the inserted element's new path can be looked up based on the naive path.

### TODO
- [ ] Better name for `reparentedToPaths`
- [ ] Move `reparentedToPaths` out from under `EditorStateCanvas.controls`
- [x] Reset `reparentedToPaths` after running the duplicate shortcut
- [x] test for the bugs fixed


editor/src/core/model/element-template-utils.ts
```diff
@@ -636,8 +655,14 @@ export function insertJSXElementChildren(
             projectContents,
             elementsToInsert.map((child) => child.uid),
           )
-          insertedChildrenPaths = elementsToInsert.map((child) =>
-            EP.appendToPath(EP.appendToPath(parentPath, newFragmentUid), child.uid),
+
+          insertedChildrenPaths = elementsToInsert.reduce(
```
seanparsons, line 659: Looks like it would be worth extracting out this logic as it's repeated 3 times.

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.tsx
```diff
@@ -76,36 +78,57 @@ export function absoluteDuplicateStrategy(
...
-            setElementsToRerenderCommand([...selectedElements, ...newPaths]),
-            updateSelectedViews('always', selectedElements),
-            updateFunctionCommand('always', (editorState, commandLifecycle) =>
-              runMoveStrategy(
-                canvasState,
-                editorState,
+            updateFunctionCommand('always', (state, commandLifecycle) => {
+              const selectedUids = selectedElements.map(EP.toUid)
+              const newParentPath = EP.parentPath(
+                Object.values(state.canvas.controls.reparentedToPaths)[0],
```
seanparsons, line 98: I'd recommend making this indexing safe, as it's assuming there's a value in that object.

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1786,27 +1773,73 @@ export function duplicate(
...
+                : getInsertionPathWithWrapWithFragmentBehavior(
+                    targetParentPath,
+                    workingEditorState.projectContents,
+                    workingEditorState.nodeModules.files,
+                    workingEditorState.canvas.openFile?.filename ?? null,
+                    workingEditorState.jsxMetadata,
+                    workingEditorState.elementPathTree,
+                  )
+
+            if (insertionPath == null) {
```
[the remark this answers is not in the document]
bkrmendy: good point, I assigned a (hopefully) helpful message to `detailsOfUpdate` before the early return in 
https://github.com/concrete-utopia/utopia/pull/3917/commits/5ae3f1c8bbcf02c69ff621db756051461e4ea291

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.tsx
bkrmendy: added it in https://github.com/concrete-utopia/utopia/pull/3917/commits/d091d0526c484342a7b5ee4b37499b63603c0ef8

editor/src/core/model/element-template-utils.ts
bkrmendy: The usages are very similar, but I think they are just different enough so that a generalised version wouldn't be much less complex (because the value added in the callback of the reduce is always different)

editor/src/components/canvas/canvas-strategies/strategies/absolute-duplicate-strategy.tsx
```diff
@@ -76,36 +78,59 @@ export function absoluteDuplicateStrategy(
...
           [
             ...duplicateCommands,
-            setElementsToRerenderCommand([...selectedElements, ...newPaths]),
-            updateSelectedViews('always', selectedElements),
-            updateFunctionCommand('always', (editorState, commandLifecycle) =>
-              runMoveStrategy(
-                canvasState,
-                editorState,
+            updateFunctionCommand('always', (state, commandLifecycle) => {
+              const selectedUids = selectedElements.map(EP.toUid)
```
balazsbajorics, line 96: instead of this whole function body, why is it not enough to do something along these lines?

```typescript
const newSelectedViews = selectedViews.map(selectedView => reparentedToPaths[EP.toString(selectedView)])
...
setElementsToRerenderCommand(newSelectedViews)
updateSelectedViews(newSelectedViews)

```

bkrmendy:
a more comprehensive solution is needed

PR #3919, opened 2023-07-10
Prevent focusing on autofocused elements.

seanparsons:
**Problem:**
Currently, it's possible to manually focus an autofocused element, but that has no real useful effect and is potentially confusing.

**Fix:**
For those elements which are auto-focusable, we prevent manual focusing with a change to `isFocusableComponent` to check if they are auto-focusable.

**Commit Details:**
- Added `isAutofocusable` to `MetadataUtils` containing the logic to check if something should be auto-focusable.
- Refactored `isFocusableComponent` to call `isAutofocusable`.
- Removed a couple of `XFromMetadata` functions as they were only being called by a function that retrieved the metadata and then called those functions.


Rheeseyb (changes requested):
Ok this change shouldn't be affecting the styling anywhere, just the double-click-to-focus behaviour. The styling in all of these places is to indicate whether the element is a component defined in the user's project, and therefore the editor provides some way to alter the definition of that component (either by automatically or manually focusing it). So it seems like the `isFocusableComponent` function is the wrong place to make that change (though maybe that should now be renamed?), but rather a new function introduced that is only used in the places where a user can choose to manually focus the component

editor/src/components/editor/component-button.tsx
```diff
@@ -30,19 +30,28 @@ export const ComponentOrInstanceIndicator = React.memo(() => {
     'focusedElementPath',
   )
 
-  const { isComponent, selectedViews } = useEditorState(
+  const autoFocusedPaths = useEditorState(
```
Rheeseyb, line 33: It doesn't look like this component is even used anymore, so it can go in the bin

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1866,7 +1866,32 @@ export const MetadataUtils = {
       return this.findNearestAncestorFlexDirectionChange(elementMap, parentPath)
     }
   },
-  isFocusableComponentFromMetadata(element: ElementInstanceMetadata | null): boolean {
+  isAutofocusable(
```
Rheeseyb, line 1869: Minor, but this should be `isAutoFocused`, because either it has automatically been focused or it hasn't. Technically all parsed components are "auto-focusable", they just need to be wrapped in a `Scene` to trigger that.

Rheeseyb (approved):

editor/src/core/model/element-metadata-utils.ts
seanparsons: What about `isAutofocusableCandidate`? As this is the logic that does the check, as opposed to the one that inspects `autoFocusedPaths`?

Rheeseyb: Hmmm yeah ok that's actually a fair point, as otherwise it would be checking against the list of autofocused paths. In that case I'd say just leave the name as it is then and ignore my suggestion 👍

PR #3925, opened 2023-07-11
floating inspector

lankaukk:
Part 2 of making everything float, the inspector!

Also tweaked the alignment/placement of toolbar, strategy picker, strategy indicator, and canvas error underlay

Screenshot 2023-07-11 at 5 06 56 PM


bkrmendy (approved):

editor/src/components/canvas/controls/elements-outside-visible.spec.browser2.tsx
```diff
@@ -106,7 +106,7 @@ describe('elements outside visible area', () => {
       const indicator = screen.queryByTestId(getIndicatorId(targetPath, ['right']))
       expect(indicator).not.toBeNull()
       expect(indicator?.style.top).toBe('182px')
-      expect(indicator?.style.left).toBe('1920px')
+      expect(indicator?.style.left).toBe('2175px')
```
Rheeseyb, line 109: Bah, it looks like this PR breaks the arrow indicator when the selected element is off-screen to the right

Rheeseyb (changes requested):
We need to take into account the Inspector's (default) width when determining the "centre" of the Canvas (for "Scroll to Element" and pasting to the storyboard)

maltenuhn:
> We need to take into account the Inspector's (default) width 

With both the inspector and navigator now sitting visually _atop_ the canvas, we can simplify and target the center _of_ the canvas, no matter what's rendered on top of it. I've just simulated this in a few permutations, and it works well.

With the code editor also floating on top of the canvas it won't be that easy, but since that'll be draggable and resizable we _might_ want a completely different approach (eg "is there something where the element would be") - but I'd not worry about that until we hit that problem.

PR #3930, opened 2023-07-13
Handle elements with text content coming from expressions

gbalint:
[**Try me**](https://utopia.pizza/p/476cc863-sticky-bovid/?branch_name=fix-leaf-expression-handling)

**Problem:**
By adding code elements to the metadata, I introduced some regressions connected to identifying text-only elements. 

1. For text-only elements we show the text content as navigator label.  To test this we check if the element has children or not. However, when there is an expression in the element (e.g. `{1+1}`), the expression as a standalone code element is the child of the original element, so the original element has a child.
2. We do very similar check to decide whether something is text editable, so elements with expressions are not text editable anymore.

**Fix:**
Instead of checking if the element has children, I check if it has non-expression descendants. When there is an expression in an element I only want to treat it as a text editable element when the expression does not generate other children elements, only text.

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -144,6 +144,7 @@ import {
   conditionalClauseInsertionPath,
   isChildInsertionPath,
 } from '../../components/editor/store/insertion-path'
+import { isFeatureEnabled } from '../../utils/feature-switches'
 
 const ObjectPathImmutable: any = OPI
 
@@ -413,7 +414,7 @@ export const MetadataUtils = {
     }
     const jsxElement = element.element.value
     // to mark something as text-like, we need to make sure it's a leaf in the metadata graph
-    const childrenElementsFromMetadata = MetadataUtils.getChildrenOrdered(
+    const childrenElementsFromMetadata = MetadataUtils.getNonExpressionDescendants(
       metadata,
       pathTree,
       target,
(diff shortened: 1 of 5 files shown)


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2100,6 +2104,38 @@ export const MetadataUtils = {
...
+      return childrenInMetadata
+    }
+
+    return childrenInMetadata.flatMap((c) =>
+      getNonExpressionDescendantsInner(metadata, pathTree, c),
+    )
+  },
+}
+
+function getNonExpressionDescendantsInner(
```
seanparsons, line 2123: The naming of this _seems_ wrong given that it's checking `isJSExpression` and including things if that is true...

seanparsons: Ah, it _skips_ the expressions and directly gets their children.

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2100,6 +2104,38 @@ export const MetadataUtils = {
...
+    metadata: ElementInstanceMetadataMap,
+    pathTree: ElementPathTrees,
+    target: ElementPath,
+  ): Array<ElementInstanceMetadata> {
+    const childrenInMetadata = MetadataUtils.getChildrenOrdered(metadata, pathTree, target)
+    if (!isFeatureEnabled('Code in navigator')) {
+      return childrenInMetadata
+    }
+
+    return childrenInMetadata.flatMap((c) =>
```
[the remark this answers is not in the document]
gbalint: I think large hierarchies don't really exist, because expressions can not really have expression children (because then they are just parsed as a single expression), and we stop the recursion when we reach a non-expression.

seanparsons (approved):

PR #3932, opened 2023-07-13
Feature/cmd g wrap in group

balazsbajorics:
This PR restores the original behavior of CMD + G: wrap in Group! (also in the context menu)

**Details:**
- For the sake of simplicity, wrap in group is achieved by using the old WRAP_IN_ELEMENT action
- WRAP_IN_ELEMENT now adds the wrapped elements to the array `trueUpGroupsForElementAfterDomWalkerRunsResult`
- I took the breaking CMD+G "wrap in fragment" tests and changed them so they are testing wrap in group

I removed Wrap in Fragment, in hindsight this was probably too dramatic and the two options should coextist. @maltenuhn any thoughts on this?

**Note:**
I realized it doesn't work correctly for Grouping non position: absolute elements. I've updated the PR to show a toast saying "Only `position: absolute` elements can be grouped for now. 🙇"


ruggi (approved):

editor/src/components/canvas/canvas-context-menu.spec.browser2.tsx
```diff
@@ -1060,37 +1062,81 @@ describe('canvas context menu', () => {
...
+      const elementBounds = element.getBoundingClientRect()
+      await openContextMenuAndClickOnItem(
+        renderResult,
+        canvasControlsLayer,
+        elementBounds,
+        'Group Selection',
+      )
+      await renderResult.getDispatchFollowUpActionsFinished()
+
+      expect(renderResult.getEditorState().editor.toasts.length).toBe(1)
```
seanparsons, line 1136: Minor: Surely just checking that the array is equal to an array containing the text is simpler?

editor/src/components/context-menu-items.ts
```diff
@@ -393,7 +387,7 @@ export const unwrap: ContextMenuItem<CanvasData> = {
 
 export const wrapInPicker: ContextMenuItem<CanvasData> = {
   name: 'Wrap in…',
-  shortcut: 'G',
+  shortcut: 'W',
```
seanparsons, line 390: Is this shortcut just wrong then I take it?

seanparsons (approved):

editor/src/components/context-menu-items.ts
balazsbajorics: does not require _cmd_ to be held!

PR #3933, opened 2023-07-13
True up groups on reparent

seanparsons:
**Problem:**
When reparenting an element that is part of a group on the canvas, the group containing the element isn't true'd up.

**Fix:**
When calling `reparentElement`, also invoke `getRequiredGroupTrueUps` which adds the necessary commands when needed.

**Commit Details:**
- Implemented `getRequiredGroupTrueUps` which adds `QueueGroupTrueUp` commands
  when an element being reparented is part of a group.
- `getReparentOutcome` now includes a call to `getRequiredGroupTrueUps`.

### DIFF
--- a/editor/src/components/canvas/commands/queue-group-true-up-command.ts
+++ b/editor/src/components/canvas/commands/queue-group-true-up-command.ts
@@ -35,7 +35,7 @@ export const runQueueGroupTrueUp: CommandFunction<QueueGroupTrueUp> = (
 }
 
 // If the target is in a group, then this will add a command for including the siblings in a trueing up.
-export function addPossibleTrueUp(
+export function getRequiredGroupTrueUps(
   metadata: ElementInstanceMetadataMap,
   pathTrees: ElementPathTrees,
   target: ElementPath,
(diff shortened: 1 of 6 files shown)


editor/src/components/canvas/commands/queue-group-true-up-command.ts
```diff
@@ -29,3 +33,18 @@ export const runQueueGroupTrueUp: CommandFunction<QueueGroupTrueUp> = (
...
+// If the target is in a group, then this will add a command for including the siblings in a trueing up.
+export function getRequiredGroupTrueUps(
+  metadata: ElementInstanceMetadataMap,
+  pathTrees: ElementPathTrees,
+  target: ElementPath,
+): Array<QueueGroupTrueUp> {
+  const parentPath = EP.parentPath(target)
+  if (treatElementAsGroupLike(metadata, pathTrees, parentPath)) {
+    const siblings = MetadataUtils.getSiblingsOrdered(metadata, pathTrees, target)
+    return siblings.map((sibling) => queueGroupTrueUp(sibling.elementPath))
```
balazsbajorics, line 46: since we already have access to the siblings here, we could instead create a `pushIntendedBoundsAndUpdateGroups` command, with the intendedBounds array containing all non-reparented siblings and their globalFrames. 
queueGroupTrueUp uses trueUpGroupsForElementAfterDomWalkerRunsResult which has the performance penalty,
pushIntendedBoundsAndUpdateGroups is "free" because it uses the frames provided to the command and does not need to wait for a dom-walker pass.

balazsbajorics (approved):

PR #3934, opened 2023-07-13
Handle carrying over UID for a commented out expression.

seanparsons:
**Problem:**
If an entire expression is commented out in the code, an error is raised about an invalid mapping coming from the parse.

**Cause:**
When attempting to carry over the UID from the element to the expression that represents the commented out block, the logic that updates the highlight bounds discovers that there's no highlight bounds for that expression. This is as a result of the construction of the highlight bounds for that expression returning a `null` and not adding one, which is itself a result of an array of expressions that are all null.

**Fix:**
For these, a fallback case has been added as `ExpressionAndText` itself does contain some bounds, so that if the nodes do not provide a useful bounds we construct it from those bounds instead.

**Commit Details:**
- Implemented `buildHighlightBoundsForExpressionsAndText` which caters for `ExpressionAndText` values using those to build bounds as a fallback.

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer-parsing.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer-parsing.ts
@@ -1928,7 +1928,7 @@ function buildHighlightBoundsForExpressionsAndText(
     let highestEnd: number | null = null
     for (const expression of expressions) {
       lowestStart = Math.min(lowestStart ?? expression.startPos, expression.startPos)
-      highestEnd = Math.min(highestEnd ?? expression.endPos, expression.endPos)
+      highestEnd = Math.max(highestEnd ?? expression.endPos, expression.endPos)
     }
     if (lowestStart == null || highestEnd == null) {
       // In this case fail outright as the bounds cannot be produced from an empty array.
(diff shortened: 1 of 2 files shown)


editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1915,6 +1911,45 @@ function buildHighlightBounds(
...
+    mapDropNulls((e) => e.expression, expressions),
+    uid,
+  )
+  if (baseHighlightBounds == null) {
+    // Find the bounds of the bounds.
+    let lowestStart: number | null = null
+    let highestEnd: number | null = null
+    for (const expression of expressions) {
+      lowestStart = Math.min(lowestStart ?? expression.startPos, expression.startPos)
+      highestEnd = Math.min(highestEnd ?? expression.endPos, expression.endPos)
```
balazsbajorics, line 1931: shouldn't this be Math.max to get the "highest" in highestEnd?

seanparsons: It definitely should, copy-paste considered dangerous...Fixed in d71a88628.

Rheeseyb (approved):

balazsbajorics (approved):

PR #3936, opened 2023-07-14
Make navigator reparent work with groups

### LINKED ISSUE #3935: make navigator reparent work with groups

ruggi:
Fixes #3935 

**Problem:**

Navigator reparenting does not true up groups.

**Fix:**

When reparenting, make sure to take into account groups and:
1. adjust the reposition coordinates accordingly
2. true up groups accordingly


seanparsons (approved):

enidemi (approved):

editor/src/core/shared/math-utils.ts
```diff
@@ -121,6 +121,12 @@ export function isFiniteRectangle<C extends CoordinateMarker>(
   return !isInfinityRectangle(r)
 }
 
+export function isNotNullFiniteRectangle<C extends CoordinateMarker>(
```
balazsbajorics, line 124: thanks!

editor/src/components/navigator/navigator.spec.browser2.tsx
```diff
@@ -3855,3 +3864,522 @@ describe('Navigator labels', () => {
...
+          height: canvasRect.height,
+        }),
+      )
+
+      await renderResult.dispatch(
+        [reorderComponents([dragmePath], rootPath, front(), canvasPoint(canvasCenter))],
+        true,
+      )
+
+      expect(getPrintedUiJsCode(renderResult.getEditorState())).toEqual(
```
balazsbajorics, line 4321: one idea: you could also assert that the globalFrame of the reparented element stays the same (as that is the goal of the behavior)

ruggi: makes sense!

ruggi: https://github.com/concrete-utopia/utopia/pull/3936/commits/53596164aaaaed31e3743832a2ab0e3c21583825

editor/src/components/editor/actions/actions.tsx
```diff
@@ -5607,3 +5623,96 @@ export function insertWithReparentStrategies(
...
+
+    const elementToInsertFrame = element?.globalFrame ?? null
+    if (elementToInsertFrame == null || !isFiniteRectangle(elementToInsertFrame)) {
+      // this should never happen
+      return elementToInsert.intendedCoordinates
+    }
+
+    if (reparentTargetParentIsGroup) {
+      const groupFrame =
+        MetadataUtils.findElementByElementPath(jsxMetadata, reparentTargetPath)?.globalFrame ?? null
```
balazsbajorics, line 5661: instead of the globalFrame, we should be using `specialSizeMeasurements.globalContentBoxForChildren`

PR #3938, opened 2023-07-14
Add "group-like" case for navigator icons.

seanparsons:
**Problem:**
Group-like elements are not specifically called out in the navigator, they're just defaulting to showing the component icon.

**Fix:**
Now ahead of all other cases, we check `treatElementAsGroupLike` and display a different icon if that returns true.

**Commit Details:**
- Added `treatElementAsGroupLike` check to `createLayoutOrElementIconResult`.


ruggi (approved):

maltenuhn (approved):
Verily, behold how it groupeth in a most group-like manner:

image

PR #3939, opened 2023-07-14
(Spike) Prevent CanvasStrategies from running on code elements

Rheeseyb:


seanparsons (approved):
Could we have a test that exercises this check please?

Rheeseyb:
> Could we have a test that exercises this check please?

Anything is possible when you put your mind to it! But no. Ok fine

gbalint (approved):

PR #3942, opened 2023-07-17
Fix navigator tests with Code in navigator FS on

gbalint:
**Description:**
Another small PR to prepare for deleting the `Code in navigator` feature switch.
Navigator browser tests fixed with `Code in navigator` feature switch on.

### DIFF
--- a/editor/src/core/shared/element-path-tree.spec.browser2.tsx
+++ b/editor/src/core/shared/element-path-tree.spec.browser2.tsx
@@ -156,4 +156,27 @@ describe('getChildrenOrdered', () => {
       'sb/sc/app:app-root/frag/cond-1/cond-1-true/cond-2/cond-2-child~~~1',
     ])
   })
+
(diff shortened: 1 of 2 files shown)


editor/src/core/shared/element-path-tree.spec.browser2.tsx
```diff
@@ -136,6 +137,9 @@ describe('Building and ordering the element path tree for a real project', () =>
   })
 })
 describe('getChildrenOrdered', () => {
+  // This test only makes sense with the code in navigator FS off, otherwise the expression is the
```
Rheeseyb, line 140: In that case I'd say it's worth adding a `Returns an expression child` test case that only runs with the FS on (unless we already have one)

gbalint: Done in 3a73678682918a4797f06f0b2a0058638209e804

Rheeseyb (approved):

PR #3945, opened 2023-07-17
Prevent contract dropdown from changing root elements.

seanparsons:
**Problem:**
Currently if the contract dropdown is operated for a root element, it will cause an error and make some unexpected changes to the project.

**Cause:**
This was as a result of the strategy removing the root element, but not correctly replacing it in the original location.

**Fix:**
Now when the contract dropdown is changed, the event handler checks if anything is the root element of a component and prevents the change displaying a toast warning the user in the process.

**Commit Details:**
- `EditorContractDropdown.onChange` checks to see if any of the selected elements are root elements of components and if so, displays a warning and prevents the change from occurring.


balazsbajorics (approved):

gbalint:
I think it would make sense to add a new test case to `editor-contract-section.spec.browser2.tsx`

Rheeseyb (approved):

seanparsons:
> I think it would make sense to add a new test case to `editor-contract-section.spec.browser2.tsx`

I agree, added one in 977ac901e.

PR #3946, opened 2023-07-17
True up groups on paste

### LINKED ISSUE #3943: True up groups on post-action paste

ruggi:
Fixes #3943 

**Problem:**

Groups are not being true'd up on post-action paste.

**Fix:**

Add the pasted paths to the ones to true up if the target element is a group.
Positioning does not need to be adjusted as it's already being forced to the center of the target element, and cut already also trues them up.

**Notes:**

I believe this should be in an incremental diff, but - paste will generate multiple undo steps due to the trueing up, which I don't think is correct/ideal.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/post-action-options/post-action-paste.ts
+++ b/editor/src/components/canvas/canvas-strategies/post-action-options/post-action-paste.ts
@@ -35,6 +35,7 @@ import { showToastCommand } from '../../commands/show-toast-command'
 import { updateFunctionCommand } from '../../commands/update-function-command'
 import { updateSelectedViews } from '../../commands/update-selected-views-command'
 import { wildcardPatch } from '../../commands/wildcard-patch-command'
+import { treatElementAsGroupLike } from '../strategies/group-helpers'
 import {
   absolutePositionForPaste,
   offsetPositionInPasteBoundingBox,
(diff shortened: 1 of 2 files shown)


editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -335,7 +335,7 @@ describe('actions', () => {
       elements: (renderResult: EditorRenderResult) => Array<ElementPaste>
       pasteInto: InsertionPath
       want: string
-      generatesUndoStep?: boolean
+      generatesUndoStep?: number
```
seanparsons, line 338: Minor:
```suggestion
      generatesUndoSteps?: number
```

seanparsons (approved):

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -1302,7 +1333,7 @@ describe('actions', () => {
 
         const canvasRoot = renderResult.renderedDOM.getByTestId('canvas-root')
 
-        await undoCheckerFn(renderResult, async () => {
+        await expectSingleUndoNSaves(renderResult, test.generatesUndoSteps ?? 2, async () => {
```
balazsbajorics, line 1336: it's a single _undo step_ and N saves! so the field should be renamed `test.generatesSaves` or `test.generatesSaveCount` or `test.triggeredSaveNTimes`

editor/src/components/canvas/canvas-strategies/post-action-options/post-action-paste.ts
```diff
@@ -233,6 +249,11 @@ function pasteChoiceCommon(
         },
       },
     }),
+    wildcardPatch('on-complete', {
+      trueUpGroupsForElementAfterDomWalkerRuns: {
+        $set: groupTrueUpPaths,
+      },
+    }),
```
balazsbajorics, line 256: this shall either be a `groupTrueUpPaths.map(p => queueGroupTrueUp(p))`, or we should change `queueGroupTrueUp` to accept an array of ElementPath

ruggi: oh! nice catch, I wanted to be consistent with the rest of the file but it's much better. Went with the former option ;) https://github.com/concrete-utopia/utopia/pull/3946/commits/933d363588fa0b3e33692c9f5c1c7df2c251fd09

balazsbajorics (approved):

PR #3948, opened 2023-07-18
Feature/paste to replace post action menu

enidemi:
https://utopia.pizza/p/30fc3aa5-charmed-foe/?branch_name=feature/paste-to-replace-post-action

**Problem:**
Paste to Replace feature is not using the Post Action menu yet.

**Fix:**
Remove "Paste to Replace" action and add it to the paste strategies.

**Commit Details:**
- added Paste To Replace options to the post action menu and paste strategies
- update context menu and global shortcuts to create post action session
- extended tests to check selected views after paste, except where a new fragment is inserted into a conditional because here the uid-s are inconsistent between test runs
- added one more test to paste to replace where multiselected elements are replaced
- minor fix in context menu for missing icon warnings for elements

### DIFF
--- a/editor/src/components/editor/global-shortcuts.tsx
+++ b/editor/src/components/editor/global-shortcuts.tsx
@@ -103,7 +103,11 @@ import {
   OPEN_INSERT_MENU,
   PASTE_TO_REPLACE,
 } from './shortcut-definitions'
-import type { EditorState, LockedElements } from './store/editor-state'
+import type {
+  EditorState,
+  LockedElements,
+  PasteToReplacePostActionMenuData,
+} from './store/editor-state'
 import { DerivedState, getOpenFile, RightMenuTab } from './store/editor-state'
 import { CanvasMousePositionRaw, WindowMousePositionRaw } from '../../utils/global-positions'
 import { pickColorWithEyeDropper } from '../canvas/canvas-utils'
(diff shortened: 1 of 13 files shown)


ruggi (approved):
\o\

editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -885,11 +893,26 @@ export function handleKeyDown(
...
+          ]).at(0)
+
+          if (defaultChoice != null) {
+            return [
+              EditorActions.startPostActionSession(pasteToReplacePostActionMenuData),
+              EditorActions.executePostActionMenuChoice(defaultChoice),
+            ]
+          }
+        }
+        return []
```
balazsbajorics, line 915: should this be a shared function between this shortcut and the context menu? I created a similar shared function called `createWrapInGroupAction` because it was no longer just dispatching a single EditorAction

balazsbajorics (approved):

editor/src/components/editor/global-shortcuts.tsx
enidemi: good idea! fixed in [6043360](https://github.com/concrete-utopia/utopia/pull/3948/commits/604336085919131d4680d5884ff116844bab0dd7)

PR #3950, opened 2023-07-18
Invalid groups warnings

### LINKED ISSUE #3949: Show warning in navigator for invalid groups

ruggi:
Fixes #3949 

**Problem:**

Groups with "invalid" children configurations are not highlighted as such in the navigator.

**Fix:**

Show a warning in the navigator for group entries if:
- any of its children is not `position: absolute`
- any of its children has percentage pins _and_ the group does not have explicit `width` and `height`

The `getGroupState` function returns a `GroupChildState` which refers to the first invalid child of the group (or `valid` if all are ok).

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -1516,16 +1516,21 @@ export const MetadataUtils = {
     metadata: ElementInstanceMetadataMap,
     path: ElementPath,
   ): JSXElement | null {
-    const element = MetadataUtils.findElementByElementPath(metadata, path)
+    return this.getJSXElementFromElementInstanceMetadata(
+      MetadataUtils.findElementByElementPath(metadata, path),
+    )
+  },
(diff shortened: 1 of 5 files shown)


enidemi (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1516,16 +1516,21 @@ export const MetadataUtils = {
     metadata: ElementInstanceMetadataMap,
     path: ElementPath,
   ): JSXElement | null {
-    const element = MetadataUtils.findElementByElementPath(metadata, path)
+    return this.getJSXElementFromElementInstanceMetadata(
```
Rheeseyb, line 1519: I know we have used it in a few places in this file, but please do avoid using `this` here
```suggestion
    return MetadataUtils.getJSXElementFromElementInstanceMetadata(
```

Rheeseyb (approved):

balazsbajorics (approved):

PR #3953, opened 2023-07-18
Fix/paste to replace on root divs

enidemi:
**Problem:**
When Paste to Replace is called on a root element it is inserted directly as a Component child. Root element insertion is not supported yet.

**Fix:**
Disable contextmenu entry "paste as value" when a root div is selected. Show toast when "Paste to Replace" is fired on selection when there is at least one root element. Toast is shown when the shortcut is used on a root div.

**Commit Details:**
- paste strategy filters selection for root divs
- show toast if there are root divs
- contextmenu is disabled on roots

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/post-action-options/post-action-paste.ts
+++ b/editor/src/components/canvas/canvas-strategies/post-action-options/post-action-paste.ts
@@ -587,14 +587,14 @@ export const PasteToReplaceWithPropsReplacedPostActionChoice = (
 function pasteToReplaceCommands(
   editor: EditorState,
   builtInDependencies: BuiltInDependencies,
-  targets: Array<ElementPath>,
+  unfilteredTargets: Array<ElementPath>,
   elementToPaste: Array<ElementPaste>,
   originalMetadata: ElementInstanceMetadataMap,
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-strategies/post-action-options/post-action-paste.ts
```diff
@@ -592,7 +592,9 @@ function pasteToReplaceCommands(
   originalMetadata: ElementInstanceMetadataMap,
   originalPathTree: ElementPathTrees,
 ): Array<CanvasCommand> {
-  const pasteCommands = targets.flatMap((target) => {
+  const targetsWithoutRoot = targets.filter((target) => !EP.isRootElementOfInstance(target))
```
Rheeseyb, line 595: It might be safer here to rename the function's input `targets` input param to `unfilteredTargets`. I'm thinking that if someone comes in here later and generates another type of command for each of the targets, the naming of that param would signify that they probably don't want to blindly use it (especially since such a change would be unlikely to show this filtering line in a PR diff)

Rheeseyb (approved):

seanparsons (approved):

PR #3957, opened 2023-07-19
icons/errored-elements

maltenuhn:
**Problem:**
For groups and in general, we didn't have good ways to indicate error states with icons

**Fix:**
- Introduced `group-ded` and `group-problematic` assets: group-ded-blue-18x18@2x and group-problematic-blue-18x18@2x
- Introduced red colourway for other element icons, eg image-red-18x18@2x
- Introduced single exclamation mark decorator: exclamationmark-blue-3x8@2x

NB: this PR does not add any of these assets to any of the `Icons.` objects.


ruggi (approved):
❗

PR #3958, opened 2023-07-19
Grouping for flex children too

balazsbajorics:


editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -607,24 +632,237 @@ export function groupConversionCommands(
...
+      if (
+        foundMetadata == null ||
+        element == null ||
+        isLeft(element) ||
+        !isJSXElement(element.value)
+      ) {
+        throw new Error(
+          `Invariant violation: ElementInstanceMetadata.element found for ${EP.toString(
+            p,
+          )} was null or Left or not JSXElement`,
```
seanparsons, line 707: We really should get rid of that Either for these...

editor/src/core/model/element-template-utils.spec.tsx
```diff
@@ -98,13 +98,13 @@ describe('guaranteeUniqueUids', () => {
     ]
     const fixedElements = guaranteeUniqueUids(exampleElements, emptySet())
 
-    const child0PropsOptic = fromArrayIndex<JSXElementChild>(0)
+    const child0PropsOptic = fromArrayIndex<JSXElementChild | null>(0)
```
seanparsons, line 101: With this then we don't need to change `isJSXElement`, which can also cause issues when composing it when the value isn't null.
```suggestion
    const child0PropsOptic = fromArrayIndex<JSXElementChild>(0).compose(notNull())
```

seanparsons (changes requested):
Looks really nice, but I don't think we should change `isJSXElement`. I had to remove some of those null checks for the optics to work nicely in the first place.

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -607,24 +632,237 @@ export function groupConversionCommands(
...
+
+  // TODO retarget Fragments
+  const result = [
+    setCssLengthProperty(
+      'always',
+      target,
+      PP.create('style', 'left'),
+      setValueKeepingOriginalUnit(localFrame.x, parentSize.width),
+      parentFlexDirection,
+      mustHavePins.includes('left') ? 'create-if-not-existing' : 'do-not-create-if-doesnt-exist',
```
ruggi, line 818: my suggestion would be to make a small generator function for these, e.g.
```ts
function setPinCssProperty(pin: 'top' | 'left' | 'right' | ' bottom', x: number, y: number)
```
and return four of those directly

ruggi (commented):
lgtm! I do agree with Sean's point about `isJSXElement` though, it would be good to revert that before merging

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
balazsbajorics: fixed!

enidemi (approved):

seanparsons (approved):

PR #3959, opened 2023-07-19
Include worker update results in undo stack

Rheeseyb:
**Problem:**
Worker updates are treated as transient, and therefore are never included in the undo history, which can in turn result in the code editor falling out of sync with the canvas when undoing under certain conditions

**Fix:**
Since we want the worker updates to remain transient _but_ be included in the undo history, we now wrap them in a `mergeWithPrevUndo`. 

This in turn highlighted that we were incorrectly treating all `MERGE_WITH_PREV_UNDO` actions as non-transient, whereas we should be determining that based on their wrapped actions, so I have fixed that as part of this, and corrected the order of `if` `else` block used when pushing to the undo stack.

Some grouping tests were broken by this change, because the tests were expecting the `TRUE_UP_GROUP` to be treated as non-transient, even though the action itself was marked as transient. I've fixed this (by marking the action as non-transient), though it sounds like in an upcoming PR that will become irrelevant anyway.

### DIFF
--- a/editor/src/components/canvas/dom-walker-caching.spec.browser2.tsx
+++ b/editor/src/components/canvas/dom-walker-caching.spec.browser2.tsx
@@ -101,11 +101,11 @@ describe('Dom-walker Caching', () => {
     expect(renderResult.getRecordedActions().map((a) => a.action)).toEqual([
       'SET_ELEMENTS_TO_RERENDER',
       'SET_CANVAS_FRAMES',
-      'UPDATE_FROM_WORKER',
+      'MERGE_WITH_PREV_UNDO',
       'SAVE_DOM_REPORT',
       'SAVE_DOM_REPORT',
       'SET_CANVAS_FRAMES',
-      'UPDATE_FROM_WORKER',
+      'MERGE_WITH_PREV_UNDO',
       'SAVE_DOM_REPORT',
       'SAVE_DOM_REPORT',
     ])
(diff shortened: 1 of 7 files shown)


enidemi (approved):

editor/src/components/canvas/dom-walker-caching.spec.browser2.tsx
```diff
@@ -101,11 +101,11 @@ describe('Dom-walker Caching', () => {
     expect(renderResult.getRecordedActions().map((a) => a.action)).toEqual([
       'SET_ELEMENTS_TO_RERENDER',
       'SET_CANVAS_FRAMES',
-      'UPDATE_FROM_WORKER',
+      'MERGE_WITH_PREV_UNDO',
```
balazsbajorics, line 104: nitpick, but we should print and assert the wrapped actions here

Rheeseyb: Yeah, that's fair, I'll fix it

editor/src/core/shared/array-utils.ts
```diff
@@ -327,7 +327,7 @@ export function immutablyUpdateArrayIndex<T>(
   return working
 }
 
-export function safeIndex<T>(array: Array<T>, index: number): T | undefined {
+export function safeIndex<T>(array: Array<T> | ReadonlyArray<T>, index: number): T | undefined {
```
balazsbajorics, line 330: I believe this can just be ReadonlyArray here
```suggestion
export function safeIndex<T>(array: ReadonlyArray<T>, index: number): T | undefined {
```

balazsbajorics (approved):

editor/src/components/canvas/dom-walker-caching.spec.browser2.tsx
Rheeseyb: Ok I've used `simpleStringifyActions` here (after moving that function to a more sensible place and tweaking it a bit to give better indentation)

editor/src/core/shared/array-utils.ts
Rheeseyb: Yep, fixed 👍

PR #3960, opened 2023-07-20
more red icons

maltenuhn:
**Problem:**
High demand for red icons means we ran out of some of them.

**Fix:**
Make some more, upload them. Like so:
lockcloseddot-red-16x16@2x
eyeopen-red-16x16@2x
eye-strikethrough-red-16x16@2x
lockclosed-red-16x16@2x
lockopen-red-16x16@2x

dot-red-16x16@2x


maltenuhn:
Merging since assets only.

PR #3961, opened 2023-07-20
Reduce group true up saves.

seanparsons:
**Problem:**
Truing up groups triggers more saves than are necessary, where one save comes from the change that the user has made but then another save originates from the change made in truing up the groups.

**Cause:**
`editorDispatch` is responsible for processing actions and performing a bunch of associated processes off the back of the updated `EditorState` and is fired potentially a total of four times from `boundDispatch`:
- The initial pass for running whatever actions the user has triggered.
- After the DOM walker has run on the result of the first pass, then a `SAVE_DOM_REPORT` is put through it.
- When truing up groups: a `TRUE_UP_GROUPS` action is triggered, set to merge the result with the previous undo entry.
- When truing up groups: After the DOM walker has been run on the result of truing up the groups, another `SAVE_DOM_REPORT` action is triggered.
For each of these undo history handling, potential saves, updates to the evaluation cache, updates to VS Code are all applied.

**Fix:**
Splitting `editorDispatch` into two halves is the solution applied here:
- `editorDispatchActionRunner`: The first runs the actions and additionally potentially clears the evaluation cache.
- `editorDispatchClosingOut`: The second runs the remainder of old implementation, handling the undo history, saving, etc.

With those two functions now `boundDispatch` runs the first function repeatedly where `editorDispatch` is currently used with the latter function run after all of those have been executed. As a result the undo and save handling get consolidated for all the things that are triggered off of a bunch of dispatched actions.

**Why The Performance Regression Results Have Changed:**
The performance regression tests have changed dramatically, but it's actually not as bad as it looks:
- Some key context here is that `workingEditorState` is shared between all the invocations of `asyncTestDispatch`.
- On master we fire that function because of a set prop in this case, then it does an `editorDispatch` on that action which in turns triggers `maybeRequestModelUpdateOnEditor` causing a promise to be queued up with a dispatch of `UPDATE_FROM_WORKER`. In the tests that "worker" is actually inline and not dependent on a real web worker.
- Somewhere after that `editorDispatch` we have a `waitForDispatchEntireUpdate` condition that if true yields the main thread to the promise mentioned earlier.
- In the middle of the call to `asyncTestDispatch` call `n`, `asyncTestDispatch` call `n + 1` is fired and runs to completion, with the key being that it's updating `workingEditorState` as it goes.
- Then `asyncTestDispatch` call n finishes its little wait and continues on, doing whatever with this newly updated `workingEditorState` but likely does a whole lot of nothing as it's all "done" because of the other call.

Whereas in this branch, because the queueing up of the `UPDATE_FROM_WORKER` action happens right at the end it only fires once the initial call to `asyncTestDispatch` has completed. In reality this makes it closer to the implementation in `boundDispatch` as the `UPDATE_FROM_WORKER` normally comes from an async web worker. Plus `boundDispatch` isn't implemented as an `async` function so never yields the main thread for anything else to fire during execution.

**Commit Details:**
- Created the two halves of `editorDispatch` from the original.
- Pulled up `reducerToSplitToActionGroups` to the top level so that it could be reused in both dispatch functions.
- Created `getParsedAfterCodeChanged` from the existing constant so that it could be reused.
- Removed `workerUpdatedModel` as it was just a duplicate.
- Commented a lot of the core save handling conditions.
- Renamed `applyProjectChanges` to `applyProjectChangesToVSCode` and created `applyProjectChangesToEditor` alongside it.
- Removed `mergeWithPrevUndo` wrapper from what was `editorDispatch` as it's unnecessary now.


ruggi (approved):
🤯

editor/src/components/editor/store/dispatch-strategies.spec.tsx
```diff
@@ -50,7 +50,7 @@ import { saveDOMReport, selectComponents, toggleProperty } from '../actions/acti
 import * as History from '../history'
 import { DummyPersistenceMachine } from '../persistence/persistence.test-utils'
 import type { DispatchResult } from './dispatch'
-import { editorDispatch } from './dispatch'
+import { editorDispatchActionRunner } from './dispatch'
```
Rheeseyb, line 53: Looks like this is unused

editor/src/components/editor/shortcuts.spec.browser2.tsx
```diff
@@ -1195,7 +1195,7 @@ describe('group selection', () => {
         EP.fromString(`${BakedInStoryboardUID}/${TestSceneUID}/${TestAppUID}:container/bbb`),
       ])
 
-      await expectSingleUndoNSaves(renderResult, 4, async () =>
+      await expectSingleUndoNSaves(renderResult, 2, async () =>
```
balazsbajorics, line 1198: nitpick, but would you mind replacing these with `expectSingleUndo2Saves`? It would make it easier to mass find how many places are expecting two saves

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -369,13 +369,82 @@ export function resetDispatchGlobals(): void {
   accumulatedProjectChanges = emptyProjectChanges
 }
 
-export function editorDispatch(
+// With this reducer we can split the actions into groups (arrays) which can be dispatched together without rebuilding the derived state.
+// Between the different group derived state rebuild is needed
+function reducerToSplitToActionGroups(
+  actionGroups: EditorAction[][],
+  currentAction: EditorAction,
+  i: number,
```
Rheeseyb, line 377: `i`?

Rheeseyb: Oh I see this entire function was just moved

seanparsons: Taken from the original code, I believe it's the index of the action.

editor/src/components/editor/shortcuts.spec.browser2.tsx
seanparsons: Fixed in d90533219.

editor/src/components/editor/store/dispatch-strategies.spec.tsx
seanparsons: Fixed in d90533219.

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -465,10 +489,38 @@ export function editorDispatch(
...
   const canSave = isLoaded && !isLoadAction && isBrowserEnvironment
+  const changesShouldTriggerSave = editorChangesShouldTriggerSave(
+    storedState.unpatchedEditor,
+    frozenEditorState,
+  )
+  // Should save in a regular unforced situation if:
+  // - Changes have been made which necessitate a save.
+  //   AND
+  // - These conditions are met:
+  //   - There changes and they are not transient.
```
Rheeseyb, line 508:
```suggestion
  //   - There are changes and they are not transient.
```

editor/src/components/canvas/dom-walker-caching.spec.browser2.tsx
```diff
@@ -199,43 +206,49 @@ describe('Dom-walker Caching', () => {
...
       'storyboard-entity/scene-1-entity/app-entity:app-outer-div/card-instance:card-outer-div',
       'storyboard-entity/scene-2-entity/same-file-app-entity:same-file-app-div',
     ])
-    expect(saveDomReportActions[2].cachedPaths).toEqual([])
-
-    expect(saveDomReportActions[3].invalidatedPaths).toEqual(['storyboard-entity/scene-2-entity'])
-    expect(saveDomReportActions[3].cachedPaths).toEqual([
+    expect(saveDomReportActions[2].cachedPaths).toEqual([
       EP.fromString('storyboard-entity/scene-1-entity/app-entity:app-outer-div/card-instance'),
       EP.fromString('storyboard-entity/scene-1-entity/app-entity:app-outer-div'),
```
Rheeseyb, line 241: As mentioned, this path shouldn't appear in both the `cachedPaths` and `invalidatedPaths` for the same dom walking result, so this definitely needs some investigating

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -65,7 +65,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`457`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`770`)
```
balazsbajorics, line 68: one thing I'd call out here: if a component change / user facing feature changes these numbers, that's a potential regression. here we are basically moving the baseline. we are very certain that the actual performance did not change, just the way we count this number

balazsbajorics (changes requested):
we should either prevent `this.storedState` to be in the namespace of boundDispatch, or we should go back to always updating `this.storedState` during the inner steps of the dispatch flow.

otherwise looks good!

editor/src/core/performance/performance-regression-tests.spec.tsx
Rheeseyb: I'd argue that any change to these numbers is a potential regression. So especially if we're changing something that _isn't_ a UI change we need to be sure that the updated figures are not a regression

seanparsons: It's also the baseline _with the test dispatch_, which in this case is a key part of the difference.

seanparsons:
> we should either prevent `this.storedState` to be in the namespace of boundDispatch, or we should go back to always updating `this.storedState` during the inner steps of the dispatch flow.
> 
> otherwise looks good!

Fixed in b449cd0ac.

balazsbajorics (approved):

editor/src/components/canvas/dom-walker-caching.spec.browser2.tsx
seanparsons: When a selective re-render is triggered, it can end up having paths that differ to what ends up in `invalidatedPaths`, which when results in the unchanged elements ending up also in `cachedPaths`.

PR #3963, opened 2023-07-20
Add groups to Wrap In… menu, but not for insertion

### LINKED ISSUE #3962: Include groups in wrap in menu, but not for insert

ruggi:
Fixes #3962

**Problem:**

`Cmd+G` already wraps elements in a group, but the same is currently not present in the Wrap In floating menu.

**Fix:**

1. Add a new entry for Groups in the insertable components
2. Add a filter for insertable components that checks suitability with the current insert mode (`insert`, `wrap`, ...), with the new `GROUPS_GROUP` matching only the `wrap` mode.

### DIFF
--- a/editor/src/components/canvas/ui/floating-insert-menu-helpers.ts
+++ b/editor/src/components/canvas/ui/floating-insert-menu-helpers.ts
@@ -1,0 +1,15 @@
+export type InsertMenuMode = 'closed' | 'insert' | 'convert' | 'wrap'
+
+export const insertMenuModes: {
+  all: InsertMenuMode[]
+  onlyClosed: InsertMenuMode[]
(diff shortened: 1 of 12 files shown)


seanparsons (approved):

editor/src/components/canvas/ui/floating-insert-menu-helpers.ts
```diff
@@ -0,0 +1,15 @@
+export type InsertMenuMode = 'closed' | 'insert' | 'convert' | 'wrap'
+
+export const insertMenuModes: {
```
Rheeseyb, line 3: I suspect this approach might be overkill, but if that's the case it wouldn't take much to change the code later anyway

Rheeseyb (approved):

PR #3969, opened 2023-07-21
Support highlight bounds on code elements

Rheeseyb:
**Problem:**
Now that we support code elements in the Navigator, we should also support highlight bounds for them

**Fix:**
The fix itself was a simple 1 liner, and that was to call `includeElement` for arbitrary JS expressions in `parser-printer.ts`. In doing so though I discovered some issues with our fast-check tests that are checking the highlight bounds contain all relevant UIDs (these are incidentally the tests that fail periodically on CI).

The problem with the fast-check tests is that, whilst we stop traversing the tree at certain types of elements when capturing the highlight bounds, we don't apply the same filter when running those tests (instead we just check at the point of testing if the element is one we are interested in, but continue walking the tree either way).

To fix the fast-check tests I've now added a filter function to decide if we want to continue walking or not, and a boolean flag to state whether we want to walk attributes of an element (which we also don't capture highlight bounds for right now). I've also removed a duplicate function here (it seems two different functions converged at some point).

Finally, since there was no reason not to (and we're checking it in the tests), I have included the highlight bounds for `element.condition` for conditional elements.

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer.test-utils.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer.test-utils.ts
@@ -210,7 +210,7 @@ export function testParseCode(
     let uids: Array<string> = []
     fastForEach(success.topLevelElements, (topLevelElement) => {
       if (isUtopiaJSXComponent(topLevelElement)) {
-        ensureElementsHaveUID(topLevelElement.rootElement, uids)
+        ensureElementsHaveUID(topLevelElement.rootElement, uids, () => true, true)
         ensureArbitraryJSXBlockCodeHasUIDs(topLevelElement.rootElement)
       }
     })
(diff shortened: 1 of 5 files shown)


gbalint (approved):

editor/src/core/workers/parser-printer/parser-printer.test-utils.ts
```diff
@@ -881,56 +887,131 @@ function walkElementsWithin(
   elementsWithin: ElementsWithin,
   includeDataUIDAttribute: IncludeDataUIDAttribute,
   walkWith: (elem: JSXElementChild) => void,
+  shouldWalkElement: (elem: JSXElementChild) => boolean,
+  shouldWalkAttributes: boolean,
```
seanparsons, line 891: This should probably be a union type like `'walk-attributes' | 'do-not-walk-attributes'`.

seanparsons (approved):

balazsbajorics (approved):

editor/src/core/workers/parser-printer/parser-printer.test-utils.ts
Rheeseyb: Good shout! Done 👍

PR #3970, opened 2023-07-24
Fix/fragments in groups

balazsbajorics:
**Problem**
1. Grouping is not allowed if one or more selected element is a Fragment. 
2. If the Group's parent is a Fragment, the Group will always become position: absolute, even if the fragment's parent is Flex, giving a flex layout to the children of the fragment.
3. If a Group has a Fragment child which has a position: absolute child, moving/resizing the group's grandchild should result in proper true-up of the Group's position and size.

**Fix**
An assortment of fixes to allow the above behaviors, and new test cases to cover them too.


editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -715,8 +718,17 @@ export function createWrapInGroupActions(
   // delete all reparented elements first to avoid UID clashes
   const deleteCommands = orderedActionTargets.map((e) => deleteElement('always', e))
 
+  // TODO this is horrible and temporary at best. Instead of this, we should fix layoutSystemForChildren for Fragments in fillGlobalContentBoxFromAncestors
```
seanparsons, line 721: Do we have a task for this?

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -773,8 +785,42 @@ export function createWrapInGroupActions(
...
+    element: JSXElement
+    metadata: ElementInstanceMetadata
+    expectedElementPath: ElementPath
+  }> = orderedActionTargets.flatMap((p) => {
+    return replaceFragmentLikePathsWithTheirChildrenRecursive(
+      metadata,
+      allElementProps,
+      elementPathTrees,
+      [p],
+    ).flatMap((maybeRetargetedPath) => {
```
seanparsons, line 798: I think this `flatMap` should just be a regular `map`, I'm also a little wary about the number of `flatMap` calls in here as well if it's on the performance hot path.

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
balazsbajorics: yes!

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
balazsbajorics: fixed!

ruggi (approved):

PR #3971, opened 2023-07-24
Fix/static reparent absolute

enidemi:
**Problem:**
Pasted/reparented elements easily end up as flow elements even in non-flex containers.

**Fix:**
Only paste/reparent as flow when the parent is a flex container or a text element.

**Commit Details:**
- update tests
- change `reparentStrategyForPaste`

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-helpers.ts
@@ -39,7 +39,10 @@ export function reparentStrategyForPaste(
 ): ReparentStrategy {
   const parentIsFlexLayout =
     MetadataUtils.findLayoutSystemForChildren(currentMetadata, pathTrees, parent) === 'flex'
-  return parentIsFlexLayout ? 'REPARENT_AS_STATIC' : 'REPARENT_AS_ABSOLUTE'
+  const isTextFromMetadata = MetadataUtils.isTextFromMetadata(
+    MetadataUtils.findElementByElementPath(currentMetadata, parent),
+  )
+  return parentIsFlexLayout || isTextFromMetadata ? 'REPARENT_AS_STATIC' : 'REPARENT_AS_ABSOLUTE'
 }
 
 export function findReparentStrategies(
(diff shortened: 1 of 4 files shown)


seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-helpers.ts
```diff
@@ -42,15 +39,10 @@ export function reparentStrategyForPaste(
...
-    currentMetadata,
-    allElementProps,
-    pathTrees,
-    parent,
+  const isTextFromMetadata = MetadataUtils.isTextFromMetadata(
+    MetadataUtils.findElementByElementPath(currentMetadata, parent),
   )
-  const reparentAsStatic = parentIsFlexLayout || flowParentReparentType === 'REPARENT_AS_STATIC'
-  return reparentAsStatic ? 'REPARENT_AS_STATIC' : 'REPARENT_AS_ABSOLUTE'
+  return parentIsFlexLayout || isTextFromMetadata ? 'REPARENT_AS_STATIC' : 'REPARENT_AS_ABSOLUTE'
```
Rheeseyb, line 45: I think we'd be better off sharing the logic used by `findReparentStrategies`, as otherwise we run the risk of the two drifting. It looks like you could do that by using this function https://github.com/concrete-utopia/utopia/blob/4752b2ea1427e3572b88adf385e4a3f38415118e/editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-parent-lookup.ts#L549

Rheeseyb (changes requested):
Please can you unify the two approaches used to determine the `ReparentStrategy` (I've left a comment inline)

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-strategy-helpers.ts
enidemi: I'm not sure about this, that function uses the same `flowParentAbsoluteOrStatic` which I just removed from the `reparentStrategyForPaste`. It looks like the canvas uses something else when reparenting, otherwise the elements should end up as non-absolute more frequently.

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -389,14 +389,14 @@ describe('actions', () => {
...
-            <div data-uid='aad'>foo</div>
+            <div data-uid='aad' style={{ top: 0, left: 0, position: 'absolute' }}>foo</div>
         </div>
 		`,
       },
       {
         name: 'multiple elements',
         startingCode: `
-        <div data-uid='aaa'>
+        <div data-uid='aaa' style={{ lineHeight: '20px' }}>
```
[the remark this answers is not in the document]
enidemi: needed for tests only, CI has different default font size

balazsbajorics (approved):

PR #3977, opened 2023-07-25
Inspector safeguards for groups

### LINKED ISSUE #3976: Safeguard against invalid groups via the inspector

ruggi:
Fixes #3976 

**Problem:**

You can create groups with invalid configurations via the inspector.

**Fix:**

We defined a set of common cases that are handled to detect invalid group configurations in the navigator:
```ts
type InvalidGroupState =
  | 'child-not-position-absolute'
  | 'child-has-percentage-pins-without-group-size'
  | 'child-has-missing-pins'
  | 'group-has-percentage-pins'
```

These same rules can be used in the inspector so that if the user tries to apply some changes there that would lead to those issues, we instead prevent the action from going through and display a toast instead.

This PR adds counterpart checks to the inspector strategies that will do that, showing a toast that contains the same error string we use in the navigator tooltips.

https://github.com/concrete-utopia/utopia/assets/1081051/1ecadfd2-8533-4b0f-809b-493b481daa94


editor/src/components/canvas/canvas-strategies/strategies/group-helpers.ts
```diff
@@ -96,38 +107,68 @@ function elementHasPercentagePins(jsxElement: JSXElement): boolean {
...
+  const width = getPin('width')
+  const height = getPin('height')
 
-  if (leftPin || rightPin) {
-    return topPin || bottomPin
+  if (leftPin != null || rightPin != null) {
+    return topPin != null || bottomPin != null || height?.unit === '%'
   } else {
-    return !(topPin || bottomPin)
+    return !(topPin != null || bottomPin != null) || width?.unit === '%'
```
seanparsons, line 124: This is equivalent right?
```suggestion
    return (topPin == null && bottomPin == null) || width?.unit === '%'
```

ruggi: yep but for some reason it was easier on my brain in that form - updating it!

editor/src/components/inspector/inspector.tsx
```diff
@@ -714,6 +720,34 @@ export const InspectorContextProvider = React.memo<{
   const onUnsetValue = React.useCallback(
     (property: PropertyPath | Array<PropertyPath>, transient: boolean) => {
       let actionsArray: Array<EditorAction> = []
+
+      const propTop = PP.create('style', 'top')
+      const propBottom = PP.create('style', 'bottom')
+      const propLeft = PP.create('style', 'left')
+      const propRight = PP.create('style', 'right')
+
+      const maybeInvalidGroupState = maybeInvalidGroupStates(
```
seanparsons, line 729: Minor: I found it a little confusing that these are so similarly named.

editor/src/components/inspector/inspector-strategies/fill-container-basic-strategy.ts
```diff
@@ -21,13 +21,18 @@ import {
...
+  maybeGroupChildWithoutFixedSizeForFill,
+  maybeInvalidGroupStates,
+} from '../../canvas/canvas-strategies/strategies/group-helpers'
 
 export const fillContainerStrategyFlow = (
   axis: Axis,
   value: 'default' | number,
   otherAxisSetToFill: boolean,
 ): InspectorStrategy => ({
-  name: 'Set tp Fill Container',
```
seanparsons, line 30: Hah, how'd that get in there? :)

editor/src/components/inspector/inspector.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/3977/commits/38d544f75a43228f7b6a4877e84247ead37cde2f

editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
```diff
@@ -3198,6 +3199,158 @@ describe('inspector tests with real metadata', () => {
...
+        'position-left-number-input',
+      )) as HTMLInputElement
+
+      expect(leftControl.value).toBe('10')
+
+      const elementFrame = getFrame(targetPath, renderResult)
+
+      await setControlValue('position-left-number-input', '25%', renderResult.renderedDOM)
+
+      expect(getFrame(targetPath, renderResult)).toBe(elementFrame)
```
balazsbajorics, line 3299: if we show a toast for this, you could also assert the toast here

balazsbajorics (approved):

PR #3978, opened 2023-07-25
wrap in group: preserve Hug pins

balazsbajorics:
**Problem:**
Wrapping in Group destroys Hug pins by replacing them with their fixed measured value

**Fix:**
For width and height if the pin is Hug, leave it as is

### DIFF
--- a/editor/src/components/inspector/inspector-common.ts
+++ b/editor/src/components/inspector/inspector-common.ts
@@ -11,6 +11,7 @@ import {
 import type {
   ElementInstanceMetadata,
   ElementInstanceMetadataMap,
+  JSXAttributes,
 } from '../../core/shared/element-template'
 import {
   isJSXElement,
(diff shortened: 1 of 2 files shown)


ruggi (approved):

enidemi (approved):

editor/src/components/inspector/inspector-common.ts
```diff
@@ -621,7 +622,7 @@ export function detectFillHugFixedState(
     getSimpleAttributeAtPath(right(element.element.value.props), PP.create('style', property)),
   )
 
-  if (simpleAttribute === MaxContent) {
+  if (isHugFromStyleAttribute(element.element.value.props, property)) {
```
Rheeseyb, line 625: `element.element.value.props` 👀

Rheeseyb (approved):

PR #3982, opened 2023-07-25
Distinguish between groups and groups that can be trued up.

seanparsons:
**Problem:**
Currently we use the same check to see if something is a group when also checking if the group should be trued up. Which means that an element might not be indicated that it is a group (in the navigator for example) when it is not eligible to be trued up.

**Fix:**
This change splits the checks into two parts:
- `treatElementAsGroupLike` - The original function, now simplified to just check if it is our `Group` element. This is purely to identify groups.
- `allowGroupTrueUp` - Newly added to determine if the group can be trued up.

Behind the scenes of this we determine if a group is in a state of `valid`, `warning` or `error`. Those that are `valid` or `warning` can be trued up, whereas those in a state of `error` cannot.

The prior check for if the children of the group are absolutely positioned has moved from `treatElementAsGroupLike` to the validity check returning `error` if they are not absolutely positioned.

**Commit Details:**
- Implemented `allowGroupTrueUp` to determine if an element is both a group
  and can be trued up.
- Added `GroupValidity` type and supporting functions to retrieve that from
  `GroupState` and `InvalidGroupState`.
- Reworked `getGroupState` to use `replaceNonDOMElementPathsWithTheirChildrenRecursive`
  so that it treats elements like fragments as transparent.
- Implemented `getGroupValidity` that just wraps around the result of `getGroupState`.
- Used `allowGroupTrueUp` in place of `treatElementAsGroupLike` where appropriate.


editor/src/components/canvas/canvas-strategies/post-action-options/navigator-reparent.ts
```diff
@@ -173,13 +175,15 @@ export const PropsReplacedNavigatorReparentPostActionChoice = (
 function adjustIntendedCoordinatesForGroups(
   jsxMetadata: ElementInstanceMetadataMap,
   pathTrees: ElementPathTrees,
+  allElementProps: AllElementProps,
   reparentTargetPath: ElementPath,
   intendedCoordinates: CanvasPoint,
   element: ElementInstanceMetadata | null,
 ): CanvasPoint {
-  const reparentTargetParentIsGroup = treatElementAsGroupLike(
+  const reparentTargetParentIsGroup = allowGroupTrueUp(
```
balazsbajorics, line 183: I think here we should be using the more optimistic new treatElementAsGroupLike, because this code tries to make the freshly reparented element be a good group citizen, and that should be the desired behavior even if the group is currently broken

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.tsx
```diff
@@ -183,9 +183,10 @@ export function absoluteResizeBoundingBoxStrategy(
                 return []
               }
 
-              const elementIsGroup = treatElementAsGroupLike(
+              const elementIsGroup = allowGroupTrueUp(
```
balazsbajorics, line 186: here too, I'd make it more relaxed and use treatElementAsGroupLike

editor/src/components/canvas/canvas-strategies/strategies/group-helpers.ts
```diff
@@ -21,30 +22,76 @@ import { importAlias } from '../../../../core/shared/project-file-types'
...
+      case 'valid':
+      case 'warning':
+        return true
+      case 'error':
+        return false
+      default:
+        assertNever(groupValidity)
+    }
+  } else {
+    return false
```
balazsbajorics, line 57: ultra nitpick but this would be nicer as an early return above the happy path to the left

balazsbajorics (changes requested):
even though your changes are a pure refactor, I commented on two places where I think the right move would be to relax the ruleset and keep using treatElementAsGroupLike

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.tsx
seanparsons: Fixed in c381620da.

editor/src/components/canvas/canvas-strategies/post-action-options/navigator-reparent.ts
seanparsons: Fixed in c381620da.

balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/strategies/group-helpers.ts
```diff
@@ -21,30 +22,76 @@ import { importAlias } from '../../../../core/shared/project-file-types'
 import { assertNever } from '../../../../core/shared/utils'
 import { styleStringInArray } from '../../../../utils/common-constants'
 import { isCSSNumber } from '../../../inspector/common/css-utils'
+import { replaceNonDOMElementPathsWithTheirChildrenRecursive } from './fragment-like-helpers'
 
+// Returns true if the element should be treated as a group,
+// even if it's configuration (including its children) means that we cannot do any
+// group manipulation operations.
```
ruggi, line 29: \o\ \o\

ruggi (approved):

PR #3983, opened 2023-07-25
Unify navigator label icon logic

Rheeseyb:
**Problem:**
A recent regression in the Navigator highlighted that we had ended up with 2 very similar functions for determining the icon to show in a Navigator row

**Fix:**
Merge the two functions

**TODO:**
- [ ] Tests that check each and every icon 😩

### DIFF
--- a/editor/src/components/inspector/common/name-and-icon-hook.ts
+++ b/editor/src/components/inspector/common/name-and-icon-hook.ts
@@ -90,6 +90,7 @@ function getNameAndIconResult(
       pathTrees,
       autoFocusedPaths,
       null,
+      allElementProps,
     ),
   }
 }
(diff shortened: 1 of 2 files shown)


editor/src/components/navigator/navigator-item/navigator-item.spec.browser2.tsx
```diff
@@ -0,0 +1,329 @@
...
+    async function checkNavigatorIcon(
+      description: string,
+      expectedIcnProps: { category: string; type: string } | 'no-icon',
+      navigatorEntry: NavigatorEntry,
+      expectWarningIcon: 'expect-warning-icon' | 'expect-normal-icon' = 'expect-normal-icon',
+    ) {
+      const testId = layoutIconTestIdForEntry(navigatorEntry)
+
+      if (expectedIcnProps === 'no-icon') {
+        await expect(async () => editor.renderedDOM.getByTestId(testId)).rejects.toThrow()
```
balazsbajorics, line 188: Nit: according to the [cheat sheet](https://testing-library.com/docs/dom-testing-library/cheatsheet/), you could use `queryByTestId` and expect it returns a null. furthermore, queryByTestId synchronously returns null, so you can remove the extra await from here

enidemi (approved):

gbalint (approved):

seanparsons (approved):

PR #3986, opened 2023-07-26
Ensure first interaction works correctly.

seanparsons:
**Problem:**
During the first interaction (dragging an element for example), several of the navigator elements turn orange erroneously. But only during that interaction, subsequent ones work completely fine.

**Cause:** 
`clearSpyCollectorInvalidPaths` wasn't working correctly because the entries in the `validPaths` set weren't referentially equal, when they were deeply equal. Likely this is due to the entries being removed from the path cache, resulting in new instances being created.

**Fix:**
Switching from `Set<ElementPath>` to `Set<string>` for the `validPaths` value, to make it completely independent of the particular instance of `ElementPath`, as `string` values are compared by value not by reference.

**Commit Details:**
- Use `Set<string>` instead of `Set<ElementPath>` for valid paths as the latter can fail because of reference equality checks that `Set` assumes.


enidemi (approved):

gbalint (approved):

balazsbajorics (commented):
+1

PR #3987, opened 2023-07-26
Map support 

gbalint:
# Description

This PR is a simple solution to create a new elefant type for maps.

It is a simple solution, because instead of finding out what are the building blocks of a map expression and parsing and storing them separately (as we did with conditionals, which are constructed from a condition, a whenTrue and whenFalse expressions), I just copied the structure of `JSExpressionOtherJavascript`. So we store the whole js code of a map expression, and we just recognize it is a special expression (and store it with a different type), but we don't store any special map specific data in them.

```typescript
export type JSXMapExpression = {
  type: 'JSX_MAP_EXPRESSION'
  originalJavascript: string
  javascript: string
  transpiledJavascript: string
  definedElsewhere: Array<string>
  sourceMap: RawSourceMap | null
  uid: string
} & WithElementsWithin
```

This solution is good enough to create a new kind of navigator item for maps. It won't be enough to implement slot-like behavior for maps though.
However, we will be able to incrementally extend the model without removing the storage of the full javascript code. And when we are ready for that, we can even throw away the full javascript representation, we'll see if that is going to be practical later or not.

# What is a map?

Unlike most of our element types, maps are not a specific syntactic construct in javascript. Map is just a specific function call, which I recognize with the following ruleset:

It is a:
- a function call
- a property access expression
- the called function (the property) is called "map"

There is no guarantee that this "map" function is the real one, and not e.g. a map function on your custom component.
If this is not good enough, we can refine this ruleset, e.g. by using a typechecker to make sure the function is accessed on an array.

# The PR includes

- parsing maps and storing them in projectcontents
- showing maps with the "Map" label in the navigator
- updated existing tests and added new tests

# Note
A lot of tests only include changes in the uid of an expression, because with the model changes all the map expressions receive a different uid than before. These kinds of updates are hard to follow and should not be necessary. The solution is to make it possible to define the uid of an expression/map-expression from code with a standard comment (similarly to conditionals)
This can come in a new PR though.


editor/src/core/shared/element-path-tree.ts
```diff
@@ -94,7 +94,7 @@ function getChildrenPaths(
     element.element.value.children.length > 0
   ) {
     childrenFromElement = element.element.value.children
-      .filter((child) => !isJSXTextBlock(child) && !isJSExpressionOtherJavaScript(child))
+      .filter((child) => !isJSXTextBlock(child) && !isJSExpressionMapOrOtherJavaScript(child))
```
seanparsons, line 97: I wonder if this condition should be an exhaustive switch/case so that any subsequent changes don't cause later pains.

editor/src/core/shared/uid-utils.ts
```diff
@@ -662,7 +666,7 @@ function isUtopiaJSXElement(
 function isUtopiaJSExpressionOtherJavaScript(
```
seanparsons, line 666: This should probably be renamed to match the function it calls.

seanparsons (approved):
Nice.

editor/src/core/shared/uid-utils.ts
```diff
@@ -662,7 +666,7 @@ function isUtopiaJSXElement(
...
   element: JSXElementChild | ElementInstanceMetadata,
 ): element is JSExpressionOtherJavaScript {
-  return isJSExpressionOtherJavaScript(element as any)
+  return isJSExpressionMapOrOtherJavaScript(element as any)
```
Rheeseyb, line 669: This is definitely cheating

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1205,25 +1221,53 @@ function createExpressionOtherJavaScript(
   sourceMap: RawSourceMap | null,
   elementsWithin: ElementsWithin,
   alreadyExistingUIDs: Set<string>,
-): WithParserMetadata<JSExpressionOtherJavaScript> {
+): WithParserMetadata<JSExpressionMapOrOtherJavascript> {
+  const isList = (() => {
```
Rheeseyb, line 1225: I know this is a draft, but so that you don't miss this one later - this is now just using `isNodeAMapExpression`

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1205,25 +1221,53 @@ function createExpressionOtherJavaScript(
...
   // Ideally the value we hash is stable regardless of location, so exclude the SourceMap value from here and provide an empty UID.
-  const value = jsExpressionOtherJavaScript(
-    javascript,
-    transpiledJavascript,
-    definedElsewhere,
-    null,
-    elementsWithin,
-    '',
-  )
+  const value = isList
```
Rheeseyb, line 1234: Huh, I'm not sure why both `value` and `expression` are almost identical here. It looks like that was maybe the result of refactoring over time (possibly from when the 2 arbitrary JS types were merged)

editor/src/components/navigator/layout-element-icons.ts
```diff
@@ -260,7 +260,18 @@ export function createElementIconPropsFromMetadata(
 
   const isExpressionOtherJavascript = MetadataUtils.isExpressionOtherJavascriptFromMetadata(element)
   if (isExpressionOtherJavascript) {
-    // TODO: need a real icon
+    // TODO: needs a real icon
+    return {
+      category: 'element',
+      type: 'lists3',
```
Rheeseyb, line 266:
```suggestion
      type: 'lists',
```

editor/src/core/shared/uid-utils.ts
gbalint: Yes-yes, I removed that

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
gbalint: Yes, fixed

Rheeseyb (approved):

PR #3993, opened 2023-07-27
Make sure to apply group child pins on unwrap

### LINKED ISSUE #3992: Make sure group children get applied pins on unwrap

ruggi:
Fixes #3992 

**Problem:**

When unwrapping group children, the inherited grandchildren may not have valid pins and/or jump around (e.g. unwrapping a flex container).

**Fix:**

Apply `createPinChangeCommandsForElement` commands in the `UNWRAP_ELEMENT` function for group children.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
@@ -787,39 +787,6 @@ export function createWrapInGroupActions(
 
   const groupPath = EP.appendToPath(parentPath.intendedParentPath, group.uid) // TODO does this work if the parentPath is a ConditionalClauseInsertionPath?
 
-  function createPinChangeCommandsForElement(
-    elementMetadata: ElementInstanceMetadata | null,
-    expectedPath: ElementPath,
-  ): Array<CanvasCommand> {
-    if (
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -835,14 +802,56 @@ export function createWrapInGroupActions(
...
+          ),
+        )
       })
     })
   })
 
   return applyCommandsAction([...deleteCommands, insertGroupCommand, ...pinChangeCommands])
 }
 
+export function createPinChangeCommandsForElement(
```
balazsbajorics, line 820: now that this is exported, I think the name should refer to elements that are now becoming group children... maybe it's too naive but how about `createPinChangeCommandsForElementBecomingGroupChild` ?

balazsbajorics (approved):
nice!!

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
ruggi: I guess it'd work! https://github.com/concrete-utopia/utopia/pull/3993/commits/1579c5e1628672cfebc4a9e31b2cb0f457976165

PR #3996, opened 2023-07-27
alignment buttons sticky again

lankaukk:
After making the inspector "float", the alignment buttons stopped behaving they way we wanted. Now they are sticky!

Bonus: Killed a rogue border above the Conditional Section that was more visible in dark mode

Screenshot 2023-07-27 at 3 00 04 PM


editor/src/components/inspector/inspector.tsx
```diff
@@ -165,9 +165,6 @@ const AlignmentButtons = React.memo((props: { numberOfTargets: number }) => {
       style={{
         justifyContent: 'space-around',
         height: UtopiaTheme.layout.rowHeight.normal,
-        position: 'sticky',
```
maltenuhn, line 168: wasn't this _meant_ to not make them scroll?

lankaukk: that... would make sense. Do we want them not to scroll then?

editor/src/components/inspector/sections/layout-section/conditional-section.tsx
```diff
@@ -336,8 +337,11 @@ export const ConditionalSection = React.memo(({ paths }: { paths: ElementPath[]
 
   return (
     <div style={{ paddingBottom: 8 }} data-testid={ConditionalSectionTestId}>
-      <InspectorSubsectionHeader
+      <FlexRow
```
Rheeseyb, line 340: This change will also need to be made in https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/inspector/sections/code-element-section.tsx

enidemi (approved):

balazsbajorics (approved):
I assume you had a good reason to kill some InspectorSubsectonHeaders, the rest of the code looks okay!

balazsbajorics (changes requested):
I spoke too soon! 
image
It looks like the `NineBlockControl` collapsed and half of it is missing

balazsbajorics (approved):

PR #4003, opened 2023-07-28
Update group frame buttons in the inspector for groups.

seanparsons:
**Problem:**
Currently the "resize to fit", "resize to fill" and "fixed size" buttons in the inspector don't make sense as they currently stand for groups.

**Fix:**
Disabled the first two buttons mentioned above and used some existing logic to convert a group to a frame if a selected element is a group.

**Commit Details:**
- Resize to fit and resize to fill buttons now disabled if the selected element is a group.
- Set to fixed size button now converts and groups to frames.

### DIFF
--- a/editor/src/components/inspector/resize-to-fit-control.spec.browser2.tsx
+++ b/editor/src/components/inspector/resize-to-fit-control.spec.browser2.tsx
@@ -1,8 +1,14 @@
+/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "expectNoAction"] }] */
 import { cmdModifier } from '../../utils/modifiers'
-import { expectSingleUndo2Saves } from '../../utils/utils.test-utils'
+import {
+  expectNoAction,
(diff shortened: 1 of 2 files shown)


ruggi (approved):

editor/src/components/inspector/resize-to-fit-control.spec.browser2.tsx
```diff
@@ -261,3 +346,59 @@ export var storyboard = (
   </Storyboard>
 )
 `
+
+const projectWithGroup = `import * as React from 'react'
```
balazsbajorics, line 350: nitpick but I think the test is more readable if both the before and after code is in the test body. but it's better if it's not the entire project code, just the relevant middle snippet. 

but at the same time the rest of the test file is structured like this, so I get the argument that this new test just follows the (admittedly suboptimally readable) pattern

balazsbajorics (approved):

PR #4004, opened 2023-07-28
Make sure INSERT_INSERTABLE adds missing pins when targeting a group

### LINKED ISSUE #4002: Make sure insertables into a group get pins

ruggi:
Fixes #4002 

**Problem:**

Inserting a `img` (or any other insertable with missing pins) into a group ends up with an invalid group because the `img` has no position pins.

https://github.com/concrete-utopia/utopia/assets/1081051/00d14798-3c09-435f-a0e5-9f1650abe1f5

**Fix:**

Make sure that `INSERT_INSERTABLE` adds the missing pins on the element. Since the metadata is not available, I added an optimistic variant of `createPinChangeCommandsForElementBecomingGroupChild` that will try to get the best pins possible from the combination of the element _and_ the target group.

- For the position pins it will try to put the element in the center of the group, or `0,0` as a fallback.
- For the dimension pins it will use the element ones, or default to the group ones as a fallback.

https://github.com/concrete-utopia/utopia/assets/1081051/3ac83e79-f3bb-4b77-8e94-78f6d2b7ba79

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -134,6 +134,7 @@ import {
   nullIfInfinity,
   isNotNullFiniteRectangle,
   localRectangle,
+  zeroRectIfNullOrInfinity,
 } from '../../../core/shared/math-utils'
 import type {
   PackageStatusMap,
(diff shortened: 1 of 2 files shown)


editor/src/components/editor/actions/actions.tsx
```diff
@@ -4771,15 +4774,39 @@ export const UPDATE_FNS = {
...
+        )
+
+        if (action.toInsert.element.type === 'JSX_ELEMENT') {
+          // this condition would need updating when we can insert fragments or conditionals with children
+          groupCommands.push(
+            ...createPinChangeCommandsForElementInsertedIntoGroup(
+              newPath,
+              right(action.toInsert.element.props),
+              groupFrame,
+              localRectangle(groupFrame),
```
balazsbajorics, line 4792: this looks wrong! a localRectangle means local relative to the offset-providing parent or ancestor. there's functions to convert a globalFrame to a local rectangle by providing the parent. or you could just access groupMetadata.localRectangle

balazsbajorics (changes requested):

editor/src/components/editor/actions/actions.tsx
ruggi: aha! would this work? https://github.com/concrete-utopia/utopia/pull/4004/commits/c57499f86b8e3b69eb855b2fc3eb6de7bca88b4a

balazsbajorics (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -4771,15 +4775,38 @@ export const UPDATE_FNS = {
...
-        ],
+
+      let groupCommands: CanvasCommand[] = []
+      if (treatElementAsGroupLike(editor.jsxMetadata, action.insertionPath.intendedParentPath)) {
+        const group = MetadataUtils.findElementByElementPath(
+          editor.jsxMetadata,
+          action.insertionPath.intendedParentPath,
+        )
+        if (group != null && action.toInsert.element.type === 'JSX_ELEMENT') {
+          // this condition would need updating when we can insert fragments or conditionals with children
```
seanparsons, line 4786: I would suggest making this throw an exception for things that aren't `JSX_ELEMENT`, otherwise we'll just pass right over this when we do support that.

seanparsons (approved):

editor/src/components/editor/actions/actions.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/4004/commits/d578af2083636126e1df32359fecc18b413f1c09

PR #4005, opened 2023-07-28
Fix text expressions in conditionals

gbalint:
**Problem:**
Normally we hide expression elements from the navigator when they only return text, e.g.

```<div>{1+1}</div>```

is shown as:

image

The reason is that the expression results in text content, and normally we don't treat text content as elefants. 

However, when text content is in a conditional branch, we show a navigator line for that text content (otherwise it the conditional branch would be empty).

The bug was that the conditional branch in the navigator was really empty when there was an expression which returned text content:
```{true ? 1 + 1 : null}```
resulted in 
image

**Fix:**
I made sure we don't filter out text only expressions when their parent is a conditional.

### DIFF
--- a/editor/src/components/navigator/navigator.spec.browser2.tsx
+++ b/editor/src/components/navigator/navigator.spec.browser2.tsx
@@ -598,6 +598,15 @@ export var storyboard = (
       {[0, 1, 2].map((i) => (
         <div>fifth {i}</div>
       ))}
+
+      <div data-uid='text-expr-cond-wrapper'>{
+        // @utopia/uid=text-expr-cond
+        true ? (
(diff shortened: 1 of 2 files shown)


balazsbajorics:
I think I'm weeks behind on the conversation / design. In the spring, the idea was that we only show conditionals in the navigator if at least one branch is JSXElement or if both branches are null. 

So we _would not_ show conditionals in the navigator if their content is {true ? text : text } or {true ? text : null} or {true ? null  text} 

Did this thinking evolve? Is the null slot more important now? Or did we change the design because the rest of the navigator changed (with the code blocks now showing up as Elefants)?

Rheeseyb (approved):

bkrmendy (approved):

PR #4010, opened 2023-07-31
Group delete behaviours

### LINKED ISSUE #4009: Group delete behaviours

ruggi:
Fixes #4009 

**Problem:**

1. Deleting a group child should select the next sibling instead of the group itself.
2. Deleting the last group child should delete the group itself (like a fragment).

**Fix:**

Implement both the above.

Kapture 2023-07-31 at 14 40 02

### DIFF
--- a/editor/src/components/editor/actions/actions.spec.browser2.tsx
+++ b/editor/src/components/editor/actions/actions.spec.browser2.tsx
@@ -363,6 +363,79 @@ describe('actions', () => {
     `,
         wantSelection: [makeTargetPath('aaa')],
       },
+      {
+        name: 'delete group child selects next sibling',
+        input: `
+          <View data-uid='view'>
+            <Group data-uid='group'>
+              <div data-uid='child1' style={{ position: 'absolute', width: 10, height: 10, top: 0, left: 0, background: 'blue' }} />
+              <div data-uid='child2' style={{ position: 'absolute', width: 10, height: 10, top: 0, left: 40, background: 'blue' }} />
+              <div data-uid='child3' style={{ position: 'absolute', width: 10, height: 10, top: 40, left: 0, background: 'blue' }} />
+              <div data-uid='child4' style={{ position: 'absolute', width: 10, height: 10, top: 40, left: 40, background: 'blue' }} />
+            </Group>
+          </View>
+        `,
+        targets: [makeTargetPath('view/group/child3')],
+        wantCode: `
+          <View data-uid='view'>
+            <Group
+              data-uid='group'
+              style={{ width: 50, height: 50 }}
+            >
+              <div data-uid='child1' style={{ position: 'absolute', width: 10, height: 10, top: 0, left: 0, background: 'blue' }} />
+              <div data-uid='child2' style={{ position: 'absolute', width: 10, height: 10, top: 0, left: 40, background: 'blue' }} />
+              <div data-uid='child4' style={{ position: 'absolute', width: 10, height: 10, top: 40, left: 40, background: 'blue' }} />
+            </Group>
+          </View>
+        `,
+        wantSelection: [makeTargetPath('view/group/child1')],
+      },
+      {
+        name: 'delete group child selects next sibling (multiple selection)',
+        input: `
+          <View data-uid='view'>
+            <Group
+              data-uid='group'
+              style={{ width: 50, height: 50 }}
+            >
+              <div data-uid='child1' style={{ position: 'absolute', width: 10, height: 10, top: 0, left: 0, background: 'blue' }} />
+              <div data-uid='child2' style={{ position: 'absolute', width: 10, height: 10, top: 0, left: 40, background: 'blue' }} />
+              <div data-uid='child3' style={{ position: 'absolute', width: 10, height: 10, top: 40, left: 0, background: 'blue' }} />
+              <div data-uid='child4' style={{ position: 'absolute', width: 10, height: 10, top: 40, left: 40, background: 'blue' }} />
+            </Group>
+            <div data-uid='foo'>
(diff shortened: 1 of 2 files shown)


editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -237,7 +237,7 @@ describe('actions', () => {
       />
     </View>
     `,
-        wantSelection: [makeTargetPath('aaa/000')],
+        wantSelection: [makeTargetPath('aaa/000/ccc')],
```
ruggi, line 240: …because now fragments are aligned with groups

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -363,6 +363,79 @@ describe('actions', () => {
...
+            <div data-uid='foo' />
+          </View>
+        `,
+        wantSelection: [makeTargetPath('view/group/child1'), makeTargetPath('view/foo')],
+      },
+      {
+        name: 'delete last group child deletes the group',
+        input: `
+          <View data-uid='view'>
+            <Group data-uid='group'>
```
balazsbajorics, line 428: can you make a test case / change this test case to contain multiple <Groups> embedded in each other, each only having a single child, so when you delete the innermost leaf, we show that the automatic deletion is recursive?

ruggi: oh yes 👍

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -363,6 +363,79 @@ describe('actions', () => {
...
+              style={{ width: 50, height: 50 }}
+            >
+              <div data-uid='child1' style={{ position: 'absolute', width: 10, height: 10, top: 0, left: 0, background: 'blue' }} />
+              <div data-uid='child2' style={{ position: 'absolute', width: 10, height: 10, top: 0, left: 40, background: 'blue' }} />
+              <div data-uid='child4' style={{ position: 'absolute', width: 10, height: 10, top: 40, left: 40, background: 'blue' }} />
+            </Group>
+            <div data-uid='foo' />
+          </View>
+        `,
+        wantSelection: [makeTargetPath('view/group/child1'), makeTargetPath('view/foo')],
```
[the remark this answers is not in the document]
ruggi: interesting, for me it feels intuitive this way - say you want to delete multiple nested elements from two groups/fragments it would be _extremely_ cumbersome to re-do the selection every time 🤔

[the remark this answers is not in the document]
balazsbajorics: haha but if you want to do _that_ you could just multiselect the multiple elements you want to delete :)

editor/src/components/editor/actions/actions.spec.browser2.tsx
ruggi: mmh I don't think it makes sense: deleting the innermost leaf of a group deletes the innermost group, but the other ones are not empty. This is extra logic we should built into this (for fragments as well)

balazsbajorics:
```javascript
<Group>
  <Group>
    <Group>
      <div>Hello!</div>
    </Group>
  </Group>
</Group>
```
If you delete Hello, all of the enclosing Groups shall be gone

ruggi: https://github.com/concrete-utopia/utopia/pull/4010/commits/a43f08057c692c871fa61f6266f11248f881696f adds recursive deletion for fragments/groups (which was not implemented), I added tests for more complex cases as well (e.g. multiselect)

balazsbajorics (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1669,26 +1707,61 @@ export const UPDATE_FNS = {
...
-            if (parentIsFragment && parentWillBeEmpty) {
-              return parentPath
+
+            if (mustDeleteEmptyParent && parentWillBeEmpty) {
+              const bubbledUp = [
+                parentPath,
+                ...deletableParents(editor.jsxMetadata, parentPath, allSelectedPaths),
+              ]
+              bubbledUpDeletions.push(...bubbledUp)
+              return bubbledUp[bubbledUp.length - 1]
```
seanparsons, line 1729: We do have `closestSharedAncestor` in `element-path.ts` which I _think_ will handle this without any need for things to be in a certain order.

ruggi: you're right, but the problem is that it's important to keep track of all ancestors (up to the shared one) to make sure the pruning of selected views works correctly

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1669,26 +1707,61 @@ export const UPDATE_FNS = {
...
           mapDropNulls((view) => {
             const parentPath = EP.parentPath(view)
+            if (behavesLikeGroupOrFragmentForDeletion(editor.jsxMetadata, parentPath)) {
+              // there may be bubbled up deletions, so find out which is the actual parent
+              // where the bubbles stopped
+              const parentsBubbledUp = [
+                view,
+                ...deletableParents(editor.jsxMetadata, parentPath, staticSelectedElements),
+              ].map(EP.parentPath)
+              const actualParent = parentsBubbledUp[parentsBubbledUp.length - 1]
```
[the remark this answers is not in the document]
ruggi: same as above, we need the entirety of the upward subtree (`ignorePaths`)

PR #4016, opened 2023-08-01
Prevent wrapping/cutting empty groups

### LINKED ISSUE #4015: Prevent empty group wrap/cut

ruggi:
Fixes #4015

**Problem:**

It's possible to wrap empty groups into groups, which is illegal. Likewise is cutting empty groups on the storyboard.

**Fix:**

- Show a toast and stop when trying to CMD+G an empty group
- Show a toast and stop when trying to CMD+X an empty group on the storyboard
- Show a toast and stop when trying to wrap via the insert menu an empty group*

(*) I deliberately extended this to all wrapping elements (not just group-into-groups) because trying to wrap an empty group with almost any other element (e.g. a `div`) would break the editor (and it does not feel conceptually correct either).


balazsbajorics (approved):
👌

bkrmendy (approved):

PR #4019, opened 2023-08-01
Unwrap multiple targets

### LINKED ISSUE #4018: Keep selection on multiselect unwrap

ruggi:
Fixes #4018 

**Problem:**

- Unwrapping multiple groups with CMD+G only selects the children of the first unwrapped group
- Unwrapping multiple groups with the context menu only unwraps one of the groups, as multiple unwrapping is not supported

Kapture 2023-08-01 at 16 38 36

This is because:
1. The unwrapping for multiple selected views works by enqueueing sequential unwraps, each of which resets the selection
2. The unwrapping for the context menu triggers the unwrap on the `0th` element of the selection

**Fix:**

- Rework the unwrapping logic so that it supports multiple targets, which solves both the above in one go
- Make sure the old tests pass
- Add specific tests

Kapture 2023-08-01 at 16 39 41

**Notes:**

I tried to touch the logic of the unwrapping as little as possible: there's quite a bit of room for improvement there, but imho this is already a relevant enough change that it should simply make sure the existing logic works fine as-is, with just the minimum set of additions to work with the multiple targets approach, without adding all the cognitive load of a larger rewrite.

### DIFF
--- a/editor/src/components/context-menu-items.ts
+++ b/editor/src/components/context-menu-items.ts
@@ -402,9 +402,7 @@ export const unwrap: ContextMenuItem<CanvasData> = {
     )
   },
   action: (data, dispatch?: EditorDispatch) => {
-    if (data.selectedViews.length > 0) {
-      requireDispatch(dispatch)([EditorActions.unwrapElement(data.selectedViews[0])], 'everyone')
-    }
+    requireDispatch(dispatch)([EditorActions.unwrapElements(data.selectedViews)], 'everyone')
   },
 }
 
--- a/editor/src/components/editor/actions/actions.spec.browser2.tsx
+++ b/editor/src/components/editor/actions/actions.spec.browser2.tsx
@@ -15,7 +15,7 @@ import {
   TestScenePath,
   TestSceneUID,
 } from '../../../components/canvas/ui-jsx.test-utils'
-import { deleteSelected, selectComponents, unwrapElement, wrapInElement } from './action-creators'
+import { deleteSelected, selectComponents, unwrapElements, wrapInElement } from './action-creators'
 import type { ElementPath } from '../../../core/shared/project-file-types'
 import type { ElementPaste } from '../action-types'
 import type { InsertionPath } from '../store/insertion-path'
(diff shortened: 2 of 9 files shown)


balazsbajorics (approved):

editor/src/components/context-menu-items.ts
```diff
@@ -402,9 +402,7 @@ export const unwrap: ContextMenuItem<CanvasData> = {
     )
   },
   action: (data, dispatch?: EditorDispatch) => {
-    if (data.selectedViews.length > 0) {
-      requireDispatch(dispatch)([EditorActions.unwrapElement(data.selectedViews[0])], 'everyone')
-    }
+    requireDispatch(dispatch)([EditorActions.unwrapElements(data.selectedViews)], 'everyone')
```
seanparsons, line 405: Lovely!

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -5873,3 +6102,7 @@ export var storyboard = (
     })
   })
 })
+
+function comparePaths(a: ElementPath, b: ElementPath): number {
```
seanparsons, line 6106: We should probably put this in `element-path.ts`.

seanparsons:
Does do the right thing if we had something like:
```
Group 1 (selected)
+ Div 1
  + Group 2 (selected)
    + Div 2
```
Then unwrap things which I'd expect to result in:
```
Div 1
+ Div 2
```

seanparsons (approved):

ruggi:
> Does this do the right thing if we had something like:
> 
> ```
> Group 1 (selected)
> + Div 1
>   + Group 2 (selected)
>     + Div 2
> ```
> 
> Then unwrap things which I'd expect to result in:
> 
> ```
> Div 1
> + Div 2
> ```

_almost_, it was not correctly cleaning up the nested group. I adjusted and added a specific test for it here https://github.com/concrete-utopia/utopia/pull/4019/commits/f68e730acf13bee7e401bdb1dedc508443ffa7e1

PR #4022, opened 2023-08-01
Fixes for control structures in the navigator

gbalint:
**Description:**
This PR was started as a simple fix to the navigator.
For conditionals blue colors were already used for the layout icon, and I extended that to code and map items too.
I also changed the code icon to be different from the map icon.

However, when working on the tests for this, I realized that `MetadataUtils.isExpressionOtherJavascriptFromMetadata` returns true not just for otherjs expressions but for maps too!

I fixed that, and then a bunch of tests broke, because it turned out we have a few features which depended on this bug, because they need the same behavior for maps as for expressions.

I also modified the label for genericcode items to `CODE` and added new tests for all kinds navigator labels.

So after this PR the three kinds of control structure navigator items (conditional, map and code) use the same unified design with different icons:

image

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item.spec.browser2.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item.spec.browser2.tsx
@@ -280,7 +280,7 @@ describe('Navigator item row icons', () => {
     await checkNavigatorIcon('Empty Slot', 'no-icon', visibleNavigatorTargets[11])
     await checkNavigatorIcon(
       'Code element',
-      { category: 'element', type: 'lists' },
+      { category: 'element', type: 'genericcode' },
       visibleNavigatorTargets[12],
     )
     await checkNavigatorIcon(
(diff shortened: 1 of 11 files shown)


editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -513,21 +515,34 @@ export const NavigatorItem: React.FunctionComponent<
...
+      if (MetadataUtils.isJSXMapExpressionFromMetadata(elementMetadata)) {
+        return 'map'
+      }
+      if (MetadataUtils.isExpressionOtherJavascriptFromMetadata(elementMetadata)) {
+        return 'code'
       }
+      return 'none'
     },
-    'NavigatorItem isConditional',
+    'NavigatorRowLabel codeItemType',
```
ruggi, line 540: I believe this should be `NavigatorItem`
```suggestion
    'NavigatorItem codeItemType',
```

ruggi (approved):

seanparsons (approved):

editor/src/components/navigator/navigator-item/navigator-item.spec.browser2.tsx
```diff
@@ -326,4 +328,47 @@ describe('Navigator item row icons', () => {
       visibleNavigatorTargets[20],
     )
   })
+
+  it('Should show the correct labels for each type of row', async () => {
```
Rheeseyb, line 332: Nice!

Rheeseyb (approved):

PR #4023, opened 2023-08-01
Capture and parse comments surrounding JSExpressions

Rheeseyb:
**Problem:**
We would like a way to provide overrides to expressions and maps for the sake of testing (override UIDs) and enabling certain capabilities in the editor.

**Fix:**
Parse the comments of JS Expressions in the same way as we do with other types, and use those comments when attempting to create a UID for the parsed expression. The bulk of the actual changes exist in `parser-printer-parsing.ts`, with most of the remaining changes being test fixes to support this.

Along the way I discovered duplicate factories for creating `JSExpressionMapOrOtherJavascript`, so I merged those. I also discovered that we have 2 almost identical functions for parsing arbitrary expressions, but merging those was less trivial so I'm saving that for a follow up PR.

**Note:**
This PR identified a bug in the way we add comments to JS expressions when printing the parsed model, which shows up in 2 of the tests. That bug already existed in the code, so that will be tackled separately after.

Also note that because of the above we aren't able to programmatically add comments to JS Expressions yet, though that will come with a follow up PR shortly.


editor/src/components/canvas/canvas-utils.spec.browser2.tsx
```diff
@@ -207,8 +207,9 @@ describe('updateFramesOfScenesAndComponents - pinFrameChange -', () => {
 
   it('no layout prop on child', async () => {
     const renderResult = await renderTestEditorWithCode(
+      // FIXME This should be fine with {{ ...props.style }} but that causes the uids to switch in tests only
```
Rheeseyb, line 210: This one is beyond me. The issue only shows up in this specific test, and only shows up when running in the test environment

bkrmendy (approved):

balazsbajorics (approved):

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -2612,6 +2616,26 @@ export function parseOutJSXElements(
       : expression.whenFalse
     const whenFalseBlock = parseClause(innerWhenFalse)
 
+    // We need to create the UID here before parsing expression.condition, because we
+    // might be stealing one specified in the comment attached to the expression.condition
```
ruggi, line 2620: yikes 😬

ruggi (approved):

gbalint (approved):

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
Rheeseyb: Actually I'll update this comment a bit. Basically the overrides we're using right now for the conditional type are actually the leading comments of the `expression.condition`, meaning that if we wish to continue doing that then we can't have separate overrides for both the conditional as a whole and the `expression.condition`. That hasn't mattered thus far, but it means right now we can't provide a hard coded UID for the `expression.condition`

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1215,59 +1244,72 @@ function createExpressionValue(
 function createExpressionOtherJavaScript(
   sourceFile: TS.SourceFile,
   node: TS.Node,
+  originalJavascript: string,
   javascript: string,
   transpiledJavascript: string,
   definedElsewhere: Array<string>,
   sourceMap: RawSourceMap | null,
   elementsWithin: ElementsWithin,
+  isList: boolean,
```
seanparsons, line 1253: Shouldn't this be a sum type?

seanparsons: Ah, it's transplanted from elsewhere, so *shrugs*.

PR #4034, opened 2023-08-02
Flatten unwrap selection

### LINKED ISSUE #4033: flatten selection for unwrap

ruggi:
Fixes #4033 

**Problem:**

This is a followup to https://github.com/concrete-utopia/utopia/pull/4019 which fixes an existing bug that would delete an entire subtree if an unwrap was triggered on multiple elements of the same subtree.

**Fix:**

Make sure to flatten the unwrap targets so that the unwrap only triggers the topmost ancestor and ignores the other elements.

- Adjusted the existing tests
- Added a test to make sure multiselect unwrap still works fine when both a subtree cluster and other elements in a different subtree are targeted.

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -490,7 +490,10 @@ import {
   getReparentOutcomeMultiselect,
   pathToReparent,
 } from '../../canvas/canvas-strategies/strategies/reparent-utils'
-import { areAllSelectedElementsNonAbsolute } from '../../canvas/canvas-strategies/strategies/shared-move-strategies-helpers'
+import {
+  areAllSelectedElementsNonAbsolute,
+  flattenSelection,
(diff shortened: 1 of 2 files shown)


editor/src/components/editor/actions/actions.tsx
```diff
@@ -2283,132 +2286,135 @@ export const UPDATE_FNS = {
...
-              workingEditor.elementPathTree,
-              target,
-            )
+        // make sure to trim descendant paths, so that unwrapping on a subtree only works on the first ancestor
+        const flattenedPaths = flattenSelection(orderedPaths)
 
-            if (!(supportsChildren || elementIsFragmentLike)) {
-              return workingEditor
-            }
+        const withViewsUnwrapped: EditorState = flattenedPaths.reduce((workingEditor, target) => {
```
ruggi, line 2295: nothing changed from here down, it's just a whitespace diff

bkrmendy (approved):

seanparsons (approved):

PR #4039, opened 2023-08-03
Avoid reference equality during deletion.

seanparsons:
**Problem:**
For some inexplicable reason a single `DELETE_SELECTED` test was failing just for me locally and only if I ran the entire browser test suite.

**Cause:**
The underlying cause for this was the use of reference equality with `ElementPath` instances, mostly with `Array.includes` which was not working as expected. With the likely trigger being the element path cache being cleared at a weird time. The reason for this triggering in this very specific case was not discovered.

**Fix:**
Switched a few places to use `EP.containsPath` or `EP.pathsEqual` where appropriate.

**Commit Details:**
- In the `DELETE_SELECTED` action handler, avoid using reference equality when checking `ElementPath` instances.
- In the `DELETE_SELECTED` action handler, avoid using reference equality for metadata but instead check the path with `EP.pathsEqual`.


Rheeseyb (approved):
Oh my word!

gbalint (approved):

PR #4040, opened 2023-08-03
Override map count with comment flag

gbalint:
**Description:**
This PR makes it possible to override the number of rendered children of a map expression:

```typescript
{
    // @utopia/map-count=3
    [0, 1, 2, 3, 4].map((i) => (
      <div
        style={{
          backgroundColor: '#aaaaaa33',
          position: 'absolute',
          left: 40,
          top: 35 + i * 100,
          width: 110,
          height: 66,
        }}
      />
    ))
  }
```

The feature is implemented by providing a `renderLimit` parameter to the `createLookupRender` function. `createLookupRender` returns a render function which just stops rendering after the limit is reached, so the map element will not be able to render more children after the limit.

For the note there is nothing map specific in this, theoretically we could set a limit on the number of children on any kind of element (but I only enabled the comment flag for maps).


ruggi (approved):
🎉

bkrmendy (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -84,11 +85,15 @@ export function createLookupRender(
...
   highlightBounds: HighlightBoundsForUids | null,
   editedText: ElementPath | null,
+  renderLimit: number | null,
 ): (element: JSXElement, scope: MapLike<any>) => React.ReactChild {
   let index = 0
 
   return (element: JSXElement, scope: MapLike<any>): React.ReactChild => {
     index++
+    if (renderLimit != null && index > renderLimit) {
+      return ''
```
balazsbajorics, line 95: have you tried return null here? just commenting without actually trying this in csb but I believe null is safer

balazsbajorics (approved):

PR #4042, opened 2023-08-04
Feature: two pass group true up

balazsbajorics:


ruggi (approved):

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -574,7 +574,7 @@ export function convertFrameToGroup(
       indexPosition: absolute(MetadataUtils.getIndexInParent(metadata, pathTrees, elementPath)),
       importsToAdd: GroupImport,
     }),
-    queueGroupTrueUp(trueUpElementChanged(childInstances[0].elementPath)), // let the editor know that the children are positioned correctly and the Group needs to be shifted/resized
+    queueGroupTrueUp([trueUpElementChanged(childInstances[0].elementPath)]), // TODO change this to trueUpChildrenOfElementChanged(elementPath)
```
seanparsons, line 577: For v2?

editor/src/components/canvas/commands/queue-group-true-up-command.ts
```diff
@@ -12,14 +12,14 @@ import { trueUpTargetToDescription } from '../../../core/model/groups'
 
 export interface QueueGroupTrueUp extends BaseCommand {
   type: 'QUEUE_GROUP_TRUE_UP'
-  target: TrueUpTarget
+  targets: Array<TrueUpTarget>
```
seanparsons, line 15: Since we have arrays of commands, is there a distinction between one `QueueGroupTrueUp` with X `targets` and X instances of `QueueGroupTrueUp` with a target in each?

balazsbajorics: I don't think so! maybe some ultra minor performance difference? it's a single editor state patch instead of `n` patches folded

seanparsons (approved):

PR #4047, opened 2023-08-04
Update React, Vite & Webpack

bkrmendy:
# Description
This PR removes the compatibility flags from React and updates Vite, Webpack, and their respective plugins.

### React
- Removed the `{ legacyRoot: true }` option from the `render` calls in tests
- Instead of `ReactDOM.render`, the recommended
```typescript
import { createRoot } from 'react-dom/client'
const root = createRoot(rootElement)
root.render(...)
```
method is used
- `@testing-library/react` is updated. Because of this, tests that dispatch multiple events in a single `act` wrapper needed to be updated. I think the reason for this is that based on [this](https://legacy.reactjs.org/docs/testing-recipes.html#act) docs page, DOM updates triggered by a user event are only reflected in the DOM after the `act` block completes.

### Vite
- The host and port of the HMR connection had to be explicitly specified (see the docs [here](https://vitejs.dev/config/server-options.html#server-hmr)), otherwise the Vite tried to connect on `ws://localhost:8001`, which was refused by the server
- `vite-plugin-html` (which injects some variables into `index.html` and `vscode-editor-iframe/index.html`) isn't updated, since the latest version (3.2.0) doesn't support injecting variables into multiples files (there's an unmerged open PR for it)

### Webpack
- The config object of `ForkTsCheckerWebpackPlugin` had to be updated


editor/src/utils/vite-hmr-config.ts
```diff
@@ -4,7 +4,7 @@
 // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
 if (import.meta.hot) {
   // this is written without the != null part to make sure Webpack (and Vite) recognizes and tree shakes it
-  import.meta.hot?.decline() // this _should_ be working but does not for some reason
+  import.meta.hot?.invalidate() // this _should_ be working but does not for some reason
```
gbalint, line 7: Is it still true that it doesn't work?

bkrmendy: It very much works, this method is the recommended replacement for `decline` as per the docs: https://vitejs.dev/guide/api-hmr.html#hot-decline

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -1180,7 +1180,6 @@ export function getComponentsRenderedWithReactDOM(
         ) {
           const propertyExpressionIdentifier: TS.Identifier = propertyAccessExpression.expression
           const propertyNameIdentifier: TS.Identifier = propertyAccessExpression.name
-          // Checks if this starts as ReactDOM.render(...).
```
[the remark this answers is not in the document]
bkrmendy: no recollection, https://github.com/concrete-utopia/utopia/pull/4047/commits/7654d44951436a73b914c2ea7ab3b8d6f441f1db puts back the commit

enidemi (approved):

PR #4048, opened 2023-08-04
feature(left-menu) pull navigator into left-panel, place it on canvas

enidemi:
**Fix:**
Moving the navigator into the left-menu and placing the left-menu where the navigator used to be.

**Commit Details:**
- replace navigator toggling with left-menu toggling in shortcuts
- update canvas-toolbar and inspector settings to toggle the left-menu instead of the navigator

Screenshot 2023-08-04 at 17 39 37


ruggi (commented):
this looks good to me but it's not taking into account the indicator arrows positioning, they will be in the wrong location on collapse. Either we merge this as is and make another PR later or embed that change into this PR (my preference would be on the latter, because imho is part of the same thing, but no blockers!)

gbalint (approved):

ruggi (approved):

PR #4054, opened 2023-08-04
Feature: Constrained Group Children

balazsbajorics:


ruggi:
imho this is good to go now: I fixed the merge conflicts _and_ added support for treating `max-content` for width and height as a constraint

seanparsons (approved):

bkrmendy (approved):

editor/src/components/canvas/commands/utils/group-resize-utils.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+export function transformConstrainedLocalFullFrameUsingBoundingBox(
+  newBoundingBox: Size,
+  currentBoundingBox: Size,
+  currentFrame: FrameWithAllPoints,
+  constrainedPoints: Array<FramePoint>,
+): FrameWithAllPoints {
+  // TODO verify that currentBoundingBox and currentFrame's dimensions match up
```
ruggi, line 59: note: I'm not exactly sure what this means - the `currentBoundingBox` relates to the group dimensions, while the `currentFrame` relates to the _child_ dimensions at the start of the interaction.
So I see alternative possible interpretations:
1. make sure that the computed sizes (at the caller side) _add up_ to the original bounding box
2. make sure that the final sizes (at the caller side) _add up_ to the final bounding box
3. possibly more?

Therefore I'm leaving this here until more light is cast on this

PR #4055, opened 2023-08-07
(perf) Remove unnecessary navigator items rerenders

### LINKED ISSUE #4056: Fix navigator performance while dragging groups

ruggi:
Fixes #4056 

**Problem:**
The editor is painfully slow when dragging stuff around (esp. groups) in the canvas.

**Fix:**
There are **multiple** reasons for this, but this PR addresses the navigator item rerenders, which should noticeably improve the situation.

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
@@ -548,6 +548,9 @@ export const NavigatorItemContainer = React.memo((props: NavigatorItemDragAndDro
     return EP.pathsEqual(firstSibling.elementPath, props.elementPath)
   }, [editorStateRef, props.elementPath])
 
+  // Note for future selves: watch out! This works because changing siblings triggers a re-render of the navigator,
+  // but if that were not to happen anymore, the references used here by the editorStateRef would not guarantee
+  // updated hook values.
   const isLastSibling = React.useMemo(() => {
     const siblings = MetadataUtils.getSiblingsOrdered(
       editorStateRef.current.jsxMetadata,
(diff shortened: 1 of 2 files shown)


bkrmendy (approved):

enidemi (approved):

editor/src/components/navigator/navigator-item/navigator-item-dnd-container.tsx
```diff
@@ -547,24 +535,32 @@ export const NavigatorItemContainer = React.memo((props: NavigatorItemDragAndDro
...
       return false
     }
 
     return EP.pathsEqual(firstSibling.elementPath, props.elementPath)
-  }, [metadata, elementPathTree, props.elementPath])
+  }, [editorStateRef, props.elementPath])
 
   const isLastSibling = React.useMemo(() => {
-    const siblings = MetadataUtils.getSiblingsOrdered(metadata, elementPathTree, props.elementPath)
+    const siblings = MetadataUtils.getSiblingsOrdered(
```
seanparsons, line 552: Does this work in the case where reparenting something in the navigator should cause this to change? Like if I dragged something from elsewhere to the end of a few siblings? Although as I write this I suspect it does work because any change to the hierarchy would cause the entire navigator to re-render.

I think this might be fine in the whole, but I'm a little wary of it because the dependencies aren't right locally and some change to how we render the navigator could cause this to render incorrectly. At the very least there should be a warning comment for our future selves.

ruggi: good shout! it seems to be working fine for me but I'll add a comment

ruggi: https://github.com/concrete-utopia/utopia/pull/4055/commits/db0af6abb8d711b7b17f7a4ff44b9a38ee30d97a and https://github.com/concrete-utopia/utopia/pull/4055/commits/15ef8b6c8b7584300a93e924172f3f25d7dce611

PR #4057, opened 2023-08-07
Error navigator item for invalid map overrides

gbalint:
**Problem:**
Show warning navigator items when the map override count is higher than the full number of children:

image

**Fix:**
I created a new kind of navigator item called `InvalidOverrideNavigatorItem` for these warnings.
Similarly to the conditional case, they receive an element path with an extra fake uid at the end, so they are addressable. Selection is disabled for these navigator items.


Rheeseyb (approved):

bkrmendy (approved):

editor/src/components/navigator/navigator-item/item-label.tsx
```diff
@@ -223,6 +226,8 @@ export const ItemLabel = React.memo((props: ItemLabelProps) => {
               ? colorTheme.dynamicBlue.value
               : isConditionalClause
               ? colorTheme.fg7.value
+              : isInvalidOverride
```
seanparsons, line 229: I feel like this should really be a function with regular if..else statements at this point, it's 4 ternaries deep. :(

seanparsons (approved):

PR #4059, opened 2023-08-07
Don't push intended bounds for group children with hug sides

### LINKED ISSUE #4058: don't update frame for group children with hug sides

ruggi:
Fixes #4058

**Problem:**

If a group child has `Hug x Hug` text children, when moving the group around the children points will be converted to fixed.

https://github.com/concrete-utopia/utopia/assets/1081051/afe8e7c1-7c0f-49a8-901e-80a05c820c9e

**Fix:**

The reality is that this happen for all `Hug x [anything]` children - so this PR branches the logic of the `getUpdateResizedGroupChildrenCommands` function so that it can discriminate between `moving` and `resizing` something, and skip hugging elements if in the `moving` situation.

The outcome is that if you move something its w/h frame points won't be touched but they will if you resize (which makes the most sense?).

https://github.com/concrete-utopia/utopia/assets/1081051/a26fe167-4098-4736-80ec-6134b5bcb9b5

Note: I've not added specific tests to this one because I think the behavior itself can be up for interpretation anyways and it would make sense to play around with it before deciding it's the right choice.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.tsx
@@ -221,6 +221,7 @@ export function absoluteResizeBoundingBoxStrategy(
                 pushIntendedBoundsAndUpdateGroups(
                   [{ target: selectedElement, frame: newFrame }],
                   'starting-metadata',
+                  'resize',
                 ),
               ]
             })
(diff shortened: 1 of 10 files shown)


Rheeseyb (approved):

Rheeseyb:
> The outcome is that if you move something its w/h frame points won't be touched but they will if you resize (which makes the most sense?).

Yeah I'm wondering about this. Maybe it _never_ makes sense to touch the w/h points for a hug x hug child when manipulating the group

seanparsons (approved):

PR #4060, opened 2023-08-07
Fix/inspector offsets for scroll-to and indicator arrows

ruggi:
**Problem:**

With the recent changes to the UI two things diverged:
- `scroll to` does not calculate the center of the viewport correctly (or, well, it does but it is not good given the floating inspector)
- the indicator arrows are unusable when overflowing on the right side of the canvas

**Fix:**

Take the inspector into the account and:
- adjust the scroll-to target `x`
- adjust the rightmost limit of the indicator arrows

https://github.com/concrete-utopia/utopia/assets/1081051/8089691f-2048-4376-9cfc-63f3a9b816e6


Rheeseyb (approved):
Nice one! We probably need to do the same with the code editor and navigator widths

enidemi (approved):

PR #4067, opened 2023-08-08
Prevent group selection/highlight when child of group selected.

seanparsons:
**Problem:**
When the child of a group is selected, clicking on the group itself should deselect the child and not select the group.

**Fix:**
A change to `replaceNonSelectablePaths` was made which relies on the selected views, however this would result in the mouse down deselecting the child and then the mouse up would select the group because at that point there were no selected views. The fix for this was to keep the selected views stable across the entire click for the logic that handles the selection change, so that the mouse up could see the selected views as they were before the interaction had started. With `useGetSelectableViewsForSelectModeFromSelectedViews` being very similar to another function with relies on editor state refs, but in this case takes the selected views as a parameter.

The above work uncovered an issue with how browser tests rely on some globals for the mouse position. The globals would carry over from one test to the next, which meant that any change in test order or the removal of a test might result in an unrelated test then failing. Now before each test the globals are reset to how they were before the editor is loaded.

**Commit Details:**
- `replaceNonSelectablePaths` now also excludes group like ancestors of the currently selected element.
- Added `useGetSelectableViewsForSelectModeFromSelectedViews` as an alternative to `useGetSelectableViewsForSelectMode` when the selected views should stay constant from the start.
- `isTargetElementHovered` value in `ReorderSliderControl` uses `containsPath` for some extra safety.
- `useSelectionArea` now uses `useGetSelectableViewsForSelectModeFromSelectedViews`.
- Added `resetGlobalPositions` utility function.
- In `mocha-setup-beforeall.js` trigger `resetGlobalPositions` before each test run.

### DIFF
--- a/editor/mocha-setup-beforeall.js
+++ b/editor/mocha-setup-beforeall.js
@@ -1,3 +1,6 @@
 beforeEach(function () {
   viewport.set(2200, 1000)
+  if ('resetGlobalPositions' in globalThis) {
+    globalThis.resetGlobalPositions()
+  }
 })
--- a/editor/src/components/canvas/controls/reorder-slider-control.tsx
+++ b/editor/src/components/canvas/controls/reorder-slider-control.tsx
@@ -57,7 +57,7 @@ export const ReorderSliderControl = controlForStrategyMemoized(
       Substores.highlightedHoveredViews,
       (store) => {
         const { hoveredViews } = store.editor
-        return target != null && hoveredViews.includes(target)
+        return target != null && EP.containsPath(target, hoveredViews)
       },
       'ReorderSliderControl isTargetElementHovered',
     )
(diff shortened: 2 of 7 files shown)


bkrmendy (approved):

editor/src/components/canvas/controls/reorder-slider-control.tsx
```diff
@@ -57,7 +57,7 @@ export const ReorderSliderControl = controlForStrategyMemoized(
       Substores.highlightedHoveredViews,
       (store) => {
         const { hoveredViews } = store.editor
-        return target != null && hoveredViews.includes(target)
+        return target != null && EP.containsPath(target, hoveredViews)
```
Rheeseyb, line 60: 😩

editor/mocha-setup-beforeall.js
```diff
@@ -1,3 +1,6 @@
 beforeEach(function () {
   viewport.set(2200, 1000)
+  if ('resetGlobalPositions' in globalThis) {
+    globalThis.resetGlobalPositions()
+  }
```
Rheeseyb, line 5: 👍

Rheeseyb (approved):

PR #4069, opened 2023-08-09
Only allow non-negative override

gbalint:
Fixes #[ticket_number]

**Problem:**

**Fix:**

**Commit Details:**
- Did [other things]

### DIFF
--- a/editor/src/core/shared/comment-flags.ts
+++ b/editor/src/core/shared/comment-flags.ts
@@ -97,9 +97,9 @@ function getUtopiaCommentFlag(c: Comment, type: UtopiaCommentFlagType): UtopiaCo
     return null
   }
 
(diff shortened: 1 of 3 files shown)


editor/src/core/shared/comment-flags.ts
```diff
@@ -97,9 +97,9 @@ function getUtopiaCommentFlag(c: Comment, type: UtopiaCommentFlagType): UtopiaCo
     return null
   }
 
-  function parseIntOrNull(value: string): number | null {
+  function parseNonNegativeIntOrNull(value: string): number | null {
```
seanparsons, line 100: Minor: But I think this rename makes sense.
```suggestion
  function parsePositiveInt(value: string): number | null {
```

seanparsons (approved):

PR #4074, opened 2023-08-10
Fix missing group trueups when changing directions during keyboard move

### LINKED ISSUE #4073: fix lost group trueups when changing directions in the same keyboard move interaction

ruggi:
Fixes #4073

**Problem:**

Changing directions during the _same_ keyboard group move interaction session causes the group to lose its trueuing up.

Kapture 2023-08-10 at 11 37 20

**Fix:**

The reason is that incremental accumulated presses with different keys inherit the not-changed portion of the frame from the _original_ frame. This PR changes this by merging incremental intended bounds based on their target element, so that changes on multiple points of the frame end up in the final intended bounds push.

Kapture 2023-08-10 at 11 38 20

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-resize.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-resize.spec.browser2.tsx
@@ -11,11 +11,18 @@ import {
   shiftCmdModifier,
   shiftModifier,
 } from '../../../../utils/modifiers'
-import { selectComponentsForTest } from '../../../../utils/utils.test-utils'
+import { selectComponentsForTest, wait } from '../../../../utils/utils.test-utils'
 import { selectComponents, setHighlightedView } from '../../../editor/actions/action-creators'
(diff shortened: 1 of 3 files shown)


bkrmendy (approved):

editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-resize.spec.browser2.tsx
```diff
@@ -256,6 +264,117 @@ describe('Keyboard Absolute Resize E2E', () => {
...
+
+    await keyDownArrowDownHoldingCmd()
+    await keyDownArrowDownHoldingCmd()
+    await keyDownArrowDownHoldingCmd()
+    await keyDownArrowDownHoldingCmd()
+    await keyDownArrowDownHoldingCmd()
+
+    clock.current.tick(KeyboardInteractionTimeout)
+
+    await wait(500)
```
seanparsons, line 333: Is this still needed with the line above?

ruggi: yup!

seanparsons (approved):

PR #4080, opened 2023-08-10
Don't set absolute position when grouping elements with hug

### LINKED ISSUE #4079: Don't set position when grouping text elements with hug contents

ruggi:
Fixes #4079 

**Problem:**

Grouping text elements with hug contents sets their w/h to a fixed number.

**Fix:**

Do the calculations for the group sizing during the true up for the new `wrap` mode, but skip them during the pin updates if having hug contents.

### DIFF
--- a/editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
+++ b/editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
@@ -5,26 +5,16 @@ import type {
   ElementInstanceMetadata,
   ElementInstanceMetadataMap,
 } from '../../../core/shared/element-template'
-import { localRectangle, roundRectangleToNearestWhole } from '../../../core/shared/math-utils'
 import type { CanvasRectangle, LocalRectangle, Size } from '../../../core/shared/math-utils'
 import {
-  MaybeInfinityCanvasRectangle,
   boundingRectangleArray,
-  canvasRectangle,
-  canvasVector,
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
```diff
@@ -242,25 +242,34 @@ function getUpdateResizedGroupChildrenCommands(
   let commandsToRun: Array<CanvasCommand> = []
   let updatedElements: Array<ElementPath> = []
 
-  Object.entries(updatedLocalFrames).forEach(([pathStr, frameAndTarget]) => {
-    const elementToUpdate = EP.fromString(pathStr)
-    const metadata = MetadataUtils.findElementByElementPath(editor.jsxMetadata, elementToUpdate)
+  Object.entries(updatedLocalFrames)
+    .filter(([path]) => {
```
seanparsons, line 246: Could fold this into the `forEach` to avoid creating an intermediary array.

seanparsons (approved):

editor/src/components/canvas/commands/push-intended-bounds-and-update-groups-command.ts
ruggi: done here https://github.com/concrete-utopia/utopia/pull/4080/commits/546100ee93205cb9587613aa33a093e565292af1

bkrmendy (approved):

PR #4098, opened 2023-08-16
Inspector section for group children constraints

### LINKED ISSUE #4099: Add inspector section for group children constraints

ruggi:
Fixes #4099 

**Problem:**

Now that we have constrained group children implemented, we need a UI piece in the inspector to set those constraints.

**Fix:**

- Add a new section in the inspector that shows up when the selection is consists (entirely) of group children
- In that new section, add two dropdowns to control the constraints of width and height
- Also add a pin controls component (only for tlbr) that controls side constraints

Screenshot 2023-08-16 at 12 29 39 PM

There is a "mixed" combo scenario, where some children have e.g. a constrained width but other don't. In that case, the dropdown will show a `Mixed` value, while the pin control will show the "bar", and clicking it once will set it for all elements.


editor/src/components/inspector/inspector.tsx
```diff
@@ -349,6 +351,16 @@ export const Inspector = React.memo<InspectorProps>((props: InspectorProps) => {
     'RootElementIndicator aRootElementIsSelected',
   )
 
+  const areGroupChildren = useEditorState(
```
seanparsons, line 354:
```suggestion
  const areChildrenOfAGroup = useEditorState(
```
Small beans, but the name makes me think this is checking the children to see if they are groups.

seanparsons (approved):

editor/src/components/inspector/inspector.tsx
ruggi: I went a little more with `allSelecteElementsdAreChildrenOfAGroup`

lankaukk (approved):

PR #4107, opened 2023-08-22
Add tests for static groups.

seanparsons:
**Problem:**
Currently there are no/limited tests for static positioned groups.

**Fix:**
Add more tests for static positioned groups.

**Commit Details:**
- Added tests for not absolutely positioned groups and flex contained groups.
- Changed `resizeElement` to throw an exception if the control could not be found.
- Extracted out `getResizeControl` and used it in tests where the expectation is that controls will not be shown.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize.test-utils.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize.test-utils.tsx
@@ -5,6 +5,14 @@ import type { EdgePosition } from '../../canvas-types'
 import { mouseDragFromPointWithDelta } from '../../event-helpers.test-utils'
 import type { EditorRenderResult } from '../../ui-jsx.test-utils'
 
+export function getResizeControl(
+  renderResult: EditorRenderResult,
+  edgePosition: EdgePosition,
+): HTMLElement | null {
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/canvas-strategies/strategies/absolute-resize.test-utils.tsx
```diff
@@ -5,6 +5,14 @@ import type { EdgePosition } from '../../canvas-types'
 import { mouseDragFromPointWithDelta } from '../../event-helpers.test-utils'
 import type { EditorRenderResult } from '../../ui-jsx.test-utils'
 
+export function getResizeControl(
+  renderResult: EditorRenderResult,
+  edgePosition: EdgePosition,
+): HTMLElement | null {
+  const testIdOfControl = `resize-control-${edgePosition.x}-${edgePosition.y}`
+  return renderResult.renderedDOM.queryByTestId(testIdOfControl)
```
bkrmendy, line 13: getByTestId automatically throws an error if it finds no elements or finds more than one element with the same test id

seanparsons: This function is also being used in `resizeElement` for the case when it's not null.

bkrmendy (approved):

ruggi (approved):

PR #4109, opened 2023-08-22
Support grouping conditionals

### LINKED ISSUE #4108: Support grouping conditionals

ruggi:
Fixes #4108 

**Problem:**

Trying to group conditional elements is not currently allowed.

**Fix:**

- Support wrapping conditionals with groups
- Support grouping conditionals with CMD+G
- Trim margins when grouping conditionals
- Explicitly don't allow grouping null conditionals (we can change this later on, when tackling zero-sized elements more specifically)

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
@@ -1,27 +1,36 @@
 import type { CSSProperties } from 'react'
+import { getLayoutProperty } from '../../../../core/layout/getLayoutProperty'
+import type { StyleLayoutProp } from '../../../../core/layout/layout-helpers-new'
 import type { PropsOrJSXAttributes } from '../../../../core/model/element-metadata-utils'
 import {
   MetadataUtils,
   getZIndexOrderedViewsWithoutDirectChildren,
 } from '../../../../core/model/element-metadata-utils'
 import { generateUidWithExistingComponents } from '../../../../core/model/element-template-utils'
+import {
+  removeMarginProperties,
+  removePaddingProperties,
+} from '../../../../core/model/margin-and-padding'
 import { arrayAccumulate, mapDropNulls } from '../../../../core/shared/array-utils'
-import { foldEither, isLeft, isRight, right } from '../../../../core/shared/either'
+import { isLeft, isRight, right } from '../../../../core/shared/either'
 import * as EP from '../../../../core/shared/element-path'
 import type { ElementPathTrees } from '../../../../core/shared/element-path-tree'
 import type {
   ElementInstanceMetadata,
   ElementInstanceMetadataMap,
   JSXAttributes,
+  JSXConditionalExpression,
   JSXElement,
   JSXElementChild,
   JSXElementLike,
 } from '../../../../core/shared/element-template'
 import {
   emptyComments,
+  isJSXConditionalExpression,
   isJSXElement,
   isJSXElementLike,
   isJSXFragment,
+  isNullJSXAttributeValue,
   jsExpressionValue,
   jsxAttributesFromMap,
   jsxElement,
@@ -35,15 +44,10 @@ import type {
   LocalRectangle,
   Size,
 } from '../../../../core/shared/math-utils'
-import { canvasVector } from '../../../../core/shared/math-utils'
-import {
-  canvasRectangle,
-  MaybeInfinityCanvasRectangle,
-  offsetRect,
-} from '../../../../core/shared/math-utils'
 import {
   boundingRectangleArray,
   canvasRectangleToLocalRectangle,
+  canvasVector,
   forceFiniteRectangle,
   isFiniteRectangle,
   isInfinityRectangle,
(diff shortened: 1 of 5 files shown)


editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -787,26 +797,69 @@ export function createWrapInGroupActions(
...
+          element: elementLike,
+          metadata: elementMetadata,
+        }
+
+      case 'JSX_CONDITIONAL_EXPRESSION':
+        // Remove any margin properties from the branches.
+        let conditional: JSXConditionalExpression = { ...element }
+        if (isJSXElementLike(conditional.whenTrue)) {
+          conditional.whenTrue = modify(
+            elementPropsOptic,
```
seanparsons, line 820: If you change this line like this you wont need the containing if statement.
```suggestion
            fromField('whenTrue').compose(elementPropsOptic),
```

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -787,26 +797,69 @@ export function createWrapInGroupActions(
...
         element == null ||
         isLeft(element) ||
-        !isJSXElementLike(element.value)
+        !elementCanBeAGroupChild(element.value)
       ) {
         throw new Error(
           `Invariant violation: ElementInstanceMetadata.element found for ${EP.toString(
             p,
-          )} was null or Left or not JSXElement`,
+          )} cannot be a group child`,
```
seanparsons, line 857: Nice.

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -687,6 +683,22 @@ export function convertGroupToFrameCommands(
...
+  element: JSXElementChild | null,
+): element is GroupChildElement {
+  if (element == null) {
+    return false
+  }
+  return (
+    isJSXElementLike(element) ||
+    (isJSXConditionalExpression(element) &&
+      // do not allow grouping zero-sized conditionals
+      !(isNullJSXAttributeValue(element.whenTrue) && isNullJSXAttributeValue(element.whenFalse)))
```
Rheeseyb, line 698: Ooooh this is a tricky one. So this is only checking that one side of the conditional is not null, but isn't checking that the active branch is not null. Even if we were to check that the active branch is not null, that doesn't protect us from the active branch being switched. Why do we actually need the not-null check? I'm wondering if it would be better to support null children now 🤔

ruggi: heh you're right and this should use the active branch check actually. This is not a requirement tho, it's just an additional defensive condition I added deliberately, but nothing is preventing from having zero-sized conditionals in groups (they will just look as invalid in the navigator). Shall I just remove the condition and call it a day?

Rheeseyb: If that doesn't break anything then yeah you might as well

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -687,6 +682,17 @@ export function convertGroupToFrameCommands(
...
 
+export type GroupChildElement = JSXElementLike | JSXConditionalExpression
+
+export function elementCanBeAGroupChild(
+  element: JSXElementChild | null,
+): element is GroupChildElement {
+  if (element == null) {
+    return false
+  }
+  return isJSXElementLike(element) || isJSXConditionalExpression(element)
```
gbalint, line 693: Shouldn't we check if the active branch  of the conditional expression can be a group child?

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
```diff
@@ -925,24 +973,50 @@ export function createPinChangeCommandsForElementBecomingGroupChild(
...
+          ? elementMetadata.element.value.whenTrue
+          : elementMetadata.element.value.whenFalse
+      return isJSXElement(branch)
+        ? commandsForPinChangeCommandForElementBecomingGroup(
+            expectedPath,
+            branch.props,
+            childLocalRect,
+            newLocalRectangleForGroup,
+          )
+        : []
```
gbalint, line 1014: I think we should not arrive to this case because we should check earlier what is the active branch of the conditional (see my comment above)

editor/src/components/canvas/canvas-strategies/strategies/group-conversion-helpers.ts
ruggi: I think this would deserve some thinking: you can end up with a inactive branch that is not a valid child, and flipping the conditional would end up with an invalid group anyways, so it would end up in the same scenario 🤔

ruggi: makes sense. would this work? [`47cfc23` (#4109)](https://github.com/concrete-utopia/utopia/pull/4109/commits/47cfc23d9f578b0225d4101078fe656ff0d73d0e)

gbalint (approved):

PR #4110, opened 2023-08-22
Do not include the absolute children outline for groups.

seanparsons:
**Problem:**
The absolute children outline for groups matches the outline for the group itself, so if there's any lag in rendering it then it looks a bit strange.

**Fix:**
Exclude group children from the absolute children outline.

**Commit Details:**
- Added `testId` to `OutlineControlProps`.
- `OutlineControl` now includes a `data-testid` value so it can be easily identified.
- `AbsoluteChildrenOutline` filters out selected views which are group like.
- Changed some types from `Array<ElementPath>` to `ReadonlyArray<ElementPath>` to make them more compatible with `arrayAccumulate`.


ruggi (approved):
\o\

editor/src/components/canvas/controls/select-mode/simple-outline-control.tsx
```diff
@@ -28,26 +28,32 @@ export const MultiSelectOutlineControl = React.memo<MultiSelectOutlineControlPro
...
-        ...localSelectedElements.map((path) => (
-          <OutlineControl
-            key={EP.toString(path)}
-            targets={[path]}
-            color='primary'
-            outlineStyle='solid'
-          />
-        )),
+        ...localSelectedElements.map((path) => {
+          const varSafeId = `multiselect-element-outline-${EP.toVarSafeComponentId(path)}`
```
Rheeseyb, line 38: I don't believe there's any need for this to be "varSafe". I actually don't think we need `toVarSafeComponentId` at all anymore, as that's left over from the old old Utopia

Rheeseyb (approved):

editor/src/components/canvas/controls/select-mode/simple-outline-control.tsx
seanparsons: ORLY?

seanparsons: Removed in b26e6a405.

Rheeseyb: Yeah, it was used for creating variable names that were valid in JS. I don't recall specifically why.

PR #4120, opened 2023-08-28
Remix support initial PR

gbalint:
**Description:**
This PR is the first step to implement Remix support in Utopia.

- I added a feature switch called 'Remix support'
- New `RemixContainer` component in `utopia-api` which you can use to insert a remix project into a utopia project
- Renderer switches `RemixContainer` components to the `RemixContainerComponent` implementation, which uses `UtopiaRemixRootComponent` to render a remix project.
- `UtopiaRemixRootComponent` uses the remix lib to render a remix project.
- To keep the PR simple, we provide a dummy routes array to the remix project, which just renders a placeholder text.


bkrmendy (approved):
🎉

ruggi (approved):

PR #4122, opened 2023-08-28
True up groups when editing code

### LINKED ISSUE #4121: True up groups when editing via vscode

ruggi:
Fixes #4121 

**Problem:**

Groups are not trued up when the changes come from editing the code rather than the canvas.

**Fix:**

- When editor decides if there are trueups to be committed, as per the array in the editor state, _also_ include a list of groups that need trueuing up. This solution will allow us to also **drop the trueups scattered throughout the code**. I don't think it should live in this PR though but in an incremental one, as to keep the scope reasonable.
- When that trueup happens, changes coming from the editor are picked up for trueups as well
- To avoid having the caret jump around (because a changed element event happens), the vscode extension now reveals the code selection only if it's not already matching the current selection's caret position.

**Notes:**
- There's still the issue with undo steps and trueups which become part of the stack (and the incidental code update), but I think it should be tackled separately.


seanparsons (commented):
Looks good to me, with the proviso that we're happy to not be able to manually create groups that aren't trued up.

seanparsons (approved):
My comment from before should've been an approve.

ruggi:
closing this for now, might get useful in the future tho

PR #4126, opened 2023-08-29
Generating route manifest for Remix

gbalint:
**Problem:**
Remix needs a route manifest, which is a flat representation of routes and their associated files and parents, parsed from the file paths.
I could not directly use Remix code to generate this, because:
- The creation of the manifest is an internal part of Remix, not part of its API.
- The manifest is constructed compile-time in Remix, because it is based on the source files of the user's project.
- There is no use-case in remix to read the source files from memory during runtime, Remix is not an IDE in the browser!
- Route manifest creation uses `node.fs` and `node.path` to read files, and these packages can not run in the browser.

**Fix:**
Copy paste the necessary functions from Remix, and adapt them to our needs:
- Instead of using node.fs, use the injected FileOps object which contain implementations for the necessary file system functions
- Use 'path-browserify' instead of node.path 
- Do not rely on code evaluation, we are going to use our utopia parsed model later to find the exported components from route modules.


bkrmendy (approved):

editor/package.json
```diff
@@ -171,6 +172,7 @@
     "object-path-immutable": "3.0.0",
     "onigasm": "2.2.4",
     "path": "0.12.7",
+    "path-browserify": "^1.0.1",
```
Rheeseyb, line 175: Can we remove the `^` from the new dependencies please

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -10,3 +21,66 @@ export const DefaultFutureConfig: FutureConfig = {
...
+  if (routesFromRemix == null) {
+    return null
+  }
+  return patchRemixRoutes(routesFromRemix)
+}
+
+function patchRemixRoutes(routesFromRemix: RouteManifest<ConfigRoute> | null) {
+  const routesFromRemixWithRoot: RouteManifest<ConfigRoute> = {
+    ...routesFromRemix,
+    root: { path: '', id: 'root', file: 'root.js', parentId: '' },
```
Rheeseyb, line 67: Is this hardcoding expected / required, or are we going to remove this at some point?

editor/package.json
gbalint: Fixed

editor/src/third-party/remix/flat-routes.ts
```diff
@@ -0,0 +1,542 @@
+/* eslint-disable */
+
+// Copy pasted and adapted from Remix: https://github.com/remix-run/remix/blob/main/packages/remix-dev/config/flat-routes.ts
```
Rheeseyb, line 3: Can you update these URLs to use the permalink (i.e. including the commit SHA) URLs please

Rheeseyb (approved):

PR #4135, opened 2023-08-31
Fix resizing pinned groups when resizing their parent

### LINKED ISSUE #4134: Fix resizing pinned groups when resizing their parent

ruggi:
Fixes #4134 

**Problem:**

If a group has position pinned to its parent, resizing the parent won't update the group.
Kapture 2023-08-31 at 12 45 42

**Fix:**

The bug is caused by the dimensions of the group component being forced (https://github.com/concrete-utopia/utopia/blob/564fee9f7e89b60209607898fdb5b5b75c422b04/editor/src/core/es-modules/package-manager/group-component.tsx#L126-L132), so instead of doing expensive calculations there this PR fixes it by mirroring the resize to the groups which are children of the interaction targets, similarly to what we do for fragments.

Instead of adding the special logic to the resize function directly, I included it as part of the retargeting + flattening logic, so we can reuse it transparently.

Kapture 2023-08-31 at 12 46 43

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
@@ -126,7 +126,6 @@ const preventOnRootElements: (metaStrategy: MetaCanvasStrategy) => MetaCanvasStr
     customStrategyState: CustomStrategyState,
   ): Array<CanvasStrategy> => {
     const selectedElements = flattenSelection(
-      canvasState.startingMetadata,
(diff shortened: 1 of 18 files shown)


ruggi:
For reviewers: this commit [`0db27be` (#4135)](https://github.com/concrete-utopia/utopia/pull/4135/commits/0db27bea552bb09f57838bc4e40826d9bb84ab20) is the relevant one.

bkrmendy (approved):

seanparsons (approved):

PR #4138, opened 2023-08-31
Handle shifting top level elements.

seanparsons:
**Problem:**
When grouping two elements, the selection is lost.

**Cause:**
The logic that attempts to copy over old UIDs from previous parse results has an issue because an import has been added, which shifts the array of top level elements slightly. As a result of that when the `UtopiaJSXComponent` is hit in the array, it's being compared to something which isn't one of those but likely is something like some unparsed code. This means all the old UIDs are effectively thrown away.

**Fix:**
When running through the top level elements, we now mostly ignore (at least when it comes to for comparisons) anything that doesn't contain UIDs.

**Commit Details:**
- Changed how `fixTopLevelElementUIDs` handles matching up entities by having it (for that purpose at least) ignore anything that does not contain UIDs.

### DIFF
--- a/editor/src/core/workers/parser-printer/uid-fix.spec.ts
+++ b/editor/src/core/workers/parser-printer/uid-fix.spec.ts
@@ -8,6 +8,8 @@ import type {
   JSXElementChild,
   TopLevelElement,
 } from '../../shared/element-template'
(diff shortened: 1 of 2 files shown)


ruggi (approved):

Rheeseyb (approved):
Nice!

PR #4145, opened 2023-09-04
Alternative canvas elements lookup.

seanparsons:
**Problem:**
When clicking on text that overflows outside of elements, selection can be lost when an ancestor of the text containing element is focused.

**Cause:**
This originates because of a discrepancy in two internal bits of the editor:
- `firstAncestorOrItselfWithValidElementPath`: Given a browser Element look up the frames for the paths attached to that and see if the mouse pointer is inside those frames. This is underpinned by `document.elementsFromPoint` which only checks the regular HTML element bounding rectangles.
- `getFramesInCanvasContextUncached`: Get all the frames of everything in the metadata, then the result of this is scanned to find which the mouse pointer is inside, including the extended bounds from overflowing text.

The first of the two above was called from a check which clears the focus in this case when the overflowing text is clicked on. As a result of that, then the second function gets called with the now trimmed metadata and doesn't find anything to select as the focus has now been cleared and the element with the overflowing text is now not present there.

**Fix:**
Created an alternative function for `document.elementsFromPoint` which also checks the bounds of text nodes within HTML elements. Which makes it consistent in behaviour to the lookups which check the content bounds including that of the text which is done exclusively from the metadata.

**Note:**
I did some cursory testing of the performance of this new function against the Before I Go project using some `console.log` lines and it takes about 4ms in that to run. However, this function does only return elements within the `CanvasContainer` component whereas `document.elementsFromPoint` was returning everything in the hierarchy from the `html` element down that was under the mouse, so we'll save a little bit from not checking the paths on the approximately 14 HTML elements above `CanvasContainer`.

**Commit Details:**
- Implemented `canvasElementsUnderPoint`, which checks the bounds of the text within an element.
- Used `canvasElementsUnderPoint` in place of `document.elementsFromPoint` in `dom-lookup.ts`.

### DIFF
--- a/editor/src/components/canvas/dom-lookup.ts
+++ b/editor/src/components/canvas/dom-lookup.ts
@@ -22,6 +22,7 @@ import * as EP from '../../core/shared/element-path'
 import { getPathsOnDomElement } from '../../core/shared/uid-utils'
 import Canvas, { TargetSearchType } from './canvas'
 import type { CanvasPositions } from './canvas-types'
+import { CanvasContainerID } from './canvas-types'
 import type { AllElementProps } from '../editor/store/editor-state'
 import Utils from '../../utils/utils'
 import { memoize } from '../../core/shared/memoize'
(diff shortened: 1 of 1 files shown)


seanparsons:
Closing because this doesn't look like an appealing approach.

PR #4162, opened 2023-09-07
Render Remix projects

bkrmendy:
## Description
This PR adds the code that parses and renders remix projects on the storyboard.

The navigator and editing interactions are out of scope for this PR.

### Details
- RemixDerivedDataFactory is injected into `deriveState`, to circumvent the `not-from-ts-to-tsx` check
- `getControlStyles` is refactored so that a circular dependency problem is avoided
- `buildResult` in `Component` in `createComponentRendererComponent` is initialized to the result of calling `buildComponentRenderResult` instead of `null` (https://github.com/concrete-utopia/utopia/pull/4162/commits/3f5b57534099133e3bacfd6711f314b511d0fd7c)


editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -99,3 +128,405 @@ export function getRoutesFromRouteManifest(
     return []
   }
 }
+
+export function createAssetsManifest(routes: RouteManifest<EntryRoute>): AssetsManifest {
+  return {
+    entry: { imports: [], module: 'TODO' },
```
Rheeseyb, line 134: What is the actual TODO here? I don't think this is something we're tracking anywhere

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -99,3 +128,405 @@ export function getRoutesFromRouteManifest(
...
+  }
+}
+
+export type RouteModuleWithFilePath = RouteModule & {
+  filePath: string
+}
+
+export interface RouteModulesWithFilePaths {
+  [routeId: string]: RouteModuleWithFilePath
+}
```
Rheeseyb, line 147: Doesn't look like these two are used anymore?

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -99,3 +128,405 @@ export function getRoutesFromRouteManifest(
...
+  routeModulesToRelativePaths: RouteModulesWithRelativePaths
+  routingTable: RemixStaticRoutingTable
+}
+
+export function getDefaultExportNameAndUidFromFile(
+  projectContents: ProjectContentTreeRoot,
+  filePath: string,
+): { name: string; uid: string } | null {
+  const file = getProjectFileByFilePath(projectContents, filePath)
+  if (file == null || file.type != 'TEXT_FILE' || file.lastParseSuccess == null) {
```
Rheeseyb, line 177: Are we deliberately using `lastParseSuccess` here? It means that this function will return a result even if the current file is a `ParseFailure`, which I don't _think_ was the intention here?

editor/src/components/canvas/remix/remix-utils.tsx
bkrmendy: thanks! removed them in [192aafb](https://github.com/concrete-utopia/utopia/pull/4162/commits/192aafb3d2ab6c6136caf78643c15d6004f6ec02)

editor/src/components/canvas/remix/remix-utils.tsx
bkrmendy: should it be `parsed`?

Rheeseyb: `file.fileContents.parsed`

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -99,3 +128,397 @@ export function getRoutesFromRouteManifest(
...
+  for (const child of children) {
+    const path = findPathToOutlet(isOutlet, child)
+    if (path != null) {
+      return [parentUid, ...path]
+    }
+  }
+  return null
+}
+
+export function isRemixOutletElement(
```
Rheeseyb, line 245: I believe we agreed that this should live in either `element-metadata-utils.ts` or `project-file-utils.ts`

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -99,3 +128,397 @@ export function getRoutesFromRouteManifest(
...
+  }
+
+  const executionScope = executionScopeCreator(projectContents)
+
+  const nameAndUid = getDefaultExportNameAndUidFromFile(projectContents, filename)
+
+  return {
+    executionScopeCreator: executionScopeCreator,
+    loader: executionScope.scope['loader'] as LoaderFunction | undefined,
+    action: executionScope.scope['action'] as ActionFunction | undefined,
```
Rheeseyb, line 443: I think we should attach a `FIXME` to these, because really we should be creating the `executionScope` at the point where we use the `loader` and `action` like we do for the module's default export component

editor/src/components/canvas/remix/remix-utils.tsx
bkrmendy: https://github.com/concrete-utopia/utopia/pull/4162/commits/cddc103ffd7070810f4ed2a17a35d4afc97a1f4d

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -99,3 +128,397 @@ export function getRoutesFromRouteManifest(
...
+
+  const rootJSRootElement = getDefaultExportedTopLevelElement(rootJsFile)
+  if (rootJSRootElement == null) {
+    return null
+  }
+
+  const routesByParentId = groupRoutesByParentId(routeManifest)
+  const routes: DataRouteObject[] = createClientRoutes(
+    routeManifest,
+    {},
```
Rheeseyb, line 482: I think this is either wrong, or missing a TODO, because `createClientRoutes` takes a `routeModulesCache` for this param, which is used when trying to determine if a module needs re-evaluating, so since this is always empty here (and never mutated) the re-evaluation function will never be able to retrieve the module

editor/src/components/canvas/remix/remix-utils.tsx
bkrmendy: added the FIXME comment in https://github.com/concrete-utopia/utopia/pull/4162/commits/cddc103ffd7070810f4ed2a17a35d4afc97a1f4d

editor/src/components/editor/store/remix-derived-data.tsx
```diff
@@ -0,0 +1,109 @@
...
+  mutableContext: { current: MutableUtopiaCtxRefData }
+  topLevelComponentRendererComponents: { current: MapLike<MapLike<ComponentRendererComponent>> }
+  resolvedFiles: { current: MapLike<Array<string>> }
+  resolvedFileNames: { current: Array<string> }
+} = {
+  mutableContext: { current: {} },
+  topLevelComponentRendererComponents: { current: {} },
+  resolvedFiles: { current: {} },
+  resolvedFileNames: { current: [] },
+}
```
Rheeseyb, line 49: Do we still need this?

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -99,3 +128,397 @@ export function getRoutesFromRouteManifest(
...
+  filePath: string,
+): { name: string; uid: string } | null {
+  const file = getProjectFileByFilePath(projectContents, filePath)
+  if (file == null || file.type != 'TEXT_FILE' || file.lastParseSuccess == null) {
+    return null
+  }
+
+  const defaultExportName =
+    file.lastParseSuccess.exportsDetail.find(
+      (e): e is ExportDefaultFunctionOrClass => e.type === 'EXPORT_DEFAULT_FUNCTION_OR_CLASS',
```
seanparsons, line 182: We could do with one of our `isX` functions here as then it would be just:
```suggestion
      isExportDefaultFunctionOrClass
```

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -99,3 +128,397 @@ export function getRoutesFromRouteManifest(
...
+
+  return (
+    file.fileContents.parsed.topLevelElements.find(
+      (t): t is UtopiaJSXComponent =>
+        t.type === 'UTOPIA_JSX_COMPONENT' && t.name === defaultExportName,
+    )?.rootElement ?? null
+  )
+}
+
+function addLoaderAndActionToRoute(
```
seanparsons, line 222: Tiny:
```suggestion
function addLoaderAndActionToRoutes(
```

editor/src/components/canvas/remix/remix-utils.tsx
seanparsons: Do we really want to add library specific stuff to those?

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -99,3 +128,397 @@ export function getRoutesFromRouteManifest(
...
+      route.loader = loader
+    } else {
+      addLoaderAndActionToRoute(route.children ?? [], routeId, loader, action)
+    }
+  })
+}
+
+function findOutletAmongJSXElementChildren(
+  parentUid: string,
+  isOutlet: (e: JSXElementChild) => boolean,
```
seanparsons, line 240: Seems like this is for finding some subset of all the outlets?
```suggestion
  outletPredicate: (e: JSXElementChild) => boolean,
```

editor/src/components/editor/store/remix-derived-data.tsx
bkrmendy: Cleaning this up would happen in a separate PR, I removed the unused props in https://github.com/concrete-utopia/utopia/pull/4162/commits/fbac022f2c91c1b9dcd98b423f0cc6712297d3a5

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -99,3 +128,397 @@ export function getRoutesFromRouteManifest(
...
+  const file = getProjectFileByFilePath(projectContents, filePath)
+  if (file != null && isTextFile(file) && isParseSuccess(file.fileContents.parsed)) {
+    return isRemixOutletAgainstImports(element, file.fileContents.parsed.imports)
+  } else {
+    return false
+  }
+}
+
+function findPathToOutlet(
+  isOutlet: (e: JSXElementChild) => boolean,
```
seanparsons, line 253: Same thing here:

```suggestion
  outletPredicate: (e: JSXElementChild) => boolean,
```

editor/src/components/canvas/remix/remix-utils.tsx
bkrmendy: a "ref" (global var in this case) is passed down in https://github.com/concrete-utopia/utopia/pull/4162/commits/fbac022f2c91c1b9dcd98b423f0cc6712297d3a5, to be cleared up in a future PR

editor/src/core/model/project-file-utils.ts
```diff
@@ -154,6 +154,48 @@ export function isSceneFromMetadata(elementInstanceMetadata: ElementInstanceMeta
...
+export function isRemixSceneAgainstImports(element: JSXElementChild, imports: Imports): boolean {
+  return isGivenUtopiaAPIElement(element, imports, 'RemixScene')
+}
+
+export function isRemixOutletAgainstImports(element: JSXElementChild, imports: Imports): boolean {
+  if (!isJSXElement(element)) {
+    return false
+  }
+
+  const utopiaAPI = imports['@remix-run/react']
```
Rheeseyb, line 166: `utopiaAPI`? 😁

editor/src/components/canvas/remix/remix-rendering.spec.browser2.tsx
```diff
@@ -49,3 +170,94 @@ describe('Remix content with feature switch off', () => {
...
+}
+
+async function expectRemixSceneToBeRendered(
+  editor: EditorRenderResult,
+  textContent: string = DefaultRouteTextContent,
+) {
+  const [remixScene] = await editor.renderedDOM.findAllByTestId(REMIX_SCENE_TESTID)
+  expect(remixScene).toBeDefined()
+
+  const [rootText] = await editor.renderedDOM.findAllByText('This is root!')
```
Rheeseyb, line 258: Minor point, it seems like the `This is root!` should be a constant in this file

editor/src/components/canvas/remix/utopia-remix-root-component.tsx
```diff
@@ -1,66 +1,116 @@
...
+        continue
+      }
+
+      const relativePath =
+        remixDerivedDataRef.current.routeModulesToRelativePaths[routeId].relativePath
+
+      const defaultComponent = (componentProps: any) =>
+        value
+          .executionScopeCreator(projectContentsRef.current)
+          .scope[nameAndUid.name]?.(componentProps) ?? <React.Fragment />
```
seanparsons, line 52: Why a Fragment?

editor/package.json
```diff
@@ -196,7 +196,7 @@
     "react-merge-refs": "1.0.0",
     "react-onclickoutside": "6.9.0",
     "react-popper": "2.2.3",
-    "react-router": "^6.15.0",
+    "react-router": "6.14.2",
```
balazsbajorics, line 199: too bad we can't have comments here, it'd be important to note that this version needs to match the version used by remix

editor/src/components/canvas/remix/remix-utils.tsx
gbalint: I think this is really just used to find outlets, so I'm not sure why is this function injected at all. @bkrmendy ?

seanparsons (approved):
Not a blocker for the PR and I can imagine it came up late in the work, but it would've been nicer to separate out the circular fix changes as they've bloated this PR a bit.

editor/src/components/canvas/remix/remix-utils.tsx
gbalint: So I believe the function is just passed in so we don't have to pass in all the information necessary to find outlets. But then basically this is a general `findAmongJSXElementChildren`, and the `isOutlet` can be any kind of predicate.

gbalint: Fixed in caddafb09445da6df2a48879e0cc405de3be0994

editor/src/components/canvas/remix/utopia-remix-root-component.tsx
gbalint: We added a task for this to the paper doc for later.

PR #4164, opened 2023-09-07
Standardise on the Git checksum.

seanparsons:
**Problem:**
Images when added to the project and pushed to a Github repository reappear as needing updates from the repository.

**Cause:**
Git uses a wrapper around a regular SHA1 checksum but in the handling for updates there were a mix of that checksum and the stock SHA1. Since those produce different values, some situations where the checksum was compared between a value produced by Github and a value produced by Utopia, they would be treated as different even if their content was identical.

**Fix:**
One part of the fix was to remove some duplication as there were more than one function called `imageFile`, with one missing the sha field, causing it to always be `undefined`. Now that field is supplied in every place that we can to ensure that even a newly added file that matches one in the repository will not cause a clash.

The `gitBlobChecksumX` functions were added to make it easier to produce the correct checksum value and `getSHA1Checksum` was replaced by one of those functions in the codebase.

Additionally these fixes were implemented as they surfaced while testing the above:
- The recent fix for parsing the incoming code from Github was carried through into the "update against Github" flow for the case when the project is already connected to a repository.
- Updating against Github also now causes new assets/images to be saved to Utopia.
- If an update is made to the project and committed before the editor gets a chance to trigger a refresh, then the `lastRefreshedCommit` value ends up out of date.

**Commit Details:**
- Moved all the utility functions for `ProjectFile` types to `project-file-types.ts`.
- Added `gitBlobSha` to `SaveAsset`.
- `updateProjectAgainstGithub` now parses the branches latest content so that
  like `updateProjectWithBranchContent` the updates get passed through into VS Code.
- Renamed `inferGitBlobChecksum` to `gitBlobChecksumFromBuffer`.
- Added `gitBlobChecksumFromBase64` and `gitBlobChecksumFromText`.
- `saveAssetsToProject` now saves the assets which have changed from the old project
  contents to the new project contents.
- `updateProjectAgainstGithub` now saves newly added assets through the use of
  `saveAssetsToProject`.
- `updateProjectAgainstGithub` updates `lastRefreshedCommit` so that an update
  done before a refresh maintains consistency.
- `mergeProjectContentsTree` tries to carry over the parsed values for the branch
  content.
- Added `gitBlobSha` to `AssetResult` and `ImageResult`.
- `extractAsset` and `extractImage` now produce the appropriate git blob checksum.


ruggi (approved):

editor/src/components/navigator/left-pane/github-pane/index.tsx
```diff
@@ -484,9 +493,17 @@ const RemoteChangesBlock = () => {
   )
   const triggerUpdateAgainstGithub = React.useCallback(() => {
     if (repo != null && branch != null && commit != null) {
-      void updateProjectAgainstGithub(dispatch, repo, branch, commit)
+      void updateProjectAgainstGithub(
```
balazsbajorics, line 496: not related, but what's with this `void` guy here?

editor/src/core/shared/file-utils.ts
```diff
@@ -78,48 +89,89 @@ export function extractText(file: File): Promise<TextResult> {
   })
 }
 
-export function extractAsset(file: File): Promise<AssetResult> {
-  return new Promise((resolve, reject) => {
+export async function extractAsset(file: File): Promise<AssetResult> {
```
balazsbajorics, line 92: why make it async? are you using `await` inside this function body? or to bring it in line with some other function's signature?

balazsbajorics (approved):

editor/src/components/navigator/left-pane/github-pane/index.tsx
seanparsons: `updateProjectAgainstGithub` is async from memory, so if we don't add that eslint complains.

editor/src/core/shared/file-utils.ts
seanparsons: It was already async because it returned a `Promise`, adding that got rid of an eslint warning.

editor/src/components/navigator/left-pane/github-pane/index.tsx
balazsbajorics: interesting!

PR #4165, opened 2023-09-07
Disable link following on click in the text editor

gbalint:
**Problem:**
During text editing anchor elements are clickable, and on click the browser tries to navigate away to the target url.

**Fix:**
I applied a preventDefault to the click event in the text editor. This felt a little dangerous, because we heavily rely on the default browser behavior handling mouse events for text editing, e.g. clicking to set cursor position, drag to select a range, etc.
However, it seems the click event is not necessary for these, and the text editing features only rely on the more basic (mouseup, mousedown) etc mouse events.

### DIFF
--- a/editor/src/components/text-editor/text-editor.spec.browser2.tsx
+++ b/editor/src/components/text-editor/text-editor.spec.browser2.tsx
@@ -1627,6 +1627,32 @@ describe('SWITCH_EDITOR_MODE', () => {
   })
 })
 
+describe('Mouse behavior during text editing', () => {
+  const beforeUnloadCallback = () => {
+    throw new Error('Should not navigate away on click')
+  }
+  before(() => window.addEventListener('beforeunload', beforeUnloadCallback))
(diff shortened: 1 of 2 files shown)


editor/src/components/text-editor/text-editor.spec.browser2.tsx
```diff
@@ -1627,6 +1627,32 @@ describe('SWITCH_EDITOR_MODE', () => {
   })
 })
 
+describe('Mouse behavior during text editing', () => {
+  const beforeUnloadCallback = () => {
+    throw new Error('Should not navigate away on click')
+  }
+  before(() => window.addEventListener('beforeunload', beforeUnloadCallback))
+  after(() => window.removeEventListener('beforeunload', beforeUnloadCallback))
```
ruggi, line 1635: nice trick

ruggi (approved):

bkrmendy (approved):

PR #4167, opened 2023-09-08
Run the DOM walker on observed mutations (in Live mode only)

Rheeseyb:
**Problem:**
If the user's application includes state and interactions that change the rendered output, and they switch to using Live mode to trigger those changes, the captured metadata (and thus the Navigator) will be stale until the user switches to Select mode (or does anything else to dispatch an action).

**Fix:**
The underlying issue here is that we only run the DOM walker in response to a dispatched action, and no actions are dispatched whilst interacting with the app in Live mode. We do however have observers that will act on the changes introduced by interacting with the app, so we can use those to dispatch a new `RUN_DOM_WALKER` action whenever such a change is observed.

At first I experimented with using that action whilst in Select mode as well, as I'm 99% sure there are cases where we'd want that to happen, but unfortunately this has very real performance implications (it results in the DOM walker being run twice as often whilst in Select mode). I'm certain we could resolve that by attempting to marry up the intended walks to prevent duplicate walks from occurring, but doing so is non-trivial, as different paths are invalidated, and we can't just assume that we can skip a DOM walk if the previous action triggered one (since that would just negate the point of the `RUN_DOM_WALKER` action anyway), so we'd need some way of knowing beforehand that nothing had actually changed since the last walk.

So since the main reason for this work was to focus on Live mode, we only dispatch the actions if the editor is in Live mode.


bkrmendy (approved):

gbalint:
I believe this should have a test (do a change in live mode and check if the metadata is updated, similarly to tests in `dom-walker.spec.browser2`)

gbalint (approved):

seanparsons (approved):

editor/src/components/canvas/dom-walker.spec.browser2.tsx
```diff
@@ -406,6 +409,166 @@ describe('Capturing closest offset parent', () => {
...
+
+    const buttonToClick = renderResult.renderedDOM.getByTestId('click-me')
+    const buttonToClickBounds = buttonToClick.getBoundingClientRect()
+    const clickPoint = {
+      x: buttonToClickBounds.width / 2,
+      y: buttonToClickBounds.height / 2,
+    }
+
+    await mouseClickAtPoint(buttonToClick, clickPoint)
+    await wait(20)
```
balazsbajorics, line 512: can this be await(0) or awaiting a requestAnimationFrame? I have a feeling we are doing us a disservice by putting non-zero awaits into the test code – maybe it's superstition on my part, but it certainly something that can become flaky in the future

Rheeseyb: tbh reducing the setTimeout to zero hadn't occurred to me. I've tried that and it works perfectly (including reproducing the issue that this PR fixes), so I've made that change 👍

PR #4168, opened 2023-09-08
No More Baked Absolute Window Coordinates in Tests + No More Cyclic Dependency Issues

balazsbajorics:
I'm sorry that I'm fixing many things at one but basically you start fixing one of these issues and the rest just comes along.

**Problem:**
1. We have a small handful of BAD BAD BAD tests that capture some coordinates measured as clientX clientY, meaning any shift on the canvas scroll position, or an offset introduced by a new UI or anything will break these tests
2. We never loaded the real editor css files for the editor so `body` had an 8px margin around it (among other sins). meaning these coordinates were all off by 8 as well
3. sometimes we run into a circular dependency that is _only_ a problem in Karma but not in the real editor

**Fix Details:**
- Moved most of the editor-entry-point.tsx into a new module called `editor-entry-point-imports.tsx`
- Import `editor-entry-point-imports.tsx` from karma config
- Created a new helper function `boundingClientRectToCanvasRectangle` to be able to quickly get an element's CanvasRectangle. The function is mostly copy pasted from dom-walker. I put it in test-utils because it relies on a hybrid of editor state and DOM, so no production code should treat it as a universally usable function.
- Fixed `InsertionPlusButton` tests which were simply expecting some global coordinates where the plus button should show up
- Fixed the FlexReparentIndicator tests which were similarly just measuring global coordinates
- Fixed absolute reparent strategy tests, some of which had a check for mid-drag global coordinates of the element
- Fixed `Paste Here` which accidentally used the mouse position OF WHEN YOU CLICKED IN THE CONTEXT MENU TO SELECT PASTE HERE! instead now it uses the mouse position of the right click that opened the context menu
- Fixed a draw-to-insert test which was drawing using window mouse coordinates – no more
- I also changed the watch-tsc to include --noEmit, maybe it makes the typescript compiler hang less often?


ruggi (approved):

editor/karma-setup.js
```diff
@@ -18,6 +18,7 @@ ReactTestingLibrary.configure({
   },
 })
 
+// this must run before importing the editor-entry-point-imports module
```
Rheeseyb, line 21: This comment should be duplicated in `karma-base.conf.js`

Rheeseyb (approved):

editor/karma-setup.js
balazsbajorics: oh!

balazsbajorics: pushed

PR #4169, opened 2023-09-11
Refactor the floating panels logic to make it easier to test.

seanparsons:
**Problem:**
The logic behind the new floating panels is currently embedded within a handful of React callbacks, which means that the steps of it cannot be unit tested.

**Fix:**
As directly as possible the logic was pulled out of the various callbacks into `floating-panels-state.tsx`.

However a refactor was also done to the structure of the data, because it was causing some bugs. Previously a combination of `location` and `locationInColumn` was used to determine where a particular panel would be, but that permitted invalid cases like a single panel in a column at position 1 or two things in the same place. Now the columns are stored in an object where the key is the location name and the value is an array of the panels within the column.

**Commit Details:**
- Moved various types, constants and utility functions to `floating-panels-state.tsx`.
- Created container type `PanelState` that holds the panel content within a field `panelContent`.
- Extracted out `updatePanelsToDefaultSizes`, `updatePanelPositionsBasedOnLocationAndSize`, `updateSizeOfPanel` and `dragPaneToNewPosition` from the callbacks.
- Implemented `getOrderedPanelsForRendering`, which returns the panels in a consistent order regardless of where they are positioned.
- Added some additional optics and utilities.


editor/src/components/canvas/floating-panels-state.tsx
```diff
@@ -0,0 +1,639 @@
...
+        y: 0,
+        width: 80,
+        height: canvasSize.height,
+      }),
+    },
+    // between the right edge and the last right menu
+    {
+      targetPanel: 'rightMenu2',
+      frame: windowRectangle({
+        x: rightColumn2Frame.x + rightColumn2Frame.width - EdgeDropAreaWidth, //GapBetweenPanels,
```
bkrmendy, line 521: is `GapBetweenPanels` still needed?

seanparsons: Comment from pre-existing code, leaving it for now.

balazsbajorics (approved):

editor/src/core/shared/optics/optic-creators.ts
```diff
@@ -79,6 +79,28 @@ export function fromArrayIndex<A>(index: number): Optic<Array<A>, A> {
...
+export function pairTupleSecond<A1, A2>(): Optic<[A1, A2], A2> {
+  return lens(
+    ([_a1, a2]) => {
+      return a2
+    },
+    ([a1, _a2], newA2) => {
+      return [a1, newA2]
+    },
+  )
+}
```
Rheeseyb, line 102: What are these for?

Rheeseyb (approved):

editor/src/core/shared/optics/optic-creators.ts
seanparsons: They were implemented at one point to get at the respective parts of the result that comes back from `Object.entries`, but they ended up not being used in the end.

PR #4175, opened 2023-09-12
pass `remixRoutingTable` to `normalisePathToUnderlyingTarget`

bkrmendy:
## Description
The Remix Routing Table is a data structure mapping the uid of the root component of each Remix route module to the path of the route module.  `normalisePathToUnderlyingTarget` uses this lookup table to check whether an element path points to a component that is the root element of a route module.

The rest of the PR is about passing the remix routing table from `DerivedState` to `normalisePathToUnderlyingTarget`.

## Out of scope
A fairly common pattern is that an action calls other actions/creates intermediate `EditorState`s inside its handler function. Along the new editor state, a new derived state should be created as well, so that the derived state reflects the changes in the new editor state. Since the primary scope of this PR is create the necessary data flows so that `normalisePathToUnderlyingTarget` can use the remix routing table, the correct derived state generation is postponed to a follow-up PR.


seanparsons (approved):
I'm a little wary of functions like `getElementFromProjectContents` ending up with something Remix specific being passed into them. But I don't particularly have an alternative in mind instead of that...

Rheeseyb (approved):

Rheeseyb:
For the visibility of others here, we've agreed to merge this PR as is, since this unblocks a lot of other work, but a follow up PR will be undoing a lot of these changes, since we realised that `normalisePathToUnderlyingTarget` can be significantly simplified. We could make that change now, but we need to prove that the simplification works (and is performant) before merging it, and we don't want to block the rest of the Remix work.

PR #4176, opened 2023-09-12
Remix valid path generation

gbalint:
**Problem:**
Remix content needs to taken into account when generating the valid paths. 

**Fix:**
Generating the valid paths by finding the outlet and using the relative paths of currently rendered route modules and their contents.

I added tests on:
- having metadata
- selection
- when there are more scenes, metadata is there for both remix scenes


editor/src/components/canvas/remix/remix-rendering.spec.browser2.tsx
```diff
@@ -233,6 +459,104 @@ describe('Remix navigation', () => {
 
     await expectRemixSceneToBeRendered(renderResult, 'About')
   })
+
+  // TODO: enable when the dom walker updates on the mutation observer
+  xit('Remix navigation updates metadata', async () => {
```
bkrmendy, line 464: this can be un-`x`-d

bkrmendy (approved):

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1934,13 +1935,22 @@ export function getParseSuccessForFilePath(
   }
 }
 
+export type RemixValidPathsGenerationContext =
+  | { type: 'disabled' }
+  | {
+      type: 'active'
```
Rheeseyb, line 1941: Surely this should be either `'disabled'` and `'enabled'`, or `'inactive'` and `'active'` 😁

editor/src/components/canvas/remix/utopia-remix-root-component.tsx
```diff
@@ -81,17 +100,66 @@ export const UtopiaRemixRootComponent = React.memo((props: UtopiaRemixRootCompon
...
       return null
     }
-
-    return createMemoryRouter(routes)
+    const initialEntries = currentEntriesRef.current == null ? undefined : currentEntriesRef.current
+    return createMemoryRouter(routes, { initialEntries: initialEntries })
   }, [routes])
 
+  const updateNavigationData = React.useCallback(
+    (routerr: RouterType, location: Location) => {
```
Rheeseyb, line 122: Can we rename `routerr` to something more sensible please (even `innerRouter` would do)

Rheeseyb (approved):

PR #4177, opened 2023-09-12
update dependecy crusier rules

bkrmendy:
## Description
The `not-from-ts-to-tsx` rule proved to be too restrictive when adding library code to `.tsx` files. With the rule present, refactoring files to work around the rule proved to be very time-consuming, so this PR scopes the rule to `*.worker.ts` files, so that Vite doesn't attempt to hot-reload them.

### DIFF
--- a/editor/.dependency-cruiser.js
+++ b/editor/.dependency-cruiser.js
@@ -186,7 +186,7 @@ module.exports = {
     },
     {
       name: 'not-from-ts-to-tsx',
-      comment: 'Prevent a .ts file from pointing to a .tsx file.',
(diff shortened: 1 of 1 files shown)


ruggi (approved):

editor/.dependency-cruiser.js
```diff
@@ -189,10 +189,11 @@ module.exports = {
       comment: 'Prevent a .ts file from pointing to a .tsx file.',
```
seanparsons, line 189: Looks like this comment needs updating from the change you've made.

seanparsons (approved):

Rheeseyb (approved):

gbalint (approved):

PR #4183, opened 2023-09-13
Fix group behavior when resizing its parent

### LINKED ISSUE #4182: Fix group behavior when resizing parent

ruggi:
Fixes #4182 

**Problem:**

Groups behavior is funky when resizing the parent, not exclusively when its position is pinned to the parent bounds.

**Fix:**

This PR removes the previous (wrong!) workaround and instead:
1) removes the forced w/h for the `Group` component, which is only set when the element misses both the `w/h` property _and_ is not pinned to a direction (`top+bottom` or `left+right`)
2) enqueues the elements for true up in the resize strategy to make sure they are trued up once the interaction completes (at the cost of one extra render)


editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
```diff
@@ -1569,32 +1614,127 @@ export var storyboard = (
...
+          formatTestProjectCode(`
+          import * as React from 'react'
+          import { Storyboard, Group } from 'utopia-api'
+
+          export var storyboard = (
+            <Storyboard data-uid='storyboard'>
+              <div data-uid='div' style={{ background: '#aaa', position: 'absolute', left: 100, top: 100, width: 200, height: 100 }}>
+                <Group data-uid='group' style={{ position: 'absolute', top: 50, bottom: 33, width: 89, left: 60, background: 'white' }}>
+                  <div data-uid='foo' style={{ position: 'absolute', width: 30, height: 11, background: 'red', top: 0, left: 0 }} />
+                  <div data-uid='bar' style={{ position: 'absolute', width: 30, height: 11, background: 'blue', top: 6, left: 59 }} />
```
balazsbajorics, line 1705: I think the `width` of foo and bar should not change, only the height! can you change the code so that the starting group+children is already trued up?

balazsbajorics:
thank you for the quick turnaround and fixing this!

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
ruggi: the idea was to have a non-trued-up group to make sure the true up works both ways, but it's fine by me to change it back :)

balazsbajorics: oh okay, that's a fine idea then! but can you please add a comment about it? :)

seanparsons (approved):

balazsbajorics (approved):

PR #4185, opened 2023-09-13
Simplify normalisePathToUnderlyingTarget

Rheeseyb:
**Problem:**
We use `normalisePathToUnderlyingTarget` to determine which file a given element resides in (from the element path), but that function was overly complex and required `nodeModules` and now the remix routing table, as the lookup was performed by traversing the entire element path, meaning we'd be hoping across files to resolve the next part of the path.

**Fix:**
Since we know that UIDs are unique across the parsed parts of the project, and we are only interested in those parsed parts, we can create a mapping from UID to file by traversing each of the parsed models in the project, and then use that mapping to skip straight to the final file. This wouldn't necessarily seem like an overall improvement in complexity, but we were already having to traverse all parsed models to extract the UIDs as part of the process we use to ensure their uniqueness, so tweaking that logic to create the mapping was trivial. This also meant that the number of types in the `NormalisePathResult` union could be drastically reduced too.

This did expose some tests that passed via pure happenstance - since the previous logic would stop as soon as it reached the last `ElementPathPart`, it would not ensure that the target element _actually_ exists in the file that it returned, meaning e.g. a `deleteElements` call with duplicate paths would still manage to resolve the file paths even if the element had already been deleted. To fix this, I introduced a `NormalisePathElementNotFound`, which is then handled in places where it was fine for the logic to continue.

**Note:**
This simplification makes most of the params redundant. To keep the logic changes clear, I'm going to create a separate PR that cleans up all of the mess, which I'll branch off of this one (and merge them both together). I'll also address the `FIXME`s in that PR.


bkrmendy (approved):

gbalint (approved):

editor/src/components/custom-code/code-file.ts
```diff
@@ -429,210 +362,43 @@ export function normalisePathSuccessOrThrowError(
...
+  const staticPath = EP.dynamicPathToStaticPath(elementPath)
+  const lastPartOfPath = EP.takeLastPartOfPath(elementPath)
 
-  const droppedPathPart = potentiallyDroppedFirstPathElementResult.droppedPathElements
-  if (droppedPathPart.length === 0) {
-    return normalisePathError(
-      `Unable to handle empty scene path part for ${optionalMap(EP.toString, elementPath)}`,
-    )
-  }
+  const allUidsWithFiles = getAllUniqueUids(projectContents)
```
seanparsons, line 377: Do we know how much of a performance impact this has on a large project?

editor/src/core/shared/element-path.ts
```diff
@@ -1051,6 +1051,21 @@ export function dropFirstPathElement(path: ElementPath | null): DropFirstPathEle
   }
 }
 
+export function takeLastPartOfPath(path: ElementPath): ElementPath
+export function takeLastPartOfPath(path: null): null
+export function takeLastPartOfPath(path: ElementPath | null): ElementPath | null {
+  if (path == null) {
```
seanparsons, line 1057: Does this really need to be nullable? You'd be surprised how much of a performance loss this causes.

seanparsons (approved):

editor/src/core/shared/element-path.ts
Rheeseyb: Will remove the `| null` because it isn't needed anymore

PR #4200, opened 2023-09-15
Feature/grid panes

balazsbajorics:
This PR reworks the floating panels code to use a grid-based layout.

**Details:**
- The layout is stored as an array of columns, with each column containing zero or more panel.
- we use react-dnd to enable reordering of the panels
- the drop-targets are actual divs drawn at various offsets from the grid column
- For now, resize is not supported: the code pane has a width of 500, and any column containing at least one of `Inspector | Navigator` has a width of 260
- I deleted a bunch of code that was talking about resizing in the floating panels environment
- There can be maximum 2 columns on the left and 2 columns on the right
- The grid reorder code makes sure you can only have a panel in "column 2" if the leftmost column is occupied. same for the right hand side.
- Reordering the panels should let them push each other to the other side of the editor
- You can also reorder within a column
- FOR NOW, panels take up equal space within the same column. (so it's a full-column height, or 50-50 split for two panels or a 33-33-33 split for three panels.)
- Please try it and give feedback!


editor/src/components/canvas/canvas-wrapper-component.tsx
```diff
@@ -217,53 +196,6 @@ export const CanvasWrapperComponent = React.memo(() => {
   )
 })
 
-const ErrorOverlayComponent = React.memo(() => {
```
balazsbajorics, line 220: moved to canvas-error-overlay.tsx

editor/src/components/canvas/camvas-floating-toolbars.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
```
seanparsons, line 1: This file is called `camvas-floating-toolbars.tsx` BTW...

editor/src/components/canvas/grid-panel.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const dropBelowElement: LayoutUpdate = React.useMemo(
+    () => ({
+      type: 'after-index',
+      indexInColumn: order,
+      columnIndex: index,
+    }),
+    [order, index],
+  )
+
+  const canDropAbove = draggedPanel != null && canDrop(draggedPanel, dropAboveElement)
```
seanparsons, line 41: Do these not need to be `useMemo` wrapped?

balazsbajorics: not because `canDropAbove` is a boolean so there is no object reference equality to save here

editor/package.json
```diff
@@ -118,6 +118,7 @@
     "@svgr/plugin-jsx": "5.5.0",
     "@tippyjs/react": "4.1.0",
     "@types/fontfaceobserver": "0.0.6",
+    "@types/lodash.findlastindex": "4.6.7",
```
seanparsons, line 121: Chrome seems to indicate to me that `Array.findLastIndex` is a real thing.

editor/src/components/canvas/grid-panels-state.tsx
```diff
@@ -0,0 +1,223 @@
...
+    return layout.map((column) => {
+      return removeAll(column, [paneToMove], (l, r) => l.uid === r.uid)
+    })
+  }
+
+  function floatColumnsTowardsEdges(layout: StoredLayout) {
+    const leftSide = layout.slice(0, IndexOfCanvas)
+    const rightSideReversed = layout.slice(IndexOfCanvas).reverse()
+    const leftSideFixed = accumulate(new Array(leftSide.length).fill([]), (acc) => {
+      let iterator = 0
```
seanparsons, line 200: It's definitely not an iterator!
```suggestion
      let index = 0
```

seanparsons (approved):

editor/src/components/canvas/grid-panels-container.tsx
```diff
@@ -0,0 +1,138 @@
...
+        // if the drop results in no change, we don't allow it
+        return false
+      }
+
+      return true
+    },
+    [panelState],
+  )
+
+  const columnWidths: Array<`${number}px`> = React.useMemo(
```
seanparsons, line 71: This doesn't compile...

Rheeseyb: This is actually valid TS!

editor/src/components/canvas/grid-panel.tsx
```diff
@@ -0,0 +1,138 @@
...
+    ),
+  )
+
+  const draggablePanelComponent = (() => {
+    switch (panel.name) {
+      case 'code-editor':
+        return (
+          <CodeEditorPane
+            panelData={props.pane.panel}
+            small={props.pane.span !== GridPanelsNumberOfRows} // it's only not small if it's full-height (span === number of rows)
```
Rheeseyb, line 69: So perhaps a better prop name would be `fullHeight`

editor/src/components/canvas/grid-panels-dnd.tsx
```diff
@@ -0,0 +1,74 @@
...
+export function useGridPanelDraggable(draggedPanel: StoredPanel): {
+  drag: ConnectDragSource
+  dragPreview: ConnectDragPreview
+} {
+  const [{ isDragging }, drag, dragPreview] = useDrag(
+    () => ({
+      // "type" is required. It is used by the "accept" specification of drop targets.
+      type: FloatingPanelTitleBarType,
+      item: { draggedPanel: draggedPanel } as FloatingPanelDragItem,
+      // The collect function utilizes a "monitor" instance (see the Overview for what this is)
```
Rheeseyb, line 19: See what Overview?

editor/src/components/canvas/grid-panels-drag-targets.tsx
```diff
@@ -0,0 +1,110 @@
...
+      }),
+      [columnIndex],
+    )
+
+    const canDropBefore = draggedPanel != null && canDrop(draggedPanel, dropBeforeColumn)
+    const canDropAfter = draggedPanel != null && canDrop(draggedPanel, dropAfterColumn)
+
+    const { drop: dropBefore, isOver: isOverBefore } = useGridPanelDropArea(
+      columnIndex,
+      9,
```
Rheeseyb, line 37: I don't like these arbitrary numbers

editor/src/components/canvas/grid-panels-state.tsx
```diff
@@ -0,0 +1,225 @@
...
+  span: number
+  index: number
+  order: number
+}
+
+export type PanelName = Menu | Pane
+
+export interface StoredPanel {
+  name: PanelName
+  type: 'menu' | 'pane'
```
Rheeseyb, line 43: Since this only appears to be used for setting the width, should we just store the width here instead?

editor/src/components/canvas/grid-panels-state.tsx
```diff
@@ -0,0 +1,225 @@
...
+/**
+ * Normalizes the index to 0,1,2,3
+ */
+function normalizeColIndex(index: number): number {
+  return mod(index, NumberOfColumns)
+}
+
+export function updateLayout(
+  stored: StoredLayout,
+  paneToMove: StoredPanel, // must be referentially equal to the stored panel!
```
Rheeseyb, line 131: 👀 that reliance sounds dangerous to me. Why the need for ref equality here? It looks to me like we could remove the old one purely based on the `name`, or even the `uid`

balazsbajorics: oh that comment is outdated, will delete

editor/src/components/canvas/grid-panels-state.tsx
balazsbajorics: nope! the width is a property of the column! this informs the default width (which is all we have since I haven't done the resizing yet) but it's not _the_ width which will be user changeable per column, not per panel item

editor/src/components/canvas/grid-panels-state.tsx
balazsbajorics: it's also not an index of the loop - it's the index of the "cursor" of the array the loop is optionally updating

editor/src/components/canvas/grid-panel.tsx
balazsbajorics: the prop is talking in terms of VSCode – small means it will get a 0.7x scaling. in the grid context, we decide that small happens when it's not full height in the grid

editor/src/components/canvas/design-panel-root.tsx
```diff
@@ -169,18 +170,7 @@ const DesignPanelRootInner = React.memo(() => {
...
-                  height={0}
-                  onResize={NO_OP}
-                  setIsResizing={NO_OP}
-                  resizableConfig={{
-                    enable: {
-                      right: true,
-                    },
-                  }}
-                />
+                <CodeEditorPane panelData={null as any} small={false} />
```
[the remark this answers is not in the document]
balazsbajorics: Either the FS is on, and this code doesn't run. or the FS is off, and the code reading panelData doesn't run.

this old code will be removed as soon as the feature switch is removed. making the prop optional means I'd need to put error handling somewhere where we don't anticipate an error to happen.

editor/src/components/canvas/grid-panels-container.tsx
balazsbajorics: oh, old bundled typescript vs new typescript running in vscode tripped me.

editor/src/components/canvas/grid-panels-state.tsx
balazsbajorics: I renamed it to "indexInAccumulator"

editor/src/components/canvas/camvas-floating-toolbars.tsx
balazsbajorics: oh boy

seanparsons (approved):

gbalint (approved):

PR #4201, opened 2023-09-15
Integrate popup menu into canvas toolbar.

seanparsons:
**Problem:**
The insert/wrap/convert pop-up menu would be cleaner if it was a part of the canvas toolbar.

**Fix:**
In case we wanted to still use the floating menu later on this PR attempts to maintain that rather than just outright destroy it to make the integration work. Which means that there's some work which can be done later on should we decide against keeping the floating menu.

Mostly the reusable innards from `floating-insert-menu.tsx` have been extracted out so that they can be utilised in the canvas toolbar.

Along with reusing the general logic, this PR also attempts to re-use as much of the styling as possible from the floating menu for the `react-select` based combo box.

**Footage:**
https://github.com/concrete-utopia/utopia/assets/217400/e0a8b447-42f3-40d2-a1b2-94a703ec2ad0

**Commit Details:**
- Exported a few types from `floating-insert-menu.tsx`.
- Extracted out `getActionsToApplyChange` from the `FloatingMenu` component, along with a couple of support functions.
- Created a `WrapInElementWith` type to put a name to an embedded type.
- Added `ToolbarSearchListing` component as a portal for the menu.
- Added `CanvasToolbarSearch` component which wraps the `react-select` component for the listing.
- Added additional event handlers for the various new actions needed.
- Created `wrapInSubmenu` function for consistent submenu design.
- Added submenus for the convert/wrap-in/insert cases.


editor/src/components/editor/canvas-toolbar.tsx
```diff
@@ -44,12 +49,140 @@ import { useToolbarMode } from './canvas-toolbar-states'
...
 
 export const InsertMenuButtonTestId = 'insert-menu-button'
 export const InsertConditionalButtonTestId = 'insert-mode-conditional'
 export const CanvasToolbarId = 'canvas-toolbar'
 
+export interface ToolbarSearchListingProps {}
+
+export const CanvasToolbarSearchPortalId = 'canvas-toolbar-search-portal'
+
+export const ToolbarSearchListing = React.memo((props: ToolbarSearchListingProps) => {
```
Rheeseyb, line 83: You can just do
```React.memo(() => {```
So there's no need for the empty `interface`

File: editor/src/components/editor/canvas-toolbar.tsx [hunk 2]
```diff
@@ -44,12 +49,140 @@ import { useToolbarMode } from './canvas-toolbar-states'
...
+
+  return (
+    <WindowedSelect
+      id={'canvas-toolbar-search'}
+      ref={selectRef}
+      components={{ Option: CustomComponentOption }}
+      openMenuOnFocus={true}
+      openMenuOnClick={true}
+      onBlur={undefined}
+      //value={value}
```
Rheeseyb, line 129:
```suggestion
```

editor/src/components/editor/canvas-toolbar.tsx
seanparsons: Ah yes, at one point it wasn't empty, fixed in bee6ce5d3.

Rheeseyb (approved):

balazsbajorics (approved):

PR #4202, opened 2023-09-18
Don't resize text group children past intrinsic bounds

### LINKED ISSUE #4214: Don't resize text items inside groups past their intrinsic size

ruggi:
Fixes #4214 

**Problem:**

Text elements inside groups can be resized past their intrinsic size.

Kapture 2023-09-20 at 10 45 35

**Fix:**

1. Store the text measurements in `specialSizeMeasurements.textBounds` during DOM walker runs
2. When pushing bounds, force the dimensions to the max between the resized bounds and the text bounds stored in `specialSizeMeasurements`

Kapture 2023-09-20 at 10 46 22


editor/src/components/canvas/dom-walker.ts
```diff
@@ -1045,9 +1038,58 @@ function getSpecialMeasurements(
...
+    }
+  }
+  return true
+}
+
+function getTextBounds(element: HTMLElement, scale: number): DOMRect | null {
+  const range = document.createRange()
+  range.selectNodeContents(element)
+  if (range.getBoundingClientRect == null) {
+    // for some reason, this can be undefined (in tests?)
```
seanparsons, line 1064: In jest tests only?

ruggi: apparently so

seanparsons (approved):
Seems a little odd to me personally that the text extends outside the group bounds, but everything here looks good to me!

editor/src/components/canvas/dom-walker.ts
seanparsons: That's likely because they run with a fake DOM, which doesn't support everything a regular browser supports.

editor/src/components/canvas/dom-walker.ts
```diff
@@ -1045,9 +1038,58 @@ function getSpecialMeasurements(
...
+      node.nodeType === Node.TEXT_NODE ||
+      (node.nodeType === Node.ELEMENT_NODE && node.nodeName === 'BR')
+    if (!isForText) {
+      return false
+    }
+  }
+  return true
+}
+
+function getTextBounds(element: HTMLElement, scale: number): DOMRect | null {
```
gbalint, line 1060: This feels redundant with `getCanvasRectangleFromElement` with the `withContent = 'with-content'` parameter value. 
And that is used to fill the `specialSizeMeasurements.globalFrameWithTextContent`, which is also something similar (the difference is that `globalFrameWithTextContent` is a union of the text bounds and the global frame).

Can we somehow unify how these are calculated? E.g. use `getTextBounds` in `getCanvasRectangleFromElement` too, so we don't have two ways to calculate text size.

ruggi: would something [`5e546a4` (#4202)](https://github.com/concrete-utopia/utopia/pull/4202/commits/5e546a47f8714eb570cf12f6c2d9d1816afe9b91) work?

Rheeseyb (approved):

PR #4204, opened 2023-09-18
Navigate home if the current route no longer matches

bkrmendy:
https://screenshot.click/18-01-a0htj-ncex2.mp4

## Problem
When a route module is deleted, but that route module is part if the rendered route, an error is thrown in the console, and no content is rendered.

## Fix
Detect when the current location cannot be resolved from the current route modules, and if it can't be, navigate back to `/`


Rheeseyb (changes requested):
The code itself is fine, but this also means that we'd never show a 404 error in legit use cases, so e.g. a `Link` with a slight typo in the `to` prop would just seemingly do nothing, rather than showing the 404 error when clicked. We need to either ensure this only happens as a result of the current route being deleted, make the navigation more deliberate (e.g. requiring the user to click a button to return home), or show a toast warning that the route didn't exist

PR #4205, opened 2023-09-18
Move dependencies from createRemixDerivedData into useGetRoutes hook

Rheeseyb:
**Problem:**
The use of `createRemixDerivedData` was causing us to repeatedly create new `router`s when nothing relevant had changed, which would cause issues with navigation in a running remix app.

**Fix:**
The problem here was really that we were trying to derive too much data here, creating the routes themselves as part of the derived state. Rather than doing that, we now effectively create placeholders for the routes, and then craft the `loader` and `action` functions (via `createExecutionScope`) as and when we need them, following the pattern that was set out with the `routeModules`. To achieve this I had to pull out a load of the dependencies of `getRemixExportsOfModule` and shift them into `createExecutionScope`, which then requires providing them when resolving the route or module. Then the final part of the puzzle was to provide those values via refs inside the `useGetRoutes` and `useGetRouteModules` hooks so that we could ensure we'd always have the up to date data without causing excessive renders.

I've also updated an existing test to navigate deeper into an app, which would otherwise fail.

Note that the Remix tests now dispatch an extra `runDOMWalker()` at the start of each test. The previously excessive rendering meant that when running those tests we would have the correct metadata, which is no longer the case, and so that extra dispatch is required. This appears to be a difference between running the tests in the test runner vs in the real app.

### DIFF
--- a/editor/src/components/canvas/remix/utopia-remix-root-component.tsx
+++ b/editor/src/components/canvas/remix/utopia-remix-root-component.tsx
@@ -1,7 +1,7 @@
 import { RemixContext } from '@remix-run/react/dist/components'
 import type { RouteModules } from '@remix-run/react/dist/routeModules'
 import React from 'react'
-import type { Location } from 'react-router'
+import type { DataRouteObject, Location } from 'react-router'
 import { createMemoryRouter, RouterProvider } from 'react-router'
 import { UTOPIA_PATH_KEY } from '../../../core/model/utopia-constants'
 import type { ElementPath } from '../../../core/shared/project-file-types'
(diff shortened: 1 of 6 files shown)


editor/src/components/canvas/remix/utopia-remix-root-component.tsx
```diff
@@ -75,7 +93,86 @@ function useGetRouteModules(basePath: ElementPath) {
...
+    fileBlobsRef,
+    hiddenInstancesRef,
+    displayNoneInstancesRef,
+  ])
+}
+
+function useGetRoutes() {
+  const routes = useEditorState(
+    Substores.derived,
+    (store) => store.derived.remixData?.routes ?? [],
```
bkrmendy, line 110: `MemoryRouter` has an invariant that it can only be passed a non-empty array of routes, so we should check the length of `routes` before the router is created later on in this file

bkrmendy (approved):

gbalint (approved):

PR #4213, opened 2023-09-20
signal routes not matching location

bkrmendy:
image

## Problem
We want to provide some extra support for the eventuality when a location doesn't match because there's no corresponding route module (and not because the loader function returns a 404). This can easily happen when a route module is deleted, but there's still a link pointing to that route in the project, and it gets clicked, thus navigating to the non-existent route.

## Fix
Add an indicator in the Remix Scene label that explains the situation.

### DIFF
--- a/editor/src/components/canvas/remix/utopia-remix-root-component.tsx
+++ b/editor/src/components/canvas/remix/utopia-remix-root-component.tsx
@@ -27,7 +27,7 @@ interface RemixNavigationContext {
 }
 
 interface RemixNavigationAtomData {
-  [pathString: string]: RemixNavigationContext
+  [pathString: string]: RemixNavigationContext | undefined
 }
 
 export const ActiveRemixSceneAtom = atom<ElementPath>(EP.emptyElementPath)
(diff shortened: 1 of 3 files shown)


gbalint (approved):

editor/src/components/canvas/remix/utopia-remix-root-component.tsx
```diff
@@ -27,7 +27,7 @@ interface RemixNavigationContext {
 }
 
 interface RemixNavigationAtomData {
-  [pathString: string]: RemixNavigationContext
+  [pathString: string]: RemixNavigationContext | undefined
```
seanparsons, line 30: Is this so that you can assign `undefined` values against a key? If that means something different to a key not existing I'd recommend putting some other marker value instead as `undefined` is the result you'd get for a non-existent index against an object.

seanparsons (approved):

editor/src/components/canvas/remix/utopia-remix-root-component.tsx
bkrmendy: I added this so that null checks are enforced when looking up values from `RemixNavigationAtom`

PR #4215, opened 2023-09-20
Remix Outlet navigator item label

gbalint:
**Description:**
Remix Outlet is a component with special behavior: during render it is replaced by the default exported component of the actual Remix route file.
But that route component is invisible in the navigator and does not appear in our metadata hierarchy.
As a minimal way to show which route component is rendered, I added the component name to the Outlet navigator item's label, e.g.:

image

### DIFF
--- a/editor/src/components/canvas/remix/remix-utils.tsx
+++ b/editor/src/components/canvas/remix/remix-utils.tsx
@@ -38,6 +38,7 @@ import { MetadataUtils } from '../../../core/model/element-metadata-utils'
 import type { ElementInstanceMetadataMap } from '../../../core/shared/element-template'
 import type { ElementPathTrees } from '../../../core/shared/element-path-tree'
 import { getAllUniqueUids } from '../../../core/model/get-unique-ids'
+import { safeIndex } from '../../../core/shared/array-utils'
 
 export const OutletPathContext = React.createContext<ElementPath | null>(null)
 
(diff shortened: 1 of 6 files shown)


editor/src/components/inspector/sections/layout-section/conditional-section.tsx
```diff
@@ -294,7 +294,7 @@ export const ConditionalSection = React.memo(({ paths }: { paths: ElementPath[]
   )
 
   const branchLabels = useEditorState(
-    Substores.metadata,
```
bkrmendy, line 297: Is this change still needed?

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -383,3 +387,32 @@ export function getRoutesAndModulesFromManifest(
...
+    return null
+  }
+
+  const outletChildren = MetadataUtils.getImmediateChildrenPathsOrdered(metadata, pathTrees, path)
+  if (outletChildren.length == 0) {
+    return null
+  }
+
+  const uidsToFilePath = getAllUniqueUids(projectContents).uidsToFilePaths
+  const filePath = uidsToFilePath[EP.toUid(outletChildren[0])]
```
seanparsons, line 407: Minor: You could use `safeIndex` the array lookup and replace the length check above with a null check on the result.

bkrmendy (approved):

editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
```diff
@@ -97,21 +98,31 @@ const elementSupportsChildrenSelector = createCachedSelector(
 )((_, navigatorEntry) => navigatorEntryToKey(navigatorEntry))
 
 export const labelSelector = createCachedSelector(
-  (store: MetadataSubstate) => store.editor.jsxMetadata,
+  (store: MetadataSubstate & ProjectContentSubstate) => store.editor.jsxMetadata,
```
seanparsons, line 101: Looks like these can all be changed to the type alias created elsewhere in this PR.
```suggestion
  (store: ProjectContentAndMetadataSubstate) => store.editor.jsxMetadata,
```

seanparsons (approved):

Rheeseyb (approved):

editor/src/components/inspector/sections/layout-section/conditional-section.tsx
gbalint: Yes, it is needed, because it calls the labelSelector which relies on projectContents too.

PR #4217, opened 2023-09-20
Remove some unused files.

seanparsons:
**Problem:**
We've got some empty and or unused files in the code.

**Fix:**
Removed those files.


Rheeseyb (approved):
We really need to write that Swift test...

balazsbajorics (approved):

PR #4221, opened 2023-09-21
Remix error handling

Rheeseyb:
**Problem:**
Currently the Utopia editor will only provide a Remix error boundary at the root level, ignoring user-defined error boundaries, meaning all errors are captured and displayed wrapped inside the `RemixScene`

**Fix:**
There are two parts to this. First is to support user-defined error boundaries. For this I've added a function `fileExportsFunctionWithName`, which quite simply checks the `ParsedModel` to see if the `exportDetails` include an exported function matching the given name. This is then used when deriving the routes to determine if each route `hasErrorBoundary`. To then render the error boundary, I've reused the `executionScopeCreator` in the same way that it is used to create and render the default component (but we only do this if `hasErrorBoundary` was true for that route).

The second part to this fix is to then provide a fallback `ErrorThrower` component, which is used if no user-defined error boundary exists, will re-throw the error, so that we can then show our regular canvas overlay.

I've included tests to check both behaviours, but you'll notice that they are annoyingly in separate files. I first attempted to write the tests as karma tests, but we couldn't find a way to prevent the tests failing on uncaught errors (which will apparently still happen with error boundaries), so I switched to Jest tests, but if both are run in the same file (even in separate `describe` blocks) the second will fail with the error `TypeError: document.elementsFromPoint is not a function`.


editor/src/components/canvas/remix/remix-error-handling-with-boundary.spec.tsx
```diff
@@ -0,0 +1,17 @@
+import { setFeatureForUnitTestsUseInDescribeBlockOnly } from '../../../utils/utils.test-utils'
+import { runTestReturningErrorBoundaries } from './remix-error-handling.test-utils'
+
+// This test and its sibling have to be kept in separate files,
```
seanparsons, line 4: Weird, but thanks for the comment!

editor/src/components/canvas/remix/remix-error-handling.test-utils.tsx
```diff
@@ -0,0 +1,128 @@
...
+    RegisteredCanvasStrategies,
+    notLoggedIn,
+    false,
+  )
+  await renderResult.dispatch([runDOMWalker()], true)
+  return renderResult
+}
+
+async function clickLinkWithTestId(editor: EditorRenderResult, testId: string) {
+  const targetElement = editor.renderedDOM.queryAllByTestId(testId)[0]
```
seanparsons, line 29: This should give a sensible error if it can't find it or there's more than one.
```suggestion
  const targetElement = editor.renderedDOM.getByTestId(testId)
```

editor/src/components/canvas/remix/remix-error-handling.test-utils.tsx
```diff
@@ -0,0 +1,128 @@
...
+  withCustomBoundary: 'with-custom-boundary' | 'without-custom-boundary',
+): Promise<{ customBoundary: HTMLElement | undefined; canvasOverlay: HTMLElement | undefined }> {
+  const project = createProject(withCustomBoundary === 'with-custom-boundary')
+
+  const renderResult = await renderRemixProject(project)
+
+  await renderResult.dispatch([switchEditorMode(EditorModes.liveMode())], true)
+  await clickLinkWithTestId(renderResult, 'remix-link')
+
+  const [customBoundary] = renderResult.renderedDOM.queryAllByText(CustomBoundaryText)
```
seanparsons, line 121: Should these allow more than one result?

editor/src/components/canvas/remix/utopia-remix-root-component.tsx
```diff
@@ -75,19 +75,25 @@ function useGetRouteModules(basePath: ElementPath) {
         continue
       }
 
+      const createExecutionScope = () =>
```
seanparsons, line 78: Does this need to be a function that then gets invoked twice?

seanparsons (approved):

bkrmendy (approved):

editor/src/components/canvas/remix/remix-error-handling.test-utils.tsx
Rheeseyb: I'm not sure why I used `queryAllByTestId` for this one...

editor/src/components/canvas/remix/remix-error-handling.test-utils.tsx
Rheeseyb: Nope! Changing them to `queryByText`

editor/src/components/canvas/remix/utopia-remix-root-component.tsx
Rheeseyb: Unfortunately it does. It's not quite being invoked twice, it's only being invoked when evaluating either the `default` or the `ErrorBoundary`. The reason is so that we can use the current project contents etc at the time of evaluating either of those

PR #4229, opened 2023-09-22
Group resize min sizes

### LINKED ISSUE #4240: Respect constraint min/max sizes when resizing

ruggi:
Fixes #4240 

**Problem:**

Resizing groups doesn't respect the inner constraints for the group dimensions.

**Fix:**

When resizing an element that contains constrained elements, force min/max dimensions for the resized elements that respect them.

https://github.com/concrete-utopia/utopia/assets/1081051/e39d7961-a9d6-40c8-93b0-dbb67f5a4807


balazsbajorics:
if a group child only has `left` constrained, we should not treat the width as constrained as well, because it results in a group that can only grow with resize, but never shrink: https://screenshot.click/27-41-mu58w-t3tbb.mp4

ruggi:
> if a group child only has `left` constrained, we should not treat the width as constrained as well, because it results in a group that can only grow with resize, but never shrink: [screenshot.click/27-41-mu58w-t3tbb.mp4](https://screenshot.click/27-41-mu58w-t3tbb.mp4)
> 
> also on the video: after the resize becomes constrained and stops following the mouse, we still show "ghost" snap guidelines where the mouse is

makes sense! it was an overside from my part, I misunderstood some bound rules. Fixed in [`1e3b043` (#4229)](https://github.com/concrete-utopia/utopia/pull/4229/commits/1e3b043ed2f0fea848e5d7694052c5192d3b1a3f)

as per the guidelines - I'll give it a thought, maybe they are fine as they are but I'm not sure if they're useful or not 🤔

balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.tsx
```diff
@@ -237,6 +254,232 @@ export function absoluteResizeBoundingBoxStrategy(
...
+    // If it's locked, return the original bounds.
+    return {
+      offset: originalOffset,
+      size: originalDimension,
+    }
+  }
+
+  // The maximum value of the given constraints, used as the upper bound for the adjustment.
+  const maxDimension = constrainedSizes.reduce((size, frame) => {
+    return frame[dimension] > size ? frame[dimension] : size
```
seanparsons, line 341:
```suggestion
    return Math.max(frame[dimension], size)
```

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.tsx
```diff
@@ -237,6 +254,232 @@ export function absoluteResizeBoundingBoxStrategy(
...
+  } else if (minBound) {
+    return frameOffset
+  } else if (maxBound && dimensionBound) {
+    return originalDimension - frameOffset
+  } else if (maxBound) {
+    return originalDimension - frameDimension
+  } else if (dimensionBound) {
+    return frameDimension
+  } else {
+    return 0
```
seanparsons, line 479: Shouldn't this return a `null` or some such? The zero looks like it might end up being the root of some weird bugs in the future.

seanparsons (approved):
A bounty of tests!

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.tsx
ruggi: that's a good point, I made it nullable here [`584fe8a` (#4229)](https://github.com/concrete-utopia/utopia/pull/4229/commits/584fe8a3c01dc7b7cdc1f79ba0ec0c53be169e54)

PR #4231, opened 2023-09-22
Allow export default identifier in remix route files

gbalint:
**Problem:**
Remix routes need to have a default export component (which is rendered in place of the Outlet).
Until now we did not accept the syntax to separately export the component in an `export default <identifer>` line.

**Fix:**
At first I thought we do not parse these exports because they were called `EXPORT_IDENTIFIER` which does not really express that these are always default exports.
So I also renamed these to `EXPORT_DEFAULT_IDENTIFIER` to make this more clear

### DIFF
--- a/editor/src/core/model/project-file-utils.ts
+++ b/editor/src/core/model/project-file-utils.ts
@@ -18,6 +18,8 @@ import type {
   TextFileContents,
   HighlightBoundsWithFileForUids,
   RevisionsStateType,
+  ExportDefaultIdentifier,
+  ExportDefaultFunctionOrClass,
 } from '../shared/project-file-types'
 import {
   RevisionsState,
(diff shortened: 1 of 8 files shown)


seanparsons (approved):

editor/src/core/model/project-file-utils.ts
```diff
@@ -18,6 +18,8 @@ import type {
   TextFileContents,
   HighlightBoundsWithFileForUids,
   RevisionsStateType,
+  ExportDefaultIdentifier,
+  ExportDefaultFunctionOrClass,
```
bkrmendy, line 22: unused import

editor/src/core/model/project-file-utils.ts
```diff
@@ -28,6 +30,8 @@ import {
   forEachParseSuccess,
   isParseSuccess,
   isExportDefaultFunctionOrClass,
+  isExportDefaultIdentifier,
```
bkrmendy, line 33: unused import as well

bkrmendy (approved):

PR #4238, opened 2023-09-25
Make `shouldRevalidate` work

bkrmendy:
## Problem
When navigating from a rendered route module to the same route module (for example, from `/posts/0` to `/posts/1` in the example project), the navigation doesn't happen and an error is thrown.

## Root cause
`DataRouteObject.shouldRevalidate` is a way to optimize data revalidation in Remix (see https://remix.run/docs/en/main/route/should-revalidate for details). It's called after actions and on client-side navigations (this is our use case).

It's optionally exported from route modules, and a factory function called `createShouldRevalidate` (https://github.com/concrete-utopia/utopia/blob/37481e71d5ab5e11a2c80ed53e0a3b2c689450ff/editor/src/third-party/remix/client-routes.tsx#L143) wraps it into a higher-order function. The wrapping happens even if `shouldRevalidate` isn't exported from the route module. This `createShouldRevalidate` picks out the route to be validated from `routeModulesCache`, which is where our code needs to be improved since we never actually fill out `routeModulesCache`, and because of this, an invariant exception is thrown. 

## Fix
When creating the route modules in `UtopiaRemixRootComponent`, write the route module into `CreateRemixDerivedDataRefsGLOBAL.routeModulesCache.current`.

## Details
As of this PR, the route module cache is a global variable (somewhat modeled on RemixBrowser, where the route module cache is a global on the window, `window.__routeModules` (https://github.com/remix-run/remix/blob/d5ec2f8e01527998470c98db6d15cb489823187e/packages/remix-react/browser.tsx#L139)). This is problematic for us because `createShouldRevalidate` also creates a variable that it captures (https://github.com/concrete-utopia/utopia/blob/bdad305658792bf4fa3738e4a70cb47b1549413a/editor/src/third-party/remix/client-routes.tsx#L148), and since we can have multiple Remix scenes, this variable can be written by otherwise unrelated `shouldRevalidate` calls. This might lead to revalidation not working correctly. Still, we should accept this as a known bug so that the problem above is fixed.


gbalint (approved):
I think the compromise is acceptable, because shouldRevalidate is stored on the Remix route, which is stored in DerivedState, so it would be a really involved refactor to make that non-global. Basically we would have to move back plenty of code to UtopiaRemixRootComponent hooks.
But I think it you should add a TODO comment about this, so we will take this into account when we migrate to Remix V2, and we rethink what can happen in the derivestate and what needs to happen separately for the different remix scenes.

ruggi (approved):

seanparsons (approved):

PR #4239, opened 2023-09-25
Chore/bump ts 5.2

gbalint:
**Description:**
Updating Typescript to 5.2.2, Eslint to 7.32 (8 is incompatible so I postponed that), @typescript/eslint to 6.7.2
The typescript library used by our parser-printer is _not_ updated in this PR.

To explain the changes I added inline comments to the PR


editor/src/components/canvas/canvas-types.ts
```diff
@@ -41,16 +39,6 @@ export enum CSSCursor {
...
-  VectorDelete = "-webkit-image-set( url( '/editor/cursors/cursor-vector-default.png ') 1x, url( '/editor/cursors/cursor-vector-default@2x.png ') 2x ) 4 4, zoom-out",
-  VectorMove = "-webkit-image-set( url( '/editor/cursors/cursor-vector-default.png ') 1x, url( '/editor/cursors/cursor-vector-default@2x.png ') 2x ) 4 4, zoom-out",
-  VertexDefault = "-webkit-image-set( url( '/editor/cursors/cursor-vertex-default.png ') 1x, url( '/editor/cursors/cursor-vector-default@2x.png ') 2x ) 4 4, zoom-out",
-  VertexInsert = "-webkit-image-set( url( '/editor/cursors/cursor-vertex-default.png ') 1x, url( '/editor/cursors/cursor-vertex-default@2x.png ') 2x ) 4 4, zoom-out",
-  VertexDelete = "-webkit-image-set( url( '/editor/cursors/cursor-vertex-default.png ') 1x, url( '/editor/cursors/cursor-vertex-default@2x.png ') 2x ) 4 4, zoom-out",
-  VertexMove = "-webkit-image-set( url( '/editor/cursors/cursor-vertex-default.png ') 1x, url( '/editor/cursors/cursor-vertex-default@2x.png ') 2x ) 4 4, zoom-out",
-  Text = 'text',
-  TextInsert = 'text',
```
gbalint, line 53: The problem with CSSCursor enum values is that they are repeated under multiple names.
Since these enum names are not used anyway, I just deleted them, but generally we should not use an enum for this at all. A better solution would to just store them in a const object.

editor/src/components/editor/server.ts
```diff
@@ -289,7 +289,7 @@ export async function saveAssets(
   const promises = assets.map((asset) =>
     saveAsset(projectId, asset.fileType, asset.base64, asset.fileName),
   )
-  return await Promise.all(promises)
+  return Promise.all(promises)
```
gbalint, line 292: Breaks the `@typescript-eslint/return-await` rule.

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -1635,7 +1632,7 @@ export function SpecialSizeMeasurementsKeepDeepEquality(): KeepDeepEqualityCall<
       parentFlexDirectionResult &&
       parentJustifyContentEquals &&
       parentFlexGapEquals &&
-      parentPaddingEquals &&
+      parentPaddingEquals.areEqual &&
```
gbalint, line 1635: These are real serious bugs: because of truthiness we did not realize that these are object values, so they are always true! Thankfully, the `@typescript-eslint/strict-boolean-expressions` rule caught them.

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -5090,7 +5090,7 @@ interface NonTrivialKeyword {
 const nontrivial: NonTrivialKeyword = { trivial: false }
 
 type ParsedPropertiesWithNonTrivial = {
-  [Property in keyof ParsedProperties]: ParsedProperties[Property] | NonTrivialKeyword
+  [Prop in keyof ParsedProperties]: ParsedProperties[Prop] | NonTrivialKeyword
```
gbalint, line 5093: The name `Property` is used in the outer scope.

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -5258,7 +5258,7 @@ export function toggleBorderEnabled(_: null, oldValue: CSSBorder): CSSBorder {
     return workingNewValue
   } else {
     const widthValue =
-      oldValue.width != null && oldValue.width.value.value > 0
+      oldValue.width != null && isCSSNumber(oldValue.width.value) && oldValue.width.value.value > 0
```
gbalint, line 5261: Stricter type checking revealed that oldValue.width.value.value can be a string, and than the `> 0` comparison is not correct.

editor/src/core/property-controls/property-controls-parser.ts
```diff
@@ -95,7 +95,7 @@ export function parseNumberInputControlDescription(
       return numberInputControlDescription
     },
     optionalObjectKeyParser(parseString, 'label')(value),
-    objectKeyParser(parseEnum(['number-input']), 'control')(value),
+    objectKeyParser(parseConstant('number-input'), 'control')(value),
```
gbalint, line 98: `parseEnum` 's typing can only guarantee that the result is a string and not a 'number-input'

editor/src/core/shared/object-utils.ts
```diff
@@ -124,7 +124,7 @@ export function objectMap<T extends Record<string, unknown>, K extends keyof T,
   return mappedObj
 }
 
-export function objectMapDropNulls<T, K extends keyof T, U>(
+export function objectMapDropNulls<T extends MapLike<any>, K extends keyof T, U>(
```
gbalint, line 127: A general problem was that we can not just assume that a general T type parameter is an object.

editor/src/core/property-controls/property-controls-parser.spec.ts
```diff
@@ -83,7 +85,7 @@ function runBaseTestSuite<T>(
       control: 'ham sandwich',
     }
     expect(parseFn(value)).toEqual(
-      left(objectFieldParseError('control', descriptionParseError('Not a member of an enum.'))),
+      left(objectFieldParseError('control', descriptionParseError(parseErrorDescription))),
```
gbalint, line 88: Fixed the tests because we switched from `parseEnum` to `parseConstant` in `property-controls-parser`

editor/src/core/shared/math-utils.ts
```diff
@@ -970,14 +970,16 @@ export function rectanglesEqual<C extends CoordinateMarker>(
 ): boolean {
   if (isInfinityRectangle(first) || isInfinityRectangle(second)) {
     return isInfinityRectangle(first) && isInfinityRectangle(second)
-  } else {
+  } else if (isFiniteRectangle(first) && isFiniteRectangle(second)) {
+    // TODO: why is this extra if necessary? They have to be both finite because of the else branch
```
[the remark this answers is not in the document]
seanparsons: I wonder if this is because `||` can short circuit if `isInfinityRectangle(first)` returns `true`, which is not to say that's "correct" I just wonder if they've had to make this more dumb because of some edge case bug.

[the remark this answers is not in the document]
bkrmendy: I think type narrowing + short circuiting is the culprit too. The `if` with the `||` operator can succeed because of either of the `isInfinityRectangle` calls, so the types of `first` or `second` cannot be narrowed at compile time.

I'd restructure the function so that we fully narrow the types of `first` and `second` with an `isFiniteRectangle(first) && isFiniteRectangle(second)` check, and otherwise return `isInfinityRectangle(first) && isInfinityRectangle(second)`.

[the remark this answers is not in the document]
gbalint: Fixed in b49041dc766e24e3b51d6c4f2782417b1e211131

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -1578,12 +1578,9 @@ export function SpecialSizeMeasurementsKeepDeepEquality(): KeepDeepEqualityCall<
...
     const parentFlexDirectionResult = oldSize.parentFlexDirection === newSize.parentFlexDirection
     const parentJustifyContentEquals = oldSize.parentJustifyContent === newSize.parentJustifyContent
-    const parentFlexGapEquals = NumberKeepDeepEquality(oldSize.parentFlexGap, newSize.parentFlexGap)
+    const parentFlexGapEquals = oldSize.parentFlexGap === newSize.parentFlexGap
     const parentPaddingEquals = SidesKeepDeepEquality(oldSize.parentPadding, newSize.parentPadding)
-    const parentHugsOnMainAxisEquals = BooleanKeepDeepEquality(
-      oldSize.parentHugsOnMainAxis,
-      newSize.parentHugsOnMainAxis,
-    )
+    const parentHugsOnMainAxisEquals = oldSize.parentHugsOnMainAxis === newSize.parentHugsOnMainAxis
```
bkrmendy, line 1583: These `===` are a little brittle because if we'd change `parentFlexGap` or `parentHugsOnMainAxis` for any reason, we won't get any type errors, so I'd keep the `*DeepEquality` calls and add the `.areEqual` prop instead

bkrmendy (approved):

ruggi (approved):

Rheeseyb (approved):

PR #4244, opened 2023-09-26
lock and eye icons fixed to the right

lankaukk:
**Problem:**
The lock and hide buttons in the navigator rows weren't visible because they were getting pushed out of the bounds of the navigator when the label was long.

**Fix:**
Now the buttons are fixed to the right edge of the navigator row, and they have the same background color as the row itself, so it covers any continuation of the label beneath it. The icons will stay fixed to the right even if the navigator is wider than the default.

exhibit-A
Screenshot 2023-09-26 at 5 21 15 PM

(*Note: For presentation purposes in the above gif, I made the left pane 268px wide. Today it is 260px and everything gets slightly cut off on the right in _all_ tabs of the left pane )

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item-components.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item-components.tsx
@@ -268,6 +268,7 @@ interface NavigatorItemActionSheetProps {
   instanceOriginalComponentName: string | null
   isSlot: boolean
   iconColor: IcnProps['color']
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):
> Today it is 260px and everything gets slightly cut off on the right in all tabs of the left pane

oh this is my fault, I'm gonna fix it today!

seanparsons (approved):

PR #4251, opened 2023-09-28
No multiselect controls for selection with the same uids

gbalint:
**Problem:**
When there are two remix scenes in the project, it is really easy to accidentally select the same element in both (by using the code editor). In this case we should a multiselect outline around those elements, which is kinda meaningless for theis special case.

**Fix:**
When the selected elements all share the same uid we don't show the multiselect outline and the strategy controls.


editor/src/core/shared/element-path.ts
```diff
@@ -1139,3 +1139,15 @@ export function getStoryboardPathFromPath(path: ElementPath): ElementPath | null
 export function appendTwoPaths(base: ElementPath, other: ElementPath): ElementPath {
   return elementPath([...base.parts, ...other.parts])
 }
+
+export function multipleElementPathsWithTheSameUid(paths: Array<ElementPath>): boolean {
```
seanparsons, line 1143: Would this be a more suitable name?
```suggestion
export function allPathsHaveTheSameUid(paths: Array<ElementPath>): boolean {
```

seanparsons (approved):

editor/src/core/shared/element-path.ts
gbalint: First I used a similar name but I wanted the function to return `false` when there is only one uid (or when the array is empty), and for these cases `allPathsHaveTheSameUid` would be misleading a name.

balazsbajorics:
I did a very poor job following the remix track lately, could you link a test project here where I can play with this?

bkrmendy (approved):

gbalint:
@balazsbajorics https://utopia.pizza/p/118b8b1e-topaz-mousepad/?branch_name=feature-all-uids-the-same-multiselection

balazsbajorics (approved):
I think it's an OK mitigation step, but comes with a big downside: if you actually want to interact with the (multi)selected element on the canvas, you have to unselect it and select it again so you have a single selection.

I think the long term solution should be that we introduce a third type of outline on the canvas, sitting between highlight (temporary) and real selection (permanent, multiseletion usually very deliberate). This "code highlight" should be more permanent than a highlight, and allow you to single click turn it into a "real" selection.

editor/src/core/shared/element-path.ts
Rheeseyb: Hmmm I agree with that reasoning, but this name now gives the impression that it is checking for the presence of any duplicate UIDs, rather than checking that all paths end in the same UID. so e.g. given the input `[ 'aaa/bbb', 'aaa/ccc', 'aaa/ccc:ddd/bbb' ]` I'd expect this function to return `true` because there are two paths ending in `bbb`. So I think for that reason it would be better to name this something `multiplePathsAllWithTheSameUID`

Rheeseyb (approved):

PR #4254, opened 2023-09-28
Hide safe gaps after drag threshold

### LINKED ISSUE #4253: Hide safe gaps after drag threshold, text safe gaps only after resize

ruggi:
Fixes #4253 

**Problem:**

1. Safe gaps are immediately hidden when an interaction starts, but that can become clunky when i.e. double-clicking a text element to enter edit mode.
2. Text elements always show safe gaps, but this is not necessarily a good thing (they are below the small element threshold).

**Fix:**

1. Hide the safe gaps only after the drag threshold is met
2. Only show safe gaps for text elements that were resized past their initial `max-content` dimensions

https://github.com/concrete-utopia/utopia/assets/1081051/8a24a5ae-6025-4ee4-8cc5-3e73ae346537

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/interaction-state.ts
+++ b/editor/src/components/canvas/canvas-strategies/interaction-state.ts
@@ -41,6 +41,7 @@ export interface DragInteractionData {
   _accumulatedMovement: CanvasVector
   spacePressed: boolean
   zeroDragPermitted: ZeroDragPermitted // Will still complete the interaction with no drag distance applied.
+  hasBeenPastThreshold: boolean
 }
 
 export interface HoverInteractionData {
@@ -201,6 +202,7 @@ export function createInteractionViaMouse(
       _accumulatedMovement: zeroCanvasPoint,
       spacePressed: false,
       zeroDragPermitted: zeroDragPermitted,
+      hasBeenPastThreshold: false,
     },
     activeControl: activeControl,
     lastInteractionTime: Date.now(),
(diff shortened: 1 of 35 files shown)


seanparsons (approved):

editor/src/components/canvas/canvas-strategies/interaction-state.ts
```diff
@@ -347,6 +352,7 @@ function updateInteractionDataViaMouse(
             _accumulatedMovement: zeroCanvasPoint,
             spacePressed: false,
             zeroDragPermitted: currentData.zeroDragPermitted,
+            hasBeenPastThreshold: false,
```
balazsbajorics, line 355: Actually `drag` is null until it exceeded the threshold (but once it did exceed it, it stays not-null even if you go back near zero magnitude), so you can just nullcheck that. there is also the unfortunately named `isNotYetStartedDragInteraction` helper that relies on the null value

balazsbajorics (approved):
I like the new behavior! but I wouldn't add `hasBeenPastThreshold` to the interactionState to avoid inflating it and instead use drag != null.

ruggi:
> I like the new behavior! but I wouldn't add `hasBeenPastThreshold` to the interactionState to avoid inflating it and instead use drag != null.

oh man, nice! yeah I added that because I did not figure out the `!= null` part, it actually solves it! removed `hasBeenPastThreshold` in [`b341491` (#4254)](https://github.com/concrete-utopia/utopia/pull/4254/commits/b3414916a9da2955513474df3248bcdda8a0e366)

PR #4258, opened 2023-09-28
Chore/grid panels horizonal resizing

balazsbajorics:
**Description:**
This PR adds resizing functionality to the grid! There are now resize handles you can drag and violently shake! 

**Commit Details:**
- Move the StoredLayout to a jotai Atom: `GridPanelsStateAtom`
- Added GridColumnResizeHandle which works very smilarly to re-resizable's resize handles, including the big mouse catcher div that defeats VSCode's insanely agressive mouse-eating behavior
- StoredLayout's columns are no longer just an Array<StoredPanel>, they now have width too
- Not just one width but two widths!!!
- One width for the case when the column only holds Panels (ie the code editor, in the future maybe preview, others) – so the Large Mode
- One width for when the column holds at least one Menu (navigator / inspector) – ie the Mini Mode
- Resizing a column resizes the active mode's corresponding width
- An empty column gets deleted which resets the settings
- Menu (ie Mini) resize is clamped between 200px - 300px


editor/src/components/canvas/grid-panels-state.tsx
```diff
@@ -223,3 +258,54 @@ export function updateLayout(
...
+  const column = panelState[normalizeColIndex(columnIndex)]
+  if (column.panels.length === 0) {
+    // empty columns are zero width
+    return 0
+  }
+  const width = column.panels.some((p) => p.type === 'menu') ? column.menuWidth : column.paneWidth
+  return width
+}
+
+// export function usePanelHorizontalReize(): (columnIndex: number, newWidth: number) => void {}
```
ruggi, line 311: can this be zapped?

ruggi (approved):

editor/src/components/canvas/grid-panels-state.tsx
balazsbajorics: oops fixed

enidemi (approved):

PR #4266, opened 2023-09-28
Delete existing empty spans when leaving text edit mode

### LINKED ISSUE #4265: Delete empty text elements when leaving text edit mode

ruggi:
Fixes #4265

**Problem:**

When leaving text edit mode, new elements are deleted, but existing ones are kept.

**Fix:**

Update the behavior so it also removes _existing_ empty elements if:
1. the element is a span
3. _and_ it does not have event handlers

### DIFF
--- a/editor/src/components/text-editor/text-editor.spec.browser2.tsx
+++ b/editor/src/components/text-editor/text-editor.spec.browser2.tsx
@@ -357,7 +357,7 @@ describe('Use the text editor', () => {
   })
   describe('blur', () => {
     describe('when the element is empty', () => {
-      it('deletes existing elements', async () => {
+      it('keeps existing elements', async () => {
         const editor = await renderTestEditorWithCode(projectWithText, 'await-first-dom-report')
 
         await enterTextEditMode(editor)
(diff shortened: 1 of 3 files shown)


balazsbajorics (changes requested):
I tried the try me link and something is very broken, the Enter acts like a delete button for divs that don't already contain text, and text insertion also immediately quits.

Bug aside, I think for non-new elements we should be more careful, and make sure it's a span and not a div and/or that it doesn't have dimensions, backgroundColor, etc. So in general, we want the feature to clean up invisible leftover, but do not cause a headache if you have something selected and you accidentally press Enter.

ruggi:
> I tried the try me link and something is very broken, the Enter acts like a delete button for divs that don't already contain text, and text insertion also immediately quits.
> 
> Bug aside, I think for non-new elements we should be more careful, and make sure it's a span and not a div and/or that it doesn't have dimensions, backgroundColor, etc. So in general, we want the feature to clean up invisible leftover, but do not cause a headache if you have something selected and you accidentally press Enter.

Updated - the behavior only triggers for span elements without styling nor event handlers!

balazsbajorics (approved):

bkrmendy (approved):

PR #4270, opened 2023-09-29
remove bg color on frame to group conversion

bkrmendy:
## Problem
When converting a frame to a group, the background color is preserved.

## Fix
Remove the background props the same way padding is removed, and show a toast that communicates this

### DIFF
--- a/editor/src/core/model/margin-and-padding.ts
+++ b/editor/src/core/model/margin-and-padding.ts
@@ -33,10 +33,25 @@ export const PaddingPropertyPaths: Array<PropertyPath> = [
   PP.create('style', 'paddingTop'),
 ]
 
+export const BackgroundPropertyPaths: Array<PropertyPath> = [
+  PP.create('style', 'background'),
+  PP.create('style', 'backgroundAttachment'),
+  PP.create('style', 'backgroundClip'),
+  PP.create('style', 'backgroundColor'),
+  PP.create('style', 'backgroundOrigin'),
+  PP.create('style', 'backgroundPosition'),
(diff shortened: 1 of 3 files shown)


editor/src/core/model/margin-and-padding.ts
```diff
@@ -33,10 +33,25 @@ export const PaddingPropertyPaths: Array<PropertyPath> = [
...
 
+export const BackgroundPropertyPaths: Array<PropertyPath> = [
+  PP.create('style', 'background'),
+  PP.create('style', 'backgroundAttachment'),
+  PP.create('style', 'backgroundClip'),
+  PP.create('style', 'backgroundColor'),
+  PP.create('style', 'backgroundOrigin'),
+  PP.create('style', 'backgroundPosition'),
+  PP.create('style', 'backgroundRepeat'),
+  PP.create('style', 'backgroundSize'),
```
Rheeseyb, line 44: `backgroundImage`, `backgroundBlendMode`, `backgroundPositionX`, and `backgroundPositionY` if we want to be exhaustive

Rheeseyb (approved):

editor/src/core/model/margin-and-padding.ts
bkrmendy: Thanks! Added them in https://github.com/concrete-utopia/utopia/pull/4270/commits/31e48c4382fb6f0dbf39fba6e0b47dde8b59147c

ruggi (approved):

seanparsons (approved):

PR #4272, opened 2023-09-29
Border radius control adds overflow hidden

gbalint:
**Problem:**
Border radius often doesn’t do anything because there’s no overflow: hidden / scroll etc set.

**Fix:**
If no overflow property present at all, add overflow hidden and fire toast (“Element now hides overflowing content”)


Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/set-border-radius-strategy.tsx
```diff
@@ -123,6 +125,16 @@ export const setBorderRadiusStrategy: CanvasStrategyFactory = (
       ? 'individual'
       : 'all'
 
+  const addOverflowHiddenCommand = addPropIfDoesntExist(
+    'always',
+    selectedElement,
+    PP.create('style', 'overflow'),
+    'hidden',
+    'Element now hides overflowing content',
+  )
```
bkrmendy, line 134: Have you checked how hard it would be to do the same thing by composing `setProperty` and `showToastCommand`, instead of creating a bespoke command?

bkrmendy (approved):

seanparsons (approved):

PR #4273, opened 2023-09-29
Ensure updating the remix router also updates the navigation data

Rheeseyb:
**Problem:**
When working on a Remix project, it is possible to get "stuck" on a 404 route (the navigation toolbars stop working) if a change is made to the project whilst one is showing

**Fix:**
The problem here is that a change to the project's routes will replace the `router` with a new one, but we weren't updating the `navigationData` when this happened, meaning the navigation functions would still be using the previous `router` until the new one had triggered any form of navigation. This meant that hitting the "back" button would still call the loader function for the correct route, but then that route would never be rendered since this was happening via a `router` that wasn't being used as the actual `RouterProvider` for the remix app.

### DIFF
--- a/editor/src/components/canvas/remix/utopia-remix-root-component.tsx
+++ b/editor/src/components/canvas/remix/utopia-remix-root-component.tsx
@@ -219,14 +219,14 @@ export const UtopiaRemixRootComponent = React.memo((props: UtopiaRemixRootCompon
     (
       innerRouter: RouterType,
       location: Location,
-      updateEntries: 'update-entries' | 'do-not-update-entries',
+      updateEntries: 'append-to-entries' | 'replace-entries',
     ) => {
       setNavigationData((current) => {
         const key = EP.toString(basePath)
         const existingEntries = current[key]?.entries ?? []
 
         const shouldUpdateEntries =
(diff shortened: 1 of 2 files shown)


gbalint (approved):

editor/src/components/canvas/remix/utopia-remix-root-component.tsx
```diff
@@ -215,16 +215,23 @@ export const UtopiaRemixRootComponent = React.memo((props: UtopiaRemixRootCompon
     return createMemoryRouter(routes, { initialEntries: initialEntries })
   }, [routes])
 
-  const updateNavigationData = React.useCallback(
-    (innerRouter: RouterType, location: Location) => {
+  const setNavigationDataForRouter = React.useCallback(
+    (
+      innerRouter: RouterType,
+      location: Location,
+      updateEntries: 'update-entries' | 'do-not-update-entries',
```
balazsbajorics, line 222: nitpick, but `append-to-entres | replace-entries` would be much clearer form e

balazsbajorics (approved):
I left a nit to pick

bkrmendy (approved):

seanparsons (approved):

PR #4276, opened 2023-09-29
realign pane tabs

lankaukk:
Fixes #[ticket_number]

**Problem:**

**Fix:**

**Commit Details:**
- Did [other things]


bkrmendy (approved):

seanparsons (approved):

Rheeseyb (commented):
A bunch of tests have failed due to the results be 0.5-1px out. It might just mean that the mouse events originate or end in a slightly different place, which is fine, but please can you check that the test changes make sense.

PR #4282, opened 2023-10-02
Exclusive/complementary constraint resolution for group children

### LINKED ISSUE #4281: When setting constraints on group children, resolve stuttering over-constraining

ruggi:
Fixes #4281 

**Problem:**

It's possible to over-specify group child constraints, e.g. setting both `left`, `right`, and `width`.

Kapture 2023-10-02 at 11 54 13

**Fix:**

1. When setting a dimension constraint (w/h), if both the side complements (tlbr) are set, remove the oldest constraint
2. When setting a side constraint (tlbr), if both the dimension complements (w/h) are set, remove the dimension complement (so side pins win)

Kapture 2023-10-02 at 11 56 51

### DIFF
--- a/editor/src/components/inspector/fill-hug-fixed-control.tsx
+++ b/editor/src/components/inspector/fill-hug-fixed-control.tsx
@@ -785,6 +785,8 @@ export function getSafeGroupChildConstraintsArray(
   return value
 }
 
+export type ConstraintsMode = 'add' | 'remove'
+
 function setGroupChildConstraint(
   dispatch: EditorDispatch,
   option: GroupChildConstraintOptionType | 'toggle',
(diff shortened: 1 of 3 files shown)


editor/src/components/inspector/fill-hug-fixed-control.tsx
```diff
@@ -817,12 +809,80 @@ function setGroupChildConstraint(
...
+      default:
+        assertNever(dimension)
+    }
+  }
+
+  return Array.from(newProps)
+}
+
+function maybeRemoveOldestComplement(
+  set: Set<any>,
```
bkrmendy, line 867: since we set the `'data-constraints'` prop via the tool, I think it'd be worth it to validate its value `getSafeGroupChildConstraintsArray` so that we only take into account the valid values, and ignore whatever else that might have made its way into the prop (and get rid the `any`s stemming from `getSafeGroupChildConstraintsArray`

ruggi: makes sense, done in [`6f34c44` (#4282)](https://github.com/concrete-utopia/utopia/pull/4282/commits/6f34c44590fdfe7ef31f57915b07e2e24555672b)

bkrmendy (approved):

balazsbajorics (approved):

PR #4289, opened 2023-10-02
only round width/height when resizing

bkrmendy:
## Problem
When resizing, the `top` or `left` prop might have changed as a result of the edit

## Fix
Remove the rounding call that rounded all of `top`/`left`/`width`/`height` of the new frame, and only round `width` and `height`

## TODO
- [ ] fix wobblyiness


seanparsons:
Does this work as well when resizing the top left corner?

balazsbajorics:
> Does this work as well when resizing the top left corner?

extending on this, don't forget the case when an element has width, height, right, bottom props, and you are resizing the bottom right corner.

seanparsons (approved):

PR #4292, opened 2023-10-02
Check for existing scope on possible cyclic dependencies

Rheeseyb:
**Problem:**
Whilst evaluating a project inside the canvas, when we reach an import statement that has already been resolved we assume that it is a cyclic dependency, and fall back to the default import handling, meaning we won't wrap the imported components and therefore can't focus and edit them.

This can be shown by having a file A import file B, and then have multiple files import file A, which is shown [here](https://utopia.pizza/p/4e142c9b-zippy-spectrograph/) - in this project, the `Text` component cannot be focused, as the import statement `import { Text } from '/src/text'` inside `card.js` is evaluated twice. To see that working on this branch, test the project [here](https://utopia.pizza/p/4e142c9b-zippy-spectrograph/?branch_name=fix-multiple-import-handling)

**Fix:**
Rather than immediately assume that we have hit upon a cyclic dependency that we can't handle, we now keep a reference to the scope created by each import, and return that if it exists. If no such scope exists (and we have already resolved that import), _then_ we will fall back to the default import handling.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas.tsx
@@ -382,17 +382,18 @@ export const UiJsxCanvas = React.memo<UiJsxCanvasPropsWithErrorCallback>((props)
     MapLike<MapLike<ComponentRendererComponent>>
   >({})
 
-  let resolvedFiles = React.useRef<MapLike<Array<string>>>({}) // Mapping from importOrigin to an array of toImport
+  let resolvedFiles = React.useRef<MapLike<MapLike<any>>>({}) // Mapping from importOrigin to resolved scopes for the imported files
   resolvedFiles.current = {}
 
   const customRequire = React.useCallback(
     (importOrigin: string, toImport: string) => {
       if (resolvedFiles.current[importOrigin] == null) {
-        resolvedFiles.current[importOrigin] = []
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -597,162 +598,171 @@ export function attemptToResolveParsedComponents(
...
-          __esModule: true,
-        }
+  return foldEither(
+    (msg) => {
+      // This import in this file has already been resolved, so we check for an existing scope
+      const existingScope = resolvedFromThisOrigin[toImport]
+      return existingScope == null ? left(msg) : right(existingScope)
+    },
+    (resolvedFilePath) => {
+      resolvedFromThisOrigin[toImport] = null // We use null as a marker to indicate that we have started resolving, but not completed yet
```
balazsbajorics, line 608: :+1

balazsbajorics (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
bkrmendy: can't we use some other marker like `'started-resolving'`? I think the `!== undefined` (from above) is very unexpressive, and the shape of the code would stay the same

bkrmendy (approved):

PR #4293, opened 2023-10-02
use `Provider` for the jotai atoms

balazsbajorics:
## Problem
The `RemixNavigationAtom` wasn't cleaned up on editor unmount, so when running the browser tests, values written by previous tests got "stuck" in the atom, which caused unwanted and confusing behavior in the tests.

## Fix
Add a Provider, which clears all atoms when remounting (as per the [docs](https://jotai.org/docs/core/provider)). This way, we don't have to work around the crosstalk issue by coming up with separate storyboard uids for separate tests

### DIFF
--- a/editor/src/templates/editor.tsx
+++ b/editor/src/templates/editor.tsx
@@ -129,7 +129,7 @@ import {
 } from '../components/editor/store/project-server-state'
 import { GithubOperations } from '../core/shared/github/operations'
 import { GithubAuth } from '../utils/github-auth'
-import { Provider } from 'jotai'
+import { Provider as JotaiProvider } from 'jotai'
 
 if (PROBABLY_ELECTRON) {
   let { webFrame } = requireElectron()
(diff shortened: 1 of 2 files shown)


editor/src/templates/editor.tsx
```diff
@@ -706,21 +707,23 @@ export const EditorRoot: React.FunctionComponent<{
...
-                  <UiJsxCanvasCtxAtom.Provider value={spyCollector}>
-                    <EditorComponent />
-                  </UiJsxCanvasCtxAtom.Provider>
-                </LowPriorityStateContext.Provider>
-              </CanvasStateContext.Provider>
-            </DomWalkerMutableStateCtx.Provider>
-          </EditorStateContext.Provider>
-        </OriginalMainEditorStateContext.Provider>
-      </DispatchContext.Provider>
+      <Provider>
```
gbalint, line 710: Maybe we should give the Provider an alias in the import, e.g. `import { Provider as JotaiProvider } from 'jotai'` because we have so many providers here.

gbalint (approved):
Awesome!

seanparsons (approved):

editor/src/templates/editor.tsx
bkrmendy: Great idea, renamed it in https://github.com/concrete-utopia/utopia/pull/4293/commits/faf9be114c8f6d7b7d3dee7f2531a18df4e69977

Rheeseyb (approved):

PR #4298, opened 2023-10-03
Don't reparent out of groups (unless outside of a scene)

### LINKED ISSUE #4297: Don't reparent out of groups by default (unless moving outside of a scene)

ruggi:
Fixes #4297 

**Problem:**

Dragging a group child outside its parent element bounds triggers a reparent which can be super confusing.

Kapture 2023-10-03 at 11 31 48

**Fix:**

- When moving a group child around, prefer the move strategy over the reparenting ones (which can still be triggered with `Tab`)…
- …unless the reparent would happen outside of a `Scene`, in which case the reparenting is preferred.

https://github.com/concrete-utopia/utopia/assets/1081051/974821b0-2bf7-4eb7-b628-b90bc272aca7

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
@@ -74,7 +74,7 @@ export function absoluteMoveStrategy(
       fitness:
         interactionSession?.interactionData.type === 'DRAG' &&
         interactionSession?.activeControl.type === 'BOUNDING_AREA'
-          ? 1
+          ? 2
           : 0,
       apply: () => {
         if (
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
@@ -74,7 +74,7 @@ export function absoluteMoveStrategy(
       fitness:
         interactionSession?.interactionData.type === 'DRAG' &&
         interactionSession?.activeControl.type === 'BOUNDING_AREA'
-          ? 2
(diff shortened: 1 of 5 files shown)


editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
```diff
@@ -74,7 +74,7 @@ export function absoluteMoveStrategy(
       fitness:
         interactionSession?.interactionData.type === 'DRAG' &&
         interactionSession?.activeControl.type === 'BOUNDING_AREA'
-          ? 1
+          ? 2
```
Rheeseyb, line 77: Hmmm this change makes me suspicious about whether it will cause issues with other strategies. I believe we have covered a ton of cases in the tests, so it should be fine, but if we suddenly start seeing other strategies failing to fire when we'd expect them to, we will want to check if this is the cause

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
ruggi: yeah I'm a bit on the verge as well, but it _looks_ like this would be fine. Maybe it's worth to add a comment here, but I wonder what it should say 🤔

balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
ruggi: [`7755cdf` (#4298)](https://github.com/concrete-utopia/utopia/pull/4298/commits/7755cdf0b958030e92f095d3fc6ad582e79b611c)

PR #4303, opened 2023-10-03
Chore/refactor FillHugFixedControl

balazsbajorics:
**Problem:**
For my new simplified layout section I need the fill hug fixed dropdown.. which has been buried into a component that accidentally grew huge.

**Fix:**
Split the component into sub-components each using hooks. I've been meaning to do this change for months now, feels so good. Feels so clean.

**Commit Details:**
- First I inlined `DimensionRow` back into `FillHugFixedControl` because it used props instead of hooks. Then,
- split out `FixedHugDropdown`, with one prop: `width | height`
- changed GroupConstraintSelect to also only take `width | height`
- spun out `GroupChildPinControl`
- and finally, `WidthHeightNumberControl`
- The end result is that `FillHugFixedControl` is just organizational, the individual controls are self-governing and using shared selectors.
- On master there was a big duplication of event handlers / callbacks and selectors because a lot of code was explicitly written for width and then copy pasted for height. On this branch, these has been changed to use the `dimension: 'width' | 'height'` prop.
- This is a pure refactor, no behavior was meant to change. All the existing test coverage shall still work as it was.

**Note:**
I think it's better if you just read the resulting code zero-based, the diff is very messy.

### DIFF
--- a/editor/src/components/inspector/fill-hug-fixed-control.tsx
+++ b/editor/src/components/inspector/fill-hug-fixed-control.tsx
@@ -425,6 +425,7 @@ function useOnSubmitFixedFillHugType(dimension: 'width' | 'height') {
     }
   }, [
     dimension,
+    axis,
     allElementPropsRef,
     dispatch,
     metadataRef,


editor/src/components/inspector/fill-hug-fixed-control.tsx
```diff
@@ -286,371 +317,241 @@ export const FillHugFixedControl = React.memo<FillHugFixedControlProps>((props)
...
-        store.editor.allElementProps,
+    (store) =>
+      detectFillHugFixedStateMultiselect(
+        dimension === 'width' ? 'horizontal' : 'vertical',
+        metadataSelector(store),
+        selectedViewsSelector(store),
       ),
-    }),
-    'FillHugFixedControl constraints',
+    'FillHugFixedControl heightCurrentValue',
```
ruggi, line 451:
```suggestion
    'FixedHugDropdown currentValue',
```

editor/src/components/inspector/fill-hug-fixed-control.tsx
```diff
@@ -286,371 +317,241 @@ export const FillHugFixedControl = React.memo<FillHugFixedControlProps>((props)
...
-DimensionRow.displayName = 'DimensionRow'
+  const type = useEditorState(
+    Substores.metadata,
+    (store) =>
+      checkGroupChildConstraint(
+        dimension,
+        store.editor.selectedViews,
+        store.editor.allElementProps,
+      ),
+    'GroupConstraintSelect checkGroupChildConstraint',
```
ruggi, line 486:
```suggestion
    'GroupConstraintSelect type',
```

editor/src/components/inspector/fill-hug-fixed-control.tsx
```diff
@@ -286,371 +317,241 @@ export const FillHugFixedControl = React.memo<FillHugFixedControlProps>((props)
...
-      </UIGridRow>
-    )
-  },
-)
+const GroupConstraintSelect = React.memo(({ dimension }: { dimension: LayoutPinnedProp }) => {
+  const dispatch = useDispatch()
+  const colorTheme = useColorTheme()
 
-DimensionRow.displayName = 'DimensionRow'
+  const type = useEditorState(
```
ruggi, line 478: for clarity and to avoid stuttering I'd rename this to `constraintType`

editor/src/components/inspector/fill-hug-fixed-control.tsx
```diff
@@ -286,371 +317,241 @@ export const FillHugFixedControl = React.memo<FillHugFixedControlProps>((props)
...
+            ? fixedSizeDimensionHandlingText(
+                metadataRef.current,
+                elementPathTreeRef.current,
+                firstSelectedView,
+                currentComputedValue,
+              )
+            : currentComputedValue
+        const strategy = strategyForChangingFillFixedHugType(
+          valueToUse,
+          dimension === 'width' ? 'horizontal' : 'vertical',
```
seanparsons, line 412: Fairly certain you can use this.
```suggestion
          axis,
```

seanparsons (approved):

PR #4306, opened 2023-10-03
Always round the result of canvas interactions to the nearest whole pixel

Rheeseyb:
**Problem:**
It's far to easy to create or edit and element resulting in a .5px value in either the position pins or the size values.

**Fix:**
Always round the results of canvas interactions to the nearest whole pixel value. The result is that an element with a .5px starting value will have that value rounded when moving or resizing it, even if the affected side would have been otherwise untouched by the interaction (so e.g. a .5px value in the `top` pin would be rounded when dragging the right side to resize).

On top of this, it also means that whilst you can still snap to elements with .5px values, the snapped element will have whole number values, meaning the snap result will be .5px off.

**Note:**
There is an existing bug that is reflected in this PR - in the above case where the `top` pin has a .5px value, when resizing via the right side the `top` value will be increased by 1.5px. This is because the existing bug increases the `top` by 1px, and then the whole number rounding kicks in. Prior to this PR the `top` value would keep increasing by 1px on every resize

Also note that this PR makes the situation worse for elements with existing .5px values due to the split between how we apply snapping and how we communicate that snapping via the canvas controls. This is because we use measured values from the DOM, which can be +/- .002px off, meaning a .5px value can be rounded up or down as a result of that.


editor/src/components/text-editor/text-handling.ts
```diff
@@ -266,5 +265,5 @@ export function fixedSizeDimensionHandlingText(
 ): number {
   // Fixed dimensions for a text containing element need to be rounded up to prevent wrapping.
   const containsText = MetadataUtils.targetTextEditableAndHasText(metadata, pathTrees, elementPath)
-  return containsText ? roundUpToNearestHalf(dimensionValue) : dimensionValue
+  return containsText ? Math.ceil(dimensionValue) : dimensionValue
```
ruggi, line 268: should this be `roundUpToNearestWhole`? 🤔

ruggi (approved):

editor/src/components/text-editor/text-handling.ts
Rheeseyb: I was going to add that, but since that was just an alias for `Math.ceil` I didn't see the point in the end

ruggi: I guess the defensive take to prefer `roundUpToNearestHalf` here would be that this is kind of buried and this whole sub-integer pixel fight doesn't sound like game over

Rheeseyb: So the reason for this rounding in the first place is so that when hugging text we don't ever round the pixel value down, as that would cause the text to wrap

maltenuhn: > So the reason for this rounding in the first place is so that when hugging text we don't ever round the pixel value down, as that would cause the text to wrap

that feels like it should be something prominently visible in code, whether through a comment or a named function call. In six months, I could totally see someone refactoring this while trying to achieve something totally text-unrelated... and this capability accidentally goes the way of the dodo. Unless there's a test that'd break ofc.

Rheeseyb: There's a comment 2 lines above this. I just checked and the test doesn't exist though, so I can totally add that

balazsbajorics (approved):

editor/src/components/canvas/guideline.ts
```diff
@@ -217,43 +247,14 @@ export const Guidelines = {
...
-    const xRight = frame.x + frame.width
-    const yTop = frame.y
-    const yBottom = frame.y + frame.height
-    const xs = includeCentre
-      ? [frame.x, frame.x + frame.width / 2, frame.x + frame.width]
-      : [frame.x, frame.x + frame.width]
-    const ys = includeCentre
-      ? [frame.y, frame.y + frame.height / 2, frame.y + frame.height]
-      : [frame.y, frame.y + frame.height]
+    const { xs, ys, xLeft, xRight, yTop, yBottom } = pointsForGuidelinesForFrame(
```
seanparsons, line 254: That's real nice.

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/canvas-strategies.spec.browser2.tsx
```diff
@@ -504,14 +504,14 @@ describe('Snapping guidelines for absolutely moved element', () => {
     const renderResult = await getGuidelineRenderResult(4)
 
     expect(await getGuidelineDimensions(renderResult, 'guideline-0')).toEqual({
-      left: '110.375px',
-      top: '156.375px',
+      left: '109.875px',
```
balazsbajorics, line 507: this looks insane :D

editor/src/components/text-editor/text-handling.ts
Rheeseyb: Ok I've just been writing a test for this, and it turns out that actually this functionality has been broken for a while. That's because wherever this function is called, we're using measured values that have already been rounded.

PR #4312, opened 2023-10-04
Better "Hug Contents" detection

gbalint:
**Problem:**
We only detect the "Hug Contents" property of an element's dimension (width/height) when width/height is `max-content`.
But there are other possible width/height values which result in hugging the parent, e.g. `min-content`.

**Fix:**
Include these other possibilities in the `isHugFromStyleAttribute` function.

### DIFF
--- a/editor/src/components/inspector/inspector-strategies/fixed-fill-hug.spec.browser2.tsx
+++ b/editor/src/components/inspector/inspector-strategies/fixed-fill-hug.spec.browser2.tsx
@@ -34,7 +34,7 @@ import {
   selectOptionLabel,
 } from '../fill-hug-fixed-control'
 import type { Axis, FixedHugFillMode } from '../inspector-common'
-import { MaxContent } from '../inspector-common'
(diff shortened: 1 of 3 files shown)


editor/src/components/inspector/inspector-strategies/fixed-fill-hug.spec.browser2.tsx
```diff
@@ -553,6 +553,69 @@ describe('Fixed / Fill / Hug control', () => {
...
+            projectWithParentWidthHeight(width, MaxContent),
+            'await-first-dom-report',
+          )
+          await select(editor, 'parent')
+
+          const hugButtons = await editor.renderedDOM.findAllByText(HugContentsLabel)
+          expect(hugButtons).toHaveLength(2)
+        }),
+      )
+      NonHugContentAttrValues.forEach((width) =>
```
balazsbajorics, line 600: it seems like this width is not actually used, only in the title!

PR #4314, opened 2023-10-04
Convert group child percentage pins to pixel based pins.

### LINKED ISSUE #4299: If a group child has a size of 100% then the group handling breaks.
- Show warning triangle.
- Auto-fix child size to `px` on first group interaction.
- When the editor is grouping an element, make sure all properties are px and not %.

seanparsons:
Fixes #4299

**Problem:**
Percentage based pins for the children of a group don't really make any sense.

**Fix:**
In the following cases we now replace the pins of the children of a group with pixel based pins regardless of whatever they had before:
- When creating a group around some existing elements.
- When adjusted the size of a group, which in turn updates properties on the children.

Should a percentage based pin be replaced in this way, we show a toast indicating that one or more of those has been replaced.

**Commit Details:**
- Added `whenReplacingPercentageValues` property to `SetCssLengthProperty`.
- `runSetCssLengthProperty` now adds an appropriate toast when replacing a percentage based property and when it is instructed to warn in this particular case.
- `setElementPins` now replaces the children's pin values with pixel based ones in all cases regardless of what was there before.
- `runPushIntendedBoundsAndUpdateGroups` ensures that any toasts are carried forward in the editor state patch.
- The patching part of `runShowToastCommand` has been extracted into a utility function `addToastPatch`.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-resize.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-resize.spec.browser2.tsx
@@ -265,7 +265,7 @@ describe('Keyboard Absolute Resize E2E', () => {
     })
   })
 
-  it('keeps trueuing up groups as directions change', async () => {
+  it('keeps trueing up groups as directions change', async () => {
     const renderResult = await renderTestEditorWithCode(
       formatTestProjectCode(
         makeTestProjectCodeWithSnippet(`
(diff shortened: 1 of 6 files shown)


editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-resize.spec.browser2.tsx
```diff
@@ -265,7 +265,7 @@ describe('Keyboard Absolute Resize E2E', () => {
     })
   })
 
-  it('keeps trueuing up groups as directions change', async () => {
```
balazsbajorics, line 268: haha it's like queueing but trueueing

balazsbajorics (approved):

Rheeseyb (approved):

PR #4316, opened 2023-10-04
Feature/simple layout section

balazsbajorics:
**Feature:**
If you enable the FS named `Simplified Layout Section`, you will be greeted with a pretty familiar looking inspector:
image

**What works**
- The L, T, W, H number controls work for single selection. They always show the value converted to pixels, and they always show left, top, width, height, regardless of the Pinning of the element
- Below Width and Height you can find the Fixed, Fill, Hug dropdown which is the same as on master, just moved into this new place. It should work most of the time
- If the element is not a Group child, and it is the child of a Frame (for our evolving definition of Frame), the Constraints section shows the Constraints / Pins of the element, which is essentially describing what is the combination of `left, top, right, bottom, width, height` props that are actually set on the element. It is abstracted away and the options match Figma, with one exception: We don't **yet** have CenterX and CenterY pinning.
- If the element selected is a Group child, we show the group Constrains section, same what is on master today. This section is now a little bit more complicated than the simplified figma-matching one, something to think about.

**What doesn't work**
- The L, T, W, H number controls are disabled (for now) if you multiselect elements
- The new Constraints section doesn't have an option for center pinning (yet)
- We don't yet have the border radius control in the top section
- We don't yet have the buttons to trigger "Hug children", "Fill parent" etc, you have to use the Fixed, Fill, Hug dropdown.

**Commit Details:**
- New `SimplifiedLayoutSubsection`
- New `FrameUpdatingLayoutSection` which always shows L T W H "measured" px values and figures out what frame properties need updating when you change a value
- New `ConstraintsSection` that is vastly simplified compared to the existing one, with a dropdown for each dimension enumerating every allowed combination.
- Vastly simplified Pinning code that translates the dropdown options to a combination of `left, top, right, bottom, width, height` props.


editor/src/components/inspector/constraints-section.tsx
```diff
@@ -0,0 +1,259 @@
...
+  )
+})
+ConstraintsSection.displayName = 'ConstraintsSection'
+
+const GroupChildConstraintsSection = React.memo(() => {
+  return (
+    <FlexColumn css={{ paddingBottom: UtopiaTheme.layout.rowHorizontalPadding }}>
+      <UIGridRow padded variant='<-auto-><----------1fr--------->'>
+        <GroupChildPinControl />
+        <FlexColumn css={{ gap: 0 }}>
```
ruggi, line 73: imho this should match the height of the pin control

editor/src/components/inspector/sections/layout-section/self-layout-subsection/simplified-layout-subsection.tsx
```diff
@@ -0,0 +1,40 @@
...
+        </FlexRow>
+      </InspectorSubsectionHeader>
+      <FlexColumn style={{ gap: 8, paddingLeft: 8, paddingRight: 8 }}>
+        <FrameUpdatingLayoutSection />
+        <UIGridRow
+          padded={false}
+          variant='<--1fr--><--1fr-->'
+          style={{ minHeight: undefined, gap: 4 }}
+        >
+          <FixedHugDropdown dimension='width' />
```
ruggi, line 30: noice! I think having an inner label here would be clearer: `H`orizontal and `V`ertical would be nice, but there would be a stuttering `H`. What about the dimension control icons?

editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-resize-strategy.tsx
```diff
@@ -111,6 +117,97 @@ function getFitness(interactionSession: InteractionSession | null): number {
...
+  originalFrame: CanvasRectangle,
+  originalIntendedBounds: Array<CanvasFrameAndTarget>,
+  edgePosition: EdgePosition,
+  movement: CanvasVector,
+): ChangeBoundsResult {
+  let commands: Array<CanvasCommand> = []
+  let intendedBounds: Array<CanvasFrameAndTarget> = originalIntendedBounds
+
+  selectedElements.forEach((selectedElement) => {
+    const element = withUnderlyingTarget(selectedElement, projectContents, null, (_, e) => e)
```
Rheeseyb, line 138: `getElementFromProjectContents`

seanparsons: This was extracted out from its original location.

gbalint: Be careful with `getElementFromProjectContents`, it only returns an element if it is a JSXElement (I'm changing that in my in-progress PR)

editor/src/components/inspector/constraints-section.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+  const activeOption =
+    dimension === 'width'
+      ? HorizontalPinChangeOptionsIncludingMixed[pins.horizontal]
+      : VerticalPinChangeOptionsIncludingMixed[pins.vertical]
+
+  return (
+    <PopupList
+      id={`frame-child-constraint-${dimension}`}
+      // eslint-disable-next-line react/jsx-no-bind
```
Rheeseyb, line 239: 👀 
Or just put it in a `useCallback`

editor/src/components/inspector/sections/layout-section/self-layout-subsection/frame-updating-layout-section.tsx
```diff
@@ -0,0 +1,271 @@
...
+  currentValue: number
+  updateFrame: (edgePosition: EdgePosition, movement: CanvasVector) => void
+  enabled: boolean
+}
+
+const FrameUpdatingLayoutControl = React.memo((props: LayoutPinPropertyControlProps) => {
+  const { property, currentValue, updateFrame } = props
+  const pointInfo = useInspectorLayoutInfo(props.property)
+
+  const [mountCount, forceUpdate] = React.useReducer((c) => c + 1, 0)
```
Rheeseyb, line 220: What's the deal with this? Can you add a comment explaining why we need this please

editor/src/components/inspector/sections/layout-section/self-layout-subsection/simplified-layout-subsection.tsx
Rheeseyb: Yeah, this definitely needs either the icons or a label to clarify what these dropdowns apply to

Rheeseyb (approved):

editor/src/components/inspector/constraints-section.tsx
balazsbajorics: fixed!

editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-resize-strategy.tsx
balazsbajorics: this is just moved code AFAIK

editor/src/components/inspector/sections/layout-section/self-layout-subsection/frame-updating-layout-section.tsx
balazsbajorics: I added a comment, and better yet, also fixed the variable name so it actually is self-explanatory

editor/src/components/inspector/constraints-section.tsx
balazsbajorics: done!

ruggi (approved):

gbalint (approved):

balazsbajorics:
I'm gonna merge this as soon as the tests pass :)

PR #4318, opened 2023-10-04
Set collapsing parent size to fixed when converting to absolute

gbalint:
**Problem:**
An element's dimension can be defined by its children (similarly to the "Hug Contents" size property in the inspector). In this case converting a child to absolute can collapse the parent to smaller or even zero size.

Note: This can also happen if you delete/reparent the child from the parent, but that is not the scope of this PR.

**Fix:**
In the convert-to-absolute strategy check if the parent is a "collapsing" one, and set its size to fixed to make sure it doesn't collapse.

Note: we decide whether a parent is a collapsing one based on its style props. Later, we will need a smarter solution, which can recognize this property from CSS rules coming from anywhere.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.tsx
@@ -77,6 +77,7 @@ import { memoize } from '../../../../core/shared/memoize'
 import { childInsertionPath } from '../../../editor/store/insertion-path'
 import type { ElementPathTrees } from '../../../../core/shared/element-path-tree'
 import { cssPixelLength } from '../../../inspector/common/css-utils'
+import type { ProjectContentTreeRoot } from '../../../assets'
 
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.tsx
```diff
@@ -613,3 +627,77 @@ function getFrameWithoutMargin(
...
+
+  return []
+}
+
+const CollapsingWidthHeightValues = ['max-content', 'min-content', 'fit-content', 'auto']
+
+function isCollapsingParent(element: JSXElement, property: 'width' | 'height') {
+  const simpleAttribute = defaultEither(
+    null,
+    getSimpleAttributeAtPath(right(element.props), PP.create('style', property)),
```
balazsbajorics, line 699: is it possible to base this on dom metadata? we could even extend specialSizeMeasurements if needed.

bkrmendy (approved):

balazsbajorics (approved):

PR #4330, opened 2023-10-05
only add `contain: layout` when necessary

bkrmendy:
## Problems
1. `contain: layout` should be added to the parent after inserting an element with the absolute reparent strategy
2. `contain: layout` should not be added to elements that don't support children when they are converted to absolute

## Fix
1. Check whether `contain: layout` is needed, and it if necessary
2. Refine the ruleset in `AddContainLayoutIfNeeded`


seanparsons:
Does this need any aligning with `fixParentContainingBlockSettings` or vice-versa?

balazsbajorics (approved):

Rheeseyb (approved):

PR #4334, opened 2023-10-06
Gap controls centered relatively to content

### LINKED ISSUE #4333: Place flex gap handles relative to the contents

ruggi:
Fixes #4333

**Problem:**

Flex gap handles are centered relatively to the container size. This means that they can become hard to find if the container.

E.g.

Screenshot 2023-10-06 at 11 46 15 AM

**Fix:**

1. Place the control handles centered relatively to the content bounds, not the container bounds
2. Make the control handles `brandNeonPink` to make them easier to spot

Screenshot 2023-10-06 at 11 47 18 AM

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/set-flex-gap-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/set-flex-gap-strategy.tsx
@@ -125,7 +125,10 @@ export const setFlexGapStrategy: CanvasStrategyFactory = (
       resizeControl,
       controlWithProps({
         control: FloatingIndicator,
-        props: { ...props, color: colorTheme.gapControls.value },
+        props: {
+          ...props,
+          color: colorTheme.brandNeonPink.value,
+        },
         key: 'padding-value-indicator-control',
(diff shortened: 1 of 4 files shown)


editor/src/components/canvas/canvas-strategies/strategies/set-flex-gap-strategy.tsx
```diff
@@ -125,7 +125,10 @@ export const setFlexGapStrategy: CanvasStrategyFactory = (
       resizeControl,
       controlWithProps({
         control: FloatingIndicator,
-        props: { ...props, color: colorTheme.gapControls.value },
```
seanparsons, line 128: Does this mean that `gapControls` should be updated?

ruggi: mmmh it's still there for the striped background, maybe worth renaming it to `gapControlsBg` or something like that?

seanparsons (approved):

Rheeseyb (approved):

PR #4335, opened 2023-10-06
Remove unnecessary DerivedState from commands

gbalint:
**Problem:**
DerivedState was added to CanvasCommands as an input parameter, because for a while it was necessary to be able to access the remix routing table from commands. However, this is unused since the refactoring of `normalisePathToUnderlyingTarget` ( https://github.com/concrete-utopia/utopia/pull/4185 ), because we no longer need the remix routing table to find a file for a uid in the project.

**Fix:**
Remove all derivedState parameters from commands and handle the fallout from that.


seanparsons (approved):

bkrmendy (approved):

ruggi (approved):
📜

PR #4338, opened 2023-10-06
Replace moveTemplate with inspector strategies

gbalint:
**Problem:**
Our legacy `moveTemplate` is only used for unwrapping. Because we would like to finetune unwrapping behavior, it is time to replace `moveTemplate` with inspector strategies.
Fortunately, unwrapping behavior is covered with tests quite well.

**Fix:**
To reduce the scope of changes, I did not want to put all of the unwrapping logic into strategies, just the part which was covered by moveTemplate. I created two new strategies, one which moves the content to the new parent as absolute, and one which converts to absolute first.
Because the unwrapping logic needed the new path which was created by the strategies, I created a new kind of InspectorStrategy called CustomInspectorStrategy which can return a custom data field, and not just the commands themselves. In a subsequent PR all InspectorStrategies can be converted to CustomInspectorStrategy (and then it can be renamed back to InspectorStrategy), but making that change in this PR would have made a lot of noise.
After making these changes I could delete all moveTemplate code and the functions which were only called from there.


bkrmendy (approved):

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -5916,16 +5916,11 @@ export var storyboard = (
...
               }}
               data-uid='qux'
             />
             <div
-              style={{
-                backgroundColor: 'orange',
-                height: 60,
-                top: 170,
-              }}
+              style={{ backgroundColor: 'orange', height: 60 }}
```
balazsbajorics, line 5923: we lost a top: 170 here, is this a problem?

balazsbajorics (approved):

Rheeseyb (approved):

PR #4344, opened 2023-10-09
Reuse strategy code in the insert popup

bkrmendy:
## Problem
The insert popup should re-use code from the strategies to insert the selected element

## Fix
Reuse `getReparentOutcome`, `autoLayoutParentAbsoluteOrStatic`, `getAbsoluteReparentPropertyChanges` and `getStaticReparentPropertyChanges`. These functions are used by the strategies as well, but the strategies themselves cannot be readily reused, since the actual insertion logic is mixed up with canvas interaction-specific code (check where the above functions are used)


editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -167,6 +167,7 @@ import type { ElementPathTrees } from '../../core/shared/element-path-tree'
 import { createPasteToReplacePostActionActions } from '../canvas/canvas-strategies/post-action-options/post-action-options'
 import { generateConsistentUID } from '../../core/shared/uid-utils'
 import { getAllUniqueUids } from '../../core/model/get-unique-ids'
+import { UTOPIA_UID_KEY } from '../../core/model/utopia-constants'
```
balazsbajorics, line 170: unused imports

editor/src/components/editor/insert-callbacks.ts
```diff
@@ -134,32 +173,266 @@ function useEnterDrawToInsertForElement(elementFactory: (newUID: string) => JSXE
   )
 }
 
-export function useToInsert(): (convertTo: InsertMenuItem | null) => void {
+const insertAsAbsoluteStrategy = (
```
balazsbajorics, line 176: if we keep the strategie(s), please move them to dedicated file(s)

editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy-multi-file.spec.browser2.tsx
```diff
@@ -309,7 +309,7 @@ export var App = () => {
       <FlexParent data-uid='flexparentinapp' />
       <AbsoluteChild
         data-uid='absolutechildinparent'
-        style={{ left: 294, top: -40 }}
+        style={{ left: 294, top: -40, position: 'absolute' }}
```
Rheeseyb, line 312: This shouldn't be needed, because the component already has `position: 'absolute'` on its root element.

editor/src/components/editor/actions/actions.tsx
```diff
@@ -615,6 +615,13 @@ export function editorMoveMultiSelectedTemplates(
   editor: EditorModel
   newPaths: Array<ElementPath>
 } {
+  if (newParent == null) {
+    return {
+      editor: editor,
+      newPaths: [],
+    }
+  }
```
Rheeseyb, line 623: The fact that this wasn't early returning before makes me very suspicious

bkrmendy: `getReparentOutcome` used to add a default for this "Cater for something being reparented to the canvas", I removed it because it's essentially feature logic tucked away in a library function

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-property-changes.ts
```diff
@@ -183,6 +180,7 @@ export function getAbsoluteReparentPropertyChanges(
 
   return [
     adjustCssLengthProperties('always', target, newParentFlexDirection, edgePropertiesToAdjust),
+    setProperty('always', target, PP.create('style', 'position'), 'absolute'),
```
Rheeseyb, line 183: This really shouldn't be needed here. We have a strategy for converting to absolute before reparenting, which is why this hasn't been needed before.

editor/src/components/editor/insert-callbacks.ts
```diff
@@ -134,32 +175,146 @@ function useEnterDrawToInsertForElement(elementFactory: (newUID: string) => JSXE
...
-        )
-        dispatch(actions, 'everyone')
+    (elementToInsert: InsertMenuItem | null) => {
+      if (elementToInsert == null) {
+        return
+      }
+      const targetParent: ElementPath | null = safeIndex(selectedViewsRef.current, 0) ?? null
+
+      if (targetParent == null) {
+        return
```
Rheeseyb, line 196: This should probably show a toast, otherwise it fails silently in legitimate cases (i.e. nothing is selected and I click the insert menu)

editor/src/components/editor/actions/actions.tsx
Rheeseyb: Interesting, it looks like that was never the case anyway.

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-property-changes.ts
bkrmendy: removed in https://github.com/concrete-utopia/utopia/pull/4344/commits/8699f0e10b0af192fd8f3bf4208eadc7d61f3d1f

editor/src/components/editor/insert-callbacks.ts
bkrmendy: [move strategies to separate files](https://github.com/concrete-utopia/utopia/pull/4344/commits/ced8a9282bbb430f89b11c24948aa51623c6ca8d)

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -6718,12 +6718,6 @@ export var storyboard = (
         </div>`,
         ),
       )
-      expect(renderResult.getEditorState().editor.toasts).toHaveLength(1)
-      const firstToast = safeIndex(renderResult.getEditorState().editor.toasts, 0)
-      expect(firstToast?.level).toEqual('INFO')
-      expect(firstToast?.message).toEqual(
-        "Added `contain: 'layout'` to the parent of the newly added element.",
```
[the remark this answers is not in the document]
bkrmendy: no clue why that got removed, thanks for noticing it, I restored it in https://github.com/concrete-utopia/utopia/pull/4344/commits/b22b6f3b6e8a3ab6ffa8131b9c95707710e67c66

gbalint (approved):

Rheeseyb (approved):

PR #4345, opened 2023-10-09
Remove findElementAtPath usages

gbalint:
**Problem:**
`findElementAtPath` is a legacy function. It traverses through the hierarchy using the element path structure, and this method is not compatible with Remix projects (where the Outlet is replaced with a Remix route dynamically). After the refactor in https://github.com/concrete-utopia/utopia/pull/4185 , `withUnderlyingTarget` can directly access the necessary element by its unique uid, and `getElementFromProjectContents ` uses `withUnderlyingTarget`. So we should replace all `findElementAtPath` usages with `getElementFromProjectContents` calls.

**Fix:**
- I renamed `getElementFromProjectContents` to `getJSXElementFromProjectContents` because it was really misleading that it only returned the found element if it was a `JSXElement` (so it was similar to `findJSXElementAtPath` and not `findElementAtPath`)
- I readded a function called `getElementFromProjectContents`, but without the filtering to `JSXElement`-s
- I removed the unused `getJSXElementTagName`,  `getJSXElementBaseName` and `findJSXElementLikeAtPath` functions
- I simplified the code in some places and structured it in "happy path to the left" style
- `moveTemplate` uses `findElementAtPath` too. Since `moveTemplate` is really legacy code, instead of refactoring it, we should try to remove its usages and rewrite those features with strategies. I removed the `moveTemplate` dependency from SET_Z_INDEX and `setsetZIndexOnSelected`. `moveTemplate` is still used in `UNWRAP_ELEMENTS`, but refactoring that is not in the scope of this PR.


editor/src/components/editor/actions/actions.tsx
```diff
@@ -1569,17 +1562,9 @@ export const UPDATE_FNS = {
...
-      action.indexPosition,
-      EP.parentPath(action.target),
-      null,
-      editor,
-      null,
-      null,
-    ).editor
+    return foldAndApplyCommandsSimple(editor, [
+      reorderElement('always', action.target, action.indexPosition),
+    ])
```
bkrmendy, line 1567: 👩‍🍳👌

bkrmendy (approved):

balazsbajorics (approved):

PR #4346, opened 2023-10-09
Fix uids before inserting an element from the insert popup

bkrmendy:
## Problem
Inserting two `img` elements of the same type during the same editing session via the insert popup causes an "irrecoverable error"

## Cause
Since we create the elements that populate the insert menu at module resolution time, each element is created only once. Because of the default uid parameter in the `jsExpressionValue` function, the uids of the props in these elements will always be the same. Even though we pass a new uid for the element, the props of the elements will have the same uids. For example, if you insert an `img` from the popup menu, the uid of the `width` prop will always be the same.
Relevant snippets:
- https://github.com/concrete-utopia/utopia/blob/1baddca50db35ee2bb9e8ee54b04895af7c83aa4/editor/src/core/shared/element-template.ts#L132
- https://github.com/concrete-utopia/utopia/blob/6b54b88ff04c85f1d6f3be63ae41cfd5b4fee565/editor/src/components/shared/project-components.ts#L296
- https://github.com/concrete-utopia/utopia/blob/6b54b88ff04c85f1d6f3be63ae41cfd5b4fee565/editor/src/core/shared/element-template.ts#L743

## Fix
Call `fixUtopiaElement` before inserting the element in `useToInsert`


gbalint (approved):

balazsbajorics (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -4744,7 +4758,14 @@ export const UPDATE_FNS = {
             let insertedElementChildren: JSXElementChildren = []
 
             insertedElementChildren.push(...action.toInsert.element.children)
-            const element = jsxElement(insertedElementName, newUID, props, insertedElementChildren)
+            const element = fixUtopiaElement(
+              jsxElement(insertedElementName, newUID, props, insertedElementChildren),
+              existingUids,
+            ).value
+
+            // const r = emptyGetAllUniqueUIDsWorkingResult()
```
balazsbajorics, line 4766: leftover logs

editor/src/components/editor/canvas-toolbar.spec.browser2.tsx
```diff
@@ -520,6 +521,77 @@ export var storyboard = (
...
+      EP.fromString(`${BakedInStoryboardUID}/${TestSceneUID}/${TestAppUID}:container`),
+    ])
+
+    await expectNoIrrecoverableErrors(editor, async () => {
+      FOR_TESTS_setNextGeneratedUids(['reserved', 'new-img'])
+      await insertViaAddElementPopup(editor, 'img')
+      await selectComponentsForTest(editor, [
+        EP.fromString(`${BakedInStoryboardUID}/${TestSceneUID}/${TestAppUID}:container`),
+      ])
+      FOR_TESTS_setNextGeneratedUids(['reserved', 'new-img-2'])
```
balazsbajorics, line 552: is this a fair test? in order to trigger the fixElement shouldn't you set this to `new-img` too? or is this `setNextGeneratedUids` here to feed the fixElement's UID generation?

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -162,6 +179,10 @@ function updateUID<T>(
             fixUIDsState.mutableAllNewUIDs,
             fixUIDsState.uidsExpectedToBeSeen,
           )
+          // console.log(`consistent uid generated for: ${oldUID}, new uid: ${uidToUse}`)
```
balazsbajorics, line 182: lots of console logs in this file!

balazsbajorics (approved):
delete the console logs :)

editor/src/components/editor/actions/actions.tsx
```diff
@@ -4701,7 +4711,11 @@ export const UPDATE_FNS = {
         newSelectedViews.push(newPath)
       }
 
-      const newUID = generateUidWithExistingComponents(editor.projectContents)
+      const existingUids = new Set(getAllUniqueUids(editor.projectContents).uniqueIDs)
+
+      const newUID = generateConsistentUID('zzz', existingUids)
```
Rheeseyb, line 4716: `'zzz'`?

editor/src/components/editor/actions/actions.tsx
bkrmendy: https://github.com/concrete-utopia/utopia/pull/4346/commits/c29d23e4c9fb9ba558a3616cc1b831e047d0154b

gbalint (approved):

editor/src/components/editor/actions/actions.tsx
bkrmendy: updated it to a more human-readable string in https://github.com/concrete-utopia/utopia/pull/4346/commits/7ca47f68755550aa1bf8b42e15ab3fb2288488ba

Rheeseyb (approved):

PR #4351, opened 2023-10-10
Only set width/height on zero-size element conversion

bkrmendy:
## Problem
When double-clicking the zero-sized element control on an element that is in a flex parent and has position: absolute set, the flexBasis prop is not taken into account, and the element stays zero-sized on one axis.

## FIx
The logic doing the conversion is refactored into two inspector strategies:
1. Convert flex child to non-zero frame
    - applied when the element is in a flex context
    - `flex`, `flexShrink`, `flexGrow` and `flexBasis` are pruned so they don't interfere with sizing
    - width: 100, height: 100 is added
2. Convert element to non-zero frame
    - width: 100, height: 100 is added
    - if the element is inline, `display: inline-block` is added


editor/src/components/canvas/controls/zero-sized-element-controls.tsx
```diff
@@ -143,14 +148,77 @@ export const ZeroSizedElementControls = controlForStrategyMemoized(
...
+        'do-not-warn',
+      ),
+    ]
+  },
+})
+
+function maybeAddDisplayInlineBlockCommands(
+  element: ElementInstanceMetadata,
+): Array<CanvasCommand> {
+  if (isIntrinsicallyInlineElement(element)) {
```
ruggi, line 187: maybe flip the logic here for an early return _or_ a ternary return

ruggi (approved):

Rheeseyb (commented):
This looks good, but please hold off on merging until https://github.com/concrete-utopia/utopia/pull/4360 is complete, as that includes a refactoring of a function `sizeToDimensionsFromFrame` that it would make sense to use here.

balazsbajorics (approved):
the new code reads _so much better_!

PR #4354, opened 2023-10-11
Setting hugging parent to fixed is optional when converting to absolute

gbalint:
**Problem:**
This is a followup PR to https://github.com/concrete-utopia/utopia/pull/4318/
The goal was to make it optional to set the size of a hugging parent to fixed.

**Fix:**
- I modified the convertToAbsoluteAndMove strategy, now it is a factory function of two different strategies: one which sets the hugging parent size to fixed and one which does not do that.
- I also added an extra parameter to the RUN_ESCAPE_HATCH strategy, so now there are two ways to invoke it.
- The fitness of the strategy which sets the hugging parent to fixed size is a little larger than the fitness of the other one, so setting the hugging parent to fixed size is still the default behavior. However, the user can manually choose the other strategy if necessary.
- As an extra I removed some redundant code and replaced the custom size fixing code with a call to `sizeToVisualDimensions`

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx
@@ -54,6 +54,7 @@ import {
 } from './fragment-like-helpers.test-utils'
 import { selectComponentsForTest } from '../../../../utils/utils.test-utils'
 import { ConvertToAbsoluteAndMoveStrategyID } from './convert-to-absolute-and-move-strategy'
+import CanvasActions from '../../canvas-actions'
 
 const complexProject = () => {
   const code = `
@@ -524,7 +525,10 @@ describe('Convert to Absolute/runEscapeHatch action', () => {
         )
 
         // CONVERT TO ABSOLUTE
-        await renderResult.dispatch([runEscapeHatch(targetsToConvert)], true)
+        await renderResult.dispatch(
+          [runEscapeHatch(targetsToConvert, 'set-hugging-parent-to-fixed')],
+          true,
+        )
 
         fastForEach(expectedAbsoluteElements, (target) => {
           const result = MetadataUtils.isPositionAbsolute(
@@ -618,7 +622,10 @@ describe('Convert to Absolute', () => {
     )
 
     const targetToConvert = EP.fromString('sb/scene/app/child')
-    await renderResult.dispatch([runEscapeHatch([targetToConvert])], true)
+    await renderResult.dispatch(
+      [runEscapeHatch([targetToConvert], 'set-hugging-parent-to-fixed')],
+      true,
+    )
 
     expect(getPrintedUiJsCode(renderResult.getEditorState())).toEqual(
       getCodeForTestProject(childTagAfter),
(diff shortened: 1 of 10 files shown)


editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx
```diff
@@ -1299,6 +1309,220 @@ describe('Convert to absolute/escape hatch', () => {
...
+        true,
+      )
+
+      await mouseUpAtPoint(canvas, offsetPoint(viewCenter, canvasPoint({ x: 15, y: 15 })))
+
+      expect(getPrintedUiJsCode(initialEditor.getEditorState())).toEqual(
+        makeTestProjectCodeWithSnippet(
+          `<View style={{ ...(props.style || {}) }} data-uid='aaa'>
+        <View
+          style={{ backgroundColor: '#aaaaaa33', width: '50%', height: '20%', right: 185, bottom: 305, top: 15, left: 15, position: 'absolute', }}
```
Rheeseyb, line 1456: This test will break, as the width and height of the target will now be fixed: https://github.com/concrete-utopia/utopia/blob/603552b7cfd199a9c07847b7729c2d85144989f8/editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx#L1174

editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx
```diff
@@ -1299,6 +1309,220 @@ describe('Convert to absolute/escape hatch', () => {
...
+        true,
+      )
+
+      await mouseUpAtPoint(canvas, offsetPoint(viewCenter, canvasPoint({ x: 15, y: 15 })))
+
+      expect(getPrintedUiJsCode(initialEditor.getEditorState())).toEqual(
+        makeTestProjectCodeWithSnippet(
+          `<View style={{ ...(props.style || {}) }} data-uid='aaa'>
+          <View
+            style={{ backgroundColor: '#aaaaaa33', width: '50%', height: '20%', right: 185, bottom: 305, top: 15, left: 15, position: 'absolute', }}
```
Rheeseyb, line 1517: Again, this test will need to be updated to use the fixed width and height values for the target element: https://github.com/concrete-utopia/utopia/blob/603552b7cfd199a9c07847b7729c2d85144989f8/editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx#L1230

editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.tsx
```diff
@@ -658,46 +701,17 @@ function createSetParentsToFixedSizeCommands(
...
-      const commands = [...setWidthCommands, ...setHeightCommands]
-      if (commands.length === 0) {
+      if (!isHuggingParent(parentElement, 'width') && !isHuggingParent(parentElement, 'height')) {
         return []
       }
+
       return [
-        ...commands,
+        ...sizeToVisualDimensions(metadata, pathTree, ancestor),
         showToastCommand('Parent is set to fixed size', 'NOTICE', 'set-parent-to-fixed-size'),
```
Rheeseyb, line 715: I suspect we don't need this toast any more if we're providing this as an explicit option

Rheeseyb (approved):

bkrmendy (approved):

PR #4357, opened 2023-10-11
pass in snap targets from the strategies

bkrmendy:
# Try it [here](https://utopia.pizza/p/5b629216-wood-newt/?branch_name=feature-resizing-per-strategy)

## Problem
The absolute resize strategy should snap to the children of the element being resized (when applicable).

## Fix
Determine the elements to snap to in the strategies, and snap based on these elements. This way, different strategies can snap to different targets. Use the new structure to snap to the parents and siblings in the move strategies, and to snap to the parents, siblings and (some) children in the resize strategies.


Rheeseyb (approved):
Very nice!

balazsbajorics (approved):

ruggi (approved):

PR #4359, opened 2023-10-11
Push intended bounds for empty elements

### LINKED ISSUE #4364: Don't shrink dynamic containers when becoming empty

ruggi:
Fixes #4364

**Problem:**

This PR extends the true up logic so that it goes beyond just group bounds.

**For reviewers:** the majority of the diff is just renames, while the actual logic changes happening in `editor/src/components/canvas/commands/push-intended-bounds-and-update-targets-command.ts`, `editor/src/components/editor/actions/actions.tsx`, and `editor/src/core/model/true-up-targets.ts` (plus the tests).

**What this does:**
1. Introduce a `PushIntendedBoundsTarget` to discriminate between different targets: currently, groups and hugging elements
2. Split the logic of the `runPushIntendedBoundsAndUpdateTargets` into multiple steps, one for each type of `PushIntendedBoundsTarget`. The new commands are triggered by `TRUE_UP_ELEMENTS` (renamed from `TRUE_UP`)
3. For groups, nothing changed
4. For hugging elements, if they are
	- empty, _and_
	- they don't cascade-delete, _and_
	- they are not fragments/groups, _and_
	- they are hugging 
→ set their x/y/w/h so that they match the last non-empty bounds.

**Notes:**
The explicit use of a `_group` type is more for backwards compatibility and in the future it could make sense to extend it to any element that can behave _like_ a group, but for now I think it's fine to have it like that so it's clear what's being changed.


editor/src/components/canvas/commands/push-intended-bounds-and-update-targets-command.ts
```diff
@@ -329,12 +489,12 @@ function getResizeAncestorGroupsCommands(
     )
   }
 
-  for (const frameAndTarget of targets) {
+  for (const v of targets) {
```
balazsbajorics, line 492: v??

editor/src/components/canvas/commands/push-intended-bounds-and-update-targets-command.ts
```diff
@@ -42,102 +45,270 @@ import {
...
-            trueUpGroupsForElementAfterDomWalkerRuns: {
-              $set: resizedGroupChildren.map((c) => trueUpElementChanged(c)),
-            },
+  let editorStatePatches: Array<EditorStatePatch> = []
+  if (commandLifecycle === 'mid-interaction') {
+    editorStatePatches.push({
+      canvas: {
+        controls: {
+          strategyIntendedBounds: {
+            $push: [...targets, ...resizeAncestorsIntendedBounds],
```
balazsbajorics, line 196: I wouldn't push the intended bounds here in the runPushIntendedBoundsAndUpdateTargetsGroup

editor/src/components/canvas/commands/push-intended-bounds-and-update-targets-command.ts
seanparsons: This.

seanparsons (approved):

balazsbajorics (changes requested):
marking it as read just in case it gets a second approve: there's some things about the approach that I'd discuss before merging, I think in general this is the good idea, but the implementation seems to mix intended bounds with the group true up too much

editor/src/components/editor/actions/actions.tsx
```diff
@@ -3764,29 +3860,29 @@ export const UPDATE_FNS = {
...
-      }
-      return {
-        frame: globalFrame,
-        target: element,
-      }
-    }, targetsToTrueUp)
-    const editorWithGroupsTruedUp = foldAndApplyCommandsSimple(editor, [
-      pushIntendedBoundsAndUpdateGroups(canvasFrameAndTargets, 'live-metadata'),
+
+    const groupTargets = targets.filter(isPushIntendedBoundsTargetGroup)
```
balazsbajorics, line 3875: is this extra step still needed? wouldn't it be better if `trueUpTargetToPushIntendedBoundsTarget` would instead return `Array<PushIntendedBoundsAndUpdateGroups | FixHuggingElements>`? is there some conceptual limtation I'm missing?

ruggi: I think returning commands here is a good refactor that improves readability, added that in [`3b83b7e` (#4359)](https://github.com/concrete-utopia/utopia/pull/4359/commits/3b83b7eb7f9b3a1db913972b97763c87ad5e9817).

balazsbajorics (approved):

editor/src/components/canvas/commands/push-intended-bounds-and-update-hugging-elements-command.ts
```diff
@@ -0,0 +1,177 @@
...
+  } else if (absoluteChildren === children.length) {
+    return 'contains-only-absolute'
+  } else if (absoluteChildren > 0) {
+    return 'contains-some-absolute'
+  } else {
+    return 'non-empty'
+  }
+}
+
+function getUpdateResizeHuggingElementsCommands(
```
Rheeseyb, line 80: This function should be renamed, because it's not just getting the commands, it is getting them and then applying them, and returning the result of that

Rheeseyb (approved):

editor/src/components/canvas/commands/push-intended-bounds-and-update-hugging-elements-command.ts
ruggi: [`6ea21c1` (#4359)](https://github.com/concrete-utopia/utopia/pull/4359/commits/6ea21c1c1b3b7bb9691ae2e869db41f01d63fa59)

PR #4360, opened 2023-10-11
Merge the core of the escape hatch commands

Rheeseyb:
**Problem:**
We have multiple versions of the "escape hatch" (for converting from flex / flow to absolute position) which were duplicating code, but all working slightly differently. This recently caused a bug where in some cases we'd fix the size of the converted element, whereas in others we wouldn't.

**Fix:**
After going back and forth on this I've come to the conclusion that some of the differences make sense (e.g. dragging a multiselection across different parts of the hierarchy to convert to absolute will first reparent all elements to a common ancestor before converting them to absolute, whereas the `x` shortcut doesn't and probably shouldn't). So the end result here was to pull out the common core from each approach (sizing the element based on its rendered frame, giving it `position: 'absolute'`, and setting the `top` and `left` based on the rendered frame), and then call that in each place.

One important change to note here - in doing this (and fixing up tests) I noticed that this was one way to add half pixel values to the size or pins. To prevent that, I've used `roundRectangleToNearestWhole` inside `getConvertIndividualElementToAbsoluteCommands` before fixing the sizing or position.


bkrmendy (approved):

editor/src/components/inspector/inspector-common.ts
```diff
@@ -1120,12 +1126,54 @@ export function toggleAbsolutePositioningCommands(
...
+  target: ElementPath,
+  jsxMetadata: ElementInstanceMetadataMap,
+  elementPathTree: ElementPathTrees,
+  frame: LocalRectangle,
+  parentFlexDirection: FlexDirection | null,
+): Array<CanvasCommand> {
+  // First round the frame so that we don't end up with half pixel values
+  const roundedFrame = roundRectangleToNearestWhole(frame)
+  return [
+    ...sizeToDimensionsFromFrame(jsxMetadata, elementPathTree, target, roundedFrame),
```
balazsbajorics, line 1176: I have a feeling that this is the point where you could clear layout props like right,bottom,flex-basis,min-width,max-width etc that could interfere with the visual layout – this is the crossover with @bkrmendy 's PR which is removing the interfering props

Rheeseyb: Yeah, that's why I commented on that PR asking that it not be merged until after this one, as that's what is already happening as part of `sizeToDimensionsFromFrame`. But it's not removing `right` and `bottom`. I'll include those 2 props in the pruning (this is also why I removed those initial layout props from a couple of tests, as those tests were providing all of the TLBRWH props!)

balazsbajorics: > as those tests were providing all of the TLBRWH props

and since the user _can_ write crazy code like that, I'd actually add those back to the initial conditions, and make sure the final result only has TLWH

Rheeseyb: Yep, that's fair. I've put them back now and included the BR pin pruning

balazsbajorics (approved):

PR #4362, opened 2023-10-11
Removing the old panel code, leaving only the Grid Panels

balazsbajorics:
**Problem:**
The 'Draggable Floating Panels' FS is now on by default, and we actually never want to turn it off. 

**Fix:**
I deleted the feature switch, and deleted all code that only ran when the FS was off. I also deleted some dead code!

**Commit Details:**
- Since the Grid panels are taking care of horizontal resizing, I removed the Resizable wrapper from the navigator, code editor and inspector
- I tweaked the wrapper div structure of the code editor so it is hopefully less likely to break and get a weird offset or half-height
- @lankaukk fixed the scrolling of the inspector
- The EditorState no longer holds the (unused) width of the code editor and the navigator
- removing a code editor resizing test
- There were a bunch of `,` commas after the panel title bars, so weird
- Tried to remove a whole bunch of `width: 100%` and `height: 100%` and replace them with flex: 1 or align-self: stretch

**Bonus fix**
The navigator root had a z-index: 1 which made it float above the Grid reorder drop target divs. WTF!!! WTF!!!!!


ruggi (approved):
🧹

enidemi (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2713,13 +2701,13 @@ export const UPDATE_FNS = {
...
-    // Expand the menu if it's not already visible.
-    if (!result.leftMenu.expanded || result.leftMenu.paneWidth <= LeftPaneMinimumWidth) {
+    // Show the menu if it's not already visible.
+    if (!result.leftMenu.expanded) {
       result = {
         ...result,
         leftMenu: {
           ...result.leftMenu,
-          paneWidth: LeftPaneDefaultWidth,
+          expanded: true,
```
Rheeseyb, line 2710: I guess `visible` would be better than `expanded`

editor/src/components/editor/editor-component.tsx
```diff
@@ -362,9 +362,8 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
           <LowPriorityStoreProvider>
             {(isChrome as boolean) ? null : <BrowserInfoBar />}
             <LoginStatusBar />
-            {unless(isFeatureEnabled('Draggable Floating Panels'), <TitleBar />)}
+            {/* TODO make sure the title bar and the navigator and the inspector are in the Low Priority Store! */}
```
Rheeseyb, line 365: Navigator and Inspector are. What do you mean by "title bar" here?

editor/src/components/editor/store/editor-state.ts
```diff
@@ -191,15 +191,7 @@ export enum LeftMenuTab {
...
-export const LeftPaneMinimumWidth = 5
-
-export const LeftPaneDefaultWidth = GridMenuWidth
-export const LeftPanelMinWidth = 240
-export const DefaultNavigatorWidth = 280
-export const LeftPanelWidthAtom = atomWithPubSub({
-  key: 'LeftPanelWidthAtom',
-  defaultValue: LeftPaneDefaultWidth,
-})
+export const DefaultNavigatorWidth = 280 // todo make it GridMenuWidth
```
Rheeseyb, line 194: Why is it not `GridMenuWidth` now?

Rheeseyb (approved):

editor/src/components/editor/editor-component.tsx
balazsbajorics: "title bar" must have been a brain fart.
```suggestion
```

editor/src/components/editor/store/editor-state.ts
balazsbajorics: I'm doing that in a separate PR #4363 because it will break a bunch of tests

PR #4376, opened 2023-10-16
Simplified layout multi-select support.

seanparsons:
**Problem:**
The new simplified layout section currently doesn't support updates for multiselection when ideally it should.

**Fix:**
Behind the scenes of the new layout section, changes have been implemented via delta updates at a high level. With single selection it's easy to take the current value and given a new value determine the delta and then feed that down into the inspector strategy. However that doesn't work as easily when handling multiselected elements with mixed values for the same property and would require a lot of that logic to be handled in the UI layer which also seems like the wrong approach.

In this PR that translation is handled by "direct" inspector strategies which layer over the regular delta based strategies, taking a particular value, calculating the delta and passing it into the logic previously handling the purely delta changes (at least for resizing).

Given the strategies were previously based on the deltas, this resulted in comparatively little changes as we've been able to reuse the logic, but it does seem a bit odd to have the direct value and then calculate the delta, which then calculates the direct value lower down.

**Commit Details:**
- Moved `changeBounds` and `ChangeBoundsResult` to `shared-absolute-resize-strategy-helpers.ts`.
- Added `resizeInspectorStrategy` and `directResizeInspectorStrategy`.
- Extracted out `getAppropriateLocalFrame` as the logic was needed/used in more than one place.
- Implemented `getDirectMoveCommandsForSelectedElement`.
- Added `startingFeatureSwitches` parameter to `renderTestEditorWithCode`.
- Changed `originalLocalFrame` to `originalLTWHValues` so that we can instead capture plurality of the positional/sizing properties.
- Added a `FrameUpdate` type to the simplified layout section to capture delta changes versus direct (where the property is set to a specific value) changes.
- `updateFrame` callback extended to support the differing types of updates by executing differing inspector strategies.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/shared-absolute-resize-strategy-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/shared-absolute-resize-strategy-helpers.ts
@@ -5,20 +5,36 @@ import type { LayoutEdgeProp, LayoutPinnedProp } from '../../../../core/layout/l
 import { framePointForPinnedProp } from '../../../../core/layout/layout-helpers-new'
 import { mapDropNulls, stripNulls } from '../../../../core/shared/array-utils'
 import { isLeft, isRight, right } from '../../../../core/shared/either'
-import type {
-  ElementInstanceMetadataMap,
-  JSXElement,
+import {
+  isJSXElement,
+  type ElementInstanceMetadataMap,
+  type JSXElement,
 } from '../../../../core/shared/element-template'
-import type { CanvasPoint, CanvasRectangle, CanvasVector } from '../../../../core/shared/math-utils'
+import { roundRectangleToNearestWhole } from '../../../../core/shared/math-utils'
+import type {
+  CanvasPoint,
+  CanvasRectangle,
+  CanvasVector,
+  LocalRectangle,
+} from '../../../../core/shared/math-utils'
 import {
   canvasPoint,
+  canvasVector,
+  isInfinityRectangle,
+  nullIfInfinity,
   pointDifference,
-  roundRectangleToNearestWhole,
+  zeroCanvasRect,
+  zeroLocalRect,
 } from '../../../../core/shared/math-utils'
 import type { ElementPath } from '../../../../core/shared/project-file-types'
 import type { AllElementProps } from '../../../editor/store/editor-state'
 import { stylePropPathMappingFn } from '../../../inspector/common/property-path-hooks'
-import { type CanvasFrameAndTarget, type EdgePosition } from '../../canvas-types'
+import {
+  EdgePositionRight,
+  type CanvasFrameAndTarget,
+  type EdgePosition,
+  EdgePositionBottom,
+} from '../../canvas-types'
 import { pickPointOnRect, snapPoint } from '../../canvas-utils'
 import type { AdjustCssLengthProperties } from '../../commands/adjust-css-length-command'
 import {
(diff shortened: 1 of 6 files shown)


editor/src/components/canvas/canvas-strategies/strategies/shared-absolute-resize-strategy-helpers.ts
```diff
@@ -274,3 +297,141 @@ function pinsFromEdgePosition(edgePosition: EdgePosition): Array<LayoutEdgeProp>
...
+      selectedElements: Array<ElementPath>,
+      _elementPathTree: ElementPathTrees,
+      _allElementProps: AllElementProps,
+    ): Array<CanvasCommand> | null => {
+      let commands: Array<CanvasCommand> = []
+      for (const selectedElement of selectedElements) {
+        const edgePosition: EdgePosition =
+          widthOrHeight === 'width' ? EdgePositionRight : EdgePositionBottom
+        const elementMetadata = MetadataUtils.findElementByElementPath(metadata, selectedElement)
+        const originalFrame = elementMetadata?.globalFrame
```
balazsbajorics, line 410: should this be the localFrame instead? are we using localFrame anywhere anymore?

editor/src/components/canvas/canvas-strategies/strategies/shared-move-strategies-helpers.ts
```diff
@@ -170,6 +171,44 @@ export function applyMoveCommon(
   }
 }
 
+function getAppropriateLocalFrame(
+  options: MoveCommandsOptions | undefined,
```
balazsbajorics, line 175: nitpick, but I'd rather just have this here: 
```suggestion
  options: { ignoreLocalFrame: boolean } | undefined,
```

editor/src/components/inspector/sections/layout-section/self-layout-subsection/frame-updating-layout-section.spec.browser2.tsx
```diff
@@ -0,0 +1,1029 @@
...
+          position: 'absolute',
+          left: '25%',
+          top: 100,
+          width: 200,
+          height: 300,
+        }}
+        data-uid={'rectangle-1'}
+      />
+    </div>`,
+    actionChange: async (renderResult) => {
```
balazsbajorics, line 215: this is nice

editor/src/components/canvas/canvas-strategies/strategies/shared-absolute-resize-strategy-helpers.ts
seanparsons: I went back and forth a little on this, IIRC the pre-existing code pushed me more in the direction of the global canvas frame.

editor/src/components/canvas/canvas-strategies/strategies/shared-move-strategies-helpers.ts
balazsbajorics: this was really just a nitpick, in general I like the idea that we rely on duck typing, to avoid turning things like MoveCommandsOptions into huge god-interfaces during a refactor.

editor/src/components/inspector/sections/layout-section/self-layout-subsection/frame-updating-layout-section.spec.browser2.tsx
```diff
@@ -0,0 +1,1029 @@
...
+import * as EP from '../../../../../core/shared/element-path'
+import { selectComponentsForTest } from '../../../../../utils/utils.test-utils'
+import { RegisteredCanvasStrategies } from '../../../../canvas/canvas-strategies/canvas-strategies'
+import { act, fireEvent } from '@testing-library/react'
+
+function downClickUp(targetElement: HTMLElement, mouseEventInit: MouseEventInit) {
+  fireEvent(targetElement, new MouseEvent('mousedown', mouseEventInit))
+  fireEvent(targetElement, new MouseEvent('click', mouseEventInit))
+  fireEvent(targetElement, new MouseEvent('mouseup', mouseEventInit))
+}
```
Rheeseyb, line 35: Please use [`mouseClickAtPoint`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/event-helpers.test-utils.tsx#L347) for this

Rheeseyb (approved):

editor/src/components/inspector/sections/layout-section/self-layout-subsection/frame-updating-layout-section.spec.browser2.tsx
seanparsons: Fixed in 01dd60249.

PR #4377, opened 2023-10-17
Fix/group children constraints section

balazsbajorics:
This PR ended up being an assortment of changes

**Details:**
- Group children are now governed by the Simplified constraint section. The dropdown contains the same options as for Frame children: "Left", "Right", "Left and Right", "Scale". Matching the existing frame-child behavior, selecting Left or Right **also** constraints the Width as well!!!
- I enabled the feature switch "Simplified Layout Section" by default, including most tests
- Some tests I kept the FS off, it'll take a bit of time to change these tests so I'd rather do them in a follow-up PR in a few days
- Since the performance tests are now running with Simplified Layout Section, I noticed a memoization regression in `FrameUpdatingLayoutSection` which I fixed 
- I also updated some tests so now they are actually testing the new section's behavior


editor/src/components/inspector/simplified-pinning-helpers.tsx
```diff
@@ -209,6 +227,77 @@ function detectPinsSet(
   return { horizontal: horizontalPins, vertical: verticalPins }
 }
 
+function detectConstraintsSetForGroupChild(
```
balazsbajorics, line 230: this is the most interesting part of the PR

bkrmendy (approved):

gbalint (approved):

PR #4384, opened 2023-10-17
Fix Code Pane Header Shrinking

lankaukk:
Preventing the Code Editor pane header from shrinking.

**Bonus Fix: adding a background to the Code Editor Pane so you can't see the canvas through the gap.

| Before  | After |
| ------------- | ------------- |
| image  | image  |


bkrmendy (approved):
🎉

gbalint (approved):

PR #4390, opened 2023-10-18
Set text element to fixed size without wrapping

bkrmendy:
# Try it [here](https://utopia.pizza/p/1c313831-hypnotic-taleggio/?branch_name=fix-non-rounded-frame)

## Problem
When converting elements from Hug to Fixed, text can be wrapped

## Cause & fix
This happens because we round bounding boxes to the nearest half, so eg 50.1 becomes 50, which in turn makes us actually use the rounded value, which wraps the text inside.

The fix is to expose a non-rounded frame, and callers can decide which one to use as a basis for measurements


editor/src/core/shared/dom-utils.ts
```diff
@@ -229,17 +229,25 @@ export function getCanvasRectangleFromElement(
...
 ): CanvasRectangle {
   const scale = canvasScale < 1 ? 1 / canvasScale : 1
 
+  const roundingFn =
+    rounding === 'nearest-half'
+      ? roundToNearestHalf
+      : rounding === 'no-rounding'
+      ? identity
+      : assertNever(rounding)
+
```
ruggi, line 242: can you rewrite this to not be a nested ternary and use a `switch` instead? (possibly inside a function I guess)

bkrmendy: makes sense, pushed it in https://github.com/concrete-utopia/utopia/pull/4390/commits/72b72d4a99c08f1ad88b4f3adb1e3952829071de

editor/src/components/inspector/inpector-selectors.tsx
```diff
@@ -64,7 +65,10 @@ export function useComputedSizeRef(prop: 'width' | 'height'): { readonly current
       return null
     }
 
-    const localFrame = MetadataUtils.getFrameOrZeroRect(elementPath, metadata)
+    const localFrame =
+      nullIfInfinity(
+        MetadataUtils.findElementByElementPath(metadata, elementPath)?.nonRoundedGlobalFrame,
```
Rheeseyb, line 70: If we're using the non-rounded frame here now, then we should rename the hook to `useNonRoundedComputedSizeRef`

Rheeseyb (approved):

ruggi (approved):

editor/src/components/inspector/inpector-selectors.tsx
bkrmendy: renamed it in https://github.com/concrete-utopia/utopia/pull/4390/commits/013797c6ac8f15fbc6fb292b6cd4228bb3bbadc7

PR #4392, opened 2023-10-18
Add flex layout to zero-sized div

bkrmendy:
## Problem
When converting a zero-sized div to a flex container, the children of the div jump to the coordinates of the zero-sized div

## Fix
Convert the zero-sized div into a full-fledged frame before adding the flex props (and the rest of the changes in `convertLayoutToFlexCommands`), as if it was a fragment

### DIFF
--- a/editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
+++ b/editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
@@ -1,26 +1,29 @@
 import { MetadataUtils } from '../../../core/model/element-metadata-utils'
-import { last, sortBy } from '../../../core/shared/array-utils'
-import { foldEither, isLeft } from '../../../core/shared/either'
+import { last, mapDropNulls, sortBy } from '../../../core/shared/array-utils'
+import { isLeft } from '../../../core/shared/either'
 import * as EP from '../../../core/shared/element-path'
 import type { ElementPathTrees } from '../../../core/shared/element-path-tree'
 import type {
   ElementInstanceMetadata,
   ElementInstanceMetadataMap,
 } from '../../../core/shared/element-template'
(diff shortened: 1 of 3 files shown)


editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
```diff
@@ -170,18 +188,61 @@ export function convertLayoutToFlexCommands(
...
+  const elementInstace = MetadataUtils.findElementByElementPath(metadata, target)
+  if (
+    elementInstace == null ||
+    isLeft(elementInstace.element) ||
+    !isJSXElementLike(elementInstace.element.value)
+  ) {
+    return []
+  }
+
+  if (type == 'fragment' || type == 'sizeless-div') {
```
balazsbajorics, line 213: is this narrowing going to be a problem? should this become a helper in `fragment-like-helpers.ts`?

gbalint, line 213: We already have a `treatElementAsFragmentLike`, should we use that here?

bkrmendy: I think the narrowing could work as long as this function is local to the convert to flex strategy (since here we only want to return actual commands if the element to be converted is a fragment or a sizeless div), but if this function would become a general helper, this would have to be thought through.

Would it work to change the name of the function to ifElementIsFragmentOrSizelessDivFirstConvertItToFrame?

bkrmendy: treatElementAsFragmentLike includes conditionals too afaik, so I wouldn't get into that in a fix PR (but we can make that work on a follow-up PR). Also, converting to flex is only allowed with DOM elements, so I'd stay consistent with that

gbalint (approved):

PR #4398, opened 2023-10-19
New kind of resize control size label

gbalint:
**Problem:**
Follow up to https://github.com/concrete-utopia/utopia/pull/4394
The resize control size label is misleading when a sizeless element is selected and the control is retargeted to its children. The size label shows the size of the child(ren), even though the element itself is zero sized.

**Fix:**
When the resize controls are retargeted to the child(ren) of a sizeless parent, add the `(Children)` prefix to the size label

image


bkrmendy (approved):

editor/src/components/inspector/fill-hug-fixed-control.tsx
```diff
@@ -85,6 +85,7 @@ export const FillFixedHugControlId = (segment: 'width' | 'height'): string =>
 
 export const FillContainerLabel = 'Fill container' as const
 export const FixedLabel = 'Fixed' as const
+export const ScaledLabel = 'Scaled' as const
```
seanparsons, line 88: This change is also in https://github.com/concrete-utopia/utopia/pull/4401/files#diff-5a58a0bbc5c3742fd34d2e6e769c411de15fb95986bda7a6baea085451067cddR88
Is that right?

seanparsons (approved):

PR #4399, opened 2023-10-19
detect center alignment when converting to flex

bkrmendy:
19-50-1tqhe-8k8uc

## Problem
When converting an absolute layout to flex, elements arranged roughly along the center line of the container aren't center-aligned in the resulting flex container.

## Fix
Add extra logic in the flex conversion strategy to cater for this

### DIFF
--- a/editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
+++ b/editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
@@ -13,10 +13,11 @@ import {
   boundingRectangleArray,
   nullIfInfinity,
   zeroCanvasRect,
+  getRectCenter,
 } from '../../../core/shared/math-utils'
 import type { ElementPath } from '../../../core/shared/project-file-types'
 import * as PP from '../../../core/shared/property-path'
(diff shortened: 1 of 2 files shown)


ruggi (approved):
🎯

editor/src/components/common/shared-strategies/convert-to-flex-strategy.ts
```diff
@@ -245,12 +247,81 @@ function ifElementIsFragmentLikeFirstConvertItToFrame(
   return []
 }
 
+const NEAR_CENTER_LINE_TOLERANCE = 0.2
```
seanparsons, line 250: Very small nitpick, but I think it makes this a tiny bit clearer:
```suggestion
const NEAR_CENTER_LINE_TOLERANCE_MULTIPLIER = 0.2
```

bkrmendy: renamed it in https://github.com/concrete-utopia/utopia/pull/4399/commits/3f6ef2de0dbe9df20f8e31d5e91f0b87623b8bfd

PR #4401, opened 2023-10-19
New FixedHugFill type: Scaled

gbalint:
**Problem:**
It is unintuitive that a width/height set to a <100% percentage is shown as 'Fixed'

**Fix:**
Create a new detected option type called 'Scaled'.


editor/src/components/inspector/inspector-common.ts
```diff
@@ -655,8 +656,12 @@ export function detectFillHugFixedState(
   const isGroupLike = treatElementAsGroupLike(metadata, elementPath)
 
   const parsed = defaultEither(null, parseCSSLengthPercent(simpleAttribute))
-  if (parsed != null && parsed.unit === '%' && parsed.value === 100) {
-    const valueWithType: FixedHugFill = { type: isGroupLike ? 'hug-group' : 'fill', value: parsed }
+  if (parsed != null && parsed.unit === '%') {
+    const fillOrScaled = parsed.value === 100 ? 'fill' : 'scaled'
+    const valueWithType: FixedHugFill = {
+      type: isGroupLike ? 'hug-group' : fillOrScaled,
+      value: parsed,
```
bkrmendy, line 663: This works but tbh it's too dense for me, could you take this apart so that the logic that decides what the type will be is a little more explicit? (ideally into a helper function)

bkrmendy (approved):

Rheeseyb (approved):

PR #4402, opened 2023-10-19
uid duplication from props

bkrmendy:
## Problem
Inserting two images from the insert pane triggers an unrecoverable error

## Cause
Same as https://github.com/concrete-utopia/utopia/pull/4346

## Fix
This PR attacks the problem from a different angle: `ComponentInfo` and `InsertableComponent` are refactored so that the element wrap is made lazy (via a function call). This, coupled with passing the attributes to the element templates in a lazy manner, makes the `jsExpressionValue` calls lazy as well, so the same element template will get new uids for its props every time it's inserted

### DIFF
--- a/editor/src/components/canvas/canvas-globals.spec.ts
+++ b/editor/src/components/canvas/canvas-globals.spec.ts
@@ -30,13 +30,20 @@ const cardComponentDescriptor: ComponentDescriptor = {
   variants: [
     {
       insertMenuLabel: 'Card Default',
-      elementToInsert: clearJSXElementWithoutUIDUniqueIDs(
-        jsxElementWithoutUID(
-          'Card',
-          [jsxAttributesEntry('title', jsExpressionValue('Default', emptyComments), emptyComments)],
-          [],
+      elementToInsert: () =>
+        clearJSXElementWithoutUIDUniqueIDs(
+          jsxElementWithoutUID(
+            'Card',
+            [
+              jsxAttributesEntry(
+                'title',
+                jsExpressionValue('Default', emptyComments),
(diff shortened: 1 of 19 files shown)


editor/src/components/canvas/canvas-globals.spec.ts
```diff
@@ -198,35 +215,7 @@ describe('validateControlsToCheck', () => {
...
-                            },
-                            "type": "ATTRIBUTE_VALUE",
-                            "uid": "",
-                            "value": "Default",
-                          },
-                        },
-                      ],
-                      "type": "JSX_ELEMENT",
-                    },
+                    "elementToInsert": [Function],
```
Rheeseyb, line 218: I'd say this part of the test is pretty useless now

editor/src/components/shared/__snapshots__/project-components.spec.ts.snap
```diff
@@ -6,17 +6,7 @@ Array [
...
-          "name": Object {
-            "baseVariable": "App",
-            "propertyPath": Object {
-              "propertyElements": Array [],
-            },
-          },
-          "props": Array [],
-          "type": "JSX_ELEMENT",
-        },
+        "element": [Function],
```
Rheeseyb, line 9: Same again with these tests

Rheeseyb (approved):
The tests that are now useless either need removing or replacing, but otherwise it looks good to me 👍

seanparsons (approved):
This is definitely the better way to solve the issue rather than munging them after the fact.

balazsbajorics (approved):

PR #4410, opened 2023-10-20
Resize near the size label

bkrmendy:
## Problem
Elements can only be resized from the bottom if the cursor is over the element (instead of slightly below it). Subjectively, it feels like the catchement area is clipped and harder to reach.

## Cause
The container of the size label seems to overlap the bottom resize control, and catch the mouse events.

## Fix
Add `pointer-events: none` to the container of the size label

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
@@ -46,6 +46,7 @@ import {
   mouseClickAtPoint,
   mouseDoubleClickAtPoint,
   mouseDownAtPoint,
+  mouseDragFromPointToPoint,
   mouseDragFromPointWithDelta,
   mouseMoveToPoint,
 } from '../../event-helpers.test-utils'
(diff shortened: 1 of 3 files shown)


ruggi (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
```diff
@@ -2943,6 +2946,58 @@ describe('Absolute Resize Strategy Canvas Controls', () => {
...
+      x: resizeCornerBounds.x + resizeCornerBounds.width / 2,
+      y: resizeCornerBounds.y + resizeCornerBounds.height - 1,
+    })
+
+    const endPoint = {
+      x: startPoint.x + dragDelta.x,
+      y: startPoint.y + dragDelta.y,
+    }
+
+    await mouseDragFromPointToPoint(canvasControl, startPoint, endPoint, { realMouseDown: true })
```
balazsbajorics, line 2985: does this test break without your actual fix?

balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.spec.browser2.tsx
bkrmendy: yes, it does break

PR #4412, opened 2023-10-20
Gap controls positioned to be consistent with element positions.

seanparsons:
**Problem:**
With pretty much anything other than elements that are full width/height in a flex container, the gap controls are invariably in an odd place like over on the left hand side when the content is aligned to the centre:
image

**Cause:**
The handles were centred in a div which was only _sized_ to that of the bounds of the children, so in the case of centre aligned elements the handles would be centred inside the div that was positioned on left at position 0 (for instance).

**Fix:**
By positioning the containing div for the handles against the accumulated boundary of the children, the handles end up centred against those positions instead, which means they are now in the right place:
image

**Commit Details:**
- `contentArea` in `FlexGapControl` is now a `CanvasRectangle` so that the position of the children as a whole can be utilised.
- `GapControlSegment` now uses the newly provided fields in `contentArea` to position the containing div so that the handles centered within it are centered against the bounds of the content.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/set-flex-gap-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/set-flex-gap-strategy.spec.browser2.tsx
@@ -1,3 +1,4 @@
+/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "checkFlexGapHandlesPositionedCorrectly"] }] */
 import { CanvasControlsContainerID } from '../../controls/new-canvas-controls'
 import {
   FlexGapControlHandleTestId,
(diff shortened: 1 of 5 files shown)


ruggi (approved):

Rheeseyb (approved):

bkrmendy (changes requested):
The controls are not positioned correctly when the element has padding (see this project: https://utopia.pizza/p/21453729-habitual-strawflower/?branch_name=fix-gap-control-positioning), also the PR is missing tests

bkrmendy (approved):

ruggi (approved):

balazsbajorics (approved):

Rheeseyb (approved):

PR #4413, opened 2023-10-20
Detect hug from css and add different types of hugs: squeeze, collapsed

gbalint:
**Problem:**
The Hug property should work well even when the style properties are derived from css.
Hug type should be split into multiple subcases:
- `hug` is `max-content`
- `squeeze` is `min-content`
- `collapsed` when the resulting size is 0
Squeeze and Collapsed can only be detected but not selected in the dropdown.

**Fix:**
- new property in `SpecialSizeMeasurements`: `computedHugProperty`. This contains a `width` and a `height` key.
- new function to derive the hug property from the style props: `hugPropertiesFromStyleMap`
- `hugPropertiesFromStyleMap` is called from the dom-walker using the browser API `element.computedStyleMap`, which returns style data coming from css too, and we fill in `SpecialSizeMeasurements.computedHugProperty` from that.
- In places where we tried to find out if something is hugging or not we just check it in `SpecialSizeMeasurements.computedHugProperty`
- In `detectFillHugFixedState` we call `hugPropertiesFromStyleMap` using the local style props too, to see if the hug property coming from the local style props is the same as the real one. If not, we set the control status to `detected-fromcss`

**Missing things:**
- `hugPropertiesFromStyleMap` is not final, it could be improved, e.g. `fit-content` has a  logic (sometimes is a `hug`, sometimes it is a `squeeze`) which is not implemented yet.
For full coverage e.g. this doc could be reviewed: https://www.w3.org/TR/css-position-3/#abs-non-replaced-width
- Missing tests for control statuses (one potential solution is to somehow propagate the info down through the PopupList so we can add a `data-controlstatus` to the label)
- I only update how we detect the different hug types, but not how we detec the other fixed/fill/detected etc types. That can happen in a followup PR. And I guess we have _tons_ of places where we should use the computedStyleMap and not just the style props themselves.


editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx
```diff
@@ -1207,137 +1214,15 @@ describe('Convert to absolute/escape hatch', () => {
       )
     })
 
-    it('works on a flow element with all pins and sets hugging parent to fixed size', async () => {
```
gbalint, line 1210: This test was wrong! This is not a hugging parent, so no need to set it fixed size

editor/src/components/canvas/canvas-strategies/strategies/convert-to-absolute-and-move-strategy.spec.browser2.tsx
```diff
@@ -1410,7 +1288,14 @@ describe('Convert to absolute/escape hatch', () => {
     it('Runs the escape hatch strategy and does not set hugging parent to fixed size', async () => {
       const initialEditor = await renderTestEditorWithCode(
         makeTestProjectCodeWithSnippet(`
-            <View style={{ ...(props.style || {}) }} data-uid='aaa'>
```
balazsbajorics, line 1413: so what is the new behavior for all these "width: auto; height: auto" elements?

editor/src/core/shared/element-template.ts
```diff
@@ -2018,6 +2018,11 @@ export function elementInstanceMetadata(
 export type DetectedLayoutSystem = 'flex' | 'grid' | 'flow' | 'none'
 export type SettableLayoutSystem = 'flex' | 'flow' | 'grid' | LayoutSystem
 export type TextDirection = 'ltr' | 'rtl'
+export type HugProperty = 'hug' | 'squeeze' | 'collapsed'
```
balazsbajorics, line 2021: I wonder if we should explicitly have a 'not-hugging' here instead of null doing double-duty

seanparsons (approved):

balazsbajorics (approved):

PR #4415, opened 2023-10-24
mini PR: update inspector group tests

balazsbajorics:
**Problem:**
Some group tests for the inspector were not switched over to the new layout section.

**Fix:**
Do that.

### DIFF
--- a/editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
+++ b/editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
@@ -3243,7 +3243,7 @@ describe('inspector tests with real metadata', () => {
       })
 
       const leftControl = (await renderResult.renderedDOM.findByTestId(
-        'position-left-number-input',
+        'frame-left-number-input',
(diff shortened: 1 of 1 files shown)


bkrmendy (approved):

editor/src/components/inspector/common/inspector-end-to-end-tests.spec.browser2.tsx
```diff
@@ -3256,10 +3250,10 @@ describe('inspector tests with real metadata', () => {
 
       const elementFrame = getFrame(targetPath, renderResult)
 
-      await setControlValue('position-left-number-input', '', renderResult.renderedDOM)
+      await setControlValue('frame-left-number-input', '', renderResult.renderedDOM)
 
+      // Nothing changed
       expect(getFrame(targetPath, renderResult)).toBe(elementFrame)
```
balazsbajorics, line 3256: I kept these tests around to ensure that nothing changed is the result, but the bigger context is that the number control doesn't allow you to enter empty value or % value anymore, so there's no need to fire a toast.

ruggi (approved):

PR #4419, opened 2023-10-24
Hide constraints section if empty

### LINKED ISSUE #4418: Hide constraints inspector section if empty

ruggi:
Fixes #4418 

**Problem:**

The `Constraints` section is shown in the inspector even when it's effectively empty.

Screenshot 2023-10-24 at 4 08 25 PM

**Fix:**

Don't show it if that's the case.

### DIFF
--- a/editor/src/components/inspector/constraints-section.spec.browser2.tsx
+++ b/editor/src/components/inspector/constraints-section.spec.browser2.tsx
@@ -211,9 +211,10 @@ describe('Constraints Section', () => {
     )
   })
 
-  it('is hidden when the selection does not contain elements for which pins are applicable', async () => {
-    const renderResult = await renderTestEditorWithCode(
-      formatTestProjectCode(`
+  describe('hidden behaviour', () => {
+    it('is hidden when the selection is made of groups', async () => {
+      const renderResult = await renderTestEditorWithCode(
+        formatTestProjectCode(`
 		  import * as React from 'react'
 		  import { Group, Storyboard } from 'utopia-api'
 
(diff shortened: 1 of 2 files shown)


bkrmendy (approved):

editor/src/components/inspector/constraints-section.tsx
```diff
@@ -51,10 +52,19 @@ export const ConstraintsSection = React.memo(() => {
     'ConstraintsSection onlyGroupChildrenSelected',
   )
 
+  const showSection = React.useMemo(() => {
+    return noGroupOrGroupChildrenSelected || onlyGroupChildrenSelected
+  }, [noGroupOrGroupChildrenSelected, onlyGroupChildrenSelected])
```
Rheeseyb, line 57: With this we should also remove the `when`s below

editor/src/components/inspector/constraints-section.spec.browser2.tsx
```diff
@@ -207,4 +210,30 @@ describe('Constraints Section', () => {
       }),
     )
   })
+
+  it('is hidden when there are no contents', async () => {
```
Rheeseyb, line 214: I know this is going to seem a bit pedantic, but I think this (and the PR title and description) are a bit misleading here. The reason that the contents was empty was because of the types of the selected elements, so now we're just not showing the entire section based on that (rather than explicitly checking if the section is empty and hiding it if so). What I mean by that is we could later end up in a situation where the constraints section is empty but still shown. So really this test should be `'is hidden when _ elements are selected'`.

editor/src/components/inspector/constraints-section.tsx
ruggi: the two `when`s are not necessarily true at the same time though, I think they should still remain (one can be true while the other is false)

editor/src/components/inspector/constraints-section.spec.browser2.tsx
ruggi: It's true, the PR is hiding the constraints section if there are no displayable portions of the section. The reason is partly because of how the e.g. pins are detected (there's a bunch of "ifs" that need to be put in place to make sure the pins mean "empty" - if that's even a case). I probably should rename the related code to something like what you suggested ^

editor/src/components/inspector/constraints-section.tsx
Rheeseyb: Sorry, my bad. Disregard that!

Rheeseyb (approved):

PR #4425, opened 2023-10-25
Fix: new layout section number controls can now work with %, add missing props, can override expressions

balazsbajorics:
**Problem:**
I was working down my test debt of converting tests that still ran with the old layout section to instead run using the new layout section. 

As part of this I unearthed a bunch of missing behavior and some bugs:

#### Remove use of adjustCssLengthProperties
Turns out adjustCssLengthProperties is just the worst, and now pretty much obsolete. Originally it was made in a world where _some_ strategies didn't calculate the expected globalFrame / localFrame of the manipulated element, so it made sense to have a "delta" strategy that can just increase/decrease a value without a larger context. There's two big problems with this: 
1. we now calculate the expected localFrame/globalFrame for everything anyways, because of Groups and also for the snapping bounds. 
2. if a value was missing, adjustCssLengthProperties just assumed it was 0 and put the delta in as the new value. Consider why this is wrong: I have an element that has the following frame properties: left, width. Now I want to set `right` but it did not exist before: we can't just assume `right` was 0 just because it's implicit! We already know what we want `right` to be because we already calculated the expected globalFrame and we know the parentFrame too, _so the entire existence of this strategy is moot*!_
So in the resize commands I replaced adjustCssLengthProperties with SetCssLengthProperty - with the exception of the _relative_ move strategy, which I decided should be out of scope for this PR as it was already getting large.

#### SetCssLengthProperty fires a toast when replacing an expression with a simple value
- The behavior on master was to treat expressions as untouchable, but the code had a "TODO Design Review" next to it – the new behavior is to replace the expressions with simple values + fire a toast.
- SetCssLengthProperty is also now rounding percentage values to 2 decimal places

#### Had to fix some unfortunate tests that are still using fake metadata
They are still using fake metadata, I just fixed the test values :( one day we should migrate them though but it's just a big chore. The tests themselves are super valuable though so it's definitely worth keeping them!

#### I fixed the fake metadata we create for a freshly inserted element.
We insert it on the canvas root, which we know has an _actual_ size of 0x0, so that's what I'm giving to these freshly inserted elements too. IT works!

#### Delete absolute-reparent-strategy-only-move.spec.tsx which is a fake metadata test
`absolute-move-strategy.spec.browser2.tsx` covers these cases anyways and has been updated to be a real metadata test


ruggi (approved):

editor/src/components/canvas/canvas-strategies/strategies/shared-move-strategies-helpers.ts
```diff
@@ -334,12 +364,11 @@ export function directMoveInspectorStrategy(
   }
 }
 
-export function createMoveCommandsForElement(
+export function createMoveCommandsForElementPositionRelative(
```
Rheeseyb, line 367: `PositionRelative`?

Rheeseyb (approved):

PR #4426, opened 2023-10-25
Convert flex layout special case 3 elements with constraints.

seanparsons:
**Problem:**
For a specific set of circumstances of 3 elements with the following sets of constraints from left to right:
- `[left, width]`
- `[left, right]`
- `[right, width]`

Which also do not overlap, we want to be preserve the layout as much as reasonably possible while still converting to a flex layout.

**Fix:**
In `convertLayoutToFlexCommands` an additional special case has been added which initially checks if there are 3 children of the selected element. Then because they're in a group and the visual positioning does not necessarily line up with the logical ordering, we sort them visually to determine the "first", "second" and "third" elements. The constraints are validated along with checking that there's no overlap.

Once the validation is all successful, then a `flexGrow` value is back-calculated for the "second" element so that it ends up horizontally positioned as close to the original starting position as possible. Along with values like the absolute positioning properties being expunged and the entries being reordered to make sense in a flex context.

**Commit Details:**
- Added `doRectanglesIntersect` and `anyRectanglesIntersect` utility functions.
- Renamed `AddRemoveLayouSystemControl` to `AddRemoveLayoutSystemControl`.
- Renamed `AddRemoveLayouSystemControlTestId` to `AddRemoveLayoutSystemControlTestId`.
- Added `checkConstraintsForThreeElementRow`, `anyOverlappingFrames` and `laidOutAsRow` utility functions to directly support the checks needed.
- Implemented special case to `convertLayoutToFlexCommands` for a group of 3 constrained elements laid out in a row.

### DIFF
--- a/editor/src/components/common/shared-strategies/convert-to-flex-strategy.spec.browser2.tsx
+++ b/editor/src/components/common/shared-strategies/convert-to-flex-strategy.spec.browser2.tsx
@@ -599,7 +599,7 @@ describe('Smart Convert To Flex', () => {
               top: 29,
               width: 'max-content',
             }}
-            data-constraints={['left']}
+            data-constraints={['left', 'right']}
           >
             Hello World
           </span>
@@ -642,7 +642,7 @@ describe('Smart Convert To Flex', () => {
     expect(getPrintedUiJsCode(editor.getEditorState())).toEqual(
       makeTestProjectCodeWithSnippet(`
 		<div style={{ ...props.style }} data-uid='a'>
-      <Group
+      <div
         data-uid='parent'
         style={{
           height: 204,
@@ -652,7 +652,7 @@ describe('Smart Convert To Flex', () => {
           width: 584,
           display: 'flex',
           flexDirection: 'row',
-          justifyContent: 'space-between',
+          gap: 32,
         }}
       >
         <div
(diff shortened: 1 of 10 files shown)


editor/src/components/common/shared-strategies/convert-to-flex-strategy.spec.browser2.tsx
```diff
@@ -572,6 +572,129 @@ describe('Smart Convert To Flex', () => {
...
+
+      'await-first-dom-report',
+    )
+
+    await convertParentToFlex(editor)
+
+    expect(getPrintedUiJsCode(editor.getEditorState())).toEqual(
+      makeTestProjectCodeWithSnippet(`
+		<div style={{ ...props.style }} data-uid='a'>
+      <Group
```
balazsbajorics, line 645: looks like there is a missing Group -> div conversion step! (probably related that the data-constraints are left on the children too)

bkrmendy (approved):

editor/src/components/common/shared-strategies/convert-to-flex-strategy.spec.browser2.tsx
seanparsons: Ah, I wasn't sure if we wanted this because it came up previously. But it certainly makes sense in this narrow case to make that change.

seanparsons: Should be fixed by e08982bb4b318b31fe5221f15fdbc35691bd20eb.

balazsbajorics (approved):

PR #4427, opened 2023-10-25
Wrap in non-zero sized div

bkrmendy:
## Problem
Wrapping in div inserts a zero-sized div

## Fix
Create a full-featured wrapping function for wrapping into divs, which wraps the selected elements in a div whose AABB matches the AABB of the selected elements

## Details
- A new component group is created specifically for divs (so they are easier to special case)

### DIFF
--- a/editor/src/components/editor/wrap-in-callbacks.ts
+++ b/editor/src/components/editor/wrap-in-callbacks.ts
@@ -48,7 +48,7 @@ type WrapInDivError =
   | 'Cannot determine the bounding box of selected elements'
   | 'Cannot insert into parent of selected elements'
 
-function wrapInDivCommands(
+export function wrapInDivCommands(
   metadata: ElementInstanceMetadataMap,
   elementPathTrees: ElementPathTrees,
   allElementProps: AllElementProps,
--- a/editor/src/components/editor/wrap-in-callbacks.ts
+++ b/editor/src/components/editor/wrap-in-callbacks.ts
@@ -69,10 +69,12 @@ export function wrapInDivCommands(
     return left('Cannot determine the bounding box of selected elements')
   }
 
-  const wrapperUid = generateConsistentUID(
-    'wrapper',
-    new Set(getAllUniqueUids(projectContents).allIDs),
-  )
+  const allIds = new Set(getAllUniqueUids(projectContents).allIDs)
+
+  const wrapperUid = generateConsistentUID('wrapper', allIds)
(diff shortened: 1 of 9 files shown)


editor/src/components/editor/wrap-in-callbacks.ts
```diff
@@ -1,33 +1,227 @@
...
+      editorRef.current.allElementProps,
+      editorRef.current.projectContents,
+      editorRef.current.selectedViews,
+    )
+
+    if (isLeft(result)) {
+      return
+    }
+
+    dispatch([applyCommandsAction(result.value)])
```
balazsbajorics, line 223: maybe nitpicking, but this code is _almost_ structured like an inspector strategy, it might as well be one, with a `executeFirstApplicableStrategy` call instead of this `applyCommandsAction`

bkrmendy: The reason I didn't use `executeFirstApplicableStrategy` was because `wrapInDivCommands` takes a `projectContents` param, and for that `executeFirstApplicableStrategy`'s signature would have had to be changed too, and I wanted to keep the PR scoped to wrapping. We can make this change if you think it's important though

balazsbajorics: image

most new inspector strategies take the extra params in their factory function, you could add projectContents there!

--------------------------

BTW this two sources of params actually look pretty confusing inside the strategy code:
image

For the Canvas strategies we already moved to a setup where the `apply()` function is almost parameterless. I was thinking that we could do the same for inspector strategies, the `strategy()` function could have zero parameters, and only pass in the params to the strategy's factory function. 

The end result would be something like this:
```
executeFirstApplicableStrategy(dispatch, [
  moveInspectorStrategy(
    projectContentsRef.current,
    metadataRef.current,
    selectedViewsRef.current,
    frameUpdate.edgeMovement,
  ),
  resizeInspectorStrategy(
    projectContentsRef.current,
    metadataRef.current,
    selectedViewsRef.current,
    frameUpdate.edgeMovement,
  ),
])
```

bkrmendy: implemented in https://github.com/concrete-utopia/utopia/pull/4427/commits/041d4294a4a206966c8b7fe2f481fcc7fe79206b

balazsbajorics (approved):

ruggi (approved):

PR #4429, opened 2023-10-26
Check whether component supports children for wrapping

bkrmendy:
## Problem
In the Wrap menu, we offer to wrap in elements that don't actually support children.

## Fix
The fix consist of two parts:
- Add a `supportsChildren` prop to `ComponentDescriptor`, supply this `supportsChildren` prop wherever possible, and filter the component list in the popup based on that.
- If there's no available `ComponentDescriptor`, this PR adds a best-effort check that tries to look up a top-level element with the same name as the one from the export, and checks whether that element uses the children prop


ruggi (approved):

editor/src/components/shared/project-components.ts
```diff
@@ -283,12 +293,14 @@ function makeHTMLDescriptor(
   extraPropertyControls: PropertyControls,
   attributes?: () => JSXAttributes,
 ): ComponentDescriptor {
+  const supportsChildren = intrinsicHTMLElementNamesThatSupportChildren.includes(tag) // TODO: maybe circ dep
```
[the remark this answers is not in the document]
bkrmendy: I removed the comment in https://github.com/concrete-utopia/utopia/pull/4429/commits/9dccb4aff90c77b6b3c110dbb387f5a32cad18f7, it seems not to cause any problems (famous last words)

seanparsons (approved):

editor/src/core/shared/project-file-types.ts
```diff
@@ -761,3 +765,14 @@ export type NodeModules = {
...
+export function getTopLevelElementByExportsDetail(
+  file: ParseSuccess,
+  nameToLookFor: string,
+): UtopiaJSXComponent | null {
+  return (
+    file.topLevelElements.find(
+      (t): t is UtopiaJSXComponent => t.type === 'UTOPIA_JSX_COMPONENT' && t.name === nameToLookFor,
+    ) ?? null
+  )
+}
```
Rheeseyb, line 778: Can you move this to `/editor/src/core/model/project-file-utils.ts` please, as we have a few functions close to this in there

editor/src/core/third-party/remix-controls.ts
```diff
@@ -12,6 +12,7 @@ export const RemixRunReactComponents: ComponentDescriptorsForFile = {
     properties: {
       to: { control: 'string-input' },
     },
+    supportsChildren: false,
```
Rheeseyb, line 15:
```suggestion
    supportsChildren: true,
```

Rheeseyb (approved):

editor/src/core/shared/project-file-types.ts
bkrmendy: moved it in https://github.com/concrete-utopia/utopia/pull/4429/commits/71c5121c27040496610686db54e94c105d6c59cb

PR #4438, opened 2023-10-27
Factor out `wrapInElement` calls from tests

bkrmendy:
## Description
This PR factors out calls to the `wrapInElement` factory function from the browser tests, so that they don't assume that this exact action is used for wrapping (since the action will be replaced with a commands/strategies-based implementation by https://github.com/concrete-utopia/utopia/pull/4435).

## Details
- `wrap in div works inside a conditional on an expression` was x-ed, because it's an extreme edge case of wrapping, and fixing it would have bloated the PR

### DIFF
--- a/editor/src/utils/utils.test-utils.tsx
+++ b/editor/src/utils/utils.test-utils.tsx
@@ -55,7 +55,7 @@ import {
 } from '../core/shared/project-file-types'
 import { foldEither, right } from '../core/shared/either'
 import Utils from './utils'
-import type { SimpleRectangle, WindowRectangle } from '../core/shared/math-utils'
+import type { SimpleRectangle } from '../core/shared/math-utils'
 import { canvasRectangle, localRectangle, negate, offsetRect } from '../core/shared/math-utils'
 import {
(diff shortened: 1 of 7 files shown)


editor/src/utils/utils.test-utils.tsx
```diff
@@ -575,3 +575,15 @@ export function boundingClientRectToCanvasRectangle(
...
 }
+
+export async function searchInFloatingMenu(editor: EditorRenderResult, query: string) {
+  const floatingMenu = editor.renderedDOM.getByTestId(CanvasToolbarSearchTestID)
+  const searchBox = queryByAttribute('type', floatingMenu, 'text')!
+
+  await act(() => {
+    fireEvent.focus(searchBox)
+    fireEvent.change(searchBox, { target: { value: query } })
+    fireEvent.blur(searchBox)
```
seanparsons, line 586: I'm not sure if it makes any difference here, but I'd normally expect `blur` to trigger after `keydown`.

seanparsons (approved):

PR #4439, opened 2023-10-27
Move width/height buttons inside constraint control.

seanparsons:
**Problem:**
We want the width and height controls to be integrated with the regular frame controls for constraints and frame children.

**Fix:**
Created a new control `CombinedPinControl` which wraps up the regular frame controls with the width and height buttons.

**Commit Details:**
- `PinControl` now has a `regularBorder` property, so that it's possible to turn off the border.
- `PinWidthControl` now attaches `props.toggleWidth` to the `onClick` handler.
- `PinHeightControl` now attaches `props.toggleHeight` to the `onClick` handler.
- Created `CombinedPinControl` that wraps up the different cross like controls with `PinWidthControl` and `PinHeightControl`.
- Slimmed down `GroupChildConstraintsSection` and `FrameChildConstraintsSection` by using `CombinedPinControl`.


balazsbajorics (approved):

editor/src/components/inspector/controls/pin-control.spec.browser2.tsx
```diff
@@ -0,0 +1,660 @@
...
+  return formatTestProjectCode(code)
+}
+
+interface TestCase {
+  baseProject: string
+  actionChange: (renderResult: EditorRenderResult) => Promise<void>
+  expectedProject: string
+}
+
+xdescribe('Group child constraints', () => {
```
Rheeseyb, line 89: Was this meant to be left as `xdescribe`?

seanparsons: It was an intentional change from this commit: https://github.com/concrete-utopia/utopia/pull/4439/commits/e5260e2be7e9dd6413f1df3e1962c3ffa2ba77dd

balazsbajorics: yes I changed it to xdescribe along with replacing the actual pin click handler with a NO_OP. there's a detailed follow-up task in the paper doc how to re-enable the actual toggling behaviors, which are different for frame children, group children etc. 

on master the pin toggling is disabled because for groups it was broken.

this PR accidentally resurrected the old-world usePinToggling which is buggy so we decided to keep "the behavior" from master: disable them for now

Rheeseyb (approved):

PR #4450, opened 2023-10-31
parse appDirectory

bkrmendy:
## Problem
We look for route modules in the `/src` directory no matter what. This is problematic for two reasons:
- it's not the default (`/app` is, according to Remix docs: https://remix.run/docs/en/main/file-conventions/remix-config)
- hence, most Remix projects cannot be loaded

## Fix
Try and get the default app directory from `remix.config.js`, failing that, use `/app` (the right default)


editor/src/components/canvas/remix/remix-navigator.spec.browser2.tsx
```diff
@@ -47,7 +47,7 @@ describe('Remix navigator', () => {
         </Storyboard>
       )
       `,
-      ['/src/root.js']: `import React from 'react'
+      ['/app/root.js']: `import React from 'react'
```
balazsbajorics, line 50: is it true that in this test you could keep the test routes `/src` and include a `remix.config.js` among the test project files?

bkrmendy: it's true, and I included such a test (called `Parses different route dir`, which had to be fixed in https://github.com/concrete-utopia/utopia/pull/4450/commits/063843a046f927a9d8ed27a9a075b47b86fd70cf).

I renamed `src` to `app` in the existing tests because including the remix config file would have been basically the same diff, but with more context, and using `app` by default reflects that `app` is the default folder

editor/src/components/canvas/remix/remix-utils.spec.browser2.tsx
```diff
@@ -28,6 +31,22 @@ export var storyboard = (
...
+  appDirectory: "src",
+  assetsBuildDirectory: "public/build",
+  ignoredRouteFiles: ["**/.*"],
+  publicPath: "/build/",
+  routes(defineRoutes) {
+    return defineRoutes((route) => {
+      route("/somewhere/cool/*", "catchall.tsx");
+    });
+  },
+  serverBuildPath: "build/index.js",
```
Rheeseyb, line 46: I'd say that whilst we're only parsing the `appDirectory` we should only include that in this test, otherwise there's a good chance that this sample config will drift from a valid config in a way that we wouldn't catch.

bkrmendy: great point, fixed in https://github.com/concrete-utopia/utopia/pull/4450/commits/79e549a060bdb3687c2cd2b699e3198a52801ac8

Rheeseyb (approved):

editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -344,7 +345,7 @@ export function getRoutesAndModulesFromManifest(
   const routeModuleCreators: RouteIdsToModuleCreators = {}
   const routingTable: RemixRoutingTable = {}
 
-  const rootJsFile = getProjectFileByFilePath(projectContents, `${ROOT_DIR}/root.js`)
+  const rootJsFile = getProjectFileByFilePath(projectContents, `${rootDir}/root.js`)
```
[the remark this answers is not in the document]
bkrmendy: I'll put this on a separate PR, I tried quickly pathing it, but (unfortunately) it's hardcoded in more than one place

balazsbajorics (approved):

PR #4453, opened 2023-10-31
root.jsx

bkrmendy:
## Problem
`root.js` is hardcoded as the root file name for Remix projects, even though `root.jsx` is valid as well.

## Fix 
Add `root.jsx` as a valid root file too.

## Details
`flatRoutes` also looks for the root file, so this logic is somewhat duplicated. However, we need the path of the root file for `patchRemixRoutes`, so this PR keeps the duplication


editor/src/components/canvas/remix/remix-utils.tsx
```diff
@@ -397,16 +419,20 @@ export function getRoutesAndModulesFromManifest(
...
 
 export function getRouteComponentNameForOutlet(
-  path: ElementPath,
+  elementPath: ElementPath,
   metadata: ElementInstanceMetadataMap,
   projectContents: ProjectContentTreeRoot,
   pathTrees: ElementPathTrees,
 ): string | null {
-  if (!MetadataUtils.isProbablyRemixOutlet(metadata, path)) {
+  if (!MetadataUtils.isProbablyRemixOutlet(metadata, elementPath)) {
```
bkrmendy, line 427: had to be renamed because `path` got imported

seanparsons (approved):

Rheeseyb (approved):

PR #4463, opened 2023-11-01
Remove old layout section

balazsbajorics:
Now that finally all layout related inspector tests are merged to use the new SimplifiedLayoutSection, it was finally time to remove the feature switch 'Simplified Layout Section' and all the code that now became unused, including some tests too. This is removing some 4-5 years old code!

**Note**
There's always a possibility that I've overdone it and we need to bring back some of this stuff from git history, but I say it's better deleted than left around unused.


Rheeseyb (approved):
🔪

bkrmendy (approved):

seanparsons (approved):

PR #4466, opened 2023-11-02
parse jsx attribute with object literal

bkrmendy:
# [Try it here](https://utopia.pizza/p/697deae1-piquant-litter/?branch_name=fix-jsx-attribute-with-object-literal)

## Problem
When parsing JS code, we don't add object literal initializers to `definedElsewhere` (see the example project above for an illustration).

## Fix
Cater for this in the parser-printer code. This PR add support for "normal" object initialization (e.g., `{ t: text }`) and shorthand initializers, e.g.
```typescript
const t = "hello there"
return <span>{greeting({ t })</span>
```


seanparsons (approved):
Looks good to me!

Rheeseyb (approved):

PR #4470, opened 2023-11-03
Fix flaky Jest test

balazsbajorics:
**Problem:**
Somehow there's one Jest test that breaks every now and then because it runs the highlight code which relies on document.elementsFromPoint and that doesn't exist in JSDOM and so an error is thrown.

**Fix:**
I decided to go with the most barebones of fixes: `useHighlightCallbacks` should be disabled if document.elementFromPoint is undefined.

**Commit Details:**
- `useHighlightCallbacks` should be disabled if document.elementFromPoint is undefined.
- I unified the two Remix Error Handling test files because they were only separate to avoid a test breakage


ruggi (approved):

bkrmendy (approved):

editor/src/components/canvas/remix/remix-error-handling.spec.tsx
```diff
@@ -1,14 +1,21 @@
...
-// as otherwise it causes issues with Jest for some reason
-
 describe('Remix error handling', () => {
+  it('Supports a user-defined error boundary', async () => {
+    const renderResult = await renderTestEditorWithCode(
+      makeTestProjectCodeWithSnippet(`<div>Helo</div>`),
+      'await-first-dom-report',
+    )
+
+    expect(renderResult.renderedDOM.getByText('Helo')).toBeDefined()
```
Rheeseyb, line 11: What's going on with this test? It's not at all the original test...

balazsbajorics: OH HAHAHA JESUS

Rheeseyb (approved):

PR #4471, opened 2023-11-03
Make the SVG loader safe.

seanparsons:
**Problem:**
Currently if a SVG is imported to a variable it will cause an exception that breaks the canvas.

**Cause:**
When a SVG file is added to the project as a `ASSET_FILE` instance, the editor does not hold a copy of the contents by default. As a result an empty string is passed to the loader as the contents which then immediately throws an exception as that is not valid SVG content.

**Investigation:**
The empty string comes from `getProjectFileContentsAsString`, which is problematic on its own as it works from the assumption that content is available locally and for `IMAGE_FILE` and `ASSET_FILE` instances will always return an empty string. As this is core to the loader infrastructure this needs a bit more thought than a patch fix.

While trying to use multiple placeholder SVGs in the presence of an empty string it was discovered that none of them would work with the existing logic for converting a SVG to a component. They would all fail with a SVG parser error.

A brief bit of work was done to try and create an alternative for converting a SVG to a component, but this started to become more involved and seemed like it could turn into a time sink.

**Fix:**
This makes the frontend consistent with the Github loading by now treating any SVG as a `ASSET_FILE` when added to the editor.

For the moment this "disables" the `SVGLoader` such that it produces a component that returns null.

**Commit Details:**
- `fileTypeFromFileName` now consistent with the Github loading by treating SVGs as `ASSET_FILE` instances.
- Effectively disable the `SVGLoader` so that it does not throw an exception whenever it is triggered.

### DIFF
--- a/editor/src/core/es-modules/package-manager/package-manager.spec.ts
+++ b/editor/src/core/es-modules/package-manager/package-manager.spec.ts
@@ -131,7 +131,7 @@ describe('ES Dependency Package Manager', () => {
     expect(document.getElementById('/node_modules/mypackage/simple.css')).toBeNull()
   })
 
-  it('resolves a svg import', () => {
+  xit('resolves a svg import', () => {
     const reqFn = getRequireFn(
       NO_OP,
       {},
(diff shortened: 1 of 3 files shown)


seanparsons:
Closing in favour of a better solution.

PR #4474, opened 2023-11-03
Remove SVG loader.

seanparsons:
**Problem:**
Importing any SVG in a project that has been imported from Github results in a failure breaking the canvas.

**Cause:**
As the SVG has been treated as a `ASSET_FILE` instead of a `TEXT_FILE` the contents are no longer held within the project definition. So the content of the SVG doesn't become available to the loaders and then an empty string gets passed to the `SVGLoader` which then promptly explodes.

**Context:**
We now longer care for being able to import a SVG as a component, instead it should present as a URL as it would for a `.png` or `.jpg`.

**Fix:**
Primarily this completely removes the `SVGLoader` so that `FileLoader` is used in its place.

Additionally any SVGs now added to the project via drag and drop are now are uploaded as an `ASSET_FILE` for consistency.

**Commit Details:**
- `fileTypeFromFileName` now consistent with the Github loading by treating SVGs as `ASSET_FILE` instances.
- Completely removed `SVGLoader` and any reference to it.


editor/src/core/webpack-loaders/loaders.spec.ts
```diff
@@ -51,5 +48,4 @@ describe('Applying loaders', () => {
     ['.avif', '.bmp', '.gif', '.jpg', '.jpeg', '.png', '.glb', '.data', '.cica'],
```
Rheeseyb, line 48: Can you add `'.svg'` to this please

Rheeseyb (approved):

editor/src/core/webpack-loaders/loaders.spec.ts
seanparsons: Fixed in 631803532.

bkrmendy (approved):

PR #4480, opened 2023-11-09
On-canvas text editing for variables

bkrmendy:
# Try it [here](https://utopia.pizza/p/00122f15-handsomely-salto/?branch_name=spike-const-edit-stega)
Make sure to turn on the `Steganography` FS

15-42-dpv98-asu24

## Description
This PR adds a narrow happy path for on-canvas text editing of string variables. This is accomplished by adding a Babel transform pass that encodes data about the source location of string literals with steganography and using this data about the source location to update the string contents from the text editor.

## Out of scope
The feature set that this PR supports is deliberately small so that incremental improvements can be added in separate PRs. Such improvements are text editing for strings passed in props, strings defined in objects, strings rendered by conditionals, etc. For the time being, only string-valued variable definitions are supported that are defined in the same file that uses them (see the tests for more details).

## Details
- Added a feature flag for the feature, which is set to `false` by default. Enabling steganography without an FS will break some snapshots (since the unprintable Unicode characters show up in string literals).
- Refactored `renderTestEditorWithCode` so that it takes an options struct instead of a default params (since a 3rd default param was added to `renderTestEditorWithCode`, and the diff created by that would have been bigger than the diff from the options struct refactor)
- Added `CODE_AHEAD_BUT_PLEASE_TELL_VSCODE_ABOUT_IT` as a temporary solution to force send updated files to VSCode. This is necessary because we splice the new string into the source file via directly manipulating the text of the file


editor/src/components/editor/store/vscode-changes.ts
```diff
@@ -106,10 +106,15 @@ export function collateProjectChanges(
           const fileMarkedDirtyButNoCodeChangeYet =
             firstUnsavedContent == null && secondUnsavedContent === firstSavedContent
 
+          const revisionStateWarrantWrite =
+            secondContents.content.fileContents.revisionsState === 'BOTH_MATCH' ||
+            secondContents.content.fileContents.revisionsState ===
+              'CODE_AHEAD_BUT_PLEASE_TELL_VSCODE_ABOUT_IT'
```
Rheeseyb, line 112: This pains me so much to see

editor/src/core/model/project-file-helper-hooks.ts
```diff
@@ -26,3 +29,55 @@ export function useReParseOpenProjectFile(): () => void {
     }
   }, [refEditorState, dispatch])
 }
+
+function spliceCode(
+  originalCode: string,
+  originalStringData: SteganoTextData,
+  updatedString: string,
+): string {
+  const currentString = originalCode.slice(
```
Rheeseyb, line 38: Nit: I think `currentString` is a bit confusing when used with `originalCode` and `originalStringData`. Can we rename it to `originalString` please

editor/src/core/model/project-file-utils.ts
```diff
@@ -499,7 +499,12 @@ export function canUpdateRevisionsState(
     case RevisionsState.ParsedAhead:
       return updated === RevisionsState.ParsedAhead || updated === RevisionsState.BothMatch
     case RevisionsState.CodeAhead:
-      return updated === RevisionsState.CodeAhead || updated === RevisionsState.BothMatch
+    case RevisionsState.CodeAheadButPleaseTellVSCodeAboutIt:
+      return (
+        updated === RevisionsState.CodeAhead ||
+        updated === RevisionsState.BothMatch ||
+        updated === RevisionsState.CodeAheadButPleaseTellVSCodeAboutIt
+      )
```
Rheeseyb, line 507: I don't know if this matters, so I'm only pointing it out in case you then realise it does...
With this fall through in the switch statement, it means that the `RevisionsState` can be updated from `CodeAheadButPleaseTellVSCodeAboutIt` to `CodeAhead`.

editor/src/core/model/project-file-helper-hooks.ts
bkrmendy: renamed in https://github.com/concrete-utopia/utopia/pull/4480/commits/5eac2a988c11e657273664a00a44a31faad7359f

editor/src/core/shared/project-file-types.ts
```diff
@@ -543,12 +543,17 @@ export function isParsedJSONFailure(result: ParsedJSONResult): result is ParsedJ
...
+  | 'PARSED_AHEAD'
+  | 'CODE_AHEAD'
+  | 'BOTH_MATCH'
+  | 'CODE_AHEAD_BUT_PLEASE_TELL_VSCODE_ABOUT_IT'
 
 export const RevisionsState = {
   ParsedAhead: 'PARSED_AHEAD',
   CodeAhead: 'CODE_AHEAD',
   BothMatch: 'BOTH_MATCH',
+  CodeAheadButPleaseTellVSCodeAboutIt: 'CODE_AHEAD_BUT_PLEASE_TELL_VSCODE_ABOUT_IT',
```
seanparsons, line 556: Don't make the change I'm suggesting, but I can imagine this would be nicer if we did:

```typescript
{
  type: 'CODE_AHEAD',
  shouldTellVSCode: boolean
}
```

But what we have is better for making sure we enumerate the cases.

seanparsons (approved):

Rheeseyb (approved):
This is awesome!

editor/src/core/shared/project-file-types.ts
Rheeseyb: That might actually be relevant to my comment (about `CODE_AHEAD` being able to replace `CODE_AHEAD_BUT_PLEASE_TELL_VSCODE_ABOUT_IT`). I think at this point that's more of a theoretical issue, but we could certainly look at this if it becomes a realised issue

bkrmendy: +1

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -790,11 +831,51 @@ function renderJSXElement(
     validPaths.has(EP.toString(EP.makeLastPartOfPathStatic(elementPath)))
   ) {
     if (elementIsTextEdited) {
+      const runJSExpressionWrapper = () => {
```
balazsbajorics, line 834: could be named runJSExpressionLazy

editor/src/components/editor/store/vscode-changes.ts
balazsbajorics: as it should :P

editor/src/components/text-editor/text-editor.spec.browser2.tsx
```diff
@@ -1695,6 +1701,245 @@ describe('Use the text editor', () => {
...
+      await expectSingleUndo2Saves(editor, async () => closeTextEditor())
+      await editor.getDispatchFollowUpActionsFinished()
+
+      expect(editor.getEditorState().editor.mode.type).toEqual('select')
+      expect(getPrintedUiJsCode(editor.getEditorState())).toEqual(
+        formatTestProjectCode(`
+        import * as React from 'react'
+        import { Storyboard, Scene } from 'utopia-api'
+        
+        const variable = 'this is not a constant'
```
balazsbajorics, line 1836: image

editor/src/core/workers/parser-printer/parser-printer-transpiling.ts
```diff
@@ -243,17 +248,72 @@ export function insertDataUIDsIntoCode(
...
+      StringLiteral(path: any) {
+        if (path.node.loc == null) {
+          return
+        }
+
+        // this call somehow messes with the snapshots
+        const smc = new SourceMapConsumer(mapToUse)
+        const originalStartPosition = smc.originalPositionFor(path.node.loc.start)
+        const originalEndPosition = smc.originalPositionFor(path.node.loc.end)
+        const original = cleanSteganoTextData(path.node.value).cleaned
```
balazsbajorics, line 284: is this a safety that you are calling cleanSteganoTextData on the user's original source code?

editor/src/core/workers/parser-printer/parser-printer-transpiling.ts
```diff
@@ -243,17 +248,72 @@ export function insertDataUIDsIntoCode(
...
+            getAbsoluteOffsetFromFile(fileLines, {
+              line: originalStartPosition.line,
+              column: originalStartPosition.column,
+            }) - 1,
+          endPosition: getAbsoluteOffsetFromFile(fileLines, {
+            line: originalEndPosition.line,
+            column: originalEndPosition.column,
+          }),
+        }
+        path.node.value = encodeSteganoData(original, data)
```
balazsbajorics, line 298: is the source remapping automatic, or is that a responsibility of your plugin?

editor/src/core/workers/parser-printer/parser-printer-transpiling.ts
balazsbajorics: I just checked and in https://github.com/concrete-utopia/utopia/blob/1d97eadf633d48cc2e92c6f0ac188be7f340ee9d/editor/src/core/workers/parser-printer/parser-printer-transpiling.ts#L68 we use `path.replaceWith` - maybe that's what we should do here instead of reassigning path.node.value too? 

also, I think we have some tests for source maps, I'm gonna dig them up!

balazsbajorics: here! https://github.com/concrete-utopia/utopia/blob/1d97eadf633d48cc2e92c6f0ac188be7f340ee9d/editor/src/core/workers/parser-printer/parser-printer.spec.ts#L5130

balazsbajorics: this is testing the source maps end-to-end https://github.com/concrete-utopia/utopia/blob/1d97eadf633d48cc2e92c6f0ac188be7f340ee9d/editor/src/components/canvas/ui-jsx-canvas-errors.spec.tsx#L5

PR #4484, opened 2023-11-14
Minimal Liveblocks Authentication.

seanparsons:
**Problem:**
We want to integrate the work previously done for Liveblocks, but need to have the authentication be integrated with ours.

**Fix:**
This includes the smallest sliver of the previously implemented Liveblocks work in the frontend, so that it has enough to prove it works but not pull in all of the implementation of the features.

The integration of the authentication does the following:
- Take the room ID and produce a project ID from it, this may fail the request instantly if the room ID is a generated one passed in early because the Liveblocks API demands _a_ room ID.
- Authorizes the user with Liveblocks, which gives us a token.
- Checks if the user is the owner of the project, if they are then creates the room in Liveblocks.
- Returns the token to the frontend.

**Commit Details:**
- `EditorComponent` pulls in `RoomProvider` for the context it provides.
- `EditorComponentInner` includes a call to `useRoom` to check that it is connecting correctly.
- Included `liveblocks.config.ts` from the earlier collaboration work.
- In the server added a `AuthLiveblocksUser` service call.
- `LiveblocksResources` are created on startup of the server as appropriate.
- `Utopia.Web.Liveblocks.Types` module added containing the code for calling Liveblocks.
- `Utopia.Web.Liveblocks.API` module added to wrap the calls from the above module.
- `Utopia.Web.Liveblocks` module added, contains the interesting logic for calling Liveblocks authorization and creating the Liveblocks room if the user is the owner.

### DIFF
--- a/editor/liveblocks.config.ts
+++ b/editor/liveblocks.config.ts
@@ -1,0 +1,140 @@
+import { createClient } from '@liveblocks/client'
+import { createRoomContext } from '@liveblocks/react'
+import type { CanvasVector } from './src/core/shared/math-utils'
+
+export const liveblocksThrottle = 100 // ms
+
+export const liveblocksClient = createClient({
+  throttle: liveblocksThrottle,
+  authEndpoint: '/v1/liveblocks/authentication',
+})
+
+// Presence represents the properties that exist on every user in the Room
+// and that will automatically be kept in sync. Accessible through the
+// `user.presence` property. Must be JSON-serializable.
+export type Presence = {
+  playerId: string | null
+  playerName: string | null
+  //color: { light: MultiplayerCursorColor; dark: MultiplayerCursorColor } | null
+  cursor: { x: number; y: number } | null
+  canvasScale: number | null
+  canvasOffset: CanvasVector | null
+}
+
+export function initialPresence(): Presence {
+  return {
+    cursor: null,
+    playerId: null,
+    playerName: null,
+    //color: null,
+    canvasScale: null,
+    canvasOffset: null,
+  }
+}
+
+// Optionally, Storage represents the shared document that persists in the
+// Room, even after all users leave. Fields under Storage typically are
+// LiveList, LiveMap, LiveObject instances, for which updates are
+// automatically persisted and synced to all connected clients.
+export type Storage = {
+  // author: LiveObject<{ firstName: string, lastName: string }>,
+  // ...
+}
+
+// Optionally, UserMeta represents static/readonly metadata on each user, as
+// provided by your own custom auth back end (if used). Useful for data that
+// will not change during a session, like a user's name or avatar.
+export type UserMeta = {
+  // id?: string,  // Accessible through `user.id`
+  // info?: Json,  // Accessible through `user.info`
+}
+
+// Optionally, the type of custom events broadcast and listened to in this
+// room. Use a union for multiple events. Must be JSON-serializable.
+export type RoomEvent = {
+  // type: "NOTIFICATION",
+  // ...
+}
+
+// Optionally, when using Comments, ThreadMetadata represents metadata on
+// each thread. Can only contain booleans, strings, and numbers.
+export type ThreadMetadata = {
+  // resolved: boolean;
+  // quote: string;
+  // time: number;
+}
+
+export const {
+  suspense: {
+    RoomProvider,
+    useRoom,
+    useMyPresence,
+    useUpdateMyPresence,
+    useSelf,
+    useOthers,
+    useOthersMapped,
+    useOthersConnectionIds,
+    useOther,
+    useBroadcastEvent,
+    useEventListener,
+    useErrorListener,
+    useStorage,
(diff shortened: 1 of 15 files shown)


editor/liveblocks.config.ts
```diff
@@ -0,0 +1,140 @@
...
+export const liveblocksClient = createClient({
+  throttle: liveblocksThrottle,
+  authEndpoint: '/v1/liveblocks/authentication',
+})
+
+// Presence represents the properties that exist on every user in the Room
+// and that will automatically be kept in sync. Accessible through the
+// `user.presence` property. Must be JSON-serializable.
+export type Presence = {
+  playerId: string | null
```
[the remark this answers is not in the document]
seanparsons: I've trimmed out the entire `Presence` type for now in [40eb8d6](https://github.com/concrete-utopia/utopia/pull/4484/commits/40eb8d634975e997f20706e00f5d7ec89352d039).

editor/src/components/editor/editor-component.tsx
```diff
@@ -504,18 +508,26 @@ export function EditorComponent(props: EditorProps) {
 
   const dispatch = useDispatch()
 
+  const roomId = useEditorState(
+    Substores.restOfEditor,
+    (store) => (projectId == null ? generateUUID() : `project-room-${projectId}`),
+    'EditorComponent roomId',
+  )
+
```
ruggi, line 516:
```suggestion
  const roomId = React.useMemo(
    () => (projectId == null ? generateUUID() : `project-room-${projectId}`),
    [projectId],
  )
```

server/src/Utopia/Web/Liveblocks.hs
```diff
@@ -0,0 +1,80 @@
...
+
+authorizeUserForLiveblocksProjectRoom :: LiveblocksResources -> DatabaseMetrics -> DBPool -> Text -> Text -> IO (Either Text Text)
+authorizeUserForLiveblocksProjectRoom liveblocksResources metrics pool userID roomID = runExceptT $ do
+  -- Try to identify the project ID for this room.
+  projectID <- except $ projectIDFromRoomID roomID
+  -- Authorize the user with Liveblocks, so we can obtain the token that the front end will use to access Liveblocks services.
+  token <- withExceptT show
+            $ ExceptT
+            $ authorizeUserWithLiveblocks liveblocksResources userID
+  -- Check if this user is the owner of the project.
```
ruggi, line 70: If this is enforced, no collaboration can happen.
Since for now we assume all projects to be multiplayer, I think just checking that the projectID exists is enough.

seanparsons: Ah, poor wording on my part, it's not enforced but the following comment I think explains what it's for better than this one does.

seanparsons: Improved the wording of the comment in fdffc6bb6.

editor/src/components/editor/editor-component.tsx
```diff
@@ -504,18 +508,24 @@ export function EditorComponent(props: EditorProps) {
 
   const dispatch = useDispatch()
 
+  const roomId = React.useMemo(
+    () => (projectId == null ? generateUUID() : `project-room-${projectId}`),
```
balazsbajorics, line 512: should / could this generateUUID() part be skipped and the RoomProvider just given a `null` (or any other set of props that makes the RoomProvider dormant / fail silently)?

balazsbajorics (approved):

editor/src/components/editor/editor-component.tsx
seanparsons: Frustratingly, the `id` field is a mandatory field of type `string`. I briefly considered optionally including it in the stack of providers based on a nullable `roomId` but was a bit dubious of what that might do to the editor and it would probably make `useRoom` break due to the `RoomProvider` not being available.

ruggi: nein, liveblocks needs an actual string :( :( :(

editor/src/components/editor/editor-component.tsx
```diff
@@ -88,6 +90,8 @@ function useDelayedValueHook(inputValue: boolean, delayMs: number): boolean {
 }
 
 export const EditorComponentInner = React.memo((props: EditorProps) => {
+  const room = useRoom()
+  console.info('room', JSON.stringify(room.getStatus()))
```
ruggi, line 94: probably worth removing these, but I guess they'll be chomped soon anyways if not

editor/src/components/editor/editor-component.tsx
balazsbajorics: ok thanks :)

editor/src/components/editor/editor-component.tsx
seanparsons: Yeah, I thought it would be useful to keep them so we can see it "working" in other environments until people put some real code in here.

PR #4489, opened 2023-11-15
Handle duplicate imports

liady:
**Problem:**
Importing an element with the same name as an existing element causes a crash (since there are two import statements with the same alias in the file).

**Reproduce:**
Add a `FlexCol` to the playground from the `utils.js` file. Then add a `FlexCol` to the playground from `utopia-api`

**Fix:**
* Change the inserting logic (across the system) to handle and change duplicate import names.
* Change `mergeImports` to return an `importsMergeResolution` - automatically renaming duplicate imports and also returning the duplicate name mapping (so callers can adjust component names if needed)
* Change `addImport` to handle duplicate names
* Hook into adding elements from the top insert menu and the draw-to-add right panel - to change component names when inserting (if needed)

Example (dragging/adding elements with the same name):

https://github.com/concrete-utopia/utopia/assets/7003853/7f1238d1-3f46-4f3c-b359-d2abf745eb3a


editor/src/components/canvas/canvas-strategies/strategies/reparent-utils.ts
```diff
@@ -87,6 +88,42 @@ export function elementToReparent(element: JSXElementChild, imports: Imports): E
...
+  const fileContents = getParseSuccessForFilePath(targetFile, projectContents)
+  if (fileContents == null) {
+    return element
+  }
+  const { imports, duplicateNameMapping } = renameDuplicateImports(
+    fileContents.imports,
+    element.imports,
+  )
+  // handle element name
+  if (element.element.type === 'JSX_ELEMENT') {
```
ruggi, line 105:
```suggestion
  if (isJSXElement(element.element)) {
```

editor/src/components/canvas/commands/reparent-element-command.ts
```diff
@@ -11,32 +11,39 @@ import { mergeImports } from '../../../core/workers/common/project-file-utils'
 import type { EditorState, EditorStatePatch } from '../../editor/store/editor-state'
 import {
   forUnderlyingTargetFromEditorState,
+  modifyUnderlyingTargetElement,
```
ruggi, line 14: unused

editor/src/components/canvas/commands/reparent-element-command.ts
```diff
@@ -11,32 +11,39 @@ import { mergeImports } from '../../../core/workers/common/project-file-utils'
...
   removeElementAtPath,
 } from '../../editor/store/editor-state'
 import type { BaseCommand, CommandFunction, WhenToRun } from './commands'
 import { getPatchForComponentChange } from './commands'
 import type { IndexPosition } from '../../../utils/utils'
 import { insertJSXElementChildren } from '../../../core/model/element-template-utils'
+import type { JSXElement } from '../../../core/shared/element-template'
+import { isJSXElement } from '../../../core/shared/element-template'
+import { renameComponent } from '../../../components/navigator/actions'
```
ruggi, line 23: unused

editor/src/components/editor/import-utils.ts
```diff
@@ -10,31 +14,40 @@ import {
...
+  Imports,
+  NodeModules,
+} from '../../core/shared/project-file-types'
+import { importAlias, importDetails, importsResolution } from '../../core/shared/project-file-types'
 import type { ProjectContentTreeRoot } from '../assets'
 import type { BuiltInDependencies } from '../../core/es-modules/package-manager/built-in-dependencies-list'
 import { withUnderlyingTarget } from './store/editor-state'
 import * as EP from '../../core/shared/element-path'
+import { renameDuplicateImportsInMergeResolution } from '../../core/workers/common/import-worker-utils'
+import { get } from 'objectPath'
```
ruggi, line 29: unused

ruggi (commented):
at first glance it looks ok to me, although I would strongly suggest to add at least unit tests for the `adjustImportNameIfNeeded` and the related logic

liady:
@ruggi thanks! yes, I'm adding exaustive tests to the logic, before officially opening the PR

editor/src/components/canvas/commands/add-imports-to-file-command.ts
```diff
@@ -31,7 +31,12 @@ export const runAddImportsToFile: CommandFunction<AddImportsToFile> = (
     command.targetFile,
     editorState,
     (parseSuccess) => {
-      const updatedImports = mergeImports(command.targetFile, parseSuccess.imports, command.imports)
+      // TODO handle duplicate name mapping
```
balazsbajorics, line 34: are these `// TODO handle duplicate name mapping` still outstanding? do you want to fix these in a follow-up PR?

editor/src/components/canvas/canvas-strategies/strategies/reparent-utils.ts
```diff
@@ -87,6 +88,43 @@ export function elementToReparent(element: JSXElementChild, imports: Imports): E
...
+  const { imports, duplicateNameMapping } = renameDuplicateImports(
+    fileContents.imports,
+    element.imports,
+    targetFile,
+  )
+  // handle element name
+  if (isJSXElement(element.element)) {
+    const elementName = MetadataUtils.getJSXElementName(element.element)?.baseVariable
+    if (elementName != null && duplicateNameMapping.has(elementName)) {
+      element.element.name.baseVariable = duplicateNameMapping.get(elementName)!
```
seanparsons, line 109: Generally we try to avoid mutating incoming parameters like this, it would be better to create a new value and return that so as not to subvert that expectation.

editor/src/core/workers/common/import-worker-utils.ts
```diff
@@ -0,0 +1,202 @@
...
+
+function adjustImportNameIfNeeded(
+  existingNames: ImportUniqueNames,
+  importName: string,
+  importSource: string,
+  type: ImportType,
+  absolutePath: (relativePath: string) => string,
+): string {
+  const existingImport = existingNames.get(importName)
+  if (existingImport! != null) {
```
seanparsons, line 105: Is that exclamation mark necessary?

seanparsons (approved):

editor/src/components/canvas/commands/add-imports-to-file-command.ts
liady: yes, this PR introduces the mechansim change (inside `mergeImports` and other places), and also the usage of it in the `reparenting` flow, I didn't want to make it way too large, so I will have follow-up PRs for specific locations

maltenuhn:
@liady as it's been a while, can you add a couple of gifs or a video with scenarios showing this now working?

editor/src/core/workers/common/import-worker-utils.ts
```diff
@@ -0,0 +1,237 @@
+import type { Imports, ImportsMergeResolution } from '../../../core/shared/project-file-types'
```
Rheeseyb, line 1: This file shouldn't be in the `workers` code since it isn't specifically for the workers. I'd say move it to /core/shared (and rename it to remove the "workers" in the name)

Rheeseyb (approved):

editor/src/components/canvas/commands/insert-element-insertion-subject.ts
```diff
@@ -39,31 +41,35 @@ export const runInsertElementInsertionSubject: CommandFunction<InsertElementInse
   let selectedViews: Array<ElementPath> = []
   const { subject, insertionPath } = command
 
+  let subjectElement: JSXElement = subject.element
```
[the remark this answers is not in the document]
liady: yes, I was worried that working with `subjectElement` for the editorPatch, but using `subject.element.uid` in the description might cause an inconsistency issue in the future (if, for example, for some reason it will be changed inside this function and then the description will have the wrong `uid` of the original one that's not actually in the patch)

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2229,10 +2231,14 @@ export const UPDATE_FNS = {
           success.imports,
           action.importsToAdd,
         )
+
+        // TODO handle duplicate name mapping
```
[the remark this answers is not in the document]
liady: Yes, these are places that use `mergeImports` but does not change the element name (for now). Since this PR is pretty large (changing the entire imports mechanism and fixing reparenting - which is a large and very delicate flow), I wanted to merge it, and then have subsequent very small PRs to handle other cases in the code (copy/paste, wrap, etc).

PR #4490, opened 2023-11-15
fix(editor): don't output entire state object to console

liady:
Fixes #[ticket_number]

**Problem:**
Multiple usages of the PPP debug button cause the editor (and dev tools) to get stuck. This is due to Chrome not handling well `console.info` of large objects

**Fix:**
Don't output the object itself, but rather hang it on the window and let the developer inspect it using `window.entireState`

**Commit Details:**
- change "PPP" behavior


balazsbajorics (approved):

bkrmendy (approved):

Rheeseyb (approved):
Nice!

PR #4492, opened 2023-11-15
Allow inserting variables in scope using the top insert menu

liady:
**Problem:**
We need to allow inserting text variables as text children of the element

**Fix:**
Detect and allow inserting text variables to the selected element

- Did [other things]

### DIFF
--- a/editor/src/components/canvas/ui/floating-insert-menu.tsx
+++ b/editor/src/components/canvas/ui/floating-insert-menu.tsx
@@ -18,10 +18,19 @@ import type {
 import {
   getInsertableGroupLabel,
   getNonEmptyComponentGroups,
+  insertableComponent,
+  insertableComponentGroupProjectComponent,
 } from '../../shared/project-components'
 import { InspectorInputEmotionStyle } from '../../../uuiui/inputs/base-input'
 import { optionalMap } from '../../../core/shared/optional-utils'
 import type { InsertMenuMode } from './floating-insert-menu-helpers'
+import {
+  jsxAttributesFromMap,
+  jsxElement,
+  jsxElementWithoutUID,
+  jsxTextBlock,
+} from '../../../core/shared/element-template'
+import { emptyImports } from '../../../core/workers/common/project-file-utils'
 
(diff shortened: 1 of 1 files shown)


bkrmendy (approved):

editor/src/components/canvas/ui/floating-insert-menu.tsx
```diff
@@ -114,7 +118,26 @@ export function useGetInsertableComponents(
...
+    'useGetInsertableComponents scopedVariables',
+  )
+
+  const insertableVariables = React.useMemo(() => {
+    if (fullPath == null) {
+      return []
+    } else {
+      return convertInsertableComponentsToFlatList(convertVariablesToElements(scopedVariables))
+    }
+  }, [fullPath, scopedVariables])
```
balazsbajorics, line 138: I _think_ the scopedVariables is not referentially equal so it will always blow this memo, if the performance is a concern we can look into wrapping `getVariablesInScope` in selectors

liady: Oh. Does `insertableComponents` dependencies have the same issue? (line 109)?

liady: Yes, good catch. Should be wrapped so it won't be re-created every time.

editor/src/components/shared/scoped-variables.ts
```diff
@@ -0,0 +1,146 @@
...
+        )
+      }),
+    }
+  })
+}
+
+function getMatchingElementForVariable(variable: Variable) {
+  switch (variable.type) {
+    case 'string':
+      return jsxElementWithoutUID('span', jsxAttributesFromMap({}), [
```
balazsbajorics, line 116: writing here for posterity,but we chatted about this. I think it was better when the code used `jsxTextBlock` here, it achieves the exact same outcome but doesn't involve hijacking unrelated types and touching the parser-printer

balazsbajorics (approved):

PR #4493, opened 2023-11-15
1st commenting PR

gbalint:
**Description:**
1st commenting PR!

The PR allows adding comments to canvas coordinates.

- shift-c enters the new comment mode and you can add comments by clicking on the canvas
- you can exit comment mode by pressing esc
- the comments are marked with a basic (not final) comment indicator (using the user initials/avatars can be unified with how it is done for presence as a next step)
- you can retrieve an existing comment thread by clicking on the indicator
- new UI elements are just draft ones
- adding comments do not cause any change in the project, so it is going to work for viewers too
- comment editing and deletion is not available yet
- only light mode design 
- the whole commenting feature is behind a feature switch
- updated the server readme with "Comments" add-on requirement for Liveblocks


editor/src/components/canvas/controls/comment-indicator.tsx
```diff
@@ -0,0 +1,94 @@
...
+                alignItems: 'center',
+                justifyContent: 'center',
+              }}
+            >
+              <div
+                style={{
+                  height: 20,
+                  width: 20,
+                  borderRadius: 10,
+                  background: '#ff00a7', // TODO: vary colors between users or use avatar
```
ruggi, line 68: the presence is using the stuff in the `multiplayer.ts` file for this kind of things (and colors are part of the presence data so you can get it from there and use it as an index for the color lists). Not to be done now, just noting it here FFR :)

ruggi: and https://github.com/concrete-utopia/utopia/pull/4495 will give you user avatar components too!

editor/src/components/canvas/controls/comment-popup.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { UtopiaTheme } from '../../../uuiui'
+import { ErrorBoundary } from '../../../utils/react-error-boundary'
+import { useCanvasCommentThread } from '../../../core/commenting/comment-hooks'
+import { isLoggedIn } from '../../editor/action-types'
+
+export const CommentPopup = React.memo(() => {
+  const mode = useEditorState(
+    Substores.restOfEditor,
+    (store) => store.editor.mode,
+    'CommentIndicator mode',
```
ruggi, line 20:
```suggestion
    'CommentPopup mode',
```

editor/src/components/canvas/controls/comment-popup.tsx
```diff
@@ -0,0 +1,99 @@
...
+  )
+
+  if (!isCommentMode(mode) || mode.location == null) {
+    return null
+  }
+  const { location } = mode
+  if (location == null) {
+    return null
+  }
```
ruggi, line 29: already checked above, this is not needed

editor/src/components/canvas/controls/comment-popup.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const createThread = useCreateThread()
+
+  // TODO: Unify getting name in different multiplayer components
+  const loginState = useEditorState(
+    Substores.userState,
+    (store) => store.userState.loginState,
+    'CommentThread loginState',
+  )
+
+  const name = isLoggedIn(loginState) ? loginState.user.name : null
```
ruggi, line 72: this can be obtained from the presence on the `name` key (and with https://github.com/concrete-utopia/utopia/pull/4495 there's a helper to normalize it)

editor/src/components/canvas/controls/comment-popup.tsx
```diff
@@ -57,8 +60,11 @@ interface CommentThreadProps {
 
 function CommentThread({ x, y }: CommentThreadProps) {
   const thread = useCanvasCommentThread(x, y)
+
   const createThread = useCreateThread()
 
+  const colorIndex = useMyMultiplayerColorIndex() ?? -1
```
ruggi, line 66: it would be better to either default this to `0` or (better?) `null` to be able to pick a good color from the colors list

ruggi (approved):

editor/src/components/canvas/controls/comment-popup.tsx
```diff
@@ -75,10 +81,10 @@ function CommentThread({ x, y }: CommentThreadProps) {
       // Create a new thread
       createThread({
         body,
-        metadata: { type: 'canvas', x: x, y: y, name: name ?? 'Anonymous' },
+        metadata: { type: 'canvas', x: x, y: y, name: name ?? 'Anonymous', colorIndex: colorIndex },
```
ruggi, line 84: can `name` be picked from the presence as well?

gbalint: I would like to pick the name from resolveUsers, so this will be removed. That is what can guarantee that the name here and in the comment popup is the same

ruggi: fine for me! maybe worth adding a TODO comment for this

editor/src/components/canvas/controls/comment-popup.tsx
gbalint: Unfortunately, the comment metadata base type does not allow nullable properties, only string, number and boolean. And I wanted the helper to return the fallback color, and not the first one (which is a hack though....)

bkrmendy (approved):

editor/src/components/canvas/controls/comment-popup.tsx
Rheeseyb: Can you add a comment here explaining this please, because I can definitely see someone changing this later without realising why the `-1` is a deliberate choice

Rheeseyb (commented):
Very nice work!

PR #4495, opened 2023-11-15
Basic multiplayer user bar

### LINKED ISSUE #4494: Basic presence list in sidebar

ruggi:
Fixes #4494 

**Problem:**

Now that we have multiplayer cursors, we need to show who else is in the room.

**Fix:**

This PR implements a basic avatar list in the right panel. The single player user bar will still be displayed as a fallback, but if the room connection succeeds the multiplayer bar will be used instead.

Screenshot 2023-11-15 at 8 02 44 PM

For backwards compatibility, clicking the self avatar (the rightmost one) will still open a link to `/projects`.

If there are no other participants, only the rightmost avatar will be shown.

If there are more than `MAX_VISIBLE_OTHER_PLAYERS` in the same room (set to `4`), an overflow counter will be shown. The tooltip for that item shows the list of names for the avatars not being shown.

Screenshot 2023-11-15 at 8 03 23 PM

As I did with the cursors, I tried to limit this to pure functionality work, with no real attention to the design bits.


editor/src/components/user-bar.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  const roomStatus = useStatus()
+
+  if (!isLoggedIn(loginState)) {
+    return null
+  } else if (roomStatus !== 'connected') {
+    return <SinglePlayerUserBar />
+  } else {
+    return (
+      <ClientSideSuspense fallback={<div />}>{() => <MultiplayerUserBar />}</ClientSideSuspense>
```
gbalint, line 32: I think it is important to wrap this in an `ErrorBoundary`, otherwise a network error can throw a top level exception.
Maybe we should even create our own `SafeLiveblocksWrapper` component which includes an ErrorBoundary and a `ClientSideSuspense`

editor/src/core/shared/multiplayer.ts
```diff
@@ -0,0 +1,86 @@
...
+  const colors =
+    getPreferredColorScheme() === 'dark' ? multiplayerColors.dark : multiplayerColors.light
+  return safeIndex(colors, colorIndex) ?? fallbackColor
+}
+
+/**
+ * Returns the initials from the given name, to be displayed on e.g. avatars or comments.
+ * If the name is made of multiple words, the first letter of the first two words will be returned ('John Doe' -> 'JD').
+ * If the name is made of a single word, the first two letters of the word will be returned ('John' -> 'JO').
+ * If the name is shorter than two letters, the result will be padded with 'X' characters ('F' -> 'FX').
```
balazsbajorics, line 56: why not just leave it shorter?

balazsbajorics (approved):

editor/src/components/user-bar.tsx
```diff
@@ -0,0 +1,163 @@
...
+    </a>
+  )
+})
+SinglePlayerUserBar.displayName = 'SinglePlayerUserBar'
+
+const MultiplayerUserBar = React.memo(() => {
+  const colorTheme = useColorTheme()
+
+  const self = useSelf()
+  const myName = React.useMemo(
```
seanparsons, line 56: I suspect you can get away with not needing to wrap this in a `useMemo`, as it looks to be a string and I'm guessing the cost of `normalizeMultiplayerName` isn't that high?

editor/src/components/user-bar.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  return (
+    <div
+      style={{
+        display: 'flex',
+        alignItems: 'center',
+        gap: 4,
+      }}
+    >
+      {visibleOthers.length > 0 && (
```
seanparsons, line 88:
```suggestion
      {when(visibleOthers.length > 0, ...
```

editor/src/components/user-bar.tsx
```diff
@@ -0,0 +1,163 @@
...
+            return (
+              <MultiplayerAvatar
+                key={`avatar-${other.id}`}
+                name={multiplayerInitialsFromName(name)}
+                tooltip={name}
+                color={multiplayerColorFromIndex(other.colorIndex)}
+              />
+            )
+          })}
+          {hiddenOthers.length > 0 && (
```
seanparsons, line 114:
```suggestion
          {when(hiddenOthers.length > 0, ...
```

editor/src/core/shared/multiplayer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+        want: 'FX',
+      },
+      {
+        name: 'no name at all',
+        input: '',
+        want: 'XX',
+      },
+    ]
+    tests.forEach((test, index) => {
+      it(`${index}/${tests.length + 1} ${test.name}`, async () => {
```
seanparsons, line 37: I believe werre trying to move away from generating the test name like this because it's not so easy to find a test when it fails.

seanparsons (approved):

editor/src/core/shared/multiplayer.spec.ts
balazsbajorics: yes please! ideally we should be writing these test cases in the shape of:

```typescript
      it('a very short name',() => {
        runTest({
          input: 'F',
          want: 'FX',
       })
     })
```

gbalint (approved):

editor/src/core/shared/multiplayer.ts
ruggi: I think at least for me the assumption is to have uniform 2-letter initials for now (there are a lot of possible further variations on this, e.g. names with interfixes so I wanted to go with the most stable solution for now

editor/src/components/user-bar.tsx
ruggi: I absolutely agree - and I think that a dedicated component that does both things (plus more, e.g. error reporting) is the way to go. Incremental PR sounds like the best option since there's a bunch of moving pieces atm

balazsbajorics (approved):
thanks for the test rewrite :)

PR #4496, opened 2023-11-16
record variables in scope in the element metadata

bkrmendy:
## Problem
We don't have a canonical way to determine what variables are in scope for a given element.

## Fix
Create a new prop on `EditorState` called `variablesInScope`, which stores the names and values of these variables. This PR only adds support for recording variables that are in the body of component definitions.


Rheeseyb (approved):

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -1773,6 +1774,8 @@ export const ElementInstanceMetadataKeepDeepEquality: KeepDeepEqualityCall<Eleme
     ConditionValueKeepDeepEquality,
     (metadata) => metadata.textContent,
     nullableDeepEquality(StringKeepDeepEquality),
+    (m) => m.variablesInScope,
+    createCallWithShallowEquals(),
```
balazsbajorics, line 1778: the shallow equals will fail if the variableInScope's value is an object. do we filter out object values? btw we moved the element props to allElementProps exactly so that we don't blow the reference equality of the rest of metadata! it sounds like the variables in scope should also be outside of metadata, next to allElementProps

editor/src/utils/deep-equality.ts
```diff
@@ -809,6 +809,116 @@ export function combine14EqualityCalls<A, B, C, D, E, F, G, H, I, J, K, L, M, N,
   }
 }
 
+export function combine15EqualityCalls<A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, X>(
```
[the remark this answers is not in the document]
bkrmendy: maybe we could even generate a bespoke equality function for each type we want to support. We have some experience with the TS AST API so I think this would be an interesting thing to try

editor/src/components/editor/store/store-deep-equality-instances.ts
bkrmendy: refactored in https://github.com/concrete-utopia/utopia/pull/4496/commits/c48b9228b5d7dceb585bc9cc46e21a61cf66c7a9

balazsbajorics (approved):
nice one!

PR #4497, opened 2023-11-16
Fix/inspector resize to fit

enidemi:
**Problem:**
Resize to fit button is not available for elements with absolute children.

**Fix:**
When an element has only absolute children resize to fit resizes the target element to its children's bounding box. With flow children the behaviour is unchanged.

**Commit Details:**  
- enable the button with absolute children (excluding groups)
- use `convertSizelessDivToFrameCommands` to apply the children bounding box as a frame
- added a test

### DIFF
--- a/editor/src/components/inspector/resize-to-fit-control.spec.browser2.tsx
+++ b/editor/src/components/inspector/resize-to-fit-control.spec.browser2.tsx
@@ -9,7 +9,7 @@ import {
 import { CanvasControlsContainerID } from '../canvas/controls/new-canvas-controls'
 import { mouseClickAtPoint, pressKey } from '../canvas/event-helpers.test-utils'
 import type { EditorRenderResult } from '../canvas/ui-jsx.test-utils'
-import { getPrintedUiJsCode } from '../canvas/ui-jsx.test-utils'
+import { TestAppUID, TestSceneUID, getPrintedUiJsCode } from '../canvas/ui-jsx.test-utils'
 import {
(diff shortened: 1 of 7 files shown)


editor/src/components/inspector/resize-to-fit-control.spec.browser2.tsx
```diff
@@ -103,6 +105,39 @@ describe('Resize to fit control', () => {
...
     expect(view.style.flexBasis).toEqual('')
   })
+  it('when a container div has only absolute children the children bounding box is added as a frame', async () => {
+    const editor = await renderTestEditorWithCode(
+      makeTestProjectCodeWithSnippet(`
+        <div data-uid='component-root'>
+          <div data-uid='container'>
+          <div data-uid='child1' style={{position: 'absolute', left: 50, top: 40, width: 122, height: 112}} />
+          <div data-uid='child2' style={{position: 'absolute', left: 60, top: 30, height: 22}}>hello div</div>
+        </div>
```
balazsbajorics, line 115: missing one indent :)

bkrmendy (approved):

Rheeseyb (approved):

editor/src/components/inspector/resize-to-fit-control.spec.browser2.tsx
enidemi: fixed

balazsbajorics (commented):
👌

PR #4500, opened 2023-11-16
Sanitize string literal contents after canvas editing

bkrmendy:
Depends on https://github.com/concrete-utopia/utopia/pull/4508 

## Problems
- When updating a string literal based on the encoded steganography data and the new text contents from the text editor, the new text contents are spliced as-is into the source file. This can break the code, for example when a newline a single quote is typed in text editor
- it turned out that the source locations from `SourceMapConsumer.originalPositionFor` were off by one in some cases. Apparently this is known thing in `source-map` (based on this issue https://github.com/mozilla/source-map/issues/359)

## Fixes
- Escape the update text by using JSON.stringify
- Come up with an alternative way of computing the source location of the string literal


Rheeseyb (approved):

editor/src/core/model/project-file-helper-hooks.ts
```diff
@@ -35,21 +36,10 @@ function spliceCode(
...
 ): string {
-  const originalString = originalCode.slice(
-    originalStringData.startPosition,
-    originalStringData.endPosition,
-  )
-  if (originalString !== originalStringData.originalString) {
-    throw new Error(`Tried to rewrite string but it was not matching the last known value.
-Current: >>>${originalString}<<<
-Expected: >>>${originalStringData.originalString}<<<
-`)
```
balazsbajorics, line 46: do you think this check was not valuable anymore?

balazsbajorics (approved):

seanparsons (approved):

PR #4502, opened 2023-11-16
Display multiplayer user avatars, fallback to initials

### LINKED ISSUE #4501: display user avatars in multiplayer bar

ruggi:
Fix #4501

**Problem:**

The multiplayer user bar always uses initials bubbles, ignoring any non-default avatars the users might have.

**Fix:**

If the user picture is provided and is not a default Auth0 tile, display that. Otherwise, fallback to the initials bubble.

https://github.com/concrete-utopia/utopia/assets/1081051/1922d9ca-69c3-4ec4-99ec-d262c932f93d

If the user picture fails to load, fallback to the initials bubble as well.

Note: the user picture is being sent with the presence data, but that should be replaced with `ResolveUsers` once we have that in place (as noted in the code).


editor/src/components/user-bar.tsx
```diff
@@ -155,7 +179,20 @@ const MultiplayerAvatar = React.memo(
             cursor: 'pointer',
           }}
         >
-          {props.name}
+          {when(picture == null, props.name)}
+          {when(
```
balazsbajorics, line 183: nit: maybe an `unless( picture == null,` is easier to read, maybe not, I'm not really sure, why am I even typing

balazsbajorics (approved):

editor/src/components/user-bar.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/4502/commits/1859cff20c60bd00d99bd280379904dbf6be40f7

Rheeseyb (approved):

PR #4503, opened 2023-11-16
Support for paths key in jsconfig.json

Rheeseyb:
**Test Project [Here](https://utopia.pizza/p/38a122fe-catkin-cloudberry/?branch_name=feature-jsconfig-paths)**

**Problem:**
Many projects use a `jsconfig.json` to provide path mappings (see https://code.visualstudio.com/docs/languages/jsconfig), which was something that Utopia did not support

**Fix:**
This PR implements basic support for that key with the following limitations:
1. The path mappings are meant to be relative to the `baseUrl`, but this PR assumes that the `baseUrl` is always `'.'` (i.e. the root level of the project)
2. The path mappings themselves each provide an array of lookup locations, whereas this PR only supports the very first value in that array.

I have opted for these trade offs because this feature had already overran, and those are by far the most common uses.

The parsing of the mappings, and the function to apply them live inside `project-file-utils.ts`, so that's where to look for the core of this PR. I have memoised the parsing of the mappings as that file is unlikely to change very often. We then have to use them in two places: during the actual import resolution, and when determining if a rendered element is a focusable component.

For the actual import resolution, we apply the mappings in `module-resolution.ts` inside `resolveModuleAndApplySubstitutions`, which is in the chain of core functions used for resolving the modules (and is the last function in that chain where the file path is still a single string, after which point it is split into segments).

When it comes to determining if a component is focusable, the core change happened inside `project-file-utils.ts` again, in the (badly named) function `isImportedComponent`, which attempts to determine if a component was imported from an npm module. At this point we don't need to apply a mapping, but just need to check if one matches (perhaps there is an edge case where a mapping and an npm module name are allowed to overlap, but if there is then I don't want to know about it!).


editor/src/core/model/element-metadata.spec.browser2.tsx
```diff
@@ -859,6 +867,90 @@ describe('isAutofocusable/isManuallyFocusableComponent', () => {
...
+    const targetPath = EP.fromString('sb/card')
+    expect(MetadataUtils.isAutofocusable(metadata, pathTrees, targetPath)).toBeFalsy()
+    expect(
+      MetadataUtils.isManuallyFocusableComponent(
+        targetPath,
+        metadata,
+        autoFocusedPaths,
+        filePathMappings,
+      ),
+    ).toBeTruthy()
```
seanparsons, line 952: Or whatever similar incantation, so that if this ends up returning an array later on this check doesn't keep giving it the thumbs up:
```suggestion
    ).toEqual(true)
```

seanparsons (approved):

editor/src/core/model/element-metadata.spec.browser2.tsx
Rheeseyb: Good shout 👍

bkrmendy (approved):

PR #4504, opened 2023-11-16
Collaborative editing.

seanparsons:
**Problem:**
We want users to be able to share their projects with each other and see changes _live_.

**Fix:**
The main crux of this is utilising a minimal representation of the project contents (limited to the `topLevelElements`) using `yjs` based types, synchronising changes back and forth. With a project ID based room connected via the Liveblocks support for `yjs`. There's also protection against ping-ponging changes by distinguishing between file changes having a remote origin and those having a local origin.

With this some additional changes were needed:
- Disabling autoforking on save.
- Preventing changes from VS Code applying to the project if the user does not own the project. This was mostly to solve a bug that was triggering code updates shortly as VS Code started up which then conflicted with the first update coming from `yjs`.
- Jest tweaks to get `yjs` and its dependencies to work in the tests.

**Commit Details:**
- Added `@liveblocks/yjs` and `yjs` to `dependencies`.
- Added `@peculiar/webcrypto` to `devDependencies`.
- Tweaked `transformIgnorePatterns` so that `yjs` and one of it's dependencies `lib0` are transformed, making it possible to use them in jest tests.
- Add value for `globalThis.crypto` to `jest-setup-beforeall.js` as a polyfill for the `crypto` global value provided by browsers.
- Re-added `room` variable to `EditorComponentInner` for use by the `LiveblocksProvider` that is created when the yjs document is available.
- Added `ApplyCollabFileUpdate` and `UpdateTopLevelElements` actions.
- Added `collateCollaborativeProjectChanges` as a sister function to `collateProjectChanges`.
- Changed `fileChanges` property of `ProjectChanges` to be of a new type called `ProjectContentProjectChanges` which holds two different values of changes.
- Added a few types prefixed with `Collab` to represent the yjs version of the project contents.
- `collaborativeEditorSupport` added to `EditorStoreShared`.
- Optionally create an instance of `CollaborativeEditingSupportSession` if Liveblocks is available.
- `modifyParseSuccessAtPath` now optionally throws errors if the current path is not parsed already.
- `filesModifiedByElsewhere` added to `EditorState` so that we can distinguish between file updates coming from another editor and those coming from the local editor.
- Implemented `updateCollaborativeProjectContents` to take changes from the regular project contents and apply those to the yjs representation.
- `addHookForProjectChanges` implemented to populate changes from yjs and send those back into the regular project contents using actions.
- Disabled automatic forking of projects when saving.
- `LOAD` action now optionally adds the callback hook `addHookForProjectChanges`.
- `UPDATE_CODE_FROM_EDITOR` now checks the project ownerships before populating changes back from VS Code.


editor/src/components/editor/actions/actions.tsx
```diff
@@ -3672,26 +3690,35 @@ export const UPDATE_FNS = {
...
-    const manualSave = action.unsavedContent == null
-    const code = action.unsavedContent ?? action.savedContent
-
-    let updatedFile: ProjectFile
-    if (existing == null || !isTextFile(existing)) {
-      const contents = textFileContents(code, unparsed, RevisionsState.CodeAhead)
-      const lastSavedContents = manualSave
-        ? null
-        : textFileContents(action.savedContent, unparsed, RevisionsState.CodeAhead)
+    // Prevents updates from from VS Code when the user is not the owner of the project.
```
ruggi, line 3695:
```suggestion
    // Prevents updates from VS Code when the user is not the owner of the project.
```

:D

editor/src/components/editor/actions/actions.tsx
```diff
@@ -5306,6 +5333,44 @@ export const UPDATE_FNS = {
 
     return updatedEditor
   },
+  APPLY_COLLAB_FILE_UPDATE: (action: ApplyCollabFileUpdate, editor: EditorModel): EditorModel => {
+    return modifyParseSuccessAtPath(
+      action.fullPath,
+      editor,
+      (success) => {
+        return success
+        // return action.update.success
```
ruggi, line 5342:
```suggestion
```

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -0,0 +1,359 @@
...
+
+export function addHookForProjectChanges(
+  session: CollaborativeEditingSupportSession,
+  dispatch: EditorDispatch,
+): void {
+  session.projectContents.observeDeep((changeEvents) => {
+    // TODO Check that this is the array change before doing anything
+    for (const changeEvent of changeEvents) {
+      switch (changeEvent.path.length) {
+        case 0: {
```
ruggi, line 208: would it be a good idea to make these values (0, 1, 2) constants with a more descriptive name?

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -0,0 +1,359 @@
...
+  })
+}
+
+export interface ArrayChanges {
+  updatesAt: Array<number>
+  deleteFrom: number | null
+}
+
+function updateEntireProjectContents(
+  session: CollaborativeEditingSupportSession,
```
ruggi, line 241: this does not seem to be used, can it be zapped?

ruggi (commented):
first pass done, it looks fantastic so far. I'll have another look in the morning but so far **so great**

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -0,0 +1,359 @@
...
+  collabFile: CollabTextFile,
+): void {
+  const collabFileTopLevelElements = collabFile.get('topLevelElements')
+  if (collabFileTopLevelElements === 'TEXT_FILE' || collabFileTopLevelElements === undefined) {
+    throw new Error('Invalid value for topLevelElements.')
+  } else {
+    const changes = calculateArrayChanges(
+      success.topLevelElements,
+      collabFileTopLevelElements,
+      // Fix this shitshow up
```
balazsbajorics, line 346: 👀 should this be fixed then?

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -0,0 +1,359 @@
...
+          collabFile.set('type', 'TEXT_FILE')
+          const topLevelElementsArray = new Y.Array<TopLevelElement>()
+          collabFile.set('topLevelElements', topLevelElementsArray)
+        }
+
+        mergeDoc.transact(() => {
+          if (!projectContentsMap.has(change.fullPath)) {
+            projectContentsMap.set(change.fullPath, collabFile)
+          }
+          synchroniseParseSuccessToCollabFile(parsedPart, collabFile)
```
balazsbajorics, line 175: does this need to run in the case when collabFile was set to a `new Y.Map()`? if not, you could move the mergeDoc.transact to inside the two branches of the if-else statement, and remove the variable `let collabFile`

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -0,0 +1,359 @@
...
+        } else if (isTextFile(firstContents.content) && isTextFile(secondContents.content)) {
+          if (
+            ParsedTextFileKeepDeepEquality(
+              firstContents.content.fileContents.parsed,
+              secondContents.content.fileContents.parsed,
+            ).areEqual
+          ) {
+            // Do nothing, no change.
+          } else {
+            const firstRevisionState = firstContents.content.fileContents.revisionsState
```
balazsbajorics, line 70: unused var – is this a problem with the logic that we are not using it? or simply leftover?

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -0,0 +1,359 @@
...
+  newContents: ProjectContentTreeRoot,
+): Array<ProjectFileChange> {
+  let changesToProcess: Array<ProjectFileChange> = []
+
+  function applyChanges(
+    fullPath: string,
+    firstContents: ProjectContentsTree,
+    secondContents: ProjectContentsTree,
+  ): boolean {
+    if (isProjectContentFile(firstContents)) {
```
balazsbajorics, line 57: sad emoji that this is not happy-path-to-the-left

editor/src/components/editor/store/collaborative-editing.ts
seanparsons: I'm not sure what we'd have for descriptive names as they were kinda identified by just trying various changes and seeing what we got.

editor/src/components/editor/store/collaborative-editing.ts
seanparsons: There's not really a single happy path in this however.

editor/src/components/editor/store/collaborative-editing.ts
seanparsons: Just a leftover, fixed in 6d755a897.

editor/src/components/editor/store/collaborative-editing.ts
seanparsons: Fixed in 6d755a897.

editor/src/components/editor/actions/actions.tsx
seanparsons: Fixed in 6d755a897.

editor/src/components/editor/actions/actions.tsx
seanparsons: Fixed in 6d755a897.

editor/src/components/editor/store/collaborative-editing.ts
Rheeseyb: Hahaha this was my original comment about the equality function below, but I dunno, maybe it's not that bad anymore 🤷

editor/src/components/editor/store/collaborative-editing.ts
Rheeseyb: Yeah it still needs to run in both cases

editor/src/components/editor/store/collaborative-editing.ts
seanparsons: Comment removed in 6d755a897.

editor/src/components/editor/store/collaborative-editing.ts
Rheeseyb: In the `new Y.Map()` case we're having to create the new `Y.Array` and insert that into the map, and then insert all of the top level elements into that. I guess we could duplicate the code that inserts the elements, but I don't think that would necessarily make the code any clearer.

editor/src/components/editor/store/collaborative-editing.ts
seanparsons: As an alternative, added an explanatory comment to each case in 0ffcad7b2.

editor/jest.config.js
```diff
@@ -116,7 +116,7 @@ module.exports = {
         '\\.[jt]sx?$': 'babel-jest',
       },
       roots: ['src'],
-      transformIgnorePatterns: ['<rootDir>/node_modules/.pnpm/(?!file+..+)'],
+      transformIgnorePatterns: ['<rootDir>/node_modules/.pnpm/(?!file+..+|yjs|lib0)'],
```
Rheeseyb, line 119: What's the deal with this? 👀

editor/src/components/editor/actions/actions.tsx
```diff
@@ -902,6 +911,7 @@ export function restoreEditorState(
     refreshingDependencies: currentEditor.refreshingDependencies,
     colorSwatches: currentEditor.colorSwatches,
     internalClipboard: currentEditor.internalClipboard,
+    filesModifiedByElsewhere: currentEditor.filesModifiedByElsewhere,
```
Rheeseyb, line 914: Ok I know I was the one that named this during the spike, but it is an awful name 😬 Can we rename it to `filesModifiedByAnotherUser` please?

editor/jest.config.js
seanparsons: That's so that the `yjs` and `lib0` libraries do get transformed to make them work with jest's module loader, as they're ES modules from memory which doesn't otherwise work.

editor/src/components/editor/actions/actions.tsx
seanparsons: Fixed in e30d6bb6c.

editor/src/components/editor/actions/actions.tsx
```diff
@@ -5306,6 +5333,43 @@ export const UPDATE_FNS = {
...
+  APPLY_COLLAB_FILE_UPDATE: (action: ApplyCollabFileUpdate, editor: EditorModel): EditorModel => {
+    return modifyParseSuccessAtPath(
+      action.fullPath,
+      editor,
+      (success) => {
+        return success
+      },
+      false,
+    )
+  },
```
Rheeseyb, line 5345: This whole action can go in the bin now that we're not using it. I think this was a leftover from the first implementation

seanparsons: Fixed in 4a4487461.

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -0,0 +1,363 @@
...
+            updateEntireProjectContents(changeEvent as Y.YMapEvent<any>, dispatch)
+          } else {
+            throw new Error(`Could not treat change event as Y.YMapEvent.`)
+          }
+          break
+        }
+        // Originally thought to be a case that would arise on a new addition of
+        // a file, left here to capture this specific case in case it does show up.
+        case 1: {
+          console.error(`Unhandled path length of 1: ${changeEvent.path}`)
```
Rheeseyb, line 220: This should probably be a `throw new Error` then so that it is less likely to be lost in the console

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -0,0 +1,363 @@
...
+}
+
+function updateEntireProjectContents(
+  changeEvent: Y.YMapEvent<any>,
+  dispatch: EditorDispatch,
+): void {
+  let actions: Array<EditorAction> = []
+  const entriesIterator = changeEvent.keys.entries()
+  let entriesIteratorResult = entriesIterator.next()
+  while (!entriesIteratorResult.done) {
```
Rheeseyb, line 253: I believe we could use a `for (const [filename, changeEntry] of entriesIterator) {` here

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -0,0 +1,363 @@
...
+  let updatesAt: Array<number> = []
+  for (let index: number = 0; index < Math.max(from.length, into.length); index++) {
+    if (!equals(from[index], into.get(index))) {
+      updatesAt.push(index)
+    }
+  }
+
+  let deleteFrom: number | null = null
+  if (into.length > from.length) {
+    deleteFrom = into.length
```
Rheeseyb, line 330: Shouldn't this be `deleteFrom = from.length`?

editor/src/components/editor/store/collaborative-editing.ts
seanparsons: True, that's more consistent with the rest of the function as well, corrected in 21cee15c7.

editor/src/components/editor/store/collaborative-editing.ts
seanparsons: That's much cleaner, changed in 21cee15c7.

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -0,0 +1,359 @@
...
+
+  return arrayChanges(updatesAt, deleteFrom)
+}
+
+function synchroniseParseSuccessToCollabFile(
+  success: ParseSuccess,
+  collabFile: CollabTextFile,
+): void {
+  const collabFileTopLevelElements = collabFile.get('topLevelElements')
+  if (collabFileTopLevelElements === 'TEXT_FILE' || collabFileTopLevelElements === undefined) {
```
Rheeseyb, line 341: `collabFileTopLevelElements === 'TEXT_FILE'` I'm guessing this is left over from a previous iteration?

seanparsons: Nope, we have to do this to get the correct type inference.

Rheeseyb (approved):

editor/src/components/editor/store/collaborative-editing.ts
Rheeseyb: Oh because of the awkward `Y.Map` type. Ok

ruggi (approved):

editor/jest.config.js
balazsbajorics: good point, @seanparsons could you leave a comment there saying so? and the same for globalThis.crypto in the other file

editor/src/components/editor/actions/actions.tsx
```diff
@@ -5306,6 +5332,33 @@ export const UPDATE_FNS = {
...
+            topLevelElements: newTopLevelElementsDeepEquals.value,
+          }
+        }
+      },
+      false,
+    )
+
+    return {
+      ...updatedEditor,
+      filesModifiedByAnotherUser: updatedEditor.filesModifiedByAnotherUser.concat(action.fullPath),
```
balazsbajorics, line 5359: something about it doesn't feel right. I think the action's name should be more specific, and contain Collab, or Remote, or some other word indicating that we are updating the top level elements here because an update came from the internet

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -646,6 +652,24 @@ export function editorDispatchClosingOut(
       updatedFromVSCodeOrParsedAfterCodeChange,
     )
     applyProjectChangesToVSCode(frozenEditorState, projectChanges)
+    if (finalStore.collaborativeEditingSupport.session != null) {
```
balazsbajorics, line 655: couldn't this block move above the creation of the original `finalStore`? so by the time we make the finalStore we have the final_final filesModifiedByAnotherUser value?

editor/src/components/editor/actions/actions.tsx
Rheeseyb: Yeah I think that's a good point. It doesn't feel right to you because from the action name it's not at all obvious that there is some logic in here that will always run but in fact should only run as a result of a collaborative editing change. I think at this point simply renaming the action is enough, but this is partly why we created `actions-from-vscode.ts`

editor/jest.config.js
seanparsons: Added some explanatory comments in e62ed6363.

editor/src/components/editor/store/collaborative-editing.ts
seanparsons: Fixed in 8dd00397d.

editor/src/components/editor/actions/actions.tsx
seanparsons: Fixed in ea4a8cf0e.

editor/src/components/editor/store/dispatch.tsx
seanparsons: Unless there's a strong case for making this change now, I'd rather leave this for a subsequent refactor.

Rheeseyb (approved):

PR #4506, opened 2023-11-16
Basic follow mode

ruggi:
#4505 

This PR introduces the basic implementation of follow mode for multiplayer.

When connected to a room, a user can click on another user's canvas by clicking on its avatar in the right panel.
While following another user, its canvas constraints (offset + scale) will be applied to all its followers.

https://github.com/concrete-utopia/utopia/assets/1081051/0676d5cf-3357-46b8-a1eb-a1071483a95c

**Notes & gotchas:**
- Like for the other initial PRs, no proper design work here, just functionality.
- During follow mode, the avatar has a `primary` ring outline (and "glow" shadow) and a permanent toast is displayed at the bottom of the screen.
- If the followed user leaves the room, following stops.
- While in follow mode, pointer events on the canvas are disabled as well as zoom and pan actions.
- The followed id is stored in the new editor mode, `follow`.
- No animations or easing for the canvas pan in this PR; they will jitter, but we should fix it separately as it's gonna be more involved than simple cursor springs.
- Split the `MultiplayerCursors` into separate components and renamed it to a more proper `MultiplayerPresence`. The file name should be changed as well, but I kept it there for now to make reviewing easier (will fix later! 🤞).


editor/src/components/canvas/multiplayer-cursors.tsx
```diff
@@ -184,3 +222,114 @@ const MultiplayerCursor = React.memo(
...
+        dispatch([
+          updateMultiplayerState({ following: null }),
+          switchEditorMode(EditorModes.selectMode(null, false, 'none'), undefined, true),
+        ])
+      }
+      return
+    }
+
+    let actions: EditorAction[] = []
+    if (followed.presence.canvasScale != null && followed.presence.canvasScale !== canvasScale) {
```
gbalint, line 272: Unnecessary null check, canvasScale is a number

gbalint: Oh sorry, you check non-equality to canvasScale, null check is necessary!

editor/src/components/canvas/multiplayer-cursors.tsx
```diff
@@ -184,3 +222,114 @@ const MultiplayerCursor = React.memo(
...
+      actions.push(
+        updateMultiplayerState({
+          following: multiplayerStateFollowing(followed.id, followed.presence.mode),
+        }),
+      )
+      switch (followed.presence.mode) {
+        case 'live':
+          actions.push(switchEditorMode(EditorModes.liveMode(null), undefined, true))
+          break
+        default:
```
gbalint, line 291: I think it is better to list all modes explicitly and assertNever in the default, to make sure we don't forget about this when we introduce a new mode.

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2030,6 +2033,9 @@ export const UPDATE_FNS = {
     }
   },
   UPDATE_EDITOR_MODE: (action: UpdateEditorMode, editor: EditorModel): EditorModel => {
+    if (!action.must && editor.multiplayer.following != null) {
```
gbalint, line 2036: I have a negative gut feeling here, can't we decide whether to really update the editor mode where we dispatch the action?

gbalint (approved):

bkrmendy (approved):

editor/src/components/canvas/controls/follow-mode/follow-mode-hooks.tsx
```diff
@@ -0,0 +1,11 @@
+import type { MouseCallbacks } from '../select-mode/select-mode-hooks'
+import { NO_OP } from '../../../../core/shared/utils'
+import { useKeepShallowReferenceEquality } from '../../../../utils/react-performance'
+
+export function useFollowModeSelectAndHover(): MouseCallbacks {
+  return useKeepShallowReferenceEquality({
```
balazsbajorics, line 6: the minorest of nitpicks, but useKeepShallowReferenceEquality is probably slower than React.useMemo, which is slower than storing the return value in a global const in the module root

ruggi: Makes sense. In https://github.com/concrete-utopia/utopia/pull/4506/commits/2c1813f2c35af51994950ebd433900cc41e4aed5 I kept it as a hook for consistency, but returning a const. Would that work for you?

balazsbajorics (approved):

editor/src/components/canvas/multiplayer-cursors.tsx
```diff
@@ -187,3 +219,90 @@ const MultiplayerCursor = React.memo(
   },
 )
 MultiplayerCursor.displayName = 'MultiplayerCursor'
+
+const FollowingOverlay = React.memo(() => {
+  const colorTheme = useColorTheme()
+  const dispatch = useDispatch()
+
+  const self = useSelf()
```
Rheeseyb, line 227: `self` here is shadowing a global, so we should rename it to something else

PR #4507, opened 2023-11-17
Clean steganography data when replacing prop values on copy

bkrmendy:
## Problem
When copy-pasting an element, we replace any props coming from variables with their runtime values. However, if these props have steganography data encoded in them, we don't clean out that data, and it shows up as gibberish characters in the pasted component

## Fix
Clean the steganography data on paste

### Out of scope
Since we don't support encoding steganography data into string literals passed in props (ie `<Button label={"OK"} ... />`), the rest of the paste code doesn't need to deal with this problem yet. Solving the cleaning problem for the scenario above is deferred to after we implement steganography for props with string literals.

### DIFF
--- a/editor/src/components/editor/actions/actions.spec.browser2.tsx
+++ b/editor/src/components/editor/actions/actions.spec.browser2.tsx
@@ -40,6 +40,7 @@ import {
   expectSingleUndoNSaves,
   searchInFloatingMenu,
   selectComponentsForTest,
+  setFeatureForBrowserTestsUseInDescribeBlockOnly,
 } from '../../../utils/utils.test-utils'
 import {
   firePasteEvent,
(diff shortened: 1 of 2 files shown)


Rheeseyb (approved):
Good catch!

balazsbajorics (approved):

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -5673,6 +5674,84 @@ export var storyboard = (
...
+                title='hello'
+                style={{ top: 0, left: 0, position: 'absolute' }}
+              />
+            </Scene>
+          </Storyboard>
+        )
+`),
+        )
+
+        expect(1).toEqual(1)
```
Rheeseyb, line 5752:
```suggestion
```
I'm guessing this was to shut VS Code up whilst you were writing the test :)

bkrmendy: whoops, sorry, thanks for the suggestion!

seanparsons (approved):

PR #4515, opened 2023-11-20
Avoid follow mode loops

### LINKED ISSUE #4514: Prevent loops in follow mode

ruggi:
Fix #4514 

**Problem:**

It should not be possible to have follow mode loops that would lock the editor for all players (e.g. `a` → `b` → `c` → `a`).

**Fix:**

Prevent those loops when turning follow mode on.

### DIFF
--- a/editor/src/core/shared/multiplayer.ts
+++ b/editor/src/core/shared/multiplayer.ts
@@ -106,3 +106,22 @@ export function isDefaultAuth0AvatarURL(s: string | null): boolean {
     (s.match(reAuth0DefaultAvatarURLEncoded) != null || s.match(reAuth0DefaultAvatar) != null)
   )
 }
+
+export function canFollowTarget(
+  selfId: string,
+  targetId: string | null,
+  others: { id: string; following: string | null }[],
+): boolean {
+  let followChain: Set<string> = new Set()
+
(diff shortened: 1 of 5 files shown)


editor/src/core/shared/multiplayer.ts
```diff
@@ -106,3 +106,22 @@ export function isDefaultAuth0AvatarURL(s: string | null): boolean {
...
+  let followChain: Set<string> = new Set()
+
+  let id = targetId
+  while (id != null) {
+    if (!followChain.has(id)) {
+      followChain.add(id)
+    }
+    const target = others.find((o) => o.id === id)
+    id = target?.following ?? null
+  }
```
Rheeseyb, line 124: I know that implicitly this should never be an infinite loop, but I dislike that there are no explicit guarantees of that in this piece of code.
If `followChain.has(id)` is `true`, then that already surely implies that there is a loop in this chain, in which case we should arguably exit the `while` loop and return `false`

ruggi: fine by me https://github.com/concrete-utopia/utopia/pull/4515/commits/3807342cbed90a3c961619596c5c33e35c765f3e

gbalint (approved):

PR #4516, opened 2023-11-20
Store user info in liveblocks room

gbalint:
**Problem:**
1. The commenting components can not show user information, because we don't provide a `resolveUsers` function to `createRoomContext`
2. `Presence` stores user related persistent information (name, avatar and colorIndex), even though the role of that should be to only show the transient state to of presence.

**Fix:**
Theoretically, we should provide an endpoint which can be used get user information from our backend. However, as a quick solution I decided to store the name/avatar/colorIndex in the Liveblocks room storage.

I deleted all redundant information from liveblocks:
- name from the thread metadata
- name, avatar and colorIndex from Presence

The extra upside of this is that in a single project your color remains the same across presence sessions.

### DIFF
--- a/editor/src/components/canvas/multiplayer-cursors.tsx
+++ b/editor/src/components/canvas/multiplayer-cursors.tsx
@@ -1,20 +1,19 @@
 import type { User } from '@liveblocks/client'
 import { motion } from 'framer-motion'
 import React from 'react'
-import type { Presence, UserMeta } from '../../../liveblocks.config'
 import {
   useOthers,
   useOthersListener,
   useRoom,
   useSelf,
+  useStorage,
   useUpdateMyPresence,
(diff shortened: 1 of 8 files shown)


bkrmendy (approved):

editor/src/components/canvas/multiplayer-cursors.tsx
```diff
@@ -305,7 +289,7 @@ const FollowingOverlay = React.memo(() => {
     }
   })
 
-  if (followed == null) {
+  if (followed == null || followedUser == null) {
```
ruggi, line 292: if `followed != null` and `followedUser == null` it would be worth writing something to the console because it should not happen

ruggi (approved):

PR #4517, opened 2023-11-20
Shadow root canvas

Rheeseyb:
**Problem:**
Utopia's CSS was affecting user projects, and user CSS was affecting Utopia

**Fix:**
Wrap the contents of the canvas in a shadow root. 

The core of the changes are in `ui-jsx-canvas.tsx`, where we use [`react-shadow-root`](https://github.com/apearce/react-shadow-root) to attach the shadow root. Because this breaks emotion support (as emotion appends `style` tags into the document's `head`) I've also introduce a wrapper component which directs emotion to instead introduce those `style` tags into a container `div` inside the shadow root.

This of course lead to tons of fallout as the contents of the shadow root isn't directly queryable in the tests, so to fix that I've added a function `getRenderedCanvas()` inside `ui-jsx-canvas.tsx` to provide a way to query the contents of the canvas itself in the same way that we were previously using `renderedDOM`. This does mean that unfortunately when running the tests we have to use `renderedDOM` to test editor UI, and `getRenderedCanvas()` to test the running test project. The reason this is a function is so that it will still return the correct shadow root if the canvas remounts during a test.

One other thing that required fixing - for some of our older tests we use `ReactDOMServer.renderToStaticMarkup` to take snapshots of the rendered DOM to check for general correctness of our spy wrapping and patching. Because of the way the shadow root is attached (via creating a placeholder element, and then using that to create and attach the shadow root in a `useEffect`), we need to switch to using a declarative shadow root for these tests (this is where the shadow root is created using a `template` element which includes the children, and is used specifically for server side rendering). Since I could find no reliable way to detect whether the code was being run in a SSR environment (the [`isSSR`](https://github.com/concrete-utopia/utopia/blob/c44e0ac5de859169d0caf25957031ba153a6503e/editor/src/core/shared/atom-with-pub-sub.ts#L8) function we use elsewhere isn't able to differentiate in this case) I had to explicitly add a `renderedInSSR` boolean param to the canvas props object, so that I could use `<ReactShadowRoot declarative={props.renderedInSSR}>`.

The remaining changes are snapshots updated to include the shadow root. If you want to view those, I highly recommend hiding whitespace changes!

Finally, there were some tests that incorrect, but somehow passing previously. These changes seem have been enough to (correctly) cause those tests to fail! Those tests are in `inspector-end-to-end-tests.spec.browser2.tsx`, and I've commented out (and `FIXME`d) the offending lines.


bkrmendy (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -571,20 +576,54 @@ export const UiJsxCanvas = React.memo<UiJsxCanvasPropsWithErrorCallback>((props)
...
-            <SceneLevelUtopiaCtxAtom.Provider value={sceneLevelUtopiaContextValue}>
-              {StoryboardRoot}
-            </SceneLevelUtopiaCtxAtom.Provider>
-          </CanvasContainer>
-        </UtopiaProjectCtxAtom.Provider>
-      </RerenderUtopiaCtxAtom.Provider>
+      <ReactShadowRoot declarative={props.renderedInSSR}>
+        <EmotionWrapper>
+          <style>{`div,
+          span,
```
seanparsons, line 585: Why are these elements specifically called out here?

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -571,20 +576,54 @@ export const UiJsxCanvas = React.memo<UiJsxCanvasPropsWithErrorCallback>((props)
...
+              </CanvasContainer>
+            </UtopiaProjectCtxAtom.Provider>
+          </RerenderUtopiaCtxAtom.Provider>
+        </EmotionWrapper>
+      </ReactShadowRoot>
+    </div>
+  )
+})
+
+const EmotionWrapper = React.memo(({ children }: { children?: React.ReactNode | undefined }) => {
```
seanparsons, line 611: Might be worth putting the explanation from the PR description against this for future people.

editor/src/core/webpack-loaders/css-loader.ts
```diff
@@ -29,14 +30,22 @@ const loadModule: LoadModule = (filename: string, contents: string) => {
           styleTag.type = "text/css";
           styleTag.id = elementId;
           styleTag.appendChild(document.createTextNode(content));
-          document.querySelector("head").appendChild(styleTag);
+
+          // If the canvas container doesn't yet exist, wait for the next frame
```
seanparsons, line 34: Ewww.

seanparsons (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
Rheeseyb: Ok good point actually, I need to add a `FIXME` here.
This styling is required, and happened to be applied to the canvas via our editor's css here <https://github.com/concrete-utopia/utopia/blob/master/editor/resources/editor/css/initial-load.css#L58-L67>. At one point it was discovered that certain tests would fail because `border-box` was missing, so it was added to all tests here: <https://github.com/concrete-utopia/utopia/blob/c44e0ac5de859169d0caf25957031ba153a6503e/editor/src/components/canvas/ui-jsx.test-utils.tsx#L606-L614>
But that no longer works because of the shadow root, so we now need to include that border-box styling here inside the shadow-root. Really we shouldn't be including this at all, but that's a separate issue to fix (hence why I need to add a `FIXME` here)

editor/src/components/canvas/ui-jsx-canvas.tsx
Rheeseyb: Good shout!

PR #4518, opened 2023-11-20
Disable non-owner editing.

seanparsons:
**Problem:**
When using the editor collaboratively, it doesn't make a lot of sense for non-owners to be able to change the project.

**Fix:**
In two places this prevents changes to the project model from being applied:
- `processAction`: The main hub of where changes are applied we protect against non-owners from applying non-transient editor actions.
- `handleStrategies`: Canvas interactions are disabled by having the entire strategies logic be disabled if the current user is not the owner of the project.

Additionally in `processAction` we prevent changes coming from non-owners via yjs/Liveblocks from being applied to the project model, as it appears initialisation of it in certain circumstances can cause those to be triggered.

**Commit Details:**
- `handleStrategies` now short circuits if the project is not owned by the current user.
- `processAction` now only permits certain actions to run depending on the ownership of the current project.

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -99,12 +99,11 @@ export type DispatchResult = EditorStoreFull & DispatchResultFields
 
 const cannotUndoRedoToastId = 'cannot-undo-or-redo'
 
-function processAction(
-  dispatchEvent: EditorDispatch,
+function checkIfActionShouldBeProcessed(
   editorStoreUnpatched: EditorStoreUnpatched,
   action: EditorAction,
-  spyCollector: UiJsxCanvasContextData,
-): EditorStoreUnpatched {
+): boolean {
+  // Default to true.
   let shouldProcessAction: boolean = true
   // By default when the current user does not own the project and the action is non-transient prevent actions from running...
   if (editorStoreUnpatched.projectServerState.isMyProject === 'no' && !isTransientAction(action)) {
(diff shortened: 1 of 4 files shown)


ruggi (approved):

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -105,168 +105,195 @@ function processAction(
...
+      action.action === 'UPDATE_FROM_WORKER' ||
+      action.action === 'UPDATE_TOP_LEVEL_ELEMENTS_FROM_COLLABORATION_UPDATE'
+    shouldProcessAction = allowedNonOwnerAction
+  }
+  // When the current user does own the project...
+  if (editorStoreUnpatched.projectServerState.isMyProject === 'yes') {
+    // ...Disallow these updates as they're coming from non-owners.
+    shouldProcessAction = action.action !== 'UPDATE_TOP_LEVEL_ELEMENTS_FROM_COLLABORATION_UPDATE'
+  }
+
```
balazsbajorics, line 122: maybe nit, but I'd change this to
`const shouldProcessAction = isMyProject || isTransientAction || allowedNonOwnerAction`

and not even dispatch UPDATE_TOP_LEVEL_ELEMENTS_FROM_COLLABORATION_UPDATE if the update doesn't come from a project owner

seanparsons: The latter bit is a bit nasty as it hangs off a callback that we assign during the `LOAD` action, at which point it would likely not be known if the user is the owner because that fires asynchronously.

balazsbajorics: in that case, you could move it inside `UPDATE_TOP_LEVEL_ELEMENTS_FROM_COLLABORATION_UPDATE` to just return the current editor

seanparsons: Fixed up in 889fd11d1.

PR #4519, opened 2023-11-20
Multiplayer shadows

### LINKED ISSUE #4525: Multiplayer shadows

ruggi:
Fix #4525 

**Problem:**

With collaborative editing enabled, it's difficult to understand who is doing what on the screen.

**Fix:**

This PR introduces collaborative shadows in the editor.
When something is being interacted with via the mouse, a shadow outline is shown to the other participants in the room.

https://github.com/concrete-utopia/utopia/assets/1081051/b0132d10-a8d9-4ee1-8847-562fd23b86dd

Two shadows per element are shown:
1. a border outline of the _original_ element being moved
2. a shadow of the _current_ element (live-ish), which also contains a string representing which action is being performed on the element (e.g. `move`, `resize`, etc.)

The colors of the outlines match the player colors.

The shadows are returned as either straight frames (where possible, to improve performance), or element paths and their frames get calculated in the multiplayer component when updating the presence.

As usual, this is a basic implementation with not much design work behind it.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.tsx
@@ -27,7 +27,7 @@ import { EdgePositionLeft, EdgePositionTop, EdgePositionTopLeft } from '../../ca
 import { isEdgePositionEqualTo } from '../../canvas-utils'
 import { pushIntendedBoundsAndUpdateGroups } from '../../commands/push-intended-bounds-and-update-groups-command'
 import { queueTrueUpElement } from '../../commands/queue-true-up-command'
-import { setActiveFrames } from '../../commands/set-active-frames-command'
+import { activeFrameTargetRect, setActiveFrames } from '../../commands/set-active-frames-command'
 import { setCursorCommand } from '../../commands/set-cursor-command'
 import { setElementsToRerenderCommand } from '../../commands/set-elements-to-rerender-command'
 import { setSnappingGuidelines } from '../../commands/set-snapping-guidelines-command'
@@ -264,7 +264,13 @@ export function absoluteResizeBoundingBoxStrategy(
                   'starting-metadata',
                 ),
                 queueTrueUpElement(childGroups.map(trueUpGroupElementChanged)),
-                setActiveFrames([{ frame: newFrame, action: 'resize' }]),
+                setActiveFrames([
+                  {
+                    action: 'resize',
+                    target: activeFrameTargetRect(newFrame),
+                    source: originalBoundingBox,
+                  },
+                ]),
               ]
             })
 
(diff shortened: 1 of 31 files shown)


editor/src/components/canvas/canvas-strategies/strategies/absolute-resize-bounding-box-strategy.tsx
```diff
@@ -263,6 +264,13 @@ export function absoluteResizeBoundingBoxStrategy(
                   'starting-metadata',
                 ),
                 queueTrueUpElement(childGroups.map(trueUpGroupElementChanged)),
+                setActiveFrames([
```
balazsbajorics, line 267: selectedElement, newFrame, ..... hmm these are also available in pushIntendedBoundsAndUpdateGroups

the action: 'resize' is the one piece that's missing

actually I wonder if the name `pushIntendedBoundsAndUpdateGroups` could be changed to `setStrategyIntent`, so it could store resize, target, frame (even originalFrame). it feels like if a strategy forgets to call `pushIntendedBoundsAndUpdateGroups` that is a mistake. I wonder if `setStrategyIntent` could be elevated to the level of `StrategyApplicationResult`, meaning it would be impossible to return from the apply function without also setting what was the intent

editor/src/components/canvas/multiplayer-cursors.tsx
```diff
@@ -89,8 +92,9 @@ export const MultiplayerPresence = React.memo(() => {
 
   return (
     <>
-      <MultiplayerCursors />
       <FollowingOverlay />
+      <MultiplayerShadows />
+      <MultiplayerCursors />
```
balazsbajorics, line 97: file badly needs renaming

balazsbajorics: or we have to move the other components to their own files

ruggi: it absolutely does, but I'd go with a separate PR to not pollute the diff

editor/src/core/shared/math-utils.ts
```diff
@@ -136,6 +136,12 @@ export function forceFiniteRectangle<C extends CoordinateMarker>(
   throw new Error('invariant: we expected a finite Rectangle')
 }
 
+export function canvasRectangleOrZeroRect(
```
balazsbajorics, line 139: haha I think you are looking for `zeroRectIfNullOrInfinity`

ruggi: lol, that's the one!

ruggi: https://github.com/concrete-utopia/utopia/pull/4519/commits/eb3b899eb2c238cf0e7b057e2fb3b7aaf2bf1d18

editor/src/components/canvas/multiplayer-cursors.tsx
```diff
@@ -331,3 +335,134 @@ const FollowingOverlay = React.memo(() => {
...
+    Substores.canvas,
+    (s) => s.editor.canvas.interactionSession?.interactionData,
+    'MultiplayerShadows interactionData',
+  )
+
+  const editorRef = useRefEditorState((store) => ({
+    jsxMetadata: store.editor.jsxMetadata,
+  }))
+
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
ruggi: https://github.com/concrete-utopia/utopia/pull/4519/commits/163cca700929156e1e0ce55a8753cdee9011284f

seanparsons (approved):

PR #4524, opened 2023-11-21
Add avatar to comment indicator

gbalint:
**Problem:**
Comment indicator should show the avatar when it is available.

### DIFF
--- a/editor/src/components/canvas/controls/comment-indicator.tsx
+++ b/editor/src/components/canvas/controls/comment-indicator.tsx
@@ -16,6 +16,7 @@ import {
   normalizeMultiplayerName,
 } from '../../../core/shared/multiplayer'
 import { MultiplayerWrapper } from '../../../utils/multiplayer-wrapper'
+import { when } from '../../../utils/react-conditionals'
 
 export const CommentIndicator = React.memo(() => {
   const projectId = useEditorState(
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/controls/comment-indicator.tsx
```diff
@@ -101,7 +103,19 @@ function CommentIndicatorInner() {
                   boxShadow: UtopiaTheme.panelStyles.shadows.medium,
                 }}
               >
-                {initials}
+                {avatar == null ? (
+                  <span>{initials}</span>
+                ) : (
+                  <img
```
ruggi, line 109: can you add an error handler like the one in `user-bar.tsx`? e.g. google login avatars tend to produce a lot of 403s for some reason, falling back to initials makes more sense than showing the broken image placeholder

ruggi (approved):

editor/src/components/canvas/controls/comment-indicator.tsx
balazsbajorics: we could create an ImgWithFallback component that wraps around Federico's error handler

gbalint: I made something similar, but just Avatar specific, called AvatarPicture

balazsbajorics (approved):

PR #4531, opened 2023-11-21
Make comment mention suggestions work

gbalint:
**Description:**
Fill in the `resolveMentionSuggestions` template function, so liveblocks can resolve mentions.
Searching for mentions searches by the name, and returns the user id.
The search relies on the collaborators in the liveblocks room storage.

image

### DIFF
--- a/editor/liveblocks.config.ts
+++ b/editor/liveblocks.config.ts
@@ -134,42 +134,46 @@ export const {
       return []
     }
 
-    const room = liveblocksClient.getRoom(projectIdToRoomId(projectId))
-    if (room == null) {
-      return []
-    }
-
-    const storage = await room.getStorage()
-
-    const collabs = storage.root.get('collaborators') as LiveObject<{ [userId: string]: User }>
-    if (collabs == null) {
-      return []
-    }
-
-    const users = Object.values(collabs.toObject()).map((u) => u.toObject())
+    const users = await getAllUsersFromRoom(projectIdToRoomId(projectId))
     return users.filter((u) => userIds.includes(u.id))
   },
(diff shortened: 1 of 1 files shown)


editor/liveblocks.config.ts
```diff
@@ -134,42 +134,46 @@ export const {
...
-
-    const storage = await room.getStorage()
-
-    const collabs = storage.root.get('collaborators') as LiveObject<{ [userId: string]: User }>
-    if (collabs == null) {
-      return []
-    }
-
-    const users = Object.values(collabs.toObject()).map((u) => u.toObject())
+    const users = await getAllUsersFromRoom(projectIdToRoomId(projectId))
```
ruggi, line 137: maybe I'm thinking wrongly about this one - but does this need to be updated when the project Id changes? (e.g. fork)

gbalint: getProjectID() always returns the current id, so this should survive a fork (and after a projectId change the room context is rerendered).

balazsbajorics: btw shouldn't getProjectID use the editorState instead of reading off of window?

balazsbajorics (approved):

ruggi (approved):

editor/liveblocks.config.ts
gbalint: I can not access editorState from the liveblocks resolveUsers function, only globals (and it is not a react component or hook, so I can not access the useEditorState hook neither)

gbalint: But in the long run we should implement a backend endpoint to retrieve users, and then this whole resolveUsers hack will become unnecessary.

PR #4533, opened 2023-11-22
feat(editor): support array and image in variables

liady:
**Feature:**
This PR support recognizing Arrays and image urls variables, and inserting correct elements (a map expression or an image element, respectively).

**Known:**
Map expression currently checks if all array elements are of the same type (and uses that type)

### DIFF
--- a/editor/src/components/editor/canvas-toolbar.spec.browser2.tsx
+++ b/editor/src/components/editor/canvas-toolbar.spec.browser2.tsx
@@ -673,7 +673,7 @@ describe('canvas toolbar', () => {
     )
   })
 
-  it.skip('can insert an image via the floating insert menu', async () => {
+  it('can insert an image via the floating insert menu', async () => {
     FOR_TESTS_setNextGeneratedUids(['myImage'])
     const editor = await renderTestEditorWithCode(
       makeTestProjectCodeWithComponentInnards(`
--- a/editor/src/components/editor/canvas-toolbar.spec.browser2.tsx
+++ b/editor/src/components/editor/canvas-toolbar.spec.browser2.tsx
@@ -718,7 +718,7 @@ describe('canvas toolbar', () => {
         data-uid='container'
       >
         <div data-uid='a3d' />
-        <img src={myImage} style={{width: 100, height: 100, top:0, left: 0, position: ''}} data-uid='ele'/>
+        <img src={myImage} style={{width: 100, height: 100, top:0, left: 0, position: 'absolute'}} data-uid='ele'/>
       </div>
     )`),
     )
--- a/editor/src/components/editor/canvas-toolbar.spec.browser2.tsx
+++ b/editor/src/components/editor/canvas-toolbar.spec.browser2.tsx
@@ -623,6 +623,107 @@ describe('canvas toolbar', () => {
     )
   })
 
+  it.skip('can insert an image via the floating insert menu', async () => {
+    FOR_TESTS_setNextGeneratedUids(['myImage'])
+    const editor = await renderTestEditorWithCode(
+      makeTestProjectCodeWithComponentInnards(`
+      const myImage = 'test.png?raw=true'
+      return (
+        <div
+        style={{
+          backgroundColor: '#aaaaaa33',
+          position: 'absolute',
+          left: 57,
+          top: 168,
+          width: 247,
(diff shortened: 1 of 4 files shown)


editor/src/core/shared/project-file-types.ts
```diff
@@ -776,8 +776,20 @@ export type AllVariablesInScope = {
...
+  | 'undefined'
+  | 'object'
+  | 'function'
+  | 'array'
+  | 'image'
+
 export interface Variable {
   name: string
-  type: 'string' | 'number' | 'bigint' | 'boolean' | 'symbol' | 'undefined' | 'object' | 'function'
+  type: VariableUtopiaType
```
balazsbajorics, line 793: I think these types should be moved to `scoped-variables` , as they are only used in there and they feel more like internal representations than API. (not super strictly, but project-file-types.ts is about storing internal-api-level types)

I also don't like the name `VariableUtopiaType`, (again this is not a hard rule more like an observation) we try to not use Utopia to describe "custom" or "special-cased". If you move these into scoped-variables.ts then it becomes easier to name it, because there it can just be called DetectedType or InsertableType, you don't even have to export it, so the name is way less important (doesn't pollute global namespace)

seanparsons (approved):

editor/src/components/shared/scoped-variables.ts
```diff
@@ -131,9 +133,59 @@ function getMatchingElementForVariable(variable: Variable) {
...
+        [],
+      )
     default:
       return jsxElementWithoutUID('span', jsxAttributesFromMap({}), [
         jsExpressionOtherJavaScriptSimple(`JSON.stringify(${variable.name})`, [variable.name]),
       ])
   }
 }
+
+function getTypeByValue(value: any): InsertableType {
```
Rheeseyb, line 156: This is a minor one, but we try our hardest to avoid `any`, preferring to use the `unknown` type instead, since `any` makes it far to easily punch a hole through the type system without realising that you've just done that

editor/src/components/shared/scoped-variables.ts
```diff
@@ -131,9 +133,59 @@ function getMatchingElementForVariable(variable: Variable) {
       return jsxElementWithoutUID('span', jsxAttributesFromMap({}), [
         jsExpressionOtherJavaScriptSimple(`JSON.stringify(${variable.name})`, [variable.name]),
       ])
+    case 'array':
+      return jsxFragmentWithoutUID(
+        [jsxTextBlock(`{${variable.name}.map((item) => <span>{JSON.stringify(item)}</span>)}`)],
```
Rheeseyb, line 138: Nice. I guess the next step is to have this recursively call `getMatchingElementForVariable` then?

editor/src/components/editor/canvas-toolbar.spec.browser2.tsx
```diff
@@ -623,6 +623,107 @@ describe('canvas toolbar', () => {
...
+        data-uid='container'
+      >
+        <div data-uid='a3d' />
+        <React.Fragment>{myArray.map((item) => (<span>{JSON.stringify(item)}</span>))}</React.Fragment>
+      </div>
+    )`),
+    )
+  })
+
+  it.skip('can insert an image via the floating insert menu', async () => {
```
Rheeseyb, line 676: Why the skipped test?

balazsbajorics: oh nice catch!

editor/src/components/shared/scoped-variables.ts
liady: No, since we can't use the actual element here (since it's a text block). I agree it's better to do it recursively, it just means that all of the `switch-case` branches will have to be text (for example - the conditional can't be an element but has to be `${variable.name} ? null : null` as a template string)

editor/src/components/editor/canvas-toolbar.spec.browser2.tsx
liady: Screenshot 2023-11-22 at 2 54 05
Still fighting with it :) (strings look identical but test fails. Tried copying the "actual" to the "expected", no luck)

balazsbajorics (approved):

PR #4556, opened 2023-11-23
Comment dragging

gbalint:
**Problem:**
Comment dragging.

https://github.com/concrete-utopia/utopia/assets/127662/e6c3ff82-5b9e-4e46-8441-fb693e9a2fd1

**Fix:**
- added useDragging hook to implement comment dragging
- new dragged position is updated on liveblocks on mouse up
- added new field to CommentMode, so during dragging we don't show the comment mode cursor


ruggi (approved):

Rheeseyb (approved):

editor/src/components/canvas/controls/comment-indicator.tsx
```diff
@@ -32,98 +41,185 @@ export const CommentIndicator = React.memo(() => {
...
+      window.addEventListener('mousemove', onMouseMove)
+      window.addEventListener('mouseup', onMouseUp)
+    },
+    [onMouseMove, onMouseUp, dispatch],
+  )
+
+  return { onMouseDown, dragPosition }
+}
+
+function windowPositionToCanvasPosition(
```
seanparsons, line 212: This doesn't feel like it should be in here and I'm a bit suspicious of those `-10` values!

seanparsons (approved):

editor/src/components/editor/editor-modes.ts
```diff
@@ -176,10 +177,11 @@ export const EditorModes = {
       selectOnFocus: selectOnFocus,
     }
   },
-  commentMode: function (location: CanvasPoint | null): CommentMode {
+  commentMode: function (location: CanvasPoint | null, isDragging: boolean): CommentMode {
```
balazsbajorics, line 180: nit: isDragging could be of type `'dragging' | 'not-dragging'`

editor/src/components/canvas/controls/comment-indicator.tsx
gbalint: It is not a general function, it is just about offsetting the position so the center of the indicator is at the mouse not the top/left corner. I cleaned this up a bit in fb021103202f6e1734952b497dc1a0c7eacb2026

balazsbajorics (approved):

editor/src/components/editor/editor-modes.ts
gbalint: ok, see 323f73f197942d609b6bc82dd5826be5b741990c

PR #4563, opened 2023-11-24
(Draft) H2 Support

Rheeseyb:


editor/src/components/canvas/remix/utopia-remix-root-component.tsx
```diff
@@ -150,25 +153,45 @@ function useGetRoutes() {
...
-                .scope[routeExport]?.(args) ?? null
+            route[routeExport] = async (args: any) => {
+              const { context: requestContext } = await getContextFromCustomServer(
+                customServerCreator,
+                projectContentsRef.current,
+                fileBlobsRef.current,
+                hiddenInstancesRef.current,
+                displayNoneInstancesRef.current,
+                metadataContext,
+                args.request,
```
balazsbajorics, line 172: these are not route-specific, right? I feel like this should not live in the inner loop, rather we should support an injected Context

editor/src/components/canvas/remix/utopia-remix-root-component.tsx
```diff
@@ -190,6 +213,45 @@ function useGetRoutes() {
   ])
 }
 
+// Super hacky way of calling the `fetch` function from `server.js` purely to get the context
+// that it provides, so that we can then provide that to the server function
+async function getContextFromCustomServer(
```
balazsbajorics, line 218: something about this doesn't sit right with me, but I don't yet have anything better to offer

Rheeseyb: Something about this "super hacky" horror show doesn't sit right with you? 😁 **ALL** of it doesn't sit right with me!

editor/src/components/canvas/remix/utopia-remix-root-component.tsx
```diff
@@ -150,25 +153,45 @@ function useGetRoutes() {
...
+                customServerCreator,
+                projectContentsRef.current,
+                fileBlobsRef.current,
+                hiddenInstancesRef.current,
+                displayNoneInstancesRef.current,
+                metadataContext,
+                args.request,
+              )
+
+              const patchedArgs = {
```
balazsbajorics, line 175: I found `patchRoutesWithContext` in @remix/testing and they only patch loader and action 

https://github.com/remix-run/remix/blob/4a8f558d16b8739d7f70903958ac5792e050f3e9/packages/remix-testing/create-remix-stub.tsx#L55

PR #4565, opened 2023-11-24
Handle imports and exports correctly in collaboration.

seanparsons:
**Problem:**
Export details are used as part of the flow for printing components, so that they get the `export` modifier against them. Without that the viewer in a collaboration setting might not have a newly added component exported which would result in the project breaking for them.

Additionally while fixing this issue, the imports were found to be suffering from a related issue. Which could result in the following code when changing the exports and related imports on the other side:
```typescript
import { Playground } from '/src/playground'
import Playground from '/src/playground'
```
The above code would then cause the viewer to be broken from the duplicate reference.

**Fix:**
This consists of a couple of intertwined fixes:
- Carrying over the imports and export details as part of the collaboration updates.
- Fixing the handling of changes from yjs types to our types by getting the yjs types to construct our arrays or maps through `.toArray()` or `.toJSON()` respectively. This removes an entire segment of logic that was using deltas from yjs to reconstruct our types.
- Completely reworking our delta updates from our type to the yjs types as they weren't working correctly in extreme cases like completely deleting the entire contents of a file.

**Commit Details:**
- Added `UPDATE_EXPORTS_DETAIL_FROM_COLLABORATION_UPDATE` and `UPDATE_IMPORTS_FROM_COLLABORATION_UPDATE` editor actions.
- `addHookForProjectChanges` now collates potentially many actions to dispatch at once.
- Reworked `ArrayChanges` as it didn't work very well in its original form.
- Refactored `updateEntireProjectContents` to produce a multitude of editor actions and dispatch all at once.
- Created `updateEditorWithArrayChanges` and `updateEditorWithMapChanges` utility functions.
- Implemented `updateTopLevelElementsOfFile`, `updateExportsDetailOfFile` and `updateImportsOfFile` to carry changes from the yjs types into the editor.
- `calculateArrayChanges` now produces step by step changes.
- Implemented `syncArrayChanges` and `syncMapChanges` utility functions.
- `synchroniseParseSuccesToCollabFile` now much more simplified than the original implementations.

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -315,6 +315,8 @@ import type {
   ScrollToPosition,
   UpdateTopLevelElementsFromCollaborationUpdate,
   DeleteFileFromCollaboration,
+  UpdateExportsDetailFromCollaborationUpdate,
+  UpdateImportsFromCollaborationUpdate,
 } from '../action-types'
 import { isLoggedIn } from '../action-types'
 import type { Mode } from '../editor-modes'
@@ -424,6 +426,8 @@ import {
 } from '../../../core/vscode/vscode-bridge'
 import { createClipboardDataFromSelection, Clipboard } from '../../../utils/clipboard'
 import {
+  ExportDetailKeepDeepEquality,
+  ImportDetailsKeepDeepEquality,
   NavigatorStateKeepDeepEquality,
   TopLevelElementKeepDeepEquality,
 } from '../store/store-deep-equality-instances'
@@ -543,7 +547,7 @@ import { resultForFirstApplicableStrategy } from '../../inspector/inspector-stra
 import { reparentToUnwrapAsAbsoluteStrategy } from '../one-shot-unwrap-strategies/reparent-to-unwrap-as-absolute-strategy'
 import { convertToAbsoluteAndReparentToUnwrapStrategy } from '../one-shot-unwrap-strategies/convert-to-absolute-and-reparent-to-unwrap'
 import { addHookForProjectChanges } from '../store/collaborative-editing'
-import { arrayDeepEquality } from '../../../utils/deep-equality'
+import { arrayDeepEquality, objectDeepEquality } from '../../../utils/deep-equality'
 import type { ProjectServerState } from '../store/project-server-state'
 import { fixParseSuccessUIDs } from '../../../core/workers/parser-printer/uid-fix'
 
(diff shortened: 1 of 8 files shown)


bkrmendy (approved):

ruggi (approved):

liady (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -5395,6 +5399,124 @@ export const UPDATE_FNS = {
...
+      }
+    }
+  },
+  UPDATE_EXPORTS_DETAIL_FROM_COLLABORATION_UPDATE: (
+    action: UpdateExportsDetailFromCollaborationUpdate,
+    editor: EditorModel,
+    serverState: ProjectServerState,
+  ): EditorModel => {
+    // When the current user does own the project...
+    if (serverState.isMyProject === 'yes') {
```
Rheeseyb, line 5416: We're starting to see this same check littered about the code, and it's making me uneasy because in some cases (like here) this will be incompatible with the concept of shared ownership of a project. I'm wondering if it makes sense to have a function (or series of functions) that provide checks around whether a certain type of edit can be made to the project, which would themselves right now check `serverState.isMyProject`, even if all of those functions are then just performing this exact check right now

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -211,129 +238,278 @@ export function addHookForProjectChanges(
         // the string `topLevelElements`.
```
Rheeseyb, line 238: I guess this comment needs updating now

editor/src/components/editor/actions/actions.tsx
seanparsons: We could have something similar to `isTransientAction` which gates the applicability of actions.

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -211,129 +238,278 @@ export function addHookForProjectChanges(
...
-      readIndex += delta.retain
-    }
-    if (delta.insert != null && Array.isArray(delta.insert)) {
-      newTopLevelElements.push(...delta.insert)
-    }
-    if (delta.delete != undefined) {
-      readIndex += delta.delete
-    }
-  }
+  const yjsValue: Y.Array<T> = (file?.get(fileKey) as any as Y.Array<T>) ?? new Y.Array()
```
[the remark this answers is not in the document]
seanparsons: Yeah, behold it's awful typing and weep.

Rheeseyb (commented):
Nice work 👍

editor/src/components/editor/actions/actions.tsx
Rheeseyb: Yeah I think that would make sense because at least then the logic would all be captured in a single place. I think we'll still have some awkwardness to deal with when we have to handle whether the sender of the change was the owner or not, but we can burn that bridge when we come to it

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -211,129 +238,278 @@ export function addHookForProjectChanges(
...
+  fromArray: Array<T>,
+  collabFile: CollabTextFile,
+  fileKey: string,
+  keepDeep: KeepDeepEqualityCall<T>,
+): void {
+  const againstArray = collabFile.get(fileKey) as any as Y.Array<T>
+  const elementChanges = calculateArrayChanges(
+    fromArray,
+    againstArray,
+    (l, r) => l != null && r != null && keepDeep(l, r).areEqual,
```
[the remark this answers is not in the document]
seanparsons: I think the better thing to do is to bin those off completely and ensure it only gets called with non-null values. Currently it's never called with both of those values being null by dint of one of those values coming from an array where the index is within the range of the array.

PR #4566, opened 2023-11-24
Strip source maps from the outgoing top level elements.

seanparsons:
**Problem:**
The messages that we send to Liveblocks for the collaboration on realistic projects tends to cause disconnections as each message can exceed 5MB.

**Cause:**
Most of the bloat comes from the source maps that are extraneous to the needs of the collaboration functionality.

**Fix:**
Utilising the replacer parameter of `JSON.stringify`, we replace the source maps content with a null.

**Commit Details:**
- Added `removeSourceMaps` to strip those from the outgoing top level elements.

### DIFF
--- a/editor/src/components/editor/store/collaborative-editing.ts
+++ b/editor/src/components/editor/store/collaborative-editing.ts
@@ -50,6 +50,11 @@ const TopLevelElementsKey = 'topLevelElements'
 const ExportsDetailKey = 'exportsDetail'
 const ImportsKey = 'imports'
 
+// FIXME: This is very slow an inefficient, but is a stopgap measure for right now.
+function removeSourceMaps(topLevelElements: Array<TopLevelElement>): Array<TopLevelElement> {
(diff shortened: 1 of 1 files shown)


balazsbajorics (approved):

ruggi (approved):

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -50,6 +50,11 @@ const TopLevelElementsKey = 'topLevelElements'
 const ExportsDetailKey = 'exportsDetail'
 const ImportsKey = 'imports'
 
+// FIXME: This is very slow an inefficient, but is a stopgap measure for right now.
+function removeSourceMaps(topLevelElements: Array<TopLevelElement>): Array<TopLevelElement> {
+  return JSON.parse(JSON.stringify(topLevelElements, (k, v) => (k === 'sourceMap' ? null : v)))
```
liady, line 55: nice hack though :)

Rheeseyb (commented):
👍

PR #4569, opened 2023-11-27
Variables Menu - Right Pane

liady:
Add a Variables menu pane to the right side of the screen. Currently only one-shot insertion (one click and not drag&draw)

These definitions:
image

Will result in this menu when an element from the playground is selected:
image

And when an element is not selected (or the variables are not in scope):
image

### DIFF
--- a/editor/src/components/shared/scoped-variables.ts
+++ b/editor/src/components/shared/scoped-variables.ts
@@ -139,7 +139,7 @@ function getMatchingElementForVariableInner(variable: Variable): InsertableCompo
     case 'boolean':
       return insertableComponentAndJSX(
         jsxConditionalExpressionWithoutUID(
-          jsIdentifierName(variable.name),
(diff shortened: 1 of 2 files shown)


Rheeseyb:
Can you add a screenshot to this PR please

bkrmendy (approved):

editor/src/components/shared/scoped-variables.ts
```diff
@@ -220,7 +230,7 @@ interface InsertableComponentAndJSX {
   jsx: string
 }
 
-type AllVariablesInScope = {
+export type AllVariablesInScope = {
```
gbalint, line 233: Why is this exported? I don't see any added usage in the diff

gbalint: Sorry, found it!

gbalint (approved):

PR #4570, opened 2023-11-27
Fix multiplayer remix follow and navigate to remix route when opening comments

gbalint:
**Problem:**
- Remix route following is flaky, sometimes works, sometimes doesn't work.
- When we open a comment (in the inspector or on the canvas), we don't navigate to the route with the comment.

**Fix:**
1. Instead of replacing the whole RemixNavigationState, expose the navigate function of the remix router on the RemixNavigationContext, and use that to navigate.
2. Add navigation code to CommentIndicator and CommentSection. Since we don't have scene indentifiers in the comment metadata (yet), we just navigate all remix scenes to the given route.

### DIFF
--- a/editor/src/components/canvas/controls/comment-indicator.tsx
+++ b/editor/src/components/canvas/controls/comment-indicator.tsx
@@ -25,6 +25,9 @@ import { MultiplayerWrapper } from '../../../utils/multiplayer-wrapper'
 import { AvatarPicture } from '../../user-bar'
 import type { ThreadData } from '@liveblocks/client'
 import { canvasPointToWindowPoint, windowToCanvasCoordinates } from '../dom-lookup'
+import { useAtom } from 'jotai'
+import { RemixNavigationAtom } from '../remix/utopia-remix-root-component'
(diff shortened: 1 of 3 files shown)


bkrmendy (approved):

editor/src/components/canvas/multiplayer-presence.tsx
```diff
@@ -101,7 +101,7 @@ export const MultiplayerPresence = React.memo(() => {
 
   React.useEffect(() => {
     // when the room changes, reset
-    dispatch([switchEditorMode(EditorModes.selectMode(null, false, 'none'))])
+    setTimeout(() => dispatch([switchEditorMode(EditorModes.selectMode(null, false, 'none'))]), 0)
```
ruggi, line 104: wowza 😬

ruggi (approved):

editor/src/components/canvas/multiplayer-presence.tsx
gbalint: oh yes, this is an unrelated fix to remove the flushSync errors

PR #4588, opened 2023-12-01
Scene comments

gbalint:
**Description:**
New comment type: scene comments!
- Scenes have a new optional prop `id`, which is used to identify a scene in the comment metadata. I made it optional so I don't break all the existing utopia projects. Later we can implement id generation when it is missing.
- Comment metadata has a new optional property `sceneId`. When this is not undefined, the comment is a scene comment, and its x and y coordinates are local coordinates in the scene.
- Instead of storing a CanvasPoint in NewComment.location, I created a new ADT `NewCommentLocation`, which is either a global location on the canvas, or a scene id and a local position in that scene.
- I updated comment mode so it inserts a scene comment when you are clicking over a scene.
- I updated comment indicator and comment popup to properly handle the position of the two kind of comments (scene and canvas)
- I updated the default project, so the scenes in them have ids
- When the scene is not found for a comment (e.g. it is deleted), then the comment is just shown on the same canvas position as its local position in the scene (maybe later we should store the scene coordinates in the comment metadata for safety).

**Missing things:**
- dragging a scene comment outside of the scene still keeps it as a comment of the original scene (no "comment reparenting")
- scenes on top of each others are not handled, z-order of scenes is not taken into account
- no UI improvements yet, you can't see if a comment is a scene comment or a canvas comment anywhere (the only difference is that scene comments move together with a dragged scene).
- scene ids are not generated, so if you don't add an id to your scenes, you can't add comments to it
- even if you insert a scene using the utopia UI, no id will be added (which is fine in my opinion, since the UI for scene comments is basically missing)

### DIFF
--- a/editor/liveblocks.config.ts
+++ b/editor/liveblocks.config.ts
@@ -82,7 +82,7 @@ export type ThreadMetadata = {
   // quote: string;
   // time: number;
   type: 'canvas'
-  x: number
+  x: number // x and y is global when sceneId is undefined, and local to the scene when sceneId is not null
   y: number
   sceneId?: string
   remixLocationRoute?: string
--- a/editor/liveblocks.config.ts
+++ b/editor/liveblocks.config.ts
@@ -84,6 +84,7 @@ export type ThreadMetadata = {
   type: 'canvas'
   x: number
   y: number
+  sceneId?: string
   remixLocationRoute?: string
 }
 
--- a/editor/src/components/canvas/controls/comment-mode/comment-mode-hooks.tsx
+++ b/editor/src/components/canvas/controls/comment-mode/comment-mode-hooks.tsx
@@ -15,6 +15,7 @@ import { Substores, useEditorState, useRefEditorState } from '../../../editor/st
 import { windowToCanvasCoordinates } from '../../dom-lookup'
 import {
   canvasPoint,
+  getLocalPointInNewParentContext,
   isNotNullFiniteRectangle,
   offsetPoint,
   pointDifference,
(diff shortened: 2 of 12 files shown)


bkrmendy (approved):

editor/liveblocks.config.ts
```diff
@@ -82,8 +82,9 @@ export type ThreadMetadata = {
   // quote: string;
   // time: number;
   type: 'canvas'
-  x: number
+  x: number // x and y is global when sceneId is undefined, and local to the scene when sceneId is not null
```
ruggi, line 85: the limitations of allowed types in the metadata object 😢

editor/src/components/canvas/controls/comment-mode/comment-mode-hooks.tsx
```diff
@@ -28,16 +50,47 @@ export function useCommentModeSelectAndHover(comment: CommentId | null): MouseCa
...
+
+        const scenesUnderTheMouse = scenes.filter((scene) => {
+          const sceneId = getIdOfScene(scene)
+          return (
+            sceneId != null &&
+            isNotNullFiniteRectangle(scene.globalFrame) &&
+            rectContainsPoint(scene.globalFrame, loc.canvasPositionRaw)
+          )
+        })
+        const scene = scenesUnderTheMouse[0]
```
ruggi, line 62: it would be good to safely check for index 0

ruggi (approved):

editor/src/components/canvas/controls/comment-mode/comment-mode-hooks.tsx
gbalint: done

PR #4601, opened 2023-12-05
Cleanup of gated actions for collaboration.

seanparsons:
**Problem:**
The codebase ended up with a variety of disparate locations for the checks that determine if an action should be run that were introduced for the collaborative editing work. Which means that it's easy to not change one of the many places when making further changes.

**Fix:**
All of the logic has now been moved to a single function `gatedActions` which is called in a single place: `processAction`.

**Commit Details:**
- Removed previous checks of `isMyProject` in the various collaboration actions.
- Removed `checkIfActionShouldBeProcessed`.
- Implemented all of the logic for handling if the actions should run into `gatedActions`.
- Incorporated `gatedActions` into `processAction`.


ruggi (approved):
love it

PR #4602, opened 2023-12-05
show remix route in comment pane

bkrmendy:
image

## Problem
The Remix route corresponding to a comment is not shown in the comment pane

## Fix
Add it to the thread preview


editor/src/components/inspector/sections/comment-section.tsx
```diff
@@ -220,3 +244,15 @@ const ThreadPreview = React.memo(({ thread }: ThreadPreviewProps) => {
   )
 })
 ThreadPreview.displayName = 'ThreadPreview'
+
+function getRemixLocationRoute(location: string | undefined): string | null {
```
gbalint, line 248: We have the same label in the toolbar, can we share the code between the two?

PR #4611, opened 2023-12-06
Don't re-scope the arguments of almost all CSS pseudo-classes

Rheeseyb:
**Problem:**
Follow on from #4604 
Some CSS pseudo-classes should have their arguments re-scoped, some should not.

**Fix:**
Prevent traversal of all CSS pseudo-class arguments, with the explicit exception of `:is` and `:where`, using css-tree's `walk.skip`. This means that we **would** rescope the arguments of `:is(h1)` to `:is(#canvas-container h1)`, but not the arguments of `:has(h1)`, or the arguments of a pseudo-class nested inside `:has(:is(h1))`


seanparsons (approved):

Rheeseyb:
Closing in favour of #4613

PR #4614, opened 2023-12-07
Store read/unread status in comment metadata

gbalint:
**Problem:**
We would like to display and use the read/unread status of a comment thread, but that is not natively supported by Liveblocks.

**Fix:**
Store the read/unread status in the room storage. Since this is global for all users and threads, we need to store it per thread per user.

- Introduce `userReadStatusesByThread` in the room storage.
- Set my user to read when I open the comment popup.
- Set only my user to read when I create a thread
- Set only my user to read when I submit a comment to a thread (so all the others will be unread again)
- Add a 'Mark as unread' button to the comment popup, to set my status back to unread.
- Show when a comment is unread in the comment inspector.
- Make the comment indicator partially transparent when the comment is already read.
- Delete the object for the thread in storage when the thread is deleted


editor/liveblocks.config.ts
```diff
@@ -86,6 +86,36 @@ export type ThreadMetadata = {
...
+  return (
+    statuses != null &&
+    typeof statuses === 'object' &&
+    Object.keys(statuses).every((v) => typeof v === 'string') &&
+    Object.values(statuses).every((v) => typeof v === 'boolean')
+  )
+}
+
+export function parseUserReadStatuses(userReadStatuses: string): UserReadStatuses {
+  try {
```
seanparsons, line 104: We have a simple parser infrastructure in `value-parser-utils.ts` that you could use in this which does all the heavy lifting.

bkrmendy (approved):

editor/src/core/commenting/comment-hooks.tsx
```diff
@@ -261,3 +261,79 @@ export function useUnresolvedThreads() {
...
+  const self = useSelf()
+  return useStorage((store) => {
+    if (thread == null) {
+      return 'unread'
+    }
+    const statusesForThread = store.userReadStatusesByThread[thread.id]
+    if (statusesForThread == null) {
+      return 'unread'
+    }
+    return statusesForThread[self.id] === true ? 'read' : 'unread'
```
Rheeseyb, line 288: If we're never setting this to `false` then we should probably just always return `'read'` at this point, since really what we're doing here is using the object to store a flag per user that has read the thread. If we start using `false` to represent that the user hasn't read the thread then that's added complexity for no gain

editor/src/core/commenting/comment-hooks.tsx
```diff
@@ -261,3 +261,79 @@ export function useUnresolvedThreads() {
...
+  })
+}
+
+export type ThreadReadStatus = 'read' | 'unread'
+
+export function useSetThreadReadStatus() {
```
Rheeseyb, line 294: This should be named `useToggleThreadReadStatus`

Rheeseyb (approved):

editor/src/core/commenting/comment-hooks.tsx
gbalint: Why toggle? Toggle means switching to the opposite value, but this returns a function with which you can set to any given value.

Rheeseyb: I think my reading comprehension has failed me today

editor/src/components/canvas/controls/comment-indicator.tsx
```diff
@@ -187,12 +188,16 @@ interface CommentIndicatorUIProps {
...
-      ? 0.25
-      : assertNever(props.opacity)
+  const opacity = (() => {
+    switch (props.opacity) {
+      case 'opaque':
+        return 1
+      case 'transparent':
+        return 0.25
+      case 'half-opaque':
+        return 0.75
```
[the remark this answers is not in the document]
gbalint: I got rid of these names for values in 9b9cce0ac3bf6f0144468d54c46b1a102918c9cd

PR #4616, opened 2023-12-07
Recognize props passed to a component

liady:
**Problem:**
Currently we don't show props in the "Variables" panel or the top bar.

**Fix:**
Identify which props were passed to the container and add them, allow inserting them.

## If an element is selected inside "Playground" (which defines some props)
image

## And `<Playground>` is being passed some real time values
image

## We know to recognize those props are in the scope of the selected element and display them
image


editor/src/components/shared/scoped-variables.ts
```diff
@@ -83,26 +99,79 @@ function getVariablesFromComponent(
...
+    variables: generateVariableTypes(jsxComponentVariables),
+  }
+}
+
+function getComponentPropsInScope(
+  topLevelElements: TopLevelElement[],
+  elementPath: ElementPath,
+  jsxMetadata: ElementInstanceMetadataMap,
+): AllVariablesInScope {
+  const jsxComponent = topLevelElements.find((el) => isUtopiaJSXComponent(el)) as UtopiaJSXComponent
```
gbalint, line 111: If you define the return value of the inline function, then you don't need to cast the type:
```typescript
const jsxComponent = topLevelElements.find((el): el is UtopiaJSXComponent =>
    isUtopiaJSXComponent(el),
  )
```

But in this case it is even simpler to directly pass in `isUtopiaJSXComponent`
```typescript
const jsxComponent = topLevelElements.find(isUtopiaJSXComponent)
```

editor/src/components/shared/scoped-variables.ts
```diff
@@ -83,26 +99,79 @@ function getVariablesFromComponent(
...
+    variables: generateVariableTypes(jsxComponentProps),
+  }
+}
+
+function getContainerPropsValue(
+  elementPath: ElementPath,
+  jsxMetadata: ElementInstanceMetadataMap,
+): VariableData {
+  const containerPathString = toComponentId(getContainingComponent(elementPath))
+  const containerMetadata = jsxMetadata[containerPathString]?.element.value
```
gbalint, line 131: I think it is nicer to use `isLeft`/`isRight`/`foldEither` or other `either` helpers than working with the raw value and checking the type of that with `typeof` later.

editor/src/components/shared/scoped-variables.ts
```diff
@@ -83,26 +99,79 @@ function getVariablesFromComponent(
...
+  topLevelElements: TopLevelElement[],
+  elementPath: ElementPath,
+  jsxMetadata: ElementInstanceMetadataMap,
+): AllVariablesInScope {
+  const jsxComponent = topLevelElements.find((el) => isUtopiaJSXComponent(el)) as UtopiaJSXComponent
+  const jsxComponentPropNamesUsed = jsxComponent?.propsUsed ?? []
+
+  const jsxComponentProps = omitWithPredicate(
+    getContainerPropsValue(elementPath, jsxMetadata),
+    (key) => !jsxComponentPropNamesUsed.includes(key as string),
```
seanparsons, line 116: Do we know this value is for sure a `string`? If not I'd add a check for that to the predicate.

editor/src/components/shared/scoped-variables.ts
```diff
@@ -83,26 +99,79 @@ function getVariablesFromComponent(
...
+
+  if (
+    containerMetadata != null &&
+    typeof containerMetadata !== 'string' &&
+    isJSXElement(containerMetadata)
+  ) {
+    containerMetadata.props.forEach((prop) => {
+      if (isJSXAttributesEntry(prop)) {
+        const value = simplifyAttributeIfPossible(prop.value)
+        if (!isJSXAttributeValue(value)) return
```
seanparsons, line 142: Is it not simpler to just enclose the next line in the conditional but flipped?
```suggestion
        if (isJSXAttributeValue(value)) {
```

seanparsons (approved):

editor/src/components/canvas/ui/floating-insert-menu.tsx
```diff
@@ -56,14 +56,13 @@ function convertInsertableComponentsToFlatList(
       options: componentGroup.insertableComponents.map(
         (componentToBeInserted, index): InsertMenuItem => {
           const source = index === 0 ? componentGroup.source : null
+          const label = componentToBeInserted.metadata?.originalName ?? componentToBeInserted.name
           return {
-            label: componentToBeInserted.name,
+            label,
```
Rheeseyb, line 61: This is very minor, but we explicitly avoid the shorthand notation here, instead opting for `label: label`. We tend to favour an explicit coding style over an implicit one in general to minimise the risk of misunderstanding when someone unfamiliar with the code is reading it.

editor/src/components/editor/variablesmenu.spec.browser2.tsx
```diff
@@ -106,6 +112,59 @@ describe('variables menu', () => {
...
+      document.execCommand('insertText', false, 'imgProp')
+      expect(getInsertItems().length).toEqual(1)
+      expect(getInsertItems()[0].innerText).toEqual('imgProp')
+
+      const filterBox = await screen.findByTestId(InsertMenuFilterTestId)
+      forceNotNull('the filter box must not be null', filterBox)
+
+      await act(async () => {
+        fireEvent.keyDown(filterBox, { key: 'Enter', keycode: 13 })
+      })
```
Rheeseyb, line 137: We have a bunch of helper functions for firing events in `event-helpers.test-utils.tsx`, including one for `pressKey`: https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/event-helpers.test-utils.tsx#L749 - If that function isn't suitable as it stands, please do tweak it to support this use case, since we want to avoid manually firing the events like this as those helpers exist to ensure that we can better simulate a real interaction by calling the combination of events that would be fired in the right order

editor/src/components/editor/variablesmenu.tsx
```diff
@@ -92,6 +97,17 @@ const Input = (props: InputProps) => {
   return <components.Input {...props} data-testid={VariablesMenuFilterTestId} />
 }
 
+const BASE_PADDING = 4
+const DEPTH_PADDING = 16
+function paddingByDepth(insertMenuItem: InsertMenuItem) {
+  const depth = insertMenuItem.value.metadata?.depth as number | null
```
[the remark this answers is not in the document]
Rheeseyb: Also, as a side note, we've ended up using the term "metadata" to refer to the metadata of an actual JSX Element, i.e. this type https://github.com/concrete-utopia/utopia/blob/851740fba86781fe1848509e152f57ecde44d62b/editor/src/core/shared/element-template.ts#L1982 As that is so common in this project, so please avoid using that term in the code to refer to anything else without explicitly calling out that it is referring to a different form of metadata

Rheeseyb (changes requested):
Please can you introduce a new type `InsertableVariable` (as mentioned in one of my comments) so that we can rely on the types to know that insertable variables have the required insert menu information (depth and original name) rather than having to rely on our knowledge about what should have been put in the `Record` (and then having to type check and cast that)

editor/src/components/shared/scoped-variables.ts
liady: Thanks!

editor/src/components/shared/scoped-variables.ts
liady: Good catch, it was actually like that and it was a result of a bad refactor. Thanks!

editor/src/components/editor/variablesmenu.tsx
```diff
@@ -92,8 +113,19 @@ const Input = (props: InputProps) => {
   return <components.Input {...props} data-testid={VariablesMenuFilterTestId} />
 }
 
-const iconByType = (insertMenuItem: InsertMenuItem): string => {
-  const variableType = insertMenuItem.value.metadata?.variableType as string
+const BASE_PADDING = 4
+const DEPTH_PADDING = 16
+function paddingByDepth(insertMenuItem: InsertMenuVariableItem) {
+  const depth = insertMenuItem.value.depth
+  const depthValue = depth == null ? 0 : depth
```
Rheeseyb, line 120: I believe `depth` should no longer be nullable

editor/src/components/editor/variablesmenu.tsx
```diff
@@ -123,19 +155,19 @@ const Option = React.memo((props: OptionProps<ComponentOptionItem, false>) => {
         rowHeight={'smaller'}
         css={{
           borderRadius: 2,
-          padding: 4,
           color: isHovered ? colorTheme.dynamicBlue.value : colorTheme.fg1.value,
           background: undefined,
           gap: 4,
           border: '1px solid transparent',
+          ...paddingByDepth(props.data as InsertMenuVariableItem),
```
Rheeseyb, line 162: We shouldn't need the explicit casting here

Rheeseyb (approved):

PR #4623, opened 2023-12-08
Replace file protocol with links in packages installed using pnpm

Rheeseyb:
**Problem:**
We have been using the `file:` protocol for local dependencies, but it in [v7.0.0](https://github.com/pnpm/pnpm/blob/main/pnpm/CHANGELOG.md#700) pnpm changed the handling of those to use hard links rather than symlinks. This has meant that since we last updated pnpm in April, any time we've tried making changes to either `utopia-api` or `utopia-vscode-common` those changes haven't automatically been reflected locally.

**Fix:**
Switch to using the `link:` protocol for those dependencies. Note that we're still using `file:` inside the `vscode-build` project, but that is built using yarn, and so we have to leave that one.

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer-comments.spec.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer-comments.spec.ts
@@ -168,11 +168,11 @@ describe('Parsing and printing code with comments', () => {
             }
             someProp2={{
               /* Comment before object key */ someKey /* Comment after object key */:
-                /* Comment before object value */ 'someValue' /* Comment after object separator */ /* Comment after object value */,
+                /* Comment before object value */ 'someValue' /* Comment after object value */ /* Comment after object separator */,
               someKey2: 'someValue2',
             }}
             someProp3={[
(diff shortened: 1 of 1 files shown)


seanparsons (approved):

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -910,7 +910,7 @@ function printCodeImpl(
               missingImports.map((i) => {
                 return TS.createImportSpecifier(
                   false,
-                  TS.createIdentifier(i.name),
+                  i.name === i.alias ? undefined : TS.createIdentifier(i.name),
```
Rheeseyb, line 913: What's this?

seanparsons: This fixes the issue where you end up with `import { X as X } from 'x'`, which it would appear Prettier previously collapsed.

Rheeseyb: Riiiight ok thanks

PR #4626, opened 2023-12-11
add text elements as 'intrinsicHTMLElementNamesThatSupportChildren'

liady:
**Problem:**
Text elements other than `span` and `div` cannot be edited using our text editor

**Fix:**
Add all text elements to `TextElements` array, and check if the tag name exists there as well.

**Reproduce:**
Add a `<p>test</p>` element to the canvas and try to edit it with a double click (or the 't' shortcut)


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -175,6 +176,8 @@ export const getChildrenOfCollapsedViews = (
   }, collapsedViews)
 }
 
+// TODO should this also be changed to simply TextElements?
+// (to include all possible text elements in the name lookup)
```
Rheeseyb, line 180: Good question. This is used to determine what to show in the Navigator, and is only used if the element has _only_ text as a child, so I'd say this should include all of `TextElements` plus `div` at the very least. @maltenuhn Should we be using all of `PossibleTextElements` here?

liady: We can also do `['div', 'span'].concat(TextElements)`

seanparsons (approved):

PR #4639, opened 2023-12-12
RemixScene getLoadContext

balazsbajorics:
**Problem:**
Hydrogen projects sneak a context into Remix using `getLoadContext` in the remix adapter's `createRequestHandler` [link](https://remix.run/docs/en/main/other-api/adapter#createrequesthandler). 

When trying to load a hydrogen project in Utopia, our entry point is _not_ the runtime Oxygen entry point located in `server.js`. Our entry point is the `RemixScene` in the storyboard.js file. So this context is entirely missing, and it makes the loaders and the routes sad.

Remix's own made-for-testing `createRemixStub` function takes an optional context ([link](https://remix.run/docs/en/main/utils/create-remix-stub#:~:text=If%20your%20loaders%20rely%20on%20the%20getLoadContext%20method%2C%20you%20can%20provide%20a%20stubbed%20context%20via%20the%20second%20parameter%20to%20createRemixStub)).

However! the Remix testing API takes Context as an object ([link](https://github.com/remix-run/remix/blob/018d5da59363980ebe53667339454e4bb347f6c3/packages/remix-testing/create-remix-stub.tsx#L93)), which is inconvenient for us, because the default Hydrogen project uses the incoming `request` to create the context. To be able to create the same context for the storyboard, we need access to the `request` at the context creation. 

**Fix:**
The solution is a sort of hybrid of createRemixStub and createRequestHandler:
Our `RemixScene` component takes an optional parameter `getLoadContext` which has a similar signature as the one passed in to `createRequestHandler`:
`getLoadContext?: (request: Request) => Promise<AppLoadContext> | AppLoadContext`

**Warning!** Unfortunately the signature of `getLoadContext` depends on the Remix adapter. Depending on the Remix adapter, getLoadContext may receive a `(req, res)` or an `(event)` or a `(request)`. 

This PR matches the Remix Hydrogen adapter ([link](https://github.com/Shopify/hydrogen/blob/dc56d296c5abf572c19046756ad5b27f8b98a7b3/packages/remix-oxygen/src/server.ts#L23)) regardless of which adapter the user's project uses. 
If this becomes untenable, we will have two options:
1. (more likely) rename the callback and keep the fixed signature. I suspect _most_ code using getLoadContext only cares about the request, but this is an unverified assumption.
2. (less likely) somehow utilize the user project's remix adapter. this _definitely_ will need some re-thinking of what we are doing here, we'll have to ask for feedback from the Remix team.

**Commit Details:** 
- `RemixScene` gets optional `getLoadContext` callback
- `patchRoutesWithContext` lifted from remix/testing, but modified to call the `getLoadContext` instead of getting context as an object 
- `UtopiaRemixRootComponent`'s `useGetRoutes` uses `patchRoutesWithContext`
- Tests!

### DIFF
--- a/editor/src/components/canvas/remix/remix-rendering.spec.browser2.tsx
+++ b/editor/src/components/canvas/remix/remix-rendering.spec.browser2.tsx
@@ -798,11 +798,76 @@ describe('Remix content', () => {
 })
 
 describe('Remix getLoadContext', () => {
-  it('Renders the remix project that relies on loader context', async () => {
+  it('Renders the remix project that relies on async loader context', async () => {
     const project = createModifiedProject({
       [StoryboardFilePath]: `import * as React from 'react'
       import { RemixScene, Storyboard } from 'utopia-api'
       const getLoadContext = async (request) => {
+        return {
+          request: request,
+          otherStuff: await Promise.resolve('hello!')
+        }
+      }
+      
+      export var storyboard = (
+        <Storyboard>
+          <RemixScene
+            style={{
+              width: 700,
+              height: 759,
+              position: 'absolute',
+              left: 212,
+              top: 128,
+            }}
+            data-label='Playground'
+            getLoadContext={getLoadContext}
+          />
+        </Storyboard>
+      )
+      `,
+      ['/app/root.js']: `import React from 'react'
+      import { Outlet } from '@remix-run/react'
+      import { json, useLoaderData } from 'react-router'
+      export function loader({params, request, context}) {
+        console.log('loader called', params, request, context)
+        return json({
+          contextEqualsRequest: context.request === request,
+        })
+      }
+      
+      export default function Root() {
+        const loaderData = useLoaderData()
+        return (
+          <div>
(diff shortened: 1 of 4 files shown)


bkrmendy (approved):

editor/src/components/canvas/remix/remix-rendering.spec.browser2.tsx
```diff
@@ -797,6 +797,138 @@ describe('Remix content', () => {
...
 })
 
+describe('Remix getLoadContext', () => {
+  it('Renders the remix project that relies on async loader context', async () => {
+    const project = createModifiedProject({
+      [StoryboardFilePath]: `import * as React from 'react'
+      import { RemixScene, Storyboard } from 'utopia-api'
+      const getLoadContext = async (request) => {
+        return {
+          request: request,
```
Rheeseyb, line 807: Are you putting the request in there just purely for testing purposes?

Rheeseyb (approved):

PR #4644, opened 2023-12-12
Session based VS Code storage.

seanparsons:
**Problem:**
If multiple tabs are open on the same computer then changes via VS Code may oddly propagate between those tabs.

**Cause:**
Currently our VS Code integration uses the project ID to create the storage that is used to communicate with VS Code as well as the filesystem for VS Code. This means that multiple tabs with the same project open can write data to each other in unexpected ways.

**Fix:**
The main fix for the above issue is to have a session based unique ID instead of something based on the project ID as the basis for the name of the IndexedDB storage which is used by VS Code and the editor. Previously we had a unique session based one as a placeholder and then once the project ID was set (crucially not necessarily when it was loaded) we would then reload VS Code with one based on the project ID.

Along with this two longstanding issues have now been fixed:
- Forever loading VS Code, this was due to a race condition between the message listeners ready message and the init project message. If the init project message was fired before the handler was created to wait for it then it would just get missed and then VS Code would never "finish" loading as a result.
- Empty VS Code, this was due to the `SET_PROJECT_ID` action firing before the `LOAD` action which ends up sending an empty project contents to VS Code. With the logic around changes to VS Code's content not being aware that nothing has been sent.

**Commit Details:**
- `VSCodeIframeContainer` now has a property `vsCodeSessionID` and assigns that to a query parameter `vs_code_session_id`.
- Removed `EditorState.vscodeBridgeId` and it's associated type.
- Simplified `SET_PROJECT_ID` action.
- `getUnderlyingVSCodeBridgeID` now returns a value which is assigned to a unique value on startup.
- `initVSCodeBridge` now only permits a single instance of the message listeners ready message to be handled.
- `setupVSCodeEventListenersForProject` repeatedly fires the message listeners ready message until it gets an init project message in return.
- `build-vscode` now clears the `./dist` and `./node_modules` folders.

### DIFF
--- a/editor/src/core/vscode/vscode-bridge.ts
+++ b/editor/src/core/vscode/vscode-bridge.ts
@@ -95,12 +95,18 @@ export function initVSCodeBridge(
   openFilePath: string | null,
 ) {
   let loadingScreenHidden = false
+  let seenMessageListenersReadyMessage = false
 
   // Remove any existing message handlers to prevent us accidentally duplicating them
   window.removeEventListener('message', registeredHandlers)
   registeredHandlers = (messageEvent: MessageEvent) => {
     const { data } = messageEvent
-    if (isMessageListenersReady(data) && messageEvent.source != null) {
+    if (
+      isMessageListenersReady(data) &&
+      messageEvent.source != null &&
+      !seenMessageListenersReadyMessage
+    ) {
+      seenMessageListenersReadyMessage = true
       // Don't store the source yet, because we don't want to send any messages
       // until the bridge is ready
 
--- a/editor/src/components/code-editor/code-editor-container.tsx
+++ b/editor/src/components/code-editor/code-editor-container.tsx
@@ -7,11 +7,11 @@ import { VSCodeLoadingScreen } from './vscode-editor-loading-screen'
 import { getEditorBranchNameFromURL, setBranchNameFromURL } from '../../utils/branches'
 import { VSCODE_EDITOR_IFRAME_ID } from '../../core/vscode/vscode-bridge'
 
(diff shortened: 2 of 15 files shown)


bkrmendy (approved):

ruggi (approved):

Rheeseyb (approved):

Rheeseyb:
The "Run Systems Tests" check will never succeed here, as the PR branch deploy doesn't include VS Code changes, so we can just go ahead and ignore that failure

editor/src/core/vscode/vscode-bridge.ts
```diff
@@ -95,12 +95,18 @@ export function initVSCodeBridge(
...
 
   // Remove any existing message handlers to prevent us accidentally duplicating them
   window.removeEventListener('message', registeredHandlers)
   registeredHandlers = (messageEvent: MessageEvent) => {
     const { data } = messageEvent
-    if (isMessageListenersReady(data) && messageEvent.source != null) {
+    if (
+      isMessageListenersReady(data) &&
+      messageEvent.source != null &&
+      !seenMessageListenersReadyMessage
```
balazsbajorics, line 107: should this throw an error if multiple ready messages arrive?

balazsbajorics (approved):

balazsbajorics:
this is really nice!

editor/src/core/vscode/vscode-bridge.ts
seanparsons: I don't _think_ so, possibly more than one could end up in the pipeline, it's not really an error state just something we need to protect against.

PR #4648, opened 2023-12-13
Scrollable comment popup

ruggi:
#4647 

**Problem:**

The comment popup does not scroll to the bottom, does not have a maximum height, does not have a maximum width, etc.

**Fix:**

This PR makes the comment popup more friendly.

- It has a max height of 350px
- It has a max width of 250px
- When opening a comment thread, it scrolls to the bottom
- When there are comments under the current scroll, show a shadow at the bottom of the list
- When there are comments above the current scroll, show a shadow at the top of the list **and** show the first comment as a sticky header (so you don't lose context)
- When comments are long, they get broken into multiple lines (horizontal overflow)

https://github.com/concrete-utopia/utopia/assets/1081051/02cda567-10b2-4baa-9f9c-5aac704f9c9f

### DIFF
--- a/editor/src/components/canvas/controls/comment-popup.tsx
+++ b/editor/src/components/canvas/controls/comment-popup.tsx
@@ -2,6 +2,7 @@ import type { CommentData } from '@liveblocks/client'
 import type { ComposerSubmitComment } from '@liveblocks/react-comments'
 import { Composer } from '@liveblocks/react-comments'
 import { useAtom } from 'jotai'
+import type { CSSProperties } from 'react'
 import React, { useRef } from 'react'
 import { useCreateThread, useStorage } from '../../../../liveblocks.config'
 import '../../../../resources/editor/css/liveblocks-comments.css'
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/controls/comment-popup.tsx
```diff
@@ -225,6 +249,27 @@ const CommentThread = React.memo(({ comment }: CommentThreadProps) => {
...
+
+    const atTop = element.scrollTop > tolerance
+    setShowShadowTop(atTop && isOverflowing)
+  }
+
+  React.useEffect(() => {
+    // when the thread id changes, re-check the scroll and set the inset shadow
+    onScroll()
+    scrollToBottom()
+  }, [threadId, scrollToBottom])
```
gbalint, line 271: Is this necessary? As far as I know the threadId can only change when we are creating a new thread by typing the first comment, and in that case we don't have to scroll (there is only a single comment in the thread).

gbalint: But why? Does it run only once at the beginning, and that is needed?

balazsbajorics (approved):

editor/src/components/canvas/controls/comment-popup.tsx
ruggi: if you have an open comment and click on another comment, the thread id changes and the scroll needs to change as well, but without the useeffect the component is not destroyed and the scroll stuff (e.g. shadows) does not change

gbalint: Ok, I did not realize that the same element without unmount/remount can be switched to a different comment thread.

PR #4657, opened 2023-12-14
Continuous new comments, escape handling

### LINKED ISSUE #4656: Continuous comment creation, handle escape

ruggi:
Fix #4656

This PR addresses some of the current issues with the comment popups and mode management.

https://github.com/concrete-utopia/utopia/assets/1081051/651fb2cd-bf34-4d86-8924-2145776767ae

What it does:

1. When in comment mode and a thread is open, clicking on the canvas closes the popup but keeps the mode as comment (instead of switching back to select mode)
2. If the open comment is a new comment and it's empty, clicking away will make the comment shake to indicate there are pending changes
3. When pressing `Escape` in comment mode and a comment is open and focused, it will close the comment (and another `Escape` will switch back to select mode, as you would expect)
4. Added a close button (`X`) to the thread tabs

### DIFF
--- a/editor/src/components/canvas/controls/comment-mode/comment-mode-hooks.tsx
+++ b/editor/src/components/canvas/controls/comment-mode/comment-mode-hooks.tsx
@@ -72,7 +72,7 @@ export function useCommentModeSelectAndHover(comment: CommentId | null): MouseCa
 
   const onMouseUp = React.useCallback(
     (event: React.MouseEvent) => {
-      if (comment == null || isExistingComment(comment)) {
+      if (comment == null) {
         const loc = windowToCanvasCoordinates(
           storeRef.current.scale,
           storeRef.current.canvasOffset,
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/controls/comment-mode/comment-mode-hooks.tsx
```diff
@@ -12,6 +12,7 @@ import type { CommentId } from '../../../editor/editor-modes'
 import {
   EditorModes,
   canvasCommentLocation,
+  isExistingComment,
```
Rheeseyb, line 15:
```suggestion
```
No longer used

Rheeseyb (approved):

editor/src/components/canvas/controls/comment-mode/comment-mode-hooks.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/4657/commits/1c972aa236c0b52f357e23754d4f96c010c4af45

bkrmendy (approved):

PR #4659, opened 2023-12-15
Comment thread icons

lankaukk:
Introducing new icons for "Mark as unread" "Resolve" and "_Resolved_" used in the comment popup and in the comments tab. 
Screenshot 2023-12-15 at 10 14 21 AM

- Applied greyscale styling to comments that are resolved
- Reorganized the layout of the comment previews in the comment tab
- Unread comments in comments tab get a blue dot
- Comments in comments tab change background color on hover
- Change to the style of the header comment when the list in the comment popup is long

image
image

Screenshot 2023-12-15 at 10 18 20 AM
Screenshot 2023-12-15 at 10 34 51 AM
Screenshot 2023-12-15 at 11 56 12 AM


editor/src/components/canvas/controls/comment-popup.tsx
```diff
@@ -493,7 +494,7 @@ const NewCommentPopup = React.memo((props: NewCommentPopupProps) => {
   )
 
   return (
-    <>
+    <div>
```
maltenuhn, line 497: what was the reason for turning this fragment into a div? do we still need the div at all?

editor/src/components/inspector/sections/comment-section.tsx
```diff
@@ -193,57 +195,69 @@ const ThreadPreview = React.memo(({ thread }: ThreadPreviewProps) => {
...
-          paddingLeft: 44,
-          paddingRight: 10,
-          marginTop: -5,
-          display: 'flex',
-          alignItems: 'center',
-          justifyContent: 'space-between',
-        }}
-      >
+            Route:{' '}
+            <span style={{ fontWeight: 500 }}>{thread.metadata.remixLocationRoute ?? ''}</span>
```
Rheeseyb, line 253: Why are we not using `remixLocationRouteLabel` here (since we already know that is not `null`)? It seems like we're not using at all, so if that's deliberate we should just change the above `when` clause to be `thread.metadata.remixLocationRoute != null`

Rheeseyb (approved):

ruggi (approved):

PR #4665, opened 2023-12-15
Baton Passing Basic Edition

seanparsons:
**Problem:**
If a user opens the editor in multiple tabs they can end up in a nightmare scenario of different editors updating their own variant of the project and potentially work can be lost.

**Fix:**
This PR implements a first come-first served baton passing system of claiming ownership over the project. While there is no user interface for relinquishing or passing the baton actively to another user, closing the tab or not contacting the server for a period of over 20 seconds will result in control being released.

The underlying implementation is a simple database table containing the project ID and a collaboration editor ID, where the latter is generated by a client on startup and remains a secret to other editors.

**Commit Details:**
- Added `Baton Passing For Control` feature switch.
- Add a call to release control of the projects to the existing `beforeunload` event listener.
- Added `claimControlOverProject` and `releaseControl` calls to the server powered by `callCollaborationEndpoint`.
- Added `allowedToEditProject` and associated hook function `useAllowedToEditProject`, to replace some existing functions like `useIsViewer`.
- Added `currentlyHolderOfTheBaton` to `ProjectServerState`.
- Updated `getProjectServerState` to call `claimControlOverProject`.
- `ProjectServerStateUpdater` now defers to a new function `restartServerStateWatcher`, which updates on an interval if the appropriate feature flag is enabled.
- Added `project_collaboration` table to the database.
- Implemented `collaborationEndpoint` in the server to dispatch the various cases of collaboration request.
- Extracted out some functions into `Utopia.Web.Endpoints.Common` module.
- Added `ClaimCollaborationOwnership` and `ClearCollaboratorOwnership` cases to service types in the backend.


ruggi (commented):
did a first pass and will re-pass, but it looks 🆗 to me!

ruggi (approved):

editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -612,12 +613,11 @@ export function handleKeyDown(
...
       [INSERT_IMAGE_SHORTCUT]: () => {
-        if (isViewer) {
-          return []
-        }
-        if (isSelectMode(editor.mode) || isInsertMode(editor.mode)) {
-          // FIXME: Side effects.
-          insertImage(dispatch)
+        if (allowedToEdit) {
+          if (isSelectMode(editor.mode) || isInsertMode(editor.mode)) {
+            // FIXME: Side effects.
```
Rheeseyb, line 618: I think this FIXME can just go in the bin because at this point the intention is pretty much lost

seanparsons: Changed it to indicate that there is a side effect, removing the "FIXME" in c2e2e1ea4.

editor/src/components/editor/persistence/persistence.ts
```diff
@@ -133,7 +134,8 @@ export class PersistenceMachine {
 
     this.interpreter.start()
 
-    window.addEventListener('beforeunload', (e) => {
+    window.addEventListener('beforeunload', async (e) => {
+      await releaseControl()
```
Rheeseyb, line 138: I don't think this should be using an `await` because the browser would just let the page close anyway, meaning the code below might not execute if this takes too long. Also, if the browser does close the page too quickly for this to execute, will control still be released server side?

gbalint (approved):

editor/src/components/editor/project-owner-hooks.ts
```diff
@@ -21,33 +20,25 @@ export function useDisplayOwnershipWarning(): void {
...
+          globalThis.requestAnimationFrame(() => {
+            dispatch([removeToast(OwnershipToastID)])
+          })
+        } else {
+          // Add the toast if we switch to a project that the user does not own.
+          globalThis.requestAnimationFrame(() => {
+            dispatch([
+              showToast(
+                notice(
+                  'Viewer Mode: As you are not the owner of this project, it is read-only.',
```
Rheeseyb, line 34: I think this warning text is now confusing, as we'll be showing this even if you are the owner of the project but have it open in another tab. If we can detect at this point that you're the owner but still in viewer mode then we should adjust the message to reflect that. If not we should either remove the part about "you are not to the owner" or include something about it maybe being open in another tab

editor/src/components/editor/persistence/persistence.ts
gbalint: `visibilitychange` is generally safer than `beforeunload`: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#usage_notes

server/src/Utopia/Web/Database.hs
```diff
@@ -437,3 +444,74 @@ lookupGithubAuthenticationDetails metrics pool userId = invokeAndMeasure (_getGi
...
+                               }
+
+maybeClaimExistingCollaborationOwnership :: Connection -> Text -> Text -> Text -> UTCTime -> UTCTime -> IO Bool 
+maybeClaimExistingCollaborationOwnership connection projectId currentCollaborationEditor newCollaborationEditor currentLastSeenTimeout currentTime 
+  | currentCollaborationEditor == newCollaborationEditor = updateLastSeen 
+  | currentLastSeenTimeout < currentTime                 = updateLastSeen
+  | otherwise                                            = pure False
+  where 
+    newTimeout = addUTCTime collaborationLastSeenTimeoutWindow currentTime
+    updateLastSeen = updateCollaborationLastSeenTimeout connection projectId newCollaborationEditor newTimeout >> pure True
```
Rheeseyb, line 477: Ok I see this answers my earlier question :)

server/src/Utopia/Web/Endpoints/Collaboration.hs
```diff
@@ -0,0 +1,72 @@
...
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE RecordWildCards       #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TypeApplications      #-}
+{-# LANGUAGE TypeOperators         #-}
+
+{-|
+  All the endpoints defined in "Utopia.Web.Types" are implemented here.
```
Rheeseyb, line 13: This comment is a **LIE**

Rheeseyb (approved):

editor/src/components/editor/persistence/persistence.ts
seanparsons: @gbalint Does that mean this save hook should be using that instead?

editor/src/components/editor/project-owner-hooks.ts
seanparsons: That's a good shout.

server/src/Utopia/Web/Endpoints/Collaboration.hs
seanparsons: Ah, copy-pasta.

editor/src/components/editor/persistence/persistence.ts
gbalint: @seanparsons yes, I think so

server/src/Utopia/Web/Endpoints/Collaboration.hs
seanparsons: Fixed in fcc60802d.

editor/src/components/editor/project-owner-hooks.ts
seanparsons: Fixed in fcc60802d.

editor/src/components/editor/persistence/persistence.ts
seanparsons: Fixed in fcc60802d.

editor/src/components/editor/persistence/persistence.ts
```diff
@@ -133,7 +133,7 @@ export class PersistenceMachine {
 
     this.interpreter.start()
 
-    window.addEventListener('beforeunload', (e) => {
+    window.addEventListener('beforeunload', async (e) => {
```
balazsbajorics, line 136: was this an eslint warning? the function doesn't seem to await anything

seanparsons: Oh it did, I can remove that now.

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -609,3 +612,119 @@ function synchroniseParseSuccessToCollabFile(
...
+  } else {
+    throw new Error(`Unexpected response: ${JSON.stringify(response)}`)
+  }
+}
+
+export async function releaseControl(): Promise<void> {
+  if (isFeatureEnabled('Baton Passing For Control')) {
+    const request = clearAllOfCollaboratorsOwnership(collaborationEditor)
+
+    const response = await callCollaborationEndpoint(request)
```
balazsbajorics, line 700: is this also somehow updating the porjectServerState?

seanparsons: It's not, but only by virtue of it only being called when the editor closes so there's not much point.

balazsbajorics: ah i see! I thought this is also being called if you tab away, or somehow send the editor to the background

seanparsons: This much is true, as a result of the switch of event handler.

That complicates things because now it needs access to the dispatch function so it can update that state, I'm actually tempted to revert that change because it means potentially losing control when the browser just fancies it rather than when the tab is closed.

seanparsons: Reverted event handler change in 4267c3e67.

balazsbajorics (approved):

PR #4668, opened 2023-12-19
Remove unused imports in `removeElementAtPath`

liady:
**Problem:**
When inserting a new element using drag&draw - the import for this element remains in `storyboard.js` (even if the element itself is being inserted elsewhere).

**Cause:**
We are not removing unused imports when we call `removeElementAtPath`

**Fix:**
Adding the ability to remove unused imports inside `removeElementAtPath` - if are not used anymore


ruggi (approved):

seanparsons (approved):

editor/src/components/editor/import-utils.ts
```diff
@@ -212,3 +223,54 @@ export function getImportsFor(
 
   return emptyImports()
 }
+
+export function removeUnusedImports(
+  element: JSXElement,
```
Rheeseyb, line 228: Minor point here, but I'd suggest renaming this to `removedElement`, and possibly even renaming the function to reflect that this is removing unused imports specifically for that removed element

liady:
Tests are passing but I want to check some edge case that popped up (with the worker update) before merging.

Rheeseyb (changes requested):
As discussed, this causes a breaking issue when dragging to insert

liady:
> As discussed, this causes a breaking issue when dragging to insert

Fixed (not removing the imports when reparenting the same file)

balazsbajorics (approved):

PR #4679, opened 2024-01-03
Get Remix to work with local webpack.

seanparsons:
**Problem:**
When loading any project using Remix in the editor with local webpack building the editor, you get a big error about `env` not being defined on `undefined`.

**Cause:**
In a function `remixLiveReloadConnect` there's some logic that checks `process.env.REMIX_DEV_ORIGIN` which webpack turns into `undefined.env.REMIX_DEV_ORIGIN` which fails for fairly obvious reasons.

**Fix:**
This provides a value for the specific case of `process.env.REMIX_DEV_ORIGIN`.

**Commit Details:**
- Assigns a value to `process.env.REMIX_DEV_ORIGIN`.

### DIFF
--- a/editor/webpack.config.js
+++ b/editor/webpack.config.js
@@ -166,7 +166,7 @@ const config = {
       'process.env.JEST_WORKER_ID': 'undefined',
       'process.env.HOT_MODE': hot,
       'process.env.HMR': false,
-      'process.env.REMIX_DEV_ORIGIN': "'http://localhost:8000'",
+      'process.env.REMIX_DEV_ORIGIN': isDev ? "'http://localhost:8000'" : 'undefined',
     }),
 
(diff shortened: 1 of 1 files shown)


editor/webpack.config.js
```diff
@@ -166,6 +166,7 @@ const config = {
       'process.env.JEST_WORKER_ID': 'undefined',
       'process.env.HOT_MODE': hot,
       'process.env.HMR': false,
+      'process.env.REMIX_DEV_ORIGIN': "'http://localhost:8000'",
```
Rheeseyb, line 169: Can you use an `isDev` check here please, so that we won't accidentally end up pointing at localhost if we ever switch to using a dev mode of remix in the deployed editor

seanparsons: Fixed in f7b48c4a8.

PR #4695, opened 2024-01-05
Clean up comment filtering in comment section

gbalint:
**Problem:**
Comment sorting and filtering in the comment sidebar is hard to understand and the code is convoluted too.

**Fix:**
Let's simplify how it works:
- comments are always sorted by last modification date
- there are three filtering view modes:
  1. "All" shows all unresolved comments 
  2. "All including resolved" shows all comments
  3. "Unread only" shows only unread and unresolved comments

This preference is persisted into the editor state, so closing and opening the comments tab keeps the previously chosen filter mode.
The same flag is used to decide whether resolved comment markers should be visible on the canvas.

**TODO:**
- Make the design better, I just added a simple PopupList
- In "Unread only" filter mode the canvas markers of the read threads should be filtered (this is only implemented for resolved/unresolved threads).

### DIFF
--- a/editor/src/components/inspector/sections/comment-section.tsx
+++ b/editor/src/components/inspector/sections/comment-section.tsx
@@ -39,6 +39,7 @@ import { canvasPointToWindowPoint } from '../../canvas/dom-lookup'
 import { CommentRepliesCounter } from '../../canvas/controls/comment-replies-counter'
 import type { SelectOption } from '../controls/select-control'
 import { assertNever } from '../../../core/shared/utils'
+import { pluck } from '../../../core/shared/array-utils'
 
 export type CommentFilterMode = 'all' | 'all-including-resolved' | 'unread-only'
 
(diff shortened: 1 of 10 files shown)


bkrmendy (approved):

seanparsons (approved):

ruggi (approved):

editor/src/components/inspector/sections/comment-section.tsx
```diff
@@ -70,61 +78,46 @@ const ThreadPreviews = React.memo(() => {
...
-      const theReadThreads = filteredThreads.filter((thread) => readThreads.includes(thread))
-      filteredThreads = [...unreadThreads, ...theReadThreads]
+    switch (commentFilterMode) {
+      case 'all':
+        return threads.filter((t) => !t.metadata.resolved)
+      case 'all-including-resolved':
+        return threads
+      case 'unread-only':
+        const readThreadIds = pluck(readThreads, 'id')
+        return threads.filter((t) => !readThreadIds.includes(t.id))
```
Rheeseyb, line 109: It seems this will still show resolved comments if they are unread

gbalint: True, fixed in 3fe97d7aa3bbdf382f0d0ad11ad190c2d92a8800

Rheeseyb (approved):

PR #4702, opened 2024-01-08
Puppeteer test for collaboration

bkrmendy:
# Description
This PR adds a Puppeteer test, which
- opens two browser windows
- with a different fake user in each one
- has one of the users delete one of the scenes in the project
- has the other assert that the scene has been deleted in the other page too

# Follow-up work after this PR
- create a project on pizza that both browsers open (both browser have to open the same project, for the collaboration to work)
- update the localhost link to the new project URL
- create a CI job that runs the collaboration test

## Testing the test locally
- Log out from the editor on localhost (by navigating to `localhost:8000/logout`)
- navigate to `http://localhost:8000/p?fakeUser=alice`
- log in (this will automatically create a project)
- substitute the new project for `http://localhost:8000/p/56a2ac40-caramel-yew` in the new test
- run `pnpm run collaboration-test` from `puppeteer-tests`

### DIFF
--- a/puppeteer-tests/src/comments/collaboration-test.spec.tsx
+++ b/puppeteer-tests/src/comments/collaboration-test.spec.tsx
@@ -21,6 +21,10 @@ async function expectNSelectors(page: Page, selector: string, n: number) {
   expect(elementsMatchingSelector).toHaveLength(n)
 }
 
+async function waitForCoeditPropagation() {
+  await wait(2000)
+}
+
 describe('Comments test', () => {
   it(
     'can place a comment',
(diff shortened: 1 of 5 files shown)


puppeteer-tests/src/comments/collaboration-test.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+async function expectNSelectors(page: Page, selector: string, n: number) {
+  const elementsMatchingSelector = await page.$$(selector)
+  expect(elementsMatchingSelector).toHaveLength(n)
+}
+
+describe('Comments test', () => {
+  it(
+    'can place a comment',
+    async () => {
+      const [{ page: page1, browser: browser1 }, { page: page2, browser: browser2 }] =
```
seanparsons, line 28: Not a huge issue but I'd find this easier to follow if it looked like this:
```typescript
const browser1Promise = setupBrowser(...)
const browser2Promise = setupBrowser(...)
const {page: page1, browser: browser1} = await browser1Promise
const {page: page2, browser: browser2} = await browser2Promise
```

seanparsons (approved):

gbalint (approved):

puppeteer-tests/src/comments/collaboration-test.spec.tsx
bkrmendy: good idea, implemented in https://github.com/concrete-utopia/utopia/pull/4702/commits/14a36f563e78c920b24a7a0fa7aeca2250789183

PR #4704, opened 2024-01-08
Feature: GitHub autofork url

balazsbajorics:
Verbose URL `https://utopia.pizza/p/?github_owner=balazsbajorics&github_repo=hydrogen-november&github_branch=main`

Shorter URL `https://utopia.pizza/p/?clone=balazsbajorics/hydrogen-november`

Shorter URL with branch name `https://utopia.pizza/p/?clone=balazsbajorics/hydrogen-november&github_branch=main`

Try Me link:
Verbose URL [link](https://utopia.pizza/p/?branch_name=feature-github-fork-url&github_owner=balazsbajorics&github_repo=hydrogen-november&github_branch=main)

**Problem**
The only (maintained, aside from the old zip-based repo checkout POC) working method for cloning a github repository was to create a new project, navigate to the github tab, then insert the owner / repo name, select a branch, press clone, overwrite contents of the new project.

We already had a URL-based entry point for the git checkout proof of concept which was cloning a git repo using a zip download. To my surprise it still works today, but it doesn't do the asset upload step, and does not connect the editor's github tab to the cloned repository. Plus I couldn't remember what other limitations it might have, so Balazs 2023 December decided to not spend time trying to understand if we should fix that instead of using the main github cloning feature.

**What did I do**
So I set out to replace the existing git auto-clone functionality with our "modern" and maintained github integration. I went down a bit of a dead end in #4660 and yesterday I decided to scrap most of my changes and come up with a vastly simplified solution. 
Basically what this PR ended up doing is that it still does load (an empty) project, and then calls the same code path that runs when a user selects a repository in the github tab.

**Flow of cloning a repo:**
The flow is orchestrated from `GithubRepositoryCloneFlow` which is a react component in editor-component. This is currently our _only_ loading flow which is not directly orchestrated from editor.tsx.

0. Editor.tsx detects that we want to load from a github repo, and sets the EditorStore accordingly
1. GithubRepositoryCloneFlow gets enabled, and makes sure the user is logged in and authenticated with github.
2. Once the user is logged in and authenticated, we fire off `cloneGithubRepo`
3.`cloneGithubRepo` creates a new totally empty project that only has a package.json. It fires `PersistenceMachine.createNew` so we get a proper Project ID, then saves this empty project to the server
4. the code awaits the `LOAD` action being dispatched to avoid race conditions
5. once the LOAD has been dispatched, we initiate the github cloning by calling `GithubOperations.updateProjectWithBranchContent` (that's the exact same code path that is called if the user clicks "Load from Branch" in the github pane)
6. Profit

**Changes to the Server**
Parallel to my editor changes, I also added a new server endpoint: the ability to clone a repo without specifying a branch. Since the default branch of git repos can be arbitrarily named, the server code first figures out the default repo branch name, and then calls the existing load-a-branch code path.

**Not logged in states**
image
image
If the user is not logged in, or not authenticated with github, we stop the flow and wait for their action.


editor/src/templates/editor.tsx
```diff
@@ -357,47 +373,65 @@ export class Editor {
           }
 
           const urlParams = new URLSearchParams(window.location.search)
-          const githubOwner = urlParams.get('github_owner')
-          const githubRepo = urlParams.get('github_repo')
+          const githubBranch = urlParams.get('github_branch')
           const importURL = urlParams.get('import_url')
 
+          const githubRepoObj: GithubRepo | null = (() => {
```
seanparsons, line 379: Could this just be turned into a regular function and have it be invoked here? Feels like a lot of URL handling logic all embedded within the `Editor` component.

seanparsons (approved):

editor/src/templates/editor.tsx
balazsbajorics: yeah very good point, it's became too massive

editor/src/components/github/github-repository-clone-flow.tsx
```diff
@@ -0,0 +1,177 @@
...
+let githubRepoAlreadyCloning = false
+async function cloneGithubRepo(
+  dispatch: EditorDispatch,
+  storeRef: { current: EditorStorePatched },
+  githubRepo: GithubRepoWithBranch,
+) {
+  if (githubRepoAlreadyCloning) {
+    return
+  }
+  githubRepoAlreadyCloning = true
```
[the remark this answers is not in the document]
balazsbajorics: hmm I think I need to give this a better name. we absolutely only want to trigger this once per editor load, no matter what

editor/src/utils/github-auth.ts
```diff
@@ -64,3 +66,11 @@ export const GithubAuth = {
   isAuthenticatedWithGithub: isAuthenticatedWithGithub,
   startGithubAuthentication: startGithubAuthentication,
 } as const
+
+export function useOnClickAuthenticateWithGithub() {
```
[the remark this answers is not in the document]
balazsbajorics: do you have a problem with the current location _other than_ it's a ts file?

editor/src/components/github/github-repository-clone-flow.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+export const GithubRepositoryCloneFlow = React.memo(() => {
+  const githubRepo = useEditorState(
+    Substores.userState,
+    (store) => store.userState.githubState.gitRepoToLoad,
+    'GithubRepositoryCloneFlow gitRepoToLoad',
+  )
+  const userLoggedIn = useEditorState(
+    Substores.userState,
+    (store) => store.userState.loginState.type === 'LOGGED_IN',
```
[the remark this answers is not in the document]
balazsbajorics: thanks! fixed

editor/src/components/github/github-repository-clone-flow.tsx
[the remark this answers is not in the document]
balazsbajorics: fixed!

editor/src/core/shared/github/endpoints.ts
```diff
@@ -18,6 +18,15 @@ export const GithubEndpoints = {
...
       branchName,
     ),
+  defaultBranchContents: (githubRepo: GithubRepo) =>
+    urljoin(
+      UTOPIA_BACKEND,
+      'github',
+      'branches',
+      githubRepo.owner,
+      githubRepo.repository,
+      'default-branch',
```
ruggi, line 28: shall we stick wi GH's (unchanged) [default](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches#about-the-default-branch) here? (`main`)

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1517,7 +1517,11 @@ function createStoryboardFileIfMainComponentPresent(
 
 function createStoryboardFileWithPlaceholderContents(
   projectContents: ProjectContentTreeRoot,
+  isItNecessary: 'create-placeholder' | 'skip-creating-placeholder',
```
seanparsons, line 1520: Be nice to extract this into a type definition as it appears to be in at least 2 places.

editor/src/components/github/github-repository-clone-flow.tsx
```diff
@@ -0,0 +1,179 @@
...
+  )
+
+  // at this point we can assume the repo is loaded and we can finally hide the overlay
+  dispatch([
+    setGithubState({
+      gitRepoToLoad: null,
+    }),
+  ])
+
+  // TODO make sure the EditorState knows we have a github repo connected!!!
```
seanparsons, line 121: Future work?

editor/src/components/github/github-repository-clone-flow.tsx
```diff
@@ -0,0 +1,179 @@
...
+    void cloneGithubRepo(dispatch, editorStoreRef, githubRepo)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  // The GitClonePseudoElement's sole job is to call cloneGithubRepo in a useEffect.
+  // I pulled it to a dedicated component so it's purpose remains clear and this useEffect doesn't get lost in the noise
+  return null
+})
+
+const awaitLoadActionDispatchedByPersistenceMachine = (): Promise<{ projectId: string }> => {
```
seanparsons, line 140: Can this be a regular function?

seanparsons (approved):

editor/src/components/editor/actions/actions.tsx
Rheeseyb: Also `isItNecessary` is a bad name. `shouldCreatePlaceholder` or just `createPlaceholder` would be better

editor/src/components/editor/store/editor-state.ts
```diff
@@ -2972,8 +2976,13 @@ const defaultDependencies = Utils.mapArrayToDictionary(
...
 export const defaultIndexHtmlFilePath = 'public/index.html'
 
+export const EmptyPackageJson = {
+  name: 'utopia-project',
+  version: '0.1.0',
+}
+
 export const DefaultPackageJson = {
-  name: 'Utopia Project',
+  name: 'utopia-project',
```
Rheeseyb, line 2985: 🙌

editor/src/components/github/github-repository-clone-flow.tsx
```diff
@@ -0,0 +1,179 @@
...
+// The git repo clone flow is initiated from the URL, which means we only ever want to do it once per editor load
+let didWeInitiateGitRepoDownloadSinceTheEditorLoaded = false
+
+async function cloneGithubRepo(
+  dispatch: EditorDispatch,
+  storeRef: { current: EditorStorePatched },
+  githubRepo: GithubRepoWithBranch,
+) {
+  if (didWeInitiateGitRepoDownloadSinceTheEditorLoaded) {
+    return
```
Rheeseyb, line 91: Should this throw an error instead? This feels like something that shouldn't fail silently

Rheeseyb (approved):

editor/src/components/editor/store/editor-state.ts
balazsbajorics: somewhat unrelated mini fix

editor/src/components/github/github-repository-clone-flow.tsx
balazsbajorics: this function gets called a couple of times because of ancestor remounts

editor/src/components/github/github-repository-clone-flow.tsx
balazsbajorics: yep, for the test I will write

editor/src/components/editor/actions/actions.tsx
balazsbajorics: I did the renaming, I did not create a type definition though because I think this is more readable. we should talk about it in the chat if you disagree :)

PR #4710, opened 2024-01-09
CI job for collab test

bkrmendy:
## Description
This PR adds a CI job that runs the test for commenting.

### Also
- makes it possible to run this test locally, the same way as implemented in https://github.com/concrete-utopia/utopia/pull/4715
- tweaks the test so that no hardcoded project name is necessary, the second browser navigates to the URL the first one is redirected to after the project is created.


seanparsons (approved):

editor/src/components/editor/persistence/persistence.ts
```diff
@@ -100,7 +103,10 @@ export class PersistenceMachine {
             )
             this.queuedActions.push(...updateFileActions)
             this.lastSavedTS = Date.now()
-            // TODO Show toasts after first server save of a previously local project
+            if (!this.projectUploadedToServer && event.source === 'server') {
```
Rheeseyb, line 106: 👍

Rheeseyb (approved):

PR #4711, opened 2024-01-09
feature(editor) Snatching of control support.

seanparsons:
**Problem:**
The current functionality of claiming control isn't really that helpful with multiple tabs as once a tab has claimed control it holds it forever until it's closed or loses connection.

**Fix:**
We want the visibility of the tabs to govern if they have control, so once a tab is hidden it releases control and when one becomes visible it will snatch control, ignoring whoever else used to have control of that tab.

To this end the following changes have been made:
- Added an `owner_id` column to the `project_collaboration` table to guard against someone releasing someone else's control over a project.
- A new snatch control call was added which overrides any prior control over the project and allocates it to the session doing the snatching, which is triggered by a session going from hidden to visible.
- A new release control call was added to immediately release control of a specific project, which is triggered by a session going from visible to hidden.
- Partial updates are now possible to `ProjectServerState`, which is used by the two calls above so that they don't have to rely on the slow polling call.
- A fix has been made to the dispatch logic so that changes to the `projectServerState` field are picked up instantly after they have been dispatched.

There's also a mostly unrelated fix but that was used during the development of this work which passes parameters through to `psql` from `run-psql`, which means you can run something like the following to watch the contents of a table:
```bash
watch -n 1 'run-psql -c "SELECT * FROM project_collaboration;"'
```

**Commit Details:** 
- Modified `UpdateProjectServerState` to take a `Partial<ProjectServerState>`.
- Renamed `releaseControl` to `clearAllControlFromThisEditor`.
- Created `CollaborationStateUpdater` component to manage the visibility changes of the editor.
- Added `SnatchProjectControl` and `ReleaseProjectControl` to `CollaborationRequest`.
- Renamed `ClaimProjectControlResult` to `RequestProjectControlResult`.
- Renamed `ClearAllOfCollaboratorsControlResult` to `ReleaseControlResult`.
- Added `snatchControlOverProject` and releaseControlOverProject`.
- Renamed `releaseControl` to `clearAllControlFromThisEditor`.
- Modified `editorDispatchClosingOut` to use `result.projectServerState` instead of `storedState.projectServerState`.
- Modified `editorDispatchInner` to take account of `projectServerState` when determining if anything has changed.
- Changed `getProjectServerState` to return a more useful value for test environments and to check if the user owns a project before attempting to claim ownership.
- Added `owner_id` column to `project_collaboration`.
- Updated some database logic to take account of `owner_id` in `project_collaboration`.
- Added `forceClaimCollaborationControl` and `releaseCollaborationControl` to the database handling.
- Added `SnatchCollaborationControl` and `ReleaseCollaborationControl` to `ServiceTypes`.
- Modified `ClearCollaboratorOwnership` to take the user ID.
- Small tweak to the `run-psql` script to support passing parameters to `psql`.


editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -681,22 +719,51 @@ export async function claimControlOverProject(projectID: string | null): Promise
...
+  }
+}
+
+export async function snatchControlOverProject(projectID: string | null): Promise<boolean | null> {
+  if (projectID == null || !isFeatureEnabled('Baton Passing For Control')) {
+    return null
+  }
+
+  const request = snatchProjectControl(projectID, collaborationEditor)
+  const response = await callCollaborationEndpoint(request)
```
ruggi, line 741: would it make sense to put this into a `try` and handle the failure?

ruggi (approved):

editor/src/components/editor/store/collaboration-state.tsx
```diff
@@ -0,0 +1,49 @@
...
+        if (projectId != null) {
+          // If the document is hidden, that means the editor is in the background
+          // or minimised, so release control. Otherwise if it has become unhidden
+          // then snatch control of the project.
+          if (document.hidden) {
+            void releaseControlOverProject(projectId).then(() => {
+              dispatch([updateProjectServerState({ currentlyHolderOfTheBaton: false })])
+            })
+          } else {
+            // Only attempt to do any kind of snatching of control if the project is "mine".
```
Rheeseyb, line 31: Does it make sense to wrap the entire thing in `if (isMyProject === 'yes')` (including the releasing part) rather than just the snatching part?

Rheeseyb (approved):

editor/src/components/editor/store/collaboration-state.tsx
seanparsons: There's been talk that we might be able to hand over control to another user, in which case we want the release control to apply in that case. It's just that currently it's invalid to claim control if you're not the owner. Specifically it will throw an exception.

editor/src/components/editor/store/collaborative-editing.ts
seanparsons: Fixed in 3d682504d.

PR #4713, opened 2024-01-09
Input/controls constraints for viewers

### LINKED ISSUE #4712: Disable inspector inputs for non-editors

ruggi:
Fixes #4712 

**Problem:**
- inspector inputs appear editable for viewers - even if they are really not doing anything
- drag/resize/etc. controls are visible for viewers on selected elements, which should not be

**Fix:**

- Hide element controls on the canvas for selected elements for viewers
- Disable the various inputs (and tangential actionable components and interactions) for viewers
	- number inputs
	- string inputs
	- draggable scrubbers
	- color pickers
	- popup lists
	- increment buttons
	- color swatches
- added the `useIsMyProject` hook which returns whether the _user_ is an owner of the current project, regardless of the baton


seanparsons (approved):

editor/src/components/editor/global-shortcuts.tsx
```diff
@@ -686,6 +686,9 @@ export function handleKeyDown(
         )
       },
       [INSERT_DIV_SHORTCUT]: () => {
+        if (!allowedToEdit) {
```
Rheeseyb, line 689: We have some other `INSERT_` shortcuts that might need this check then

editor/src/components/editor/store/collaborative-editing.ts
```diff
@@ -717,3 +717,11 @@ export function useAllowedToEditProject(): boolean {
     'useAllowedToEditProject',
   )
 }
+
+export function useIsMyProject(): boolean {
```
Rheeseyb, line 721: Does this clash with the baton passing work? I'm thinking that a lot of the places using this check should be instead using `useAllowedToEditProject`?

Rheeseyb (approved):
I'm approving because the code and feature looks good, but please check the question around whether `useIsMyProject` is suitable for all the cases here before merging

editor/src/components/editor/store/collaborative-editing.ts
ruggi: I don't think it clashes, but it definitely makes it more granular. My take would be to wait that there are no baton-opened PRs to revisit and update

PR #4719, opened 2024-01-10
More commenting test

gbalint:
**Description:**
I wrote a few tests for commenting.
I'm not decided on the general pattern on how to do it, because
1. I prefer granular tests of well-defined features
BUT
2. The tests have a huge overhead to launch, because they open a new browser and load the utopia app and an empty project, which takes a lot of time.

So I wrote a longer test, which shows how we could write unified tests to test longer workflows, and I also wrote small tests, which only test a single feature. We can decide which pattern to follow.


puppeteer-tests/src/comments/comment-utils.ts
```diff
@@ -0,0 +1,37 @@
+import type { Page } from 'puppeteer'
+import type { UtopiaPuppeteerBrowser } from './test-utils'
+
+export const TIMEOUT = 120000
+
+const BRANCH_NAME = process.env.BRANCH_NAME ? `&branch_name=${process.env.BRANCH_NAME}` : ''
+const BASE_URL = process.env.BASE_URL ?? 'http://localhost:8000'
+
+export async function init(utopiaBrowser: UtopiaPuppeteerBrowser) {
```
ruggi, line 9: I think this would deserve a more specific name, e.g. `initBrowserTest`

ruggi (approved):
🙌

bkrmendy (approved):

PR #4724, opened 2024-01-11
Use Liveblocks to trigger control updates.

seanparsons:
**Problem:**
We use polling to handle the control updates, but this can mean a pause between an editor dropping control and another regaining it.

**Fix:**
Liveblocks has support for broadcasting events to a "room", which we map to our projects. When an editor takes control over a room, it sends a broadcast event to the other instances with that project open to tell them a change has occurred so they can attempt to claim (not snatch) control and as a result potentially gain or lose control of the project.

**Commit Details:** 
- Added a type `ControlChangedRoomEvent`, making that the only case of `RoomEvent`.
- `CollaborationStateUpdater` now uses `useBroadcastEvent` to notify other connections that a change has been triggered to who controls the room.
- `CollaborationStateUpdater` uses `useEventListener` to listen for updates to who controls the room.
- Removed a check for if the document has focus from `claimControlOverProject`.

### DIFF
--- a/editor/src/components/editor/store/collaboration-state.tsx
+++ b/editor/src/components/editor/store/collaboration-state.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import type { EditorAction, EditorDispatch } from '../action-types'
 import {
+  claimControlOverProject,
   displayControlErrorToast,
   releaseControlOverProject,
   snatchControlOverProject,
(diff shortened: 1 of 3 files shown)


editor/src/components/editor/store/collaboration-state.tsx
```diff
@@ -22,28 +29,58 @@ export const CollaborationStateUpdater = React.memo(
...
+    const broadcast = useBroadcastEvent()
+
+    // Handle events that appear to have come from the above broadcast call.
+    useEventListener((data) => {
+      if (data.event.type === 'CONTROL_CHANGED') {
+        if (isMyProject === 'yes') {
+          void claimControlOverProject(projectId)
+            .then((controlResult) => {
+              const newHolderOfTheBaton = controlResult ?? false
+              handleControlUpdate(newHolderOfTheBaton ?? false)
```
Rheeseyb, line 57: Looks like the extra `??` here is no longer necessary

Rheeseyb (approved):

PR #4726, opened 2024-01-11
Rename duplicate JSX element names

liady:
**Problem:**
Address more locations where we recognize duplicate imports (`mergeImports` returns a `duplicateNameMapping` that requires renaming the element)

**Fix:**
Make sure to also rename the inserted element in these cases


seanparsons (approved):

editor/src/components/custom-code/code-file.test-utils.ts
```diff
@@ -80,3 +86,24 @@ export function getTextFileByPath(projectContents: ProjectContentTreeRoot, path:
     throw new Error(`Unable to find a text file at path ${path}.`)
   }
 }
+
+export function getCodeForFile(
```
Rheeseyb, line 90: I'd suggest renaming this to reflect the fact that this isn't just getting the code, but is actually printing the parsed model

Rheeseyb: Oh sorry, I hadn't realised this was just moved from a different file. I think it's still worth renaming though

Rheeseyb (approved):

PR #4740, opened 2024-01-12
Share Button

lankaukk:
Share the project link with the click of a button!

share

- creating a new "Share" button with the current User's avatar in it
- clicking it copies the project url to the clipboard (for now until we have a proper dialog pop up with permissions settings)
- clicking it fires a toast
- deleted the old TitleBar bc I thought it was causing my errors, turns out it wasn't, but we don't need it anymore anyway!
- ✨ Bonus: fixed the avatar alignment so we don't see those awkward colored borders around them anymore

Screenshot 2024-01-12 at 5 59 32 PM


editor/src/components/user-bar.tsx
```diff
@@ -53,7 +56,21 @@ export const UserBar = React.memo(() => {
...
+const SinglePlayerUserBar = React.memo(() => {
+  const dispatch = useDispatch()
+
+  const url = window.location.href
+  const handleCopyToClipboard = React.useCallback(async () => {
+    try {
+      let actions: EditorAction[] = []
+      actions.push(showToast(notice('Project link copied to clipboard!', 'NOTICE', false)))
+      await window.navigator.clipboard.writeText(url)
+      dispatch(actions)
```
balazsbajorics, line 68: Nitpick: you don't need to create an array here, we only do that in edge case situations where we conditionally push actions into the array, but want to keep it a single dispatch.

For this simple case, you can directly dispatch the action:

```suggestion
      await window.navigator.clipboard.writeText(url)
      dispatch([showToast(notice('Project link copied to clipboard!', 'NOTICE', false))])
```

balazsbajorics (approved):

seanparsons (approved):

PR #4742, opened 2024-01-15
Rename JSXElementChildren

liady:
**Problem:**
Currently when we rename elements, we only support renaming JSXElements

**Fix:**
Support renaming all JSXElementChildren

### DIFF
--- a/editor/src/core/model/element-template-utils.spec.tsx
+++ b/editor/src/core/model/element-template-utils.spec.tsx
@@ -51,6 +51,8 @@ import {
   jsxTextBlock,
   utopiaJSXComponent,
   JSXAttributes,
+  jsxConditionalExpression,
+  jsxFragment,
 } from '../shared/element-template'
 import {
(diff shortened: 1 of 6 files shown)


Rheeseyb (approved):

editor/src/core/model/element-template-utils.spec.tsx
```diff
@@ -1954,3 +1957,56 @@ describe('transformJSXComponentAtPath', () => {
     expect((updatedElement.children[0] as JSXTextBlock).text).toEqual('hello')
   })
 })
+
+describe('renameJSXElementChild', () => {
+  it('should rename a simple element', () => {
+    const jsxElementToRename = jsxElement('Flex', 'flex-1', [], [])
```
balazsbajorics, line 1963: ultra nitpick, but `flex-1` would be an illegal javascript name which would throw an exception :)

balazsbajorics (approved):

PR #4743, opened 2024-01-15
Mass disable controls for non-editors: usePermissions, DisableControlsInSubtree 

balazsbajorics:
**Context**
On master we mass-disabled the various input controls (widgets?) of the editor (the string input, the number input, the PopupList, etc) if the user is not an editor. Specifically, they all individually check `const isMyProject = useIsMyProject()` and disable themselves based on this.

**Problems with this**
1. We want to enable some of these controls (widgets?) even for viewers of the project, the simplest example are controls (checkboxes and PopupList) used in the settings pane. We don't want to tie these components to user permissions! 
2. Zooming out, the controls (widgets?) used by the inspector and other parts of the UI are generic (as in: they are our internal "ui library") ((See [Presentational Components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0).)) and do not hold any business logic related to the editor. Current master breaks this by introducing checks for `isMyProject` inside these otherwise generic components. The genericity of these components is important, because it means we can use them in various contexts (inspector sections, settings pane, future other UI) and we can trust that these will behave like they are generic UI components, and don't require hidden knowledge of editor internals to use them!
3. (premature optimization) We expect "The user owns this project" to be the wrong abstraction level to talk about these stuff, and instead we want to talk about "the user has edit permission on this project". This will come in handy in the future once we have more roles with overlapping (and not hierarchical!!!) permissions.

**My proposed solution**
1. Decouple permissions from mass-disabling components. Mass disabling components can be expressed in a way that matches how a component library would think about this – not tied to any editor specific business logic, just to a Context whether the controls (widgets?) are mass disabled or not
2. Feed mass disabling at the granularity where it is needed. Currently this can be very high up, e.g. at the root of various panes. Mass disabling can be overridden in children, so this does not put any limitations on design / UX.
3. Talk in terms of the user's permissions _for the given project_ instead of talking about the user's role.


bkrmendy (approved):

editor/src/components/inspector/inspector.tsx
```diff
@@ -355,73 +357,75 @@ export const Inspector = React.memo<InspectorProps>((props: InspectorProps) => {
...
-                    <StyleSection />
-                    <WarningSubsection />
-                    <ImgSection />
-                    <EventHandlersSection />
-                  </>,
-                )}
-              </>,
-            )}
-          </FlexCol>
+          <DisableControlsInSubtree disable={!canEdit}>
```
liady, line 360: A small proposal - what if we would have a `PermissionsWrapper` that will be used like this:
```
<PermissionsWrapper permissions={['EDIT']}>
 ... inspector components
</PermissionsWrapper>
```
And inside the permissions wrapper itself it will have the disable logic?
This might be more semantic (even though more high-level). And this way whoever uses it only needs to declare which permissions are interesting, and not have to have to break down the logic themselves.
For example - we might want to have a tooltip for the disabled controllers that says why it is disabled ('you don't have an editing persmission'), and if we use a <PermissionsWrapper> we can make all the changes there instead of every location that is breaking down this disable logic.
What do you think?

editor/src/components/editor/store/permissions.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type Permissions = {
+  edit: boolean // Edit the open project via the canvas, inspector, code editor
+  comment: boolean // Add comments
+}
+
+export function usePermissions(): Permissions {
+  return {
+    edit: useIsMyProject(),
+    comment: useIsLoggedIn(),
```
gbalint, line 12: 👍 
After this PR is merged, I'm going to replace the isLoggedIn guards for the comment ui to use this permission flag.

editor/src/components/inspector/inspector.tsx
balazsbajorics: we did discuss a variant of this in the Discord! I think eventually we may introduce this but only as an abstraction _on top of_ the `DisableControlsInSubtree` – so that the internals of the controls don't need to know about anything permissions. But i'd do any kind of PermissionsWrapper as a follow-up task, once we see how many places we actually end up using `DisableControlsInSubtree`

editor/src/components/inspector/controls/color-control.tsx
```diff
@@ -82,7 +83,8 @@ export const ColorControl = React.memo((props: ColorControlProps) => {
 
   const closePopup = React.useCallback(() => setPopupOpen(false), [setPopupOpen])
 
-  const isMyProject = useIsMyProject()
+  const controlsDisabled = useControlsDisabledInSubtree()
+  const disabled = controlsDisabled
```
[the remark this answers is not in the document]
balazsbajorics: it was intentional for me to write it this way, see the rest of the controls! there is a pattern that we `||` multiple disabled conditions together, and I suspect we will change color-control in the future too, I wanted to establish a place where we can join multiple conditions

editor/src/components/inspector/inspector.tsx
Rheeseyb: I do like this idea, but agree with Balazs that it should come in a follow up (if we choose to go with it)

Rheeseyb (approved):

editor/src/components/inspector/inspector.tsx
liady: Oh, definitely. We can use the `<DisableControlsInSubtree>` as a lower-level abstraction and introduce high level wrappers later on.

PR #4744, opened 2024-01-15
Show ghost pointer when it's over a Remix scene on a different route

bkrmendy:
## Improvement
Show a ‘Ghost’ variant of the [pointer + user label] component for users whose cursors is on / over the same scene but on a different route 

## How it Works
- we store a `{ userId: { scene data label: current route } }` mapping in the room storage `see liveblocks.config.ts`. This means that we don't track the route of Remix scenes that don't have the `data-label` prop set.
- whenever a navigation happens in any of the Remix scenes, the `{ scene data label: current route }` entry is updated for the current user id (see `<RemixScenePathUpdater ... />`)
- on the multiplayer cursors side, if a multiplayer cursor hasn't moved "in a while" (this value is set to 1s now), we check whether that cursor is hovering over a Remix scene, and whether the route of that is the same for the current user and user of multiplayer cursor (see `useIsOnSameRemixRoute`)

### DIFF
--- a/editor/src/core/shared/multiplayer.ts
+++ b/editor/src/core/shared/multiplayer.ts
@@ -199,6 +199,9 @@ export function useUpdateRemixSceneRouteInLiveblocks() {
 export function useIsOnSameRemixRoute() {
   const id = useMyUserId()
   const remixSceneRoutes = useStorage((store) => store.remixSceneRoutes)
+  if (id == null) {
+    return () => true
+  }
 
   return (params: { otherUserId: string; remixSceneDataLabel: string }): boolean => {
     const remixScenePathForMe: string | null =


editor/src/components/canvas/controls/select-mode/remix-scene-label.tsx
```diff
@@ -87,12 +90,36 @@ export const RemixSceneLabelControl = React.memo<RemixSceneLabelControlProps>((p
...
+      return
+    }
+
+    updateRemixScenePathInLiveblocks({
+      sceneDataLabel: sceneDataLabelRef.current,
+      location: currentPath,
+    })
+  }, [currentPath, sceneDataLabelRef, updateRemixScenePathInLiveblocks])
+
+  return <></>
```
ruggi, line 113: why not `return null`?

editor/src/core/shared/multiplayer.ts
```diff
@@ -175,3 +182,34 @@ export function sortThreadsByDescendingUpdateTimeInPlace(
...
+      const myNewSceneRoutes = new LiveObject({ [params.sceneDataLabel]: params.location })
+      sceneRoutes.set(self.id, myNewSceneRoutes)
+    } else {
+      mySceneRoutes.set(params.sceneDataLabel, params.location)
+    }
+  }, [])
+}
+
+export function useIsOnSameRemixRoute() {
+  const { id } = useSelf()
```
ruggi, line 200: it would be best to not use `useSelf` unless you need the presence data, you can instead get the id via `useMyUserId`

editor/src/components/canvas/controls/select-mode/remix-scene-label.tsx
bkrmendy: Great idea, I completely glossed over that, added it in https://github.com/concrete-utopia/utopia/pull/4744/commits/a9e0b34476318fc3248abfc575bea8775c9af71d

editor/src/core/shared/multiplayer.ts
bkrmendy: thanks for the heads up, added it in https://github.com/concrete-utopia/utopia/pull/4744/commits/097d49be5b3746eafe015acea08bb99fad715aef

gbalint (approved):

ruggi (approved):

PR #4753, opened 2024-01-16
Ensure that the /utopia directory exists before writing to it

Rheeseyb:
**Problem:**
There exists a race condition in the VSCode build vs extension projects, whereby we could end up attempting to write project files into the `/utopia` directory before it exists. This can be triggered by sufficiently sized projects, and when it does happens the code editor fails to load.

**Fix:**
Add an `await ensureDirectoryExists(RootDir)` inside the function that initialised the bridge on the VSCode side.

### DIFF
--- a/utopia-vscode-common/package.json
+++ b/utopia-vscode-common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "utopia-vscode-common",
-  "version": "0.1.4",
+  "version": "0.1.5",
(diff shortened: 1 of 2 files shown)


bkrmendy (approved):

balazsbajorics (approved):

gbalint (approved):
🎉

PR #4755, opened 2024-01-16
Server database tests for project control.

seanparsons:
**Problem:**
There are no tests covering the database functionality in the server, which has lead to a bug in the project control functionality.

**Fix:**
The small part of the work was to add tests to the server tests.

The larger part of the work was the thread pulling necessary to support that, which involved the following pieces of work:
- Pulling out some logic for managing the test database from some existing logic, which really pull on some existing tech debt around the startup and shutdown of the server/database in tests.
- It was discovered that there were some smaller issues around parallelising tests as required.
- Switching away from Hspec to Tasty ended up fixing the two issues above, which also solved the problem that we had a mix of the two in the tests.
- In the process now rather than some awkward logic woven through the test definitions, now starting and stopping the server/database is done via a single function at the lowest level.
- Now the test executable is easily accessible when it's built so it can be run repeatedly that way if that's helpful.
- Parameters can now also be passed to the test executable via the `shell.nix` scripts.

**Commit Details:**
- Production and development servers now close their database connections actively when shutting down.
- Broke apart `sameProjectAndOwner` into a further function called `sameProject`.
- Fixed `maybeClaimExistingCollaborationControl` to handle the case where another user is currently holding the project control.
- Changed `maybeClaimCollaborationControl` and `forceClaimCollaborationControl` to take an instance of `IO UTCTime` as the way of identifying the "current" time.
- Separated out the database specific testing needs into the `Test.Utopia.Web.Database.Utils` module.
- Switched over any Hspec tests to Tasty.
- Added tests for project control in `Test.Utopia.Web.Database` module.
- Made tests sequential or parallel running as appropriate.
- Removed Hspec dependencies.
- Bumped the cabal index version so that more recent libraries are available.
- Changed `test-server-inner` to copy the executable to the `server` folder and run it from there.
- In `shell.nix` permit parameters to be passed through scripts like `test-server`.


server/test/Test/Utopia/Web/Database.hs
```diff
@@ -0,0 +1,69 @@
...
+            assertBool "First claim result should be successful." firstClaimResult
+            secondClaimResult <- maybeClaimCollaborationControl metrics pool sixteenthOfJanuaryMorning "secondowner" "testproject" "secondeditor"
+            assertBool "Second claim result should fail." $ not secondClaimResult
+        , withTestPool "when a user claims control, they can claim it a second time immediately after" $ \pool -> do
+            metrics <- getDatabaseMetrics
+            firstClaimResult <- maybeClaimCollaborationControl metrics pool sixteenthOfJanuaryMorning "testowner" "testproject" "testeditor"
+            assertBool "First claim result should be successful." firstClaimResult
+            secondClaimResult <- maybeClaimCollaborationControl metrics pool sixteenthOfJanuaryMorning "testowner" "testproject" "testeditor"
+            assertBool "Second claim result should be successful." secondClaimResult
+        , withTestPool "when a user has claimed control, another can only claim it once the expiry has been exceeded" $ \pool -> do
```
balazsbajorics, line 62: is owner here based on userID? so if I have the same project open in multiple tabs, that is multiple instances of "firstowner", right? so this "secondowner" is theoretial in the sense that the UI is not exposing this in any way, right? as in: a project in Utopia today cannot have two owners

seanparsons: Yes, owner is user ID. I should probably change this to user. In this case the check for project owner is done at the level above this logic, so the call would never be made otherwise with a non-owner user ID.

server/test/Test/Utopia/Web/Database.hs
```diff
@@ -0,0 +1,69 @@
...
+import           Test.Tasty
+import           Test.Tasty.HUnit
+import           Test.Utopia.Web.Database.Utils
+import           Test.Utopia.Web.Executors.Test
+import           Utopia.ClientModel
+import           Utopia.Web.Database
+import           Utopia.Web.Database.Types
+import           Utopia.Web.Executors.Common
+
+sixteenthOfJanuaryMorning :: IO UTCTime
```
Rheeseyb, line 33: _thinks back_
Ahhh I remember it well!

balazsbajorics (approved):

Rheeseyb (approved):

PR #4771, opened 2024-01-19
Connection tracking and avatar ordering

### LINKED ISSUE #4775: Keep track of active connections, sort avatars

ruggi:
Fix #4775 

**Problem:**

Multiplayer avatars are not sorted in a clear way / following a fixed pattern.

**Fix:**

This PR does two things:

1. It stores the player connections in the room storage as a map where the keys are player IDs and the values are arrays of `ConnectionInfo` objects.
The `ConnectionInfo` objects contain three fields: the `id` of the connection (the incremental number provided by LB), and two timestamps, one for when the session started and one for the last heartbeat of that connection (`lastSeen`).
A couple of new hooks are used for this: one that stores the connection and one that monitors the connections.
The monitoring hook makes sure that the `lastSeen` field is kept up to date, and it also cleans up expired connections from the storage so that it does not grow indefinitely.
For convenience, I added also a new feature switch (`Debug - Connections`) that will print out log lines for the connection monitoring when enabled.

2. It uses the new connection info to sort the avatars with the following logic:
	- if a user is being followed, its avatar is pinned to the left
	- the active users are shown after the followed ones, with the leftmost being the user with the oldest active connection
	- the avatars move back into place when going in/out of follow mode

https://github.com/concrete-utopia/utopia/assets/1081051/914730cf-7df7-4ccc-9f50-b06265bea952

### DIFF
--- a/editor/src/core/shared/multiplayer-hooks.tsx
+++ b/editor/src/core/shared/multiplayer-hooks.tsx
@@ -11,6 +11,10 @@ import { isLoggedIn } from '../../common/user'
 import { isFollowMode } from '../../components/editor/editor-modes'
 import type { Connections } from '../../../liveblocks.config'
 import { useStorage, useMutation } from '../../../liveblocks.config'
+import { LiveObject } from '@liveblocks/client'
+
+const ConnectionHeartbeatInterval = 10 * 1000 // ms
+const ConnectionExpiredTime = 60 * 1000 // 1 minute
 
 export function useRemixPresence(): RemixPresence | null {
   const [activeRemixScene] = useAtom(ActiveRemixSceneAtom)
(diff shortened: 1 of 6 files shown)


editor/src/core/shared/multiplayer-hooks.tsx
```diff
@@ -37,3 +55,180 @@ export function useMyUserId(): string | null {
...
+export function useIsBeingFollowed() {
+  const mode = useEditorState(
+    Substores.restOfEditor,
+    (store) => store.editor.mode,
+    'useIsBeingFollowed mode',
+  )
+
+  const isBeingFollowed = React.useCallback(
+    (connectionId: number) => {
+      return isFollowMode(mode) && mode.connectionId === connectionId
```
seanparsons, line 68: Does this also need to compare the user ID as well because of the not unique connection IDs?

ruggi: ah yes, I actually chomped it

editor/src/core/shared/multiplayer-hooks.tsx
```diff
@@ -37,3 +55,180 @@ export function useMyUserId(): string | null {
...
+      // if the current connection is not stored yet...
+      if (!selfConns.some((c) => c.id === self.connectionId)) {
+        // ...store it...
+        const now = Date.now()
+        selfConns.push({
+          id: self.connectionId,
+          startedAt: now,
+          lastSeen: now,
+        })
+        storage.get('connections').update({ [self.id]: selfConns })
```
seanparsons, line 139: I surmise that there's the possibility that something else changes this after the lookup and then we just overwrite it, not sure if we actually care, but it seems worth mentioning.

ruggi: indeed and it would be great to lock the thing... I did this to avoid having `N` update calls, but maybe they would be less error prone 🤔

editor/src/core/shared/multiplayer-hooks.tsx
```diff
@@ -37,3 +55,180 @@ export function useMyUserId(): string | null {
...
+
+  const updateLastSeen = useMutation(
+    ({ storage, self }) => {
+      if (!isLoggedIn(loginState)) {
+        return
+      }
+
+      const selfConns: ConnectionInfo[] = storage.get('connections').get(self.id) ?? []
+      // if the current connection is stored, bump its lastSeen field
+      const connIndex = selfConns.findIndex((c) => c.id === self.connectionId)
```
seanparsons, line 211: Doesn't look like you use `connIndex` as an index, so you could do:
```suggestion
      const doUpdate = selfConns.some((c) => c.id === self.connectionId)
```

editor/src/core/shared/multiplayer-hooks.tsx
seanparsons: I don't _think_ it's necessary, because we're just gonna get variations on a theme of the values I believe. They'd just get slightly different times.

seanparsons (approved):

editor/src/core/shared/multiplayer-hooks.tsx
ruggi: ahhh stupid remnant of a previous impl. thanks!

gbalint (approved):

PR #4774, opened 2024-01-19
Logged out handling for collaboration features.

seanparsons:
**Problem:**
The collaboration functionality makes server calls often without the user initiating anything beyond loading the editor, which are being triggered when the user isn't logged on causing those services to reject the requests. Similarly some of these calls and those like the project metadata calls are being triggered in tests, which we definitely don't want as that could cause weird issues if a server is actually running.

**Fix:**
For when the editor runs normally, the main change is a check to the login state in the editor and only allow those endpoints to be called when the user is logged in. Which has been done at the level above the functions that make those service calls.

For tests, the minor part of this fix was making sure that all of the calls that are being triggered (like claiming/snatching control or the project metadata endpoint) are guarded with a check to `IS_TEST_ENVIRONMENT`.

Around the above changes are mostly fixes to maintain certain expectations about the results that are given in certain states, like when the user isn't logged in but the project still has a project ID, they should still be the "owner" of the project.

**Commit Details:**
- Test dispatch function now waits for 2 calls to update the project server state to apply before continuing.
- Added `loggedIn` parameter to `checkProjectOwned`.
- `checkProjectOwned` now handles if the user is not logged in and if it's running in a test environment to return true against `isOwner`. This is because actions need to run but at a point that does not regularly occur when running normally as the user is either logged in or the project exists locally.
- Added a `loggedIn` condition in `CollaborationStateUpdater`.
- `allowedToEditProject` and `useAllowedToEditProject` now check the login status of the user.
- Removed a call to `updateProjectServerStateInStore` that was running on any change to `SET_LOGIN_STATE`, because it was effectively running at a random point in time.
- `getProjectServerState` now doesn't itself handle `IS_TEST_ENVIRONMENT` as that handling has been pushed down into the various functions it calls.
- Added timestamps to `test-browser-full-output`.
- `getLoginState`, `checkProjectOwnership` and `fetchProjectMetadata` now all check `IS_TEST_ENVIRONMENT` so that they don't try to make calls to the server during tests.

### DIFF
--- a/editor/src/components/canvas/ui-jsx.test-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx.test-utils.tsx
@@ -106,7 +106,7 @@ import {
   treeToContents,
 } from '../assets'
 import { testStaticElementPath } from '../../core/shared/element-path.test-utils'
-import { createFakeMetadataForParseSuccess } from '../../utils/utils.test-utils'
+import { createFakeMetadataForParseSuccess, wait } from '../../utils/utils.test-utils'
 import {
   mergeWithPrevUndo,
   saveDOMReport,
@@ -146,7 +146,10 @@ import { carryDispatchResultFields } from './editor-dispatch-flow'
 import type { FeatureName } from '../../utils/feature-switches'
 import { setFeatureEnabled } from '../../utils/feature-switches'
 import { unpatchedCreateRemixDerivedDataMemo } from '../editor/store/remix-derived-data'
-import { emptyProjectServerState } from '../editor/store/project-server-state'
+import {
+  emptyProjectServerState,
+  getUpdateProjectServerStateInStoreRunCount,
+} from '../editor/store/project-server-state'
 
 // eslint-disable-next-line no-unused-expressions
 typeof process !== 'undefined' &&
@@ -626,6 +629,8 @@ label {
     </React.Profiler>,
   )
 
+  const beforeLoadUpdateStoreRunCount = getUpdateProjectServerStateInStoreRunCount()
+
   await act(async () => {
     await new Promise<void>((resolve, reject) => {
       void load(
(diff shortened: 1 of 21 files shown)


editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -626,6 +629,8 @@ label {
     </React.Profiler>,
   )
 
+  const beforeLoadUpdateStoreRunCount = getUpdateProjectServerStateInStoreRunCount()
```
seanparsons, line 632: I'm going to add a comment to the code that explains this...

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -656,6 +661,11 @@ label {
     })
   })
 
+  while (getUpdateProjectServerStateInStoreRunCount() <= beforeLoadUpdateStoreRunCount + 1) {
```
seanparsons, line 664: ...and what it does in connection to this.

editor/src/components/canvas/ui-jsx.test-utils.tsx
seanparsons: Fixed in 1345ea052.

editor/src/components/canvas/ui-jsx.test-utils.tsx
seanparsons: Fixed in 1345ea052.

editor/src/core/model/element-metadata.spec.browser2.tsx
```diff
@@ -37,6 +37,7 @@ import { elementOnlyHasTextChildren } from './element-template-utils'
 import { BakedInStoryboardUID, BakedInStoryboardVariableName } from './scene-utils'
 import type { HugPropertyWidthHeight } from '../shared/element-template'
 import { contentsToTree } from '../../components/assets'
+import { CanvasContainerID } from '../../components/canvas/canvas-types'
```
ruggi, line 40:
```suggestion
```

ruggi (approved):

editor/src/core/model/element-metadata.spec.browser2.tsx
seanparsons: Ah yes!

PR #4786, opened 2024-01-23
Ensure scene id props exist

bkrmendy:
## Problem
We link threads to scenes by adding the value of a scene's `id` prop to the thread metadata. If a scene doesn't have an `id` prop, we won't be able to link comments to it. Not all scenes may have an `id` prop, so we need a mechanism that adds the `id` prop to scenes without it.

## Fix
Walk the contents tree after each dispatch, and ensure that all scenes have `id`s and their values are all unique. If a scene doesn't have an `id` prop, or another scene already uses it, it's overwritten by the scene's uid.

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -1,4 +1,8 @@
-import { PERFORMANCE_MARKS_ALLOWED, PRODUCTION_ENV } from '../../../common/env-vars'
+import {
+  IS_TEST_ENVIRONMENT,
+  PERFORMANCE_MARKS_ALLOWED,
+  PRODUCTION_ENV,
+} from '../../../common/env-vars'
 import { isParseSuccess, isTextFile } from '../../../core/shared/project-file-types'
 import {
   codeNeedsParsing,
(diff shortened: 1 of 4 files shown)


editor/src/components/editor/store/dispatch.tsx
```diff
@@ -798,6 +814,141 @@ function applyProjectChangesToEditor(
   }
 }
 
+function walkJSXElementChild(
```
balazsbajorics, line 817: this definitely does not belong in dispatch.tsx!

Rheeseyb (approved):

editor/src/components/editor/store/dispatch.tsx
Rheeseyb: Good point, and in fact we already have a `walkElement` function that can maybe be used instead https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/element-template.ts#L2245

bkrmendy: moved it and renamed it in https://github.com/concrete-utopia/utopia/pull/4786/commits/b6aa64e7e2b5e74a04dbdf89142ddd2ae676c6aa

PR #4789, opened 2024-01-23
log waits in puppeteer tests

bkrmendy:
## Description
This PR adds logging to the commenting and collaboration tests, so they are easier to debug if they fail

### DIFF
--- a/puppeteer-tests/src/comments/comment-utils.ts
+++ b/puppeteer-tests/src/comments/comment-utils.ts
@@ -2,6 +2,7 @@ import type { ElementHandle, Page } from 'puppeteer'
 import type { UtopiaPuppeteerBrowser } from './test-utils'
 
 export const TIMEOUT = 120000
+export const PUPPETEER_TIMEOUT = 30000
 
 const BRANCH_NAME = process.env.BRANCH_NAME ? `&branch_name=${process.env.BRANCH_NAME}` : ''
 const BASE_URL = process.env.BASE_URL ?? 'http://localhost:8000'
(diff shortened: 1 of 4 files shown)


puppeteer-tests/src/comments/comment-utils.ts
```diff
@@ -1,47 +1,61 @@
...
 const BRANCH_NAME = process.env.BRANCH_NAME ? `&branch_name=${process.env.BRANCH_NAME}` : ''
 const BASE_URL = process.env.BASE_URL ?? 'http://localhost:8000'
 
+export async function getElementWithSelector(
+  page: Page,
+  selector: string,
+): Promise<ElementHandle<Element> | null> {
+  console.info('waiting for element with selector', selector)
+  const element = await page.waitForSelector(selector)
+  return element!
```
seanparsons, line 16: Is this forcing the result to not be null? When the result type allows that.

seanparsons (approved):

puppeteer-tests/src/comments/comment-utils.ts
bkrmendy: good catch, I wanted it force it not to be null, I'll update the return type

bkrmendy: removed the `| null` in https://github.com/concrete-utopia/utopia/pull/4789/commits/0dc014050e525ee8bc39999483b35a2a5e9bbf16

seanparsons: I'd recommend using `forceNotNull` as I believe the `!` just coerces the type not the value itself.

PR #4791, opened 2024-01-24
Store canvas position of scene comments.

gbalint:
**Problem:**
When a scene is deleted, the scene comments are orphaned, and we are not able to position them on the canvas.

**Fix:**
Change the liveblocks thread representation to store the local position of the thread in the scene (`sceneX` and `sceneY`) and the canvas position (`x` and `y`) both.

Note, that we can only add strings/numbers/booleans to the thread metadata type, so there is no way to create a more structured typing.

The advantage of this representation is that `x` and `y` always represents the canvas position of the thread, which was a more ambiguous earlier (canvas position for canvas comments and local position for scene comments).

So we do the best effort to position the comment thread in its scene, but if the scene is not available for some reason (e.g. it has been deleted), we still put it on the same canvas position. We do not delete the scene information, so when the user presses undo to put back the scene, the comment thread automatically becomes a scene thread again.

When the scene is moved, we update the canvas location of its scene comments using a React component called CommentMaintainer (which doesn't render anything, its role is just this maintenance step).

### DIFF
--- a/editor/liveblocks.config.ts
+++ b/editor/liveblocks.config.ts
@@ -98,10 +98,11 @@ export type RoomEvent = ControlChangedRoomEvent
 export type ThreadMetadata = {
   // quote: string;
   // time: number;
-  type: 'canvas'
-  x: number // x and y is global when sceneId is undefined, and local to the scene when sceneId is not null
+  x: number
   y: number
   sceneId?: string
+  sceneX?: number
+  sceneY?: number
   remixLocationRoute?: string
   resolved: boolean
 }
--- a/editor/liveblocks.config.ts
+++ b/editor/liveblocks.config.ts
@@ -107,6 +107,22 @@ export type ThreadMetadata = {
   resolved: boolean
 }
 
+export type SceneThreadMetadata = {
+  // quote: string;
+  // time: number;
+  x: number
+  y: number
+  sceneId: string
+  sceneX: number
+  sceneY: number
+  remixLocationRoute?: string
+  resolved: boolean
+}
+
+export function isSceneThreadMetadata(metadata: ThreadMetadata): metadata is SceneThreadMetadata {
+  return metadata.sceneId != null && metadata.sceneX != null && metadata.sceneY != null
+}
+
 export const {
   suspense: {
     RoomProvider,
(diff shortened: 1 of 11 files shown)


editor/liveblocks.config.ts
```diff
@@ -98,14 +98,31 @@ export type RoomEvent = ControlChangedRoomEvent
 export type ThreadMetadata = {
   // quote: string;
   // time: number;
-  type: 'canvas'
-  x: number // x and y is global when sceneId is undefined, and local to the scene when sceneId is not null
+  x: number
   y: number
   sceneId?: string
+  sceneX?: number
```
balazsbajorics, line 104: I don't like this! what's the point of these being optional here and then have a separate SceneThreadMetadata where they are declared as non optional?

gbalint: The reason is that canvas comments don't have these properties (sceneId, sceneX, sceneY), and we are only allowed to have string, number and boolean properties here, so no way to create a structured type.

balazsbajorics: `why not `CanvasThreadMetadata` without these fields and `SceneThreadMetadata` with these fields, and then `ThreadMetadata = CanvasThreadMetadata | SceneThreadMetadata`

but better yet, why not make it a discriminated union type with a type: 'canvas' | 'scene'? I see on master there was some notion to do that

gbalint: I can not use discriminated union types because I can only add string/boolean/number fields.
Undiscrimated union types are not that nice to work with, but you can be right maybe this solution is better: 68b241864f82462319480bf0d845bfdef0f00511

gbalint: It turned out the problem was not discriminated union types, but that I used interfaces and not types.... Here is the proper solution: dcc67bdbb4071c8c5dd7c6c7e6b4db5f94202cf2

balazsbajorics (approved):

editor/src/core/commenting/comment-maintainer.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return null
+})
+CommentMaintainerInner.displayName = 'CommentMaintainerInner'
+
+function useMaintainComments() {
+  const { threads } = useThreads()
+  const scenes = useScenes()
+  const editThreadMetadata = useEditThreadMetadata()
+
+  threads.forEach(async (t): Promise<void> => {
```
seanparsons, line 40: I was about to ask if subsequent calls to this can trample over each other, but I don't _think_ they will because the promises would all be queued up in order.

seanparsons (approved):

PR #4795, opened 2024-01-25
Link comments to scenes based on the `commentId` prop

bkrmendy:
## Problem
The relationship between the `id` prop on Scenes and the position of Scene comments is not evident. If a user deletes the `id` prop from a scene, any comments attached to that scene would be reparented to the canvas, without any indication from us.

## Fix
Link the comments on scenes based on the value of the `commentId` prop, which better communicates the purpose of the prop


editor/src/core/model/scene-id-utils.ts
```diff
@@ -11,10 +11,10 @@ import * as PP from '../shared/property-path'
 import { assertNever } from '../shared/utils'
 import { isSceneAgainstImports, isRemixSceneAgainstImports } from './project-file-utils'
 
-const IdPropName = 'id'
+export const SceneCommentIdPropName = 'data-comment-id'
```
maltenuhn, line 14: Since it's ONLY for scenes, could we make this just `commentId`? Scene _is_ our component, so we we can make it a bit less clunky

bkrmendy: sure, I updated it in https://github.com/concrete-utopia/utopia/pull/4795/commits/7ebe371b3241a3cb295dc4d1d53d79b422cbbee3.

The reason I went with `data-comment-id` is because it fit the mold of the other Utopia-specific props on Scene (`data-label`, `data-uid`), and using `data-comment-id` didn't make it necessary to tweak the code of `Scene` (without the tweaking, a React error is triggered: `React does not recognize the 'commentId' prop on a DOM element`, since all props on Scene are forwarded to the underlying `div`).

editor/src/core/model/scene-id-utils.ts
```diff
@@ -11,10 +11,10 @@ import * as PP from '../shared/property-path'
 import { assertNever } from '../shared/utils'
 import { isSceneAgainstImports, isRemixSceneAgainstImports } from './project-file-utils'
 
-const IdPropName = 'id'
+export const SceneCommentIdPropName = 'commentId'
```
liady, line 14: small suggestion - maybe `comment-target-id`?
(since it's not the id of a comment, but the id of the scene as a comment target)

editor/src/core/third-party/utopia-api-components.ts
```diff
@@ -81,7 +81,11 @@ const BasicUtopiaSceneDescriptor = (
           jsxElementWithoutUID(
             name,
             [
-              jsxAttributesEntry('id', jsExpressionValue('scene', emptyComments), emptyComments),
+              jsxAttributesEntry(
+                'commentId',
```
liady, line 85: there isn't a way to use `SceneCommentIdPropName` here, right?

liady (approved):

editor/pnpm-lock.yaml
```diff
@@ -7831,7 +7831,7 @@ packages:
     dev: true
 
   /component-indexof/0.0.3:
-    resolution: {integrity: sha1-EdCRMSI5648yyPJa6csAL/6NPCQ=}
+    resolution: {integrity: sha512-puDQKvx/64HZXb4hBwIcvQLaLgux8o1CbWl39s41hrIIZDl1lJiD5jc22gj3RBeGK0ovxALDYpIbyjqDUUl0rw==}
```
seanparsons, line 7834: Huh?

seanparsons (approved):

PR #4798, opened 2024-01-26
Convert liveblocks thread metadata to our own typed representation

gbalint:
**Problem:**
Liveblocks does not allow us to create structured typing for the thread metadata.

**Fix:**
- Create our own discriminated union based type system for threads.
- Create conversion functions from liveblocks to utopia and the other way around
- Create a converter from utopia to liveblocks for the partial types, which is useful for editing the metadata.

### DIFF
--- a/editor/src/core/commenting/comment-types.tsx
+++ b/editor/src/core/commenting/comment-types.tsx
@@ -22,31 +22,13 @@ export type SceneThreadMetadata = BaseThreadMetadata & {
 }
 
 export function canvasThreadMetadata(
-  position: CanvasPoint,
-  resolved: boolean = false,
+  data: Omit<CanvasThreadMetadata, 'type'>,
 ): CanvasThreadMetadata {
-  return {
-    type: 'canvas',
-    position: position,
-    resolved: resolved,
-  }
+  return { ...data, type: 'canvas' }
 }
 
-export function sceneThreadMetadata(
-  position: CanvasPoint,
-  sceneId: string,
-  scenePosition: LocalPoint,
-  remixLocationRoute?: string,
-  resolved: boolean = false,
-): SceneThreadMetadata {
-  return {
-    type: 'scene',
-    position: position,
-    sceneId: sceneId,
-    scenePosition: scenePosition,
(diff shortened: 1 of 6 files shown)


editor/src/core/commenting/comment-types.tsx
```diff
@@ -0,0 +1,132 @@
...
+}
+
+type PartialNullable<T> = {
+  [P in keyof T]?: T[P] | null | undefined
+}
+
+export function utopiaThreadMetadataToLiveblocksPartial(
+  metadata: PartialNullable<UtopiaThreadMetadata>,
+): PartialNullable<ThreadMetadata> {
+  if (metadata.type === 'scene') {
```
bkrmendy, line 112: I would go with a switch/case here so that `assertNever` can throw an error if we ever add a new comment type

editor/src/core/commenting/comment-types.tsx
```diff
@@ -0,0 +1,132 @@
...
+  resolved: boolean = false,
+): CanvasThreadMetadata {
+  return {
+    type: 'canvas',
+    position: position,
+    resolved: resolved,
+  }
+}
+
+export function sceneThreadMetadata(
```
ruggi, line 35: what about passing the parameters as an `Omit` for the `SceneThreadMetadata`, excluding the `type`? imho it would require less fiddling around going forward

```ts
function sceneThreadMetadata(data: Omit<SceneThreadMetadata, "type">) {
	return { ...data, type: 'scene' }
}
```

(same for the canvas metadata constructor)

ruggi (approved):

editor/src/core/commenting/comment-types.tsx
gbalint: assertNevert doesn't work here because of the Partial type

bkrmendy (approved):

editor/src/core/commenting/comment-types.tsx
gbalint: Fixed in 22f3b2b8b4afe10bd4e81515a7c806d3ffa08d61

PR #4799, opened 2024-01-26
Support images and assets in collaboration.

seanparsons:
**Problem:**
The editor does not handle image and asset file updates when using the collaboration features.

**Fix:**
Support was added for the two cases of image and asset files, whereby the adding/updating of entries via yjs is done. Albeit slightly differently to the way we handle parsed text files which are the nearest analogue, where we send a selection of fields if each field has changed. For images and assets we send the entire entry from the project contents, but only if the `base64` value is empty. Without that property the entries are just a handful (at most) of primitive fields which aren't very large at all.

Other than this some small changes were essential to the internals of how we handle `UPDATE_FILE` actions so that we didn't have to create a specific action for updates to those files coming from the collaboration logic.

**Commit Details:**
- Added `fromCollaboration` property to `UpdateFile` action.
- `applyFileChangeToMap` extended to include `IMAGE_FILE` and `ASSET_FILE` handling.
- Added `clearIfNotExpectedType` utility function.
- Added functions for populating the project contents map from `ImageFile` or `AssetFile` values.
- Added functions for creating `EditorAction` values from the yjs representation.
- Extended `CollabFile` to include union cases of `CollabImageFile` and `CollabAssetFile`.
- Changed `gatedActions` to check the `fromCollaboration` property of `UPDATE_FILE` actions.
- Modified `updateFileIfPossible` to prevent infinite looping of image and asset file updates.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-code-execution.spec.browser2.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-code-execution.spec.browser2.tsx
@@ -119,7 +119,7 @@ describe('Updating a transitive dependency', () => {
       (oldFile as TextFile).versionNumber + 1,
     )
 
-    await dispatch([updateFile(indirectFilePath, updatedIndirectFile, false, false)], true)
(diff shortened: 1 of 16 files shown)


ruggi (approved):
\o

PR #4801, opened 2024-01-29
Prevent Slate errors when adding comments.

seanparsons:
**Problem:**
When adding a comment, an exception is thrown from Slate when confirming the comment contents.

**Cause:**
In the `blur` handler of `ReactEditor` in `react-slate`, there's a lookup which attempts to find the underling DOM node of the text editor in question. For whatever reason (potentially related to the weakmap it uses under the hood of this) it is unable to match the editor and then throws an exception as a result.

**Fix:**
This doesn't really fix the failure because it seems like that potentially involves fixing some of the internals of slate and it's React integration, but merely hides the exception so that it doesn't get surfaced to the user.

**Commit Details:**
- Adds patch that catches exception in the `blur` handler of `slate-react`.

### DIFF
--- a/editor/package.json
+++ b/editor/package.json
@@ -413,7 +413,8 @@
       "react@18.1.0": "patches/react@18.1.0.patch",
       "reselect@4.1.7": "patches/reselect@4.1.7.patch",
       "use-context-selector@1.3.9": "patches/use-context-selector@1.3.9.patch",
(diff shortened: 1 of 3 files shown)


ruggi (approved):

seanparsons:
A better fix has since been suggested and implemented.

PR #4802, opened 2024-01-29
Simple error handling for multiplayer

gbalint:
**Problem:**
In case of multiplayer connection errors we just silently don't render any multiplayer UI.

**Fix:**
I use the liveblocks hooks `useLostConnectionListener` and ` useErrorListener` to listen to different connection errors.
I also added an extra guard in `useCanComment` to not render comment UI when the connection is lost because calling the `useThreads` hook triggers a liveblocks API request all the time, which really floods the browser with pending requests.


editor/src/core/shared/multiplayer-hooks.tsx
```diff
@@ -250,3 +258,49 @@ export function useMonitorConnection() {
...
+              notice('Reconnecting to other users...', 'WARNING', true, 'reconnecting-liveblocks'),
+            ),
+          ])
+        }
+        break
+
+      case 'restored':
+        dispatch([
+          removeToast('lost-liveblocks-connection'),
+          removeToast('reconnecting-liveblocks'),
```
bkrmendy, line 282: I'd put these in a const

bkrmendy (approved):

Rheeseyb (approved):

PR #4807, opened 2024-01-30
Remix BFF setup

ruggi:
This PR introduces the Remix BFF.

- The BFF code lives inside the `utopia-remix` folder.
- The needed configuration is explained in `utopia-remix/readme.md`.
- On localhost, the BFF runs on port `8002`.
- The BFF is **only enabled on localhost**, as per the changes in `env-vars.ts`.
- The BFF proxies all API calls directly to the actual backend server (8002 → 8001), **without added or changed logic**.
- The Remix app output is available in the tab `3` of tmux.
Screenshot 2024-01-30 at 12 28 29 PM

**Notes**
- For now, the BFF is _only_ enabled on localhost - I think it makes sense for now, as we figure out the various integration further


shell.nix
```diff
@@ -549,20 +555,22 @@ let
         new-window -n "Scratchpad" \; \
         new-window -n "Server" \; \
         send-keys -t :2 watch-server C-m \; \
+		new-window -n "Remix" \; \
```
seanparsons, line 558: Weird indentation.

shell.nix
```diff
@@ -591,20 +599,22 @@ let
...
```
seanparsons, line 602: Here too...

shell.nix
ruggi: damn tabs<>spaces

utopia-remix/app/routes/v1.github.authentication.finish.tsx
```diff
@@ -0,0 +1,10 @@
+import { LoaderFunctionArgs } from "@remix-run/node";
+import { handle, handleOptions } from "../util/api.server";
+import { proxy } from "../util/proxy.server";
+
+export async function loader(args: LoaderFunctionArgs) {
+  return handle(args.request, {
+    OPTIONS: handleOptions,
+    GET: proxy,
+  });
+}
```
balazsbajorics, line 10: since this exact `loader` is present in all the "just proxy to Haskell" endpoints, shouldn't we use something with even less code repetition? 

is something like this possible?

```
import { proxyToHaskellBackendLoader } from "../util/proxy.server";
export const loader = proxyToHaskellBackendLoader
```

ruggi: Yeah I also thought about making it a little less verbose (we _always_ need an `OPTIONS` loader because of CORS), but that loader is not always present, some proxy the `GET` methods (like this one) as well.
The reason for it is that since we're going route-by-route (and route-by-method for the handlers) it makes sense, imo, to be explicit in what methods we're proxying and what we are not.

We can of course use another function to which you only pass the methods you want to proxy, but it would be basically the same as using the `handle` function directly, with yet-another-layer of indirection 🤔

balazsbajorics: > and route-by-method for the handlers

can I challenge this? I think it would be very confusing if we change GET to handled by remix and keep the POST method for the haskell server

ruggi: I don't follow 🤔 we're not changing anything here, the method checks have two purposes:
1. you are not allowed to call a different method than the exact one we allow in haskell
2. if we need to customize/have additional endpoints/etc. we can do something like this, which imo is very readable

```ts
return handle(req, () => {
  GET: listDogs(),
  POST: createDog(),
  PUT: updateDog(),
  DELETE: deleteDog(),
}
```

utopia-remix/public/.keep
Rheeseyb, line 1: Do we need to keep this?

utopia-remix/remix.config.js
```diff
@@ -0,0 +1,8 @@
+/** @type {import('@remix-run/dev').AppConfig} */
+export default {
+  ignoredRouteFiles: ["**/.*"],
+  // appDirectory: "app",
+  // assetsBuildDirectory: "public/build",
+  // publicPath: "/build/",
+  // serverBuildPath: "build/index.js",
```
Rheeseyb, line 7: Should these still be commented out?

utopia-remix/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "build": "remix build",
+    "dev": "remix dev --manual",
+    "lint": "eslint --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint .",
+    "start": "remix-serve ./build/index.js",
+    "typecheck": "tsc",
+    "format": "prettier --write ."
+  },
+  "prettier": {},
+  "dependencies": {
+    "@remix-run/css-bundle": "^2.5.1",
```
Rheeseyb, line 16: Please can you remove all of the `^` prefixes from these dependencies so that we don't end up with build discrepancies between team members

utopia-remix/.eslintrc.cjs
```diff
@@ -0,0 +1,85 @@
+/**
```
Rheeseyb, line 1: I'm assuming this was created by default when setting up the remix project? If so please can we bin this off and use the eslint config from the editor (https://github.com/concrete-utopia/utopia/blob/master/editor/.eslintrc-lite.js and https://github.com/concrete-utopia/utopia/blob/master/editor/.eslintrc.js)

utopia-remix/public/.keep
ruggi: I thought it was a good way of reminding myself something like "ah yeah, that's where I should put the favicon", but if you think it's redundant I'll chuck it

utopia-remix/.eslintrc.cjs
ruggi: correct, it was default remix config. Should I just copypaste the other config or is there a way to link them (or do they even bubble down)?

utopia-remix/remix.config.js
ruggi: this came from the default remix config, so I did not touch it

utopia-remix/.gitignore
```diff
@@ -0,0 +1,7 @@
+node_modules
+
+/.cache
+/build
+/public/build
+.env
+vegeta
```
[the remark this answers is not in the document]
ruggi: I figured it's way better to do it now and forget about it, rather than forgetting about it now and committing a million outputs later :D

utopia-remix/package.json
ruggi: https://github.com/concrete-utopia/utopia/pull/4807/commits/799a48208aea95a6c1412f5ecb72781dae3e8321

Rheeseyb (approved):
This is beautiful!

utopia-remix/.eslintrc.cjs
Rheeseyb: This is a bit gross, but we run the editor's config on all other projects like so https://github.com/concrete-utopia/utopia/blob/master/editor/package.json#L46

website-next/components/common/env-vars.ts
```diff
@@ -29,6 +25,39 @@ export const IS_TEST_ENVIRONMENT: boolean =
...
+
+export const BACKEND_TYPE: BackendType = DEVELOPMENT_ENV ? 'bff' : 'direct'
+
+export function isBackendBFF(): boolean {
+  return BACKEND_TYPE === 'bff'
+}
+
+export const UTOPIA_BACKEND =
+  (PRODUCTION_OR_STAGING_CONFIG
+    ? BASE_URL
```
balazsbajorics, line 48: is this right? IIUC the original code was `BASE_URL + 'v1/'`

gbalint (approved):

website-next/components/common/env-vars.ts
ruggi: yep, in this case it's the same (`BASE_URL + v1/`) on everything except localhost, where it's `scheme://<host>:<port>/ + v1/`

utopia-remix/app/util/methods.server.ts
```diff
@@ -0,0 +1,6 @@
+const methods = ["GET", "POST", "PUT", "DELETE", "OPTIONS"] as const;
+export type Method = (typeof methods)[number];
+
+export function isMethod(s: string): s is Method {
```
seanparsons, line 4: What is this for? It misses out on some standard ones like `PATCH` at least...

seanparsons (approved):

utopia-remix/app/util/methods.server.ts
ruggi: > It misses out on some standard ones like PATCH at least 

sure, I just included the methods we actually support in the endpoints, so I omitted PATCH/HEAD and the other ones, but I can add them (`PATCH` makes sense to add, probably `HEAD` or `CONNECT` don't 😅 )

> What is this for?

I added that utility method so we can instantiate type-safely the handler methods (https://github.com/concrete-utopia/utopia/blob/799a48208aea95a6c1412f5ecb72781dae3e8321/utopia-remix/app/util/api.server.ts#L29), and then check them as well (https://github.com/concrete-utopia/utopia/blob/799a48208aea95a6c1412f5ecb72781dae3e8321/utopia-remix/app/util/api.server.ts#L36) against the same type. WDYT?

ruggi: https://github.com/concrete-utopia/utopia/pull/4807/commits/042acfb2c3c3dab3712585533c06e2d2e6c72286

PR #4809, opened 2024-01-30
Run comments & collaboration tests on staging deploys

bkrmendy:
## Problem
The Puppeteer tests for commenting and collaboration don't run after a new version of Utopia has been deployed to staging.

## Fix
Run the tests on the `trigger-staging-checks` event (renamed from `trigger-staging-screenshot` in https://github.com/concrete-utopia/utopia-deploy/pull/48)

### DIFF
--- a/.github/workflows/check-staging.yml
+++ b/.github/workflows/check-staging.yml
@@ -93,7 +93,7 @@ jobs:
       - name: Run Comments Test
         id: run-comments-test
         env:
-          BASE_URL: 'https://utopia.fish'
+          BASE_URL: 'https://utopia.pizza'
         run: |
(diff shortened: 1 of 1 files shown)


balazsbajorics (approved):

Rheeseyb (commented):
This is good but right now it won't notify us of failures. For that, you'd need to post a notification to discord like we do here: <https://github.com/concrete-utopia/utopia/blob/1cb424a3f71626a75e3f6a36fd574b1f4c641bae/.github/workflows/master-pushes.yml#L161-L174>

Rheeseyb (approved):

PR #4810, opened 2024-01-30
refactor VariableData to contain VariableMetadata

balazsbajorics:
Today `EditorState.variablesInScope` contains a map like this:
```
{
  'sb/scene/pg:app': {
    'variableName': 15 // we store a map where the key is the variable's name, and the value is the spied value
  }
}
```

this is making it hard to store extra information about a given variable, and it also makes it hard to store the variable name for a variable in scope where we do not want to store the spied value (e.g. because it may be a function reference etc)

This PR is changing the code to this:
```
{
  'sb/scene/pg:app': {
    'variableName': { spiedValue: 15 }
  }
}
```

this extra wrapper object helps in the following ways:
- makes refactoring easier, because the outer map is no longer name -> any, so it makes it possible in the future to search for usages of spiedValue
- creates a place where we can add extra fields in the future
- makes it possible to make spiedValue optional in the future (if we don't want to store a spiedValue for something)

### DIFF
--- a/editor/src/components/shared/scoped-variables.ts
+++ b/editor/src/components/shared/scoped-variables.ts
@@ -141,7 +141,7 @@ function getContainerPropsValue(
       if (isJSXAttributesEntry(prop)) {
         const value = simplifyAttributeIfPossible(prop.value)
         if (isJSXAttributeValue(value)) {
-          runtimeProps[prop.key] = value.value
+          runtimeProps[prop.key] = { spiedValue: value.value as unknown }
         }
       }
     })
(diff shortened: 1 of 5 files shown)


editor/src/components/shared/scoped-variables.ts
```diff
@@ -141,7 +141,7 @@ function getContainerPropsValue(
       if (isJSXAttributesEntry(prop)) {
         const value = simplifyAttributeIfPossible(prop.value)
         if (isJSXAttributeValue(value)) {
-          runtimeProps[prop.key] = value.value
+          runtimeProps[prop.key] = { spiedValue: value.value as unknown }
```
balazsbajorics, line 144: originally value.value is `any` here which defeated the type checker, so I put this `as unknown` here as hardening.. maybe it's just lipstick on a pig though

bkrmendy (approved):

gbalint (approved):

editor/src/components/shared/scoped-variables.ts
liady: it's definitely better than `any`, thanks for this fix

PR #4812, opened 2024-01-30
chore/remove `domtoimage`

liady:
**Problem:**
`domtoimage` caused issues when it was imported as an external inside a web worker.

**Fix:**
The underlying externals issue still remains, but we removed `domtoimage` as we don't use it anymore (and it was our only external dependecy)


balazsbajorics (approved):

Rheeseyb (approved):

editor/pnpm-lock.yaml
seanparsons, line 1: Weird, the lock file got much bigger as a result of removing something...

seanparsons (approved):

editor/pnpm-lock.yaml
balazsbajorics: oh this _is_ weird!

balazsbajorics: Liad did you run this in our nix environment? maybe you had a different node version or a different `pnpm` version that caused the lockfile to be re-done

balazsbajorics (changes requested):
blocking this until we figure out the huge lockfile change

editor/pnpm-lock.yaml
liady: yes, I ran it in our nix environment, I removed some dependencies so expected a lockfile diff, but you're right that the lockfile result seems weird. Will check it again.

liady: Seems like it's mostly prettier changes 
image
image

balazsbajorics (approved):

editor/pnpm-lock.yaml
balazsbajorics: thanks!

PR #4813, opened 2024-01-30
Better unique scene comment generation

bkrmendy:
## Problem
Unique scene id genaration breaks down in the following case:
- if a scene uses its `uid` as is `commentId`
- this scene is duplicated, and the duplicate is inserted before the original scene
In this case, the duplicate's `commentId` will be unique, but the original scene's `commentId` won't be, because of its duplicate. When falling back the element's uid to provide a unique commentId, the code doesn't stop and check whether that fallback commentId is really unique.

## Fix
When falling back to the element's uid, check whether that value is already in use. If it is, generate a unique id based on the fallback value.


balazsbajorics (approved):
I guess maybe in the future we want these ids to be more like the three-letter uid?

bkrmendy:
> I guess maybe in the future we want these ids to be more like the three-letter uid?

Absolutely, `generateUniqueSceneCommentId`'s interface can accommodate that, so we can swap the implementation out for another one anytime

gbalint (approved):

Rheeseyb (approved):

seanparsons (approved):

liady (approved):

PR #4815, opened 2024-01-31
Better checks against reparenting a component into its own definition

bkrmendy:
## Problem
It's possible to reparent elements into their own definitions. Consider the following project:
```typescript
function Container({ children, style }) {
  return (
    <div data-uid='0cd' style={style}>
      {children}
    </div>
  )
}

export var storyboard = (
  <Storyboard data-uid='aab'>
    <Scene>
      <Container />
    </Scene>
  </Storyboard>
)
```

In this project, if the Scene element is duplicated, and reparented into the root div of `Container`, then `Container` would start an infinite loop when rendered.

## Root cause
The (incorrect) code that checks against reparenting a component into its own definition doesn't check the children of the element being reparented. In the example above, a Scene is being reparented, but its child, the Container causes the trouble.

## Fix
Check all children of the element being reparented


editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.spec.browser2.tsx
```diff
@@ -933,6 +866,54 @@ export var ${BakedInStoryboardVariableName} = (props) => {
       ),
     )
   })
+  it('does not reparente scene into other component in scene', async () => {
```
Rheeseyb, line 869:
```suggestion
  it('does not reparent scene into other component in scene', async () => {
```

bkrmendy: thanks 😬

Rheeseyb (approved):

gbalint (approved):

PR #4817, opened 2024-01-31
Assorted Commenting test improvements

balazsbajorics:
- moving test timeout to global jest config instead of specifying it in every single test `it()`
- printing every browser console message to the node.js console for better debugging (should probably put all console messages behind a verbose flag)
- change the chrome window size to match the tab viewport size
- I noticed some `it()`s were weirdly comma separated, fixed that which changed their indentation, please use "hide whitespace"
- while I was there, I added a newline between every `it()` because why weren't there before?


bkrmendy (approved):

gbalint (approved):

Rheeseyb:
> while I was there, I added a newline between every it() because why weren't there before?
I wonder if prettier removed newlines because of the commas?

Rheeseyb (approved):

PR #4821, opened 2024-02-01
Support building via nix

Rheeseyb:
**Problem:**
We use nix for local development, but a series of scripts to build and deploy the actual project, leading to discrepancies between both environments that tend to cause painful issues when they arise. This has become more important now that we want to be able to use separate nix packages for building vscode vs the rest of the editor, so that we can support different node versions.

**Fix:**
This PR introduces the required scripts to be able to build and deploy the application using nix exclusively. It also introduces a new `vscode-build/shell.nix` for building vscode with a different version of node.

**Note:**
As part of this I also upped the node version to v18 for the rest of the editor. I'm happy to drop that back down to v16 and introduce that change in a separate PR.


ruggi (approved):

shell.nix
```diff
@@ -224,25 +225,62 @@ let
...
+      import Control.Monad
+      import Data.Foldable
+      import Data.List
+      import Data.Monoid
+      import System.Exit
+      import System.Process
+
+      expectedToolVersions :: [(String, [String], [String])]
+      expectedToolVersions =
+        [ ("pnpm", ["--version"], ["7.14.2"])
```
seanparsons, line 257: Should we add `nodejs` to this?

shell.nix
```diff
@@ -649,20 +651,17 @@ let
     (pkgs.writeScriptBin "build-server" ''
       #!/usr/bin/env bash
       set -e
+      cabal-update
```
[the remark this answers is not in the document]
seanparsons: Disregard my temporary insanity.

balazsbajorics (approved):

seanparsons (approved):

shell.nix
Rheeseyb: Yeah, that's not a bad shout

PR #4824, opened 2024-02-01
try to make Navigator drag and drop tests more reliable

balazsbajorics:
**Problem**
Two Navigator karma tests recently became flaky, one of them was `attempt to reparent non-reparentable item`

**Fix**
While I did not get to the bottom of exactly -what- caused the flakiness, I noticed that the `dragElementWithDNDEvents` helper function did not fire multiple staggered move events like it's sibling `mouseDragFromPointToPoint`. So I introduced the same staggering mechanism, dispatching 5 `drag` events before the `drop` event and it fixed all navigator tests, except one which became reliably breaking.
Turns out `'attempt to reparent non-reparentable item'` _should have been_ failing for a few months now but it was not reliable enough and instead it produced a false positive. I deleted this test.


bkrmendy (approved):

editor/src/components/canvas/event-helpers.test-utils.tsx
```diff
@@ -1035,70 +1034,84 @@ export async function dragElementWithDNDEvents(
...
-        clientY: endPoint.y,
-        movementX: dragDelta.x,
-        movementY: dragDelta.y,
-        buttons: 1,
-      }),
-    )
-  })
-
-  await wait(0)
+  const delta: Point = {
```
ruggi, line 1037: I believe you can use `Utils.pointDifference` here

ruggi (approved):

PR #4826, opened 2024-02-01
Insert scoped map parameters

seanparsons:
**Problem:**
The variables inside map expressions are currently not available for insert the way that component props currently are in the `Variables` tab.

**Fix:**
The following changes were made to support the use of map expression function parameters:
- The names of the various values are attached to the map expression, extracted as part of the parse.
- Capturing of the values associated with those parameters was added to the canvas handling, with those values added to the scoped variables.
- Handling of the insertion was added, with constraints added to prevent the insertion from happening at a level in the hierarchy above where parameters are defined.

**Notes:**
There's a test marked with `xit` which relates to an issue found while testing this, that for a single frame an error indicating that the inserted parameter could not be found flashes up. Mitigating the issue in the test appeared to be impossible without some unrelated changes and the actual issue doesn't seem to be related to the changes in this PR. That issue will be fixed in a subsequent PR.

**Commit Details:**
- Refactored `ReparentTargetForPaste` into the component types and created some typeguards for the subtypes.
- Extracted out the checks for valid insertion from `pasteIntoParentOrGrandparent` into `canInsertIntoTarget`.
- Added `applyElementCeilingToReparentTarget` which can take a potential reparent target and ensure that the target does not get applied to an element above a ceiling target.
- Added `insertionCeiling` field to `VariableMetadata`.
- Refactored `JSXAttributeOtherJavaScriptKeepDeepEqualityCall` into an instance specifically for the named type and created a new instance which combines those two types.
- Added `valuesInScopeFromParameters` to `JSXMapExpression`, along with the necessary additions to the parser to extract those.
- Implemented `getOtherJavaScriptTypeFromExpression` and `getOtherJavaScriptType` utility functions.


editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -104,6 +104,7 @@ export const ElementsToRerenderGLOBAL: { current: ElementsToRerender } = {
 
 export interface VariableMetadata {
   spiedValue: unknown
+  insertionCeiling: ElementPath | null
```
balazsbajorics, line 107: this is a great idea!

liady: Definitely!

liady (approved):

editor/src/components/editor/variablesmenu.spec.browser2.tsx
```diff
@@ -162,6 +162,149 @@ describe('variables menu', () => {
...
+      document.execCommand('insertText', false, 'value')
+      expect(getInsertItems().length).toEqual(1)
+      expect(getInsertItems()[0].innerText).toEqual('value')
+
+      const filterBox = await screen.findByTestId(InsertMenuFilterTestId)
+      forceNotNull('the filter box must not be null', filterBox)
+
+      await pressKey('Enter', { targetElement: filterBox })
+
+      expect(getPrintedUiJsCode(editor.getEditorState(), '/src/app.js')).toEqual(
```
balazsbajorics, line 188: if we want to assert that nothing changed, I think it would be beter to compare the printedUiJsCode to a reference the starting code, instead of an inline string literal here

editor/src/components/editor/variablesmenu.spec.browser2.tsx
```diff
@@ -162,6 +162,149 @@ describe('variables menu', () => {
...
+      await openVariablesMenu(editor)
+
+      document.execCommand('insertText', false, 'value')
+      expect(getInsertItems().length).toEqual(1)
+      expect(getInsertItems()[0].innerText).toEqual('value')
+
+      const filterBox = await screen.findByTestId(InsertMenuFilterTestId)
+      forceNotNull('the filter box must not be null', filterBox)
+
+      await pressKey('Enter', { targetElement: filterBox })
```
balazsbajorics, line 186: I tried the feature branch and the code also shows a toast 'Cannot find a suitable parent', I think we could also expect that here!

editor/src/components/shared/scoped-variables.ts
```diff
@@ -153,12 +153,14 @@ function getContainerPropsValue(
 function generateVariableTypes(variables: VariableData): Variable[] {
   return Object.entries(variables).flatMap(([name, valueMetadata]) => {
     const value = valueMetadata.spiedValue
+    const insertionCeiling = valueMetadata.insertionCeiling
     const type = getTypeByValue(value)
-    const variable = {
+    const variable: Variable = {
```
balazsbajorics, line 158: 👌

Rheeseyb (approved):

editor/src/components/editor/variablesmenu.spec.browser2.tsx
seanparsons: Fixed in c8cccb26a.

PR #4829, opened 2024-02-02
Now store the session as JSON.

seanparsons:
**Problem:**
Our server serialises the content of the session with a binary serialisation which means that it's difficult for anything else to look into the content of that session even if it has access to the database.

**Fix:**
This adds an additional column `session_json` to the `persistent_session` table, which is populated at these points:
- When the session is created.
- When the session is updated.
- When a lookup is made against the session and if the `session_json` column is NULL.

The `session_json` column is stored as the PostgreSQL native `jsonb` type which means it can be manipulated directly in PostgreSQL itself, rather than being stored as a text type.

**Commit Details:**
- Add a `session_json` column to `persistent_session`.
- `PersistentSessionFields` now includes the type for `session_json`.
- `sessionToTable` now takes a value of type `Value` to populate the `session_json` column with.
- Added `sessionMapToJSON` to validate the `SessionMap` value we expect as well as to construct the JSON that goes into `session_json`.
- `lookupSession` backfills the `session_json` column if it is NULL.
- `insertSession` and `replaceSession` functions of the `Storage SessionStorage` instance now streamline in a value for the `session_json` column.


ruggi (approved):
I was wondering if we should make the new column not nullable for consistency
(`alter table ... add column session_json jsonb not null default '{}'`) but I think it's fine this way.

Rheeseyb (approved):

PR #4832, opened 2024-02-02
sigh

balazsbajorics:
**Problem:**
There's a commenting test `can reparent comment` that is noticeably flaky. 

I was looking into what are the most brittle parts of the actual test and one stood out to me:

We move the mouse to trigger an animation and await "animation duration + 1ms". That's a very slim error margin!

**Fix:**
the REAL fix would be to not use a timer here and actually await the animation to finish.

for now though, I just change this timer to 1 second


ruggi (approved):
😅

Rheeseyb (commented):
Yuck! But 👍

PR #4836, opened 2024-02-05
Fix: Change dns resolution order to ipv4 inBFF server

liady:
**Problem:**
In Node > 17, the default dns resolution is ipv6, which maps `localhost` to `::1`, and breaks the proxy.

**Fix:**
Setting the default dns resolving to ipv4 (the other workaround is to aloways call `127.0.0.1`, but that's a more robust fix)


Rheeseyb (approved):
I'm approving this to unblock, but we need to track this hack and if possible wrap it in a check to see if this is running locally

gbalint (approved):

PR #4846, opened 2024-02-06
Remix projects groundwork

### LINKED ISSUE #4855: BFF projects setup

ruggi:
Fix #4855 

This PR introduces the ground work to use the new Remix app (https://github.com/concrete-utopia/utopia/pull/4807) to build more non-editor pages while being directly connected on the backend with our database.

This PR offers a new `/projects` page in the Remix app that can sunset the legacy NextJS equivalent.

**Commit details**

1. Introduced Prisma as an ORM to directly talk with our backend. The schema is directly imported from the existing PG schema with `prisma db pull`.
2. Included Vanilla Extract+Sprinkles for styling in the Remix app, which will allow us to iterate quickly _and_ extract/import/extend the existing design system bits from the editor. Additionally, added a couple of boilerplate/example usages of the various libraries to make it easier to get started (`sprinkle`, `style`, `recipe`, etc.).
3. Added some utility functions (thanks to https://github.com/concrete-utopia/utopia/pull/4829) to retrieve a user's details from incoming cookies, which can be used to authenticate both REST API calls as well as page loads.
4. Added integration tests for the backend bits, and hooked them up with the CI.
5. Due to a deliberate design choice by Prisma (which I strongly don't agree with), passing `undefined` values for `where` keys in db queries effectively ignores that key. The TS compiler can help us, but only to a certain extent. To mitigate this, the Prisma client instantiated in the singleton (https://github.com/concrete-utopia/utopia/pull/4846/files#diff-bce98eb811edaad1caedc672aefff54fb52ea92aade6213e73dd68cb3a8ca987R29) is extended so that any query that does a selection will throw an error if the passed `where` clause contains undefined values for defined keys.
6. Added a simple error boundary for the non-v1-routes which will show the error message. We can happily make this nicer/more informative later but at least there's the entry point.
7. Added the `/projects` page in the Remix app so that it returns a list of the current user's projects, the user details (avatar + email), and some basic pagination logic. This is left intentionally unstyled.
8. Added a placeholder `showcase` route, that returns nothing (at the moment, will revisit later).
9. Added a placeholder project thumbnail route that returns a dummy Utopia pyramid from the CDN, until we have the screenshot service up and running again.
10. Added a bunch of test utils to make sure writing tests is decently ergonomic.

**Notes**
- Due to coexistence problems between Prisma and PNPM, the `prisma-client` is actually generated and linked via `node_modules` with `pnpm exec prisma generate`
- On local machines a simple Docker container is used for the supporting PG database – this is due to limitations of the NixOS testing library on macOS. Likewise, to keep consistency as much as possible, the CI workflow for PRs also uses a simple `service` for PG and invokes the PNPM test directly instead of via Nix, but because of how everything is structured there should be no discrepancies between environments, as far as I can tell.
-


bkrmendy (approved):

utopia-remix/app/singleton.server.ts
```diff
@@ -0,0 +1,13 @@
+// Since the dev server re-requires the bundle, do some shenanigans to make
+// certain things persist across that 😆
+// Borrowed/modified from https://github.com/jenseng/abuse-the-platform/blob/2993a7e846c95ace693ce61626fa072174c8d9c7/app/utils/singleton.ts
+
+export const singleton = <Value>(
```
liady, line 5: nice

.github/workflows/pull-requests.yml
```diff
@@ -155,7 +155,7 @@ jobs:
         id: fc-initial
         with:
           issue-number: ${{ github.event.pull_request.number }}
-          comment-author: 'github-actions[bot]'
+          comment-author: "github-actions[bot]"
```
Rheeseyb, line 158: It looks like for some reason this file was formatted with different prettier settings, or maybe without prettier? This change will be reverted by the next person to save a change in this file.

shell.nix
```diff
@@ -516,12 +528,13 @@ let
...
     '')
-	(pkgs.writeScriptBin "watch-remix" ''
+    (pkgs.writeScriptBin "watch-remix" ''
       #!/usr/bin/env bash
       set -e
       cd $(${pkgs.git}/bin/git rev-parse --show-toplevel)/utopia-remix
       ${pnpm}/bin/pnpm install
-	  PORT=8002 pnpm run dev
+      ${pnpm}/bin/pnpm exec prisma generate
+      PORT=8002 pnpm run dev
```
Rheeseyb, line 537:
```suggestion
      PORT=8002 ${pnpm}/bin/pnpm run dev
```
Double check this works before accepting the suggestion, but there's no reason why it shouldn't. The reason for this is that it ensures this step is using the exact same version of `pnpm` as all others, as opposed to another version that may happen to be on the PATH

liady (approved):

utopia-remix/app/db.server.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+/**
+ * The singleton instance of the Prisma client, extended to suit our needs.
+ * Specifically:
+ * - apply the rejectUndefinedValues to all selecting queries, so that if by any chance we
+ * 	pass a select clause where the value is undefined an error is thrown and the statement/transaction
+ *  is aborted.
+ */
```
Rheeseyb, line 28: Amazing

Rheeseyb: Genuinely. Thanks for catching and explicitly handling this, because that would have been (and must have been!) a nightmare to discover and debug

.github/workflows/pull-requests.yml
ruggi: mmh we have a mixture of single and double quotes in that file tho

utopia-remix/app/db.server.ts
ruggi: yeah right? It startled me when I saw it, and then I found  of GH discussions questioning this, but I guess that design decision is there to stay ¯\_(ツ)_/¯

shell.nix
ruggi: yesss I forgot to update this one 🤦

ruggi: https://github.com/concrete-utopia/utopia/pull/4846/commits/48b6821ce59a623c0a4dc7d6dcc9d2035f531cd7

utopia-remix/app/singleton.server.ts
```diff
@@ -0,0 +1,13 @@
...
+  name: string,
+  valueFactory: () => Value,
+): Value => {
+  const g = global as unknown as { __singletons: Record<string, unknown> };
+  g.__singletons ??= {};
```
seanparsons, line 10: Took me a while to find any mention of this operator and what it does.

ruggi:
note about the formatting: we found out the problem is the `eslint-config-prettier` dep in the default Remix package.json, I'll make an incremental PR to fix all of these annoyances so not to pollute this already large diff

utopia-remix/app/singleton.server.ts
ruggi: this comes straight from the Remix code, fwiw

utopia-remix/app/handlers/listProjects.ts
```diff
@@ -0,0 +1,35 @@
+import { listProjects } from "../models/project.server";
```
seanparsons, line 1: Definitely not using our regular Prettier formatting as we bin off the semicolons.

Rheeseyb (approved):

utopia-remix/package-lock.json
seanparsons, line 1: Do we even need this if we're using pnpm?

seanparsons (approved):

utopia-remix/package-lock.json
ruggi: https://github.com/concrete-utopia/utopia/pull/4846/commits/645a5c2d14c765e22e2739e5e2b709043c35319f

PR #4847, opened 2024-02-06
Fix Flaky Comment Tests

seanparsons:
**Problem:**
A couple of the comments tests fail somewhat randomly, especially when run on the CI servers.

**Cause:**
The root of this seems to be (from examining the state with logging) of changes being fed into the thread metadata then reverting shortly afterwards. Since the tests run at an inhumanely fast pace, those changes fire rapidly and potentially end up in a state of conflict behind the scenes in however Liveblocks store that data.

**Fix:**
The two tests that appear to suffer from this issue have now had a series of short pauses added after user events are fired, to give the change a chance to resolve before moving onto another.

**Commit Details:**
- Tests have pauses added after each action to hopefully give things a chance to settle.

### DIFF
--- a/puppeteer-tests/src/comments/commenting.spec.tsx
+++ b/puppeteer-tests/src/comments/commenting.spec.tsx
@@ -221,7 +221,7 @@ describe('Comments test', () => {
     expect(commentsTabs).toHaveLength(0)
   })
 
-  it.only('can reparent comment', async () => {
+  it('can reparent comment', async () => {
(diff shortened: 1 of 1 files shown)


liady (approved):
We can of course have the number as a variable so we could tweak it, but I'm not sure how valuable it will be.

seanparsons:
> We can of course have the number as a variable so we could tweak it, but I'm not sure how valuable it will be.

That's a good idea actually, I'll add that now.

gbalint (approved):

PR #4851, opened 2024-02-06
Only include port on localhost

Rheeseyb:
**Problem:**
With the introduction of the Remix BFF https://github.com/concrete-utopia/utopia/pull/4807 we began including port numbers in requests when we shouldn't have. This is (I believe) causing CORS errors.

**Fix:**
Only include the port numbers when running locally

### DIFF
--- a/website-next/components/common/env-vars.ts
+++ b/website-next/components/common/env-vars.ts
@@ -5,7 +5,7 @@ export const SCHEME: string = typeof window === 'undefined' ? 'http' : window.lo
 export const BARE_HOSTNAME = HOSTNAME.startsWith('www.') ? HOSTNAME.slice(4) : HOSTNAME
 const isLocalHost = BARE_HOSTNAME.startsWith('localhost')
 export const PORT: string =
-  typeof window === 'undefined' || window.location.port.length === 0 || !isLocalHost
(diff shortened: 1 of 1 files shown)


website-next/components/common/env-vars.ts
```diff
@@ -44,9 +45,7 @@ export function isBackendBFF(): boolean {
 }
 
 export const UTOPIA_BACKEND =
-  (PRODUCTION_OR_STAGING_CONFIG
```
Rheeseyb, line 47: The issue is that this is never `true` in the context of the website-next project. The correct fix is to pass through the environment variable

website-next/components/common/env-vars.ts
```diff
@@ -44,9 +45,7 @@ export function isBackendBFF(): boolean {
...
-    ? BASE_URL
-    : `${SCHEME}//${HOSTNAME}:${LOCAL_BACKEND_PORTS[BACKEND_TYPE]}/`) + 'v1/'
+  (isLocalHost ? `${SCHEME}//${HOSTNAME}:${LOCAL_BACKEND_PORTS[BACKEND_TYPE]}/` : BASE_URL) + 'v1/'
```
liady, line 48: Can we use DEVELOPMENT_ENV here?

PR #4853, opened 2024-02-07
Avoid re-creating remix routes

bkrmendy:
## Problem
Some Remix tests intermittently fail when the tests try to navigate forward via a canvas toolbar or the Remix scene toolbar.

### The test failures
image

## Root cause
If the routes are re-generated for any reason, a new router is created. When we create a new router (with the `createMemoryRouter` call), we supply it with the navigation stack via the `initialEntries` option, so the navigation stack is preserved. However, it turns out that when a new router is created, the currently active route defaults to be the last element in `initialEntries`, meaning that forward navigation won't work, because there's no route to move to.

For more info on `createMemoryRouter`, see https://reactrouter.com/en/main/router-components/memory-router.

We haven't debugged why the routes are (sometimes) recreated in the tests.

## Fix
To avoid needlessly recreating the routes, this PR adds a more thorough, but heuristic check based on the `id` prop of the routes. The check is implemented in [`useRoutesWithIdEquality`](https://github.com/concrete-utopia/utopia/blob/70e3029d0e430eb2cf208899c6635797da6296ae/editor/src/components/canvas/remix/utopia-remix-root-component.tsx#L217)


ruggi (approved):

gbalint (approved):
🎉

PR #4859, opened 2024-02-08
Get Prisma working on Linux/NixOS.

seanparsons:
**Problem:**
As Prisma is a Rust binary, it suffers from the usual prebuilt binary issues that can occur with linking.

**Fix:**
Appears the somewhat official fix for this is to provide the environment variables that point to the various binaries: https://github.com/prisma/prisma/issues/3026#issuecomment-927117819

**Commit Details:**
- Added interpolated environment variables to pull in the Prisma binaries.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -761,4 +761,11 @@ in pkgs.mkShell {
   NODE_OPENSSL_OPTION = "--openssl-legacy-provider";
   # Required for node-gyp, apparently
   npm_config_force_process_config = true;
+
+  # Make Prisma work.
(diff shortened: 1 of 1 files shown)


bkrmendy (approved):

ruggi (approved):

Rheeseyb (approved):
I guess we'll need to include these packages in the container when deploying this

seanparsons:
> I guess we'll need to include these packages in the container when deploying this

The answer to that is maybe, it's not a NixOS based image we base things on so it might "just work", it depends on what that image was based on. So it's mostly just a case of trying it and fixing it if it doesn't work.

PR #4862, opened 2024-02-08
Remix: remove pagination, return all projects

### LINKED ISSUE #4861: Don't paginate remix projects

ruggi:
Fix #4861

**Problem:**

The Remix projects page should not bother with paginating the results, as the client should be doing it (and even for large projects collections it should not be a problem for a long while).

**Fix:**

Don't limit/offset the query selection (and make sure the content is filtered out!).


bkrmendy (approved):

liady (approved):
LGTM, but great pagination code to remove, though :-)

Rheeseyb (approved):

PR #4867, opened 2024-02-09
Remix: rename projects

ruggi:
Fix #4866 

**Problem:**

The Remix projects page will need a way to rename projects.

**Fix:**

- Add an action route to rename projects
- Add the related backend `update` query
- Made sure that Prisma errors coming from handlers are wrapped inside `ApiErrors` and displayed nicely
- Add tests for everything above

**Note**
- ⚠️ The diff is made larger by an adjustment to the first argument of the `handle` function, which now gets an interface instead of a plain `Request`, therefore https://github.com/concrete-utopia/utopia/commit/8af265db1718ea752a31394195dbd4ac2ccd6e31 could be pretty much ignored for reviewing purposes.

PR #4870, opened 2024-02-09
Check for unpinned dependencies

ruggi:
**Problem:**

Somebody (🙈 🙋‍♂️) always seems to be forgetting to pin dependencies when opening PRs (latest proof: https://github.com/concrete-utopia/utopia/pull/4867#discussion_r1484429968).

Therefore, let's have the robots fix ~~my~~ our mistakes, for once.

**Fix:**

1. Add a pre-commit check via husky that will exit with a non-0 code if there are unpinned deps in any of the folders listed as targets in `check-unpinned-deps.js`
2. Add a CI step that runs on PRs that does the same thing **if for some reason the pre-commit hook did not do its job**.

**Note:** I commented out the checks on `devDependencies`, because I'm not sure whether we should enforce it there or not.


bkrmendy (approved):

check-unpinned-deps.js
```diff
@@ -0,0 +1,25 @@
+const fs = require('fs')
+const { join } = require('path')
+
+/**
+ * Scan dependencies in the package.json files in these folders.
+ */
+const folders = ['.', './editor', './utopia-remix', './puppeteer-tests']
```
Rheeseyb, line 7: Can you add `utopia-api`, `utopia-vscode-common`, and `utopia-vscode-extension` to this list please

Rheeseyb (approved):
Thank you :)

check-unpinned-deps.js
```diff
@@ -0,0 +1,25 @@
...
+
+const unpinnedDeps = folders.flatMap((projectPath) => {
+  const path = join(projectPath, 'package.json')
+  const packageJson = JSON.parse(fs.readFileSync(path), 'utf8')
+  const allDeps = {
+    ...packageJson.dependencies,
+    // ...packageJson.devDependencies,
+  }
+  return Object.entries(allDeps)
+    .filter(([, version]) => version.match(/^[\^~]/))
```
seanparsons, line 17: Not sure I understand what the regex is checking, can we not just do `version.startsWith('^')`?

ruggi: yeah it's pointlessly unreadable, updated in https://github.com/concrete-utopia/utopia/pull/4870/commits/190d1ae2c4170ff0fa1b1f24592c782c7af82a82

liady (approved):

Rheeseyb (approved):

PR #4871, opened 2024-02-12
Unify our edit permissions mechanism

liady:
A small PR to unify our permissions mechanisms.
Instead of using two separate mechanisms (`hasEditPermissions` and `allowedToEditProject`) - now the permissions check is in `permissions.ts` (via the `usePermissions` hook) and the semantic checks  (`canEdit`, `canComment`, etc) can be in the specific usage areas.


editor/src/components/navigator/left-pane/index.tsx
```diff
@@ -93,7 +93,7 @@ export const LeftPaneComponent = React.memo<LeftPaneComponentProps>((props) => {
           }}
         >
           {when(
-            isMyProject,
+            isProjectOwner,
```
liady, line 96: this later needs to be changed to a semantic check (`canEdit`/`canViewPanes` or something along that lines), but I wanted to keep this PR focused

PR #4872, opened 2024-02-12
Scene dragging: do not save new scene comment canvas position in every frame

gbalint:
**Problem:**
`CommentMaintainer` saves the new global position of scene comments when there is a position change, even during interactions. Updating liveblocks like this is waaay too frequent, and causes bugs and test failures.

**Fix:**
Do not do anything during canvas interactions. The new position is stored right after the interaction is over.

### DIFF
--- a/editor/src/core/commenting/comment-maintainer.tsx
+++ b/editor/src/core/commenting/comment-maintainer.tsx
@@ -6,6 +6,7 @@ import { getIdOfScene } from '../../components/canvas/controls/comment-mode/comm
 import * as EP from '../shared/element-path'
 import { isNotNullFiniteRectangle } from '../shared/math-utils'
 import { isCanvasThreadMetadata, liveblocksThreadMetadataToUtopia } from './comment-types'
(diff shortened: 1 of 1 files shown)


bkrmendy (approved):

editor/src/core/commenting/comment-maintainer.tsx
```diff
@@ -34,6 +35,16 @@ function useMaintainComments() {
...
+    Substores.canvas,
+    (store) => store.editor.canvas.interactionSession != null,
+    'useMaintainComments isInteraction',
+  )
+
+  if (isInteraction) {
+    return
+  }
+
   threads.forEach(async (t): Promise<void> => {
```
balazsbajorics, line 48: mhmmmmmm, are we directly calling editThreadMetadata from render? shouldn't this part be wrapped in a useEffect?

gbalint: This triggers an async liveblocks update but does not wait for it, so it did not seem dangerous to me.

seanparsons (approved):

PR #4882, opened 2024-02-12
Scoped variables show destructured parameters.

seanparsons:
**Problem:**
With code like the following in a component:
```typescript
export function App({ style: { background } }) {
```
The destructured values aren't made available into the scoped variables menu.

**Fix:**
The main fixes needed for this to work were the following:
- `createComponentRendererComponent` now uses `propertiesExposedByParam` to identify the properties that are brought into scope from the destructuring, which are added to the dictionary of scoped variables.
- As a result of the above work, it was observed that `applyPropsParamToPassedProps` wasn't properly handled destructured fields that are renamed.

In the process of doing the above work I noticed that `getVariablesInScope` was looking for the first component it could find, not the targeted component. To perform the lookup `findContainingComponent` was implemented.

**Commit Details:**
- Added `propertiesExposedByParam` utility function.
- Added `findContainingComponent` utility function.
- `getVariablesInScope` now finds the correct component and not just the first one.
- Fixed `applyPropsParamToPassedProps` to correctly handle destructured values and repeated field usages.
- `createComponentRendererComponent` includes values exposed by the props param when building `spiedVariablesInScope`.

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
@@ -37,6 +37,7 @@ import { usePubSubAtomReadOnly } from '../../../core/shared/atom-with-pub-sub'
 import { JSX_CANVAS_LOOKUP_FUNCTION_NAME } from '../../../core/shared/dom-utils'
 import { objectMap } from '../../../core/shared/object-utils'
 import type { ComponentRendererComponent } from './component-renderer-component'
+import { mapArrayToDictionary } from '../../../core/shared/array-utils'
 
 function tryToGetInstancePath(
   maybePath: ElementPath | null,
@@ -133,12 +134,18 @@ export function createComponentRendererComponent(params: {
 
     let spiedVariablesInScope: VariableData = {}
     if (utopiaJsxComponent.param != null) {
-      for (const paramName of propertiesExposedByParam(utopiaJsxComponent.param)) {
-        spiedVariablesInScope[paramName] = {
-          spiedValue: scope[paramName],
-          insertionCeiling: instancePath,
-        }
-      }
+      spiedVariablesInScope = mapArrayToDictionary(
+        propertiesExposedByParam(utopiaJsxComponent.param),
+        (paramName) => {
+          return paramName
+        },
+        (paramName) => {
+          return {
+            spiedValue: scope[paramName],
+            insertionCeiling: instancePath,
+          }
+        },
+      )
     }
 
     let codeError: Error | null = null
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
@@ -5,6 +5,7 @@ import {
   isUtopiaJSXComponent,
   isSVGElement,
   isJSXElement,
+  propertiesExposedByParam,
 } from '../../../core/shared/element-template'
 import { optionalMap } from '../../../core/shared/optional-utils'
 import type {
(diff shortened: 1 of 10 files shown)


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -194,12 +203,23 @@ export function createComponentRendererComponent(params: {
...
         utopiaJsxComponent.arbitraryJSBlock,
         scope,
       )
+
+      spiedVariablesInScope = {
+        ...spiedVariablesInScope,
+        ...objectMap(
+          (spiedValue) => ({
+            spiedValue: spiedValue,
+            insertionCeiling: null,
```
balazsbajorics, line 218: I know this is just moved code, but in line 139 above insertionCeiling is set to `instancePath`, shouldn't this also be `instancePath`? or in this case null and instancePath mean the same thing?

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-props-utils.ts
```diff
@@ -38,39 +38,36 @@ export function applyPropsParamToPassedProps(
...
                   innerBoundParam.defaultExpression,
                 )
-                remainingKeys = remainingKeys.filter((k) => k !== paramName)
                 delete remainingValues[paramName]
               }
             }
+          } else {
+            // e.g. `{ prop: renamedProp }` or `{ prop: { /* further destructuring */ } }`
+            // Can't spread if we have a property name
+            const innerValue = valueAsRecord[propertyName]
```
balazsbajorics, line 64: can you explain what was the actual fix here? I see you introduced valueAsRecord which seems to be the same as remainingValues except we don't delete entries from it in the loop. but I don't see how did that fix the behavior

editor/src/components/shared/scoped-variables.ts
```diff
@@ -87,12 +90,11 @@ function getTopLevelVariables(topLevelElements: TopLevelElement[], underlyingFil
 }
 
 function getVariablesFromComponent(
-  topLevelElements: TopLevelElement[],
+  jsxComponent: UtopiaJSXComponent | null,
   elementPath: ElementPath,
   variablesInScopeFromEditorState: VariablesInScope,
 ): AllVariablesInScope {
   const elementPathString = toComponentId(elementPath)
-  const jsxComponent = topLevelElements.find(isUtopiaJSXComponent)
```
balazsbajorics, line 95: wow nice find!

balazsbajorics (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-props-utils.ts
seanparsons: Previously it was using `remainingValues` and deleting from that object, but it's totally legitimate to refer to a field more than once. Like you could destructure `style` but also refer to `style` itself in the same case. As `remainingValues` was getting trimmed this meant that every use of a field after the first was getting an `undefined`.

balazsbajorics: ahaa I see

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
seanparsons: In this case it doesn't make any particular difference from memory, but yes I don't want to change it in this PR, even though I think setting it to that would be beneficial even if just as a belt and braces change.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -130,6 +131,16 @@ export function createComponentRendererComponent(params: {
...
 
+    let spiedVariablesInScope: VariableData = {}
+    if (utopiaJsxComponent.param != null) {
+      for (const paramName of propertiesExposedByParam(utopiaJsxComponent.param)) {
+        spiedVariablesInScope[paramName] = {
+          spiedValue: scope[paramName],
+          insertionCeiling: instancePath,
+        }
+      }
+    }
```
Rheeseyb, line 142: This is a bit picky but can this be restructured a bit so that we're not using mutation here to build the `spiedVariablesInScope` above.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-props-utils.ts
Rheeseyb: Oh wow, nice catch!

Rheeseyb (approved):

PR #4886, opened 2024-02-13
Nested picker UI for nested objects

liady:
Adjust the picker UI so it will display nested objects in a nested way (and will also display the objects themselves)
image

### DIFF
--- a/editor/src/components/inspector/sections/component-section/component-section.tsx
+++ b/editor/src/components/inspector/sections/component-section/component-section.tsx
@@ -432,51 +432,71 @@ const DataPickerPopup = React.memo(
           <div style={{ fontSize: 14, fontWeight: 400, marginBottom: 16 }}>
             <span>Data</span>
           </div>
-          {variableNamesInScope.map(({ variableName, definedElsewhere, value }, idx) => {
-            return (
-              <Button
-                data-testid={VariableFromScopeOptionTestId(idx)}
-                key={variableName}
-                onClick={onTweakProperty(variableName, definedElsewhere)}
-                style={{ width: '100%' }}
-              >
-                <UIGridRow
-                  padded={false}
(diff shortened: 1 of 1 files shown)


editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -1164,54 +1184,81 @@ export class ComponentSection extends React.Component<
...
-
-    return [{ variableName: name, definedElsewhere: name, value: '[array]' }, ...values]
   }
 
-  const values = Object.entries(value).flatMap(([key, field]) =>
-    valuesFromVariable(`${name}['${key}']`, field).map((variable) =>
-      patchDefinedElsewhereInfo(variable),
+  return [
+    patchDefinedElsewhereInfo({
+      displayName,
```
balazsbajorics, line 1232: nitpick, but we don't like to mix the verbose object literal and the shorthand version. we actually prefer the verbose `displayName: displayName` over the shorthand

liady: good point, fixed!

liady:
@balazsbajorics made the UI more responsive for long / nested objects and values. thanks!

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -1164,54 +1187,89 @@ export class ComponentSection extends React.Component<
...
   value: string
+  depth?: number
 }
 
 function valuesFromObject(
   name: string,
   objectName: string,
   value: object | null,
+  depth: number = 0,
+  displayName: string = name,
```
bkrmendy, line 1201: We should avoid using default parameters because they muddle up the function's API for a limited degree of convenience

PR #4889, opened 2024-02-13
Data picker - order variables according to relevance

bkrmendy:
# Try it [here](https://utopia.fish/p/40ef424c-rowan-constellation/?branch_name=feature-variable-value-ordering)

## Problem
The options in the data picker are shown in the order they were inserted into `variablesInScope`. This order might not make sense, especially if there's a style prop involved:
image

## Fix
Use a heuristic ordering that gives priority to some options. The heuristics are the following:
- Any variables that have a control description (from `registerModule` and friends) that match the type of that variable are moved to the top.
- Any variables that match the type of the current property value are moved to the middle
- Every other variable is moved to the bottom

### Out of scope
This PR only implements the heuristic for strings, numbers, objects, arrays, and recursive combinations of these (ie, an array of objects). We can tweak/extend the heuristic when we have real-world data about how useful/correct it is.


balazsbajorics (approved):

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -372,6 +387,105 @@ function getSectionHeightFromPropControl(
...
+function usePropertyControlDescriptions(): Array<ControlDescription> {
+  return useGetPropertyControlsForSelectedComponents().flatMap((controls) =>
+    Object.values(controls.controls),
+  )
+}
+
+function arrayShapesMatch(left: Array<unknown>, right: Array<unknown>): boolean {
+  if (left.length === 0 || right.length === 0) {
+    return true
+  }
```
ruggi, line 399: this feels semantically wrong - we're asking if the array shapes match, but we are returning "yes!" if there's a difference (potential?) between them 🧐

ruggi: it may make sense functionally but then I'd either take it out of the function, or give it an explicit name

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -372,6 +387,105 @@ function getSectionHeightFromPropControl(
...
+}
+
+function variableMatchesArrayControlDescription(
+  variable: Array<unknown>,
+  controlDescription: ArrayControlDescription,
+): boolean {
+  if (variable.length === 0) {
+    return true
+  }
+
```
ruggi, line 437: similar here

ruggi (approved):

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -372,6 +387,105 @@ function getSectionHeightFromPropControl(
...
+    Object.values(controls.controls),
+  )
+}
+
+function arrayShapesMatch(left: Array<unknown>, right: Array<unknown>): boolean {
+  if (left.length === 0 || right.length === 0) {
+    return true
+  }
+
+  return variableShapesMatch(left[0], right[0])
```
gbalint, line 401: shouldn't we check all the elements?

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -372,6 +387,105 @@ function getSectionHeightFromPropControl(
...
+
+function variableMatchesArrayControlDescription(
+  variable: Array<unknown>,
+  controlDescription: ArrayControlDescription,
+): boolean {
+  if (variable.length === 0) {
+    return true
+  }
+
+  return variableMatchesControlDescription(variable[0], controlDescription.propertyControl)
```
gbalint, line 438: Same here, why only the first one is checked?

PR #4890, opened 2024-02-13
Image preview in string prop control

gbalint:
**Description:**
Simple image preview for string property controls: it shows the image when the string property is a url.

image

### DIFF
--- a/editor/src/components/inspector/sections/component-section/property-control-controls.tsx
+++ b/editor/src/components/inspector/sections/component-section/property-control-controls.tsx
@@ -55,8 +55,7 @@ import {
   normalisePathToUnderlyingTarget,
 } from '../../../custom-code/code-file'
 import { useDispatch } from '../../../editor/store/dispatch-context'
-import { isImageUrl } from '../../../../core/shared/file-utils'
-import { when } from '../../../../utils/react-conditionals'
+import { isImage } from '../../../../core/shared/utils'
 
 export interface ControlForPropProps<T extends BaseControlDescription> {
   propPath: PropertyPath
@@ -462,7 +461,7 @@ export const StringInputPropertyControl = React.memo(
           controlStyles={propMetadata.controlStyles}
           focus={props.focusOnMount}
         />
-        {when(isImageUrl(safeValue), <ImagePreview url={safeValue} />)}
+        <ImagePreview url={safeValue} />
       </FlexColumn>
     )
   },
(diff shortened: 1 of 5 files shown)


editor/src/core/shared/file-utils.ts
```diff
@@ -259,3 +259,8 @@ export function isCssFile(filename: string): boolean {
 export function isJsOrTsFile(filename: string): boolean {
   return isJsFile(filename) || isTsFile(filename)
 }
+
+export function isImageUrl(url: string): boolean {
```
liady, line 263: there's also `isImage` function, we might be able to unify them
https://github.com/concrete-utopia/utopia/blob/3f73ea881b22126e3fe104d04ef19219c5588ff9/editor/src/core/shared/utils.ts#L212-L214

editor/src/components/inspector/sections/component-section/property-control-controls.tsx
```diff
@@ -437,28 +439,68 @@ const NumberWithSliderControl = React.memo(
...
+      </FlexColumn>
     )
   },
 )
 
+interface ImagePreviewProps {
+  url: string
+}
+const ImagePreview = React.memo((props: ImagePreviewProps) => {
+  const [imageCanBeLoaded, setImageCanBeLoaded] = React.useState(true)
```
liady, line 475: won't that always start with `true` (even if the url hadn't changed) and so it will always first try to the render the `` on every re-render (even if we know the url is bad)?

balazsbajorics: it would start with true on every re-mount, but not on every re-render

editor/src/core/shared/file-utils.ts
balazsbajorics: nice catch!

editor/src/components/inspector/sections/component-section/property-control-controls.tsx
gbalint: What Balazs wrote and when the url is bad we don't even mount the ImagePreview (see line 465)
But I can integrate that check into the ImagePreview component itself (by initializing imageCanBeLoaded with `isImageUrl(url)`)

bkrmendy (approved):

balazsbajorics (approved):

liady (approved):

editor/src/components/inspector/sections/component-section/property-control-controls.tsx
liady: yes, of course, re-mount was what I meant 🤦‍♂️ in any case it's not critical at all

PR #4911, opened 2024-02-15
Data picker - dim options that don't match

bkrmendy:
## Description

This PR improves that data picker so that options that don't match any control descriptions / the existing property value are dimmed out. 

image

### DIFF
--- a/editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
+++ b/editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
@@ -17,6 +17,7 @@ function valuesFromObject(
   value: object | null,
   depth: number,
   displayName: string,
+  advanced: boolean,
 ): Array<VariableOption> {
   if (value == null) {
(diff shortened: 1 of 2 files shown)


editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
```diff
@@ -17,6 +17,7 @@ function valuesFromObject(
   value: object | null,
   depth: number,
   displayName: string,
+  advanced: boolean,
```
balazsbajorics, line 20: I think advanced is a little hard to understand, makes me feel like this is a very advanced property, only for pro use :) 

what about `matchingType` or `typeMatches` or `matchesProp` or `highlighted` or `valueMatchesPropType`?

bkrmendy: fair point, I'll go with `valueMatchesPropType`

PR #4923, opened 2024-02-19
Remix projects: collaborators

### LINKED ISSUE #4922: Show collaborators in Remix projects

ruggi:
Fix #4922 

**Problem:**

The Remix projects page should show the collaborator avatars in the bottom right corner of the cards.

**Fix:**

Screenshot 2024-02-19 at 1 07 49 PM

This PR adds the collaborator avatars to the projects page. Because of how we currently store that data, the PR uses a lazy approach for loading collaborators for existing projects.

The PR adds a new table, `project_collaborators`, which is basically a join table between the projects and the user details. When the Remix projects page loads, the list of collaborators per project is retrieved directly from the DB, and displayed consequently.

What about storing those records though? → The collaborator data currently lives in the Liveblocks room storage (https://liveblocks.io/document/storage) and is not persisted in our DB. A solution could be to get that data from LB via its REST API when loading the projects, but that would be _extremely_ time-consuming and prone to rate limiting, also considering we currently don't use any caches. We _could_ use PG as a cache, but I don't see the need for it given the solution I implemented here: the PR adds a DB call (only for the BFF, so nothing against the Haskell server directly) in the `useAddMyselfToCollaborators` editor hook so that when a user connects to a room the backend will retrieve the collaborators (fully) from LB and upsert their records in `project_collaborators`.

This has the benefit that only a minimum amount of calls to LB are performed, and they can be easily managed on our side in the future – we could even drop that storage portion entirely and solely rely on our DB moving forward. We _could_ preload the collaborators per project, but again, it does not make much sense.

Careful readers will notice that the `updateCollaborators` call does not check the ownership of the project for the user making the call, but that's on purpose: because we currently don't have permissions management at the DB level, and the user joining a project doesn't need to own it anyways, the call to update the collaborators only requires a authenticated call, regardless of access policies; the LB call only updates (basically the equivalent of a `touch` command) the project in a vacuum, so the user performing the call does not have a say in the operation's logic (= no input).

**Notes**
- As usual, no proper design work with the avatars, they're just displayed in a flex row.


server/migrations/007.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE project_collaborators(
+    id serial,
+    project_id character varying NOT NULL REFERENCES project_i_d(proj_id) ON DELETE CASCADE,
+    user_id character varying NOT NULL REFERENCES user_details(user_id) ON DELETE CASCADE,
+    created_at timestamp with time zone NOT NULL DEFAULT NOW()
+);
+
+CREATE INDEX "idx_project_collaborators_project_id" ON "public"."project_collaborators"(project_id);
```
ruggi, line 8: just using the shorthand syntax here, this uses `BTREE` by default on Postgres

utopia-remix/jest.config.js
```diff
@@ -3,4 +3,5 @@ export default {
   transform: {
     '\\.[jt]sx?$': 'babel-jest',
   },
+  transformIgnorePatterns: ['node_modules/(?!@ngrx|(?!deck.gl)|ng-dynamic)'],
```
seanparsons, line 6: Huh?

seanparsons (approved):

utopia-remix/jest.config.js
ruggi: I had to do this because it was giving jest was having a stroke due to some imports in `url-join` 😬

seanparsons: I'd throw a comment in there about that then, otherwise we'll never know in 10 months time when coming back to this.

ruggi: [`c574a18` (#4923)](https://github.com/concrete-utopia/utopia/pull/4923/commits/c574a181ba3204ea8bf11b5c4d890bb0f5719922)

Rheeseyb (approved):

PR #4925, opened 2024-02-20
relevant object / array elements

bkrmendy:
## Description

This PR improves the ordering of options in the data picker in two ways:
- Object properties / array elements that match the target prop are moved up in the list
- Objects / arrays that have any props/elements that match the target prop are moved up in the list

The main part of the PR is the `orderVariablesForRelevance` function

### Details
The PR also adds an explicit types for the variables that populate the data picker, so it's easier to see what's going on in `orderVariablesForRelevance`

image
image

### DIFF
--- a/editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
+++ b/editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
@@ -9,20 +9,20 @@ import { useRefEditorState } from '../../../editor/store/store-hook'
 import { UIGridRow } from '../../widgets/ui-grid-row'
 import { DataPickerPopupTestId, VariableFromScopeOptionTestId } from './component-section'
 import * as EP from '../../../../core/shared/element-path'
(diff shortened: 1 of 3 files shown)


balazsbajorics (approved):

liady (commented):
nice refactor as well 👏

liady (approved):

editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
```diff
@@ -121,37 +150,37 @@ interface ValueRowProps {
 function ValueRow({ variableOption, idx, onTweakProperty }: ValueRowProps) {
   const colorTheme = useColorTheme()
   const [selectedIndex, setSelectedIndex] = React.useState<number>(0)
-  const childrenLength = variableOption.variableChildren?.length ?? 0
+
+  const childrenLength = variableOption.type === 'array' ? variableOption.children.length : 0
   const [childrenOpen, setChildrenOpen] = React.useState<boolean>(
```
liady, line 155: should the default value here take into account if there are `matches` inside the object?

PR #4928, opened 2024-02-20
Data picker - open object if it has matching prop

bkrmendy:
## Description

This PR tweaks the data picker so that objects are opened if the object has a matching prop (or any descendant of an object matches).

image


balazsbajorics (approved):

liady (approved):
nice! we might need to rethink the depth/childrenLength logic (it's pretty arbitrary now), but that will be trial and error i guess

PR #4929, opened 2024-02-21
Project visibility

liady:
## Projects visibility (private/public)
This PR adds the capability to define a project as public (open to all), or private (open only to owner).
This works with [Okta's FGA](https://docs.fga.dev/modeling/basics/modeling-basics), and is the foundation for adding inremental support for fine grained permissions.
Refer to our paper doc ("Multiplayer II: Fine Project! Grained Access?") for the detailed spec.

## Implementation Contents
* A new table was added to the database to save project's visibility status (currently `PRIVATE`, `PUBLIC`, `WITH_LINK`), and a matching definition was added in Prisma's schema.
* Authorization model was written in Okta's FGA, to reflect the possible permissions.
* Infrastructure for contacting Okta's FGA (`fgaService`), and a `permissionsService` that abstracts it.
* When changing a project's visibility, we store it in our DB (for our usage), and also send a tuple to Okta that defines that `user:*` (which means "all users") have/doesn't have a `viewer` relation to the project, and the `can_view` permission is derived from it.
* Every route now specifies a `validator` to access this route, which on project routes rely on project permissions.
* When calling our `/project` endpoint, a check is being made to see if there is a `can_view` permission for this project, and if not - a 404 page is being generated.
* UI added - project visibility UI in the Projects page (menu and badge).
* Tests that mock Okta's FGA service (since there is no official mocking for now)

Current Auth Model:
image

Example 404 behavior:
<video src="https://github.com/concrete-utopia/utopia/assets/7003853/033d053b-f65c-49cb-ab01-1941ae1fd3d4"/>


server/migrations/008.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE if not exists project_access (
+    id serial,
+    project_id text not null references project(proj_id) ON DELETE CASCADE,
```
seanparsons, line 3: Does this imply that there will be multiple entries for each project?

liady: no - there is actually a unique constraint for it (`ADD CONSTRAINT "unique_project_access" UNIQUE ("project_id")`)

utopia-remix/app/services/fgaService.server.ts
```diff
@@ -0,0 +1,103 @@
+import { CredentialsMethod, OpenFgaClient } from '@openfga/sdk'
+import { AccessLevel } from '../types'
+
+export const fgaClient = new OpenFgaClient({
+  apiScheme: 'https',
+  apiHost: 'api.us1.fga.dev',
+  storeId: process.env.FGA_STORE_ID!,
```
seanparsons, line 7: I believe there's some handling for environment variables in this project, so that it's possible to make them required. Adding `!` on something in TypeScript just makes it think the type isn't nullable even if the value is null.

seanparsons (approved):

utopia-remix/app/env.server.ts
```diff
@@ -28,9 +32,13 @@ export const BrowserEnvironment: BrowserEnvironment = {
   EDITOR_URL: mustEnv('REACT_APP_EDITOR_URL'),
 }
 
-function mustEnv(key: string): string {
+function mustEnv(key: string, envFallback: Partial<Record<AppEnv, string>> = {}): string {
```
ruggi, line 35: you can use `optionalEnv` instead of `mustEnv`

utopia-remix/app/models/projectAccess.server.ts
```diff
@@ -0,0 +1,24 @@
+import { AccessLevel } from '../types'
+import { prisma } from '../db.server'
+import * as permissionsService from '../services/permissionsService.server'
+
+export async function setProjectAccess(params: {
+  projectId: string
+  accessLevel: AccessLevel
+}): Promise<void> {
+  await prisma.projectAccess.upsert({
```
ruggi, line 9: there two writes should be wrapped in a transaction

utopia-remix/app/routes/internal.projects.$id.access.tsx
```diff
@@ -0,0 +1,37 @@
...
+  )
+
+  try {
+    await setProjectAccess({ projectId: id, accessLevel: accessLevelNumber })
+  } catch (e) {
+    console.error('error', e)
+    return new Response('error', {
+      status: Status.INTERNAL_ERROR,
+    })
+  }
```
ruggi, line 34: all endpoints should return `ApiError`s for consistency, but you don't need to try/catch here since it's inside a `handle`

utopia-remix/app/routes/internal.projects.$id.access.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { setProjectAccess } from '../models/projectAccess.server'
+import { asNumber } from '../util/common'
+import { AccessLevel } from '../types'
+
+export async function action(args: ActionFunctionArgs) {
+  return handle(args, { POST: handleChangeProjectAccess })
+}
+
+export async function handleChangeProjectAccess(req: Request, params: Params<string>) {
+  // const user = await requireUser(req)
```
ruggi, line 14: shouldn't you'd need to be authenticated to perform this? 🤔

utopia-remix/app/routes/projects.tsx
```diff
@@ -860,6 +925,8 @@ const ActiveOperation = React.memo(({ operation }: { operation: Operation }) =>
         return 'Renaming'
       case 'restore':
         return 'Restoring'
+      case 'changeAccess':
+        return 'Changing access for'
```
ruggi, line 929: as a heads up, this has been changed so you'll probably need to resolve some conflict when merging w/ master after https://github.com/concrete-utopia/utopia/pull/4979 is merged

utopia-remix/app/services/fgaService.server.ts
```diff
@@ -0,0 +1,104 @@
...
+  projectId: string,
+  userId: string,
+  permission: UserProjectPermission,
+) {
+  const { allowed } = await fgaClient.check({
+    user: `user:${userId}`,
+    relation: permission,
+    object: `project:${projectId}`,
+  })
+  return !!allowed
```
ruggi, line 71: fyi I believe we never use double negations in the codebase to avoid confusion

ruggi (commented):
did a first pass, will return to this tomorrow!

utopia-remix/app/env.server.ts
liady: it's not optional for proper function, though (just uneeded for tests)

ruggi: I would definitely not put it there though imho, a `must` means it _must_ be set, you can set the related variable in the environment when running the tests

utopia-remix/app/services/fgaService.server.ts
```diff
@@ -0,0 +1,104 @@
...
+  'can_request_access',
+] as const
+
+type UserProjectPermission = (typeof userProjectPermission)[number]
+
+async function checkUserProjectPermission(
+  projectId: string,
+  userId: string,
+  permission: UserProjectPermission,
+) {
```
ruggi, line 65: could you add explicit return types here, in the `can<Action>Project` functions please? I'm trying to be more diligent in that regard myself, I think it'd be beneficial to have it consistent across the board

utopia-remix/app/util/common.ts
```diff
@@ -0,0 +1,6 @@
+export function asNumber(value?: unknown): number {
+  if (!['string', 'number'].includes(typeof value) || value === '') {
```
ruggi, line 2: shouldn't this be a simpler try/catch `parseInt` if it's a string, and just the number otherwise?

utopia-remix/app/routes/projects.tsx
```diff
@@ -724,7 +726,14 @@ const ProjectCard = React.memo(
           }}
         >
           <div style={{ display: 'flex', flexDirection: 'column', padding: 10, gap: 5, flex: 1 }}>
-            <div style={{ fontWeight: 600 }}>{projectTitle}</div>
+            <div style={{ fontWeight: 600, display: 'flex', gap: '10px', alignItems: 'center' }}>
+              <span>{projectTitle}</span>
+              <ProjectBadge
+                accessLevel={
+                  (project.ProjectAccess?.access_level as AccessLevel) || AccessLevel.PRIVATE
```
ruggi, line 733: these `as AccessLevel` look pretty suspicious to me, especially because they're a recipe for issues if we have a discrepancy between the data on the DB and the enum. I would instead suggest to have a helper function that gets the right `AccessLevel` (with a switch/assertNever) given a number

utopia-remix/app/routes/v1.project.$id.tsx
```diff
@@ -1,14 +1,40 @@
...
     OPTIONS: handleOptions,
-    GET: proxy,
+    GET: getProject,
   })
 }
 
+export async function getProject(req: Request, params: Params<string>) {
+  const { id: projectId } = params
+  ensure(projectId != null, 'project is null', Status.BAD_REQUEST)
+  let userId
```
ruggi, line 19: I think you can safely add a separate function that is the non-require variant of `requireUser` (which basically just does `getUserFromSessionId`) and use the `user|null` return. Moreover I don't understand if `ANON` is actually useful or not, if not I think it makes more sense to just drop it and rely on the user being null or not

utopia-remix/app/routes/v1.project.$id.tsx
```diff
@@ -1,14 +1,40 @@
...
+    userId = 'ANON'
+  }
+  const projectData = (await proxy(req, params)) as { ownerId: string }
+  const creatorId = projectData.ownerId
+  const isCreator = creatorId === userId
+
+  const allowed =
+    isCreator ||
+    (await hasUserProjectPermission(projectId, userId, UserProjectPermission.CAN_VIEW_PROJECT))
+  ensure(allowed, 'Access denied', Status.NOT_FOUND)
```
ruggi, line 33: If we're going with the decision of returning a 404 for unaccessible projects, we should also not leak information such as `Access denied` here but rather an equally generic `not found`-ish message (or instead going the opposite way and return a 403).

utopia-remix/app/routes/v1.project.$id.tsx
```diff
@@ -1,14 +1,40 @@
...
+  const { id: projectId } = params
+  ensure(projectId != null, 'project is null', Status.BAD_REQUEST)
+  let userId
+  try {
+    const user = await requireUser(req)
+    userId = user.user_id
+  } catch (e) {
+    userId = 'ANON'
+  }
+  const projectData = (await proxy(req, params)) as { ownerId: string }
```
ruggi, line 26: I'd drop the proxying here, since we're returning the data directly without streaming it and hence losing the related benefits. You can just grab the data directly from the DB at this point

ruggi (changes requested):
Aside from the specific comments I left on the implementation (all minor things) I think the PR changes are fine, except for the access control in the `v1.project.$id.tsx` route: what we want is introduce an access layer to the project operations, and it should be treated as such: we don't perform operations on a project unless the access layer is passed.
Therefore I would instead tackle this as a middleware-ish approach applied to our handlers, probably making it pluggable.
You have the benefit of not having to change the routes, only the authorization middleware, so we can work on it transparently as the business logic of the routes does not (and should not, unless for special cases) care about it.

If you add a middleware that works like this (just a stub)
```ts
function authorizeProjectAccess(req: Request, params: Params) {
  const projectId: string = <grab the project id from the params>
  const user: UserDetails | null = getOptionalUser()
  const projectOwnerId: string = getProjectOwnerId(projectId) // from the DB
  const canAccess = user.user_id === projectOwnerId || hasUserProjectPermission(...)
  ensure(canAccess, 'not found', 404)
}
```
you can attach it to the handler like a generic middleware where we want it.

Adding more elegant things (e.g. caching) is easier when all the logic in a single place like this one

liady:
@ruggi I agree, we should defintely have a middleware layer for endpoints. It was left out of this initial PR since it only handles the "can_view" permission (and not editing/collaborating).

utopia-remix/app/routes/v1.project.$id.tsx
liady: so is it better to just query the DB directly? the reason I chose to proxy here is just to allow for any logic the Haskell server does that the BFF is not (and should not be) aware of

ruggi: I just felt with the code involved here the proxying felt a bit unnecessary, but if going with the middleware solution I'd indeed just keep the proxying there for now (I'd lean on the latter)

ruggi:
> @ruggi I agree, we should defintely have a middleware layer for endpoints. It was left out of this initial PR since it only handles the "can_view" permission (and not editing/collaborating).

sure, but I don't see a reason for not going for the better option right now – also considering that with this change in its current form we already introduce technical debt (and a behavior discrepancy between that route and any other route used by the editor and the BFF). As this is an introductory PR it feels like a great place for stubbing a very simple middleware that we can improve upon going forward, rather than modifying a route that will need to be re-changed later :)

utopia-remix/app/env.server.ts
Rheeseyb: +1 we shouldn't be giving an inline fallback to `mustEnv` like this. We should either use `mustEnv` as is and provide the test value when running the tests, or use `optionalEnv` to provide the fallback inline. The danger of doing it like this is that it implicitly makes the thing optional.

utopia-remix/app/env.server.ts
```diff
@@ -8,8 +8,10 @@ declare global {
 
 export type AppEnv = 'local' | 'stage' | 'prod' | 'test'
 
+const serverEnvironemnt: AppEnv = mustEnv('APP_ENV') as AppEnv
```
ruggi, line 11: ?

utopia-remix/app/env.server.ts
ruggi: this is not resolved yet, right? 🤔

utopia-remix/app/util/api.server.ts
```diff
@@ -35,25 +36,47 @@ interface HandleRequest {
   params: Params<string>
 }
 
+export type ProjectAccessValidator = (request: Request, params: Params<string>) => Promise<boolean>
```
ruggi, line 39: given how they are used, there's not a reason for returning a value - either they succeed or not, and when failing the chain stops. Therefore either this returns a "true" value that is handled explicitly (e.g. throwing an exception) or it does not return anything because it will implicitly explode with e.g. `ensure` (or any other exception)

utopia-remix/app/util/api.server.ts
```diff
@@ -142,3 +165,13 @@ export async function requireUser(
     throw error
   }
 }
+
+export async function getUser(request: Request): Promise<UserDetails | null> {
```
ruggi, line 169: can you please use this function inside `requireUser`? this is a copy of the code over there, and it's _very_ important we don't have diverging behaviors for this stuff

utopia-remix/app/handlers/validators.ts
```diff
@@ -0,0 +1,37 @@
...
+): AccessValidator {
+  return async function (req: Request, params: Params<string>) {
+    const { id: projectId } = params
+    ensure(projectId != null, 'project id is null', Status.BAD_REQUEST)
+
+    const projectData = await getProjectById({ id: projectId })
+    ensure(projectData != null, 'Project not found', Status.NOT_FOUND)
+
+    let user = await getUser(req)
+    let userId = user?.user_id ?? 'ANON'
```
Rheeseyb, line 26: I really dislike this use of a hardcoded string here for an anonymous user. Since we're only using it in one place it's fine, but I'd argue that if we have to do this then we should probably provide the fallback string inside `checkUserProjectPermission` rather than inside the validator

utopia-remix/app/services/permissionsService.server.ts
```diff
@@ -0,0 +1,36 @@
+import { assertNever } from '../util/assertNever'
```
Rheeseyb, line 1: Is there any reason to keep this file and `fgaService.server.ts` separate?

utopia-remix/app/util/api.server.ts
```diff
@@ -35,25 +36,47 @@ interface HandleRequest {
...
+  handler: Handler
+  validator: AccessValidator
+}
+
+function toValidatedHandler(descriptor: Handler | ValidatedHandler): ValidatedHandler {
+  if (typeof descriptor === 'function') {
+    return { handler: descriptor, validator: ALLOW }
+  }
+  return descriptor
+}
```
ruggi, line 54: I think this is a bit too complex for what we need: imo it would make sense to have the handler data to include a list of extra handlers that run before (and after?), something like:
```ts
handle(args, {
  POST: {
    handler: handleDeleteProject,
    before: [
      authorizeProjectAccess,
      doSomethingElse,
      doAnotherThing
    ],
  }
})
```

and so on. This way we 1) don't need special handler types 2) we immediately think of them as a chain of independent (but sequential!) functions.

Sure this means updating all the existing routes, but imo a bit of repetition is better than a sacrifice in clarity.

utopia-remix/app/handlers/validators.ts
```diff
@@ -0,0 +1,37 @@
...
+  }: {
+    errorMessage?: string
+    status?: number
+  } = {},
+): AccessValidator {
+  return async function (req: Request, params: Params<string>) {
+    const { id: projectId } = params
+    ensure(projectId != null, 'project id is null', Status.BAD_REQUEST)
+
+    const projectData = await getProjectById({ id: projectId })
```
ruggi, line 22: we don't need the whole project data, only the ownership data, so the related function should only get that field from the db (and be renamed accordingly)

utopia-remix/app/routes/internal.projects.$id.access.tsx
```diff
@@ -0,0 +1,35 @@
...
+export async function action(args: ActionFunctionArgs) {
+  return handle(args, { POST: handleChangeProjectAccess })
+}
+
+export async function handleChangeProjectAccess(req: Request, params: Params<string>) {
+  const user = await requireUser(req)
+  const { id } = params
+  ensure(id != null, 'id is null', Status.BAD_REQUEST)
+
+  const project = await getProject({ id: id, owner_id: user.user_id })
```
ruggi, line 19: same here, we don't need the whole project, we just want to know if the user owns that project, which can be done with a dedicated function that returns the `owner_id` field from the DB

utopia-remix/app/models/projectAccess.server.ts
```diff
@@ -0,0 +1,26 @@
+import { AccessLevel } from '../types'
+import { prisma } from '../db.server'
+import * as permissionsService from '../services/permissionsService.server'
+
+export async function setProjectAccess(params: {
+  projectId: string
+  accessLevel: AccessLevel
+}): Promise<void> {
+  await prisma.$transaction(async (tx) => {
```
Rheeseyb, line 9: Nice

utopia-remix/app/routes/v1.project.$id.tsx
ruggi: I also don't think this is resolved, see also @Rheeseyb 's comment about `ANON`

utopia-remix/app/handlers/validators.ts
liady: Yeah, we need it since Okta needs some kind of string here, but I agree that inside the check function might be a better idea. Will change.

utopia-remix/app/util/api.server.ts
liady: Yes, I'll update all the routes (I actually think that an explicit `validator` key is important here, since it will enforce always having one)

utopia-remix/app/env.server.ts
liady: I just extracted it, since later calls to`mustEnv` inside the `ServerEnvironment` object need the server environment variable, so it can't be a key in the object initializtion

utopia-remix/app/util/api.server.ts
liady: I unified the session id fetching logic, but `requireUser` and `getUser` have slightly different behaviors (`requireUser` throws an error if there is an empty session id or no user, while `getUser` only returns null and it's ok)

utopia-remix/app/util/api.server.ts
ruggi: I think this is still overly-specifying responsibilities. I think an even simpler array of handlers would be a more elegant and flexible solution: I really wouldn't want to have to define another key in that type to support a different type of middleware which is functionally exactly the same as a project validator, but has the "unfortunate" property of doing "something else" with a request before going to the actual business logic.

Likewise if I think of _after_ handlers, the same logic would apply.

This is a stub of what I mean (that _also_ does something useful for the future - which is context passing) with a couple of examples: https://gist.github.com/ruggi/32d7e703b739f8b84bae33e4b8a09ba6

utopia-remix/app/routes/internal.projects.$id.access.tsx
ruggi: I don't think this is resolved, could you have a look at updating it please?

utopia-remix/app/handlers/validators.ts
ruggi: here too please (https://github.com/concrete-utopia/utopia/pull/4929#discussion_r1514195378)

utopia-remix/app/routes-test/internal.projects.$id.access.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    await createTestUser(prisma, { id: 'bar' })
+    await createTestSession(prisma, { key: 'the-key', userId: 'foo' })
+    await createTestProject(prisma, { id: 'one', ownerId: 'foo', title: 'project-one' })
+    await createTestProject(prisma, { id: 'two', ownerId: 'bar', title: 'project-two' })
+    jest.spyOn(permissionsService, 'setProjectAccess').mockResolvedValue()
+  })
+
+  it('changes access level to private', async () => {
+    const formData = new FormData()
+    formData.append('accessLevel', '0')
```
ruggi, line 35: it would be better to use the enum value here rather than hardcoding `0` (same for the other tests below)

utopia-remix/app/routes-test/internal.projects.$id.access.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  createTestProject,
+  createTestSession,
+  createTestUser,
+  newTestRequest,
+} from '../test-util'
+import { ApiError } from '../util/errors'
+import * as permissionsService from '../services/permissionsService.server'
+import { AccessLevel } from '../types'
+
+describe('handleChangeAccess', () => {
```
ruggi, line 15: could you please also test the case where the user is not authenticated (since it's part of the endpoint logic)

utopia-remix/app/services/fgaService.server.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export const fgaClient = new OpenFgaClient({
+  apiScheme: 'https',
+  apiHost: 'api.us1.fga.dev',
+  storeId: ServerEnvironment.FGA_STORE_ID,
+  // authorizationModelId: 'YOUR_MODEL_ID', // Optionally, you can specify a model id to target, which can improve latency
+  credentials: {
+    method: CredentialsMethod.ClientCredentials,
+    config: {
+      apiTokenIssuer: 'fga.us.auth0.com',
```
ruggi, line 13: honest question because I don't know: should this (and below) be configurable/put in the enviornment vars?

utopia-remix/app/services/permissionsService.server.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as fgaService from './fgaService.server'
+
+const ANONYMOUS_USER_ID = '__ANON__'
+
+export async function hasUserProjectPermission(
+  projectId: string,
+  requestUserId: string | null,
+  permission: UserProjectPermission,
+) {
+  const userId = requestUserId ?? ANONYMOUS_USER_ID
```
ruggi, line 12: again I think there's no need for this `__ANON__` string. You can just update the `canViewXXX` functions to receive a `null` and return false immediately (it would be good to also add the return type to the function signature)

utopia-remix/app/util/api.server.spec.ts
```diff
@@ -52,3 +52,48 @@ describe('requireUser', () => {
...
+    const request = newTestRequest({ method: 'POST' })
+    const response = await (handle(
+      { request: request, params: {} },
+      {
+        POST: {
+          handler: async () => {
+            return { data: 'success' }
+          },
+          validator: async () => {
+            throw new ApiError('failed', 400)
```
ruggi, line 87: for consistency, could you use `Status.[THE METHOD]` instead of hardcoded numbers when dealing with http status codes?

utopia-remix/app/handlers/validators.ts
```diff
@@ -0,0 +1,34 @@
+import { AccessValidator, ensure, getUser } from '../util/api.server'
+import { UserProjectPermission } from '../types'
+import { Params } from '@remix-run/react'
+import { Status } from '../util/statusCodes'
+import { hasUserProjectPermission } from '../services/permissionsService.server'
+import { getProjectById, getProjectOwnerById } from '../models/project.server'
```
ruggi, line 6:
```suggestion
import { getProjectOwnerById } from '../models/project.server'
```

utopia-remix/app/models/project.server.ts
```diff
@@ -22,6 +23,42 @@ export async function listProjects(params: { ownerId: string }): Promise<Project
   })
 }
 
+export async function getProject(params: {
```
ruggi, line 26: this function can be removed

utopia-remix/app/models/project.server.ts
```diff
@@ -22,6 +23,42 @@ export async function listProjects(params: { ownerId: string }): Promise<Project
...
+    select: { owner_id: true },
+    where: {
+      proj_id: params.id,
+      OR: [{ deleted: null }, { deleted: false }],
+    },
+  })
+  return project?.owner_id ?? null
+}
+
+export async function getProjectById(params: {
```
ruggi, line 51: this function can be removed too

utopia-remix/app/services/permissionsService.server.ts
liady: but this is not actually the case, since in public projects we do allow anonymous users to view them. So when a project is confgured so that `user:*` can view it, we need to send to Okta's FGA some user id (even if it's anonymous) for it to return `true`

utopia-remix/app/routes-test/internal.projects.$id.access.spec.ts
liady: should we test it here? since the validator is actually the one who checks it (and now not the endpoint itself)

utopia-remix/app/routes/internal.projects.$id.access.tsx
liady: removed (the validator checks it now)

utopia-remix/app/routes-test/internal.projects.$id.access.spec.ts
liady: fixed

utopia-remix/app/util/api.server.spec.ts
liady: fixed

utopia-remix/app/handlers/validators.ts
liady: fixed

utopia-remix/app/models/project.server.ts
liady: fixed

utopia-remix/app/models/project.server.ts
liady: fixed

utopia-remix/app/routes-test/internal.projects.$id.access.spec.ts
ruggi: the endpoint has a `requireUser` call which is checking authentication, not authorization, and imo the tests should reflect that. e.g. https://github.com/concrete-utopia/utopia/pull/4929#discussion_r1507942054 ← this should have had the tests fail

utopia-remix/app/services/fgaService.server.ts
liady: added to env vars, just to prepare for future changes (currently those are the only values, but they can definitely change)

utopia-remix/app/services/permissionsService.server.ts
ruggi: huh ok, I did not know about okta's requiring a user id string

utopia-remix/app/routes/internal.projects.$id.access.tsx
```diff
@@ -18,13 +17,10 @@ export async function action(args: ActionFunctionArgs) {
 }
 
 export async function handleChangeProjectAccess(req: Request, params: Params<string>) {
-  const user = await requireUser(req)
+  await requireUser(req)
```
ruggi, line 20: wait! by removing the user ownership check this means that any user can change the project access of any other user's projects 😱 (or am i hallucinating?) (if that's the case this should **absolutely** tested)

ruggi: ah yeah I just checked: there should be absolutely a test for making sure bob cannot change alice's project access

liady: no, since the validator is checking it:
```
export async function action(args: ActionFunctionArgs) {
  return handle(args, {
    POST: {
      handler: handleChangeProjectAccess,
      validator: validateProjectAccess(UserProjectPermission.CAN_MANAGE_PROJECT),
    },
  })
}
```
this was the actual concept of the validator - to take the functionality of access check from the handler

ruggi: Sure. Can you add a test for this then? the validator I mean.

_Note for future selves: it'd be good to have full tests for the whole actions, not just handlers._

liady: in any case I added a test for the entire `action` that checks for unauthorized access (though again it mainly checks the validator logic, and not this specific route)

liady: yes, I started testing entire actions (for the `/project` endpoint and the `/access` endpoint), this should definitely be the standard now (since it checks access validation as well)

liady: (fixed)

utopia-remix/app/routes/internal.projects.$id.restore.tsx
```diff
@@ -3,9 +3,15 @@ import { Params } from '@remix-run/react'
...
 import { ensure, handle, requireUser } from '../util/api.server'
 import { Status } from '../util/statusCodes'
+import { ALLOW } from '../handlers/validators'
 
 export async function action(args: ActionFunctionArgs) {
-  return handle(args, { POST: handleRestoreDeletedProject })
+  return handle(args, {
+    POST: {
+      handler: handleRestoreDeletedProject,
+      validator: ALLOW,
```
ruggi, line 12: this should also validate the project access

utopia-remix/app/routes/v1.github.save.$projectId.tsx
```diff
@@ -10,6 +11,6 @@ export async function loader(args: LoaderFunctionArgs) {
 
 export async function action(args: ActionFunctionArgs) {
   return handle(args, {
-    POST: proxy,
+    POST: { handler: proxy, validator: ALLOW },
```
ruggi, line 14: this should also validate the project access

utopia-remix/app/routes/v1.project.$id.$filename.tsx
```diff
@@ -1,10 +1,11 @@
 import { LoaderFunctionArgs } from '@remix-run/node'
 import { proxy } from '../util/proxy.server'
 import { handle, handleOptions } from '../util/api.server'
+import { ALLOW } from '../handlers/validators'
 
 export async function loader(args: LoaderFunctionArgs) {
   return handle(args, {
     OPTIONS: handleOptions,
-    GET: proxy,
+    GET: { handler: proxy, validator: ALLOW },
```
ruggi, line 9: this should also validate the project access

utopia-remix/app/routes/v1.project.$id.metadata.tsx
```diff
@@ -1,10 +1,11 @@
 import { LoaderFunctionArgs } from '@remix-run/node'
 import { proxy } from '../util/proxy.server'
 import { handle, handleOptions } from '../util/api.server'
+import { ALLOW } from '../handlers/validators'
 
 export async function loader(args: LoaderFunctionArgs) {
   return handle(args, {
     OPTIONS: handleOptions,
-    GET: proxy,
+    GET: { handler: proxy, validator: ALLOW },
```
ruggi, line 9: this should also validate the project access

utopia-remix/app/routes/v1.project.$id.owner.tsx
```diff
@@ -1,10 +1,11 @@
 import { LoaderFunctionArgs } from '@remix-run/node'
 import { proxy } from '../util/proxy.server'
 import { handle, handleOptions } from '../util/api.server'
+import { ALLOW } from '../handlers/validators'
 
 export async function loader(args: LoaderFunctionArgs) {
   return handle(args, {
     OPTIONS: handleOptions,
-    GET: proxy,
+    GET: { handler: proxy, validator: ALLOW },
```
ruggi, line 9: this should also validate the project access

utopia-remix/app/routes/v1.projectid.tsx
```diff
@@ -10,6 +11,9 @@ export async function loader(args: LoaderFunctionArgs) {
 
 export async function action(args: ActionFunctionArgs) {
   return handle(args, {
-    POST: proxy,
+    POST: {
+      handler: proxy,
+      validator: ALLOW,
```
ruggi, line 16: this should also validate the project access

utopia-remix/app/handlers/validators.ts
```diff
@@ -0,0 +1,34 @@
...
+  {
+    errorMessage,
+    status,
+  }: {
+    errorMessage?: string
+    status?: number
+  } = {},
+): AccessValidator {
+  return async function (req: Request, params: Params<string>) {
+    const { id: projectId } = params
```
ruggi, line 19: the problem here is that this is requiring the route to be defined with `$projectId` in the route name. for example, all `internal.project.` routes use instead `$id`. To make things simpler either we rename all the routes to use `$projectId` (but it means that we need to remember to do it like this _every time_), or we instead pass the right param key when invoking the validator

Rheeseyb: This is doing the opposite (asserting that the route is named `$id` and then renaming the param to `projectId`), but the point about the naming being implicitly required is valid here, because there will definitely be routes (e.g. `v1.github.save.$projectId`) where it makes more sense to use `$projectId` rather than `$id`.

ruggi: yeah, sorry, flipped the naming but the argument stands

liady: isn't it actually requiring `$id` in the project name and just renaming it to `projectId`?

ruggi: yeah, see @Rheeseyb comment

utopia-remix/app/routes/v1.projectid.tsx
Rheeseyb: This route doesn't take a param - it's for creating (and returning) a new project ID, so there is nothing to validate

ruggi: abort abort

utopia-remix/app/routes/v1.projects.tsx
```diff
@@ -1,10 +1,15 @@
...
+import { validateProjectAccess } from '../handlers/validators'
+import { UserProjectPermission } from '../types'
 
 export async function loader(args: LoaderFunctionArgs) {
   return handle(args, {
     OPTIONS: handleOptions,
-    GET: handleListProjects,
+    GET: {
+      handler: handleListProjects,
+      validator: validateProjectAccess(UserProjectPermission.CAN_VIEW_PROJECT),
```
Rheeseyb, line 12: How is the validator working here when there aren't any params?

utopia-remix/app/routes/v1.projectid.tsx
liady: this doesn't have a project yet, correct? ( this only creates a new id)

utopia-remix/app/routes/internal.projects.$id.restore.tsx
liady: fixed

utopia-remix/app/routes/v1.github.save.$projectId.tsx
liady: fixed

utopia-remix/app/routes/v1.project.$id.$filename.tsx
liady: fixed

utopia-remix/app/routes/v1.project.$id.owner.tsx
liady: fixed

utopia-remix/app/handlers/validators.ts
liady: added an option to pass additional config to the validator

utopia-remix/app/routes/v1.projects.tsx
liady: defintiely right. It seems that this route indeed shouldn't work (shouldn't have used a validator at all, was a bad search/replace when added all of them), 
but looks like perhaps it is not actually used (since `/projects` leads to our new `projects.tsx` screen)?
in any case I reverted it back to `ALLOW`, since perhaps there is a place that calls it using `v1`.
I also added a way to pass the project param to the validator (according to Rheese's comment about `x.$projectId.ts` routes). We discussed it and perhaps the proper solution there will probably be to have another handler that extracts the project details before validation, but seems like it is out of scope of this "Project Visibility" PR. (but can defintiely be a subsequent "BFF middleware infra" PR).

utopia-remix/app/routes/v1.project.$id.metadata.tsx
liady: fixed

utopia-remix/app/models/project.server.ts
```diff
@@ -23,12 +23,15 @@ export async function listProjects(params: { ownerId: string }): Promise<Project
   })
 }
 
-export async function getProjectOwnerById(params: { id: string }): Promise<string | null> {
+export async function getProjectOwnerById(
+  params: { id: string },
+  config: { allowDeleted: boolean } = { allowDeleted: false },
```
ruggi, line 28: the wording is a bit off, this is not "allowing: but rather _"including"_ the soft-deleted prpject

utopia-remix/app/handlers/validators.ts
```diff
@@ -10,16 +10,20 @@ export function validateProjectAccess(
   {
     errorMessage,
     status,
+    idRouteParam = 'id',
```
ruggi, line 13: please let's not give an arbitrary default here and instead make this explicit (sure sure it'll need updating across the board but it's good)

utopia-remix/app/models/project.server.ts
```diff
@@ -23,12 +23,15 @@ export async function listProjects(params: { ownerId: string }): Promise<Project
...
+): Promise<string | null> {
   const project = await prisma.project.findFirst({
     select: { owner_id: true },
     where: {
       proj_id: params.id,
-      OR: [{ deleted: null }, { deleted: false }],
+      OR: config.allowDeleted ? [] : [{ deleted: null }, { deleted: false }],
```
ruggi, line 34: this looks fine, but _**please**_ add a test for this. Prisma can be surprisingly non-intuitive and we 10000000% want to be sure we don't get incorrect selections on soft-deleted projects

ruggi: case in point for super scary, counter-intuitive Prisma behaviors: https://github.com/concrete-utopia/utopia/blob/77adcec58e49177f58b5c8a514530bb5b38353bf/utopia-remix/app/db.server.ts#L18-L24

liady: yes! added tests 🙏

utopia-remix/app/handlers/validators.ts
liady: fixed by passing a `getProjectId` explicit function

utopia-remix/app/models/project.server.ts
liady: fixed

utopia-remix/app/handlers/validators.ts
liady: (writing here as well for completion - changed to a function that gets the project id)

utopia-remix/app/models/project.server.ts
```diff
@@ -22,6 +23,26 @@ export async function listProjects(params: { ownerId: string }): Promise<Project
   })
 }
 
+export async function getProjectOwnerById(
+  params: { id: string },
+  config: { includeDeleted: boolean } = { includeDeleted: false },
+): Promise<string | null> {
+  const whereOptions: {
```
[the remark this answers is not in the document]
liady: I changed it to `let`, but I'm wondering how relevant will be a very specific function here in the scope of the file (since if this is the case - we will probably want to support more flags in this generic function)

utopia-remix/app/test-util.ts
```diff
@@ -87,6 +101,17 @@ export async function truncateTables(models: DeletableModel[]) {
   }
 }
 
+export async function clearDb(prisma: UtopiaPrismaClient) {
```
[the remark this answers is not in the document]
liady: I actually added it to have a place where you know it "clears everything", instead of having your tests fail in the future just because someone added a new table that is not being cleared with the explicit `truncateTables([...])` call...
(so for example if we're adding a new `ProjectAccessRequest` table, we only need to make sure to clear it here instead of having to chase all the tests that are calling `truncateTable` and changing them)
wdyt? don't you think it's more robust to have one place that is in charge of "resetting" the db for tests?

utopia-remix/app/types.ts
```diff
@@ -33,6 +41,44 @@ export function userToCollaborator(user: UserDetails): Collaborator {
...
+export const AccessLevel = {
+  PRIVATE: 0,
+  PUBLIC: 1,
+  WITH_LINK: 2,
+} as const
+
+export type AccessLevel = (typeof AccessLevel)[keyof typeof AccessLevel]
+
+export function asAccessLevel(accessLevel: number | undefined | null): AccessLevel | null {
+  if (accessLevel == null) {
```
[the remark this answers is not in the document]
liady: yup, fixed

utopia-remix/app/routes/v1.project.$id.tsx
```diff
@@ -1,16 +1,31 @@
 import { ActionFunctionArgs, LoaderFunctionArgs } from '@remix-run/node'
 import { proxy } from '../util/proxy.server'
 import { handle, handleOptions } from '../util/api.server'
+import { Status } from '../util/statusCodes'
+import { UserProjectPermission } from '../types'
+import { ALLOW, validateProjectAccess } from '../handlers/validators'
```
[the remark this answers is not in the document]
liady: fixed!

utopia-remix/app/test-util.ts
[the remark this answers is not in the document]
ruggi: I mean, the "really" clean way of doing a reset would be to `truncate` the db and apply the migrations to start from a clean slate.
My point is: do you want a single function that clears everything? Cool, but then let's use it _everywhere_. The current solution/convention applied consistently in all test files is to be very explicit and deliberate about the db set up and teardown.
The motivations for a function to which you explicitly pass a list of tables to clean up are multiple, it forces you to be deliberate and think about the tables that are involved in the things you're testing (e.g. including cascades) and it scales well.

However a single cleanup function option that zaps everything can be fine too (e.g. in the future we need to seed data globally for the tests, in that case it can get handier). I picked my option when writing the initial tests but I'm fine with changing it, but please let's pick an option and be consistent with it.

My suggestion would be to not inflate this already huge diff with this kind of refactors and have a separate PR for them instead.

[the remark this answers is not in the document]
liady: good point about keeping this PR focused - changed it back to `truncateTables` 👍

Rheeseyb (approved):

PR #4932, opened 2024-02-21
Adding property access, element access and JS identifiers.

seanparsons:
**Problem:**
We currently can't express expressions like `props.something`, `props.something[1]` or even an identifier without them being treated as arbitrary blocks which makes them black boxes to our understanding.

**Fix:**
Created the representations of property accessors, element accessors and identifiers and made them part of the `JSExpression` type. But critically this does not change the parser, so currently these values are not produced by anything. That and a couple of `FIXME` cases will be addressed in subsequent work.

**Commit Details:**
- Added `JSIdentifier`, `JSPropertyAccess` and `JSElementAccess` types along with supporting factory functions.
- Incorporated the new types as new cases of `JSExpression`.
- Filled in a lot of cases that handle the new types in various functions.

### DIFF
--- a/editor/src/core/shared/jsx-attributes.ts
+++ b/editor/src/core/shared/jsx-attributes.ts
@@ -3,7 +3,17 @@ import type { MapLike } from 'typescript'
 import { UtopiaUtils } from 'utopia-api/core'
 import { findLastIndex, uniqBy } from './array-utils'
 import type { Either } from './either'
-import { isLeft, isRight, left, mapEither, reduceWithEither, right, traverseEither } from './either'
+import {
+  applicative2Either,
+  flatMapEither,
+  isLeft,
+  isRight,
+  left,
+  mapEither,
+  reduceWithEither,
+  right,
+  traverseEither,
+} from './either'
 import type {
   JSXArrayElement,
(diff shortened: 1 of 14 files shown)


bkrmendy (approved):

gbalint (approved):

editor/src/core/shared/jsx-attributes.ts
```diff
@@ -186,6 +221,17 @@ export function jsxAttributeToValue(
   switch (attribute.type) {
     case 'ATTRIBUTE_VALUE':
       return attribute.value
+    case 'JS_IDENTIFIER':
+      return requireResult[attribute.name] ?? inScope[attribute.name]
```
balazsbajorics, line 225: what is requireResult in our case? does it represent an import? from just reading the code here it looks to me like we want to prioritize inScope (assuming that contains all the nested scopes, including the closest scope around this identifier expression)

seanparsons: Ah, yes, I think that was my intention, but not what I wrote.

balazsbajorics (approved):

PR #4936, opened 2024-02-21
Remix projects: don't use colors directly

ruggi:
**Problem:**

The `colors` object is used directly in multiple places inside `style` props, but that _may_ be responsible for the hydration problems we're seeing in production.

**Fix:**

This PR gets rid of that and instead uses `sprinkle` classnames.

(Also reverted to a more sane usage of radix UI's component imports).

### DIFF
--- a/utopia-remix/app/routes/projects.tsx
+++ b/utopia-remix/app/routes/projects.tsx
@@ -549,7 +549,7 @@ const ProjectCard = React.memo(
                   className={sprinkles({
                     boxShadow: 'shadow',
                     color: 'white',
-                    background: 'primary',
+                    backgroundColor: 'primary',
                   })}
                 >
                   {when(collaborator.avatar === '', multiplayerInitialsFromName(collaborator.name))}
--- a/utopia-remix/app/routes/projects.tsx
+++ b/utopia-remix/app/routes/projects.tsx
@@ -1,8 +1,5 @@
 import React from 'react'
-import {
-  Root as DropdownMenuRoot,
-  Trigger as DropdownMenuTrigger,
-} from '@radix-ui/react-dropdown-menu'
+import * as DropdownMenu from '@radix-ui/react-dropdown-menu'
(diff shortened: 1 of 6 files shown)


lankaukk (approved):

bkrmendy (approved):

utopia-remix/app/routes/projects.tsx
```diff
@@ -551,7 +546,11 @@ const ProjectCard = React.memo(
                     filter: project.deleted === true ? 'grayscale(1)' : undefined,
                   }}
                   title={collaborator.name}
-                  className={sprinkles({ boxShadow: 'shadow' })}
+                  className={sprinkles({
+                    boxShadow: 'shadow',
+                    color: 'white',
+                    background: 'primary',
```
Rheeseyb, line 552: Should this be `backgroundColor`?

liady (approved):

utopia-remix/app/styles/contextMenu.css.ts
```diff
@@ -29,10 +29,9 @@ export const contextMenuDropdown = recipe({
     sprinkles({
       borderRadius: 'small',
       background: 'white',
-      color: 'lightModeBlack'
+      color: 'lightModeBlack',
```
balazsbajorics, line 32: does it mean we don't yet run eslint / prettier for these files?

ruggi: I think this was untouched after we enabled prettier correctly

utopia-remix/app/routes/projects.tsx
ruggi: yeah it should have been

ruggi: [`5999810` (#4936)](https://github.com/concrete-utopia/utopia/pull/4936/commits/59998107213f557c7040bbfb6352208ea8864607)

PR #4937, opened 2024-02-21
Remix projects: sort and filter hooks

ruggi:
This PR extracts the sorting and filtering logic out of the projects component and makes it reusable with a couple of hooks, which also don't duplicate the project elements.


utopia-remix/app/util/useSortCompareProject.tsx
```diff
@@ -0,0 +1,45 @@
+import moment from 'moment'
```
gbalint, line 1: Minor thing, but we use kebab-case in filenames not camel-case

gbalint: Ignore me, this is the remix app :)

liady (approved):

utopia-remix/app/util/useSortCompareProject.tsx
ruggi: no it's good, I just realized filename casing is mixed 🤢 I'll use kebab for this one and have another cleanup PR

utopia-remix/app/util/use-sort-compare-project.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const sortAscending = useProjectsStore((store) => store.sortAscending)
+
+  return React.useCallback(
+    (a: ProjectWithoutContent, b: ProjectWithoutContent): number => {
+      switch (sortCriteria) {
+        case 'title':
+          return sortAscending ? a.title.localeCompare(b.title) : b.title.localeCompare(a.title)
+        case 'dateCreated':
+          return sortAscending
+            ? moment(a.created_at).unix() - moment(b.created_at).unix()
```
gbalint, line 18: Is there a reason to use a lib for this? Date.getTime() is enough for this, isn't it?

ruggi: Yep, but those are not dates – they are strings, Prisma is lying unfortunately. So instead of doing optimistic parsing of whatever strings come out of the ORM, I used `moment` to be sure the parsing is reliable

PR #4946, opened 2024-02-22
Collaborators from the DB

### LINKED ISSUE #4944: BFF collaborators in the DB

ruggi:
Fix #4944

This is a followup to https://github.com/concrete-utopia/utopia/pull/4923, that turns the Utopia DB into the source of truth for collaborators, dropping them from the storage entirely.

1. When the user opens a project, the editor will load the collaborators while **contextually** adding the user to the project collaborators, if missing.
2. When the editor loads, the collaborators are loaded and a listener on the room is added so that the state collaborators are refreshed when needed.
3. If the editor is not using the BFF, for backwards compatibility the collaborators are loaded from the LB client. ⚠️ This is only temporary to allow for smooth sailing in `.app` while we don't enable the BFF there too.

The `collaborators` field in the LB object is kept **just as a compatibility measure** until we enable the BFF in production. A consequent PR should drop entirely that field _and_ the other related things marked as `_DEPRECATED`.


editor/src/core/commenting/comment-hooks.tsx
```diff
@@ -237,25 +238,21 @@ export function useAddMyselfToCollaborators() {
...
     }
-    maybeUpdateCollaborators()
-  }, [addMyselfToCollaborators, collabs, projectId, maybeUpdateCollaborators])
+  }, [addMyselfToCollaborators, collabs, projectId])
 }
 
 export function useCollaborators() {
-  return useStorage((store) => store.collaborators)
+  return useEditorState(
+    Substores.restOfEditor,
```
gbalint, line 252: I suggest to add a new Substore for multiplayer related stuff.
Or maybe use an atom for collaborators instead of the editor state (it seems to be quite independent from other parts of the state)

ruggi: @gbalint I added a multiplayer substate here [`fbb22ca` (#4946)](https://github.com/concrete-utopia/utopia/pull/4946/commits/fbb22ca6154824df4ed206709effd2c9c59a746f) . would that work for you or would you prefer the atom instead?

website-next/components/common/env-vars.ts
```diff
@@ -36,7 +36,7 @@ type BackendType =
   | 'direct' // direct calls to the backend
 
 const LOCAL_BACKEND_PORTS: { [type in BackendType]: number } = {
-  direct: 8001,
+  direct: 8000,
```
Rheeseyb, line 39: What is with this change?

ruggi: while I was trying this out locally I was hitting some CORS issues with the haskell server, which were solved by hitting the same origin as the editor

Rheeseyb: Riiight ok 👍

Rheeseyb (approved):

bkrmendy (approved):

PR #4951, opened 2024-02-26
Change the 404 page

liady:
A better design for the 404 page (when a project is private / not found).
To test - just try `/p/some-made-up-project` (later when the visibility PR will be merged it will also be shown for private projects)

image


editor/src/templates/project-not-found/ProjectNotFound.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react'
+import PyramidLight404 from './404_pyramid_light.png'
```
balazsbajorics, line 2: we have a `resources` folder in the editor, images saved there are served like so `https://github.com/concrete-utopia/utopia/blob/34a6bb591ba1fa30c380c9904d8d87139a27d5a5/editor/src/templates/index.html#L150`. 

I believe the images in `resources` are served with proper caching and other such stuff as opposed to being expensively served by the haskell server.

editor/src/templates/project-not-found/ProjectNotFound.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <div style={{ fontSize: '42px' }}>Project not found.</div>
+        <div style={{ fontSize: '22px', width: '430px', textAlign: 'center', lineHeight: '40px' }}>
+          Either this project does not exist, or you do not have access to it.
+        </div>
+        <a
+          href={`${UTOPIA_BACKEND_BASE_URL}projects`}
+          rel='noopener noreferrer'
+          style={{ textDecoration: 'none' }}
+        >
+          <div
```
ruggi, line 51: wouldn't it be better to use a `button` here? (including our own `Button` if possible)

editor/src/templates/project-not-found/ProjectNotFound.tsx
```diff
@@ -0,0 +1,72 @@
...
+        style={{
+          display: 'flex',
+          flexDirection: 'column',
+          gap: '40px',
+          alignItems: 'center',
+          paddingBottom: '30px',
+        }}
+      >
+        <div
+          style={{ fontSize: '160px', fontWeight: 600, fontStyle: 'italic', lineHeight: 'normal' }}
```
ruggi, line 38: I believe `lineHeight` is redundant here

editor/src/templates/project-not-found/ProjectNotFound.tsx
```diff
@@ -0,0 +1,72 @@
...
+              display: 'flex',
+              alignItems: 'center',
+              justifyContent: 'center',
+              color: 'white',
+              backgroundColor: '#0075f9',
+              borderRadius: '90px',
+              padding: '10px 30px',
+              transition: 'background-color 0.3s',
+            }}
+            onMouseOver={(e) => (e.currentTarget.style.backgroundColor = '#5DA9FF')}
```
ruggi, line 63: I think you can achieve the same thing with `css={{...}}` and the `&:hover` prop

bkrmendy (approved):

gbalint (approved):

PR #4956, opened 2024-02-26
Data picker - special case the `children` prop

bkrmendy:
# Try it [here](https://utopia.fish/p/cc22c5f3-charming-saturnalia/?branch_name=feature-children-prop-dedicated-picking)

## Problem
When the `children` prop is requested via `registerModule`/`registerInternalComponent`, and the data picker is being used to pick a variable for the `children` prop, it puts the picked variable in the wrong place.

For example, consider this component:

```tsx
<NavLink key={item.id} prefetch="intent" style={activeLinkStyle} to={url}>
  <p>Children should go here</p>
</NavLink>
```

When picking a variable to replace the `p` inside `Navlink`, the result becomes this:
```tsx
<NavLink key={item.id} prefetch="intent" style={activeLinkStyle} to={url} children={item.title}>
  <p>Children should go here</p>
</NavLink>
```

instead of:

```tsx
<NavLink key={item.id} prefetch="intent" style={activeLinkStyle} to={url}>
  {item.title}
</NavLink>
```

## Fix
Special case the `children` prop, and instead of setting a prop on the selected element, put the selected variable into the slot for `children`.


editor/src/components/editor/actions/actions.tsx
```diff
@@ -2231,6 +2232,30 @@ export const UPDATE_FNS = {
...
   },
+  INSERT_ATTRIBUTE_OTHER_JAVASCRIPT_INTO_ELEMENT: (
+    action: InsertAttributeOtherJavascriptIntoElement,
+    editor: EditorModel,
+  ): EditorModel => {
+    const withNewElement = modifyUnderlyingTargetElement(action.parent, editor, (element) => {
+      switch (element.type) {
+        case 'JSX_CONDITIONAL_EXPRESSION':
+        case 'JSX_FRAGMENT':
+          // only JSX elements are handled in this action
```
balazsbajorics, line 2243: should this throw an Error?

editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
```diff
@@ -70,6 +70,14 @@ function valueToDisplay(option: VariableOption): string {
   }
 }
 
+function isChildrenProp(path: PropertyPath): boolean {
+  return (
+    path.propertyElements.length > 0 &&
```
balazsbajorics, line 75: what if the path is longer than 1 element? it feels like we should error out in that case (not necessarily here, but this code reminded me of the problem)

balazsbajorics (approved):
looks good, left some minor questions

editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
bkrmendy: fair concern, I'll address this in follow-up PR so the decision around this edge isn't muddled together with the code on this PR

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2231,6 +2232,30 @@ export const UPDATE_FNS = {
       selectedViews: newSelectedViews,
     }
   },
+  INSERT_ATTRIBUTE_OTHER_JAVASCRIPT_INTO_ELEMENT: (
```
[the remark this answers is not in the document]
bkrmendy: how about `INSERT_JS_INTO_JSX_CHILDREN`? Trying to balance name length vs information density here :D

seanparsons (approved):

PR #4959, opened 2024-02-26
Identifier cartouche

balazsbajorics:
image

**Problem**
In the component inspector by default the controls for props show the spied value of the property. We show them in black if it's a 'simple' value, and we show them in blue if the value's source is a javascript expression. 
Since we have the data picker, if we pick some data, say `collection[0].image`, we will show `https://.../.../long-image-url.jpg` in blue. There is a disconnect between what the picker shows and what the actual choice ends up looking like. Also, knowing the value in case of an expression is not as important as seeing the actual expression.

**Aside**: types of supported "data" from the data picker:
1. `prop={myData}` - js identifier
2. `prop={myData[0]}` - js element access
3. `prop={myData.attr}` - js property access

**Fix**
For the data types we support and recognize, show a "Cartouche" control, a blue pill control which indicates that we recognize your expression and associate with an option of the data picker.
Clicking on the cartouche also opens the data picker

**Not part of this PR**
- when opening the data picker, we should indicate the currently selected option with a highlight / default selection
- We should give the option to delete a cartouche
- We should still show the runtime values in a second row or as a tooltip option
- We should show a cartouche for sub-fields, for example `prop={{image: images[0]}}` here the row `image` should show the cartouche for `images[0]`
- We should show the cartouche for Array and Object in place of the now-empty row


editor/src/components/canvas/ui-jsx-canvas-renderer/jsx-element-child-to-text.ts
```diff
@@ -0,0 +1,133 @@
+import type { JSXElementChild } from '../../../core/shared/element-template'
+import { assertNever } from '../../../core/shared/utils'
+
+export function jsxElementChildToText(
```
balazsbajorics, line 4: this is just `jsxElementChildToText` from`ui-jsx-canvas-element-renderer-utils.tsx`, moved verbatim to a dedicated file

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -119,50 +126,82 @@ function useComponentPropsInspectorInfo(
...
-        return <VectorPropertyControl {...props} controlDescription={controlDescription} />
-      case 'jsx':
-        return <JSXPropertyControl {...props} controlDescription={controlDescription} />
-      default:
-        return null
+  }
+
+  const attributeExpression = props.propMetadata.attributeExpression
+
+  if (attributeExpression != null && PP.depth(props.propPath) === 1) {
```
balazsbajorics, line 136: I recommend hiding whitespace for this

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -358,14 +399,20 @@ const RowForBaseControl = React.memo((props: RowForBaseControlProps) => {
         variant='<--1fr--><--1fr-->|-18px-|'
       >
         {propertyLabel}
-        <div ref={dataPickerButtonData.setReferenceElement}>
+        <div
+          style={{
+            minWidth: 0, // this ensures that the div can never expand the allocated grid space
```
liady, line 404: Nice fix. 
off topic - at the future I think we should actually fix it in the `UIGridRow` component, as I imagine it will be a common issue (since the use of `fr` can yield unexpected results)

Rheeseyb (approved):

seanparsons (approved):

gbalint (approved):

liady (approved):

editor/src/components/inspector/sections/component-section/component-section.tsx
balazsbajorics: good point!

PR #4963, opened 2024-02-27
Fix grid width by setting 'minmax' in fr units

liady:
**Problem:**
Grid `1fr` columns don't always behave as expected.
image

**Fix:**
In places where we want the exact "split" behavior we should use `minmax(<min>, 1fr)`
image

See here:
https://css-tricks.com/you-want-minmax10px-1fr-not-1fr/


gbalint:
Nice! Can you add a little padding to the right too? So it looks similarly to the overflowing string control (like in the Title prop control).

ruggi (approved):

gbalint (approved):

seanparsons (approved):

PR #4966, opened 2024-02-27
Parsing New JS Expression Types.

seanparsons:
**Problem:**
We have the types for the new expression types that we want, but there's nothing to actually parse those from the code a user writes.

**Fix:**
Two places were the core of this which is `parseAttributeExpression` and `parseOutJSXElements`, which are responsible for parsing the properties of a JSX element and parsing the JSX elements and content within them. Both had changes to incorporate some common parsing of the 3 new types, with some adaptations to their particular call sites.

Otherwise there were a lot of changes relating to editing text, as this crosses over in a few places. With this necessitating a change to trigger a print-parse _within_ the `UPDATE_TEXT` action handler. This was due to needing to have a fully parsed (more detailed) model from an arbitrary block that we create in the `UPDATE_TEXT` action. For example if `props.backgroundColor` is changed to `props.otherColors[bestColor]` then we need to be able to parse out the need for whatever `bestColor` is as it has been newly introduced. But the arbitrary expression we create for that does not refer to it and would cause a canvas error if we attempted to render it.

The parse-print cycle in the action above does have the downside of the editor bundle now referring to the parser-printer worker bundle.

Otherwise there's a couple of small bug fixes from the previous PR and some smaller improvements like those needed for the navigator.

**Commit Details:**
- Added fake spy entries for the 3 new types and incorporated them into the existing text editing handling in `renderCoreElement`.
- `jsxElementChildToText` now takes a parameter of `outermost`, which helps to determine if the entire result should be wrapped in curly braces.
- Expanded handling of the new types into `jsxElementChildToText`, where previously they were not handled.
- Fixed a bug in `runJSExpression` where the `block` parameter was passed instead of the scope.
- The `UPDATE_TEXT` handler now has some additional typing to help avoid covariant function parameters which allow some bad values to be added into the model.
- `UPDATE_TEXT` action handler now also triggers a print-parse directly of the files that were changed so that we get a more detailed parsed result into the model.
- Added `isTextEditablePropertyAccessElementAccessOrIdentifier` utility function.
- Fixed some functions in `MetadataUtils` around if a value is text editable or not as they needed extra cases added for the 3 new types.
- Added support for the 3 new types to `getChildrenPaths` and `getReorderedPaths`.
- Added a few typeguards for the new types.
- Reworked `getDefinedElsewhereFromAttribute`.
- Added `jsxConditionalExpressionConditionOptic`.
- Fixed a bug in `jsxAttributeToValue`, changing the order of deference between the two scopes used.
- Broke apart `parseUID` function into two smaller functions.
- Added `getUIDFromCommentsOrValue`, `createJSElementAccess`, `createJSPropertyAccess`, `createJSIdentifier`, `parseIdentifier`, `parseElementAccessExpression`, `parsePropertyAccessExpression` and `merge2WithParseMetadata` functions to the parsing logic.
- `parseAttributeExpression` now will parse out the new types.
- Refactored `mergeHighlightBounds` so that it permits any number of bounds to combine.
- `parseOutJSXElements` updated to parse out the new types.
- Removed dependency cruiser rule.


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1840,6 +1844,16 @@ function getValidElementPathsFromElement(
...
+    return paths
+  } else if (
+    isJSExpression(element) &&
+    (isJSIdentifier(element) || isJSPropertyAccess(element) || isJSElementAccess(element))
+  ) {
+    const uid = getUtopiaID(element)
+    const path = parentIsInstance
+      ? EP.appendNewElementPath(parentPath, uid)
+      : EP.appendToPath(parentPath, uid)
+    let paths = [path]
```
gbalint, line 1856: This just looks like a `return [path]`

editor/src/core/model/conditionals.ts
```diff
@@ -16,7 +17,7 @@ import { getUtopiaID } from '../shared/uid-utils'
 import type { Optic } from '../shared/optics/optics'
 import { fromField, fromTypeGuard } from '../shared/optics/optic-creators'
 import { findUtopiaCommentFlag, isUtopiaCommentFlagConditional } from '../shared/comment-flags'
-import { isLeft, isRight } from '../shared/either'
+import { foldEither, isLeft, isRight } from '../shared/either'
```
gbalint, line 20: Added imports, but no code changes.

gbalint (approved):

bkrmendy (approved):

editor/src/components/canvas/canvas-utils.ts
seanparsons: Fixed in 0590320a3.

editor/src/core/model/conditionals.ts
seanparsons: Fixed in 0590320a3.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -634,7 +648,15 @@ function jsxElementChildToText(
   prevElement: JSXElementChild | null,
   nextElement: JSXElementChild | null,
   expressionContext: 'jsx' | 'javascript',
+  outermost: 'outermost' | 'inner',
 ): string {
+  function outermostWrapInParens(value: string): string {
```
Rheeseyb, line 653: `outermostWrapInBraces`

seanparsons: Fixed in 80d167006.

editor/src/components/inspector/sections/component-section/component-section.spec.browser2.tsx
```diff
@@ -645,12 +650,26 @@ describe('Set element prop via the data picker', () => {
...
+          }
+        } else {
+          throw new Error(`Expected JSIdentifier for onValue but got: ${element.onValue}`)
+        }
+      } else {
+        throw new Error(`Expected JSElementAccess but got: ${element}`)
+      }
+    } else {
+      throw new Error(`Unexpected value: ${possibleElement.value}`)
+    }
```
balazsbajorics, line 672: me sobs about happy path to the left

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1096,6 +1102,21 @@ export const MetadataUtils = {
...
+      return true
+    } else if (isJSPropertyAccess(element)) {
+      return MetadataUtils.isTextEditablePropertyAccessElementAccessOrIdentifier(element.onValue)
+    } else if (isJSElementAccess(element)) {
+      return (
+        MetadataUtils.isTextEditablePropertyAccessElementAccessOrIdentifier(element.onValue) &&
+        MetadataUtils.isTextEditablePropertyAccessElementAccessOrIdentifier(element.element)
+      )
+    } else {
+      return false
```
balazsbajorics, line 1117: shouldn't we also return true for Arbitrary JS blocks too? othwerwise the following will not be text editable:
`<p>{element.["arbitrary" + "Js"]}</p>`

seanparsons: I tried to be as conservative as possible initially, I _think_ that should be text editable however. Lets expand it in a later PR along with some tests to validate that.

PR #4970, opened 2024-02-28
Remix projects pending operations

### LINKED ISSUE #4969: remix projects pending operations

ruggi:
Fix #4969

**Problem:**

In the Remix projects it's impossible to know whether a pending operation is completed or not, and fetchers gets canceled when unmounted, not triggering data reloads.

**Fix:**

1. Use a specialized version of fetchers that adds an ongoing operations for the project actions
2. Display the ongoing actions stacked in the bottom right corner of the page
3. When the fetchers complete, remove the pending operations
4. Turn on the future `v3_fetcherPersist` feature of Remix (https://remix.run/docs/en/main/file-conventions/remix-config#future) that allows the fetchers to keep going after being unmounted _and_ being available through the `useFetchers` hook
5. Refactor the `ProjectsStore` object so it's composed of persistable and non-persistable fields, so we don't e.g. persist pending operations (and other transient things like the selected category)

The video below should give a good example of all of the above in action:

https://github.com/concrete-utopia/utopia/assets/1081051/87c042a5-dc7f-4102-b58d-e4b82221838f

(Delay added for dramatic purposes!)

### DIFF
--- a/utopia-remix/app/routes/projects.tsx
+++ b/utopia-remix/app/routes/projects.tsx
@@ -826,9 +826,6 @@ ProjectCardActions.displayName = 'ProjectCardActions'
 
 const ActiveOperations = React.memo(() => {
   const operations = useProjectsStore((store) => store.operations)
-  if (operations == null) {
-    return null
-  }
 
   function getOperationVerb(op: Operation) {
     switch (op.type) {
--- a/utopia-remix/app/routes/projects.tsx
+++ b/utopia-remix/app/routes/projects.tsx
@@ -3,20 +3,22 @@ import {
   Trigger as DropdownMenuTrigger,
 } from '@radix-ui/react-dropdown-menu'
 import {
+  CubeIcon,
   DashboardIcon,
   DotsHorizontalIcon,
   HamburgerMenuIcon,
   MagnifyingGlassIcon,
   TrashIcon,
-  CubeIcon,
 } from '@radix-ui/react-icons'
-import React from 'react'
 import { LoaderFunctionArgs, json } from '@remix-run/node'
 import { useFetcher, useLoaderData } from '@remix-run/react'
+import { motion } from 'framer-motion'
 import moment from 'moment'
 import { UserDetails } from 'prisma-client'
+import React from 'react'
 import { ProjectContextMenu } from '../components/projectActionContextMenu'
 import { SortingContextMenu } from '../components/sortProjectsContextMenu'
+import { useCleanupOperations } from '../hooks/useFetcherWithOperation'
 import { useIsDarkMode } from '../hooks/useIsDarkMode'
 import { listDeletedProjects, listProjects } from '../models/project.server'
 import { getCollaborators } from '../models/projectCollaborators.server'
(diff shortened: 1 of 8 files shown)


utopia-remix/app/routes/projects.tsx
```diff
@@ -819,3 +823,74 @@ const ProjectCardActions = React.memo(({ project }: { project: ProjectWithoutCon
...
+        margin: 10,
+        display: 'flex',
+        flexDirection: 'column',
+        gap: 10,
+      }}
+    >
+      {operations
+        .sort((a, b) => -(a.startedAt - b.startedAt))
+        .map((op) => {
+          return (
```
gbalint, line 860: I think it would be nice to create a separate component for a single operation.

bkrmendy (approved):

utopia-remix/app/routes/projects.tsx
ruggi: done in [`db64baf` (#4970)](https://github.com/concrete-utopia/utopia/pull/4970/commits/db64baf71e96b24f0261c0feed2f1f4dfdaff7f0)

Rheeseyb (approved):
Nice!

utopia-remix/app/routes/projects.tsx
```diff
@@ -819,3 +823,78 @@ const ProjectCardActions = React.memo(({ project }: { project: ProjectWithoutCon
...
+        bottom: 0,
+        right: 0,
+        margin: 10,
+        display: 'flex',
+        flexDirection: 'column',
+        gap: 10,
+      }}
+    >
+      {operations
+        .sort((a, b) => -(a.startedAt - b.startedAt))
```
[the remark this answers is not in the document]
seanparsons: Also `Array.sort` sorts in place, is that OK for this case?

utopia-remix/app/store.tsx
```diff
@@ -1,44 +1,115 @@
...
+            operations: operations.filter((other) => other.key !== key),
+          }))
+        },
       }),
       {
         name: 'store',
+        partialize: (fullStore) => {
+          const nonPersistedKeys = Object.keys(
+            initialProjectsStoreStateNonPersisted,
+          ) as (keyof ProjectsStoreStateNonPersisted)[]
```
seanparsons, line 92: I suspect it's better to make this a type ascription, `as` will just tell TypeScript that it's definitely that type and if there's any divergence you'll have a bad time.

utopia-remix/app/types.ts
```diff
@@ -31,3 +31,23 @@ export function userToCollaborator(user: UserDetails): Collaborator {
...
+export interface Operation {
+  projectId: string
+  projectName: string
+  type: OperationType
+}
+
+export type OperationType = 'rename' | 'delete' | 'destroy' | 'restore'
+
+export function operationsEqual(a: Operation, b: Operation): boolean {
+  return a.projectId === b.projectId && a.type === b.type
```
seanparsons, line 44: Should this include `projectName`? If not I'd be tempted to change what this function is called so it doesn't give people a false impression of what it checks.

seanparsons (approved):

utopia-remix/app/types.ts
ruggi: good point!

ruggi: [`7962936` (#4970)](https://github.com/concrete-utopia/utopia/pull/4970/commits/79629367287c8ba48ffa9ae21acac55b827c53f9)

utopia-remix/app/routes/projects.tsx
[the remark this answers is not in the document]
ruggi: [`676c512` (#4970)](https://github.com/concrete-utopia/utopia/pull/4970/commits/676c5121d2669114b05b9d721c07be1119e43dc6)

utopia-remix/app/store.tsx
ruggi: I was having a hard time with the TS compiler around this one, do you have a direct example? (I thought this was fine since the keys of that object are definitely the keys of that type, but that's the best I got)

seanparsons: I _believe_ this should work and will error if the types don't line up:
```typescript
const nonPersistedKeys: Array<keyof ProjectsStoreStateNonPersisted> = ...
```

ruggi: ah yeah, I tried it but…
```
Type 'string[]' is not assignable to type '(keyof ProjectsStoreStateNonPersisted)[]'.
  Type 'string' is not assignable to type 'keyof ProjectsStoreStateNonPersisted'.ts
```

PR #4971, opened 2024-02-28
Fix deep equality checks for new expression types.

seanparsons:
**Problem:**
In recent work, new types were added to `JSExpression`, but the keep deep equality calls were not updated to cater for this.

**Fix:**
The extra cases were added to the two deep equality calls and one of the calls was renamed to line it up with the type.

**Commit Details:**
- Renamed `JSXAttributeKeepDeepEqualityCall` to `JSExpressionKeepDeepEqualityCall`.
- Added `JSIdentifierKeepDeepEquality`, `JSPropertyAccessKeepDeepEquality` and `JSElementAccessKeepDeepEquality`.
- Added handling for the new expression types to `JSXElementChildKeepDeepEquality`.
- Added handling for the new expression types to `JSExpressionKeepDeepEqualityCall`.

### DIFF
--- a/editor/src/components/editor/store/store-deep-equality-instances.ts
+++ b/editor/src/components/editor/store/store-deep-equality-instances.ts
@@ -1100,37 +1100,57 @@ export const JSExpressionKeepDeepEqualityCall: KeepDeepEqualityCall<JSExpression
   oldAttribute,
   newAttribute,
 ) => {
-  if (isJSXAttributeValue(oldAttribute) && isJSXAttributeValue(newAttribute)) {
-    return JSXAttributeValueKeepDeepEqualityCall(oldAttribute, newAttribute)
-  } else if (
-    modifiableAttributeIsAttributeOtherJavaScript(oldAttribute) &&
-    modifiableAttributeIsAttributeOtherJavaScript(newAttribute)
-  ) {
-    return JSExpressionOtherJavaScriptKeepDeepEqualityCall()(oldAttribute, newAttribute)
-  } else if (
-    modifiableAttributeIsAttributeNestedArray(oldAttribute) &&
-    modifiableAttributeIsAttributeNestedArray(newAttribute)
-  ) {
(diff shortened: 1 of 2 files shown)


editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -1215,28 +1274,25 @@ export function JSXElementChildKeepDeepEquality(): KeepDeepEqualityCall<JSXEleme
...
+    } else if (isJSExpressionFunctionCall(oldElement) && isJSExpressionFunctionCall(newElement)) {
       return JSXAttributeFunctionCallKeepDeepEqualityCall()(oldElement, newElement)
+    } else if (isJSIdentifier(oldElement) && isJSIdentifier(newElement)) {
+      return JSIdentifierKeepDeepEquality()(oldElement, newElement)
+    } else if (isJSPropertyAccess(oldElement) && isJSPropertyAccess(newElement)) {
+      return JSPropertyAccessKeepDeepEquality()(oldElement, newElement)
+    } else if (isJSElementAccess(oldElement) && isJSElementAccess(newElement)) {
+      return JSElementAccessKeepDeepEquality()(oldElement, newElement)
     } else {
       return keepDeepEqualityResult(newElement, false)
```
balazsbajorics, line 1297: shouldn't we instead of this catch-all put an exhaustive check here so in the future we can't forget adding new deep equality results?

balazsbajorics (approved):
check my comment about the exhaustive check!

editor/src/components/editor/store/store-deep-equality-instances.ts
seanparsons: Yeah, that would be better.

seanparsons: Fixed in 857d3890a.

gbalint (approved):

PR #4978, opened 2024-02-29
Fix issue with Prisma binaries being incompatible.

seanparsons:
**Problem:**
Prisma has some binary components and they need to match versions with the node libraries that use them or it'll just explode with a mystery error when it tries to use an incompatible API.

**Cause:**
The current version of nixpkgs that we're tracking is 22.11 which has the 4.xx version of the `query-engines` package, but our Remix app is using 5.xx.

**Fix:**
The 5.xx version of `query-engines` is available in the 23.11 nixpkgs so this PR adds that as an additional reference and uses it for the environment variables. The environment variables themselves were also updated as they have changed between 4.xx and 5.xx.

**Commit Details:**
- Added new nixpkgs version to `release.nix`.
- Updated the Prisma environment variables for version 5.xx.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -12,7 +12,15 @@ let
     sha256 = "11w3wn2yjhaa5pv20gbfbirvjq6i3m7pqrq2msf0g7cv44vijwgw";
   }) { inherit config; };
 
+  recentPkgs = import (builtins.fetchTarball {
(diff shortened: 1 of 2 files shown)


ruggi (approved):
glory glory

liady (approved):
Nice!

PR #4979, opened 2024-02-29
Remix projects: optimistic UI

### LINKED ISSUE #4977: Remix projects optimistic UI

ruggi:
Fix #4977

This is a follow up to https://github.com/concrete-utopia/utopia/pull/4970: now that we have pending operations, this PR adds optimistic patterns to the project operations (rename, delete, destroy, and restore).

In a nutshell: when an operation is ongoing, hide or update the UI counterparts immediately, reverting back if it fails.

https://github.com/concrete-utopia/utopia/assets/1081051/64080bdd-abbd-43bd-afb8-bf440a247a0a

**Note**
The bulk of this work is in the updated `useCleanupOperations` hook, which will take care of updating or removing pending operations based on the fetcher state changes.

There are two `useEffect` hooks there: the first listens to the fetcher state changes (**no other way around it**) coming from the network and keeps a list of fetchers that completed (meaning they transitioned from `loading` to `idle` while having data, and then were removed from the `useFetchers` response). The second one performs the actual processing of the operations on a list of idle fetchers that was stored in a `useState` hook to avoid infinite dependency loops. To the best of my ability (and my **strong, committed will** to not use `useEffect` hooks) this was the best (and possibly the only possible?) approach to solve this, but it'd be very good to have an extra pair of eyes on it.


utopia-remix/app/components/projectActionContextMenu.tsx
```diff
@@ -16,16 +24,20 @@ type ContextMenuEntry =
...
+  const restoreFetcher = useFetcherWithOperation(project.proj_id, 'restore')
+  const renameFetcher = useFetcherWithOperation(project.proj_id, 'rename')
 
   const selectedCategory = useProjectsStore((store) => store.selectedCategory)
 
   const deleteProject = React.useCallback(
     (projectId: string) => {
-      deleteFetcher.submit({}, { method: 'POST', action: `/internal/projects/${projectId}/delete` })
+      deleteFetcher.submit(
+        operationDelete(project),
```
Rheeseyb, line 37: Since we're creating a fetcher specifically for each type of operation, it's probably worth trying to create those in a way that means we don't have to supply the operation itself (i.e. we know this is a deletion only fetcher, so we should be able to supply it with the type of operation, method, and action at the time of creating it, and just tell it which project to delete when using it)

Rheeseyb: Actually, saying that, I've just realised that we're only supplying the operation type to use in the string `key`, so maybe just ignore the above and leave it as is :)

ruggi: yeah I was about to reply: I had to play around with this a bit last night - the main problem being that the operation should also contain the information it's receiving _when_ submitted, which is unfortunately no available before that (and can obviously change every time :( )

utopia-remix/app/types.ts
```diff
@@ -32,22 +33,74 @@ export function userToCollaborator(user: UserDetails): Collaborator {
...
   return a.projectId === b.projectId && a.type === b.type
 }
 
-export function operation(project: ProjectWithoutContent, type: OperationType): Operation {
-  return {
-    projectId: project.proj_id,
-    projectName: project.title,
-    type: type,
+export function getOperationVerb(op: Operation) {
+  switch (op.type) {
```
liady, line 94: minor - but maybe it will be better for the operation caller to specify the full message text, instead of inferring it like this (if we'll have more complex texts this can easily break)

utopia-remix/app/util/errors.ts
```diff
@@ -22,3 +22,8 @@ export function isErrorWithStatus(u: unknown): u is ErrorWithStatus {
   const maybe = u as ErrorWithStatus
   return maybe.status != null && maybe.statusText != null && maybe.data != null
 }
+
+export function isLikeApiError(u: unknown): u is ApiError {
+  const maybe = u as ApiError
+  return maybe.status != null && maybe.message != null
```
Rheeseyb, line 28: We should be checking that this is actually not null and an object before checking the fields, like we do here <https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/canvas/canvas-component-entry.tsx#L154>

ruggi: [`d199a17` (#4979)](https://github.com/concrete-utopia/utopia/pull/4979/commits/d199a173ecaf1db12f9c92177c73081d795fad8d)

ruggi: also done it for another helper

utopia-remix/app/routes/projects.tsx
```diff
@@ -842,59 +890,71 @@ const ActiveOperations = React.memo(() => {
...
-      <div>
-        {getOperationVerb(operation)} project {operation.projectName}
+        className={sprinkles({
+          boxShadow: 'shadow',
+          borderRadius: 'small',
+          backgroundColor: operation.errored ? 'error' : 'primary',
+          color: 'white',
+        })}
+      >
+        {when(!operation.errored, <Spinner className={sprinkles({ backgroundColor: 'white' })} />)}
```
seanparsons, line 931:
```suggestion
        {unless(operation.errored, <Spinner className={sprinkles({ backgroundColor: 'white' })} />)}
```

seanparsons (approved):

utopia-remix/app/routes/projects.tsx
ruggi: [`730f55f` (#4979)](https://github.com/concrete-utopia/utopia/pull/4979/commits/730f55f00a11a6bc7af7f914b773022f52bac8d8)

balazsbajorics (approved):

PR #4986, opened 2024-03-04
Nested render prop navigator

gbalint:
image
image

## Description

This PR adds navigator support for render props.  With these changes, the navigator renders labels for render props, and renders the JSX hierarchy in each render prop. 

### Out of scope
This PR only implement displaying render props and their contents in the navigator, editability/interactions are out of scope.

### Details
- a feature flag is added for this feature (`Render Props in Navigator`, set to true by default)
- `getNavigatorTargets` adds the JSX elements in props to the navigator targets
- `NavigatorItem` detects and renders the synthetic navigator entries for render props
- the `childOrAttribute` prop is removed from `RenderPropNavigatorItemContainerProps`


editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -737,6 +737,8 @@ export const NavigatorItem: React.FunctionComponent<
 
   const isComponentScene = useIsProbablyScene(navigatorEntry) && childComponentCount === 1
   const isRenderProp = isRenderPropNavigatorEntry(navigatorEntry)
+  const isRenderPropSlot =
+    isSyntheticNavigatorEntry(navigatorEntry) && navigatorEntry.renderProp != null
```
balazsbajorics, line 741: is the synthetic-ness here the thing that makes this a slot? because if it would be filled, we would have a non-synthetic entry for the render prop?

editor/src/components/navigator/navigator-utils.ts
```diff
@@ -119,19 +124,47 @@ export function getNavigatorTargets(
...
+            const entries = [
+              renderPropNavigatorEntry(fakePath, prop),
+              syntheticNavigatorEntry(
+                EP.appendToPath(path, `prop-slot-${prop}`),
+                propValue ?? jsExpressionOtherJavaScriptSimple('null', []),
+                propValue == null ? prop : null,
+              ),
+            ]
+            navigatorTargets.push(...entries)
+            visibleNavigatorTargets.push(...entries)
```
balazsbajorics, line 147: is it possible to do an early return here?

editor/src/components/navigator/navigator-utils.ts
```diff
@@ -256,7 +289,20 @@ export function getNavigatorTargets(
           }
         }
       } else {
-        fastForEach(Object.values(subTree.children), (child) => {
+        const notProcessedChildren = Object.values(subTree.children).filter(
```
balazsbajorics, line 292: maybe something like notRenderPropChildren or normalChildren better describes that the interesting thing about these children isn't that they are unprocessed (raw) but rather the fact that we didn't deem them to be render props
```suggestion
        const notRenderPropChildren = Object.values(subTree.children).filter(
```

editor/src/components/navigator/navigator-item/navigator-item.tsx
bkrmendy: "Synthetic" should have been called "Label" or something similar, it means that it's not a navigator entry for something actually in the project, but just something that was "synthesised". The navigator entry for the (empty) slot is independent from the corresponding synthetic navigator entry (the label)

editor/src/components/navigator/navigator-utils.ts
```diff
@@ -256,7 +289,20 @@ export function getNavigatorTargets(
...
+          (c) => !processedPathsAsRenderProp.has(c.pathString),
+        )
+        if (
+          propertyControls != null &&
+          Object.keys(propertyControls).length > 0 &&
+          notProcessedChildren.length > 0 &&
+          renderPropFound // only show a dedicated label for the children prop if the component has render props too
+        ) {
+          const entry = renderPropNavigatorEntry(EP.appendToPath(path, 'children'), 'children')
```
balazsbajorics, line 302: I don't think it's worth fixing now, but I think there is something lopsided in the idea that we decide to show / hide a label here in the navigator entries generation. I'd rather we always have the label and then hide it conditionally. but I guess that would make drag and drop less straightforward

editor/src/components/navigator/navigator-utils.ts
bkrmendy: https://github.com/concrete-utopia/utopia/pull/4986/commits/ec94818460bc276c593b2a5822300a55cf3aa0c2

editor/src/components/navigator/navigator-utils.ts
bkrmendy: adding a label for children is very noisy visually, I think it's better to add a label for the children prop only if there's a render prop right above in the navigator

image

balazsbajorics: > I'd rather we always have the label and then hide it conditionally
so what I meant is to add the label in the navigator entries, and then hide it via css if there is no render prop

editor/src/components/navigator/navigator-item/navigator-item.tsx
balazsbajorics: > The navigator entry for the (empty) slot is independent from the corresponding synthetic navigator entry (the label)

but the check is literally isSyntheticNavigatorEntry && navigatorEntry.renderProp, so I'm still confused then.

editor/src/components/navigator/navigator-utils.ts
bkrmendy: I tried that too, the reason it doesn't work is that because the `AutoSizer` in the Navigator component needs to know the size of each item beforehand, hiding the child label via css would leave a gap in the place of the child label:

image


(of course if you have a specific solution in mind, feel free to share it)

balazsbajorics: nah, I think we are good with the current one, just wanted to write that if _you_ have an idea of how to separate these two concerns, I'd prefer that. but it's good as is

balazsbajorics (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -822,12 +818,12 @@ export const NavigatorItem: React.FunctionComponent<
             style={{
               width: 140,
               height: 19,
-              borderRadius: 20,
-              backgroundColor: colorTheme.dynamicBlue10.value,
-              color: colorTheme.navigatorResizeHintBorder.value,
+              fontWeight: 600,
+              textTransform: 'uppercase',
+              color: 'gray',
```
Rheeseyb, line 823: We should definitely be using the `colorTheme` here

seanparsons (approved):

PR #4989, opened 2024-03-04
Restore errors for the new expression types.

seanparsons:
**Problem:**
One side effect of the new expression types is that if the expression produce an error (like if `element` in `element.property` is undefined) then only the resultant exception message (and potentially a wrong one) is presented to the user without the indicator showing where in the code it came from.

**Fix:**
When the expression throws an exception, it is re-run as an arbitrary block which with the associated source map produces the correct code highlight for the error location.

**Commit Details:**
- Added `originalJavascript` to `JSPropertyAccess` and `JSElementAccess`.
- Added `sourceMap` to `JSIdentifier`, `JSPropertyAccess` and `JSElementAccess`.
- Expanded `jsxAttributeToValue` to defer to `jsExpressionOtherJavaScript`
  to get proper stack traces if the expression as a whole fails.


bkrmendy (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.spec.browser2.tsx
```diff
@@ -94,8 +94,8 @@ export const App2 = (props) => {
         width: 350,
         height: 400,
       }}
-      data-uid='aaa'
-      data-testid='aaa'
+      data-uid='div'
```
balazsbajorics, line 97: why this seemingly unrelated change? :)

balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.spec.browser2.tsx
seanparsons: Ah yes, I did mean to write something about this but forgot in the end.

In functions like `createJSElementAccess` we create a version of the value without a UID or source map so that if it lacks a `data-uid` property we can hash the value to produce the UID. However, the fields like `JSElementAccess.onValue` were being passed through with their UIDs and source maps intact. As a result for each case of `props.style` for example, when it hashed the `onValue` field it was getting a different value for each case of `props` (in a `JSIdentifier`) as those had their UIDs and source maps which were included in the hashing. Now this means that each instance of `props.style` wherever it is should produce the same hash value and when there are clashes with existing values we start at `aaa` and rotate by character (to `aab` and so on) from there.

I changed some of these tests to values which are unlikely to be produced in this way, after finding some ending up like `aac`, or similar, as a result of these changes.

PR #4990, opened 2024-03-04
Parser support for render props

bkrmendy:
## Problem

In order to support editing operations on render props in the inspector, and making it possible to show render props in the navigator, we need to parse render props as JSX elements

## Fix
Parse them as JSX elements.

### Details
- The parser is changed so that if it detect a prop that's a JSX element, it parses it as such. The corresponding change is made in the printer too.
- the PR extends `JSExpression` with `JSXElement`. We'll need to support fragments and conditionals too in the future, this PR takes an incremental step toward that goal.
- Because of the change above, `jsxAttributeToValue` calls `renderCoreElement`. To make that possible, `jsxAttributeToValue` is passed all the arguments `renderCoreElement` needs
- The change above caused a circular dependency problem, so some functions have been shuffled around to work around that. `text-editor-utils.tsx` and `jsx-attribute-utils.tsx` were created to break to circular dependency chains


editor/src/core/model/element-template-utils.ts
```diff
@@ -1275,7 +1284,12 @@ export function elementChildSupportsChildrenAlsoText(
   metadata: ElementInstanceMetadataMap,
   elementPathTree: ElementPathTrees,
 ): ElementSupportsChildren | null {
-  if (isJSExpression(element)) {
+  if (
+    isJSExpression(element) &&
+    // the condition below is a kludge to make the reparent tests pass, discuss
```
seanparsons, line 1289: Still needs a discussion?

seanparsons (approved):

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
```diff
@@ -47297,10 +47297,7 @@ exports[`UiJsxCanvas render renders correctly when a component is passed in via
           style=\\"background-color: pink; width: 100%; height: 20px\\"
           data-path=\\"eee/fff/app-entity:aaa/bbb:ddd\\"
         >
-          <div
-            data-uid=\\"ccc~~~1\\"
-            data-path=\\"eee/fff/app-entity:aaa/bbb/ccc~~~1\\"
-          >
+          <div data-uid=\\"ccc\\" data-path=\\"eee/fff/app-entity:aaa/bbb/ccc\\">
```
balazsbajorics, line 47300: is this change desired here?

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -751,10 +762,12 @@ export function utopiaCanvasJSXLookup(
...
 
 function runJSExpression(
-  filePath: string,
-  requireResult: MapLike<any>,
   block: JSExpression,
+  elementPath: ElementPath | null,
   currentScope: MapLike<any>,
+  renderContext: RenderContext,
+  uid: string | undefined,
+  codeError: Error | null,
```
balazsbajorics, line 770: I would leave a comment here that this should probably deleted in a further refactor

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -2366,6 +2366,39 @@ function getAttributeExpression(
...
+      if (
+        parseResult.value.value.length !== 1 ||
+        parseResult.value.value[0].value.type !== 'JSX_ELEMENT'
+      ) {
+        return left('Cannot parse jsx element')
+      }
+
+      return right(
+        withParserMetadata(
+          parseResult.value.value[0].value,
```
balazsbajorics, line 2396: shouldn't we assign `parseResult.value.value` to a named const for readability?

editor/src/core/workers/parser-printer/uid-fix.ts
```diff
@@ -71,6 +71,8 @@ const expressionFunctionCallUIDOptic: Optic<JSExpressionFunctionCall, string> =
 const expressionOtherJavaScriptUIDOptic: Optic<JSExpressionOtherJavaScript, string> =
   fromField('uid')
 
+const expressionJSXElementOptic: Optic<JSXElement, string> = fromField('uid')
```
balazsbajorics, line 74: I think this is missing a "uid" from the name 
```suggestion
const expressionJSXElementUIDOptic: Optic<JSXElement, string> = fromField('uid')
```

editor/src/components/canvas/__snapshots__/ui-jsx-canvas.spec.tsx.snap
gbalint: Yes. Before this jsx render props were wrapped into an expression (which just returned the element), which made them a generated element.
After this change these elements will not be considered generated (which is a prerequisite to make them editable).

editor/src/utils/value-parser-utils.ts
```diff
@@ -325,6 +325,15 @@ export function parseJsx(_: unknown, value: unknown): ParseResult<JSXParsedValue
         name: type,
       })
     }
+    if (value.hasOwnProperty('props')) {
```
balazsbajorics, line 328: is this a somewhat unrelated change?

balazsbajorics (approved):
looks very nice :)

editor/src/core/workers/parser-printer/uid-fix.ts
gbalint: I removed it, it was unused.

editor/src/utils/value-parser-utils.ts
gbalint: Yes, this was just a small issue we found during the more extensive jsx prop testing.
However, after this PR we may rethink on only relying on the runtime value and not on the parsed model in `parseJsx`. But that can come in a subsequent PR.

PR #4997, opened 2024-03-06
Chore/hydrogen test case

balazsbajorics:
This PR adds a first experimental test for a real-life Remix Hydrogen mock-shop project.
The project itself is loaded from the test editor's memory, however once the project is loaded, it has userland code that starts making queries to the mock.shop backend to load content. Which means the test requires an online connection to work! Hopefully it will not be too brittle in real life. I tried to write the code defensively so it always awaits and retries until things are loaded.

**Commit Details:**
- Added the entire PersistentModel of the hydrogen-november test project as a mega javascript object in a `.test-utils.ts` file so it's banned from being accidentally pulled into the production editor
- Added a mouse click helper function called `mouseClickAtElementCenter` which throws an error if the element you want to click on would not be clickable by a real mouse (by checking if its center is obstructed)
- Added a first test `can navigate in play mode to a detail page` which waits for the Mock Shop to load and then clicks on Unisex -> Hoodie -> Large and awaits the large button becoming selected.


gbalint (approved):

bkrmendy (approved):

Rheeseyb (commented):
🥳

PR #5003, opened 2024-03-06
Unify parsing of arbitrary js

Rheeseyb:
**Problem:**
We had 2 very slightly different functions for parsing arbitrary JS, one of which wasn't injecting the data-uids generated (during the parse) into the `javascript` field of the parsed object. This means that in those cases we aren't able to apply changes to the parsed `elementsWithin` when printing the javascript back out.

**Fix:**
Merge the two together by renaming `parseJSExpression` to `parseAttributeOtherJavaScript` (as `parseJSExpression` was the one that included the `data-uid` injection), and removing the previous version of `parseAttributeOtherJavaScript`.

This alone wasn't enough as that caused issues with the source maps, because it turns out that `parseJSExpression` was just binning off the original source maps. To fix that I then had to update `insertDataUIDsIntoCode` to take in the original source map, so that it could update that.

I also had to include a fix from https://github.com/concrete-utopia/utopia/pull/4696 into `insertDataUIDsIntoCode` to attempt wrapping the code in an anonymous function before transpiling if wrapping it in parens wasn't enough (as without that change that function is unable to handle top level `if` statements).

**Note:** 
I need to also pull across changes to the printing flow from here <https://github.com/concrete-utopia/utopia/pull/4975/files#diff-f87f630e16e8aebfd1166b663c79da12ee5d7104d93e2d0ba74e3b2851ea6ecd> so that we actually are streaming in the changes to the `elementsWithin` when printing, but that's a much simpler change. I just wanted to separate that change from this painful PR.


seanparsons (approved):
Yay!

editor/src/components/canvas/ui-jsx-canvas-errors.spec.tsx
```diff
@@ -496,29 +496,44 @@ export var ${BakedInStoryboardVariableName} = (props) => {
...
           "stackFrames": Array [
             Object {
-              "columnNumber": 13,
+              "columnNumber": 20,
               "fileName": "code.tsx",
-              "lineNumber": 1,
+              "lineNumber": 8,
               "originalCode": Array [
                 ScriptLine {
-                  "content": "",
```
balazsbajorics, line 504: oh nice, you fixed a buggy error message!

bkrmendy (approved):

ruggi (approved):

editor/src/core/workers/parser-printer/__snapshots__/parser-printer.spec.ts.snap
```diff
@@ -909,17 +907,16 @@ Object {
             "n",
           ],
           "elementsWithin": Object {},
-          "javascript": "n > 0",
+          "javascript": "n > 0;",
```
gbalint, line 910: Why are all these extra line ending semicolons necessary?

balazsbajorics (approved):

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
```diff
@@ -1443,19 +1352,18 @@ function getCommentsOnExpression(sourceText: string, expression: TS.Node): Parse
   }
 }
 
-// FIXME It looks like this should now be merged with parseAttributeOtherJavaScript
-function parseJSExpression(
+export function parseAttributeOtherJavaScript(
```
gbalint, line 1355: I think we should call this `parseJSExpressionMapOrOtherJavascript`, attributeOtherJavascript is kind of a legacy name which is misleading (because it is not just for attributes)

editor/src/core/workers/parser-printer/__snapshots__/parser-printer.spec.ts.snap
Rheeseyb: This is because the javascript has now gone through a babel transpile (whereas before it was just the same as the `originalJavascript`). The `javascript` field should be renamed to `javascriptWithUIDs`, as the naming here is terrible, but I'll do that in a follow up PR

editor/src/core/workers/parser-printer/parser-printer-parsing.ts
Rheeseyb: Yeah good suggestion. I'll do that in another PR to save this one from becoming too noisy

PR #5011, opened 2024-03-08
Use YAML files to store our auth data, local dev

liady:
WIP - This PR adds `.yml` files in order to store our model data and test it. The purpose of it is to have our CI update the authorization model data, instead of changing it directly in our Okta dashboard.
This PR also supports local FGA dev server.

Test docs here: https://docs.fga.dev/modeling/testing
CLI docs here: https://github.com/openfga/cli?tab=readme-ov-file#write-authorization-model


ruggi:
closed in favor of https://github.com/concrete-utopia/utopia/pull/5059 which built on / extended this

PR #5012, opened 2024-03-08
adding karma-time-stats-reporter

balazsbajorics:
Shelving this because as a side effect it prints all test logs. I think we'll need to patch / fork the plugin to make it silent.


balazsbajorics:
Shelving this because as a side effect it prints all test logs. I think we'll need to patch / fork the plugin to make it silent.

PR #5013, opened 2024-03-08
Highlight selected data picker expression.

seanparsons:
**Problem:**
It's helpful to be able to see what expression is currently being used in the data picker.

**Fix:**
The entire hierarchy of the expression (should any apply) is italicised in the data picker if it matches.
image

For an expression we produce a simple path representation that is matched against the same path representation produced for the variables listed in the data picker.

**Commit Details:**
- `useDataPickerButton` takes two more properties so that `useComponentPropsInspectorInfo` can be called within it.
- Added `valuePath` field to `PrimitiveOption`, `ArrayOption`, `ObjectOption` and `JSXOption`.
- `DataPickerPopup` now takes an additional `propExpressionPath` property.
- `ValueRow` takes a `currentPropExpressionPath` to represent what is currently being used for a property if an appropriate expression is being used.
- Span in `ValueRow` gets italicised if the current expression matches.
- Slightly rejigged the fields of the `VariableInfo` types to make them a little clearer and get the appropriate values needed now.
- Added `isPrefixOf` utility function.
- Added `modifiableAttributeToValuePath` utility function.

### DIFF
--- a/editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
+++ b/editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
@@ -22,12 +22,14 @@ import type {
 } from './variables-in-scope-utils'
 import { useVariablesInScopeForSelectedElement } from './variables-in-scope-utils'
 import { assertNever } from '../../../../core/shared/utils'
+import { isPrefixOf } from '../../../../core/shared/array-utils'
 
 export interface PrimitiveOption {
   type: 'primitive'
   variableInfo: PrimitiveInfo
   definedElsewhere: string
   depth: number
+  valuePath: Array<string | number>
 }
 
 export interface ArrayOption {
@@ -36,6 +38,7 @@ export interface ArrayOption {
   depth: number
   definedElsewhere: string
   children: Array<VariableOption>
+  valuePath: Array<string | number>
 }
 
 export interface ObjectOption {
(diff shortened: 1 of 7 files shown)


editor/src/core/shared/element-template.ts
```diff
@@ -961,6 +960,32 @@ export function isRegularJSXAttribute(attribute: ModifiableAttribute): attribute
   )
 }
 
+export function modifiableAttributeToValuePath(
```
balazsbajorics, line 963: could you add some test coverage for this? I _think_ this function will be key for the data tracing features next week

seanparsons: Yup.

editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
```diff
@@ -272,7 +294,7 @@ function ValueRow({ variableOption, idx, onTweakProperty }: ValueRowProps) {
                   opacity: variableOption.variableInfo.matches ? 1 : 0.5,
                 }}
               >
-                {variableOption.variableInfo.displayName}
```
balazsbajorics, line 275: is this a related change?

bkrmendy (approved):

balazsbajorics (approved):

PR #5017, opened 2024-03-08
Radix themes

lankaukk:
**Problem:**

**Fix:**

- Did [other things]

Fixes #[ticket_number]  (<< pls delete this line if it's not relevant)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "./editor",
   "directories": {},
   "devDependencies": {
-    "husky": "^7.0.2"
+    "husky": "7.0.2"
   },
   "scripts": {
     "prepare": "husky install",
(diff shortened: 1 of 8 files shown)


bkrmendy (approved):

package.json
```diff
@@ -19,5 +19,9 @@
   "bugs": {
     "url": "https://github.com/concrete-utopia/utopia/issues"
   },
-  "homepage": "https://utopia.app"
+  "homepage": "https://utopia.app",
+  "dependencies": {
+    "@radix-ui/themes": "2.0.3",
```
balazsbajorics, line 24: I think these are misplaced, you probably accidentally ran `pnpm install @radix-ui/themes` in the root directory instead of the utopia-remix folder. we should revert this and the package-lock change, and re-run `pnpm install ...` inside `utopia-remix`.

I can help if you need it :)

balazsbajorics (approved):

PR #5018, opened 2024-03-11
Refer FGA to a local mock

liady:
Run FGA against a local mock to allow for offline/local work

### DIFF
--- a/utopia-remix/app/services/fgaService.server.ts
+++ b/utopia-remix/app/services/fgaService.server.ts
@@ -2,21 +2,29 @@ import { CredentialsMethod, OpenFgaClient } from '@openfga/sdk'
 import { ServerEnvironment } from '../env.server'
 import { AccessLevel } from '../types'
 
-export const fgaClient = new OpenFgaClient({
-  apiScheme: 'https',
-  apiHost: ServerEnvironment.FGA_API_HOST,
-  storeId: ServerEnvironment.FGA_STORE_ID,
-  // authorizationModelId: 'YOUR_MODEL_ID', // Optionally, you can specify a model id to target, which can improve latency
-  credentials: {
-    method: CredentialsMethod.ClientCredentials,
-    config: {
-      apiTokenIssuer: ServerEnvironment.FGA_API_TOKEN_ISSUER,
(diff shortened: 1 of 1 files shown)


bkrmendy (approved):

utopia-remix/app/services/fgaService.server.ts
```diff
@@ -2,21 +2,29 @@ import { CredentialsMethod, OpenFgaClient } from '@openfga/sdk'
...
-    },
-  },
-})
+const localFgaClient = {
+  write: async () => {},
+  check: async () => ({ allowed: true }),
+}
+
+export const fgaClient =
+  ServerEnvironment.environment === 'local'
```
ruggi, line 11: this should also include tests (by default?) I think it would be best to have this check it's not `prod | stage`

Rheeseyb: We should definitely be checking if the credentials have been provided, and only fall back to the `localFgaClient` if they haven't been.

Rheeseyb: +1 to Balazs's point too - it's much safer

balazsbajorics: actually _should we_ have no auth for tests? there's already multiple project access tests! those ought to break if we always allow project access

ruggi: tests currently don't break because they're not actually using the middleware

liady: @balazsbajorics I think that we need to mock the client only for local dev (not even tests, since they already mock a lower-level api)

liady: @Rheeseyb  We do need to put some sort of a more complex mechanism to allow someone with credentials to toggle whether they want to work against an auth server, or to have "always allow" auth (for whoever will be developing locally).
For now I used the server env to always default local dev to 'allow' just to release the blocker from everyone (for example if someone already changed their `.env` file).
I will have a follow-up PR later that will sort local auth dev as well

Rheeseyb (approved):

PR #5019, opened 2024-03-11
Don't animate comment indicator if ref is null

ruggi:
**Problem:**

When loading comment indicators, if the `avatarRef` is null the whole wrapper component breaks after trying to animate it.

**Fix:**

1. Do not run the animation for the avatar if the ref is null
2. Track whether the animation should play at all, meaning the indicator has been hovered at least once, so to avoid a loading animation glitch (the animation would go from 1 to 0 immediately after loading the component)

### DIFF
--- a/editor/src/components/canvas/controls/comment-indicator.tsx
+++ b/editor/src/components/canvas/controls/comment-indicator.tsx
@@ -1,7 +1,7 @@
 import type { ThreadData } from '@liveblocks/client'
-import { Comment } from '@liveblocks/react-comments'
 import type { CommentProps } from '@liveblocks/react-comments'
-import { AnimatePresence, motion, useAnimate } from 'framer-motion'
+import { Comment } from '@liveblocks/react-comments'
+import { AnimatePresence, motion } from 'framer-motion'
 import type { CSSProperties } from 'react'
 import React from 'react'
 import type { ThreadMetadata, UserMeta } from '../../../../liveblocks.config'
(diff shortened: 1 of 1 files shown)


bkrmendy (approved):

editor/src/components/canvas/controls/comment-indicator.tsx
```diff
@@ -538,24 +541,28 @@ function useHover() {
...
 const CommentIndicatorWrapper = React.memo((props: CommentIndicatorWrapper) => {
-  const { thread, expanded, user, forceDarkMode, ...commentProps } = props
+  const { thread, expanded, user, forceDarkMode, animatable, ...commentProps } = props
 
   const [avatarRef, animateAvatar] = useAnimate()
 
   const animDuration = 0.1
 
   React.useEffect(() => {
+    if (avatarRef.current == null || !animatable) {
```
balazsbajorics, line 563: when is this ref null? is this deterministic? is there a particular state / order of events that leads to a null ref?

ruggi: looking at the flow, it seems like it's when the component is mounted

ruggi: oh I agree too (and if you look at the code I always use the declarative approach), but there was definitely a reason for this `useAnimate`, which… I don't remember, and I looked for references in the code/past PRs… with no avail, so I just did that (and the bug is fixed too) [`954ad5a` (#5019)](https://github.com/concrete-utopia/utopia/pull/5019/commits/954ad5a892d6d69282a51a1016e7dd30bef3b86f)

happy monday folks!

Rheeseyb (approved):

editor/src/components/canvas/controls/comment-indicator.tsx
Rheeseyb: Win!

balazsbajorics: woohoooo!

balazsbajorics (approved):

PR #5024, opened 2024-03-12
Suport local FGA dev

liady:
This PR allows developing locally with FGA in 3 ways:
1. With a mock client - meaning that all requests are approved
2. Against the remote server
3. Against a local FGA server

This PR also introduces `safeFgaWrite` which mitigates an issue with current FGA API that throws a hard error upon writing on existing tuple or deleting a non-existing tuple (which we want to be idempotent)


utopia-remix/app/services/fgaClient.server.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+  return 'remote'
+}
+
+export async function safeFgaWrite(params: object) {
+  try {
+    await fgaClient.write(params)
+  } catch (err) {
+    if (((err as FgaApiValidationError).apiErrorCode = ErrorCode.WriteFailedDueToInvalidInput)) {
```
Rheeseyb, line 58: `===`!

balazsbajorics: how could this pass the tests? are we using the same linter config as for the editor?

Rheeseyb: Our disgusting way of running eslint has come back to haunt us!
The CI version only runs on the main editor: https://github.com/concrete-utopia/utopia/blob/master/editor/package.json#L18
And the commit hook doesn't include the new `utopia-remix` package: https://github.com/concrete-utopia/utopia/blob/master/editor/package.json#L59-L99

liady: yes, we saw that the remix config is not the same 🤦  it also doesn't warn on hooks dependencies array or let/const...

utopia-remix/.env.sample
```diff
@@ -8,14 +8,16 @@ AUTH0_ENDPOINT=""
 AUTH0_CLIENT_ID=""
 AUTH0_REDIRECT_URI=""
 
+GITHUB_OAUTH_CLIENT_ID=""
```
ruggi, line 11: I think this would need adjusting with a rebase

utopia-remix/app/services/fgaClient.server.ts
```diff
@@ -0,0 +1,65 @@
+import { CredentialsMethod, ErrorCode, FgaApiValidationError, OpenFgaClient } from '@openfga/sdk'
+import { ServerEnvironment } from '../env.server'
+
+const mockOpenFgaClient = {
+  write: async () => {},
+  check: async () => ({ allowed: true }),
+}
+
+export const fgaClient = createFgaClient()
```
ruggi, line 9: I think this would be a great place to reuse the `singleton` helper available in `singleton.server.ts`

utopia-remix/app/services/fgaClient.server.ts
```diff
@@ -0,0 +1,65 @@
...
+      return 'local_mock'
+    }
+    if (ServerEnvironment.FGA_API_HOST.includes('localhost')) {
+      return 'local'
+    }
+  }
+  return 'remote'
+}
+
+export async function safeFgaWrite(params: object) {
```
ruggi, line 54: I'm 100% sure my future self will forget to call this function and instead just call `fgaClient.write` 🙈 
What if instead we extend `OpenFgaClient` and override `write`?

```ts
class CoolOpenFgaClient extends OpenFgaClient {
  // ...

  write(...) {
    try {
      await super.write(...)
    } catch (err) {
      // do stuff
    }
  }

  // ...
}
```

utopia-remix/app/env.server.ts
```diff
@@ -19,12 +19,13 @@ export const ServerEnvironment = {
   AUTH0_CLIENT_ID: optionalEnv('AUTH0_CLIENT_ID', '<AUTH0_CLIENT_ID>'),
   AUTH0_REDIRECT_URI: optionalEnv('AUTH0_REDIRECT_URI', '<AUTH0_REDIRECT_URI>'),
   // FGA Credentials
-  FGA_STORE_ID: optionalEnv('FGA_STORE_ID', '<FGA_STORE_ID>'),
-  FGA_CLIENT_ID: optionalEnv('FGA_CLIENT_ID', '<FGA_CLIENT_ID>'),
-  FGA_SECRET: optionalEnv('FGA_SECRET', '<FGA_SECRET>'),
-  FGA_API_HOST: optionalEnv('FGA_API_HOST', '<FGA_API_HOST>'),
-  FGA_API_TOKEN_ISSUER: optionalEnv('FGA_API_TOKEN_ISSUER', '<FGA_API_TOKEN_ISSUER>'),
-  FGA_API_AUDIENCE: optionalEnv('FGA_API_AUDIENCE', '<FGA_API_AUDIENCE>'),
+  FGA_STORE_ID: optionalEnv('FGA_STORE_ID', ''),
```
seanparsons, line 22: Seems a little strange to default the value that would come from an environment variable to an empty string.

utopia-remix/app/services/fgaClient.server.ts
```diff
@@ -0,0 +1,65 @@
...
+    storeId: ServerEnvironment.FGA_STORE_ID,
+    // authorizationModelId: 'YOUR_MODEL_ID', // optional
+    credentials: getCredentials(fgaClientMode),
+  })
+}
+
+function getCredentials(fgaClientMode: 'local' | 'remote') {
+  // if working against a local server, we don't need credentials
+  if (fgaClientMode === 'local') {
+    return { method: CredentialsMethod.None } as const
```
seanparsons, line 28: Would it not make sense to define the return type and then I don't believe you'd need the `as const` suffixes?

seanparsons (approved):

utopia-remix/app/services/fgaClient.server.ts
liady: The problem is that openfga doesn't export the return type at all. I can just define it to be a union, though

utopia-remix/app/env.server.ts
liady: the idea here is that in most of the scenarios we dont need this value in local env file

Rheeseyb (approved):

ruggi (approved):

PR #5026, opened 2024-03-12
Component inspector: Array row and Object row shows Cartouche

balazsbajorics:
**Problem:**
For a component `<Component data={data} pull={5} />` we would show this component inspector:
image
Notice how the row for the `data` prop does not show a cartouche! The object's children elements show up as blue, but no point indicates that this is in fact a `data` variable being fed into the `data` prop.
Side note: opening the data picker would correctly highlihgt (with italics) that the selected data was indeed `data`
image

**Fix:**
image
Show a Cartouche for Object controls. And also for Array controls:
image


editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -840,23 +844,32 @@ const RowForObjectControl = React.memo((props: RowForObjectControlProps) => {
...
-                  fontWeight: 500,
-                  gap: 4,
                   cursor: props.disableToggling ? 'default' : 'pointer',
                 }}
               >
                 {title}
                 {unless(props.disableToggling, <ObjectIndicator open={open} />)}
               </PropertyLabel>
+              <div onClick={stopPropagation}>
+                <ControlForProp
```
balazsbajorics, line 863: this is the interesting change, RowForObjectControl renders ControlForProp which will render a Cartouche (if applicable) or return `null` if no cartouche was needed

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -563,28 +567,37 @@ const RowForArrayControl = React.memo((props: RowForArrayControlProps) => {
                 )}
               </SquareButton>
             ) : null}
+            <ControlForProp
```
balazsbajorics, line 570: this is the interesting change, RowForArrayControl renders ControlForProp which will render a Cartouche (if applicable) or return `null` if no cartouche was needed

editor/src/components/inspector/common/control-status.ts
```diff
@@ -174,7 +176,9 @@ function isControlled(
       return spiedValue != null
     } else if (
       modifiableAttributeIsAttributeValue(modifiableAttributeResult.value) ||
-      modifiableAttributeIsPartOfAttributeValue(modifiableAttributeResult.value)
+      modifiableAttributeIsPartOfAttributeValue(modifiableAttributeResult.value) ||
+      modifiableAttributeIsAttributeNestedArray(modifiableAttributeResult.value) ||
+      modifiableAttributeIsAttributeNestedObject(modifiableAttributeResult.value)
```
balazsbajorics, line 181: array literals and object literals are not considered controlled, because they are expandable without clearing their existing content

editor/src/components/inspector/sections/component-section/component-section.tsx
```diff
@@ -531,18 +535,18 @@ const RowForArrayControl = React.memo((props: RowForArrayControlProps) => {
...
-          <FlexRow style={{ gap: 5 }} ref={dataPickerButtonData.setReferenceElement}>
-            <PropertyLabel
-              target={[propPath]}
-              style={{ textTransform: 'capitalize', paddingTop: 2 }}
-            >
+          <FlexRow
+            style={{ flex: 1, gap: 5, justifyContent: 'space-between' }}
+            ref={dataPickerButtonData.setReferenceElement}
+          >
+            <PropertyLabel target={[propPath]} style={objectPropertyLabelStyle}>
```
balazsbajorics, line 546: I unified the look and feel of object and array sections, working on both of them highlighted just how different they looked, and it felt wildly offputting

seanparsons (approved):
Very nice!

editor/src/components/inspector/sections/component-section/component-section.tsx
Rheeseyb: That's good but I can see this drifting again. I wonder if we can extract a lot of this into a shared component to prevent that

Rheeseyb (approved):

PR #5028, opened 2024-03-13
Project Page Review and More Radix Components

lankaukk:
Making many small updates to the Projects Page:
- Adding more Radix Theme components
- First pass at the Sharing Dialog

Screenshot 2024-03-15 at 12 30 34 AM
Screenshot 2024-03-15 at 12 30 56 AM
Screenshot 2024-03-14 at 4 33 35 PM

https://github.com/concrete-utopia/utopia/assets/47405698/41f47d0e-2313-4dea-9e8a-4f44cb03a42a


utopia-remix/app/components/projectActionContextMenu.tsx
```diff
@@ -165,41 +166,117 @@ export const ProjectContextMenu = React.memo(({ project }: { project: ProjectWit
...
+                          {accessLevel === AccessLevel.PUBLIC ? 'Public' : 'Private'}
+                          <CaretDownIcon />
+                        </Button>
+                      </DropdownMenu.Trigger>
+                      <DropdownMenu.Content>
+                        <DropdownMenu.CheckboxItem
+                          style={{ height: 28, fontSize: 12, paddingLeft: 30 }}
+                          checked={accessLevel === AccessLevel.PUBLIC}
+                          disabled={accessLevel === AccessLevel.PUBLIC ? true : false}
+                          /* eslint-disable-next-line react/jsx-no-bind */
```
seanparsons, line 233: Why does this rule need to be disabled?

utopia-remix/app/components/projectActionContextMenu.tsx
```diff
@@ -173,59 +168,56 @@ export const ProjectContextMenu = React.memo(
...
-                onClick={onClick}
-                className={contextMenuItem()}
-              >
-                {entry.text}
-              </DropdownMenu.Item>
+              <Dialog.Root key={`separator-${index}`}>
+                <Dialog.Trigger>
+                  <ContextMenu.Item
+                    style={{ height: 28, fontSize: 12 }}
+                    /* eslint-disable-next-line react/jsx-no-bind */
```
ruggi, line 188: I think this has already been fixed on master, could you update/remove it?

utopia-remix/app/components/projectActionContextMenu.tsx
```diff
@@ -173,59 +168,56 @@ export const ProjectContextMenu = React.memo(
...
+                    onSelect={preventDefault}
+                  >
+                    <Flex justify={'between'} align={'center'} width={'100%'}>
+                      <Text>Share</Text>
+                      {pendingAccessRequests.length > 0 ? (
```
ruggi, line 193:
```ts
when(pendingAccessRequests.length > 0, ...
```

utopia-remix/app/components/projectActionContextMenu.tsx
```diff
@@ -173,59 +168,56 @@ export const ProjectContextMenu = React.memo(
...
+              </Dialog.Root>
             )
-          })}
-        </DropdownMenu.Content>
-      </DropdownMenu.Portal>
+          }
+          return (
+            <ContextMenu.Item
+              key={`entry-${index}`}
+              /* eslint-disable-next-line react/jsx-no-bind */
```
ruggi, line 208: this one should also go

utopia-remix/app/routes/projects.tsx
```diff
@@ -647,115 +594,139 @@ const ProjectCard = React.memo(
...
+
+    React.useEffect(() => {
+      const action = `/internal/projects/${project.proj_id}/access/requests`
+      accessRequestsFetcher.submit({}, { method: 'GET', action: action })
+      if (accessRequestsFetcher.state === 'idle' && accessRequestsFetcher.data != null) {
+        if (isProjectAccessRequestWithUserDetailsArray(accessRequestsFetcher.data)) {
+          setAccessRequests(accessRequestsFetcher.data)
+        }
+      }
+    }, [accessRequestsFetcher, project])
```
ruggi, line 610: This is causing/will cause an endless loop of requests to that fetcher! 😱 
1. The effect should not submit the fetcher, but instead _only_ be used to process new data on the fetcher (lines 605-609)
2. The submit should only happen on an explicit action – I think the right place would when the context menu is opened, see https://github.com/concrete-utopia/utopia/blob/4bbd34c90173a22e7d13a1cb13641c4a64a08f44/utopia-remix/app/routes/projects.tsx#L905-L906

ruggi (changes requested):
The code looks good overall 👍 

However the `setRequestAccess` logic which needs updating otherwise that poor server will explode in a million shards 😅

Rheeseyb (approved):

PR #5034, opened 2024-03-13
BFF reasonable env vars

ruggi:
**Problem:**

The env vars situation is getting out of control and makes local development kind of miserable, because it forces you to have certain env vars that are not necessarily required for your work.

**Fix:**

Use `mustEnvOrLocalFallback` instead of `mustEnv` and `optionalEnv`: for production we _will_ need all env vars configured correctly, but for local development we don't. E.g. you don't need the FGA auth creds if you are not going to use it.

If an env var is not configured on production, it will scream, but locally it will default to noops.
You can of course continue to have custom values for all the env vars in the `.env.sample` (which are now commented out, to signal they are not required) or the `readme`.

From now on we should always rely on `mustEnvOrLocalFallback` unless we really, _really_, **really** have good reasons.


utopia-remix/app/services/fgaClient.server.ts
```diff
@@ -87,14 +90,14 @@ function getCredentials(fgaClientMode: 'local' | 'remote'): CredentialsConfig {
 function getFgaClientMode(): 'local_mock' | 'local' | 'remote' {
   if (ServerEnvironment.environment === 'local') {
     if (ServerEnvironment.FGA_API_HOST == '') {
-      console.log('Using mock FGA client')
+      console.info('Using mock FGA client')
```
ruggi, line 93: this is because of lint

utopia-remix/app/env.server.ts
```diff
@@ -10,27 +10,29 @@ declare global {
...
-  AUTH0_CLIENT_ID: optionalEnv('AUTH0_CLIENT_ID', '<AUTH0_CLIENT_ID>'),
-  AUTH0_REDIRECT_URI: optionalEnv('AUTH0_REDIRECT_URI', '<AUTH0_REDIRECT_URI>'),
-  // FGA Credentials
-  FGA_STORE_ID: optionalEnv('FGA_STORE_ID', ''),
-  FGA_CLIENT_ID: optionalEnv('FGA_CLIENT_ID', ''),
-  FGA_SECRET: optionalEnv('FGA_SECRET', ''),
-  FGA_API_HOST: optionalEnv('FGA_API_HOST', ''),
-  FGA_API_TOKEN_ISSUER: optionalEnv('FGA_API_TOKEN_ISSUER', ''),
-  FGA_API_AUDIENCE: optionalEnv('FGA_API_AUDIENCE', ''),
-  FGA_API_SCHEME: optionalEnv('FGA_API_SCHEME', 'https'),
```
ruggi, line 30: I figured we don't really need this - as we'll always have ssl on prod and never on local

utopia-remix/app/env.server.ts
```diff
@@ -39,19 +41,40 @@ export type BrowserEnvironment = {
...
 }
 
-function optionalEnv(key: string, defaultValue: string): string {
-  const value = process.env[key]
-  return value ?? defaultValue
+function mustEnvOrLocalFallback(key: string, localFallback: string) {
+  const value = process.env[key] ?? ''
+  if (value !== '') {
+    return value
+  }
```
balazsbajorics, line 75: I think the `?? ''` here is not needed

```suggestion
  const value = process.env[key]
  if (value != null) {
    return value
  }
```

balazsbajorics: or maybe even 
```suggestion
  if (key in process.env) {
    return process.env[key]
  }
```

ruggi: the lookup can give back an undefined tho, _or_ you can pass an env var that is defined but is an empty string

utopia-remix/.env.sample
```diff
@@ -1,26 +1,23 @@
-APP_ENV="local"
-BACKEND_URL="http://127.0.0.1:8002"
-CORS_ORIGIN="http://localhost:8000"
-DATABASE_URL="postgres://<username>:postgres@localhost:5432/utopia"
-REACT_APP_EDITOR_URL="http://localhost:8000"
+# APP_ENV="local"
+# BACKEND_URL="http://127.0.0.1:8002"
```
Rheeseyb, line 2: Since we can now locally use `localhost` here again (I forget why we couldn't before), and this is just `EXPRESS_PROXY_TARGET_HOST` plus `EXPRESS_PROXY_TARGET_PORT`, I'd say we don't need a separate variable for these - we should either have one for `BACKEND_URL` and calculated the proxy target from that, or vice versa

utopia-remix/.env.sample
```diff
@@ -1,26 +1,23 @@
...
+# CORS_ORIGIN="http://localhost:8000"
+# REACT_APP_EDITOR_URL="http://localhost:8000"
```
Rheeseyb, line 4: I believe these two (plus `UTOPIA_CDN_URL`) will always be the same, so unless I'm missing something there's no need for multiple variables

ruggi: [`d8f6218` (#5034)](https://github.com/concrete-utopia/utopia/pull/5034/commits/d8f621809da9058ff7b5c0aa90a620216c7e0493)

ruggi: the CDN URL can be different on prod tho

ruggi:
@Rheeseyb addressed your comments, thanks! I updated the vars, we'll need to rename the `REACT_APP_EDITOR_URL` to `EDITOR_URL` on prod, but I think it's a tolerable annoyance considering the readability benefits

utopia-remix/app/env.server.ts
liady: wait, on local sometimes you want to work against a remote fga server ('https') and sometimes against a local one ('http'). It's actually important (we can try and set it up somewhere else, but we do need it)

utopia-remix/app/services/fgaClient.server.ts
```diff
@@ -46,7 +46,10 @@ export function createFgaClient() {
     return mockOpenFgaClient
   }
   return new WriteSafeOpenFgaClient({
-    apiScheme: ServerEnvironment.FGA_API_SCHEME,
+    apiScheme:
+      ServerEnvironment.environment === 'local' || ServerEnvironment.environment === 'test'
```
liady, line 50: no, this is not actually the the correct logic... It should be according to `fgaclientMode` if we want to decide it here

liady (changes requested):

liady (approved):

PR #5033, opened 2024-03-13
Collaborative projects

liady:
This PR adds the ability to:
* Define a project as collaborative
* Ask for access for this project
* Grant access requests

This builds on #5039 .

The flow (with temp UI):

<video src="https://github.com/concrete-utopia/utopia/assets/7003853/14ba3484-0215-4329-9df4-441f4269f7c3" />

PR #5043, opened 2024-03-14
Share popup

liady:
**Problem:**

**Fix:**

- Did [other things]

Fixes #[ticket_number]  (<< pls delete this line if it's not relevant)


utopia-remix/app/types.ts
```diff
@@ -43,6 +42,20 @@ export function userToCollaborator(user: UserDetails): Collaborator {
   }
 }
 
+export interface AccessRequest {
+  status: AccessRequestStatus
+  token: string
+  user_id: string
+  project_id: string
```
Rheeseyb, line 49: I'm assuming this is because you wanted to match the keys from Prisma, but unless there's a good reason to, please don't use snake case here

utopia-remix/app/components/SharePopup.tsx
```diff
@@ -0,0 +1,223 @@
+/* eslint-disable react/jsx-no-bind */
```
Rheeseyb, line 1: You shouldn't need this. If you ever do need to disable rules like this, please only do so for specific lines where possible

utopia-remix/app/types.ts
liady: yes, but removed completely.

Rheeseyb (approved):

seanparsons (approved):

utopia-remix/app/components/SharePopup.tsx
```diff
@@ -0,0 +1,189 @@
...
+          </Dialog.Close>
+        </Flex>
+        <Flex justify='between'>
+          <Text size='1'>Project Visibility</Text>
+          <VisibilityDropdown
+            accessLevel={accessLevel}
+            changeProjectAccessLevel={changeProjectAccessLevel}
+          />
+        </Flex>
+        {accessRequests.length > 0 ? (
```
ruggi, line 86: `when(accessRequests.length > 0, ...)`

ruggi (approved):

liady:
Merged into #5033

PR #5045, opened 2024-03-14
Property controls now have required and default value fields.

seanparsons:
**Problem:**
For the cartouche functionality we need a way to specify that a property is required and what the default value should be.

**Fix:**
For all the regular controls, we now have `required` and `defaultValue` fields.

Currently this does not drive any functionality in itself, that will be part of the follow up work.

**Commit Details:**
- For every type that is part of `RegularControlDescription` there are now `required` and `defaultValue` fields.
- Added `flatMapParser` utility function.
- Added `applicative10Either` utility function.
- Added `parseEnumValueOrBasicControlOption`, `parseVector2`, `parseVector3`, `parseVector4`, `parseEuler`, `parseMatrix3` and `parseMatrix4` functions which are used to parse the default value fields.
- The parsers for the `RegularControlDescription` types now parse the `required` and `defaultValue` fields.
- Expanded the property controls parser tests to check the `required` and `defaultValue` fields.


Rheeseyb (commented):
👍

PR #5046, opened 2024-03-14
Do not show extra CHILDREN in navigator without jsx element render props.

gbalint:
**Problem:**
1. We only show render props with jsx elements in them
2. We show CHILDREN label before normal children when there ar render props before them
3. We accidentally show the children label even when it has other render props, but not with jsx elements in them (so they are not shown)

**Fix:**
Only set renderPropFound to true when we can really insert something into the navigator targets


seanparsons (approved):

Rheeseyb:
Can you add a sample project to this please, because I'm struggling to understand this one

PR #5047, opened 2024-03-14
Dont allow requesting access as a logged out user

liady:


Rheeseyb:
Looks like this has been succeeded by https://github.com/concrete-utopia/utopia/pull/5054 (please reopen if this work should be continued regardless)

PR #5050, opened 2024-03-14
insert parsed preferred children

bkrmendy:
## Problem
Inserting elements specified in preferredChildComponents is half-finished:
- inserting the element as render prop triggers an error flash on the canvas (since the element is inserted as a JS expression, not as a JSX element, and until the source file is re-parsed, the JSX element syntax will trigger a parsing error when it's evaluated)
- inserting a non-intrinsic element breaks the canvas, because the import statement that imports the element isn't added to the file.

## FIx
These two problems have a common root case, which is that `ComponentDescriptor.preferredChildComponents` isn't parsed in `componentDescriptorForComponentToRegister`. This PR adds component descriptor parsing support for `ComponentDescriptor.preferredChildComponents`, and improves the render prop insert popup to use the parsed info from `ComponentDescriptor.preferredChildComponents`.

### Details
- added parsing support for `ComponentDescriptor.preferredChildComponents` in `componentDescriptorForComponentToRegister`
- make it possible to specify `importsToAdd` in the `SET_PROP` action, and add those imports to the affected file in the action handler. 
- adjust the render prop popup accomodate the new internal preferred component type


editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -236,6 +251,169 @@ function variantsForComponentToRegister(
...
+  }
+}
+
+async function makePropertyDescriptors(
+  properties: PropertyControlsFromUtopiaAPI,
+  context: { moduleName: string; workers: UtopiaTsWorkers },
+): Promise<PropertyDescriptorResult<PropertyControls>> {
+  let result: PropertyControls = {}
+
+  for await (const [propertyName, descriptor] of Object.entries(properties)) {
```
ruggi, line 398: TIL you can `for await` 😮

ruggi (approved):

editor/src/core/property-controls/property-controls-local.ts
seanparsons: Wat.

editor/src/components/editor/action-types.ts
```diff
@@ -727,6 +727,7 @@ export interface SetProp {
   target: ElementPath
   propertyPath: PropertyPath
   value: JSExpression
+  importsToAdd?: Imports
```
gbalint, line 730: Usually we prefer non-optional function parameters, so you can not forget about them accidentally.

editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
```diff
@@ -77,6 +86,12 @@ interface RenderPropPickerProps {
   id: string
 }
 
+interface PreferredChildToInsert {
+  elementToInsert: (uid: string) => JSXElementChild
+  additionalImports: Imports | null
```
gbalint, line 91: do we really need the null? what is the difference between null and empty object?

editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
```diff
@@ -85,35 +100,67 @@ export const RenderPropPicker = React.memo<RenderPropPickerProps>(({ key, id, ta
...
       e.preventDefault()
 
+      const uid = generateConsistentUID(
+        'prop',
+        new Set(getAllUniqueUids(projectContentsRef.current).uniqueIDs),
+      )
+
+      const element = preferredChildToInsert.elementToInsert(uid)
+      if (element.type !== 'JSX_ELEMENT') {
+        throw new Error('only JSX elements are supported as preferred components')
```
gbalint, line 117: Is it sure? E.g. fragments could make sense or even expressions or just a string

editor/src/components/editor/action-types.ts
bkrmendy: Absolutely, but I wanted to avoid adding a ton of unrelated changes to the PR because of this. Also, the command constructor functions use the same pattern. I don't have strong opinions on this, I can update the PR if we don't want the optional param here

editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
bkrmendy: This specifically refers to elements coming from `preferredChildComponents`, and this is only valid until we add support for fragments and conditionals in render props (which we intend to do soon)

editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
bkrmendy: Turns out we don't need the `| null`, fixed it in https://github.com/concrete-utopia/utopia/pull/5050/commits/9f522b69ad78d99d72d84521c8f50d9e8f3f7f50

editor/src/components/editor/action-types.ts
bkrmendy: I made `importsToAdd` non-optional in the action type, and defaulted it to an empty imports object https://github.com/concrete-utopia/utopia/pull/5050/commits/71276a4a4eeca3b26443c699cab9ab9e5072feae

gbalint (approved):

editor/src/components/inspector/sections/component-section/component-section.spec.browser2.tsx
```diff
@@ -835,11 +835,22 @@ describe('Controls from registering components', () => {
...
+                      importedAs: null,
+                      importedFromWithin: [
+                        {
+                          alias: 'span',
+                          name: 'span',
+                        },
+                      ],
+                      importedWithName: null,
+                    },
+                  },
```
Rheeseyb, line 852: This is unexpected. Why are we importing `span` from the storyboard

editor/src/components/navigator/navigator.spec.browser2.tsx
```diff
@@ -919,6 +919,105 @@ export default Components
...
+    ['/utopia/components.utopia.js']: `const Components = {
+      '/utopia/storyboard': {
+        Card: {
+          properties: {
+            header: {
+              control: 'jsx',
+              preferredChildComponents: [
+                {
+                  name: 'Heading',
+                  additionalImports: '../src/heading',
```
Rheeseyb, line 994: This should be absolute, right? Otherwise it only works in specific files

editor/src/core/property-controls/property-controls-local.ts
Rheeseyb: 😮

editor/src/components/navigator/navigator.spec.browser2.tsx
bkrmendy: Thanks for catching this, I updated the code so that no imports are added for built-in elements https://github.com/concrete-utopia/utopia/pull/5050/commits/6f71152ea5296bbb221507c86e78878e3b73ed56

editor/src/components/inspector/sections/component-section/component-section.spec.browser2.tsx
bkrmendy: Thanks for catching this, I updated the code so that no imports are added for built-in elements https://github.com/concrete-utopia/utopia/pull/5050/commits/6f71152ea5296bbb221507c86e78878e3b73ed56

Rheeseyb (commented):
👍

PR #5052, opened 2024-03-15
Express: handle pipe errors gracefully

### LINKED ISSUE #5051: Handle express pipe errors

ruggi:
Fix #5051 

**Problem:**

The express server can crash during proxied requests, if they cause unmanaged errors during the piping.

**Fix:**

1. Add an explicit handler for errors happening on the piped requests, returning a graceful error to the client
2. Additionally, make sure that uncaught errors on the whole process are logged out instead of making the whole thing explode, so the lights stay on

### DIFF
--- a/utopia-remix/server.js
+++ b/utopia-remix/server.js
@@ -143,7 +143,14 @@ function proxy(originalRequest, originalResponse) {
     },
   )
 
-  originalRequest.pipe(proxyRequest)
+  originalRequest
+    .pipe(proxyRequest)
+    // if the request fails for any non-explicit reason (e.g. ERRCONNREFUSED),
+    // handle the error gracefully and return a common status code back to the client
+    .on('error', (err) => {
+      console.error('failed proxy request', err)
+      originalResponse.status(502).json({ error: 'Service temporarily unavailable' })
+    })
 }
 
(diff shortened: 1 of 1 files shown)


Rheeseyb (approved):

utopia-remix/server.js
```diff
@@ -191,3 +198,12 @@ app.listen(environment.PORT, listenCallback(environment.PORT))
 if (process.env.NODE_ENV === 'development') {
   app.listen(environment.PORT + 1, listenCallback(environment.PORT + 1))
 }
+
+// To make sure everythin keeps working and the server doesn't crash, if
+// there are any uncaught errors, log them out gracefully
+process.on('uncaughtException', (err) => {
```
seanparsons, line 204: Probably want to put this at the top of this file, so they're setup before anything comes online and starts doing things.

seanparsons (approved):

utopia-remix/server.js
ruggi: [`ccb086e` (#5052)](https://github.com/concrete-utopia/utopia/pull/5052/commits/ccb086efb0198a0347b5f6f72c145481703cd27b)

PR #5054, opened 2024-03-15
Only show the Request Access button when the user is logged in

### LINKED ISSUE #5053: Only show request access for logged in users

ruggi:
Fix #5053 

**Problem:**

The `ProjectNotFound` page shows the `Request access` button if the project is collaborative, but it should only show it if the user is _also_ logged in.

**Fix:**

Add a flag for it. I passed the value coming from the persistence's context – the same could also be done with `getLoginState` directly inside the component but it felt more consistent this way.
Happy to change it if somebody feels otherwise!

### DIFF
--- a/editor/src/templates/project-not-found/ProjectNotFound.tsx
+++ b/editor/src/templates/project-not-found/ProjectNotFound.tsx
@@ -69,7 +69,7 @@ export default function ProjectNotFound({
             <ActionButton text='Return Home' />
           </a>
           {when(
-            projectId != null && loggedIn === true,
+            projectId != null && loggedIn,
             <ActionButton
               text={accessRequested ? 'Access Requested' : 'Request Access'}
               onClick={requestAccess}
--- a/editor/src/templates/project-not-found/ProjectNotFound.tsx
+++ b/editor/src/templates/project-not-found/ProjectNotFound.tsx
@@ -13,8 +13,8 @@ export default function ProjectNotFound({
   projectId,
   loggedIn,
 }: {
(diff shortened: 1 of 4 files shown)


editor/src/templates/project-not-found/ProjectNotFound.tsx
```diff
@@ -1,15 +1,21 @@
...
 
 const PyramidLight404 = `${process.env.UTOPIA_DOMAIN}/editor/404_pyramid_light.png?hash=${process.env.UTOPIA_SHA}`
 
-export default function ProjectNotFound({ projectId }: { projectId?: string }) {
+export default function ProjectNotFound({
+  projectId,
+  loggedIn,
+}: {
+  projectId?: string
+  loggedIn?: boolean
```
Rheeseyb, line 17: Do these need to be optional?

Rheeseyb (approved):

seanparsons (approved):

editor/src/templates/project-not-found/ProjectNotFound.tsx
ruggi: nah I don't think they do – I adapted to the existing `?:` but it makes sense to have them explicit

ruggi: fixed in [`be43df5` (#5054)](https://github.com/concrete-utopia/utopia/pull/5054/commits/be43df5e02c673827cff314efb0900d231fbca39)

PR #5060, opened 2024-03-17
Move eslint config to a separate package

liady:
**Problem:**
`utopia-remix` didn't have an eslint config, which caused the local IDE not to alert on esling errors (that fail the build).

**Fix:**
* In order to have `editor` and `utopia-remix` to use the same config, the file `.eslintrc-lite.js` was used as the base config, it was moved to its own package so that its dependencies could be installed with it.
* An install step was added to `start-minimal`, to install the dependencies of the new config.
* the `lint-other-projects` command now uses this new package as the eslint config


utopia-remix/app/components/sharingDialog.tsx
```diff
@@ -127,6 +127,7 @@ function AccessRequests({
         <Flex key={request.token} justify='between'>
           <Text size='1'>{request.User?.name ?? request.User?.email ?? request.user_id}</Text>
           {status === AccessRequestStatus.PENDING ? (
+            // eslint-disable-next-line react/jsx-no-bind
```
liady, line 130: since it's inside a map function, the only way to avoid this errors here and in :192,  is to extract it to a different component. I'm not sure it's not an overkill here

editor/.eslintrc-lite.js
```diff
@@ -1,186 +1,3 @@
 module.exports = {
```
balazsbajorics, line 1: do we still need this file for anything?

balazsbajorics (approved):

editor/.eslintrc-lite.js
liady: I replaced now all of its occurences with references to `eslint-config-utopia`

PR #5064, opened 2024-03-18
Send user permissions object to the editor

liady:
This PR adds the ability to send all the permissions to the editor as an object - so it can be used for fine grained auth (for example - turnning on/off editing abilities or commenting functionality).
This PR adds a new route, `/project/{id}/permissions` - that returns an object with all of the fga permissions and their values.


balazsbajorics (approved):

editor/src/components/editor/server.ts
```diff
@@ -325,6 +326,40 @@ export async function saveThumbnail(thumbnail: Buffer, projectId: string): Promi
...
+      }
+    case 'LOGIN_NOT_YET_KNOWN':
+    case 'NOT_LOGGED_IN':
+    case 'LOGIN_LOST':
+    case 'OFFLINE_STATE':
+    case 'COOKIES_OR_LOCALFORAGE_UNAVAILABLE':
+      return emptyUserPermissions()
+    default:
+      const _exhaustiveCheck: never = loginState
+      throw new Error(`Unknown login state.`)
```
ruggi, line 359: `assertNever` should be equivalent here

editor/src/components/editor/server.ts
```diff
@@ -325,6 +326,40 @@ export async function saveThumbnail(thumbnail: Buffer, projectId: string): Promi
   }
 }
 
+export async function getUserPermissions(
```
ruggi, line 329: this _must_ check that the BFF is not enabled, because this would otherwise explode on production

editor/src/templates/editor.tsx
```diff
@@ -325,9 +326,13 @@ export class Editor {
     )
 
     void getLoginState('cache').then((loginState: LoginState) => {
+      const projectId = getProjectID()
       startPollingLoginState(this.boundDispatch, loginState)
       this.storedState.userState.loginState = loginState
-      void getUserConfiguration(loginState).then((shortcutConfiguration) => {
+      void Promise.all([
```
ruggi, line 332: see the comment about checking the BFF enabled in `getUserPermissions` or this whole promise will throw

utopia-remix/app/services/fgaService.server.ts
```diff
@@ -35,6 +35,18 @@ async function checkUserProjectPermission(
   return allowed ?? false
 }
 
+export async function getAllPermissions(projectId: string, userId: string) {
+  const { relations } = await fgaClient.listRelations({
+    user: `user:${userId}`,
+    object: `project:${projectId}`,
+    relations: fgaUserProjectPermission.map((permission) => permission),
```
ruggi, line 42: is this `map` needed?

website-next/components/projects/projects-page-component.tsx
```diff
@@ -529,13 +529,15 @@ export class ProjectsPage extends React.Component<EmptyProps, ProjectsState> {
                 <label>Sort</label>
                 <SortButton
                   selected={this.state.sortMode === 'date'}
+                  // eslint-disable-next-line react/jsx-no-bind
```
ruggi, line 532: I think these are remnants and can be removed

ruggi (changes requested):

utopia-remix/app/services/fgaService.server.ts
liady: it was there since TS didn't allow casting from a const array into `string[]`. I changed it now to have an explicit casting using `as unknown as string[]`, to move this issue to the type world instead of the runtime :)

website-next/components/projects/projects-page-component.tsx
liady: actually all this code is inside an old React class (😳), and now we're linting it properly so it throws this error. I'm not sure if we need to fix it though

utopia-remix/app/services/permissionsService.server.ts
```diff
@@ -39,6 +40,22 @@ export async function setProjectAccess(projectId: string, accessLevel: AccessLev
...
 }
 
+export async function getAllPermissions(
+  projectId: string,
+  userId: string | null,
+  ownerId: string,
+): Promise<Record<fgaService.FgaUserProjectPermission, boolean>> {
+  const fgaPermissions = await fgaService.getAllPermissions(projectId, userId ?? ANONYMOUS_USER_ID)
+  // this is here since it's our internal logic, not part of FGA
+  if (userId === ownerId) {
```
Rheeseyb, line 50: Seems we could avoid the `fgaService` call altogether then if the user is the owner, and instead use the same logic as `emptyUserPermissions` in that specific case

Rheeseyb (approved):

utopia-remix/app/services/fgaService.server.ts
```diff
@@ -35,6 +35,28 @@ async function checkUserProjectPermission(
...
+    relations: fgaUserProjectPermission as unknown as string[],
+  })
+  return fgaUserProjectPermission.reduce((acc, permission, index) => {
+    acc[permission] = relations.includes(permission)
+    return acc
+  }, {} as Record<FgaUserProjectPermission, boolean>)
+}
+
+export function getPermissionsOverride(value: boolean): Record<FgaUserProjectPermission, boolean> {
+  return fgaUserProjectPermission.reduce((acc, permission) => {
```
seanparsons, line 54: In the editor source, we have a function that does this kind of construction called `mapArrayToDictionary` BTW.

seanparsons (approved):

ruggi (approved):

PR #5070, opened 2024-03-18
File-based registerComponent

bkrmendy:
**Problem**
`registerComponent` is called by the evaluation of user code, and a dispatch is triggered by that.
It is really hard to control when and how many times the dispatches are called, and we have multiple errors based on different circumstances:
1. sometimes registerComponent doesn't work from some files
2. sometimes a deleted registerComponent does not remove the specific properties

Try it here: https://utopia.fish/p/a755ed92-savory-planet/?branch_name=spike-new-register-component

**Fix**

This PR fixes the above problems by introducing a different way of registering components: sidecar files.

- component controls are described in descriptor files (which are postfixed by `.utopia.js` and reside in the `utopia` folder`)
- the lifecycle management is easy, we update the property controls when a descriptor file has been changed
- we store which file registered which controls in a cache, so we know what to overwrite/delete after a change
- the `updatePropertyControlsInfo` action overwrites all the property controls, to eliminate further lifecycle management issues
- Example component descriptor file:
```javascript
const Components = {
  '/src/card-with-image': {
    CardWithImage: {
      supportsChildren: false,
      properties: {
        title: {
          control: 'string-input',
        },
        imageUrl: {
          control: 'string-input',
        },
      },
      variants: [],
    },
  },
  '/src/card-with-image-2': {
    CardWithImageV2: {
      supportsChildren: false,
      properties: {
        data: {
          control: 'object',
          object: {
            title: {
              control: 'string-input',
            },
            imageUrl: {
              control: 'string-input',
            },
          },
        },
      },
      variants: [],
    },
  },
}

export default Components
```
Missing:
- `registerModule`, `registerInternalComponent` and `registerExternalComponent` are not removed from `utopia-api` yet, because we have lots of test code, default code, etc which includes an import with these, and I wanted to reduce the size of the PR
- the execution scope is not passed in when the descriptor file is evaluated, so you can not import from other files


editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -74,6 +61,130 @@ async function parseInsertOption(
...
+    return
+  }
+  const combinedTopLevelArbitraryBlock = componentsFile.parseResult.combinedTopLevelArbitraryBlock
+
+  if (combinedTopLevelArbitraryBlock == null) {
+    return
+  }
+
+  try {
+    // TODO: this isn't the right function to call, as `resolveParamsAndRunJsCode` doesn't actually give us access to the exports/default exports
```
Rheeseyb, line 87: I think we'd need to take a similar approach as the canvas here, where we using the `editor.codeResultCache.curriedRequireFn` with the current project contents to create a `require` function that we can use to evaluate a file: https://github.com/concrete-utopia/utopia/blob/2372876aec02dbb1f524a5adcbfc1d9e2bff0790/editor/src/components/canvas/ui-jsx-canvas.tsx#L387-L390

editor/src/components/custom-code/code-file.ts
```diff
@@ -144,6 +144,7 @@ export function componentDescriptor(
 
 export interface ComponentDescriptorWithName extends ComponentDescriptor {
   componentName: string
+  moduleName: string
```
balazsbajorics, line 147: for components that exist in the project, is this some sort of normalized path? If yes, I would put this fact in the name!

editor/src/components/editor/store/dispatch.tsx
```diff
@@ -349,6 +349,7 @@ function maybeRequestModelUpdate(
         const updates = parseResult.map((fileResult) => {
           return parseResultToWorkerUpdates(fileResult)
         })
+        void maybeUpdateComponentDescriptor(parseResult, workers, dispatch)
```
balazsbajorics, line 352: do not dispatch from the dispatch please!

editor/src/components/custom-code/code-file.ts
```diff
@@ -151,7 +151,7 @@ export function componentDescriptor(
 
 export interface ComponentDescriptorWithName extends ComponentDescriptor {
   componentName: string
-  moduleName: string
+  filenameNoExtension: string
```
balazsbajorics, line 154: is this the file path with folders? or just the filename? won't it clash if there's two files with the same name? I searched the codebase and nothing seems to _set_ componentName or filenameNoExtension

gbalint: FYI we discussed this, and agreed on moduleName and a comment about it

editor/src/components/inspector/sections/component-section/component-section.spec.browser2.tsx
```diff
@@ -968,52 +970,6 @@ describe('Controls from registering components', () => {
   })
 })
 
-/*
-describe('Delete cartouche handling', () => {
```
Rheeseyb, line 972: Please don't remove the commented out tests. Please can you instead try restoring them, since they _should_ work now with this new approach. They were commented because the old approach was causing unrelated tests to fail when these ran

editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -74,6 +61,164 @@ async function parseInsertOption(
...
+    componentDescriptorUpdates,
+  )
+  dispatch([EditorActions.updatePropertyControlsInfo(updatedPropertyControlsInfo)])
+}
+
+interface ComponentDescriptorFileLookup {
+  [path: string]: ComponentDescriptorWithName[]
+}
+
+const COMPONENTS_DESCRIPTOR_FILE_CACHE: { current: ComponentDescriptorFileLookup } = {
```
Rheeseyb, line 170: (Not for this PR) I'm thinking we should move this into the editor store so that we can avoid globals here (for the sake of tests)

Rheeseyb (approved):
Nice work!

seanparsons (approved):
All of the thumbs up.

PR #5073, opened 2024-03-18
Fix eslint-plugin-react versions

liady:
**Problem:**
VSCode eslint didn't catch functions in `react/jsx-no-bind` rule, whereas the prettier config did catch them.

**Fix:**
Match the versions of `eslint-plugin-react=*`


editor/tsconfig.json
```diff
@@ -25,5 +25,6 @@
       "worker-imports": ["src/core/workers/worker-import-utils.ts"]
     }
   },
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules"],
+  "include": ["src", "test", ".eslintrc.js"]
```
Rheeseyb, line 29: Why do we need this now?

liady: without it there is a TS error in the actual .eslintrc.js file, but it's not important nor it blocks anything (removed it for now)

seanparsons (approved):

liady:
Temporarily closing as Spike

PR #5074, opened 2024-03-18
Express: fix hot reload

ruggi:
**Problem:**

Hot reloading can break on Remix due to a racy update to the build artifacts.

**Fix:**

Give the handler a little time to breathe with a 1s timeout, and hot reload works again.

### DIFF
--- a/utopia-remix/server.js
+++ b/utopia-remix/server.js
@@ -81,8 +81,12 @@ installGlobals()
 async function createDevRequestHandler(initialBuild) {
   let build = initialBuild
   async function handleServerUpdate() {
+    // 0. give the rebuild time to breathe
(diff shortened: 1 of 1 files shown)


Rheeseyb (approved):

bkrmendy (approved):

utopia-remix/server.js
```diff
@@ -81,8 +81,12 @@ installGlobals()
 async function createDevRequestHandler(initialBuild) {
   let build = initialBuild
   async function handleServerUpdate() {
+    // 0. give the rebuild time to breathe
+    await new Promise((res) => setTimeout(res, 1000))
```
balazsbajorics, line 85: is there something we could explicitly await? I am always afraid of these fixed timeouts

PR #5069, opened 2024-03-18
Copy link in share dialog

ruggi:
Fix #5068 

**Problem:**

When the share dialog is open and the project is collaborative, to copy the project link you need to close the dialog and go back to the menu, which is clunky.

**Fix:**

Show a selectable/copyable input field (read only) with the project link when the share modal is open for public/collaborative projects.

https://github.com/concrete-utopia/utopia/assets/1081051/9971ca49-a8bb-4422-8d9e-77574a47c812

Also, this PR fixes two tangential problems:
1. The hot-reload breaks due to racy updates to the build artifacts, so a 1s timeout is added which makes everything works fine
2. The env vars are currently injected with a hacky solution into the `window` object; use zustand instead (non-persisted) so it's always available across the application.

**Notes**
- As usual, no design work besides the minimum.

PR #5075, opened 2024-03-18
Remix: use zustand store for env instead of window

ruggi:
**Problem:**

We're currently relying on a hacky solution for the env variables in the client portion of Remix, injecting them into the `window` object, and this can lead to issues in our components as I found out while developing https://github.com/concrete-utopia/utopia/pull/5069.

**Fix:**

Use the zustand store (non persisted) instead, so we don't rely on `dangerouslySetInnerHTML` and still can access all the env vars. They are initialized at the mount of the root, as it was before when checking `window.ENV`.


utopia-remix/app/components/projectActionContextMenu.tsx
```diff
@@ -102,10 +104,7 @@ export const ProjectActionsMenu = React.memo(
             'separator',
             {
               text: 'Copy Link',
-              onClick: (selectedProject) => {
-                window.navigator.clipboard.writeText(projectEditorLink(selectedProject.proj_id))
-                // TODO notification toast
-              },
+              onClick: (selectedProject) => copyProjectLink(selectedProject.proj_id),
```
Rheeseyb, line 107: Oops!

ruggi: what am I not seeing here? :D

utopia-remix/app/util/use-copy-project-link.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
```
Rheeseyb, line 1: This is meant to be part of #5069 too

utopia-remix/app/root.tsx
```diff
@@ -50,6 +46,11 @@ export const headers: HeadersFunction = () => ({
 export default function App() {
   const data = useLoaderData<typeof loader>()
   const isDarkMode = useIsDarkMode()
+  const setEnv = useProjectsStore((store) => store.setEnv)
```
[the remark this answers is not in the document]
ruggi: ha, makes sense. I did a bit of type gymnastics and ended up with [`0d72593` (#5075)](https://github.com/concrete-utopia/utopia/pull/5075/commits/0d72593709186ec94d79dee45502415270c847da), which is based on https://docs.pmnd.rs/zustand/guides/initialize-state-with-props . How does it look?

balazsbajorics (approved):
looks good!

bkrmendy (approved):

PR #5076, opened 2024-03-18
Test project validator

ruggi:
(sharded out of https://github.com/concrete-utopia/utopia/pull/5072 for scope's sake)

**Problem:**

The project validator is not tested itself.

**Fix:**

1. Refactor the validator and its logic so it's easier to test and to follow (it can and will get complicated)
2. Simplify it a bit so it always returns consistent status codes (404 by default, 403 if the project is not accessible but the user can request access to it _and_ the related flag is set)
3. Test it

### DIFF
--- a/utopia-remix/app/handlers/validators.spec.ts
+++ b/utopia-remix/app/handlers/validators.spec.ts
@@ -16,6 +16,8 @@ import { Status } from '../util/statusCodes'
 
 describe('validators', () => {
   describe('validateProjectAccess', () => {
+    let hasUserProjectPermissionMock: jest.SpyInstance
+
     afterEach(async () => {
       await truncateTables([
         prisma.projectID,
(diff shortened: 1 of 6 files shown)


utopia-remix/app/handlers/validators.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      })
+      const got = await validateProjectAccess(perm, { getProjectId: () => 'one' })(
+        newTestRequest({ authCookie: 'bob-key' }),
+        {},
+      )
+      expect(got.ok).toBe(true)
+    })
+
+    it('does nothing if the user has access', async () => {
+      jest.spyOn(permissionsService, 'hasUserProjectPermission').mockResolvedValue(true)
```
liady, line 78: instead of repeating `jest.spyOn(permissionsService, 'hasUserProjectPermission').<method>` we can save the instance at the top:
`const hasUserProjectPermissionMock= jest.spyOn(permissionsService, 'hasUserProjectPermission')` 
and just use `hasUserProjectPermissionMock.<method>` everywhere.
(safer from typos, easier to read, etc)

bkrmendy (approved):

PR #5086, opened 2024-03-19
Proxy project editor entry links through remix

liady:
This PR creates routes for `/p/projectId` and `/project/projectId` in Remix, so we will be able to return 404/403 pages directly from these routes.


utopia-remix/server.js
```diff
@@ -180,8 +180,8 @@ app.use('/authenticate', proxy)
 app.use('/editor', proxy)
 app.use('/hashed-assets.json', proxy)
 app.use('/logout', proxy)
-app.use('/p', proxy)
-app.use('/project', proxy)
+app.use('/p/?$', proxy)
```
liady, line 183: this will leave only the empty `/p` links to go through our regular proxy (and not remix)

Rheeseyb: It's worth adding a comment to that effect here

utopia-remix/app/routes/loaders/editorLoader.server.tsx
```diff
@@ -0,0 +1,68 @@
...
+    GET: {
+      handler: getProjectForEditor,
+      validator: ALLOW,
+    },
+  })
+}
+
+const validator = validateProjectAccess(UserProjectPermission.CAN_VIEW_PROJECT, {
+  canRequestAccess: true,
+  getProjectId: (params) => params.id?.split('-')[0],
```
seanparsons, line 23: Probably worth extracting out this logic into a function, I can see the same logic being used 3 lines later.

utopia-remix/app/routes/p.$id.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { UserProjectPermission } from '~/types'
+import { type Params, redirect } from '@remix-run/react'
+
+export async function loader(args: LoaderFunctionArgs) {
+  return getProjectForEditor(args.request, args.params)
+}
+
+const validator = validateProjectAccess(UserProjectPermission.CAN_VIEW_PROJECT, {
+  canRequestAccess: true,
+  getProjectId: (params) => params.id?.split('-')[0],
```
seanparsons, line 13: ...and here again.

seanparsons (approved):

utopia-remix/app/components/projectNotFound.tsx
```diff
@@ -0,0 +1,124 @@
...
+  text: string
+  onClick?: () => void
+  disabled?: boolean
+}) {
+  return (
+    <Button
+      disabled={disabled}
+      onClick={onClick}
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
balazsbajorics, line 103: 🤨

utopia-remix/app/components/projectNotFound.tsx
```diff
@@ -0,0 +1,124 @@
...
+      css={{
```
ruggi, line 104: I think we decided not to use emotion at all in Remix 🤔 (and use Vanilla Extract/Sprinkles instead)

ruggi: (@maltenuhn should know more about the whole rationale)

utopia-remix/server.js
```diff
@@ -180,13 +180,16 @@ app.use('/authenticate', proxy)
...
 app.use('/logout', proxy)
-app.use('/p', proxy)
-app.use('/project', proxy)
 app.use('/share', proxy)
 app.use('/sockjs-node', proxy)
 app.use('/v1/javascript/packager', proxy)
 app.use('/vscode', proxy)
 
+// this will use the proxy only for `/p` and `/project` paths
+// for `/p/{id} and `/project/{id}` we will use Remix routes
```
balazsbajorics, line 189: thanks for the comment!

balazsbajorics (approved):

utopia-remix/app/components/projectNotFound.tsx
liady: yeah, TS is shouting because `css` is not a defined HTML property, but we know (and TS don't) that emotion uses this prop. I'll try find another solution for it (it's a known issue, but in the editor it doesn't happen)

utopia-remix/app/components/projectNotFound.tsx
liady: oh, ok, I took the page as is from the editor. I'll remove the emotion then

ruggi: yeah makes sense, we do use emotion in the editor but for how remix treats assets it does not play nicely with it https://remix.run/docs/en/main/styling/css-in-js

utopia-remix/app/components/projectNotFound.tsx
liady: fixed by removing emotion and sprinkles (since it doesn't play well with Radix anyway)

utopia-remix/app/components/projectNotFound.tsx
liady: removed emotion altogether

utopia-remix/app/components/projectNotFound.css.ts
```diff
@@ -0,0 +1,7 @@
+import { style } from '@vanilla-extract/css'
+
+export const hover = style({
+  ':hover': {
+    backgroundColor: '#5DA9FF !important',
```
ruggi, line 5: is `important` needed? (just to double check so we don't have to chase them)

utopia-remix/app/routes/p.$id.tsx
```diff
@@ -0,0 +1,30 @@
...
+})
+
+export async function getProjectForEditor(req: Request, params: Params<string>) {
+  const validatorResult = await validator(req, params)
+  if (validatorResult.ok) {
+    const proxyResponse: Response = (await proxy(req, {
+      rawOutput: true,
+    })) as Response
+    return new Response(proxyResponse.body, {
+      headers: { 'content-type': 'text/html' },
```
ruggi, line 24: it would be good to return all response headers here (except `host`, see the implementation in `proxy`), maybe overriding the content type. But isn't the content type already returned with the response headers?

utopia-remix/app/components/projectNotFound.css.ts
liady: unfotunately here yes, since the Radix classes override the vanilla-extract ones. I think this will be an issue that we'll need to solve in a more "strategic" level when we decide which design system we want to use and how we want to configure it

ruggi: 😢

utopia-remix/app/routes/p.$id.tsx
liady: I actually tried returning the `proxyResponse` itself but it didn't work. I'll try to maybe just use `handleMethod` logic here and see if I can unify both

ruggi (approved):

PR #5087, opened 2024-03-20
Internal property controls

bkrmendy:
## Problem
Consider the following `registerInteralComponent` call:
```typescript
registerInternalComponent(Chapter, {
  supportsChildren: false,
  properties: {
    title: {
      control: 'jsx',
      preferredChildComponents: [
        {
          name: 'Heading',
          additionalImports: 'src/ui-ib/heading',
          variants: [
            code: '<Heading>Chapter 1</Heading>
          ]
        }
      ]
    },
  },
  variants: []
}
```

To correctly insert the `Heading` element into the `title` prop, we need to insert the imports specified in the `additionalImports` corresponding to the `preferredChildComponents` in the `title` prop. However, in the `registerInternalComponent` schema, `PreferredChildComponent.additionalImports` is a string, so it needs to be parsed and turned into an `Imports` object that the editor can work with. This would be all well and good (we already do this for `ComponentToRegister.variants`, however, this isn't straightforward for `JSXControlDescription.preferredChildComponents`, since the types for the control definitions are shared between the editor and `utopia-api`. To have a parsed representation of `additionalImports` (and the actual elements to insert in `variants`), we need a different type for the JSX control in the editor (a type where `additionalImports` is `Imports`, instead of `string`).  This separation already exists for components passed to `registerInternalComponent` (`ComponentInsertOption` vs `ComponentDescriptor`).

## Fix

For now, duplicate the property control types in `utopia-api/src/property-control/property-controls.ts` to a new file, `editor/src/components/custom-code/internal-property-controls.ts`. In the editor, update all imports importing a property control to import the same thing from `internal-property-controls.ts`.

### DIFF
--- a/editor/src/core/property-controls/property-controls-local.ts
+++ b/editor/src/core/property-controls/property-controls-local.ts
@@ -7,7 +7,7 @@ import type {
   ComponentInsertOption,
   PreferredChildComponent,
 } from 'utopia-api/core'
-import type { PropertyControls } from '../../components/custom-code/internal-property-controls'
+import type { PropertyControls } from 'utopia-api/core'
 import type { ProjectContentTreeRoot } from '../../components/assets'
 import { packageJsonFileFromProjectContents } from '../../components/assets'
 import type {
@@ -106,7 +106,6 @@ async function componentDescriptorForComponentToRegister(
   const parsedVariantsUnsequenced = await Promise.all(parsedInsertOptionPromises)
   const parsedVariants = sequenceEither(parsedVariantsUnsequenced)
 
-  // TODO
   return mapEither((variants) => {
     return {
       componentName: componentName,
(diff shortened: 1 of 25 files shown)


editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -106,6 +106,7 @@ async function componentDescriptorForComponentToRegister(
   const parsedVariantsUnsequenced = await Promise.all(parsedInsertOptionPromises)
   const parsedVariants = sequenceEither(parsedVariantsUnsequenced)
 
+  // TODO
```
Rheeseyb, line 109: TODO what?

Rheeseyb (approved):

editor/src/core/property-controls/property-controls-local.ts
bkrmendy: forgot to push the commit that removes it https://github.com/concrete-utopia/utopia/pull/5087/commits/d0ae89e2a65ba204f048cb70e0cbc8e5ccac7c17

PR #5097, opened 2024-03-21
Remix projects: projects loader return pending requests flag

ruggi:
**Problem:**

The pending requests badge/info needs to be available for all projects, immediately.

**Fix:**

This is a PR required for incremental work.

1. Return the `hasPendingRequests` flag with the projects obtained via `listProjects`, and use that instead of checking the requests list
2. The type naming is getting tricky so, as part of this PR, as I think it should be part of **this** PR, the naming is adjusted a bit:
	- types related to Prisma DB data are marked with the `FromDB` suffix
	- the `ProjectListing` is used as a decoration type on top of that, so it includes the DB data _and_ the accessory data we add
	- the legacy `ProjectListing` used by Haskell is renamed to `ProjectListingV1`

Again I know the type renaming is a bit noisy but I think it should definitely be part of this PR as the scope is all here.

### DIFF
--- a/utopia-remix/app/routes/projects.tsx
+++ b/utopia-remix/app/routes/projects.tsx
@@ -12,15 +12,17 @@ import {
   TrashIcon,
 } from '@radix-ui/react-icons'
 import { Badge, Button, ContextMenu, DropdownMenu, Flex, Text, TextField } from '@radix-ui/themes'
-import { type LoaderFunctionArgs, json } from '@remix-run/node'
+import { json, type LoaderFunctionArgs } from '@remix-run/node'
 import { useFetcher, useLoaderData } from '@remix-run/react'
 import moment from 'moment'
 import type { UserDetails } from 'prisma-client'
 import React from 'react'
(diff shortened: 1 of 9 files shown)


utopia-remix/app/routes/projects.tsx
```diff
@@ -804,7 +804,7 @@ const ProjectCard = React.memo(
             </div>
           </div>
         </ContextMenu.Trigger>
-        <ProjectActionsMenu project={project} accessRequests={accessRequests} />
+        <ProjectActionsMenu project={project} />
         <SharingDialogWrapper project={project} accessRequests={accessRequests} />
```
liady, line 808: can we also move the `accessRequestFetcher` login inside `SharingDialogWrapper` (since now, after the change, it's the only place that uses it)

liady (approved):

gbalint (approved):

utopia-remix/app/routes/projects.tsx
ruggi: yep, coming in the next PR which does a larger set of improvements (including this one!)

PR #5099, opened 2024-03-21
Remix projects: more robust sharing dialog

ruggi:
**Problem:**

The sharing dialog has a prominent issue: the main one is that since the dialog is per-project, if a project from a category is moved to a different category (e.g. changing its access level) the dialog never gets cleaned up properly, leaving an invisible overlay on the page.

**Fix:**

The fix is to move the sharing dialog outside of the project components, and instead placing it in the main root of the projects page. It's still governed by the `sharingProjectId` property. To make everything work and have the logic less scattered, the access requests are fetched when the dialog is opened, from the context menu, placed in the store (non-persisted), and then consumed by the dialog component. This way the PR makes it also easy to give a better UX when loading a project, which does not have jumpy changes if/when the requests are loaded asynchronously.

https://github.com/concrete-utopia/utopia/assets/1081051/01fe3888-52ad-4cec-91ba-75f1c086600f


liady:
A general question - if the pending requests are displayed in async manner anyway - what is the reason to trigger the fetch from the context menu, and not the dialog itself?
(We can also have a small spinner there, until the requests arrive)

ruggi:
> A general question - if the pending requests are displayed in async manner anyway - what is the reason to trigger the fetch from the context menu, and not from the dialog component itself? (We can also have a small spinner there, until the requests arrive)

Because it's the only place where we have a user action that can trigger the fetch explicitly, without having to use a `useEffect` in the dialog.

seanparsons (approved):

bkrmendy (approved):

PR #5100, opened 2024-03-21
Restore cartouche delete and image preview tests

gbalint:
**Problem:**
These tests have been commented out for a while because they mysteriously caused dom walker tests to fail.
I removed them in https://github.com/concrete-utopia/utopia/pull/5070 , because there were lot of changes in this test file, and I did not want to increase the PR size by updating dead code.

However, these tests are valid and successful tests, so I updated them to check if they still break the dom walker tests.

Unfortunately, they do, but I still think they are valuable tests which we should fix later, so I readded their updated version and commented them out again.


balazsbajorics (approved):
:(

bkrmendy (approved):

PR #5101, opened 2024-03-21
Working towards a better render prop insertion popup

Rheeseyb:
**Problem:**
We want a better component picker for render props and for general insertion.

**Fix:**
This first pass improves the styling of the render prop picker, and also extracts the core into a component that can later be used inside the general Insert menu pane:

image

image

This is populated using the `preferredChildren` of registered components only. The "All Components" tab currently just shows the same preferred components. The model here needs a bit of adapting, as [`PreferredChildComponent`](https://github.com/concrete-utopia/utopia/blob/master/utopia-api/src/helpers/helper-functions.ts#L10-L14) differs substantially from the Insert Menu's [`InsertableComponent`](https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/shared/project-components.ts#L59-L66), so we'll need to fix that up before being able to support "All Components" (and use this picker inside the Insert menu)

As well as that, we want to highlight the topmost element that is already being used inside that prop, but that will come in a separate PR.

[**Try it here**](https://utopia.fish/p/2e3389de-emphasized-trumpet/?branch_name=feature-render-prop-insert-picker)


bkrmendy (approved):

editor/src/components/navigator/navigator-item/component-picker.tsx
```diff
@@ -0,0 +1,295 @@
...
+  const unfilteredComponentsToShow =
+    selectedTab === 'preferred' ? props.preferredComponents : props.allComponents
+  const componentsToShow =
+    filter.trim() === ''
+      ? unfilteredComponentsToShow
+      : unfilteredComponentsToShow.filter((v) =>
+          v.name.toLocaleLowerCase().includes(filter.toLocaleLowerCase().trim()),
+        )
+
+  return (
```
gbalint, line 65: In my opinion this returned jsx is way too long for a single component. Especially because there are good candidates in it which could be extracted into components, e.g. the view of the components themselves, maybe even the variants, or all the components under the search field together.

Rheeseyb: Yeah, that's fair. I have a couple of follow up tasks for another PR I'll be submitting later today, so I'll do a bit of refactoring here as part of that too

gbalint (approved):

PR #5104, opened 2024-03-22
Remove global property controls cache

gbalint:
**Problem:**
The new sidecar based register component solution uses much less global variables than the previous one, but it still uses a single global cache called `COMPONENTS_DESCRIPTOR_FILE_CACHE`.
However, this cache is very similar to `EditorState.propertyControlsInfo`, the only extra information it provides it that it stores the controls indexed by the descriptor file name. This is necessary, because we have to know which property controls have to be deleted/updated when a certain descriptor file changes.

**Fix:**
Include the source descriptor file information in EditorState.propertyControlsInfo, and use that directly instead of the global cache.

**Todo:**
- add tests for lifecycle management: register components using a descriptor file and then update that file and check the changes


editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -576,6 +579,7 @@ import type { CommentFilterMode } from '../../inspector/sections/comment-section
 import type { Collaborator } from '../../../core/shared/multiplayer'
 import type { MultiplayerSubstate } from './store-hook-substore-types'
 import type { PropertyControls } from '../../custom-code/internal-property-controls'
+import { string } from 'prop-types'
```
Rheeseyb, line 582: Oops!

Rheeseyb (approved):

PR #5106, opened 2024-03-22
Always include owners in shared with me

ruggi:
**Problem:**

The `listSharedWithMeProjectsAndCollaborators` logic might not return the owner as part of the collaborators, which should implicitly be part of that list.

**Fix:**

- Make sure owner data is patched onto the collaborators map if missing
- Rework the Prisma querying so it's still a single query but is clearer, and uses the collaborators table as the main source for the join to work around some Prisma building quirks
- Add a lot of comments explaining what the different pieces of the query do, so everyone can understand what's going on _and why_
- Adjust the tests accordingly

### DIFF
--- a/utopia-remix/app/models/project.server.ts
+++ b/utopia-remix/app/models/project.server.ts
@@ -166,32 +166,53 @@ export async function listSharedWithMeProjectsAndCollaborators(params: {
   projects: ProjectListing[]
   collaborators: CollaboratorsByProject
 }> {
-  // 1. grab the project IDs where the user is the collaborator,
-  // is not an owner, are not deleted, and have the COLLABORATIVE access.
-  const projects = await prisma.project.findMany({
+  // 1. grab the projects IDs for which the user is a collaborator, are not deleted, are
+  // collaborative, and the user is not the owner.
+  const projects = await prisma.projectCollaborator.findMany({
     where: {
-      ProjectAccess: { access_level: AccessLevel.COLLABORATIVE },
-      ProjectCollaborator: { some: { user_id: params.userId } },
-      owner_id: { not: params.userId },
-      OR: [{ deleted: null }, { deleted: false }],
+      user_id: params.userId,
+      Project: {
+        OR: [{ deleted: null }, { deleted: false }],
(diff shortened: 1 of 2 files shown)


utopia-remix/app/models/project.server.ts
```diff
@@ -168,30 +168,57 @@ export async function listSharedWithMeProjectsAndCollaborators(params: {
...
+    const projectId = project.proj_id
+
+    collaboratorsByProject[projectId] = project.ProjectCollaborator.map(({ User }) =>
+      userToCollaborator(User),
+    )
+
+    // the owner of a project should always appear in the list of collaborators, so
+    // make sure to append it if it's missing
+    const collaboratorsIncludeOwner = collaboratorsByProject[projectId].some(
+      (collaborator) => collaborator.id === project.owner_id,
```
balazsbajorics, line 211: why is it sometimes missing?

balazsbajorics (approved):

utopia-remix/app/models/project.server.ts
ruggi: simplest example is for existing (= old) projects that have not been opened by the owner

utopia-remix/app/models/project.server.ts
```diff
@@ -168,30 +168,57 @@ export async function listSharedWithMeProjectsAndCollaborators(params: {
...
+          ProjectCollaborator: { select: { User: true } },
+        },
+      },
     },
-    orderBy: { modified_at: 'desc' },
+    orderBy: { Project: { modified_at: 'desc' } },
+  })
+  const projects = projectsAndCollaborators.map(({ Project }) => Project)
+
+  // 2. grab the owner details in case some are missing from the collaborators
```
Rheeseyb, line 194: > in case some are missing from the collaborators

The owner details will always be missing from the collaborators, as they have been explicitly excluded from the query above

Rheeseyb: Oh sorry, I realise this is referencing the next step

Rheeseyb (approved):

PR #5113, opened 2024-03-25
404/403 design

liady:
Adjust the 404/403 page designs

PR #5115, opened 2024-03-25
Add a login route

liady:
This adds a new `/login` route to Remix, that calls the Auth0 url method in order to login and redirect if needed.
This is required for completion of #5113.

The route supports `/login?redirectTo={path}&fakeUser={alice|bob}`:

<video src="https://github.com/concrete-utopia/utopia/assets/7003853/fddd5ca3-4442-4546-aa8f-8f4d8b0f1e3c"/>

### DIFF
--- a/utopia-remix/app/routes-test/login.spec.ts
+++ b/utopia-remix/app/routes-test/login.spec.ts
@@ -3,17 +3,6 @@ import { handleLogin } from '../routes/login'
 import { ServerEnvironment } from '../env.server'
 
 describe('handleLogin', () => {
-  let corsOrigin: string
-  beforeAll(() => {
-    corsOrigin = ServerEnvironment.CORS_ORIGIN
-    // this is since in the tests we set CORS_ORIGIN to '*',
-    // so we can't rely on it in URLs
-    ServerEnvironment.CORS_ORIGIN = 'http://localhost:8000'
-  })
-  afterAll(() => {
-    ServerEnvironment.CORS_ORIGIN = corsOrigin
-  })
-
   it('should redirect to auth0 login with the correct redirectTo param', async () => {
     const request = newTestRequest({
       method: 'GET',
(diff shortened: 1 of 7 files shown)


utopia-remix/app/routes/login.tsx
```diff
@@ -0,0 +1,25 @@
...
+export async function loader(args: LoaderFunctionArgs) {
+  return handle(args, {
+    GET: {
+      validator: ALLOW,
+      handler: handleLogin,
+    },
+  })
+}
+
+export async function handleLogin(req: Request, params: Params<string>) {
```
ruggi, line 16: I think that if the user is already logged in this should do nothing and return immediately (redirecting to the `redirectTo` maybe)

utopia-remix/app/routes-test/login.spec.ts
```diff
@@ -0,0 +1,69 @@
+import { newTestRequest } from '../test-util'
+import { handleLogin } from '../routes/login'
+import { ServerEnvironment } from '../env.server'
+
+describe('handleLogin', () => {
+  let corsOrigin: string
+  beforeAll(() => {
+    corsOrigin = ServerEnvironment.CORS_ORIGIN
+    // this is since in the tests we set CORS_ORIGIN to '*',
```
ruggi, line 9: note: this is actually a leftover bug, we should have it already as `localhost:8000` (maybe it's worth doing a mini PR for this one!) (and it should not be named `CORS_ORIGIN` anymore but `REMIX_URL` or something like that)

ruggi (approved):

utopia-remix/app/routes/login.tsx
liady: so I think that usually in SaaS systems the `/login` route always takes you to the login page even if you are logged-in (for example if you want to change accounts, or have a new session). We can decide to do it differently here of course

ruggi: yeah, we should align with whatever pattern makes more sense at the moment – this is fine by me

Rheeseyb (approved):

PR #5118, opened 2024-03-26
Validate whether registered component exists

bkrmendy:
## Description
This PR adds a validation pass to the sidecar file registration process. With this PR, each registered component has to be imported into the sidecar file, and the validation pass performs the following checks:
- the imported component has to be "defined" (`!== undefined`, so if it's renamed or deleted, the stale registration info doesn't linger around)
- the registration key has to match the name of the imported component. For example,
```tsx
import { CardWithImage } from '/src/card-with-image'
...
const Components = {
  '/src/card/: {
    CardWithImage: {
      component: CardWithImage
       ...
     }
  }
}
```
the `CardWithImage` component has to be registered under the `CardWithImage` key (otherwise the registered controls won't show up for the actual `CardWithImage` component).

## Out of Scope
We can't yet tell the original name of 3rd party components, so in that case, `isComponentRegistrationValid` returns true.


editor/src/core/property-controls/property-controls-local.spec.tsx
```diff
@@ -1133,9 +1219,12 @@ describe('Lifecycle management of registering components', () => {
...
     it('Renaming a regular file to a component descriptor name adds the property controls', async () => {
-      const descriptorFileContent1 = `const Components = {
+      const descriptorFileContent1 = `import { Card } from '../src/card'
+      
+      const Components = {
       '/src/card': {
         Card: {
+          component: Card,
           supportsChildren: false,
           properties: {
```
gbalint, line 1229: I would add a new test case where the validation fails and the component is not registered.

gbalint (approved):

Rheeseyb (approved):

editor/src/core/property-controls/property-controls-local.spec.tsx
```diff
@@ -144,14 +167,81 @@ describe('registered property controls', () => {
...
+    }
+    
+    export default Components
+  `,
+      }),
+      'await-first-dom-report',
+    )
+    const editorState = renderResult.getEditorState().editor
+
+    // /src/card is not here
```
balazsbajorics, line 226: we could expect() this instead!

editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -173,6 +178,21 @@ export function createModuleEvaluator(editor: EditorState): ModuleEvaluator {
...
+): boolean {
+  if (typeof registration.component === 'undefined') {
+    return false
+  }
+
+  if (isComponentRendererComponent(registration.component)) {
+    return registration.component.originalName === registrationKey
+  }
+
+  return true
```
balazsbajorics, line 193: this will need to be re-written if we want to add a third check

editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -173,6 +178,21 @@ export function createModuleEvaluator(editor: EditorState): ModuleEvaluator {
...
+function isComponentRegistrationValid(
+  registrationKey: string,
+  registration: ComponentToRegister,
+): boolean {
+  if (typeof registration.component === 'undefined') {
+    return false
+  }
+
+  if (isComponentRendererComponent(registration.component)) {
+    return registration.component.originalName === registrationKey
```
balazsbajorics, line 190: this will need to be re-written if we want to add a third check (sorry, I wrote the comment on the wrong line before)

seanparsons (approved):

editor/src/core/property-controls/property-controls-local.ts
bkrmendy: We'll have to modify this function anyway when we add support for 3rd party components, and knowing that the function consists on 2 ifs and has full coverage it's safe to refactor

PR #5120, opened 2024-03-26
Parse and handle early returns.

seanparsons:
**Problem:**
We want components which have early return conditions to be manipulable by the editor at least when they are returning their regular result. Currently those components become big arbitrary blobs that are partly opaque to the editor.

**Fix:**
The arbitrary blocks that are present at the start of a component previously just ran and returned the "scope" created by that block, but we needed something that differentiates between the block running to the end with the scope and if a `return` was issued in the middle of the block that bails out of it early. This has resulted in an arbitrary block like this:
```typescript
if (props.hidden) {
  return null
}
const someValue = {
  a: 1
}
```
Being transpiled into this:
```typescript
return (() => {
  if (props.hidden) {
    return utopiaCanvasEarlyReturnResult(null)
  }
  const someValue = {
    a: 1
  }
  return utopiaCanvasBlockRanToEnd({
    someValue: someValue
  })
})()
```
Then when we run that code we can tell if an early return was triggered, or otherwise render as we used to updating the scope for the component.

As part of this the following changes were made:
- The parser for the arbitrary block got better at not creating empty arbitrary blocks which don't actually have any code in them.
- Two fixes relating to source map handling were made:
  - A bug fix where a `SourceMapGenerator` was being passed around with the value of `RawSourceMap`, resulting in an error when the source map was then parsed.
  - Replacing calls against the `SourceMapGenerator` of `JSON.parse(generator.toString())` with `generator.toJSON()` as that made much more sense.
- Disabled a test relating to exported default anonymous functions as it appears they weren't being handled correctly in master, so are probably better fixed in a separate piece of work.

**Notes:**
This has nothing user facing just yet as that would balloon this into a bit of a mega PR.

**Commit Details:**
- Added `ArbitraryBlockResult` type which distinguishes between the various ways to exit from an arbitrary block.
- `resolveParamsAndRunJsCode` now returns an instance of `ArbitraryBlockResult`.
- `runBlockUpdatingScope` now handles the new return type of `resolveParamsAndRunJsCode`, so that running to the end results in the scope being updated.
- Created `applyBlockReturnsFunctions` utility function for putting the 3 arbitrary block functions into scope.
- Updated more of the tests checking the transpiled code.
- Fixed a couple of places that were passing around a `SourceMapGenerator`
  instead of the JSON that it generates.
- Replaced some cases that were using `SourceMapGenerator.toString()` and
  then parsing the JSON, by calling `SourceMapGenerator.toJSON()` directly
  instead.
- Disabled a test for now so that the underlying issue can be fixed separately.
- Added `failOnNullResult` utility function for the parser.
- `parseOtherJavaScript` now permits a null value in the right side of the either.
- `parseArbitraryNodes` now supports not creating a blank arbitrary node which only has whitespace in it.
- Added `rewriteBlockToCaptureEarlyReturns` Babel plugin which turns the arbitrary blocks into ones that use our special functions to capture what sort of result there was.


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -220,6 +224,9 @@ export function createComponentRendererComponent(params: {
       highlightBounds: highlightBounds,
       editedText: rerenderUtopiaContext.editedText,
     }
+
+    const buildResult = React.useRef<React.ReactElement | null>(null)
```
balazsbajorics, line 228: do we ever read this? maybe we could just simply return the value from the `switch (arbitraryBlockResult.type)` without setting this ref

editor/src/core/property-controls/property-controls-local.spec.tsx
```diff
@@ -74,70 +74,7 @@ describe('registered property controls', () => {
...
-                  "importedWithName": null,
-                },
-              },
-              "insertMenuLabel": "ID Card",
-            },
-          ],
-        },
-      }
-    `)
+    expect(editorState.propertyControlsInfo['/src/card']).toMatchInlineSnapshot(`undefined`)
```
balazsbajorics, line 77: this seems like a regression to fix TODO

editor/src/core/shared/element-template.ts
```diff
@@ -2362,6 +2362,42 @@ export function createNotImported(path: string, variableName: string): ImportInf
...
+}
+
+export function arbitraryBlockRanToEnd(scope: MapLike<unknown>): ArbitraryBlockRanToEnd {
+  return {
+    type: 'ARBITRARY_BLOCK_RAN_TO_END',
+    scope: scope,
+  }
+}
+
+export type ArbitraryBlockResult = EarlyReturnVoid | EarlyReturnResult | ArbitraryBlockRanToEnd
```
balazsbajorics, line 2399: nice

editor/src/core/workers/parser-printer/__snapshots__/parser-printer-arbitrary-elements.spec.tsx.snap
```diff
@@ -97,22 +100,28 @@ export var storyboard = (
...
+    if (n <= 0) {
+      return 0;
+    } else {
+      return a(n - 1);
+    }
   }
-}
-return { a: a, b: b };",
+
+  return utopiaCanvasBlockRanToEnd({
```
balazsbajorics, line 120: ooh, this is very elegant!

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
seanparsons: The value of it is returned by the `ComponentRendererComponent` as what should be rendered. I wanted to avoid changing this too much, at least for this PR. I'll make a note to have a look at this separately as it might clean this up a little.

editor/src/components/editor/actions/actions.spec.tsx
```diff
@@ -408,6 +408,8 @@ describe('UPDATE_FILE_PATH', () => {
     })
     expect(filesAndTheirImports).toMatchInlineSnapshot(`
       Object {
+        "/package.json": Array [],
```
seanparsons, line 411: I'll have a little look into why these have been included all of a sudden, but they don't appear to be causing any harm.

editor/src/core/performance/__snapshots__/performance-regression-tests.spec.tsx.snap
```diff
@@ -2773,8 +2755,6 @@ Array [
   "/UtopiaSpiedFunctionComponent(View)/div/Symbol(react.forward_ref)(SpyWrapper)/UtopiaSpiedFunctionComponent(ComponentRenderer(App))",
   "/div/Symbol(react.forward_ref)(SpyWrapper)/UtopiaSpiedFunctionComponent(ComponentRenderer(App))/Symbol(react.forward_ref)(SpyWrapper)",
   "/div/Symbol(react.forward_ref)(SpyWrapper)/UtopiaSpiedFunctionComponent(ComponentRenderer(App))/Symbol(react.forward_ref)(SpyWrapper)",
-  "/div/Symbol(react.forward_ref)(SpyWrapper)/UtopiaSpiedFunctionComponent(ComponentRenderer(App))/Symbol(react.forward_ref)(SpyWrapper)",
```
balazsbajorics, line 2776: Sean do you know what change made the SpyWrapper re-render less?

editor/src/core/workers/parser-printer/parser-printer-uids.spec.tsx
```diff
@@ -113,7 +113,7 @@ describe('parseCode', () => {
     )
 
     const result = getAllUniqueUids(projectContents)
-    expect(result.uniqueIDs).toHaveLength(8)
+    expect(result.uniqueIDs).toHaveLength(7)
```
balazsbajorics, line 116: is this number going down because of the reduced number of empty arbitrary blocks?

balazsbajorics (approved):
this looks very nice!!

gbalint (approved):

editor/src/core/workers/parser-printer/parser-printer-uids.spec.tsx
seanparsons: Yeah, these are the arbitrary blocks that used to be a handful of spaces and a line feed.

editor/src/core/performance/__snapshots__/performance-regression-tests.spec.tsx.snap
seanparsons: This comes from the overly eager rendering from this which is now a null value in its new location, as it should always update on that first render anyway. Restoring it, makes these renders return to the regression test.

editor/src/components/editor/actions/actions.spec.tsx
seanparsons: These appear to be being added because the parser is getting a little further along rather than just bailing out immediately. As a result something like `package.json` has a parse success, but everything is effectively empty in that result.

editor/src/core/performance/__snapshots__/performance-regression-tests.spec.tsx.snap
seanparsons: Context: https://github.com/concrete-utopia/utopia/pull/5120/files#diff-9be29e1233edaec82e4b10de584ebe2839adf2e913bdb4e8a53a79bdb9bb6b23R228

balazsbajorics: okay!

PR #5123, opened 2024-03-26
Remix optimistic access change

ruggi:
**Problem:**

The sharing dialog jumps around when changing the access level.

**Fix:**

Apply optimistic updates to it when updating the project access level. In case of failures, the dialog will revert back to the current state.

https://github.com/concrete-utopia/utopia/assets/1081051/28137787-62da-4dd4-9809-34dea4dbff97

(delay added for dramatisation)


utopia-remix/app/components/sharingDialog.tsx
```diff
@@ -59,26 +60,40 @@ function SharingDialog({ project }: { project: ProjectListing | null }) {
   const setSharingProjectId = useProjectsStore((store) => store.setSharingProjectId)
   const accessRequests = useProjectsStore((store) => store.sharingProjectAccessRequests)
 
+  // keep track of possible new access levels to avoid jumpy single-frame updates
+  const [currentAccessLevel, setCurrentAccessLevel] = React.useState<AccessLevel | null>(null)
```
gbalint, line 64: Can we set to the real access level immediately? And then we don't need the nullability and the null checks.

Rheeseyb (approved):

utopia-remix/app/components/sharingDialog.tsx
```diff
@@ -59,26 +60,35 @@ function SharingDialog({ project }: { project: ProjectListing | null }) {
...
 
-  const clearSharingProjectId = React.useCallback(() => {
-    if (!projectAccessMatchesSelectedCategory) {
-      setSharingProjectId(null)
-    }
-  }, [setSharingProjectId, projectAccessMatchesSelectedCategory])
+  const changeAccessFetcherCallback = React.useCallback(
+    (data: unknown) => {
+      if (isLikeApiError(data)) {
+        setAccessLevel(projectAccessLevel)
```
liady, line 74: just to make sure - if the change call succeeds, the `projectAccessLevel` is also updated to the new access level (and doesn't stay in a stale mode)?

liady (approved):

utopia-remix/app/components/sharingDialog.tsx
ruggi: yeah – if the call succeeds the access level stay put on the one set there, so yeah it does not get stale (and if other components reference the project they receive its fresh data via the loader update)

PR #5126, opened 2024-03-27
Better error handling for component registration

bkrmendy:
image

## Problem
Errors that occur during component registration are not shown to users

## Fix
- Create a type that encodes component registration errors (`ComponentDescriptorRegistrationError`).
- Reuse the existing error overlay and the existing error types to surface the component registration errors. Since it only uses the interface provided by the `ComponentDescriptorRegistrationError` type, this code can easily be swapped out for a different error reporting mechanism in the future


editor/src/core/property-controls/property-controls-local.spec.tsx
```diff
@@ -233,6 +278,117 @@ describe('registered property controls', () => {
...
+            source: 'eslint',
+            startColumn: null,
+            startLine: null,
+            type: '',
+          },
+        ],
+      },
+    })
+
+    // /src/card is not here
```
balazsbajorics, line 326: I still think it would be nicer to expect() that /src/card is not in the array, instead of checking what _is_ in the array. 

if we add a new default builtin test dependency, all of these tests will fail

bkrmendy: makes sense, refactored the code in https://github.com/concrete-utopia/utopia/pull/5126/commits/eeace749ca3cf3e056d4163d89d2ee483eeb1c58

balazsbajorics (approved):

seanparsons (approved):

Rheeseyb (approved):

PR #5130, opened 2024-03-27
Remix: share dialog collaborators design pass

ruggi:
This PR aligns the share dialog a bit more with the intended design.

1. Show the owner in the collaborators list at the bottom of the share dialog
2. Show the collaborator avatars on the left side of the rows
3. If the project has gone private, show a disclaimer line and grey out the non-owner collaborator rows
4. Make the collaborator rows reactive to the optimistic access level implemented earlier
5. If the project is not collaborative, don't show the approve button at all for pending requests but show `pending` instead

https://github.com/concrete-utopia/utopia/assets/1081051/7ac0fe9e-1063-4535-9bb5-8854d256ca52

Note: the operations have been changed to just accept a project ID string instead of the whole object since 1) we don't need it and 2) it makes it easier to use those operations with just that value (e.g. how `AccessRequestsList` creates the new approval operation).


balazsbajorics (approved):

Rheeseyb (approved):

utopia-remix/app/components/sharingDialog.tsx
```diff
@@ -133,69 +134,128 @@ function SharingDialog({ project }: { project: ProjectListing | null }) {
...
+const OwnerCollaboratorRow = React.memo(() => {
+  const myUser = useProjectsStore((store) => store.myUser)
+  if (myUser == null) {
+    return null
+  }
+
+  return (
+    <CollaboratorRow
+      picture={myUser?.picture ?? null}
+      name={`${myUser?.name ?? myUser?.email ?? myUser?.id} (you)`}
```
seanparsons, line 226: Should this have a final default value? I think it can end up with `null (you)` right?

seanparsons (approved):

utopia-remix/app/components/sharingDialog.tsx
ruggi: nope, in the _worstest_ case it will be your user id

ruggi: (those `?` are actually incorrect since the not-null is checked above, will make a PR to clean up)

PR #5131, opened 2024-03-27
(chore) CDN link helpers

ruggi:
This PR does a bit of minor cleanup as a followup to https://github.com/concrete-utopia/utopia/pull/5130 and https://github.com/concrete-utopia/utopia/pull/5129.

1. Add a helper function and hook to generate CDN links
2. Remove unnecessary (and confusing) question marks from the user fields used in `OwnerCollaboratorRow`

### DIFF
--- a/utopia-remix/app/util/cdnLink.tsx
+++ b/utopia-remix/app/util/cdnLink.tsx
@@ -1,0 +1,17 @@
+import urlJoin from 'url-join'
+import { useAppStore } from '../stores/appStore'
+import React from 'react'
+
+export function cdnLink(host: string | null, relativePath: string): string {
+  return urlJoin(host ?? '', relativePath)
+}
+
+export function useCDNLink() {
+  const env = useAppStore((store) => store.env)
+  return React.useCallback(
+    (relativePath: string) => {
+      return cdnLink(env?.UTOPIA_CDN_URL ?? null, relativePath)
+    },
+    [env],
+  )
+}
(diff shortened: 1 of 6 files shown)


seanparsons (approved):

utopia-remix/app/util/cdnLink.tsx
```diff
@@ -0,0 +1,17 @@
+import urlJoin from 'url-join'
+import { useAppStore } from '../stores/appStore'
+import React from 'react'
+
+export function cdnLink(host: string | null, relativePath: string): string {
+  return urlJoin(host ?? '', relativePath)
+}
+
+export function useCDNLink() {
```
liady, line 9: Is it possible to have the API as `useCDNLink(filename)` directly?

ruggi: this way this is just a reusable function that can be called indefinitely.
what would be the benefit of the opposite?

balazsbajorics: if `env` is in the global scope then we don't need the useStore hook then we don't need the useCDNLink hook. if env is _not_ in the global scope, then we need all this

ruggi: there are two variants: a general purpose function that receives a base host (for when you don't have access to the `env`, and the hook variant for easy use inside components, that have access to the context that provides the `env` data. I think what @liady was asking for is a version of the hook that doesn't return a callback but rather just the string from a relative path passed to the hook itself

```ts
const theLink = useCDNLink("/foo.png")
```

But I think having a callable function makes more sense in the hook scenario

liady: yes, was just proposing that it might be easier to read to do it in one step - since really the only reason it's a hook is because we rely on the AppStore, but the extra step of returning a function might be unnecessary. But it's really minor here :)

liady (approved):

PR #5135, opened 2024-03-28
Split project creation methods

liady:
**Problem:**
Today creation of a new project and saving an existing one both call the PUT method on `/project/{id}`, which makes it difficult to handle creation of access levels upon project creation

**Fix:**
This prep PR splits the methods so that the PUT endpoint will have the access level handling code, while the POST method will not. This will allow us to split the calling code in the editor as well

### DIFF
--- a/utopia-remix/app/routes-test/v1.project.$id.spec.ts
+++ b/utopia-remix/app/routes-test/v1.project.$id.spec.ts
@@ -140,7 +140,7 @@ describe('create new project', () => {
     const userId = 'user1'
 
     let projectProxy: jest.SpyInstance
-    let setProjectAccessMock: jest.SpyInstance
+    let createProjectAccessMock: jest.SpyInstance
     afterEach(async () => {
       await truncateTables([
         prisma.projectAccess,
@@ -152,7 +152,7 @@ describe('create new project', () => {
       ])
 
       projectProxy.mockClear()
-      setProjectAccessMock.mockClear()
+      createProjectAccessMock.mockClear()
     })
 
     beforeEach(async () => {
@@ -160,12 +160,12 @@ describe('create new project', () => {
       await createTestSession(prisma, { key: 'the-key', userId: userId })
 
       projectProxy = jest.spyOn(proxyServer, 'proxy')
-      setProjectAccessMock = jest.spyOn(projectAccessModel, 'setProjectAccess')
+      createProjectAccessMock = jest.spyOn(projectAccessModel, 'createProjectAccess')
     })
 
     it('should set access level for a new project', async () => {
       projectProxy.mockResolvedValue({ id: projectId, ownerId: userId })
-      setProjectAccessMock.mockResolvedValue(null)
+      createProjectAccessMock.mockResolvedValue(null)
       const req = newTestRequest({
         method: 'PUT',
         authCookie: 'the-key',
@@ -178,14 +178,14 @@ describe('create new project', () => {
       }) as Promise<ApiResponse<{ id: string; projectId: string }>>)
       const project = await response.json()
       expect(project).toEqual({ id: projectId, ownerId: userId })
-      expect(setProjectAccessMock).toHaveBeenCalledWith({
+      expect(createProjectAccessMock).toHaveBeenCalledWith({
         projectId: projectId,
         accessLevel: AccessLevel.PUBLIC,
       })
     })
-    it('shouldnt set access level for a new project if not provided', async () => {
+    it('should set access level to PRIVATE for a new project if not provided', async () => {
       projectProxy.mockResolvedValue({ id: projectId, ownerId: userId })
-      setProjectAccessMock.mockResolvedValue(null)
+      createProjectAccessMock.mockResolvedValue(null)
       const req = newTestRequest({
         method: 'PUT',
         authCookie: 'the-key',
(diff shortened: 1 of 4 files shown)


utopia-remix/app/routes/v1.project.$id.tsx
```diff
@@ -20,8 +23,47 @@ export async function loader(args: LoaderFunctionArgs) {
...
+  // handle access level setting
+  const url = new URL(req.url)
+  const accessLevelParam = url.searchParams.get('accessLevel')
+  const accessLevel: AccessLevel | null = accessLevelParamToAccessLevel(accessLevelParam)
+  if (accessLevel != null) {
+    await setProjectAccess({ projectId: id, accessLevel: accessLevel })
+  }
+
+  return project
+}
```
ruggi, line 52: I think this should _always_ create the project access, with a default to `PRIVATE`. A project should _always_ have an access (and we should address the nullability separately, but the concept stands).
Of course if the `accessLevel` param is passed, it overrides the default (if it has a valid value).

utopia-remix/app/routes/v1.project.$id.tsx
```diff
@@ -20,8 +23,47 @@ export async function loader(args: LoaderFunctionArgs) {
...
+      validator: ALLOW,
+    },
+    POST: {
       handler: proxy,
       validator: ALLOW,
     },
   })
 }
+
+async function createProject(req: Request, params: Params<string>) {
```
ruggi, line 36: Note: this currently is not transactional, but I think we can address it later.

utopia-remix/app/routes/v1.project.$id.tsx
```diff
@@ -20,8 +23,47 @@ export async function loader(args: LoaderFunctionArgs) {
...
+  const accessLevelParam = url.searchParams.get('accessLevel')
+  const accessLevel: AccessLevel | null = accessLevelParamToAccessLevel(accessLevelParam)
+  if (accessLevel != null) {
+    await setProjectAccess({ projectId: id, accessLevel: accessLevel })
+  }
+
+  return project
+}
+
+function accessLevelParamToAccessLevel(accessLevelParam: string | null): AccessLevel | null {
```
ruggi, line 54: would it make sense to just pass the same integer values of the `AccessLevel` enum instead? So we'd have to do less parsing (and it can be put in the request body rather than the query params) 🤔

liady: the request body is the project data itself, and since the request is being passed "as-is" to the proxy (and I know there is a type validation on the project data there) I was not sure that I could add arbitrary fields directly in the body. If we can tweak it somehow then it's better of course.
As for the enum vs string - I thought it's better for the client to not know the internal enum representation of the access levels, but rather use a descriptive easy-to-read strings, do you think we need to opt for the enum values here?
btw in any case we would have to do some parsing in order to "cast" the number to the AccessLevel type (like `asAccessLevel` in `types.ts`)

utopia-remix/app/routes/v1.project.$id.tsx
liady: yes, I think ideally we would want to proxy code (actual server code) to create the project and set the access level in the same transaction

utopia-remix/app/routes/v1.project.$id.tsx
liady: ok, will do. It means though that if someone is calling PUT on a public project by accident (without setting access level) it will be reverted back to PRIVATE

ruggi: I don't think we should – the call should be an upsert with a `DO NOTHING` if the access level already exists, because we have separate endpoints for changing the access level (this is just metadata bootstrapping)

utopia-remix/app/routes/v1.project.$id.tsx
ruggi: oooh you're right, the body is the project data! nvm, nvm. And the reasoning for the value makes sense, up to you :+1:

seanparsons (approved):

utopia-remix/app/routes/v1.project.$id.tsx
liady: changed it to default to PRIVATE and on conflict do nothing

utopia-remix/app/routes-test/v1.project.$id.spec.ts
```diff
@@ -119,3 +120,125 @@ describe('getProject', () => {
...
+        context: {},
+      }) as Promise<ApiResponse<{ id: string; projectId: string }>>)
+      const project = await response.json()
+      expect(project).toEqual({ id: projectId, ownerId: userId })
+      expect(createProjectAccessMock).toHaveBeenCalledWith({
+        projectId: projectId,
+        accessLevel: AccessLevel.PRIVATE,
+      })
+    })
+    it('should set access level to PRIVATE for a new project if access level is invalid', async () => {
```
Rheeseyb, line 206: I think this is ok, but I'm wondering if we should hard fail the request immediately if the access level is invalid

Rheeseyb (approved):

utopia-remix/app/routes-test/v1.project.$id.spec.ts
seanparsons: Probably better that than creating things with the "wrong" access level.

ruggi (approved):

PR #5137, opened 2024-03-28
Component descriptor validation for external components

gbalint:
**Problem:**
https://github.com/concrete-utopia/utopia/pull/5118 introduced component validation in component descriptor parsing.
However, that solution only worked for internal components, and this PR extends the validation to external components too.

**Fix:**
We use a custom require function when we parse the component descriptor file. I extended this require function so it attached two new symbol properties to the exported object/function: the exported name of the component and where it was imported from.
These new properties can be used in `isComponentRegistrationValid` to compare the name of the imported component and the key in the descriptor file.

**Todo:** 
Validate the module name too!

### DIFF
--- a/editor/src/core/property-controls/property-controls-local.ts
+++ b/editor/src/core/property-controls/property-controls-local.ts
@@ -125,11 +125,13 @@ export function setRequireInfoOnComponent(exported: any, name: string, moduleNam
 
 function extendExportsWithInfo(exports: any, toImport: string): any {
   Object.entries(exports).forEach(([name, exp]) => {
-    try {
-      ;(exp as any)[exportedNameSymbol] = name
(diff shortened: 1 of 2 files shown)


bkrmendy (approved):

editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -161,6 +195,10 @@ export function createModuleEvaluator(editor: EditorState): ModuleEvaluator {
         },
         resolvedParseSuccess,
       )
+      const absoluteFilenameOrPackage = isRight(filePathResolveResult)
```
seanparsons, line 198: I believe this should work:
```suggestion
      const absoluteFilenameOrPackage = defaultEither(toImport, filePathResolveResult)
```

seanparsons (approved):

PR #5148, opened 2024-03-31
Dont call set project access if not provided

liady:
**Problem:**
Currently our editor calls PUT `/project/project-id` for every project save. We have a code in the PUT method that sets the access level to PRIVATE if not provided - and while it knows not to override it in the DB - it sets it in FGA on every save.

**Fix:**
This is a TEMP fix - change the route code not to set the access level if not provided - and check in the `createProjectAccess` method that the project access is indeed empty.

The proper fix will be to have the editor not call the PUT endpoint on every project save

### DIFF
--- a/utopia-remix/app/models/projectAccess.server.ts
+++ b/utopia-remix/app/models/projectAccess.server.ts
@@ -29,6 +29,16 @@ export async function createProjectAccess(params: {
   projectId: string
   accessLevel: AccessLevel
 }): Promise<void> {
+  // check if access level already exists
+  const projectAccess = await prisma.projectAccess.findUnique({
+    where: {
+      project_id: params.projectId,
+    },
+  })
+  if (projectAccess != null) {
+    console.error(`Project access already exists for project ${params.projectId}`)
+    return
+  }
   await prisma.$transaction(async (tx) => {
     await tx.projectAccess.upsert({
       where: {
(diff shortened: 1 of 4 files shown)


utopia-remix/app/routes/v1.project.$id.tsx
```diff
@@ -44,9 +44,10 @@ async function createProject(req: Request, params: Params<string>) {
   // handle access level setting
   const url = new URL(req.url)
   const accessLevelParam = url.searchParams.get('accessLevel')
-  const accessLevel: AccessLevel =
-    accessLevelParamToAccessLevel(accessLevelParam) ?? AccessLevel.PRIVATE
-  await createProjectAccess({ projectId: id, accessLevel: accessLevel })
+  const accessLevel: AccessLevel | null = accessLevelParamToAccessLevel(accessLevelParam)
+  if (accessLevel != null) {
```
liady, line 48: this is so that we won't run `createProjectAccess` on every project save.

balazsbajorics (approved):

liady:
hotfixing since currently every project change makes it private again

utopia-remix/app/models/projectAccess.server.ts
```diff
@@ -29,6 +29,16 @@ export async function createProjectAccess(params: {
   projectId: string
   accessLevel: AccessLevel
 }): Promise<void> {
+  // check if access level already exists
+  const projectAccess = await prisma.projectAccess.findUnique({
```
ruggi, line 33: this should be / have been put inside the transaction below. Moreover with this check in place (if this is the implementation way we want to go for), the upsert is unnecessary and can just be a `create` when `projectAccess == null`.

liady: good point, changed in #5156

PR #5150, opened 2024-03-31
Project creation endpoint

liady:
> This is a prep PR for allowing the editor to call a dedicated route for creating a new project (instead of the same route for saving a project)

This PR creates a new endpoint - `PUT /v1/project` - that accepts this structure:
```
  name: 'new-project-name',
  content: {} // new project content,
  accessLevel?: 'public' | 'private' | 'collaborative'
```
It removes the creation logic from the old `PUT /v1/project/{project-id}`, and moves it to a dedicated endpoint that does the following:
* generates a new id
* creates a new project with this id
* assigns access level
* sets the user as the creator in fga
* returns the project id and the owner id (similar response to the current call).


utopia-remix/app/routes/v1.project._index.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+function withPath(req: Request, path: string): URL {
+  const url = new URL(req.url)
+  url.pathname = path
+  return url
+}
+
+async function createProjectId(req: Request) {
+  // const response = await fetch(req)
```
liady, line 87: I'm actually not sure if it's better to use here the internal method call or the fetch call. using the internal call saves us an HTTP request, but exposes internal implementation.

utopia-remix/app/routes/v1.project._index.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+}
+
+function withPath(req: Request, path: string): URL {
+  const url = new URL(req.url)
+  url.pathname = path
+  return url
+}
+
+async function createProjectId(req: Request) {
```
seanparsons, line 86: Probably better to give this a return type rather than casting the result at the callsite.

seanparsons (approved):

utopia-remix/app/routes/v1.project._index.tsx
```diff
@@ -0,0 +1,104 @@
...
+    headers: req.headers,
+  })
+  const projectIdResponse = (await createProjectId(projectIdRequest)) as { id: string }
+  const { id } = projectIdResponse
+
+  // prepare data for creating the project
+  const requestData = (await req.json()) as { content: object; name: string; accessLevel: string }
+  const { content, name, accessLevel: accessLevelParam } = requestData
+  const headers = new Headers(req.headers)
+  headers.delete('content-length')
```
ruggi, line 40: how come this needs to be deleted?

utopia-remix/app/routes/v1.project._index.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+async function createProject(req: Request, params: Params<string>) {
+  // const response = await fetch(req)
+  const response = (await createProjectInternal({
+    request: req,
+    params: params,
+    context: {},
+  })) as Response
+  return await response.json()
+}
```
ruggi, line 104: These two functions are equivalent, I don't think there's a need to have 2, just one helper that does the given request and returns its json response (with a generic function name)

ruggi: I think that calling the actions directly is not great – I'd suggest to either call the handlers themselves, or do a proper HTTP call with `fetch`.

liady: actually since we'll be probably removing `projectid` and `PUT /project/{id}` in the near future, maybe it's better to even call `proxy` directly here. changed.

utopia-remix/app/routes/v1.project._index.tsx
liady: good point! but now changed it to call the `proxy` directly.

bkrmendy (approved):

PR #5152, opened 2024-04-01
Fix right click on projects

liady:
This PR fixes the behavior where right click on a project deselects it

### DIFF
--- a/utopia-remix/app/routes/projects.tsx
+++ b/utopia-remix/app/routes/projects.tsx
@@ -692,6 +692,16 @@ const ProjectCard = React.memo(
       }
     }, [openShareDialog, canOpenShareDialog])
 
+    const onMouseDown = React.useCallback(
+      (event: React.MouseEvent) => {
+        // on right click only allow selection (not-deselction)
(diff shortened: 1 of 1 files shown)


utopia-remix/app/routes/projects.tsx
```diff
@@ -720,7 +730,7 @@ const ProjectCard = React.memo(
                 position: 'relative',
                 filter: project.deleted === true ? 'grayscale(1)' : undefined,
               }}
-              onMouseDown={onSelect}
```
balazsbajorics, line 723: does it mean calling `onSelect()` somehow leads to a de-select?

liady: Yes, it's actually toggling the selection (maybe should've been called `toggleSelection`)

seanparsons (approved):

ruggi (approved):

utopia-remix/app/routes/projects.tsx
balazsbajorics: sorry I missed this! could we rename this to toggleSelection then?

PR #5156, opened 2024-04-02
(refactor) insert project access check into transaction

liady:
Minor - insert the project access null check into the transaction, and change to `create` (as per @ruggi  comment from #5148 )


utopia-remix/app/models/projectAccess.server.ts
```diff
@@ -29,28 +29,24 @@ export async function createProjectAccess(params: {
...
-    },
-  })
-  if (projectAccess != null) {
-    console.error(`Project access already exists for project ${params.projectId}`)
-    return
-  }
   await prisma.$transaction(async (tx) => {
-    await tx.projectAccess.upsert({
+    // check if project access already exists
+    const projectAccess = await prisma.projectAccess.findUnique({
```
ruggi, line 34: this needs to use `tx`

```suggestion
    const projectAccess = await tx.projectAccess.findUnique({
```

ruggi (approved):

bkrmendy (approved):

PR #5157, opened 2024-04-02
(mini pr) no test should change the project URL

balazsbajorics:
We previously only prevented the URL rewriting if we were in a _debug_ test environment, but I realized there's no reason to ever rewrite the URL in _any_ test environment. 

This makes debugging Karma tests easier if we just append the `--debug` flag to any npm script


ruggi (approved):

Rheeseyb (approved):
Good catch!

PR #5158, opened 2024-04-02
Show selection only on right click

liady:
Instead of showing selection on left-clicks (which can be confusing) - we're showing the blue "selected" highlight only when a context menu is opened.

<video src="https://github.com/concrete-utopia/utopia/assets/7003853/c8371344-cd0b-4f14-90ed-3a612ad4d057" />

### DIFF
--- a/utopia-remix/app/routes/projects.tsx
+++ b/utopia-remix/app/routes/projects.tsx
@@ -246,7 +246,6 @@ const Sidebar = React.memo(({ user }: { user: UserDetails }) => {
   const setSearchQuery = useProjectsStore((store) => store.setSearchQuery)
   const selectedCategory = useProjectsStore((store) => store.selectedCategory)
   const setSelectedCategory = useProjectsStore((store) => store.setSelectedCategory)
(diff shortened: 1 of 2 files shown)


utopia-remix/app/routes/projects.tsx
```diff
@@ -698,18 +679,12 @@ const ProjectCard = React.memo(
...
-    const onMouseDown = React.useCallback(
-      (event: React.MouseEvent) => {
-        // on right click only allow selection (not de-selction)
-        if (event.button !== 2 || !selected) {
-          onSelect()
-        }
-      },
-      [onSelect, selected],
-    )
+    const toggleContextMenuOpen = React.useCallback((isOpen: boolean) => {
```
balazsbajorics, line 682: this doesn't look like a toggle to me!

liady: Hmm, I've always thought of "toggle" as also "toggle on" or "toggle off" and not just "change to the opposite of the current state"

PR #5159, opened 2024-04-02
Expose Early Return Values.

seanparsons:
**Problem:**
We momentarily capture early return values from components, but there's no way to see what they were after the fact.

**Fix:**
Our metadata is now slightly extended to include an early return value, so that it can be stored at the point where it's captured during the rendering of the component. The type of that field is `EarlyReturn | null` so that it's possible to capture the case where the early return itself is null.

There was also a small fix to `insertDataUIDsIntoCode` to make it try a series of "wrapping" approaches when processing the code. This was done because it can be quite situational which is best to try, so having particular call sites picking and choosing didn't make much sense.

**Notes:**
This does not represent the early returns in the UI, that will be in some follow up work.

**Commit Details:**
- In `createComponentRendererComponent`, call `addFakeSpyEntry` when an early return value is returned from a component.
- Added an `EarlyReturn` type to cover the two types of early returns.
- Remove the early returning code when an early return comes into `createComponentRendererComponent`.
- Added `earlyReturn` field to `ElementInstanceMetadata`.
- Reworked `insertDataUIDsIntoCode` to try the various approaches for wrapping the expression instead of the previous style of specifying but having a single alternative approach.

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer-transpiling.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer-transpiling.ts
@@ -532,48 +532,46 @@ export function insertDataUIDsIntoCode(
   code: string,
   map: RawSourceMap,
   elementsWithin: ElementsWithinInPosition,
-  wrapInParens: boolean,
   rootLevel: boolean,
   filename: string,
 ): Either<string, CodeWithMap> {
-  if (wrapInParens) {
-    const wrappedInParensResult = insertDataUIDsIntoCodeInner(
-      sourceFileName,
-      sourceFileText,
-      code,
-      map,
-      elementsWithin,
-      'wrap-in-parens',
-      rootLevel,
-      filename,
-    )
-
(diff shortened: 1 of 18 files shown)


editor/src/core/workers/parser-printer/parser-printer-transpiling.ts
```diff
@@ -532,48 +532,46 @@ export function insertDataUIDsIntoCode(
...
+    sourceFileText,
+    code,
+    map,
+    elementsWithin,
+    'wrap-in-parens',
+    rootLevel,
+    filename,
+  )
+  if (isRight(wrappedInParensResult)) {
+    return wrappedInParensResult
```
balazsbajorics, line 549: I like this combination of eithers and early returns!

balazsbajorics (approved):
superb!

bkrmendy (approved):

PR #5160, opened 2024-04-03
Change "trash" to "archive"

liady:
Fix a small issue where the string "trash" needed to change to "archive"

### DIFF
--- a/utopia-remix/app/util/use-sort-compare-project.tsx
+++ b/utopia-remix/app/util/use-sort-compare-project.tsx
@@ -57,7 +57,7 @@ export function useProjectIsOnActiveOperation() {
           case 'private':
           case 'sharing':
             return op.type === 'delete' && op.projectId === project.proj_id
-          case 'trash':
+          case 'archive':
             return (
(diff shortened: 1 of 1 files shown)


ruggi (approved):

utopia-remix/app/util/use-sort-compare-project.tsx
```diff
@@ -57,7 +57,7 @@ export function useProjectIsOnActiveOperation() {
           case 'private':
           case 'sharing':
             return op.type === 'delete' && op.projectId === project.proj_id
-          case 'trash':
+          case 'archive':
```
ruggi, line 60: it'd be nice to make this type safe with a guard

liady: interestingly, it threw an error but wasn't caught in the CI
image

ruggi: o.O

bkrmendy (approved):

PR #5162, opened 2024-04-03
Sidecar file docs

bkrmendy:


utopia-api/component-registration.md
```diff
@@ -0,0 +1,94 @@
...
+          startingWidth: Utopia.numberControl(),
+          incrementSize: Utopia.numberControl(),
+          placeholderWidth: Utopia.numberControl(),
+        }),
+      },
+
+      // `supportsChildren` tells Utopia whether this component uses the `children` prop or not.
+      // This makes Utopia's Navigator and canvas reparenting features work better with this component
+      supportsChildren: false,
+
```
bkrmendy, line 63: I left out `preferredChildComponents` because we want to re-think that for sure

PR #5165, opened 2024-04-03
Arbitrary block entries.

seanparsons:
**Problem:**
Arbitrary blocks are big nebulous blobs of code to the editor, so any detailed understanding is effectively impossible.

**Fix:**
This is the first step which breaks the arbitrary block into a series of statements. Our arbitrary block types get a field of statements which contains an array of entries for each statement in the block, which holds the code and the parsed value of what is defined in there and what is defined elsewhere.

**Commit Details:**
- Added `JSArbitraryStatement` type.
- Added `statements` field to `JSExpressionOtherJavaScript`.
- Added `statements` field to `ArbitraryJSBlock`.
- Extracted out `identifyValuesDefinedInNode`.
- `parseOtherJavaScript` now optionally parses each individual statement, where the optionality also prevents infinite recursion.


balazsbajorics (approved):
🎉

PR #5166, opened 2024-04-03
Create project with different access levels (client side)

liady:
Allow creating projects with different access levels (client side couterpart to #5150).

This PR splits the initial creating state of the state machine:
- If the user is anonymous - the flow remains the same (this code will be removed since we won't support anonymous users)
- If the user is logged in - the new `PUT /project` endpoint is being called and is being passed the access level, the project id and project itself is created behind the scenes (no need for two http calls), and the user is assigned as the creator in FGA.

<video src="https://github.com/concrete-utopia/utopia/assets/7003853/27b6d9e5-0f15-49c0-beec-d751a720000b"/>


editor/src/components/editor/persistence/generic/dummy-persistence-backend.ts
```diff
@@ -24,6 +25,14 @@ function loadProject<ModelType>(projectId: string): Promise<ProjectLoadResult<Mo
   })
 }
 
+function createNewProjectInServer<ModelType, FileType>(
+  projectModel: ProjectModel<ModelType>,
+): Promise<ProjectCreationResult<ModelType, FileType>> {
+  return Promise.resolve(
+    projectCreationResult(`Project_${projectCounter++}`, projectWithFileChanges([], projectModel)),
```
ruggi, line 32: instead of doing this replacement I would instead await `getNewProjectId`

editor/src/components/editor/url-utils.ts
```diff
@@ -0,0 +1,7 @@
+export function readParamFromUrl(param: string): string | null {
```
ruggi, line 1: can we explicitly state that this is not only reading the param from the URL but also removing it from the current window location, since this function call is not idempotent? (or just do the reading in this function and do the window replacement at the caller's site)

balazsbajorics (approved):

balazsbajorics (approved):

ruggi (approved):

PR #5168, opened 2024-04-04
Fix inserting empty element with no additional imports

bkrmendy:
## Problem
if the `(empty)` option is picked from the render prop picker, an error is thrown (`DependencyNotFoundError: Could not find 'undefined` relative to '...'`)

## Cause
Due to how string interpolation works, if a render prop's `preferredChild` doesn't have `additionalImports` filled out, the code adds an `import {...} from 'undefined'` import line

## Fix
Check whether `additionalImports` filled out

### DIFF
--- a/editor/src/components/navigator/navigator.spec.browser2.tsx
+++ b/editor/src/components/navigator/navigator.spec.browser2.tsx
@@ -5025,6 +5025,47 @@ describe('Navigator', () => {
   })
 
   describe('render props', () => {
+    it('can insert an empty intrinsic element into render prop', async () => {
+      const renderResult = await renderTestEditorWithModel(
+        projectWithRenderProp(''), // <- no render prop
+        'await-first-dom-report',
+      )
+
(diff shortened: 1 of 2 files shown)


editor/src/components/navigator/navigator.spec.browser2.tsx
```diff
@@ -5041,6 +5082,8 @@ describe('Navigator', () => {
       )
       await mouseClickAtPoint(renderPropOptionElement, { x: 3, y: 3 })
 
+      await wait(100000)
```
balazsbajorics, line 5085: hello!

bkrmendy: 😬 fixed https://github.com/concrete-utopia/utopia/pull/5168/commits/80a6148392038aab747f994a1d93553c54cc1829

balazsbajorics (approved):

Rheeseyb (approved):

PR #5171, opened 2024-04-04
Simple picker for preferred components

Rheeseyb:
**Problem:**
Our current render prop picker is too much when there are only a few defined "preferred" options.

**Fix:**
Introduce a simple version, with a "More..." button to show the fuller version
image

We still need to add a heuristic to only show this simpler option if there are only a few preferred options, as well as offer _all_ components in the fuller version, but that is to come later.

One thing I want to draw attention to here is that both pickers are using `react-contexify`, with a rather nasty hack for rendering the fuller version in place of the simple version (using a ref to capture the location of the first menu so that we can use that when opening the second menu). It's gross, and I'm not proud of it, but this is working around the fact that clicking a button will close the menu, and that we don't want to open the second menu using the mouse position when clicking the "More..." button


bkrmendy (approved):

editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
```diff
@@ -63,27 +74,147 @@ const usePreferredChildrenForTargetProp = (
...
+        <div
+          style={{
+            width: 18,
+            height: 18,
+            display: 'flex',
+            justifyItems: 'center',
+            alignItems: 'center',
+            position: 'relative',
+          }}
+        ></div>{' '}
```
seanparsons, line 163: What's with the weird space?

seanparsons (approved):

editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
gbalint: I suspect it is a prettier artifact.

Rheeseyb: Urgh yes. I'll remove it.

PR #5174, opened 2024-04-04
Editor share dialog

### LINKED ISSUE #5175: Editor share dialog (iframe)

ruggi:
Fix #5175

**Problem**

The editor should offer a share dialog equivalent to the one available in the projects dashboard.

**Fix**

https://github.com/concrete-utopia/utopia/assets/1081051/720e227f-3163-4e0f-9c1d-f0e85889a77b

To make things easier, we decided to go with an iframe for the sharing dialog so the same UI can pretty much be shared between the editor and the remix projects.

_The diff is a bit large but it's pretty much self-contained stuff (e.g. the editor state changes are 10 files alone) that's not really meant to be split up in order not to lose context_. To simplify reviewing each commit holds a different piece of the puzzle.

1. Once in a multiplayer environment, clicking on the `Share` button in the top right corner of the right pane of the editor will open up a dialog centered on the screen with the visibility/copy/access lists.
2. All operations can be performed like they happen on the projects page (changing visibility, accepting requests, etc).
3. Extract the sharing dialog content component so it can be reused between the dashboard and the iframe (which has its own route).
4. To mitigate network latency, a loading spinner is shown and the iframe height is adjusted dynamically to avoid a jumpy/glitchy visual experience.
5. Non-owner users don't see the `Share` button but rather a `Fork` button.

**Future todos**

1. Add some visual feedback for ongoing operations (although it could be tricky since it's an iframe…)
2. Show a badge for pending requests on the share button


utopia-remix/app/routes/internal.editor.projects.$id.sharing.tsx
```diff
@@ -0,0 +1,89 @@
...
+  return json(
+    {
+      user: user,
+      project: sharingDetails,
+    },
+    { headers: { 'cache-control': 'no-cache' } },
+  )
+}
+
+const SharingIframe = React.memo(() => {
```
Rheeseyb, line 37: I don't think there's any need for us to call it `*Iframe` here

bkrmendy (approved):

PR #5179, opened 2024-04-05
Reset Github state on auth mismatch

### LINKED ISSUE #5178: Reset GH auth on mismatch

ruggi:
Fix #5178 

**Problem:**

If the Github authentication fails on a previously connected project, every time the user does something GH-related (explicitly, like opening the GH pane, or implicitly with the background refresh) there will be error toasts shown in the editor.

**Fix:**

The easiest way to trigger this is to manually revoke the GH auth from the GH dashboard (https://github.com/settings/applications).

This PR adjusts a few things to make sure these kinds of situations are handled more gracefully:
1. when a GH operation fails (since the backend is not giving specific error results back), check whether the user details are available; if they are not, reset the GH state.
2. when the GH data is refreshed, enforce the presence of the user details _alongside_ the authenticated flag; if they are not there, reset the GH state.
3. currently, just hitting the `Authenticate with Github` button in the GH pane will assume the user _is_ authenticated even if there's a cookie mismatch; if that's not the case, instead, just assume it's not fully authenticated yet.

### DIFF
--- a/editor/src/core/shared/github/helpers.ts
+++ b/editor/src/core/shared/github/helpers.ts
@@ -19,6 +19,7 @@ import type { EditorAction, EditorDispatch } from '../../../components/editor/ac
 import {
   deleteFile,
   removeFileConflict,
+  setGithubState,
   showToast,
   updateBranchContents,
   updateFile,
@@ -38,7 +39,6 @@ import type {
 import {
   emptyGithubData,
   emptyGithubSettings,
-  FileChecksums,
   githubOperationPrettyName,
   projectGithubSettings,
 } from '../../../components/editor/store/editor-state'
@@ -71,6 +71,10 @@ export function dispatchPromiseActions(
   promise: Promise<Array<EditorAction>>,
 ): Promise<void> {
   return promise.then((actions) => dispatch(actions, 'everyone'))
+}
+
+async function promisifyEditorActions(actions: EditorAction[]): Promise<EditorAction[]> {
+  return actions
 }
 
 export function parseGithubProjectString(maybeProject: string): GithubRepo | null {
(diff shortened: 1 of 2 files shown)


editor/src/core/shared/github/helpers.ts
```diff
@@ -850,10 +880,17 @@ export async function refreshGithubData(
...
+  if (!githubAuthenticated) {
+    promises.push(Promise.resolve([updateGithubData(emptyGithubData())]))
+  } else if (githubUserDetails === null) {
+    const userDetails = await getUserDetailsFromServer()
+    const actions: EditorAction[] =
+      userDetails == null
+        ? resetGithubStateAndDataActions()
+        : updateUserDetailsFromServerActions(userDetails)
+
+    promises.push(promisifyEditorActions(actions))
```
balazsbajorics, line 892: this is very much nitpick / conversation, but to me an obscure helper function always means I have to mentally jump there and then figure out what it does, and then "pop the stack" and continue reading the code. the only exception are "Standard Library" helper functions which we use in a lot of places.

in this specific case, I don't think the readability is worse if we just use the builtin way to promisify something:
```suggestion
    promises.push(Promise.resolve(actions))
```

balazsbajorics (approved):

editor/src/core/shared/github/helpers.ts
ruggi: oh nice I did not know it existed 👍

Rheeseyb (approved):

editor/src/core/shared/github/helpers.ts
```diff
@@ -846,14 +872,21 @@ export async function refreshGithubData(
   previousCommitSha: string | null,
   originCommitSha: string | null,
   operationContext: GithubOperationContext,
-): Promise<void> {
+): Promise<Array<Promise<Array<EditorAction>>>> {
```
seanparsons, line 875: This type feels wrong, a Promise of an array of Promises is functionally equivalent to an array of Promises.

I think this should be refactored slightly so that the `await` is reworked into a `.then` as this function is effectively blocking with the `await` and then turning the result into a resolved promise.

seanparsons (approved):

editor/src/core/shared/github/helpers.ts
balazsbajorics: nice catch!

PR #5181, opened 2024-04-05
New component descriptor errors

gbalint:
**Problem:**
Component descriptor file errors are registered as eslint errors, even though they are not that.

**Fix:**
- Introduced a new kind of error in `EditorState.codeEditorErrors` called `componentDescriptorErrors`
- Added a new action to set component descriptor errors
- Added these errors to the error overlay
- Introduced a new error source `component-descriptor`
- Modified the errors coming from component descriptor parsing/validation so they contain `component-descriptor` as source
- Changes the error severity of these errors from fatal to warning, so they don't block the canvas from rendering

### DIFF
--- a/editor/src/third-party/react-error-overlay/containers/CompileErrorContainer.tsx
+++ b/editor/src/third-party/react-error-overlay/containers/CompileErrorContainer.tsx
@@ -31,7 +31,6 @@ interface CompileErrorContainerProps {
 class CompileErrorContainer extends PureComponent<CompileErrorContainerProps, {}> {
   render() {
     const { currentBuildErrorRecords, editorHandler, onOpenFile } = this.props
-    console.log('currentBuildErrorRecords', currentBuildErrorRecords)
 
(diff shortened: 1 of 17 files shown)


editor/src/third-party/react-error-overlay/containers/CompileErrorContainer.tsx
```diff
@@ -31,6 +31,7 @@ interface CompileErrorContainerProps {
 class CompileErrorContainer extends PureComponent<CompileErrorContainerProps, {}> {
   render() {
     const { currentBuildErrorRecords, editorHandler, onOpenFile } = this.props
+    console.log('currentBuildErrorRecords', currentBuildErrorRecords)
```
balazsbajorics, line 34: shouldn't this be a console.info or console.WARN?

balazsbajorics (approved):

bkrmendy (approved):

PR #5184, opened 2024-04-08
Small fixes to projects selection

liady:
- Clicking a project does not toggle selection
- Clicking on an area that is not a project card deselects all project

<video src="https://github.com/concrete-utopia/utopia/assets/7003853/5c22f437-6b23-4b86-a8b2-8862ee0d48ca" />

### DIFF
--- a/utopia-remix/app/routes/projects.tsx
+++ b/utopia-remix/app/routes/projects.tsx
@@ -17,7 +17,7 @@ import { json, type LoaderFunctionArgs } from '@remix-run/node'
 import { useFetcher, useLoaderData } from '@remix-run/react'
 import moment from 'moment'
 import type { UserDetails } from 'prisma-client'
-import React from 'react'
+import React, { useEffect } from 'react'
 import { ProjectActionsMenu } from '../components/projectActionContextMenu'
 import { SharingDialogWrapper } from '../components/sharingDialog'
 import { SortingContextMenu } from '../components/sortProjectsContextMenu'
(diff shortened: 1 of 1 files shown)


utopia-remix/app/routes/projects.tsx
```diff
@@ -207,6 +207,29 @@ const ProjectsPageInner = React.memo(() => {
     return activeProjects.find((p) => p.proj_id === sharingProjectId) ?? null
   }, [activeProjects, sharingProjectId])
 
+  const containerRef = React.useRef<HTMLDivElement | null>(null)
+
+  const setSelectedProjectId = useProjectsStore((store) => store.setSelectedProjectId)
+
+  // deselect all projects when clicking outside of a project
+  useEffect(() => {
+    const containerRefCurrent = containerRef.current
```
gbalint, line 216: Do we really need the useEffect? Can't we just add the event listener to deselect on selection? And remove it on deselection?

ruggi: +1, I think it's just fine to add the click handler to the project entries with a stop propagation, and a click event for deselection on the component containing the entries

liady: yes, generally stopping propagation is not always a good practice since you're preventing other ancestors to respond to clicks for future purposes (if they want), but here it's actually ok since no one else cares about it :) changed it to use the stop propagation method. the mousedown handler fires also on context menu clicks, so added a check that it's indeed inside the container

utopia-remix/app/routes/projects.tsx
```diff
@@ -207,6 +207,26 @@ const ProjectsPageInner = React.memo(() => {
...
 
+  const setSelectedProjectId = useProjectsStore((store) => store.setSelectedProjectId)
+
+  const containerRef = React.useRef<HTMLDivElement | null>(null)
+
+  const onContainerMouseDown = React.useCallback(
+    (e: React.MouseEvent) => {
+      const target = e.target as HTMLElement
+      // check if target is inside the projects' container
+      // since this handler fires also on context menu clicks
```
seanparsons, line 218: I believe you can exclude context menu clicks by checking `e.type !== 'contextmenu'` BTW if that would change the logic.

seanparsons (approved):

utopia-remix/app/routes/projects.tsx
liady: Yes, but maybe in the future we will also open it from a button, etc, so I just wanted to give the proper check here (that the clicked element is indeed inside the projects page container)

gbalint (approved):

PR #5188, opened 2024-04-08
Move gtag init code to root

liady:
**Problem:**
Trying to load the root page (i.e `utopia.app` or `utopia.pizza`) and consent to cookies breaks this page forever

**Fix:**
Fix the `gtag` initialization code as per Remix's recommendation:
https://github.com/remix-run/examples/blob/main/google-analytics/app/root.tsx

### DIFF
--- a/utopia-remix/app/root.tsx
+++ b/utopia-remix/app/root.tsx
@@ -63,6 +63,23 @@ export default function App() {
         <Links />
       </head>
       <body className={styles.root}>
+        <script async src='https://www.googletagmanager.com/gtag/js?id=G-QM0KPN0RNV'></script>
+        <script
+          dangerouslySetInnerHTML={{
+            __html: `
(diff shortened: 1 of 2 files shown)


ruggi (approved):

ruggi (commented):
wait, this is setting the gtag thing on _all_ remix routes, isn't it? why can't this be on the _index page alone?

Rheeseyb (approved):

liady:
@ruggi it doesn't work there, probably due to how Remix is doing SSR vs CSR in inner routes. 
(see Remix's recommendation on where to initialize it) 
It's just the initialization code though (no events are being fired).
The alternative is to set it using `useEffect` in the component itself.

ruggi (approved):

Rheeseyb (approved):

PR #5189, opened 2024-04-08
Duplicated UID Error Message Improvements

balazsbajorics:
**Problem:**
image
The duplicate UID error message is cryptic, and it persists on the screen even after the project no longer has duplicated UIDs. This makes it hard to turn this into an actionable bug report.

**Context:**
If we can detect that there is a duplicated UID, we _could_ also run code that fixes the UIDs. 
The reason we don't do that today is that it may lead to selection loss, and also it is a performance-costly and memory intensive operation. So we definitely don't want this to be a silent problem with a silent self-healing mechanism because that would lead to silently degraded editor performance.

**Proposal:**
1. try to run an automatic UID fix mitigation, as a fallback, to increase the editor's resilience.
2. Ultimately our goal should be to not have to run the mitigation and instead make sure the actions themselves are not emitting code with duplicated UIDs.

**This PR:**
As a first step, I'd make the error message clearer, and contain the offending action names so we can create bug reports and repro cases. Although this is more noisy in the short term, I hope that in a few days this should weed out a complete list of the problematic actions that have been plaguing us recently:
image 
To reduce noise a bit, I turned these toasts into non-sticky, meaning they disappear after a few seconds.

### DIFF
--- a/editor/src/components/editor/store/dispatch.tsx
+++ b/editor/src/components/editor/store/dispatch.tsx
@@ -824,7 +824,7 @@ function applyProjectChangesToEditor(
 }
 
 export const UTOPIA_DUPLICATE_UID_ERROR_MESSAGE = (dispatchedActions: string) =>
-  `A dispatched action resulted in dupliacted UIDs. Suspicious actions: ${dispatchedActions}`
+  `A dispatched action resulted in duplicated UIDs. Suspicious actions: ${dispatchedActions}`
 function editorDispatchInner(
   boundDispatch: EditorDispatch,
   dispatchedActions: EditorAction[],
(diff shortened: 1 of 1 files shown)


editor/src/components/editor/store/dispatch.tsx
```diff
@@ -823,7 +823,8 @@ function applyProjectChangesToEditor(
   }
 }
 
-export const UTOPIA_IRRECOVERABLE_ERROR_MESSAGE = `Utopia has suffered from an irrecoverable error, please reload the editor.`
+export const UTOPIA_DUPLICATE_UID_ERROR_MESSAGE = (dispatchedActions: string) =>
+  `A dispatched action resulted in dupliacted UIDs. Suspicious actions: ${dispatchedActions}`
```
Rheeseyb, line 827:
```suggestion
  `A dispatched action resulted in duplicated UIDs. Suspicious actions: ${dispatchedActions}`
```

Rheeseyb (approved):

gbalint (approved):

PR #5195, opened 2024-04-08
Feat/editor requests badge

### LINKED ISSUE #5194: Add notification badge for pending requests on the editor

ruggi:
Fix #5194 

**Problem:**

Pending requests on a project should show as a badge in the editor's share button similarly to what happens in the projects page.

**Fix:**

Screenshot 2024-04-08 at 6 56 47 PM

This PR expands the functionality of the metadata endpoint to also return extra metadata; specifically, in this case, just a single optional `hasPendingRequests` field. The field is populated only if the user owns the project.

The metadata endpoint is called at fixed intervals, so fresh-ish data is always available, so the new badge becomes reactive to changes to access requests statuses.


liady (approved):

editor/src/components/user-bar.tsx
```diff
@@ -196,6 +196,12 @@ const MultiplayerUserBar = React.memo(() => {
     return ownerId === myUser.id
   }, [ownerId, myUser])
 
+  const projectMetadata = useEditorState(
+    Substores.projectServerState,
+    (store) => store.projectServerState.projectData,
```
balazsbajorics, line 201: why not 
```suggestion
    (store) => store.projectServerState.projectData?.hasPendingRequests,
```
this way we only need to re-render if the boolean flag flips

balazsbajorics (approved):

editor/src/components/user-bar.tsx
ruggi: ah yeah ofc [`ab6abaa` (#5195)](https://github.com/concrete-utopia/utopia/pull/5195/commits/ab6abaad26aa43f4dd5d3817254f027d37e6dca2)

utopia-remix/app/models/project.server.ts
```diff
@@ -264,3 +264,27 @@ export async function updateGithubRepository(params: {
...
+      ProjectAccessRequest: true,
+    },
+  })
+  if (project == null) {
+    return null
+  }
+
+  return {
+    hasPendingRequests: project.ProjectAccessRequest.some(
+      (r) => r.status === AccessRequestStatus.PENDING,
```
seanparsons, line 287: Shouldn't this be part of the query to prisma?

seanparsons (approved):

utopia-remix/app/models/project.server.ts
ruggi: mmmh I think that if we put that in the query we cannot do further checks on the access request statuses down the line

PR #5206, opened 2024-04-09
Component API - Navigator/Inspector configuration

bkrmendy:
## Description

This PR adds new props to `ComponentToRegister`, which can be used by the inspector or the navigator to provide better UX for the registered components. The API change is in `utopia-api/src/helpers/helper-functions.ts`.

### Details
- I went with the following pattern for defining union types:
```typescript
export const FocusOptions = ['default', 'always', 'never'] as const
export type Focus = (typeof FocusOptions)[number]
```
This way, the values can be validated by the `as const` array, and the union type can be created through the `typeof` + `[number]` pattern
- In the constructor function of `ComponentDescriptor` (`componentDescriptor`) I'm passing the optional params as `T | undefined` instead of `param?: T` so that we get type errors wherever these params aren't passed (for example, in the equality instance definitions)

### DIFF
--- a/editor/src/components/custom-code/code-file.ts
+++ b/editor/src/components/custom-code/code-file.ts
@@ -130,10 +130,20 @@ export interface ComponentDescriptor {
   preferredChildComponents: Array<PreferredChildComponentDescriptor>
   variants: ComponentInfo[]
   source: ComponentDescriptorSource
-  focus?: Focus
-  inspector?: InspectorSpec
-  emphasis?: Emphasis
-  icon?: Icon
+  focus: Focus
+  inspector: InspectorSpec
+  emphasis: Emphasis
+  icon: Icon
+}
+
+export const ComponentDescriptorDefaults: Pick<
+  ComponentDescriptor,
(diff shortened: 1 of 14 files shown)


liady (approved):

editor/src/components/custom-code/code-file.ts
```diff
@@ -129,6 +130,10 @@ export interface ComponentDescriptor {
   preferredChildComponents: Array<PreferredChildComponentDescriptor>
   variants: ComponentInfo[]
   source: ComponentDescriptorSource
+  focus?: Focus
```
balazsbajorics, line 133: why not `focus: Focus | undefined` here too?

balazsbajorics (approved):

gbalint (approved):

editor/src/components/custom-code/code-file.ts
gbalint: because this is the API type, and we want to allow the users to skip setting focus.

seanparsons (approved):

PR #5209, opened 2024-04-09
Pages Tab v0.1

balazsbajorics:
image

This PR introduces the Pages tab in the left pane. 

The way it works is that the user needs to add an array of Featured Routes (which we currently store in the package.json), like so:
```json
{
  "... rest of package json",
  "utopia": {
    "featuredRoutes": [
      "/",
      "/collections/unisex",
      "/products/beanie",
      "/blogs/news",
      "/blogs/news/making-liquid"
    ]
  }
}
```

(this is not the final API to add the featuredRoutes, we'll iterate on this)

We take the array of routes, and show them in the Pages tab. We show routes that actually match a route (these are the "pages") with the file icon, and paths that don't match any route are considered "folders". (in the screenshot, `/products` is like that)

If a dynamic route segment is hovered or selected, we show the URL parameter name.

Instead of the file-based routing's dot-delimiters, we show the `/` as delimiter, to match the RemixScene's navigation bar.


editor/src/components/navigator/left-pane/pages-pane.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+  const remixRoutes: Array<RouteMatch> = useEditorState(
+    Substores.derived,
+    (store) => {
+      const result = uniqBy(
+        registeredExampleRoutes.flatMap((exampleRoute): Array<RouteMatch> => {
+          const matchResult = matchRoutes(store.derived.remixData?.routes ?? [], exampleRoute) ?? []
+
+          if (
+            matchResult[matchResult.length - 1]?.pathname !==
```
ruggi, line 46: can the index go to `-1`?

editor/src/components/navigator/left-pane/pages-pane.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const [activeRemixScene] = useAtom(ActiveRemixSceneAtom)
+
+  const onClick = React.useCallback(() => {
+    void navigationControls[EP.toString(activeRemixScene)]?.navigate(props.resolvedPath)
+  }, [navigationControls, activeRemixScene, props.resolvedPath])
+
+  const resolvedRouteSegments = props.resolvedPath.split('/')
+  const templateRouteSegments = props.routePath.split('/')
+
+  const lastResolvedSegment = resolvedRouteSegments[resolvedRouteSegments.length - 1]
```
ruggi, line 137: same question

ruggi (approved):

editor/src/components/navigator/left-pane/pages-pane.tsx
balazsbajorics: yep if there's no match

ruggi: would it make sense then to use `safeIndex`?

editor/src/components/navigator/left-pane/pages-pane.tsx
balazsbajorics: here, I _think_ not

editor/src/components/navigator/left-pane/pages-pane.tsx
balazsbajorics: Actually I wanted to use Array.prototype.get(-1), but it needs bumping our TS config.

I'm forgot what safeIndex does. would it enforce the question mark?

editor/src/components/navigator/left-pane/pages-pane.tsx
```diff
@@ -0,0 +1,234 @@
...
+            return [
+              {
+                path: exampleRoute,
+                resolvedPath: exampleRoute,
+                matchesRealRoute: false,
+              },
+            ]
+          }
+
+          return matchResult?.map(
```
gbalint, line 58: As far as I see `matchResult` can not be null (it is initialized with `?? []`), so the `?` operator is unnecessary

editor/src/components/navigator/left-pane/pages-pane.tsx
ruggi: no, it would give you back an `| undefined` that you can handle instead of throwing because of the negative index

editor/src/components/navigator/left-pane/pages-pane.tsx
gbalint: .split() always returns an array with >0 length (even if the splitter character is not in the string, it returns the whole string itself), so the index here can not be negative

editor/src/components/navigator/left-pane/pages-pane.tsx
balazsbajorics: image
I thought arrays don't throw on negative index!

editor/src/components/navigator/left-pane/pages-pane.tsx
```diff
@@ -0,0 +1,234 @@
...
+      >
+        {lastTemplateSegment}
+      </span>
+    </FlexRow>
+  )
+})
+
+function fillInGapsInRoute(routes: Array<string>): Array<string> {
+  // if we find a route /collections, and a route /collections/hats/beanies, we should create an entry for /collections/hats, so there are no gaps in the tree structure
+  const result: Array<string> = []
```
gbalint, line 221: Our convention is to use `let` when we mutate the variable later (even though I don't like that convention :) )

gbalint (approved):

PR #5210, opened 2024-04-10
Support pasting on StringInput

ruggi:
**Problem:**

The StringInput components don't react to paste events because they miss the setter changes for their values.

**Fix:**

This PR expands the `StringInput` events handling so that it can automatically handle paste events if
- the event is pure text and _not_ utopia content
- the component has a defined `onChange` event handler


editor/src/uuiui/inputs/string-input.tsx
```diff
@@ -77,6 +78,46 @@ export const StringInput = React.memo(
 
       const placeholder = getControlStylesAwarePlaceholder(controlStyles) ?? initialPlaceHolder
 
+      const onChange = React.useCallback(
```
liady, line 81: btw, would we expect this logic (`onChange`, `onPaste`) to be on the StringInput or the HeadlessStringInput?

liady (approved):

editor/src/uuiui/inputs/string-input.spec.browser2.tsx
```diff
@@ -77,4 +83,106 @@ describe('StringInput', () => {
...
+        </TestInspectorContextProvider>,
+      )
+
+      const inputElement = await result.findByTestId('the-input')
+      if (inputElement == null || !(inputElement instanceof HTMLInputElement)) {
+        throw new Error('Could not find input element.')
+      }
+
+      firePasteTextEvent(inputElement, 'well hello there!')
+      await wait(100)
```
balazsbajorics, line 172: could this be replaced by an `await firePasteTextEvent`? if not, then `await result.getDispatchFollowUpActionsFinished()` ? 

the await(0) is sometimes acceptable as it means we wait a tick. but awaiting a concrete number of milliseconds should never be used because it means there is possibility for the test to become flaky

ruggi: the render result is not a `EditorRenderResult`, so I just changed the awaits to be 0s

PR #5211, opened 2024-04-10
component API: children prop

bkrmendy:
# [Project with an updated components file](https://utopia.fish/p/f65a8658-walnut-midnight/?branch_name=feature-component-registration-api-children)

## Description
This PRs adds some new configuration options to the component registration API. The new stuff is:
- `children`: makes it possible to specify whether the element supports children, and if it does, what elements should be rendered by that component and what placeholders should be used if there are no children. The `supportsChildren` and `preferredChildComponents` props have been removed from the public-facing API, and rolled into the `children` prop.
- `variants`: this prop was already present in the previous version of the API, but now it's possible to add variants in a more streamlined way. See the `ComponentExample` for details.
- it's possible to add placeholders and component examples to JSX controls too, using the same that `children` uses.

## Scope
The sole aim of this PR is to add the new config props and the code involved in parsing/validating them. Any other feature involved will come on a separate PR.

## Manual Tests
I hereby swear that:


editor/src/components/custom-code/code-file.ts
```diff
@@ -124,10 +124,18 @@ export function componentInfo(
   }
 }
 
+export type PlaceholderSpec =
+  // a placeholder that inserts a span with that wraps `contents`. This will be
```
balazsbajorics, line 128: needs better copy

balazsbajorics (approved):

editor/src/components/custom-code/code-file.ts
```diff
@@ -124,10 +124,18 @@ export function componentInfo(
   }
 }
 
+export type PlaceholderSpec =
+  // a placeholder that inserts a span that wraps `contents`. This will be
+  // editable with the text editor
+  | { type: 'text'; contents: string }
+  // inserts a `div` with the specified width and height
+  | { type: 'spacer'; width: number; height: number }
```
Rheeseyb, line 132: We also want one which completely fills the available space

bkrmendy: added it in https://github.com/concrete-utopia/utopia/pull/5211/commits/298cef9a97a3b6588a485b7595070a1d1f75349a

Issue #5243, opened 2024-04-14
Inserting a render prop causes UID error


bkrmendy:
This one:

![Image](https://github.com/concrete-utopia/utopia/assets/2945037/9a7bda77-0f17-418b-a6e2-182f40286c8b)

PR #5244, opened 2024-04-14
Inserting a render prop causes UID error

bkrmendy:
## Problem

https://github.com/concrete-utopia/utopia/issues/5243

## Cause
The uids within the props of the inserted element weren't fixed, which led to duplicate uids creeping into the project.

## Fix
Fix the inserted element with `fixUtopiaElement` before inserting it, which "fixes" the render prop uids too.

## Manual Tests:
I hereby swear that:

### DIFF
--- a/editor/src/components/navigator/navigator-item/render-prop-selector-popup.spec.browser2.tsx
+++ b/editor/src/components/navigator/navigator-item/render-prop-selector-popup.spec.browser2.tsx
@@ -5,7 +5,7 @@ import {
   getPrintedUiJsCodeWithoutUIDs,
   renderTestEditorWithModel,
 } from '../../canvas/ui-jsx.test-utils'
-import { StoryboardFilePath } from '../../editor/store/editor-state'
+import { StoryboardFilePath, navigatorEntryToKey } from '../../editor/store/editor-state'
 import * as EP from '../../../core/shared/element-path'
 import {
   mouseClickAtPoint,
@@ -489,7 +489,7 @@ describe('The navigator render prop picker', () => {
     )
   })
 
-  it('deduplicated prop UIDs', async () => {
+  it('deduplicates render prop UIDs', async () => {
     const editor = await renderTestEditorWithModel(
       createModifiedProject({
         [StoryboardFilePath]: `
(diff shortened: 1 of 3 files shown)


seanparsons (approved):

editor/src/components/navigator/navigator-item/render-prop-selector-popup.spec.browser2.tsx
```diff
@@ -488,4 +488,140 @@ describe('The navigator render prop picker', () => {
...
+  it('deduplicates render prop UIDs', async () => {
+    const editor = await renderTestEditorWithModel(
+      createModifiedProject({
+        [StoryboardFilePath]: `
+        import * as React from 'react'
+        import * as Utopia from 'utopia-api'
+        import { Storyboard, Scene } from 'utopia-api'
+
+        export function Flex({ hello }) {
+          return <div>{hello.gr}</div>
```
balazsbajorics, line 501: this is probably not important but hello.gr will render a null. but maybe that changes the navigator outcome?

Rheeseyb: Looks like this was a typo then, since it's meant to be `hello.greeting`.

Rheeseyb (approved):

editor/src/components/navigator/navigator-item/render-prop-selector-popup.spec.browser2.tsx
bkrmendy: thanks for noticing it, I fixed it in https://github.com/concrete-utopia/utopia/pull/5244/commits/71080c08d99bc8445d11f46f59012aab7a97a756

editor/src/components/navigator/navigator.spec.browser2.tsx
```diff
@@ -5049,7 +5049,7 @@ describe('Navigator', () => {
         'regular-sb/scene/pg:dbc',
         'regular-sb/scene/pg:dbc/78c',
         'render-prop-sb/scene/pg:dbc/78c/prop-label-header-header',
-        'regular-sb/scene/pg:dbc/78c/pro', // <- the inserted render prop
```
balazsbajorics, line 5052: seems like by adding the to-be-inserted uid to `uniqueIds.add(uid)` we make it so that the inserted uid can _never_ appear in the final project because the fixUtopiaElement will treat the UID as taken.

is this unavoidable?

Rheeseyb (approved):

balazsbajorics (approved):

PR #5247, opened 2024-04-15
do not parse unsupported files

bkrmendy:
## Problem
We attempt to parse non js/ts files with the parsing code. 

## Fix
Only parse an allowlisted set of file extensions. The allowlist is implemented in the `isParseableFile` function.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/core/workers/parser-printer/parser-printer.ts
+++ b/editor/src/core/workers/parser-printer/parser-printer.ts
@@ -92,7 +92,7 @@ import {
 import * as PP from '../../shared/property-path'
 import { assertNever, fastForEach } from '../../shared/utils'
 import { addImport, emptyImports } from '../common/project-file-utils'
-import { FileExtensionsToLint, lintCode } from '../linter/linter'
+import { FileExtensionsToParse, lintCode } from '../linter/linter'
 import type { FunctionContents, WithParserMetadata } from './parser-printer-parsing'
 import {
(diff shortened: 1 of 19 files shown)


seanparsons (approved):

editor/src/core/workers/parser-printer/parser-printer.ts
```diff
@@ -1940,6 +1942,11 @@ export function lintAndParse(
   shouldTrimBounds: 'trim-bounds' | 'do-not-trim-bounds',
   applySteganography: SteganographyMode,
 ): ParsedTextFile {
+  const fileExtension = getFileExtension(filename)
+  if (!FileExtensionsToParse.includes(fileExtension.toLowerCase())) {
```
Rheeseyb, line 1946: Does `getFileExtension` include the `.`?
I'd say it's better to use `isJsFile` declared here since that also includes mjs and cjs files <https://github.com/concrete-utopia/utopia/blob/c2724ce54a8b6ca8db5f40c1672e8f4871966b89/editor/src/core/shared/file-utils.ts#L246-L253>

bkrmendy: thanks, pushed the change in https://github.com/concrete-utopia/utopia/pull/5247/commits/4399d71f4e134a9f482b660b1564d34564271940

Rheeseyb (approved):

PR #5253, opened 2024-04-15
Turn down connection amount for Haskell server.

seanparsons:
**Problem:**
Collectively we're consuming too many of the PostgreSQL connection pool connections.

**Fix:**
Turn down the number of connections the Haskell server opens.

**Commit Details:**
- In `createDatabasePoolFromConnection` turn the connections down to 2 stripes of 1 connection each for an absolute maximum of 2.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/server/src/Utopia/Web/Database.hs
+++ b/server/src/Utopia/Web/Database.hs
@@ -109,7 +109,9 @@ getDatabaseConnectionString = lookupEnv "DATABASE_URL"
 createDatabasePoolFromConnection :: IO Connection -> IO DBPool
 createDatabasePoolFromConnection createConnection = do
   let keepResourceOpenFor = 10
-  createPool createConnection close 2 keepResourceOpenFor 1
+  let poolStripes = 2
+  let connectionsPerStripe = 1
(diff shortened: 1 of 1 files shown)


ruggi (approved):

balazsbajorics (approved):

server/src/Utopia/Web/Database.hs
```diff
@@ -109,7 +109,7 @@ getDatabaseConnectionString = lookupEnv "DATABASE_URL"
 createDatabasePoolFromConnection :: IO Connection -> IO DBPool
 createDatabasePoolFromConnection createConnection = do
   let keepResourceOpenFor = 10
-  createPool createConnection close 3 keepResourceOpenFor 3
+  createPool createConnection close 2 keepResourceOpenFor 1
```
Rheeseyb, line 112: Can you add a pair of `let`s for these two numbers please since they are, or add a comment saying what they are

seanparsons: Fixed in 4ce3eb773.

PR #5257, opened 2024-04-15
Remove URL param after clone

liady:
Removing the 'clone' URL param after project creation

### DIFF
--- a/editor/src/components/editor/editor-component.tsx
+++ b/editor/src/components/editor/editor-component.tsx
@@ -75,7 +75,7 @@ import { useIsLoggedIn, useLiveblocksConnectionListener } from '../../core/share
 import { ForkSearchParamKey, ProjectForkFlow } from './project-fork-flow'
 import { isRoomId, projectIdToRoomId } from '../../utils/room-id'
 import { SharingDialog } from './sharing-dialog'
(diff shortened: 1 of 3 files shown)


balazsbajorics (approved):

editor/src/components/editor/editor-component.tsx
```diff
@@ -91,8 +91,10 @@ function pushProjectURLToBrowserHistory(
     // …but if it's forking, remove the fork param
     queryParams.delete(ForkSearchParamKey)
   }
-  // remove accessLevel param
-  queryParams.delete(AccessLevelParamKey)
+  // remove one-time creation params
+  ;[AccessLevelParamKey, CloneParamKey].forEach((key) => {
+    queryParams.delete(key)
+  })
```
Rheeseyb, line 97: It feels pretty unnecessary to wrap this in a `forEach` rather than just calling the function twice explicitly

Rheeseyb (approved):

editor/src/components/editor/editor-component.tsx
balazsbajorics: well for now it's two params, but we may want to add more in the future

Issue #5252, opened 2024-04-15
Deleting a render prop from a generated element doesn't work, should


maltenuhn:
We currently don't allow for deletion of generated components - but you should be able to delete (the content of) _their_ render props

![Image](https://github.com/concrete-utopia/utopia/assets/2945037/e91047c4-51f1-4a7e-9e42-4fc4c04b8bae)

PR #5260, opened 2024-04-15
Delete render prop from element in map

bkrmendy:
**Problem:**
https://github.com/concrete-utopia/utopia/issues/5252

**Fix:**
Make it possible

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/editor/actions/actions.spec.browser2.tsx
+++ b/editor/src/components/editor/actions/actions.spec.browser2.tsx
@@ -571,9 +571,9 @@ describe('actions', () => {
                 ["test"].map(() => (
                   <Card
                     data-uid='card'
-                    header={<h3 data-uid='b82'>woot</h3>}
+                    header={<h3 data-uid='render-prop-element'>woot</h3>}
                   >
-                    <p data-uid='render-prop'>Card contents</p>
+                    <p data-uid='child-element'>Card contents</p>
                   </Card>
                 ))
               }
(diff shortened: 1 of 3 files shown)


editor/src/components/editor/actions/actions.tsx
```diff
@@ -1419,7 +1419,7 @@ function toastOnGeneratedElementsTargeted(
   }
 
   if (!generatedElementsTargeted || allowActionRegardless) {
-    result = actionOtherwise(result)
+    result = actionOtherwise(editor)
```
bkrmendy, line 1422: I changed this to `editor` from `result` because if `result` is passed in, `generatedElementsTargeted` is false (so it becomes true after the negation), and `allowActionRegardless` is true, both the toast and the result of the action would be added to the resulting editor state, and we only want one of those

gbalint (approved):

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2056,6 +2056,31 @@ export const MetadataUtils = {
...
+    const element = MetadataUtils.findElementByElementPath(componentMetadata, parentPath)
+    if (
+      element == null ||
+      !isRight(element.element) ||
+      element.element.value.type !== 'JSX_ELEMENT'
+    ) {
+      return false
+    }
+
+    const elementInProps = element.element.value.props.some(
```
balazsbajorics, line 2075: this maybe the first actual real life case when we are disadvantaged by the element path not being able to express a render prop relationship

balazsbajorics (approved):

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -531,6 +534,104 @@ describe('actions', () => {
...
+              data-uid='scene'
+              data-testid='scene'
+              commentId='sce'
+            >
+              {
+                // @utopia/uid=map
+                ["test"].map(() => (
+                  <Card
+                    data-uid='card'
+                    header={<h3 data-uid='b82'>woot</h3>}
```
balazsbajorics, line 574: instead of b82, please give this a more noticeable uid, like render-prop-to-be-deleted :)

bkrmendy: added a different uid in https://github.com/concrete-utopia/utopia/pull/5260/commits/ef26dccd82d9d08ccbbd2986ed8f1f88f4aa13d1

PR #5264, opened 2024-04-16
Remix support: renaming (home) to /

balazsbajorics:
**Problem**
On master, we display the root route as `(home)` in the canvas topmenu's navigation bar, and in the Routes section of the Pages tab.

This is non-standard, and once we want to allow free text input in the navigation bar, this will become weird. In the Routes section it's also sticking out from the rest of the routes which are displayed in standard format.

**Fix**
image

Show the root route as `/` everywhere

### DIFF
--- a/editor/src/components/navigator/left-pane/pages-pane.tsx
+++ b/editor/src/components/navigator/left-pane/pages-pane.tsx
@@ -5,10 +5,17 @@ import { jsx } from '@emotion/react'
 
 import { useAtom } from 'jotai'
 import React from 'react'
+import ReactDOM from 'react-dom'
 import { matchRoutes } from 'react-router'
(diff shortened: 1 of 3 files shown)


editor/src/components/navigator/left-pane/pages-pane.tsx
```diff
@@ -5,10 +5,17 @@ import { jsx } from '@emotion/react'
```
ruggi, line 5: 👀

ruggi (approved):

liady (approved):

editor/src/components/navigator/left-pane/pages-pane.tsx
balazsbajorics: oops

PR #5271, opened 2024-04-16
do not switch tabs on select

bkrmendy:
Prior art: https://github.com/concrete-utopia/utopia/pull/4381 (noting it for posterity)

## Problem
We don't always want to switch to the navigator tab when selection changes.

The new ruleset is the following: 
- if all elements are deselected (selected views are []): do not switch to the navigator
- if something is selected and the pages tab is active: do not switch to the navigator
- otherwise: switch to the navigator

## Fix
Tweak the `UpdateSelectedViews` command, the `SELECT_COMPONENTS` action, the `SELECT_ALL_SIBLINGS` action and the `CLEAR_SELECTION` action so that they don't change the left panel.

### Details
- The PR adds a function called `nextSelectedTab`, which implements the rules on when to switch tabs
- the following actions still switch back to the Navigator tab: `INSERT_JSX_ELEMENT`, `WRAP_IN_ELEMENT`, `UNWRAP_ELEMENTS`, `INSERT_INSERTABLE`

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/navigator/left-pane/left-pane-utils.ts
+++ b/editor/src/components/navigator/left-pane/left-pane-utils.ts
@@ -1,0 +1,17 @@
+import type { ElementPath } from '../../../core/shared/project-file-types'
+import { LeftMenuTab } from '../../editor/store/editor-state'
+
+export function nextSelectedTab(
+  currentLeftMenuTab: LeftMenuTab,
+  newSelectedPaths: ElementPath[],
+): LeftMenuTab {
+  if (newSelectedPaths.length === 0) {
+    return currentLeftMenuTab
+  }
+
+  if (currentLeftMenuTab === LeftMenuTab.Pages) {
+    return LeftMenuTab.Pages // same as LeftMenuTab.Pages
+  }
+
+  return LeftMenuTab.Navigator
+}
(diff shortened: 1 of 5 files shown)


editor/src/components/navigator/left-pane/left-pane-utils.ts
```diff
@@ -0,0 +1,17 @@
...
+export function nextSelectedTab(
+  currentLeftMenuTab: LeftMenuTab,
+  newSelectedPaths: ElementPath[],
+): LeftMenuTab {
+  if (newSelectedPaths.length === 0) {
+    return currentLeftMenuTab
+  }
+
+  if (currentLeftMenuTab === LeftMenuTab.Pages) {
+    return LeftMenuTab.Pages // same as LeftMenuTab.Pages
```
balazsbajorics, line 13: can confirm it is indeed same as LeftMenuTab.Pages :)

balazsbajorics (approved):

editor/src/components/navigator/left-pane/left-pane-utils.ts
bkrmendy: technically true https://github.com/concrete-utopia/utopia/pull/5271/commits/31e85e2eeaa2d76a8f3b7e36e14d8f1ab03655e2

liady (approved):

gbalint (approved):

Rheeseyb (approved):

PR #5273, opened 2024-04-16
Check offline state before some service calls.

### LINKED ISSUE #5230: Reduce unnecessary connection-lost errors
Client-side, today we immediately show the 'you're offline' on (1) save error, or (2) polling for login state failing with an exception. Since the editor does allow you to continue editing (and saves updates), this is too eager. (Also, seeing other people's mouse pointers move while the editor tells you it's disconnected feels dissonant)

Client-side:
- [x] bin off those two "checks" (because they can potentially pull in other types of errors and have say a super simple HEAD request to the server for checking to see if we're online)
- [x] Only show error bar after 3 failed attempts over 15 seconds, and clear it as soon as the check succeeds again
- [ ] Verify this works on a branch by disconnecting the wifi

Log-in status
A lost connection to the server (eg due to network error, or server restarting) is treated as the equivalent of a user having been logged out, and the inspector aggressively shows the log-in button right away.

- [ ] The logged-in check instead should be decoupled from, but immediately preceded by, the is-connected check
- [ ] and ideally be positively affirmed ('I received a response from the server that this person is / is not logged in'). A _non_-response shouldn't itself trigger the disconnection warning.

Server side
- [ ] Investigate if there are server-side reasons for this as well that _aren't_ due to server deploys having happened shortly before

seanparsons:
**Problem:**
When the editor goes offline, the inspector switches to an offline state before the offline banner appears, which looks a bit strange.

**Fix:**
Now the checks for login state and project server state are prefixed with a check to see if the server is online, then those calls will only proceed if the server is online. Otherwise the state isn't updated at all and will be left as it is.

**Notes:**
There's an issue with the low priority store which causes the offline bar to show up later than it should, this will be fixed in a separate PR.

**Commit Details:**
- Extract out the online state checking into `checkOnlineState` from `startOnlineStatusPolling`.
- `getProjectServerState` checks the result of `checkOnlineState` initially and if that indicates the server is offline it doesn't check the project state.
- `startPollingLoginState` checks the result of `checkOnlineState` initially and if that indicates the server is offline it doesn't check the login state.
- Added `/online-status` endpoint to the Haskell server.
- Remix `/online-status` endpoint checks the same URL on the Haskell server.

**Manual Tests:**
I hereby swear that:

Fixes #5230

### DIFF
--- a/utopia-remix/app/routes/online-status.tsx
+++ b/utopia-remix/app/routes/online-status.tsx
@@ -1,3 +1,15 @@
-export async function loader() {
+import type { LoaderFunctionArgs } from '@remix-run/node'
+import { proxy } from '../util/proxy.server'
+
+export async function loader({ request }: LoaderFunctionArgs) {
+  // Call the old server to check if it's online.
+  const url = new URL(request.url)
+  url.pathname = '/online-status'
+  const onlineStatusRequest = new Request(url, {
+    method: 'GET',
+  })
+  await proxy(onlineStatusRequest, { rawOutput: true })
+
+  // Return a simple response.
   return new Response('Online', { headers: { 'content-type': 'text/plain' } })
 }
--- a/utopia-remix/app/routes/online-status.tsx
+++ b/utopia-remix/app/routes/online-status.tsx
@@ -11,5 +11,7 @@ export async function loader({ request }: LoaderFunctionArgs) {
   await proxy(onlineStatusRequest, { rawOutput: true })
 
   // Return a simple response.
-  return new Response('Online', { headers: { 'content-type': 'text/plain' } })
+  return new Response('Online', {
+    headers: { 'content-type': 'text/plain', 'cache-control': 'no-cache' },
+  })
 }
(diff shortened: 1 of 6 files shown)


utopia-remix/app/routes/online-status.tsx
```diff
@@ -1,3 +1,15 @@
...
+  // Call the old server to check if it's online.
+  const url = new URL(request.url)
+  url.pathname = '/online-status'
+  const onlineStatusRequest = new Request(url, {
+    method: 'GET',
+  })
+  await proxy(onlineStatusRequest, { rawOutput: true })
+
+  // Return a simple response.
   return new Response('Online', { headers: { 'content-type': 'text/plain' } })
```
ruggi, line 14: I missed this, we should include `cache-control: no-cache` here to please the cloudfront deities

ruggi: (since this is not going through `handle`)

Rheeseyb (approved):

balazsbajorics (approved):

ruggi (approved):

gbalint (approved):

PR #5278, opened 2024-04-16
Use moduleName when looking up registered component

### LINKED ISSUE #5263: We need to include the module when looking up a registered component

Rheeseyb:
**Problem:**
In the render prop picker we are currently looking up registered components using just the component name, meaning we can return a matching component from different module.

**Fix:**
With the merge of #5268 we can now rely on the `moduleName` to lookup the correct registered component. This change also pulls through that `moduleName` into the internal `PreferredChildComponentDescriptor` (as that had drifted), and then adds a function for creating the default `Imports` for the default empty variant, using that `moduleName`.

To be able to find the correct registered component for the target of the picker (i.e. the component containing the slot that was clicked), I've also very slightly tweaked the way we store `ImportInfo` in the metadata. Currently, the import statement:
```js
import { Card } from '../../src/Card'
```
would result in `ImportInfo.filename` being `'../../src/Card'`, which is useless in that form. This PR will convert that relative filepath to an absolute one at the point of creating the metadata, so that we can now use it to look up the registered component.

**Manual Tests:**
I hereby swear that:

Fixes #5263

### DIFF
--- a/editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
+++ b/editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
@@ -53,7 +53,7 @@ const usePreferredChildrenForTargetProp = (
   const selectedElement = useEditorState(
     Substores.metadata,
     (store) => MetadataUtils.findElementByElementPath(store.editor.jsxMetadata, target),
-    'usePreferredChildrenForSelectedElement selectedElement',
+    'usePreferredChildrenForTargetProp selectedElement',
   )
 
   const preferredChildrenForTargetProp = useEditorState(
--- a/editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
+++ b/editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
@@ -266,6 +266,14 @@ function iconPropsForIcon(icon: Icon): IcnProps {
   }
 }
 
+export function labelTestIdForComponentIcon(
+  componentName: string,
+  moduleName: string,
+  icon: Icon,
+): string {
+  return `variant-label-${componentName}-${moduleName}-${icon}`
+}
+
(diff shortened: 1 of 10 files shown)


editor/src/utils/path-utils.ts
```diff
@@ -83,7 +83,7 @@ export function absolutePathFromRelativePath(
   originIsDir: boolean,
   relativePath: string,
 ): string {
-  if (relativePath.startsWith('/') || !relativePath.includes('/')) {
+  if (!relativePath.startsWith('.')) {
```
ruggi, line 86: shouldn't this be just
```ts
if (!relativePath.startsWith('/')) {
// ...
}
```
?

editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
```diff
@@ -59,22 +50,26 @@ const usePreferredChildrenForTargetProp = (
   target: ElementPath,
   prop: string,
 ): Array<PreferredChildComponentDescriptorWithIcon> => {
-  const selectedJSXElement = useEditorState(
+  const selectedElement = useEditorState(
     Substores.metadata,
-    (store) => MetadataUtils.getJSXElementFromMetadata(store.editor.jsxMetadata, target),
-    'usePreferredChildrenForSelectedElement selectedJSXElement',
+    (store) => MetadataUtils.findElementByElementPath(store.editor.jsxMetadata, target),
+    'usePreferredChildrenForSelectedElement selectedElement',
```
ruggi, line 56:
```suggestion
    'usePreferredChildrenForTargetProp selectedElement',
```

ruggi (approved):

balazsbajorics (approved):

seanparsons (approved):

editor/src/components/navigator/navigator-item/render-prop-selector-popup.tsx
Rheeseyb: Done 👍

PR #5280, opened 2024-04-16
Moved store updates outside of dom walker conditional.

### LINKED ISSUE #5274: The low priority store appears to be being updated at the wrong time, resulting in UI updating irregularly or later than expected.

seanparsons:
**Problem:**
It was observed that updates reliant on the low priority store seemed to be triggered later than would be expected.

**Cause:**
The updates to the low and regular stores were inside a conditional for running the DOM walker, which meant that updates not related to the canvas would only affect the editor when an update to the canvas was triggered.

**Fix:**
Now updates to the stores are done outside of the conditional which specifies if the DOM walker should be triggered.

A smaller fix was done in the process of testing this as updates to the project server state were triggering updates when they shouldn't have done, as no change actually occurred.

**Commit Details:**
- Tweaked `runUpdateProjectServerState` to not always cause updates.
- In `boundDispatch`, the updates to the regular and low priority stores are now triggered outside of a conditional for running the DOM walker.

**Manual Tests:**
I hereby swear that:

Fixes #5274

### DIFF
--- a/editor/src/components/editor/store/editor-update.tsx
+++ b/editor/src/components/editor/store/editor-update.tsx
@@ -29,6 +29,7 @@ import { InitialOnlineState } from '../online-status'
 import type { Optic } from '../../../core/shared/optics/optics'
 import { fromField } from '../../../core/shared/optics/optic-creators'
 import { modify } from '../../../core/shared/optics/optic-utilities'
+import { ProjectServerStateKeepDeepEquality } from './store-deep-equality-instances'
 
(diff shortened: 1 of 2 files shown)


balazsbajorics (approved):
since no tests cover these changes, let's keep an eye out in the next couple of days for weird errors

ruggi (approved):

gbalint (approved):

Rheeseyb (commented):
👍

Issue #5275, opened 2024-04-16
Github error messages are (still?) annoying


bkrmendy:
(MN: I have recollection of this having been addressed, but seemingly not - only relevant ticket I found was https://github.com/concrete-utopia/utopia/pull/5179)

We fire Github error messages inappropriately. Reported cases:
- [ ] user forks a project but doesn't have access to the github repo, or is not authenticated with github. 
- [ ] user disconnects github auth
- [ ] editor is offline or cannot otherwise connect to github

We don't want to completely refactor this, but I propose the following mitigations (some of which had previously been logged and might have already been implemented!)

- [ ] we never try to connect to Github if the user isn't authenticated
- [ ] we only show the toasts for user-initiated actions (those from buttons, in the UI); we don't fire error messages for system-initiated polling-type failures 
- [ ] We change the copy: "Couldn't xxxx (see console)" - ie "Couldn't list Github pull requests (see console)" - to read friendlier and not line-wrap
- [ ] We bump down the level from error to primary (to avoid the "IT'S RED EVERYTHING IS BROKEN")

![Image](https://github.com/concrete-utopia/utopia/assets/16385508/4952de3a-eb34-485f-9f47-ea205a7a373c)

PR #5282, opened 2024-04-17
Fix Github error messages

bkrmendy:
## Problem
fixes https://github.com/concrete-utopia/utopia/issues/5275

## Fix with details
(based on the issue)
- We change the copy: "Couldn't xxxx (see console)" - ie "Couldn't list Github pull requests (see console)" - to read friendlier and not line-wrap: implemented as described
- We bump down the level from error to primary (to avoid the "IT'S RED EVERYTHING IS BROKEN"): implemented as described
- We never try to connect to Github if the user isn't authenticated: implemented by calling `GithubHelpers.getUserDetailsFromServer` at the beginning of `runGithubOperation`, and early returning if the result is `null`. `GithubHelpers.getUserDetailsFromServer` fires off a request internally
- We only show the toasts for user-initiated actions (those from buttons, in the UI); we don't fire error messages for system-initiated polling-type failures: this is implemented by passing down a flag to `runGithubOperation` (`'polling' | 'user-initiated'`), and the toast is only shown if the flag is `user-initiated`. 

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/core/shared/github/helpers.ts
+++ b/editor/src/core/shared/github/helpers.ts
@@ -192,13 +192,11 @@ export async function runGithubOperation(
     if (userDetails == null) {
       actions.push(...resetGithubStateAndDataActions())
     } else {
-      actions.push(
-        showToast(notice(`${opName} failed. See the console for more information.`, 'ERROR')),
-      )
+      actions.push(showToast(notice(`Couldn't ${opName} (see console)`, 'ERROR')))
     }
 
     dispatch(actions, 'everyone')
-    console.error(`[GitHub] operation "${opName}" failed:`, error)
+    console.error(`Couldn't ${opName}`, error)
     throw error
   } finally {
     dispatch([updateGithubOperations(operation, 'remove')], 'everyone')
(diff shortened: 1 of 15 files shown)


editor/src/core/shared/github/helpers.ts
```diff
@@ -305,36 +325,38 @@ function updateUserDetailsFromServerActions(userDetails: GithubUser): EditorActi
   return [updateGithubData({ githubUserDetails: userDetails })]
 }
 
-export async function getUserDetailsFromServer(): Promise<GithubUser | null> {
-  const url = GithubEndpoints.userDetails()
+export const GithubHelpers = {
```
ruggi, line 328: what's the reason for moving the function into this const? is it for easier stubbing of the method in the tests? (in `github-operations.test-utils.ts`) if that's the case, wouldn't it be easier to just leave it as a function (since it's already inside a helpers file) and stub it with a `*` import?

seanparsons (approved):

editor/src/core/shared/github/helpers.ts
bkrmendy: It's exactly for mocking, I tried other approaches but right now the best price-to-value option is exporting through a const

PR #5299, opened 2024-04-17
Implement focus setting in the component descriptor

gbalint:
**Problem:**
Implement setting `focus` to `default`, `always` or `never` in the component descriptor.

**Fix:**
- Add element instances of `always` focused components to `derived.autoFocusedPaths`
- Always return false to `isManuallyFocusableComponent` if component is registered and `focus` is not `default`
- Make sure autofocused components and their contents are not set to purple (leave the color as it would be without the component)
- Add the contents of autofocused components to valid paths.

**Manual Tests:**
I hereby swear that:

**Missing:**
Valid path generation contains duplicated logic to make scenes with a single path autofocused. This logic should be removed in a subsequent PR.

Fixes https://github.com/concrete-utopia/utopia/issues/5289


editor/src/core/model/element-metadata-utils.ts
```diff
@@ -168,6 +169,11 @@ import { treatElementAsGroupLikeFromMetadata } from '../../components/canvas/can
 import type { RemixRoutingTable } from '../../components/editor/store/remix-derived-data'
 import { exists } from '../shared/optics/optic-utilities'
 import { eitherRight, fromField, fromTypeGuard, notNull } from '../shared/optics/optic-creators'
+import { Prop } from 'xstate/lib/model.types'
```
seanparsons, line 172: This import looks a bit suspicious...

editor/src/core/shared/element-path.ts
```diff
@@ -1115,6 +1104,32 @@ export function isExplicitlyFocused(
   return isNotAutoFocused && isFocused(focusedElementPath, path)
 }
 
+export function isInExplicitlyFocusedSubtree(
+  focusedElementPath: ElementPath | null,
+  autoFocusedPaths: Array<ElementPath>,
+  path: ElementPath,
+): boolean {
+  // we exclude children of a potentially focused component as children are not part of the focus system
```
seanparsons, line 1112: Love the comments!

seanparsons (approved):

bkrmendy (approved):

PR #5303, opened 2024-04-17
Fix UIDs when changing elements inside maps

### LINKED ISSUE #5214: Selection lost when making a change via the Inspector to a single selected element inside a map
selecting a Size in the sample project and setting a prop via the inspector results in deselection.

Or select a Money, and change the boolean prop showCurrency to true or false

(this probably happens because the UID of the edit element changes and our uid-fix code can't fix it, but we are not sure) 

to reproduce: make sure the 'print UIDs' feature switch is turned off

liady:
**Problem:**
When changing a prop in an element inside a code block (map/arbitrary code block) - the selection was lost. This was due to the element UID changing without being able to be fixed by the `fix-uids` methods.
<video src="https://github.com/concrete-utopia/utopia/assets/7003853/943b16d4-d0b3-4c5d-899e-2a1a3426ab16" />

**Fix:**
1. Instead of matching the old and new expressions according to the old UID (which wasn't present in the new expression) - a fallback was added to match it according to order as well
2. This fix exposed another issue where if `elementsWithin` ids are fixed - there was still a mistmatch with other string fields (i.e `originalJavascript`, `transpiledJavascript`, etc) where the old UIDs where present. Another fix was made to change the UIDs in them as well.
3. Added a test that fails without the fix (and passes with it)

Fixes #5214


editor/src/core/workers/parser-printer/uid-fix.spec.ts
```diff
@@ -307,6 +320,31 @@ describe('fixParseSuccessUIDs', () => {
...
+    const base = lintAndParseAndValidateResult(
+      'test.js',
+      baseTernary,
+      null,
+      emptySet(),
+      'trim-bounds',
+    )
+    const baseTernaryChangedText = baseTernaryText.replace(
+      /<Money /g,
+      `<Money withoutCurrency={false}`,
```
balazsbajorics, line 334: this is neat!

balazsbajorics (approved):

seanparsons (approved):

Rheeseyb (approved):

PR #5309, opened 2024-04-18
Preload package dependencies.

### LINKED ISSUE #5205: When deploying, ensure the Hydrogen dependencies are installed
**Problem:** Cloning new large github projects, and loading large projects (eg the Sample Project) is so slow that it's unclear if the load will succeed. This is due to a number of factors, but a first-load-after-new-deploy is particularly bad, and even worse if multiple people try to do that at the same time. And that's a common scenario.

The reason is that deploys nuke our cache, which not only slows down the first subsequent project load, but also makes multiple people loading the same project even slower.

**So:** Each deploy should install the packages in the sample projects. It's not yet clear where those will live and how many we will need. It is clear that they will change a lot in the next 90 days, including what dependencies they use. The list of `package.json` files to load the dependencies from will come from an environment variable.

seanparsons:
**Problem:**
Cloning new large github projects, and loading large projects (eg the Sample Project) is so slow that it's unclear if the load will succeed. This is due to a number of factors, but a first-load-after-new-deploy is particularly bad, and even worse if multiple people try to do that at the same time. And that's a common scenario.

**Fix:**
Now the server checks for an environment variable `PRELOAD_DEPENDENCIES`, which it splits based on commas expecting URLs for package.json files in the value. The dependencies in the project are then cached as if the editor had requested them.

**Results:**
Testing this locally by timing the point at which a reload of the mock shop project is started until the canvas displays fully:
- Without preloading: `00:34.74`
- With preloading: `00:20.95`

**Commit Details:**
- Moved a few functions relating to the packager from `Common.hs` to `NPM.hs`.
- Added preload trigger to `startup`.
- Extracted out `packageJSONToDependencies` from `getProjectDependenciesFromPackageJSON`.
- Implemented `preloadNPMDependencies` to lookup dependencies and cache them.
- Added `logException` utility function.
- Included `async` project in server.

**Manual Tests:**
I hereby swear that:

Fixes #5205

### DIFF
--- a/server/src/Utopia/Web/Packager/NPM.hs
+++ b/server/src/Utopia/Web/Packager/NPM.hs
@@ -9,31 +9,46 @@
 module Utopia.Web.Packager.NPM where
 
 import           Conduit
-import           Control.Lens              hiding ((.=))
+import           Control.Concurrent.Async
+import           Control.Concurrent.ReadWriteLock
+import           Control.Lens                     hiding ((.=))
 import           Control.Monad.Catch
 import           Data.Aeson
-import qualified Data.ByteString.Lazy      as BL
-import           Data.Conduit.Combinators  hiding (encodeUtf8)
+import qualified Data.ByteString.Lazy             as BL
+import qualified Data.Conduit                     as C
+import           Data.Conduit.Combinators         hiding (encodeUtf8, foldMap,
+                                                   print)
+import qualified Data.Conduit.Combinators         as C hiding (concatMap)
 import           Data.Generics.Product
 import           Data.Generics.Sum
-import qualified Data.HashMap.Strict       as Map
+import qualified Data.HashMap.Strict              as Map
+import qualified Data.HashSet                     as HS
 import           Data.IORef
-import           Data.List                 (isSuffixOf, stripPrefix)
-import           Data.Text                 (pack)
+import           Data.List                        (isSuffixOf, stripPrefix)
+import           Data.String                      (String)
+import           Data.Text                        (pack, splitOn)
 import           Data.Time.Clock
-import           Protolude                 hiding (catch, finally, mapM)
-import           RIO                       (readFileUtf8)
+import qualified Network.Wreq                     as WR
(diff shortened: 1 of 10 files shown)


shell.nix
```diff
@@ -309,7 +309,7 @@ let
         , ("ghc", ["--version"], ["The Glorious Glasgow Haskell Compilation System, version 9.0.2"])
         , ("cabal", ["--version"], ["cabal-install version 3.8.1.0", "compiled using version 3.8.1.0 of the Cabal library "])
         ]
-
+
```
Rheeseyb, line 312:
```suggestion

```

Rheeseyb (approved):

server/src/Utopia/Web/Packager/NPM.hs
```diff
@@ -176,8 +199,77 @@ getProjectDependenciesFromPackageJSON decodedProject = either (const []) identit
...
+getPackageJSONContent packageJSONURL = do
+  response <- WR.get $ toS packageJSONURL
+  either fail pure $ eitherDecode $ view WR.responseBody response
+
+packageJSONToDependencies :: MinimalPackageJSON -> [ProjectDependency]
+packageJSONToDependencies MinimalPackageJSON{..} =
+  let dependenciesMap = fromMaybe mempty (dependencies <> devDependencies)
+  in  Map.foldMapWithKey (\key value -> [ProjectDependency key value]) dependenciesMap
+
+cachePackagerContent :: (MonadResource m, MonadMask m) => PackageVersionLocksRef -> Text -> ConduitBytes m -> IO (ConduitBytes m, UTCTime)
```
ruggi, line 215: there's a non-zero chance that under load this will really stress out the I/O of the actual server storage, but I guess it's a problem for tomorrow

seanparsons: This is the already existing function that we haven't seen an issue with so far, just moved to a different file.

ruggi (approved):

PR #5313, opened 2024-04-18
Navigator html + head fixes

### LINKED ISSUE #5291: Tweak `html` and `head` element handling in Navigator
- The `html` row in the navigator should by default be non-selectable (using the lock icon), so `body` becomes the ‘outermost div’ you can select
- The `head` element should not expand its children by default

ruggi:
**Problem:**

The `html` entry of the navigator should be auto-locked, and the `head` entry should be collapsed by default.

**Fix:**

- Disable the select callback for the `html` entry
- Add the `html` entry check in `updateSimpleLocks`

**Manual Tests:**
I hereby swear that:

Fixes #5291

### DIFF
--- a/editor/src/components/editor/use-default-collapsed-views.tsx
+++ b/editor/src/components/editor/use-default-collapsed-views.tsx
@@ -11,7 +11,7 @@ export function useDefaultCollapsedViews() {
 
   // keep track of the previously collapsed element paths, or they will be re-collapsed
   type PreviouslyCollapsed = { [path: string]: boolean }
-  const [previouslyCollapsed, setPreviouslyCollapsed] = React.useState<PreviouslyCollapsed>({})
+  const previouslyCollapsed = React.useRef<PreviouslyCollapsed>({})
 
   const elementNamesByElementPath = useEditorState(
     Substores.metadata,
@@ -39,7 +39,7 @@ export function useDefaultCollapsedViews() {
         // the collapsed views don't already contain the element
         !collapsedPaths.has(path) &&
         // the collapsed view wasn't already collapsed before
-        previouslyCollapsed[path] == null &&
+        previouslyCollapsed.current[path] == null &&
         // the element is a `head`
         elementNamesByElementPath[path] === 'head'
       )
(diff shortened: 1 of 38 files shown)


editor/src/components/navigator/navigator.tsx
```diff
@@ -323,3 +328,53 @@ export const NavigatorComponent = React.memo(() => {
...
+
+  React.useEffect(() => {
+    const newCollapsedViews = Object.keys(elementNamesByElementPath).filter((path) => {
+      return (
+        // the collapsed views don't already contain the element
+        !collapsedPaths.has(path) &&
+        // the collapsed view wasn't already collapsed before
+        previouslyCollapsed[path] == null &&
+        // the element is a `head`
+        elementNamesByElementPath[path] === 'head'
```
liady, line 367: won't that result only in the "head" path (i.e an array with max length of 1)?

ruggi: If you mean it will only contain the `<head>` path of `<head><foo /><bar /></head>` (and not `foo` and `bar`), yes but that's desired.
Or am I misunderstanding and you mean something else?

editor/src/components/editor/use-default-collapsed-views.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { isJSXElement } from '../../core/shared/element-template'
+import * as EP from '../../core/shared/element-path'
+import { addCollapsedViews } from './actions/action-creators'
+
+export function useDefaultCollapsedViews() {
+  const dispatch = useDispatch()
+
+  // keep track of the previously collapsed element paths, or they will be re-collapsed
+  type PreviouslyCollapsed = { [path: string]: boolean }
+  const [previouslyCollapsed, setPreviouslyCollapsed] = React.useState<PreviouslyCollapsed>({})
```
balazsbajorics, line 14: could this be a React.useRef? with the current value updated in the useEffect

ruggi: I guess it can! sec

balazsbajorics (approved):

liady (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -739,18 +740,40 @@ export const NavigatorItem: React.FunctionComponent<
...
+      const isHTMLTagElement =
+        element != null &&
+        isRight(element.element) &&
+        isJSXElement(element.element.value) &&
+        element.element.value.name.baseVariable === 'html'
+
+      return !isHTMLTagElement
+    },
+    'NavigatorItem canBeSelected',
+  )
```
Rheeseyb, line 760: Since we already have the functionality, we should be using the locking mechanism for this. You can add a check into `updateSimpleLocks` <https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/shared/element-locking.ts#L8-L33>

Rheeseyb (approved):

PR #5321, opened 2024-04-18
Ellipse too-long project titles through grid powers

maltenuhn:
**Problem:**

**Fix:**
Why does this work? Well, for ellipsis to kick in you need three things:

1. `text-overflow: ellipsis`
2. `overflow: hidden`
3. _and_ a fixed-width container! Which flex containers are not. But grid columns _are_ - even if they are set to dynamic widths like `1fr`. So, use that (and make the last (optional) column with github buttons collapse if empty.

Large GIF (544x444)

### DIFF
--- a/editor/src/components/titlebar/title-bar.tsx
+++ b/editor/src/components/titlebar/title-bar.tsx
@@ -58,7 +58,6 @@ const ProjectTitle: React.FC<React.PropsWithChildren<ProjectTitleProps>> = ({ ch
         whiteSpace: 'nowrap',
         textOverflow: 'ellipsis',
       }}
-      css={undefined}
     >
       {children}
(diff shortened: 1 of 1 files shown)


lankaukk (commented):
nice! But can you show the ellipses for long branch titles too? When the project is connected to a repo

maltenuhn:
> nice! But can you show the ellipses for long branch titles too? When the project is connected to a repo

Unlesss I'm mistaken, this PR already does that (but I can't verify it, since trying to log in with github locally doesn't work for me)

balazsbajorics (approved):

bkrmendy (approved):

seanparsons (approved):

PR #5334, opened 2024-04-19
EditorComponentInner viewer mode  useSelectorWithCallback  only dispa…

balazsbajorics:
**Problem:**
Dispatch is called from useSelectorWithCallback. Since out dispatch flow calls React.flushSync, we get a big warning from React:
image

In general, we shouldn't call dispatch in code which was called because of a dispatch, this kind of recursive dispatching makes debugging the dispatch flow very hard.

**Fix:**
As the react warning tells us, we can move the code to a [microtask](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask). It could be a setTimeout(0) or await wait(0), but I think we can safely use queueMicrotask for this job for more explicitness!

### DIFF
--- a/editor/src/components/editor/editor-component.tsx
+++ b/editor/src/components/editor/editor-component.tsx
@@ -407,18 +407,20 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
     Substores.userStateAndProjectServerState,
     (store) => ({ projectServerState: store.projectServerState, userState: store.userState }),
     (state) => {
-      let actions: EditorAction[] = []
-      const permissions = getPermissions(state)
-      if (!permissions.edit && permissions.comment) {
-        actions.push(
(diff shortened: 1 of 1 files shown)


liady (approved):
Do we have flows that rely on the actions to be done synchronously? (for example another synchronous code running afterwards that expects to already see the result of the dispatched actions)

balazsbajorics:
> Do we have flows that rely on the actions to be done synchronously? (for example another synchronous code running afterwards that expects to already see the result of the dispatched actions)

the main one is the inside of the dispatch flow which flushSync renders the canvas, runs the dom-walker, then flushSync renders the rest of the editor. that's why we have the flushSync and the limitation

but since I introduced that a few years ago, we did add some rare dispatches that rely on the fact that the dispatch function returns synchronously

seanparsons (approved):

PR #5343, opened 2024-04-21
When renaming an navigation entry, don't show buttons

### LINKED ISSUE #5323: while renaming a navigator entry, don't show the lock / hide / add buttons
Problem:
Image

Solution:
While editing, hide them! They are not needed when editing! Of what use are they while you are looking at your keyboard anyway, thinking about the next letter you want, and then trying to find the key that will reveal it. Plus, there are already a million conditions for what to show when where in `navigator-item`, what's one more?

liady:
When renaming a navigation entry - don't show action buttons.
Fixes #5323 
<table>
<tr>
<td>
<b>Before:</b>
<video style="display:block" src="https://github.com/concrete-utopia/utopia/assets/7003853/57e5189c-c84c-4d4b-a680-f937bd4a9256" />
</td>
<td>
<b>After:</b>
<video style="display:block" src="https://github.com/concrete-utopia/utopia/assets/7003853/4b7ab18c-0ed6-4143-90d3-1726728aa126" />
</td>
</tr>
</table>

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item.tsx
@@ -1034,11 +1034,6 @@ export const NavigatorRowLabel = React.memo((props: NavigatorRowLabelProps) => {
     ? colorTheme.componentPurple.value
     : undefined
 
-  const currentlyRenaming = useMemo(
-    () => EP.pathsEqual(props.renamingTarget, props.navigatorEntry.elementPath),
-    [props.renamingTarget, props.navigatorEntry.elementPath],
-  )
-
   return (
     <div
       style={{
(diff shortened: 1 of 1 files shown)


bkrmendy (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -844,6 +844,11 @@ export const NavigatorItem: React.FunctionComponent<
     ? 'component'
     : resultingStyle.iconColor
 
+  const currentlyRenaming = useMemo(
+    () => EP.pathsEqual(props.renamingTarget, props.navigatorEntry.elementPath),
+    [props.renamingTarget, props.navigatorEntry.elementPath],
+  )
+
```
gbalint, line 851: I'm not sure it is worth using useMemo for this simple equality check, but if have this, let's use it in line 1060 too, to pass it in as a prop to ItemLabel.

gbalint (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
balazsbajorics: nope it's not important to use a useMemo for this! 
1. the result is a boolean so it will be value compared so we don't need to worry about losing reference equality
2. ElementPath is very heavily memoized and pathsEqual is probably faster than react's memo check

balazsbajorics (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
liady: good point, removed the unnecessary `useMemo`

liady:
When renaming a navigation entry - don't show action buttons.
Fixes #5323

PR #5345, opened 2024-04-22
In view-only mode, do not show the code editor

bkrmendy:
# [Try it here](https://utopia.fish/p/619f6880-accidental-munchkin/?branch_name=feature-do-not-show-github-panel-logged-out)

## Problem
The code editor is shown for viewers.

## Fix
If a user doesn't have edit permission, the code editor panel is filtered from the available grid panels.

Since we don't support more than 1 editor yet, this PR doesn't handle adding back the code editor if a user is somehow granted edit permissions

**Manual Tests:**
I hereby swear that:

Fixes https://github.com/concrete-utopia/utopia/issues/5327

### DIFF
--- a/editor/src/components/canvas/grid-panels-state.tsx
+++ b/editor/src/components/canvas/grid-panels-state.tsx
@@ -74,9 +74,9 @@ export function useGridPanelState() {
   const permissions = usePermissions()
   React.useEffect(() => {
     if (!permissions.edit) {
-      setState(filterCodeEditorFromPanelState(state))
+      setState((s) => filterCodeEditorFromPanelState(s))
     }
-  }, [permissions.edit, setState, state])
+  }, [permissions.edit, setState])
 
   return stateAtom
 }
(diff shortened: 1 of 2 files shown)


ruggi (approved):

editor/src/components/canvas/grid-panels-state.tsx
```diff
@@ -63,6 +71,13 @@ export function useGridPanelState() {
     void saveUserPreferencesProjectLayout(projectId, state)
   }, [loaded, state, projectId])
 
+  const permissions = usePermissions()
+  React.useEffect(() => {
+    if (!permissions.edit) {
+      setState((s) => filterCodeEditorFromPanelState(s))
```
balazsbajorics, line 77: won't this mean the code editor is re-openable manually?

bkrmendy: No, the closed panels UI is already using this hook as the source of truth

balazsbajorics (approved):

PR #5369, opened 2024-04-23
Fix vite hosts for local dev via BFF

ruggi:
**Problem:**

Hot reload in local development is glitchy.

**Fix:**

Use `127.0.0.1` instead of `0.0.0.0` for local development. (https://stackoverflow.com/questions/70694187/vite-server-is-running-but-not-working-on-localhost)

**Manual Tests:**
I hereby swear that:


balazsbajorics (approved):
wat

Rheeseyb (approved):

PR #5374, opened 2024-04-23
Support replacing an element in the navigator

Rheeseyb:
**Problem:**
We want users to be able to add placeholders in their insertion variants, and to then replace those placeholders with real elements via the Navigator

**Fix:**
To achieve this, I've tweaked the `InsertJSXElement` action so that it can now specify whether to replace the target or insert the new element as a child of it. I've also then made a similar change to the component insertion context menu.

Awkwardly, because of the way `react-contexify` requires context menus to be created with all props upfront, and then rendered via a hook that just uses the id of the menu, this meant I've had to create 2 almost identical context menus (one to insert as a child, one to replace the target). 

I've also added a button to the Navigator for showing the picker with the `replace-target` functionality.

https://github.com/concrete-utopia/utopia/assets/1044774/f2f4ed1d-b7fb-4d3f-95d5-b7bb9fbe09c6


editor/src/components/navigator/navigator.spec.browser2.tsx
```diff
@@ -4757,8 +4757,8 @@ describe('Navigator', () => {
           'regular-utopia-storyboard-uid/scene-aaa/app-entity:root/new-container',
           'regular-utopia-storyboard-uid/scene-aaa/app-entity:root/old-container',
           'regular-utopia-storyboard-uid/scene-aaa/app-entity:root/old-container/wrapper',
-          'regular-utopia-storyboard-uid/dragme1',
           'regular-utopia-storyboard-uid/dragme2',
+          'regular-utopia-storyboard-uid/dragme1',
```
Rheeseyb, line 4761: I have no idea why the order of the derived targets has changed here, but I've tested this interaction on this project in the editor and the navigator (and resulting code) was ordered correctly...

editor/src/components/editor/actions/actions.spec.browser2.tsx
```diff
@@ -82,6 +82,7 @@ import { getDomRectCenter } from '../../../core/shared/dom-utils'
 import { FloatingPostActionMenuTestId } from '../../canvas/controls/select-mode/post-action-menu'
 import { safeIndex } from '../../../core/shared/array-utils'
 import { updateSelectedViews } from '../../canvas/commands/update-selected-views-command'
+import { wait } from '../../../core/model/performance-scripts'
```
ruggi, line 85:
```suggestion
```

ruggi (approved):

PR #5386, opened 2024-04-24
Datapicker improvements

balazsbajorics:
This is a grab bag of a few UX improvements for the data picker

- improved styling
- "Preferred" mode only shows local scoped data, filters out enclosing outer scopes
- Also matching data based on name! for example, when picking data for `props.title`, `review.title` will be considered a Best Match

Before:
image

After: 
image


editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
```diff
@@ -60,6 +75,21 @@ export interface JSXOption {
...
+    case 'all':
+      return 'All'
+    case 'preferred':
+      return 'Preferred'
+    default:
+      assertNever(mode)
+  }
+}
+
+export const DataPickerPreferredAllAtom = atom<DataPickerFilterOption>('preferred')
```
seanparsons, line 91: Why is this in an atom? Just to persist between showings of the inspector?

editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
```diff
@@ -310,9 +336,40 @@ const filterObjectPropFromVariablesInScope =
...
+        deepestInsertionCeiling,
+        EP.fullDepth(variable.insertionCeiling),
+      )
+    })
+
+    const result: VariableData = {}
+    Object.entries(variablesInScope).forEach(([key, variable]) => {
+      if (variable.insertionCeiling == null) {
+        result[key] = variable
+        return
```
seanparsons, line 358: Why not `else if` after this `if`?

editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
```diff
@@ -224,32 +225,54 @@ function variableInfoFromVariableData(variableNamesInScope: VariableData): Array
   return info
 }
 
+function keepOnlyFirstMatch(variables: Array<VariableInfo>): Array<VariableInfo> {
+  let foundFirstMatch: boolean = false
+  let result: Array<VariableInfo> = []
+  for (const variable of variables) {
+    const matches = foundFirstMatch ? false : variable.matches
+    result.push({ ...variable, matches })
+    if (variable.matches) {
```
ruggi, line 234: I think it would be more readable to just check `matches` instead of `variable.matches` (and maybe move it right after declaring it)

ruggi (approved):

editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
balazsbajorics: yep to persist

editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
bkrmendy: The suggestion makes sense, let's implement it as a follow-up improvement (for posterity: at the time of writing we were under time pressure to merge this PR)

PR #5402, opened 2024-04-25
Editable Remix URL In Navigation Bar.

### LINKED ISSUE #5325: Make the URL bar capable of typing into, and add the star button there, too
Make this editable:
image

Add in the star for making it a favourite like this:
image

seanparsons:
**Problem:**
We want the URL bar for Remix to function a lot more like the URL bar in a browser, supporting editing and setting favourites directly from the URL bar. 

**Fix:**
The groundwork for this change was the following:
- Separating out the control for marking favourites into its own file.
- Customising `StringInput`: Firstly to expand automatically with the content so that it imitates the label that was used previously; Secondly to visually imitate the previous look by hiding the box shadows that ordinarily are shown.

The star control needed very little additional changes as it's mostly self contained once supplied with a URL.

Some more complicated handling for various cases were added when committing the URL, as it may be invalid or non-existent in the current project.

**Commit Details:**
- Added handler functions to `RemixNavigationBar` that manage updates and committing the URL to the now editable field.
- Extracted out `StarUnstarIcon` component into its own file.
- Added a small fix to `updatePackageJsonInEditorState` to increment the version number of the `package.json` file.
- `RemixNavigationBar` now uses `StringInput` to control the URL.
- `StarUnstarIcon` also included in `RemixNavigationBar`.
- Added a `growInputAutomatically` option to `StringInput` which creates an invisible span which is used to gauge the size that the input field should be.
- Added `includeBoxShadow` option to `StringInput` so that the box shadows can be optionally hidden.
- `testId` is made mandatory in `HeadlessStringInput`.

**Manual Tests:**
I hereby swear that:

Fixes #5325


balazsbajorics:
(I'll make a follow-up ticket for this) the text field has broken copy-paste, which I think is pretty important as URL writing is frequently done using cut-copy-paste (one of our global event listeners seem to be the culprit)

editor/src/components/canvas/starunstaricon.tsx
```diff
@@ -0,0 +1,82 @@
+/** @jsxRuntime classic */
```
balazsbajorics, line 1: minor: could you rename this to star-unstar-icon.tsx please?

editor/src/components/editor/remix-navigation-bar.tsx
```diff
@@ -45,6 +66,53 @@ export const RemixNavigationBar = React.memo(() => {
 
   const label = getRemixUrlFromLocation(navigationControls[EP.toString(activeRemixScene)]?.location)
```
balazsbajorics, line 67: we should probably rename this from label to `activeUrl` or `activeRemixUrl`

editor/src/components/editor/remix-navigation-bar.tsx
```diff
@@ -45,6 +66,53 @@ export const RemixNavigationBar = React.memo(() => {
...
+  }, [label])
+
+  const onInputChange = React.useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+    setCurrentURL(e.target.value)
+  }, [])
+
+  const changeRemixURLInEditor = React.useCallback(
+    (url: string) => {
+      try {
+        const baseDomain = `domainthatisveryunlikelytoberandomlyused.com`
```
ruggi, line 84: nice trick :D

ruggi (approved):

editor/src/components/editor/remix-navigation-bar.tsx
balazsbajorics: image

🤔

editor/src/components/editor/remix-navigation-bar.tsx
seanparsons: Fixed in 247525474.

editor/src/components/editor/remix-navigation-bar.tsx
seanparsons: Stop it! :)

Issue #5405, opened 2024-04-26
we don't show the canvas flexGap control between generated items


bkrmendy:
**Problem** 
The on-canvas gap control doesn't show up between items that are generated, even if those are not in a fragment

Code
```
<div
        style={{
          height: 'max-content',
          position: 'absolute',
          left: 48,
          top: 112,
          display: 'flex',
          flexDirection: 'row',
          width: 'max-content',
          gap: 29,
        }}
      >
     <div>MyComponent</div>
        {[1, 2, 3].map(({ a, b, c }) => (
          <img
            src='https://github.com/concrete-utopia/utopia/blob/master/editor/resources/editor/pyramid_fullsize@2x.jpg?raw=true'
            alt='Utopia logo'
            style={{ width: 118, height: 150 }}
          />
        ))}
        <View
          src='https://github.com/concrete-utopia/utopia/blob/master/editor/resources/editor/pyramid_fullsize@2x.jpg?raw=true'
          alt='Utopia logo'
          style={{ width: 118, height: 150 }}
        />
        <img
          src='https://github.com/concrete-utopia/utopia/blob/master/editor/resources/editor/pyramid_fullsize@2x.jpg?raw=true'
          alt='Utopia logo'
          style={{ width: 118, height: 150 }}
        />
      </div>
```

![Image](https://github.com/concrete-utopia/utopia/assets/2945037/8aeefed7-dd4c-4077-9648-5b87182a9c38)

PR #5407, opened 2024-04-26
show the flex gap control in generated elements

bkrmendy:
# [Try it here](https://utopia.fish/p/3b3ccd0b-citrine-coaster/?branch_name=fix-flex-gap-control-generated-elements)

https://github.com/concrete-utopia/utopia/issues/5405

## Problem
The flex gap handle doesn't show up for generated elements

## Fix
Make it show up for generated elements by recursing into any map/fragment valued children of flex containers when checking whether there are enough children to show a gap control for

### Manual Tests
I hereby swear that:


editor/src/components/canvas/gap-utils.ts
```diff
@@ -197,3 +189,30 @@ export function maybeFlexGapFromElement(
...
+): Array<ElementInstanceMetadata> {
+  let result: Array<ElementInstanceMetadata> = []
+  for (const elementKey in elements) {
+    const element = elements[elementKey]
+    const elementPath = element.elementPath
+    if (EP.isChildOf(elementPath, target) && !EP.isRootElementOfInstance(elementPath)) {
+      const isMapOrFragment = defaultEither(
+        false,
+        mapEither(
+          (e) => e.type === 'JSX_MAP_EXPRESSION' || e.type === 'JSX_FRAGMENT',
```
balazsbajorics, line 205: I'd use `treatElementAsFragmentLike` (with sizeless-div-**not**-considered-fragment-like as the flag) instead of special casing _just_ map and fragment, so the same fix can apply to ternaries and in the future more stuff

ruggi (approved):
aside from @balazsbajorics 's comment on `recurseIntoChildrenOfMapOrFragment`, but everything else looks good from my pov

balazsbajorics (approved):

PR #5421, opened 2024-04-29
Make swap and add use the component picker

Rheeseyb:
**Problem**
We want to replace all uses of the insert menu with the new component picker.

**Fix**
This PR updates the "Add Element" and "Swap Element With" context menu options (and their respective shortcuts) to use the new component picker. To do that, I've had to rejig the hooks used to show the context menu to now return a callback which can be called with the `target` and `insertionTarget`, as the places calling this are unable to use hooks. 

This PR doesn't touch the "Wrap In" functionality, as that requires extra work to ensure that existing children aren't lost during the process.

**Note:**
As mentioned in #5427 the component picker does not yet support keyboard interactions, and so bunch of existing tests have been disabled with this PR. Once that issue has been resolved we can re-enable those tests.


gbalint (approved):

editor/src/components/context-menu-items.ts
```diff
@@ -341,32 +343,29 @@ export const insert: ContextMenuItem<CanvasData> = {
...
-    requireDispatch(dispatch)([
-      setFocus('canvas'),
-      EditorActions.openFloatingInsertMenu({
-        insertMenuMode: 'insert',
-        parentPath: null,
-        indexPosition: null,
-      }),
-    ])
+  action: (data, _dispatch, _coord, event) => {
+    // FIXME Render prop support
```
balazsbajorics, line 347: is this fixme for a future PR?

Rheeseyb: Crud, this was a reminder for me to fix this as part of this PR. Though technically render prop support isn't something that would make sense for "Add Element", it certainly is required for "Swap Element"

Rheeseyb: Ok change of plan, I'm going to merge this PR as is and create follow up issues so that the remaining work here can be parallelised

Issue #5424, opened 2024-04-29
Allow insertion of non-jsxelements (conditionals, maps, fragments) from the preferredContents in the component annotations.


gbalint:
We already allow conditionals/maps in the component picker, but it still doesn't work from the context menu, picking from the preferredContents from the component annotations.

PR #5431, opened 2024-04-30
Non-jsx element insertion from the context menu for preferredContents

gbalint:
**Problem:**
(https://github.com/concrete-utopia/utopia/pull/5417) introduced inserting non-jsxelements (e.g. conditionals) from the component picker.
However, non-jsxelements still can not be added from preferredContents using the context menu.

**Fix:**
First of all, non-jsxelements are not even valid according to the component descriptor parser, so I extended those to all `ComponentElementToInsert` (which is fragment, jsx elements, conditionals and maps)

I also unified the preferredContent and component picker insertion by converting ElementToInsert objects to ComponentElementToInsert ones.

**Manual Tests:**

Fixes https://github.com/concrete-utopia/utopia/issues/5424

### DIFF
--- a/editor/src/core/property-controls/property-controls-local.spec.tsx
+++ b/editor/src/core/property-controls/property-controls-local.spec.tsx
@@ -1476,6 +1476,83 @@ describe('registered property controls', () => {
         }
       `)
     })
+    it('can use conditional for children', async () => {
+      const renderResult = await renderTestEditorWithModel(
+        project({
+          ['/utopia/components.utopia.js']: `import { Card } from '../src/card'
+          import { Card2 } from '../src/card2'
+          
+          const Components = {
+        '/src/card': {
+          Card: {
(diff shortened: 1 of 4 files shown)


ruggi (approved):

editor/src/core/property-controls/property-controls-local.spec.tsx
```diff
@@ -1476,6 +1476,83 @@ describe('registered property controls', () => {
...
+        '/src/card': {
+          Card: {
+            component: Card,
+            properties: { },
+            children: {
+              preferredContents: [
+                { component: 'span', variants: { name: 'span' } },
+                {
+                  component: 'Conditional',
+                  moduleName: '',
```
bkrmendy, line 1495: this can be omitted

bkrmendy (approved):

PR #5438, opened 2024-04-30
Iconic fixes

lankaukk:
- Fixing some icons that were incorrectly named by removing hyphens from the colors
- Uploading and using new `convert-action` icons in the navigator, that are the proper size and in more colors

Look at that! the convert icon button is _entirely blue_
Screenshot 2024-04-30 at 11 21 35 AM


editor/src/components/navigator/navigator-item/navigator-item-components.tsx
```diff
@@ -250,12 +250,11 @@ const ReplaceElementButton = React.memo((props: ReplaceElementButtonProps) => {
       }}
     >
       <Icn
-        category='tools'
+        category='navigator-element'
```
maltenuhn, line 253: Eventually not all new icons should live in navigator-element (that's for _element_ icons). Not something to fix now, because we don't have a good system in place today either.

maltenuhn (approved):

seanparsons (approved):

PR #5445, opened 2024-04-30
Map expressions change to support swapping the source.

seanparsons:
**Problem:**
We want to be able to swap over the source (i.e. the left hand side of) a map expression like this:
```typescript
orders.map(order => <Order order={order} />)
```

**Fix:**
The main change is parsing a bit more of the structure of a map expression into:
- The value/expression on which `map` is being called.
- The function which transforms the contents.
Whereas previously the entire expression was treated as effectively like a anonymous JS block, with some specialisation for map expressions.

The parser now checks to see if an expression can be treated as a map expression first instead of the more general case, with some more specialised logic for getting the fields that need parsing.

**Note:**
Because this was turning into a monster piece of work, this provides the underlying infrastructure to be able to implement the UI for swapping the source in subsequent work.

**Commit Details:**
- Changed representation of `JSXMapExpression` to have `valueToMap` and `mapFunction` fields, removing the prior fields holding strings of JavaScript.
- Added (as yet unused) `UPDATE_MAP_EXPRESSION` editor action.
- Updated a lot of utility functions to specifically handle map expressions.
- Removed `JSXMapExpression` from the type of `JavaScriptContainer`.
- Removed the `OtherJavaScriptType` type and related code from the parser.
- Added `parseOutMapExpressionParts`, which attempts to parse out the various important fields for a map expression.
- `parseJSExpressionMapOrOtherJavascript` uses `parseOutMapExpressionParts` first to see if the expression can be treated as our map expression representation.
- Added `mapParserMetadata` utility function.

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1748,16 +1751,18 @@ function getValidElementPathsFromElement(
   uiFilePath: string,
   isOnlyChildOfScene: boolean,
   parentIsInstance: boolean,
+  includeElementInPath: boolean,
```
balazsbajorics, line 1754: boo for one undecypherable more boolean flag

seanparsons: Yeah, that should be changed.

balazsbajorics (approved):

Rheeseyb (approved):

PR #5447, opened 2024-04-30
Store render prop name in metadata

### LINKED ISSUE #5440: Support render props in S to Swap
Follow up from [this FIXME](https://github.com/concrete-utopia/utopia/pull/5421/files#diff-5b1943a23fbb0693f2e514b4ba72d79d93d022c65b119c517aca465fd2463b5fR367) in #5421

Rheeseyb:
**Problem**
The new component insert menu needs to know if it is targeting a render prop (and which) in order to provide suitable options and to actually replace the target. This was fine when it was only invoked from the Navigator (as the navigator items included this information), but not when it is invoked from the canvas or keyboard shortcuts, as this information isn't readily available.

**Fix**
To get around this I have now introduced the name of the prop that the element is assigned to in its metadata as a `string | null`. We capture this by providing it to the spy wrapper at render time, and then look it up before showing the component picker via the keyboard shortcut or canvas context menu.

I appreciate that the name used here `prop` isn't great, so I'm open to alternatives there.

Fixes #5440


editor/src/core/shared/element-template.ts
```diff
@@ -2533,6 +2533,7 @@ export interface ElementInstanceMetadata {
   conditionValue: ConditionValue
   textContent: string | null
   earlyReturn: EarlyReturn | null
+  prop: string | null
```
gbalint, line 2536: I think prop is not a very helpful name, especially because a jsx element has props itself.
Can this be something more verbose and clear? E.g. containingRenderProp or something like that?

Rheeseyb: Yeah as mentioned I know the name is bad. I'm going to go with `assignedToProp` (or `assignedToRenderProp` perhaps) and see how that feels

Rheeseyb: Ok done - I went with `assignedToProp`. How do you feel about that?

gbalint: `assignedToProp` is fine by me!

gbalint (approved):

bkrmendy (approved):

ruggi (approved):

PR #5456, opened 2024-05-02
Paste handlers for floating insert menu and remix url bar

ruggi:
Continuing from https://github.com/concrete-utopia/utopia/pull/5439, this PR makes the floating insert menu's search field and the remix url bar inputs able to receive paste events.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/navigator/navigator-item/component-picker.tsx
+++ b/editor/src/components/navigator/navigator-item/component-picker.tsx
@@ -14,6 +14,7 @@ import { assertNever } from '../../../core/shared/utils'
 import { insertableComponent } from '../../shared/project-components'
 import type { StylePropOption, InsertableComponent } from '../../shared/project-components'
 import type { Size } from '../../../core/shared/math-utils'
+import { dataPasteHandler } from '../../../utils/paste-handler'
 
 export interface ComponentPickerProps {
   allComponents: Array<InsertMenuItemGroup>
@@ -176,6 +177,7 @@ const FilterBar = React.memo((props: FilterBarProps) => {
       onChange={handleFilterChange}
       value={filter}
       data-testId={componentPickerFilterInputTestId}
+      {...dataPasteHandler(true)}
     />
   )
 })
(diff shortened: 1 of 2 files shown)


liady (approved):

editor/src/components/navigator/navigator-item/component-picker.tsx
```diff
@@ -176,6 +177,7 @@ const FilterBar = React.memo((props: FilterBarProps) => {
       onChange={handleFilterChange}
       value={filter}
       data-testId={componentPickerFilterInputTestId}
+      {...dataPasteHandler(true)}
```
gbalint, line 180: Is there a reason to use a helper in one of the cases, and just put the prop there in the other case?

ruggi: because this is not a `StringInput` but a regular `input` :)

Rheeseyb: tbh I'm not sure it's worth using this helper rather than just inlining `[DataPasteHandler]: true` each time, since at the very least this is going to be recreating and spreading an object, which I _think_ would be cause issues with any `React.memo` components 🤔

liady: just a thought - maybe we can have higher level components (such as StringInput) just pass all the `data-` props down (and use only `data-pasteHandler`?

ruggi: tbh if we prefer not to use the spread (I did not know it could cause issues with memoization!) we can/should just use `data-pastehandler` directly :/

liady: @Rheeseyb I think that React.memo checks equality of the props themselves, regardless of how they were created

Rheeseyb: Oh sorry of course it won't mess with the memoization!

Rheeseyb (approved):

Issue #5441, opened 2024-04-30
Replacing an element should in some cases retain existing children


Rheeseyb:
The component picker (<https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/navigator/navigator-item/component-picker-context-menu.tsx> and <https://github.com/concrete-utopia/utopia/blob/master/editor/src/components/navigator/navigator-item/component-picker.tsx>) when opened with the insertion target set to `'replace-target'` will completely replace the target, children and all. Sometimes we want this, sometimes we don't (and we are yet to fine tune which we want based on the circumstances).

We need to create two different types of "replace-target":

1. The current implementation, `replace-target`, which completely replaces the target and its subtree
2. A more surgical implementation, `replace-target-keep-children-and-style` (verbose, I know, but we can change that), which will only replace the target element itself, but will keep the existing children and `style` prop.

- Note that with the second implementation:
  - If the target already has children
    - we should probably filter out components that don't support children
    - we should also probably filter out variants that would add children
  - If the target has styling, we should probably filter out components that don't support styling
- We can expose this second implementation by adding another button to the canvas and navigator context menu.
- Right now we believe that the existing "replace" Navigator button should go with the first implementation, replacing any children

Doing this will allow the use of nested placeholder code, as well as drawing to insert `div`s before later converting them to the actual desired components.

**UI affordances**

- [ ] Context menu: Replace This - replaces target only, keeps children / style
- [ ] Context menu: Replace Everything - replaces target and children
- [ ] Row button: Replace Everything (nb might change because 'delete manually and add' is the equivalent
- [ ] Delete / disable the 'S for Swap' keyb shortcut

Rheeseyb:
@maltenuhn please can you confirm this behaviour. I believe this is what we want in all cases, but maybe this is just what we want under certain circumstances

Rheeseyb:
I've updated the description and removed the Blocked label

PR #5463, opened 2024-05-02
New swap feature to keep children and style

gbalint:
**Problem:**
See https://github.com/concrete-utopia/utopia/issues/5441

**Fix:**
I added a new context menu item called ('Replace Element With'). This includes the old behavior! And the existing 'Swap Element With' triggers this new behavior: it keeps style and children of the original element.

The replace button in the navigator item keeps the old behavior (and replaces the whole subtree).

The S (swap) keyboard shortcut is removed (suggested by Malte).

This PR does not include any kind of filtering suggested in the issue:
image

These can come in subsequent PR(s)

**Manual Tests:**
I hereby swear that:

Fixes https://github.com/concrete-utopia/utopia/issues/5441

### DIFF
--- a/editor/src/components/context-menu-items.ts
+++ b/editor/src/components/context-menu-items.ts
@@ -366,6 +366,19 @@ export function showReplaceComponentPicker(
   return showComponentPicker(target, insertionTarget)
 }
 
+export function showSwapComponentPicker(
+  targetElement: ElementPath,
+  jsxMetadata: ElementInstanceMetadataMap,
+  showComponentPicker: ShowComponentPickerContextMenuCallback,
+): ShowComponentPickerContextMenu {
+  const element = MetadataUtils.findElementByElementPath(jsxMetadata, targetElement)
+  const prop = element?.assignedToProp
+  const target = prop == null ? targetElement : EP.parentPath(targetElement)
+  const insertionTarget: InsertionTarget =
+    prop == null ? EditorActions.replaceKeepChildrenAndStyleTarget : renderPropTarget(prop)
+  return showComponentPicker(target, insertionTarget)
+}
+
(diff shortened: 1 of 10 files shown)


editor/src/components/context-menu-items.ts
```diff
@@ -366,7 +366,27 @@ export function showReplaceComponentPicker(
...
+      data.selectedViews.every((path) => {
+        const element = MetadataUtils.findElementByElementPath(data.jsxMetadata, path)
+        return (
+          element != null && isRight(element.element) && isJSXElementLike(element.element.value)
+        )
+      })
+    )
+  },
+  action: (data, _dispatch, _coord, event) => {
+    // FIXME Render prop support
```
Rheeseyb, line 382: Render prop support is now possible, so you can follow a similar pattern to that used below in "Replace Element With"

editor/src/components/navigator/navigator-item/component-picker-context-menu.tsx
```diff
@@ -127,12 +144,12 @@ export function preferredChildrenForTarget(
 
   // TODO: we don't deal with components registered with the same name in multiple files
   if (registeredComponent != null) {
-    if (insertionTarget === 'insert-as-child' || insertionTarget === 'replace-target') {
+    if (isChildInsertionTarget(insertionTarget) || isReplaceInsertionTarget('replace-target')) {
```
Rheeseyb, line 147: I know that this is soon going to be replaced after merging with my PR, but please can you double check that this tautology is gone!
```
isReplaceInsertionTarget('replace-target')
```

bkrmendy (approved):

Rheeseyb (approved):

PR #5466, opened 2024-05-02
Support Switching The Source Of A Map

### LINKED ISSUE #5391: Map / List: Replace data source using Inspector Data Picker
Success Criteria / demo scenario: 

```
export default function ExampleComponent() {
  const { reviews, recommendedProducts } = useLoaderData()
  return (
    <>
      {[1, 2, 3].map((v) => {
        return <Placeholder />
      })}
    </>
  )
}
```

Select the map in the navigator.
Using the inspector, pick a different data source - `reviews` from `const { reviews } = useLoaderData()` 
The code is now `reviews.map((v) => { return <Placeholder />})`

seanparsons:
**Problem:**
We want to be able to swap the source of a map expression, using their newly refactored representation.

**Fix:**
To support this change, a few bits of logic have been adapted to cater for the more specialised case of updating the source of the map expression. So as to work for both the property of an element case and for the map expression source case.

**Commit Details:**
- Added `variablesInScope` parameter to `addFakeSpyEntry` to be included
  in the metadata.
- Changed `expression` on `UpdateMapExpression` to take a `JSExpression`.
- `DataPickerPopup` now takes a `pickerType` to distinguish between differing
  types of targets for the picker.
- Extracted out `DataPickerPopupSubvariables` component.
- `ValueRow` also now uses `pickerType`.
- Functions like `usePropertyControlDescriptions` now support an optional property.
- Added `ListSection` component to handle the map expressions.
- `CodeElementSection` doesn't show anymore for map expressions.

**Manual Tests:**
I hereby swear that:

Fixes #5391


editor/karma-base.conf.js
```diff
@@ -44,6 +44,10 @@ module.exports = function (config) {
     basePath: '',
 
     browserNoActivityTimeout: 1000000,
+    pingTimeout: 15000,
```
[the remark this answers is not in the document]
seanparsons: Fixed in 79da2f3bb.

editor/src/components/inspector/code-element-section.spec.browser2.tsx
```diff
@@ -117,18 +117,12 @@ describe('The Inspector code element section', () => {
...
   })
 
-  it('Is the only section displayed when a map is selected', async () => {
-    const editor = await renderTestEditorWithCode(testProjectCode, 'await-first-dom-report')
-    await clickElementWithTestId(editor, 'ddd')
-
-    // Check that the code element section is displayed
-    const codeElementSection = editor.renderedDOM.getByTestId(CodeElementSectionTestId)
-    expect(codeElementSection).toBeDefined()
+  // it('Is not displayed when a map is selected', async () => {
```
Rheeseyb, line 120: Does this test still need to be commented out?

seanparsons: Probably shouldn't be.

editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
```diff
@@ -395,7 +398,11 @@ export function useVariablesInScopeForSelectedElement(
       // if the selected element doesn't have an associacted variablesInScope, we assume it's safe to walk up the path within the same component
       let path = selectedViewPath
       while (
-        EP.pathsEqual(EP.getContainingComponent(path), EP.getContainingComponent(selectedViewPath))
+        EP.pathsEqual(
```
Rheeseyb, line 401: I'm like 90% sure we used to have a function which would create all of the ancestor paths of a given element path. Maybe we deleted that at some point, but that sounds much safer than using a `while` loop here

Rheeseyb (approved):

editor/src/components/inspector/code-element-section.spec.browser2.tsx
seanparsons: Restored in 83ae48b86.

editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
ruggi: what about using `EP.getAncestors` (or possibly `findAmongAncestorsOfPath`) and filtering over those results?

editor/src/components/inspector/sections/layout-section/list-section.tsx
```diff
@@ -0,0 +1,308 @@
...
+    control: 'array',
+    propertyControl: {
+      control: 'none',
+    },
+  }
+
+  const { popupIsOpen, DataPickerOpener, DataPickerComponent, setReferenceElement, openPopup } =
+    useDataPickerButton(paths)
+
+  // early return, no hooks below this point!
```
ruggi, line 199: This does not seem relevant/correct anymore?

ruggi (approved):

editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
balazsbajorics: fixed in ade133944!

PR #5483, opened 2024-05-02
Parse destructuring of assignment statements.

seanparsons:
**Problem:**
We need more detail in the assignment statements so that we can track destructuring.

**Fix:**

**Commit Details:**

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-props-utils.ts
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-props-utils.ts
@@ -3,6 +3,7 @@ import type {
   Param,
   BoundParam,
   JSExpressionMapOrOtherJavascript,
+  JSExpression,
 } from '../../../core/shared/element-template'
 import {
   isRegularParam,
@@ -27,7 +28,7 @@ export function applyPropsParamToPassedProps(
   function getParamValue(
     paramName: string,
     value: unknown,
-    defaultExpression: JSExpressionMapOrOtherJavascript | null,
+    defaultExpression: JSExpression | null,
   ): unknown {
     if (value === undefined && defaultExpression != null) {
       return jsxAttributeToValue(
--- a/editor/src/components/editor/store/store-deep-equality-instances.ts
+++ b/editor/src/components/editor/store/store-deep-equality-instances.ts
@@ -991,7 +991,7 @@ export const RawSourceMapKeepDeepEquality: KeepDeepEqualityCall<RawSourceMap> =
 export function JSAssignmentKeepDeepEqualityCall(): KeepDeepEqualityCall<JSAssignment> {
   return combine2EqualityCalls(
     (assignment) => assignment.leftHandSide,
-    JSIdentifierKeepDeepEquality(),
+    BoundParamKeepDeepEquality(),
     (assignment) => assignment.rightHandSide,
     JSExpressionKeepDeepEqualityCall,
     jsAssignment,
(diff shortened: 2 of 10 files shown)


editor/src/core/data-tracing/data-tracing.ts
```diff
@@ -436,6 +440,87 @@ function walkUpInnerScopesUntilReachingComponent(
...
+    case 'JS_PROPERTY_ACCESS':
+    case 'JS_ELEMENT_ACCESS':
+    case 'JSX_ELEMENT':
+    case 'JSX_MAP_EXPRESSION':
+    case 'ATTRIBUTE_VALUE':
+    case 'ATTRIBUTE_NESTED_ARRAY':
+    case 'ATTRIBUTE_NESTED_OBJECT':
+      return null
+    case 'ATTRIBUTE_FUNCTION_CALL':
+      if (expression.functionName.startsWith('use')) {
```
bkrmendy, line 457: Is it possible to use the hook call regexp here as well?

seanparsons: We could but it would be slower and I don't need the ability to capture which part of the string is the function name because it's just this value.

bkrmendy (approved):
the code looks good, I'd add a test to the data tracing test suite that uses the new functionality (starting from here https://github.com/concrete-utopia/utopia/blob/40c3869bbfe351fa5edc6cb0643c6e40aa54f2ee/editor/src/core/data-tracing/data-tracing.ts#L611) with a test case that the old code couldn't have handled

editor/src/core/data-tracing/data-tracing.ts
```diff
@@ -436,6 +440,87 @@ function walkUpInnerScopesUntilReachingComponent(
   )
 }
 
+const hookCallRegex = /^(use[A-Za-z]+)\(/
```
ruggi, line 443: in theory this could also contain digits and underscores, right?

`use[a-zA-Z0-9_]+`

seanparsons: Yes, in fact I think I can go quite wide on this because it's guaranteed to be an identifier, maybe.

editor/src/core/data-tracing/data-tracing.spec.ts
```diff
@@ -719,6 +755,35 @@ describe('Data Tracing', () => {
...
+
+      const traceResult = traceDataFromProp(
+        EPP.create(EP.fromString('sb/app:my-component:component-root'), PP.create('title')),
+        editor.getEditorState().editor.jsxMetadata,
+        editor.getEditorState().editor.projectContents,
+        [],
+      )
+
+      expect(traceResult).toEqual(
+        dataTracingToLiteralAttribute(EP.fromString('sb/app:my-component'), PP.create('title'), []),
```
balazsbajorics, line 784: nice!

editor/src/core/data-tracing/data-tracing.spec.ts
```diff
@@ -493,6 +493,42 @@ describe('Data Tracing', () => {
...
+        EPP.create(EP.fromString('sb/app:my-component:component-root'), PP.create('title')),
+        editor.getEditorState().editor.jsxMetadata,
+        editor.getEditorState().editor.projectContents,
+        [],
+      )
+
+      expect(traceResult).toEqual(
+        dataTracingToAHookCall(EP.fromString('sb/app:my-component:component-root'), 'useArray', [
+          '2',
+          'rest',
```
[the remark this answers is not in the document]
seanparsons: Yeah, I'm of two minds about it. The path makes sense for what we've parsed out but looks weird against the data.

editor/src/core/data-tracing/data-tracing.ts
seanparsons: Improved in 5c8e5d0db.

balazsbajorics (approved):

Issue #5485, opened 2024-05-03
Deleting TwoFeatureCallout empties the entire canvas


maltenuhn:
(not visible in video, but initiated via a verified- _single_ press of the delete button

https://screenshot.click/03-37-udju8-4qdqw.gif

PR #5515, opened 2024-05-03
Correctly check unused imports after element deletion

bkrmendy:
## Problem
When removing unused imports after an element is deleted, we didn't check whether the removed element was used in any render props. As a result, sometimes we removed the import even if the element was used inside a render prop

## Fix
Check render props too

## Manual Tests
I hereby swear that:

https://github.com/concrete-utopia/utopia/issues/5485


balazsbajorics (approved):
well that's a nice quick fix for a disastrous bug!

gbalint (approved):

PR #5526, opened 2024-05-03
Selection on click navigator render prop items

### LINKED ISSUE #5489: Clicking on a slot label should select its contents
Today's behaviour: total selection loss (unexpected)
Future desired behaviour: slots and slot labels are independently selectable / focusable, with their own behaviour
Temporary bugfix: select its contents instead (there's a related to ticket to already _highlight_ the content)

image

ruggi:
**Problem:**

Clicking on a render prop navigator item clears the selection and does not select the immediate child.

**Fix:**

1. Add a `childPath` prop to the `RenderPropNavigatorEntry` containing the element path of its immediate child
2. When the navigator item is clicked, select that path

This PR uses the extra field to take into account synthetic entries which don't have their path in the jsx metadata, without requiring to do complex/tedious/weird lookups based on whether their parent have children in the metadata or not.

https://github.com/concrete-utopia/utopia/assets/1081051/87e10eb3-8a04-4d76-96e5-7fef902617f7

**Manual Tests:**
I hereby swear that:

Fixes #5489


bkrmendy (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -924,6 +931,7 @@ export const NavigatorItem: React.FunctionComponent<
         onMouseMove={highlight}
         onDoubleClick={focusComponent}
       >
+        {/* {EP.toString(navigatorEntry.elementPath)} */}
```
Rheeseyb, line 934:
```suggestion
```

Rheeseyb (approved):

PR #5539, opened 2024-05-03
support replacing with maps

bkrmendy:
## Problem
https://github.com/concrete-utopia/utopia/issues/5523 Lists are not available as an element to replace with

## Fix
Make them available and improve `insertComponentPickerItem` so that it can actually replace a target with a list.

### Scope
- Replacing an element with a list works if the element is in the `children` array of another element\
- Replacing an element with a list works if the element is in a conditional branch
- Replacing an element with a list does not work if the element is in a render prop, because for now we only support JSXElements in render props
- Replacing an element with a list does not work if the element is the returned by the map function in a list, because for now we only support JSXElements there (through `ElementsWithin`)

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/navigator/navigator-item/component-picker-context-menu.tsx
+++ b/editor/src/components/navigator/navigator-item/component-picker-context-menu.tsx
@@ -449,12 +449,12 @@ function insertComponentPickerItem(
     if (isReplaceTarget(insertionTarget)) {
       if (
         MetadataUtils.isJSXMapExpression(EP.parentPath(target), metadata) &&
-        elementWithoutUID.type === 'JSX_MAP_EXPRESSION'
+        elementWithoutUID.type !== 'JSX_ELEMENT'
       ) {
         return [
           showToast(
             notice(
-              'Nested Lists are not supported yet',
+              'Only JSX elements are supported inside Lists',
               'INFO',
               false,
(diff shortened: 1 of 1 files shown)


editor/src/components/navigator/navigator-item/component-picker-context-menu.tsx
```diff
@@ -441,6 +446,34 @@ function insertComponentPickerItem(
       ]
     }
 
+    if (isReplaceTarget(insertionTarget)) {
+      if (
+        MetadataUtils.isJSXMapExpression(EP.parentPath(target), metadata) &&
+        elementWithoutUID.type === 'JSX_MAP_EXPRESSION'
+      ) {
+        return [
```
balazsbajorics, line 454: I think we should show a toast for `elementWithoutUID.type !== 'JSX_ELEMENT'` because now Fragment and Conditional fail silently, and only List shows a message.

balazsbajorics (approved):

balazsbajorics:
heads up: this PR only fixes Replace Everything. for Replace With to work, we need Wrap In List support which is missing AFAIK. so once this is merged, the ticket should be rewritten to Replace With doesn't support Lists and be put back to the Malte Backlog

ruggi (approved):

Issue #5477, opened 2024-05-02
Setting minHeight prop on Section selects the next section


maltenuhn:
STR in the video https://screenshot.click/02-35-wl9np-qgyee.gif

Issue is that after ticking the minHeight box, we have two `<Section padded minHeight />` in the project - same hash. We then lose selection, and select the **second** one.

Rheeseyb:
This one will not be solvable in a 2 hour timebox, so do we increase the timebox or de-prioritise it @maltenuhn ?

PR #5547, opened 2024-05-04
Use bounds of element in hash used for UID

Rheeseyb:
**Problem:**
When there are identical elements in a file, updating the first will result in selection jumping to the next, as the UIDs will have been shifted.

**Fix:**
I've included the bounds of each element in the hash used to generate the UID, so that the position (in code) of the element affects the UID. The actual change itself is in `editor/src/core/workers/parser-printer/parser-printer-parsing.ts`, with the rest of the changes being test updates (because of the new UIDs)

**Manual Tests:**
I hereby swear that:

Fixes: #5477


balazsbajorics (approved):
so many snapshots!

seanparsons (approved):

Issue #5535, opened 2024-05-03
Bait and Switch! Beware! Changing a borderRadius on Column (via canvas controls) actually converts the grandparent to absolute


maltenuhn:
The humanity!

![Image](https://screenshot.click/03-34-a5dwi-bv2xc.gif)

![Image](https://github.com/concrete-utopia/utopia/assets/2945037/eb4334b1-7500-4271-a7ef-b5b269f70ff2)

PR #5548, opened 2024-05-04
Harden the ancestor metastrategy

bkrmendy:
## Problem
https://github.com/concrete-utopia/utopia/issues/5535

## Fix
The cause of the problem was that even though the border radius control was actiev, the ancestor metastrategy took precedence. This PR tweaks the ancestor metastrategy so that ti's only fit if the BOUNDING_AREA control is active

### Manual Tests
I hereby swear that:


balazsbajorics (approved):
wow TIL onlyFitWhenDraggingThisControl

ruggi (approved):

ruggi (approved):

PR #5558, opened 2024-05-04
Navigator emphasized color

### LINKED ISSUE #5482: Improve navigator icons for elements and annotated components
**Problem**
Our navigator shows a lot of things with the Frame icon, when they don't really behave like frames. There are three reasons:
- component annotations only have a few icons to choose from
- we don't assign good default icons to built-in elements
- we don't have a good enough set of icons to highlight and distinguish between common things

**Where this is problematic**
- "Spacer" here feels like a frame you can select / draw into - you cannot!
- our various Title components look like frames - not titles
- Suspense and Await look like frames
- Many components don't have _anything_ for users to configure - they are components, sure, but not in the "components are things with props and possibly children you can configure and use to display things" sense
- conversely, various`Button`s  are components, but what really matter is that they're a button - like the function of `Row` and `Column` being to lay things out.

Image

**Approach**
1. Add to our library of icons, and make them about the _function_ of things in the navigator first, and less about what they are
2. Map these (and existing) icons to all major built-in HTML elements
3. Add to the sample project component annotations

**Assets Needed**
NB these should be called after their _function_ rather than elements. We can keep existing ones that are called 'div' or 'zero-sized div' etc, but going forward we should think about 'what they do'. Sometimes that's 1:1 (headlines / headings / titles), sometimes it's really not (placeholder components)

- [x] Frame (filled)
- [x] Frame (dashed) 
- [x] Placeholder (using the dashed frame icon as a base, but subtly different)
- [x] Input 
- [x] Form
- [x] Page
- [x] Image
- [x] Video
- [x] Folder
- [x] Sheet
- [x] Paragraph
- [x] Text
- [x] Headline
- [x] Section
- [x] Row
- [x] Column
- [x] IrregularLayout (not the layout one, still too regular - think 3 boxes randomly placed
- [x] SFX
- [x] Link (same as Remix Link)

**Map these to elements**
- [x] h1...h6: Headline
- [x] p: Paragraph
- [x] all types of `input`: Input
- [x] `form`: Form
- [x] `head`: Folder
- [x] `a`: Link

**Extend the Component Descriptor API to be able to reference all of these icons**
How? Shrug!

**Use these with components from the sample project**
- `Spacer`: Frame (dashed)
- PageTitle and all other titles: Title
- WomanSeeking: Component (emphasized)

ruggi:
Fixes #5482 

Relates to https://github.com/concrete-utopia/hydrogen-november/pull/10

This PR adds handling of the `"emphasized"` `Emphasis` value for components in the navigator so it shows a different color like we do for the `"subdued"` value.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-item.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-item.tsx
@@ -1214,6 +1214,8 @@ export const NavigatorRowLabel = React.memo((props: NavigatorRowLabelProps) => {
               ? 'component'
               : props.emphasis === 'subdued'
               ? 'subdued'
+              : props.emphasis === 'emphasized'
+              ? 'primary'
               : props.iconColor
           }
           elementWarnings={props.elementWarnings}
--- a/editor/src/components/navigator/navigator-item/layout-icon.tsx
+++ b/editor/src/components/navigator/navigator-item/layout-icon.tsx
@@ -123,6 +123,7 @@ export const LayoutIcon: React.FunctionComponent<React.PropsWithChildren<LayoutI
             width={12}
             height={12}
             testId={iconTestId}
+            color={props.color}
           />
         )
       }
@@ -226,6 +227,7 @@ export const LayoutIcon: React.FunctionComponent<React.PropsWithChildren<LayoutI
       iconTestId,
       addAbsoluteMarkerToIcon,
       props.override,
+      props.color,
     ])
 
     const marker = React.useMemo(() => {


bkrmendy (approved):

editor/src/components/navigator/navigator-item/navigator-item.tsx
```diff
@@ -1214,6 +1214,8 @@ export const NavigatorRowLabel = React.memo((props: NavigatorRowLabelProps) => {
               ? 'component'
               : props.emphasis === 'subdued'
               ? 'subdued'
+              : props.emphasis === 'emphasized'
+              ? 'primary'
```
balazsbajorics, line 1218: isn't primary a blue? I think that's a little too strong for emphasized

ruggi: yes it's blue, but that's also shared with other elements in the sample components so I wanted to keep it consistent 🤔

balazsbajorics: ok let's merge and we can tweak it!

balazsbajorics (approved):

PR #5564, opened 2024-05-04
Give remix scenes a scene-id for dom invalidation

### LINKED ISSUE #5554: Canvas highlight for the sample store is incorrect
When CMD+hovering over the title bar, it is unselectable, but instead the Layout component (which is not where the mouse is) becomes highlighted on the canvas. 

However, once the Layout component becomes focused, the bounding box is correct.

(I suspect this is because of non-standard behaviour of the Layout component, but it still feels strange and suspicious)

Image

Rheeseyb:
**Problem:**
Remix scenes aren't being correctly invalidated by the resize and mutation observers, meaning metadata (and therefore the frames used for highlighting on the canvas) is stale.

**Fix:**
The problem here is that we invalidate paths for the dom walker using the scene ID (so that every descendant is updated), but weren't putting the scene ID on Remix scenes.

**Manual Tests:**
I hereby swear that:

Fixes #5554

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
@@ -789,7 +789,7 @@ function renderJSXElement(
     : elementProps
 
   const elementPropsWithSceneID =
-    elementIsScene && elementPath != null
+    (elementIsScene || elementIsRemixScene) && elementPath != null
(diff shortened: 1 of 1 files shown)


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-element-renderer-utils.tsx
```diff
@@ -789,7 +789,7 @@ function renderJSXElement(
     : elementProps
 
   const elementPropsWithSceneID =
-    elementIsScene && elementPath != null
```
balazsbajorics, line 792: oh wow

balazsbajorics (approved):

bkrmendy (approved):

Issue #5561, opened 2024-05-04
Changing page shows previous canvas controls


maltenuhn:
NB: this is probably connected to a number of other bugs where canvas controls are not being drawn in the correct locations (#5554)

**Problem**
When changing a page, we preserve the previous canvas controls - looks confusing (and I have no idea what happens if I were to interact with them, I'd rather not find out)

![Image](https://screenshot.click/04-10-b8h58-yhaup.gif)

PR #5566, opened 2024-05-05
clear selection on Remix route change

bkrmendy:
## Problem
https://github.com/concrete-utopia/utopia/issues/5561

## Fix
This PR adds an editor-wide effect that clears the selection whenever a remix route changes.

### Caveats
- The selection is cleared for all remix scenes even if there are multiple remix scenes and the route is only changed in one of them. We could only keep selected paths that aren't descended from the remix container where the navigation happened, but this would need some more involved logic and I'd like to keep this PR simple. Also this problem is only relevant when there's a multiselection that spans Remix scenes, which is not that common
- `pathname` is used to check whether a navigation has happened, because a change in query params doesn't mean a full navigation, so we should keep the selection in that case

### Manual Tests
I hereby swear that:

### DIFF
--- a/editor/src/components/editor/editor-component.tsx
+++ b/editor/src/components/editor/editor-component.tsx
@@ -454,6 +454,8 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
 
   useGithubPolling()
 
+  useClearSelectionOnNavigation()
+
   const portalTarget = document.getElementById(CanvasContextMenuPortalTargetID)
 
   return (
@@ -562,7 +564,6 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
         keyUp={onWindowKeyUp}
       />
       <CommentMaintainer />
-      <ClearSelectionOnNavigation />
     </>
   )
 })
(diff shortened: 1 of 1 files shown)


editor/src/components/editor/editor-component.tsx
```diff
@@ -558,6 +562,7 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
         keyUp={onWindowKeyUp}
       />
       <CommentMaintainer />
+      <ClearSelectionOnNavigation />
```
liady, line 565: Can this be a hook instead of a component?
```tsx
function useClearSelectionOnNavigation() {
  const dispatch = useDispatch()
  const [remixNavigation] = useAtom(RemixNavigationAtom)
  const paths = Object.values(remixNavigation)
    .map((n) => n?.location.pathname ?? '')
    .join('-')

  React.useEffect(() => {
    dispatch([EditorActions.clearSelection()])
  }, [dispatch, paths])
}
```

liady (approved):

editor/src/components/editor/editor-component.tsx
bkrmendy: good idea, pushed it https://github.com/concrete-utopia/utopia/pull/5566/commits/f7497fb27601f2296f270bb8f94a701bd13d517f

gbalint (approved):
The downside I see is that we clear the selection even if it is not in the Remix scene of the updated route.

PR #5573, opened 2024-05-05
Chore: more api icons

balazsbajorics:
**Problem:**
More icons that can be used as the `ComponentToRegister.icon`

**Commit Details:**

- Malte added more options
- now **every** icon option corresponds to an icon file, so there's no need for indirection in `iconPropsForIcon`
- there were two `iconPropsForIcon`s with a FIXME that one was the clone of the other, so I just deleted one


maltenuhn (approved):

Rheeseyb (approved):
So many icons!

Issue #5577, opened 2024-05-06
the List source data picker's Preferred mode should only show arrays as options


balazsbajorics:
here on the screenshot, only reviews should be visible.

![Image](https://github.com/concrete-utopia/utopia/assets/2226774/d3e95fb5-b0f1-4b8a-8b72-8aa886e9117e)

Details:
the data picker currently receives a null for PropertyPath when it comes to picking data for list sources. Instead it should receive a flag 'list-source' so it can switch modes (or use a different data collection helper).

The list-source mode _ideally_ should be actually retrieve the current value for the list's source array, so it can do the usual shape matching. however that is out of scope for today because it would be a larger change.

What we can do as a quick hack is to only filter the list to arrays (as picking anything other than an array will surely result in a runtime type error)

PR #5596, opened 2024-05-06
data source for maps

bkrmendy:
# [Try it in the sample store](https://utopia.fish/p/15d63825-absorbed-botany/?branch_name=fix-map-picker-only-map-options)

## Problem
https://github.com/concrete-utopia/utopia/issues/5577

## Fix
Show only arrays and object that have arrays somewhere in their props.

Before: 
image

After:
image
(testimonials was the only array in scope at the time of taking the screenshot)

### Other fixes so it doesn't look/feel broken
- The matching logic is updated so that if a map is selected, we show arrays as matching vars
- When picking a data source for maps, objects are shown (if they have an array somewhere inside them), but clicking on them doesn't actually set the data source

### Manual Tests
I hereby swear that:


editor/src/components/inspector/sections/layout-section/list-section.tsx
```diff
@@ -90,6 +91,37 @@ function getMapExpressionMetadata(
...
+function filterVariableOption(option: VariableOption): VariableOption | null {
+  switch (option.type) {
+    case 'array':
+      return {
+        ...option,
+        children: filterKeepArraysOnly(option.children),
+        disabled: false,
+      }
+    case 'object':
+      const chilren = filterKeepArraysOnly(option.children)
```
ruggi, line 103:
```suggestion
      const children = filterKeepArraysOnly(option.children)
```

seanparsons (approved):

balazsbajorics (approved):

editor/src/components/inspector/sections/layout-section/list-section.tsx
bkrmendy: thanks, fixed the typo in https://github.com/concrete-utopia/utopia/pull/5596/commits/e69c97c1cdcb808f8f9ed6ce1b7b01f5ceb2275b

Issue #5556, opened 2024-05-04
Data Cartouche: purple if connected to useLoaderData()


balazsbajorics:
minor prerequisite (do not lose time on this part):

1 Give a new option to `DataCartoucheInner`: contentIsComingFromServer: boolean
2 If contentIsComingFromServer is true, the data picker should have a green color, as seen below.

actual task:
**3 Feed contentIsComingFromServer in all 4 cartouche locations using data tracing**
3A for the non-list case, you can use `traceDataFromProp` and if it returns DataTracingToAHookCall then you know we need green
3B for the list case, consult Balazs

![Image](https://github.com/concrete-utopia/utopia/assets/2226774/a7088f4b-4f78-4ad7-8289-73f4841038c1)

PR #5609, opened 2024-05-06
Data Cartouche: purple if connected to useLoaderData()

bkrmendy:
# [Try it in the sample store](https://utopia.fish/p/b09f9156-electric-stocking/?branch_name=feature-show-cartouche-source)

 **Problem:**
https://github.com/concrete-utopia/utopia/issues/5556

**Fix:**
See https://github.com/concrete-utopia/utopia/issues/5556 with the following caveats:
- the green color wasn't available in the codebase yet, so this PR uses purple
- the data tracing code cannot trace yet from the `children` prop and the variable being mapped in a List (so for example in `reviews.map(...)`, we can't yet trace from `reviews`), so those data references aren't turned purple by this PR

Cartouches that are affected by this PR:
image
image

**Manual Tests:**
I hereby swear that:


editor/src/components/navigator/navigator-utils.ts
```diff
@@ -132,7 +133,12 @@ export function getNavigatorTargets(
       ) {
         if (elementFromProjectContents != null) {
           // add synthetic entry
-          const dataRefEntry = dataReferenceNavigatorEntry(path, elementFromProjectContents)
+          const dataRefEntry = dataReferenceNavigatorEntry(
+            path,
+            EP.parentPath(path),
+            elementFromProjectContents,
+            PP.create('children'),
```
balazsbajorics, line 140: hmmmm I don't like this, 
1. it doesn't currently work
2. I think this will likely never work because the children prop can receive more than a single JSX element as child, see for example:
```
      <Heading>Hello {copy}</Heading>
```

where we already render 'Hello ' and {copy} as two sibling cartouches in the navigator!


Instead of `renderedByElementPath` and `propertyPath`, it'd be neater if `DataReferenceNavigatorEntry` would take a singe, _optional_ ElementPropertyPath:

```typescript
export interface DataReferenceNavigatorEntry {
  type: 'DATA_REFERENCE'
  elementPath: ElementPath
  childOrAttribute: JSXElementChild
  renderedAt: ElementPropertyPath | null
}
```

and here for this case, provide a null for it

bkrmendy: implemented it in https://github.com/concrete-utopia/utopia/pull/5609/commits/097458c020c6a602d59417f0db64026d6fe40e42

balazsbajorics (approved):

editor/src/components/inspector/sections/component-section/cartouche-control.tsx
```diff
@@ -28,6 +46,7 @@ export const IdentifierExpressionCartoucheControl = React.memo(
         onDelete={onDeleteCartouche}
         testId={testId}
         inverted={false}
+        contentIsComingFromServer={isDataComingFromHookResult}
```
Rheeseyb, line 49: Surely this should be `contentIsComingFromAHook`?

editor/src/components/inspector/sections/component-section/data-reference-cartouche.tsx
```diff
@@ -42,13 +44,31 @@ export const DataReferenceCartoucheControl = React.memo(
 
     const dataPickerButtonData = useDataPickerButton(
       [EP.parentPath(elementPath)],
-      PP.fromString('children'), // TODO
+      props.renderedAt == null ? PP.create('children') : props.renderedAt.propertyPath, // TODO
```
Rheeseyb, line 47: Is this one still `TODO`? What do the two `TODO`s even mean?

Rheeseyb (approved):
Looks good, but I dislike the assertion that a hook means the data is coming from the server. If we want to assert that we should be checking for `useLoaderData`

editor/src/components/inspector/sections/component-section/cartouche-control.tsx
bkrmendy: I added it based on the instructions in the original issue https://github.com/concrete-utopia/utopia/issues/5556

Issue #5584, opened 2024-05-06
Can't remove prop using property controls


maltenuhn:
We should have a context menu with Unset, see here. Figure out why that's gone. Please timebox to 30 minutes, if this is a new feature we do NOT need this for today

![Image](https://github.com/concrete-utopia/utopia/assets/2945037/8e9d6a26-c153-4004-b6f0-a716db352c2a)

![Image](https://github.com/concrete-utopia/utopia/assets/2945037/ac096105-aad5-4898-afc5-58d9256e4b5d)

PR #5613, opened 2024-05-06
Ugly hack to fix inspector context menu

gbalint:
**Problem:**
react-contexify takes the mouse event and positions the context menu to the global position found in the event.
However, we put the context menu inside the local element where we need it, so the real position will be offset by the global position of that element. I think this is absolutely incompatible with how react-contexify works.

Why does it work mostly still? Because if we launch it from a div which is close to top/left of the screen, then this difference doesn't matter (if we move the inspector there, the context menu works perfectly too, but normally it is on the right, so the menu is somewhere far off-screen.).

**Fix:**
First I thought just use offsetX and offsetY from the mouse event to position the menu, but unfortunately that only works for the x coordinate, because the MenuProvider is outside of the UIGridRow of the item, so it is positioned from the top of the inspector, but the target of the mouse event is the row item itself, so its y position is different.

Because a real solution would be a rather involved refactor (I guess using Portals), I just created a quick hack to make it visible in the inspector. It should not affect other context menus.

**Manual Tests:**
I hereby swear that:

Fixes https://github.com/concrete-utopia/utopia/issues/5584

### DIFF
--- a/editor/src/components/context-menu-wrapper.tsx
+++ b/editor/src/components/context-menu-wrapper.tsx
@@ -279,7 +279,7 @@ export const MenuProvider: React.FunctionComponent<React.PropsWithChildren<MenuP
   const onContextMenu = React.useCallback(
     (event: React.MouseEvent<HTMLDivElement>) => {
       if (props.itemsLength > 0) {
-        show(event)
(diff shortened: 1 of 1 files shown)


ruggi (approved):
:D

Issue #5428, opened 2024-04-29
Add tabs to component picker


Rheeseyb:

maltenuhn:
@lankaukk could you take a look at this, and see if there's a version with (some) icons across the top, for _common_ categories?

benwolfram:
Design feedback:

• add -> before selected list item
```
<Icn
          category='semantic'
          type='classicarrow-right'
          width={16}
          height={16}
          style={{ opacity: isFocused ? 1 : 0 }}
          color='dynamic'
        />
```
• selected list item gets color: colorTheme.dynamic.value and no background
• smaller gap between arrow, cube icon and label (try ~4px)
• selected tab gets background: colorTheme.primary.value and color: colorTheme.white.value

make popup theme-aware - entire popup gets:
• backgroundColor: colorTheme.bg0.value,
• color: colorTheme.fg1.value,
• padding: 4,
• boxShadow: UtopiaStyles.shadowStyles.low.boxShadow,
• border: .2px solid ${colorTheme.popupBorder.value},
• borderRadius: 10,

PR #5629, opened 2024-05-07
Add Tabs to Component Picker

benwolfram:
**Problem:**
We want to add tabs to quickly jump to categories within the component picker.

**Fix:**
Focus should to be managed by a single element, and then arrow key navigation within as there may be a large list of categories and we don't want to force users to cycle through them all.

Within the focused element, we highlight an active selection and upon Enter keypress scroll this option in the list into view.

Also, applying design changes as requested to the list items.

**Manual Tests:**
I hereby swear that:

^ didn't do these and would love context on what needs to be done if it's relevant to my PR

Relates to https://github.com/concrete-utopia/utopia/issues/5428

### DIFF
--- a/editor/src/components/navigator/navigator-item/component-picker.tsx
+++ b/editor/src/components/navigator/navigator-item/component-picker.tsx
@@ -3,7 +3,7 @@
 import { jsx } from '@emotion/react'
 import React, { useCallback, useMemo } from 'react'
 import debounce from 'lodash.debounce'
-import { Icn, type IcnProps } from '../../../uuiui'
+import { colorTheme, Icn, type IcnProps } from '../../../uuiui'
 import { dark } from '../../../uuiui/styles/theme/dark'
 import type { JSXElementChild } from '../../../core/shared/element-template'
 import type { ElementPath, Imports } from '../../../core/shared/project-file-types'
@@ -15,13 +15,22 @@ import type {
 } from '../../canvas/ui/floating-insert-menu'
 import { UIGridRow } from '../../../components/inspector/widgets/ui-grid-row'
 import { FlexRow, type Icon } from 'utopia-api'
-import { assertNever } from '../../../core/shared/utils'
 import { insertableComponent } from '../../shared/project-components'
 import type { StylePropOption, InsertableComponent } from '../../shared/project-components'
 import type { Size } from '../../../core/shared/math-utils'
 import { dataPasteHandler } from '../../../utils/paste-handler'
 import { sortBy } from '../../../core/shared/array-utils'
 import { iconPropsForIcon } from './component-picker-context-menu'
+
+const FILTER_CATEGORIES = [
+  'Layout',
+  'Forms',
+  'HTML',
+  'Shopify',
+  'Advanced',
+  'Fragment',
+  'Miscellaneous',
+]
 
 export interface ComponentPickerProps {
   allComponents: Array<InsertMenuItemGroup>
@@ -152,6 +161,37 @@ export const ComponentPicker = React.memo((props: ComponentPickerProps) => {
     [flatComponentsToShow, highlightedComponentKey, onItemClick, selectIndex],
   )
 
+  const fakeCategorizedComponents = [
+    {
+      label: FILTER_CATEGORIES[0],
+      items: flatComponentsToShow.slice(0, 1),
+    },
+    {
+      label: FILTER_CATEGORIES[1],
+      items: flatComponentsToShow.slice(1, 5),
+    },
+    {
+      label: FILTER_CATEGORIES[2],
+      items: flatComponentsToShow.slice(5, 7),
+    },
+    {
+      label: FILTER_CATEGORIES[3],
+      items: flatComponentsToShow.slice(7, 8),
+    },
+    {
+      label: FILTER_CATEGORIES[4],
+      items: flatComponentsToShow.slice(8, 9),
+    },
+    {
+      label: FILTER_CATEGORIES[5],
+      items: flatComponentsToShow.slice(9, 10),
+    },
+    {
+      label: FILTER_CATEGORIES[6],
+      items: flatComponentsToShow.slice(10),
+    },
+  ]
+
   return (
     <div
       style={{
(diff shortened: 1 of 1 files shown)


editor/src/components/navigator/navigator-item/component-picker.tsx
```diff
@@ -152,6 +161,37 @@ export const ComponentPicker = React.memo((props: ComponentPickerProps) => {
...
+    },
+    {
+      label: FILTER_CATEGORIES[5],
+      items: flatComponentsToShow.slice(9, 10),
+    },
+    {
+      label: FILTER_CATEGORIES[6],
+      items: flatComponentsToShow.slice(10),
+    },
+  ]
```
benwolfram, line 193: Will come from API in the future, all fake.

editor/src/components/navigator/navigator-item/component-picker.tsx
```diff
@@ -266,8 +316,145 @@ const FilterBar = React.memo((props: FilterBarProps) => {
...
+          setFocusedIndex((prev) => Math.max(prev - 1, 0))
+        } else if (event.key === 'Enter') {
+          document
+            .getElementById(FILTER_CATEGORIES[Math.max(focusedIndex - 1, 0)])
+            ?.scrollIntoView({ block: 'start', behavior: 'smooth' })
+        } else {
+          return
+        }
+        event.stopPropagation()
+        event.preventDefault()
```
benwolfram, line 357: So keyboard navigation is a bit funky in this component and I am not sure why and what is going on behind the scenes with key events.

Basically, I think this popover container should have focus trapping and it doesn't. When I press tab to move focus, there are things in the background happening unrelated to the popover. Only things inside the popover should be affected until closed or Esc.

As a result, I have been focusing the Filter... text input and then pressing shift+tab to reverse focus into the category buttons.

maltenuhn: so IIRC it's possible that we handle some key events at editor level, even before they get to the popover container:  React Contexify I think should handle keyboard navigation out of the box if you use their built-in components - and it does not where we use it. If that's the case, it's obvs an uphill battle - and we probably should exempt context menus from this. Tomorrow, ping @ dev and someone can walk you through our global event handlers. (disclaimer: might be a totally different issue!)

benwolfram: It does seem that way, but I assume tab should be exempt from any global key handlers, as it should fallback to the browser's handling for proper accessible navigation. Good convo to have tomorrow.

balazsbajorics:
Hey Ben! Since you worked on this we made some changes to the actual menu design, I think these got reverted on your PR because of that botched revert I did. (sorry for that again!)

Could you please re-apply the changes from current master, so the menu items look like this?

image

Since we still don't have the API changes to provide you with the categories, I propose these two changes so we can merge this pull request:

1. if there are no categories, the top bar should be hidden, or just have a single "Components" entry
2. delete the fake categories, so we can merge it and use it with the "empty state" until someone implements the category API and wires it in.

After these changes, I think the PR will be good to go!

editor/src/components/navigator/navigator-item/component-picker.tsx
```diff
@@ -152,6 +158,13 @@ export const ComponentPicker = React.memo((props: ComponentPickerProps) => {
     [flatComponentsToShow, highlightedComponentKey, onItemClick, selectIndex],
   )
 
+  const categorizedComponents = [
+    {
+      label: 'Everything',
+      items: flatComponentsToShow,
+    },
+  ]
```
benwolfram, line 166: To address @balazsbajorics feedback, it's actually not super simple to just nuke all categories because the data needs to be formatted in categories still for it to show up based on how it's rendered now.

Here, I am just using a single category now, and only showing the tabs when there are more than 1 category, so the component is now hidden.

benwolfram:
@balazsbajorics per discussion in #design channel, I've removed my style changes.

image

editor/src/components/navigator/navigator-item/component-picker.tsx
balazsbajorics: sounds like a good solution to me!

balazsbajorics (approved):

Rheeseyb:
> ^ didn't do these and would love context on what needs to be done if it's relevant to my PR

Basically you need to clone the main branch of https://github.com/concrete-utopia/hydrogen-november, switch to live mode, and click on one of the links in that project, so that you can ensure the project loads and can be navigated. This PR would have no impact at all on that, but we keep these manual tests as a blanket protection

editor/src/components/navigator/navigator-item/component-picker.tsx
Rheeseyb: Was this discussed? I'm thinking we should merge this PR and address the tabbing as a separate PR (since it sounds like that would be an editor-wide change anyway)

editor/src/components/navigator/navigator-item/component-picker.tsx
```diff
@@ -293,62 +442,45 @@ const ComponentPickerComponentSection = React.memo(
...
-            >
-              <FlexRow css={{ gap: 10, height: 28, alignItems: 'center' }}>
-                <Icn
-                  {...iconPropsForIcon(component.value.icon ?? 'component')}
-                  width={12}
-                  height={12}
-                />
-                <label>{component.label}</label>
+        {components.reduce((acc, category) => {
+          const categoryMarkdown = category.items.map((component, index) => {
```
Rheeseyb, line 446: Looks like a `flatMap` would simplify it a bit here

benwolfram: Never used it before! Thanks for the suggestion.

benwolfram:
> Basically you need to clone the main branch of https://github.com/concrete-utopia/hydrogen-november, the easiest is to use this query param: http://localhost:8000/p/?clone=concrete-utopia/hydrogen-november , then switch to live mode, and click on one of the links in that project, so that you can ensure the project loads and can be navigated. This PR would have no impact at all on that, but we keep these manual tests as a blanket protection

Ah thanks Rheese, yeah I assume no impact with this PR but was just curious to learn so I can check the boxes. Didn't know about live mode, cool stuff!

Rheeseyb (approved):

Issue #5634, opened 2024-05-08
Restore best match functionality for data picker.


seanparsons:
The previous best match functionality wouldn't really mark a best match, but often just shows a lot of fields.

Balazs' suggestion:
```
the matches: boolean should be replaced with a
matches: 'no' | 'yes' | 'best-match'
and in case there is no best match, highlight all matches: yes-es
in case there is a best match, only highlight the best match
```

PR #5638, opened 2024-05-08
Better Matching By Properties In Data Picker

### LINKED ISSUE #5603: Data picker: certain values are a good match (eg strings) but are not indicated as insertable
eg the `value` field here contains a string, and should be highlighted as insertable

Image

seanparsons:
**Problem:**
Certain values are a good match (eg strings) but are not indicated as insertable.

**Fix:**
When a specific property is being targeted, the logic in `orderVariablesForRelevance` was not drilling down into that property. Now the control description used comes from the property specified if there is one.

**Commit Details:**
- Removed `keepOnlyFirstMatch` as it conflicts with the needs we have right now.
- Added `getTargetPropertyFromControlDescription`.
- `orderVariablesForRelevance` uses `getTargetPropertyFromControlDescription` to drill down into the relevant part of the variable.

**Manual Tests:**
I hereby swear that:

Fixes #5603

### DIFF
--- a/editor/src/components/inspector/sections/component-section/variables-in-scope-utils.spec.ts
+++ b/editor/src/components/inspector/sections/component-section/variables-in-scope-utils.spec.ts
@@ -1,0 +1,121 @@
+import { maybeToArray } from '../../../../core/shared/optional-utils'
+import type { ControlDescription } from '../../../custom-code/internal-property-controls'
+import type { PropertyValue, VariableInfo } from './variables-in-scope-utils'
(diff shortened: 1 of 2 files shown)


balazsbajorics:
there's a follow up ticket [#5634](https://github.com/concrete-utopia/utopia/issues/5634) about restoring the broken and now-removed Best Match functionality

balazsbajorics (approved):

bkrmendy (approved):

Issue #5649, opened 2024-05-09
Component API should support string field for `folder`


maltenuhn:
- [x] Instead of folder control, each prop should support an (optional) `folder` field (string)
- [x] the folder control need no longer be supported by the API at all
- [x] No parsing of the label to create subfolders; hence, no sub-folder support
- [ ] the inspector is populated _first_ by folder-less components, then by foldered ones, _in order of appearance_  ('first time the particular folder label shows up when iterating over the props').
- [ ] folders are open by default, except if label matches exactly 'Advanced' (that is closed by default)

bkrmendy:
This PR adds the folder labels/removes the folder control https://github.com/concrete-utopia/utopia/pull/5651

PR #5651, opened 2024-05-09
Support folder labels in the component API

bkrmendy:
https://github.com/concrete-utopia/utopia/issues/5649

## Problem
Organizing properties into folders via the folder control is not ergonomic in practice

## Fix
Use a folder label (in the form of a `folder?: string` prop) instead to organise props into folders

### Scope
To keep the PR scoped, this PR only adds the new prop to the property controls and removes the folder control. Automatically assembling the folders on the UI will come on a subsequent PR

### Commit Details
- the folder property control is removed
- `GenericControlProps` is added to group the props shared by all property controls (`folder` is added on here)

### Manual Tests
I hereby swear that:

### DIFF
--- a/editor/src/core/property-controls/property-controls-local.spec.tsx
+++ b/editor/src/core/property-controls/property-controls-local.spec.tsx
@@ -1782,7 +1782,7 @@ describe('registered property controls', () => {
     })
   })
 
-  describe('folderes', () => {
+  describe('folders', () => {
     it('can specify a folder prop', async () => {
       const renderResult = await renderTestEditorWithModel(
         project({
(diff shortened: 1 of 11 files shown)


editor/src/core/property-controls/property-controls-local.spec.tsx
```diff
@@ -1781,6 +1781,68 @@ describe('registered property controls', () => {
       `)
     })
   })
+
+  describe('folderes', () => {
```
ruggi, line 1785:
```suggestion
  describe('folders', () => {
```

ruggi (approved):

editor/src/core/property-controls/property-controls-local.spec.tsx
bkrmendy: thanks! committed it

balazsbajorics (approved):

PR #5660, opened 2024-05-13
Prep PR - move some types to `utopia-shared`

liady:
This PR moves some types to `utopia-shared`, and uses `ImageFile`, `AssetFile`, `RevisionState` in the editor.

NOTE:
1. This does not add a build step to build `utopia-shared` - it assumes that its code is being used as-is (direct imports)
2. This PR is not complete - we still need to figure out a way to extract `TextFile` and `ParseSuccess` without importing all of their type dependencies
3. Due to # 2 - This PR *does not* replace all of the usages of extracted types in the `editor` module


editor/src/core/shared/project-file-types.ts
```diff
@@ -7,6 +7,11 @@ import type {
 } from './element-template'
 import type { ErrorMessage } from './error-messages'
 import { arrayEqualsByValue, objectEquals } from './utils'
+import type { ImageFile } from 'utopia-shared/src/types-project-contents'
```
liady, line 10: this is to check that imports work properly

utopia-shared/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "utopia-shared",
+  "version": "1.0.0",
+  "description": "A shared module between the editor and Remix",
+  "scripts": {
+    "build": "tsc"
```
liady, line 6: This is not currently used since the code is being imported as-is (no need to build it and use it as a package)

balazsbajorics (approved):

editor/src/core/shared/project-file-types.ts
```diff
@@ -7,6 +7,15 @@ import type {
 } from './element-template'
 import type { ErrorMessage } from './error-messages'
 import { arrayEqualsByValue, objectEquals } from './utils'
+import type {
```
liady, line 10: this is to check that imports work properly in this PR

ruggi (approved):

PR #5664, opened 2024-05-13
Fixing annotation of Typography.Text

gbalint:
**Problem:**
Components which are not directly imported (e.g. `Tyopgraphy.Text` in `antd`) can not be annotated in Utopia.

**Fix:**
We need to make sure names with `.` in them are parsed as JSXElementName, and their baseVariable is imported, but their last part is used as component name.

**Missing:**
Validation does not fully work for these cases:
[ ] For internal components, we only check whether the last part of the name is the component name in ComponentRendererComponent
[ ] For external components the require annotation method doesn't work, because we don't require directly the component itself.
Fortunately, validation is just a best effort, worst case is that we will not find some of the component annotation errors, but valid component annotation files will work.

**Manual Tests:**
I hereby swear that:

Fixes https://github.com/concrete-utopia/utopia/issues/5552

### DIFF
--- a/editor/src/core/property-controls/property-controls-local.ts
+++ b/editor/src/core/property-controls/property-controls-local.ts
@@ -94,7 +94,12 @@ import {
 } from '../../components/editor/actions/action-creators'
 import type { ProjectContentTreeRoot } from '../../components/assets'
 import type { JSXElementChildWithoutUID } from '../shared/element-template'
-import { jsxAttributesFromMap, jsxElement, jsxTextBlock } from '../shared/element-template'
+import {
+  jsxAttributesFromMap,
+  jsxElement,
+  jsxElementNameFromString,
+  jsxTextBlock,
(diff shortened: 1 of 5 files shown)


editor/src/core/property-controls/property-controls-local.ts
```diff
@@ -615,9 +638,10 @@ function componentInsertOptionFromExample(
           code: `<${typed.name} />`,
         }
       }
+      const jsxName = jsxElementNameFromString(typed.name)
       return {
         label: typed.name,
-        imports: `import {${typed.name}} from '${moduleName}'`,
+        imports: `import {${jsxName.baseVariable}} from '${moduleName}'`,
```
balazsbajorics, line 644: is this going to work for antd Typography.Text -like stuff? Wouldn't this become `import { Text } from 'antd'` ?

seanparsons (approved):

Issue #5491, opened 2024-05-03
Insertion on canvas and navigator should respect the API "supports children" annotation


maltenuhn:
**Problem**
We had a number of situations where components get inserted and then disappear, and where things get inserted into non-children-supporting elements that then throw a canvas error. See eg https://screenshot.click/03-25-yv7fw-ia3di.gif

**When a component (via the API) or element (via our annotation; eg. `img`) does not support children**
- [x] The "insert" button in the navigator should not show
- [x] Canvas draw-to-insert should not offer it as a valid target
- [x] Canvas reparent should not offer it as a valid target either
- [x] Pasting should paste as sibling (no further checks required, we can assume _existence_ of a child means it's safe for it to have a sibling)
- [x] Any other single-shot insertion eg of data should not work
- [x] Navigator D&D should not 'latch'
- [x] the on-canvas "plus" button should not show, _after_ the changes here: https://github.com/orgs/concrete-utopia/projects/6/views/1?pane=issue&itemId=62009824

_(If there other forms of insertion I have missed, please mention in a channel)_

PR #5671, opened 2024-05-14
targetSupportsChildren is false when component annotation says so

gbalint:
**Problem:**
Insertion on canvas and navigator should respect the API "supports children" annotation.

**Fix:**
We have a few targetSupportsChildren related functions in MetadataUtils. Most of the PR is about passing down propertyControlsInfo to these functions, so they can check the annotation, and return false when `supportsChildren` is false (regardless of any other circumstances).

**Manual Tests:**
I hereby swear that:

Fixes https://github.com/concrete-utopia/utopia/issues/5491

### DIFF
--- a/editor/src/core/model/element-metadata-utils.spec.tsx
+++ b/editor/src/core/model/element-metadata-utils.spec.tsx
@@ -426,6 +426,7 @@ describe('targetElementSupportsChildren', () => {
       path,
       { [EP.toString(path)]: element },
       {},
+      {},
     )
     expect(actualResult).toEqual(true)
   })
@@ -439,6 +440,7 @@ describe('targetElementSupportsChildren', () => {
       {},
       path,
       { [EP.toString(path)]: element },
+      {},
       {},
     )
     expect(actualResult).toEqual(true)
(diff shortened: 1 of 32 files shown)


editor/src/core/model/element-metadata-utils.spec.tsx
```diff
@@ -711,9 +730,70 @@ export const App = (props) => {
       path,
       { [EP.toString(path)]: element },
       {},
+      {},
     )
     expect(actualResult).toEqual(true)
   })
+  it('returns false for a component with property controls info including supportsChildren false', () => {
```
Rheeseyb, line 737: Can you add a test for both the true case and the default (i.e. unspecified) cases please

Rheeseyb (approved):

bkrmendy (approved):

editor/src/core/model/element-metadata-utils.spec.tsx
gbalint: I added a test using the supportsChildren=true annotation. Regarding the default case: all the existing tests in the file already test that case thoroughly.

balazsbajorics (approved):

maltenuhn:
@gbalint because there's a lot of entry points to insertion etc, could you reference the ticket and verify that all of the things in it actually work - ideally in this PR, so if someone looks back and maybe finds something we missed, it's clear what was in scope?

PR #5675, opened 2024-05-14
Feature/navigator child data tracing

balazsbajorics:
Before:
image

After:
image

**Problem:**
The data tracing could only start running from a property at an ElementPropertyPath, but in our representation the jsx children are not the same as passing data into `props.children`, which means the data tracing couldn't trace for data passed in _as_ children to elements.
This limitation meant that #5609 only worked for navigator cartouches that belonged to data passed in as props, not children of elements. You can see in the above before/after screenshots that cartouches representing data coming from useLoaderData are displayed in blue on master.

**Fix**
#5666 Introduced the ability to start the data tracing from an arbitrary element reference and an ElementPath pointing at their containing scope. This PR uses `traceDataFromElement` for the navigator data cartouches.

**Details:**
- The navigator's `DataReferenceCartoucheControl` takes a new prop `surroundingScope` and calls `traceDataFromElement` instead of `traceDataFromProp`
- Added `surroundingScope: ElementPath` to `DataReferenceNavigatorEntry`
- Clarified the code that calls `useDataPickerButton` in `DataReferenceCartoucheControl`


editor/src/components/inspector/sections/component-section/data-reference-cartouche.tsx
```diff
@@ -40,12 +42,18 @@ export const DataReferenceCartoucheControl = React.memo(
       'DataReferenceCartoucheControl contentsToDisplay',
     )
 
-    // TODO get the actual data _value_ to display in the cartouche, and only show the data reference in the tooltip
+    const renderedAt =
+      props.renderedAt ??
+      // if the renderedAt is not provided, we assume that the element is rendered at the parent's children
+      EPP.create(
+        EP.parentPath(elementPath),
```
bkrmendy, line 49: we should turn this assumption into an explicit type, because in `getNavigatorTargets` we have all the information we need, so we can specify the behaviour there instead of guessing here

seanparsons (approved):

editor/src/components/inspector/sections/component-section/data-reference-cartouche.tsx
balazsbajorics: I talked to Berci, we decided to postpone this to a future PR as we actually want to change useDataPickerButton in a way that we won't need this fallback here

PR #5676, opened 2024-05-14
chore(editor) Remove our console proxy and the console tab

### LINKED ISSUE #5631: kill proxy-console and remove the console tab from the vscode pane
this may involve some test snapshot breakage

this will **definitely** improve stack traces of test errors, so it's worth it

Rheeseyb:
**Problem:**
The console pane is useless.

**Fix:**
Remove it.

**Manual Tests:**
I hereby swear that:

Fixes #5631


ruggi (approved):
🧹

liady (approved):

editor/src/core/shared/runtime-report-logs.tsx
```diff
@@ -98,68 +89,6 @@ export function setReactRouterErrorHasBeenLogged(value: boolean): void {
...
-        const firstLine = log.data[0]
-        // ...is a string...
-        if (typeof firstLine === 'string') {
-          // ...and it starts with these prefixes...
-          if (
-            firstLine.startsWith(ReactRouterErrorPrefix) ||
-            firstLine.startsWith(ReactRouterAwaitErrorPrefix)
-          ) {
-            // ...Mark these as having been seen.
-            setReactRouterErrorHasBeenLogged(true)
```
seanparsons, line 145: In PR #5084 this was added so that if a change caused an exception in something like an `<Await>` component, that a further change would result in the error being cleared. Does that now work without this logic?

Issue #5643, opened 2024-05-08
Standardise our context menus


maltenuhn:
**This is partially an investigation task to 'take stock' of what we have, how it's implemented, and how it needs to be refactored.**

**Problem**
Our "context menus" are non-standardised, hard to implement quickly, and even harder to fix because of the 'which one' problem. But they are also on the critical path to expose capabilities quickly, and in a way that is not embarassing.

**What do we want?**

1. An easy, fast, and standardised way to create a new "Popout" menu:
- to populate: either with a "list of things you can click on", or a completely custom UI, 
- and that's correctly positioned vs the element invoking it. (NB sometimes that element is a button - probably SquareButton -, sometimes it's something more custom like an avatar)
- if it's populated from data, that data / labels possibly need to be contextual ('Show Inspector' vs 'Hide Inspector', disabled entries etc
- a way to invoke this that doesn't require adding the new menu to a bunch of places

Some examples of things we've talked about
- Avatar that shows "Show Projects", "Add Github authentication" and "Log out"
- the new Component Picker and Data picker menu, but correctly positioned (!)
- A "Panel" button in the toolbar that shows what panels are shown and hidden
- A "..." SquareButton in the Component part of the inspector, with options "Jump to component definition" and "Jump to source"
- The 'add Page' button correctly tied to the Plus button (btw check out how this was incorrectly implemented using the `item-context-menu` because that was the easiest to find

2. An easier way to specify the missing 'Pages tab' context menus for routes and favorites ('Remove from Favorites', 'Add to Favorites', 'Jump to source', 'Copy URL')

3. Cleaner, more modular, more modern code, and standardised keyboard handling across the editor for these

5. 4. Standard styles for Menu Entries, including the common variants:
(a) just-items (no space for icons), 
(b) some-or-all-elements-have-icons, 
(c) w/ optional space for optional keyboard shortcuts or accessors, and 
(d) with indentation and expansion arrows 
(e) the 'checkbox' version we have in PopupListControl

NB those Menu Entries are also _eerily similar_ to List Items elsewhere! Please propose ways to standardise these more, universally (cf the implementation of `<GridRow  variant='...` that we really like)

This is probably 40% investigation, 15% implementation, and 78% refactoring

**FAQ to get you started**

**How are context menus implemented?**
In a number of ways, mainly by using a legacy React Contexify and then building a wrapper for it, but also in other, more surprising, ways. They've accrued over the years!

**What context menus do we have today?**
[...]

PR #5679, opened 2024-05-14
Refactor ContextMenuWrapper

benwolfram:
**Problem:**
Modernizing this file to show it some love in preparation of work to improve context menus and upgrade react-contexify, ie. keyboard navigation.

**Fix:**
Updating to modern react hooks and functional components.

Only note, is that I chose to remove the memoization (formerly `shouldComponentUpdate`, replace with `memo` in one of my earlier commits). I don't see a performance impact from this and it's better to not pre-emptively memoize. Typescript wasn't able to properly infer the generic types when wrapping with memo (see the commit related to removing memo).

s/o to @Rheeseyb for the help

**Manual Tests:**
I hereby swear that:

Relates to https://github.com/concrete-utopia/utopia/issues/5643


editor/src/components/context-menu-wrapper.tsx
```diff
@@ -18,142 +27,125 @@ import type { WindowPoint } from '../core/shared/math-utils'
...
+}
+
+interface SimpleItem<T> {
+  item: Item<T>
+  type: 'simple'
+}
+
+type MenuItem<T = null> = Submenu<T> | SimpleItem<T> | null
+
+export type Item<T = null> = ContextMenuItem<T> | null
```
benwolfram, line 43: Low key, I don't love how I've defaulted to null here, but it works.

Haven't dealt much with generics, used to preferring to have explicit typing — so open to any suggestions. Particularly I hadn't encountered component's having types to them, ie. `export class MomentumContextMenu<T> extends ReactComponent<ContextMenuProps<T>>` where this first `T` is coming from and how to seamlessly translate to functional component refactor.

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -18,206 +27,195 @@ import type { WindowPoint } from '../core/shared/math-utils'
...
+
+export type ContextMenuData<T> = Item<T>[] | null
+
 export interface ContextMenuWrapperProps<T> {
-  id: string
-  items: Array<ContextMenuItem<T>>
-  dispatch?: EditorDispatch
-  data: T
-  renderTag?: string
+  children?: ReactNode
```
Rheeseyb, line 48: Normally we'd use `React.PropsWithChildren` but this is fine tbh

Rheeseyb (approved):

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -18,206 +27,195 @@ import type { WindowPoint } from '../core/shared/math-utils'
...
+              if (!(triggerEvent instanceof MouseEvent)) return null
+              return windowPoint({ x: triggerEvent.clientX, y: triggerEvent.clientY })
+            })()
+            item?.action(getData(), dispatch, rightClickCoordinate, event)
+            contextMenu.hideAll()
+          }}
+          hidden={isHidden(item)}
+          style={{
+            height: item?.isSeparator ? 9 : 28,
+            display: 'flex',
```
maltenuhn, line 140: Maybe too soon / out of scope for this refactor: but I wonder if we'd be better off with a `grid` here, or even a variable for the list-item layout that we want (just label, label with accesor, icon+ label, icon+label+accessor etc).

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -18,206 +27,195 @@ import type { WindowPoint } from '../core/shared/math-utils'
...
+            alignItems: 'center',
+            borderRadius: 4,
```
maltenuhn, line 142: some of these should come from `UtopiaTheme` (same for height)

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -18,206 +27,195 @@ import type { WindowPoint } from '../core/shared/math-utils'
...
+      style={style}
+      onMouseDown={stopPropagation}
+      onMouseUp={stopPropagation}
+      onClick={stopPropagation}
+      ref={forwardRef}
+    >
+      <MenuProvider id={id} itemsLength={items.length} key={`${id}-provider`} style={providerStyle}>
+        {children}
+      </MenuProvider>
+      <MomentumContextMenu dispatch={dispatch} getData={getData} id={id} items={items} key={id} />
```
maltenuhn, line 216: small thing, but can we rename this to just `ContextMenu` (or `UtopiaContextMenu` as much as that feels redundant)? Momentum was an old product name we used in the mid-1990s

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -18,206 +27,195 @@ import type { WindowPoint } from '../core/shared/math-utils'
...
+                  {item.label}
+                </span>
+              }
+              arrow={<Icons.ExpansionArrowRightWhite style={{ marginLeft: 8 }} />}
+            >
+              {item.items.map(renderItem)}
+            </SubmenuComponent>
+          )
+        } else {
+          if (item === null) return null
```
gbalint, line 175: Minor styling thing, but we avoid using single line if statements

gbalint (approved):

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -2,7 +2,16 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/react'
 import fastDeepEquals from 'fast-deep-equal'
-import React, { Component as ReactComponent } from 'react'
+import React, {
+  type CSSProperties,
```
liady, line 6: minor - we need to decide at some point whether we're going with `React.<Something>` or `import { Something } from 'react'`

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -238,75 +236,79 @@ export const InspectorRowHoverCSS = {
...
+  id,
+  itemsLength,
+  localXHack_KILLME,
+  style,
+}: MenuProviderProps) => {
+  const { show } = useContextMenu({ id })
+
+  const onContextMenu = useCallback(
+    (event: MouseEvent<HTMLDivElement>) => {
+      if (itemsLength <= 0) return
```
liady, line 297: single-line here as well :)

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -238,75 +236,79 @@ export const InspectorRowHoverCSS = {
...
+
+      show(
+        event,
+        localXHack_KILLME === 'local-x-coord-KILLME'
```
liady, line 301: again, minor - but maybe extract it to 
```ts
  const options = localXHack_KILLME === 'local-x-coord-KILLME' 
      ? { position: { x: event.nativeEvent.offsetX, y: event.nativeEvent.pageY } } 
      : undefined

  show(event, options)
```

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -18,206 +27,195 @@ import type { WindowPoint } from '../core/shared/math-utils'
...
 
-  onShown = () => {
-    document.body.classList.add(BodyMenuOpenClass)
-  }
-  onHidden = () => {
-    document.body.classList.remove(BodyMenuOpenClass)
-  }
+  const renderItem = useCallback(
+    (item: Item<T>, index: number) => {
+      return (
```
liady, line 123: how easy/hard will it be to extract it to its own component instead of a callback?

liady (approved):
lgtm, left some code style comments

editor/src/components/context-menu-wrapper.tsx
benwolfram: Sounds good, I was curious what Momentum meant tbh, didn't really understand!

editor/src/components/context-menu-wrapper.tsx
benwolfram: Roger that, feels weird to me, but I gotta learn all the preferences here, so grill me on them all! Thanks.

editor/src/components/context-menu-wrapper.tsx
benwolfram: I probably should do this.. I'll update!

editor/src/components/context-menu-wrapper.tsx
benwolfram: Done in it's own commit

editor/src/components/context-menu-wrapper.tsx
benwolfram: So... this felt easy and like a no brainer for organization. But I feel like there are just hidden traps waiting for me in this component 😂. I spent all afternoon trying to get it and making it a new component broke the 'Paste' functionality and tests. So I've reverted that.

If you think it's important, maybe we can follow up later.

PR #5682, opened 2024-05-15
Restore react-router error handling

Rheeseyb:
**Problem:**
#5676 removed some handling we put in place in #5084 specific for a couple of react-router errors that would cause issues with the canvas.

**Fix:**
Rather than resurrecting the console proxying code for the sake of this, I've narrowed the logic down to only operate on the `console.error` method. We then listen for those specific errors (checking the first arg of the function, which is what we were doing previously).

**Note:**
Unlike before, this will no longer immediately restore the canvas. Some form of action is required to restore the canvas, such as a manual save, or even scrolling the canvas (in the following video I wait a bit before scrolling the canvas to demonstrate this): 
https://github.com/concrete-utopia/utopia/assets/1044774/d92e3eba-4d7f-413a-b484-8c34228d0adf
I believe this is because with the previous console proxying we would have re-rendered the canvas after any log line due to the (horrible) way the proxying was wired in.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/core/shared/runtime-report-logs.tsx
+++ b/editor/src/core/shared/runtime-report-logs.tsx
@@ -155,7 +155,7 @@ export function useErrorOverlayRecords(): ErrorOverlayRecords {
   return { errorRecords, overlayErrors }
 }
 
-export function filterOldPasses(errorMessages: Array<ErrorMessage>): Array<ErrorMessage> {
+function filterOldPasses(errorMessages: Array<ErrorMessage>): Array<ErrorMessage> {
   let passTimes: { [key: string]: number } = {}
   fastForEach(errorMessages, (errorMessage) => {
     if (errorMessage.passTime != null) {
--- a/editor/src/core/shared/runtime-report-logs.tsx
+++ b/editor/src/core/shared/runtime-report-logs.tsx
@@ -154,3 +154,39 @@ export function useErrorOverlayRecords(): ErrorOverlayRecords {
 
   return { errorRecords, overlayErrors }
 }
+
+const ReactRouterErrorPrefix = `React Router caught the following error during render`
+const ReactRouterAwaitErrorPrefix = `<Await> caught the following error during render`
+
+export function listenForReactRouterErrors(targetConsole: Console): void {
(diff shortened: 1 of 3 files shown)


editor/src/core/shared/runtime-report-logs.tsx
```diff
@@ -154,3 +154,39 @@ export function useErrorOverlayRecords(): ErrorOverlayRecords {
...
+
+  // Squirrel away the original method for unpacking later.
+  const originalMethod = targetConsoleAny.error
+  targetConsoleAny.originalErrorMethod = originalMethod
+  targetConsoleAny.error = function (...args: Array<any>) {
+    // Call the original method first.
+    originalMethod(...args)
+
+    // If the first part of the log line is a string
+    const firstLine = args[0]
```
Rheeseyb, line 180: Previously we would use `console-feed` to parse, then encode, then decode (?) the `args` (https://github.com/concrete-utopia/utopia/pull/5676/files#diff-e3967556a9487a929c1e232412b7704ffd6fb9d427e4722eb2cb2322963f7c56L55-L57), which creates an object used by that library. We would then check the `data` field of that object, which is just the `args` array (https://github.com/concrete-utopia/utopia/pull/5676/files#diff-a4c8338855ca2b6cba1cde108cc950253be1063727a563185098bbd573c2bea7L134-L145). So since we no longer need to proxy to a console pane, we don't need to do any of that, and can just check the `args` directly.

bkrmendy (approved):

editor/src/core/shared/runtime-report-logs.tsx
```diff
@@ -154,3 +154,62 @@ export function useErrorOverlayRecords(): ErrorOverlayRecords {
 
   return { errorRecords, overlayErrors }
 }
+
+function filterOldPasses(errorMessages: Array<ErrorMessage>): Array<ErrorMessage> {
```
Rheeseyb, line 158: I've moved this function (without making any changes to it) as it was causing a cyclic dependency in the tests

balazsbajorics (approved):

Issue #5654, opened 2024-05-09
Parser performance improvement: Simplify the hashing function by using full length UUIDs and remove the need to check for clashes


Rheeseyb:
We originally restricted the UIDs to be 3 characters to minimise the amount that we impact the users' code, but since we no longer print them we can make them as long as we want. If we ensure we're always using the code bounds in that hash, we can keep them consistent and don't need to worry about clashing, meaning we can completely parallelise that part.
We could probably simplify it further by just hashing the filename and code bounds rather than the actual JSX Element, which I'd bet would speed it up.

See https://github.com/concrete-utopia/utopia/pull/5547/files#diff-82cdd3e4e2dfcd2a312838a7c4c4ba895c26a9d03dd5b60957248a16915a557d for where I have introduced the code bounds into one case where we're generating the hash

PR #5683, opened 2024-05-15
Faster UID hashes

ruggi:
#5654 

**Problem:**

Our current hashing function, from `object-hash`, can be quite slow.

**Fix:**

This PR replaces `object-hash` with Murmur hashing (https://en.wikipedia.org/wiki/MurmurHash), which shows a dramatic performance improvement when the scale increases (roughly 50% on the same hydrogen project).

**Manual Tests:**
I hereby swear that:


bkrmendy (approved):

seanparsons (approved):
Lovely!

liady (approved):

Issue #5690, opened 2024-05-16
Update child node with the data picker


bkrmendy:
Consider the following element:

```tsx
<div>Hello, {name}!<div>
```

We already show a cartouche in the navigator for `{name}`, but when a new variable is picked, instead of only replacing `name`, we remove all the child elements and replace them with the newly picked variable, leading to something like

```tsx
<div>{updated.data.name}<div>
```

(notice that `Hello, ` and `!` are gone).

The expected outcome of the update is the following:
```tsx
<div>Hello, {updated.data.name}!<div>
```

Issue #5684, opened 2024-05-15
Data Picker: be able to replace children elements in an expression, be able to get the runtime value of an element that is not a prop


balazsbajorics:
The data picker's API evolved very organically, starting from only being able to pick data for a specific property of a selected element. This meant that the data picker relies on some assumptions that are no longer true:
1. the "current value" is no longer guranteed to be found in `allElementProps`
2. the target is no longer guaranteed to have a valid `usePropertyControlDescriptions` which we can use to get at the shape of the data, because propertyControls only describe props. 
3. When replacing the current data with the new data, it's either a `setProp_UNSAFE` for an ElementPropertyPath, or a `updateMapExpression` (which changes the `valueToMap` field) – but setProp_UNSAFE doesn't work correctly for replacing a single child in an array of children.

The `DataPickerPopupProps` already introduced the concept of the `DataPickerType` that can be `DataPickerForAProperty` or `DataPickerForAMapExpressionValue`. But then the internal code doesn't carry this information forward, instead just uses `null` for the PropertyPath wherever it would be applicable. 

Instead of the nullable propertyPath passed into `usePropertyControlDescriptions` and `usePropertyValue`, we should refactor the internal code too so there's a clear separation of concerns: the -actual- data picker logic needs: 
1. a currently spied value, if exists
2. the shape of the object we are looking for, if exists.

For the case of `DataPickerForAProperty`, the old code should be used.
For `DataPickerForAMapExpressionValue`, we have two options: we can keep it `map.valueToMap` specific, or we can do a similar refactor to the one in `DataReferenceCartoucheControl`, which uses `childOrAttribute: JSXElementChild; surroundingScope: ElementPath` as the input. 
I'd personally advocate for taking and exploring this latter route:
1. based on `{ childOrAttribute: JSXElementChild; surroundingScope: ElementPath }` we can get a runtime value using the fresher `variablesInScope` system
2. Instead of a map-specific `updateMapExpression` action, we could have a generic `REPLACE_ELEMENT_IN_SCOPE` action which uses `{ childOrAttribute: JSXElementChild; surroundingScope: ElementPath }` to be able to hunt down the exact JSXElementChild inside the relevant `ParseSuccess`. this could fix the bug of not being able to replace a _specific_ child in an array of children nodes, but also work for `map.valueToMap`. 
3. The generalized code could be applicable everywhere we want to show cartouches, because in every place it is correct to talk about a `childOrAttribute: JSXElementChild` as the replacement target.

[...]

PR #5695, opened 2024-05-16
Update child nodes from cartouches

bkrmendy:
# [Try it here](https://utopia.fish/p/a0e7a39b-cookie-nova/?branch_name=feature-replace-child-node-only)

## Problem
https://github.com/concrete-utopia/utopia/issues/5690

## Fix
Create an action that makes it possible to update child nodes on a node-by-node basis.

https://screenshot.click/16-30-dnbig-i1h8a.mp4

### Commit Details

- Added the `ReplaceElementInScope` action. This action takes a "replacement path" and uses that to actually perform the replace operation. Additional replacement paths will be added for setting the variables in properties or setting the left hand side of map expressions (see https://github.com/concrete-utopia/utopia/issues/5684 for details)
- A new data picker type (`DataPickerForChildNode`) is added as a stopgap measure until `ReplaceElementInScope` isn't powerful enough to handle to express all the data picker modes
- `RenderedAt` in `DataReferenceCartoucheControl` and `DataReferenceNavigatorEntry` is added as an explicit type, to specify where the cartouche is rendered. This might not be 100% correct when referring to the left hand sides of map expressions, dealing with that case is deferred to a follow-up PR

### Manual Tests
I hereby swear that:

### DIFF
--- a/editor/src/components/editor/action-types.ts
+++ b/editor/src/components/editor/action-types.ts
@@ -165,6 +165,18 @@ export interface ReplaceMappedElement {
   importsToAdd: Imports
 }
 
+export type ElementReplacementPath = {
+  type: 'replace-child-with-uid'
+  uid: string
+  replaceWith: JSXElementChild
+}
+
+export interface ReplaceElementInScope {
+  action: 'REPLACE_ELEMENT_IN_SCOPE'
+  target: ElementPath
+  replacementPath: ElementReplacementPath
+}
+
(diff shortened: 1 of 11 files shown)


seanparsons (approved):
Begone all the TODOs!

ruggi (approved):

balazsbajorics (commented):
+1

PR #5705, opened 2024-05-17
Data picker - code cleanup

bkrmendy:
## Problem
The data picker control is structured in a way that makes it hard to use for use cases other than picking data for properties

## Fix
This PR refactors the data picker so that both the variables it displays and the callback that's called when a variable is clicked are passed in from the call site. This way, we can provide the variables that make sense for properties, map expression values or child nodes.

### Commit Details
- the `UpdateMapExpression` action is removed and its functionality is rolled into `ReplaceElementInScope`
- Two new modes are added to `ElementReplacementPath` to cater to properties and mapped expression, and the `ReplaceElementInScope` action is updated to interpret these modes
- `useDataPickerButton` and `DataPickerComponent` are refactored so that the variable options to display and the callback that's called when an option is picked are passed in from the outside
- `DataPickerType` and `useDataPickerButtonListSource` are removed, because the functionality they implemented is superseded by passing in the variables and the callback

### Out of scope
https://github.com/concrete-utopia/utopia/issues/5684 mentions updating the logic that gets the runtime value of variables. This PR doesn't directly fix that, but since the logic related to variables is contained in `useVariablesInScopeForSelectedElement`, this PR unblocks making this change (because with this, the various call sites in different contexts can pass in whatever variables they deem fit).

### Manual Tests
I hereby swear that:

Related to https://github.com/concrete-utopia/utopia/issues/5684


editor/src/components/inspector/sections/component-section/data-picker-popup.tsx
```diff
@@ -481,8 +329,6 @@ function ValueRow({
           height: 28,
           marginTop: variableChildren != null && variableOption.depth === 0 ? 6 : 0, // add some space between top-level variables
           cursor: variableOption.variableInfo.matches ? 'pointer' : 'default',
-          background: currentExpressionExactMatch ? colorTheme.primary.value : undefined,
-          color: currentExpressionExactMatch ? colorTheme.white.value : undefined,
```
bkrmendy, line 485: Since the data picker UX isn't final yet, I didn't fix this up

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -127,7 +127,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`859`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`907`)
```
balazsbajorics, line 130: do you know what caused the render count increase? first glance looks like a regressed memoization

balazsbajorics (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
bkrmendy: after some discussion we agreed that this is OK, since we're still iterating on the data picker anyway

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2431,16 +2430,43 @@ export const UPDATE_FNS = {
   },
   REPLACE_ELEMENT_IN_SCOPE: (action: ReplaceElementInScope, editor: EditorModel): EditorModel => {
     return modifyUnderlyingTarget(action.target, editor, (element) => {
-      if (element.type !== 'JSX_ELEMENT' && element.type !== 'JSX_FRAGMENT') {
-        return element
-      }
+      const replacementPath = action.replacementPath
+      if (replacementPath.type === 'replace-child-with-uid') {
```
seanparsons, line 2434: Why not use a switch statement here?

seanparsons (approved):

editor/src/components/editor/actions/actions.tsx
bkrmendy: Thanks for calling it out, neither is ideal so I refactored it to a switch + helper functions in https://github.com/concrete-utopia/utopia/pull/5705/commits/0b1d170011d7854b02235684b07439cd43a7e4a4

Issue #5697, opened 2024-05-16
Fix styling of the sample store in utopia


gbalint:
You can find the utopia-compatible sample store here: https://github.com/concrete-utopia/hydrogen-may/tree/sehl-utopia
You can compare it to the real store in local dev environment or on https://utopia-1e8ff9861eb8a6796b14.o2.myshopify.dev/

Main page looks like this:

![Image](https://github.com/concrete-utopia/utopia/assets/127662/d0c5d81e-5e76-43dd-8f74-ee92663b14ec)

Instead of this:

![Image](https://github.com/concrete-utopia/utopia/assets/127662/683e0c78-6100-4c81-abdf-766c4d610521)

And the product page (click on the tote bag) is even worse, see e.g. 

![Image](https://github.com/concrete-utopia/utopia/assets/127662/568aa748-7193-4b21-986f-2fce47b1e06a)

instead of:

![Image](https://github.com/concrete-utopia/utopia/assets/127662/13de8f2b-9afc-46c0-b863-2bc6fe1d7ff9)

Rheeseyb:
Final layout issues fixed by removing `whitespace-pre-wrap` <https://github.com/concrete-utopia/hydrogen-may/pull/1/commits/67d7c14231a13fdb2c7d2421c1bfd3cc9edd11ef>

PR #5708, opened 2024-05-17
Scope editor css to only affect the editor

Rheeseyb:
**Problem:**
Editor CSS is leaking into the user projects.

**Fix:**
By wrapping all editor CSS in `@scope` rules that define upper and lower bounds, we can create what's referred to as a "donut scope" which can be applied to everything but the Canvas Container (https://developer.mozilla.org/en-US/docs/Web/CSS/@scope#scope_roots_and_scope_limits).

It's best to view this diff with whitespace hidden.

Part of #5697

### DIFF
--- a/editor/resources/editor/css/initial-load.css
+++ b/editor/resources/editor/css/initial-load.css
@@ -1,62 +1,5 @@
-@font-face {
-  font-family: 'utopian-inter';
-  src: url('../fonts/Inter-Regular.woff2') format('woff');
-  font-weight: 400;
-}
-
-@font-face {
-  font-family: 'utopian-inter';
-  src: url('../fonts/Inter-Italic.woff2') format('woff');
-  font-weight: 400;
-  font-style: italic;
-}
-
-@font-face {
-  font-family: 'utopian-inter';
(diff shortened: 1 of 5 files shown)


editor/resources/editor/css/initial-load.css
```diff
@@ -1,62 +1,5 @@
...
-  line-height: 17px;
-  overflow: hidden;
-  outline: none;
-  overscroll-behavior-x: contain;
-}
-
-a {
-  text-decoration: none;
-  color: #007aff;
 }
```
Rheeseyb, line 3: This was required as the Hydrogen project includes the `body` tag, which by default has a margin. We've always been setting the margin to zero and accidentally relying on the editor CSS bleeding

liady (approved):
very nice

editor/resources/editor/css/initial-load.css
```diff
@@ -65,39 +8,111 @@ img,
 ul,
 li,
 label {
-  user-select: none;
-  -webkit-user-select: none;
   box-sizing: border-box !important;
```
liady, line 11: doesn't this affect the canvas?

gbalint (approved):

PR #5710, opened 2024-05-20
Insert React Import For Modules

### LINKED ISSUE #5699: Check why we needed to add lots of React imports to the sample store project
I added React imports to basically all files until these errors disappeared:

Image

See https://github.com/concrete-utopia/hydrogen-may/commit/e0510771076eee95e57b5a543b3cf3923d9cdc4c

But valid react projects should work without additional imports

seanparsons:
**Problem:**
We found that a sample project required a lot of additional React imports to be added into the code when run in the editor. With it failing to a lot of `React is not defined` errors seemingly from `React.createElement` calls.

**Cause:**
In most configurations, an import for React is added when transpiling the JSX markup. Ordinarily when using our component rendering functionality this is not an issue as we effectively control that part anyway. But if the entire file is loaded through our evaluators then it needs to include those.

**Fix:**
If the code appears to be a JSX like file but does not include an import of React, we add one to the top of the file and execute that code instead.

**Notes:**
I spent some time trying to write a unit test that covers this case. However it appears that Babel is doing some kind of introspection of the environment it's running in and the behaviour is quite different where it introduces a function `_interopRequireWildcard` which it applies to the result of the `require` call which somehow breaks the React import.

**Commit Details:**
- `evaluateJs` now streamlines an import for React into the code, if there isn't already one.
- Added `addReactImport`, `includesReactImport` and `isJSXLikeFilePath` utility functions.

**Manual Tests:**
I hereby swear that:

Fixes #5699

### DIFF
--- a/editor/package.json
+++ b/editor/package.json
@@ -132,6 +132,7 @@
   },
   "dependencies": {
     "@babel/code-frame": "7.14.5",
+    "@babel/parser": "7.24.5",
     "@babel/plugin-external-helpers": "7.14.5",
     "@babel/plugin-proposal-class-properties": "7.16.7",
     "@babel/plugin-proposal-export-namespace-from": "7.14.5",
--- a/editor/pnpm-lock.yaml
+++ b/editor/pnpm-lock.yaml
@@ -45,6 +45,7 @@ specifiers:
   '@babel/cli': 7.23.4
   '@babel/code-frame': 7.14.5
   '@babel/core': 7.23.6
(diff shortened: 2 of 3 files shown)


balazsbajorics (commented):
could you add a test case to evaluator.spec.ts?

seanparsons:
> could you add a test case to evaluator.spec.ts?

There's a note in the description explaining why I couldn't. :)

Rheeseyb (approved):

PR #5702, opened 2024-05-17
Longer UIDs without existing lookup

ruggi:
**Problem:**
Our UIDs were too short.

**Fix:**
I made them longer

**Commit Details:** (< vv pls delete this section if's not relevant)

- Took each UID
- Added characters
- Put it back

**Manual Tests:**
I hereby swear that:

- [ ] the UIDs are longer
- [ ] I have not created duplicate UIDs

Fixes #[ticket_number] (<< pls delete this line if it's not relevant)

PR #5712, opened 2024-05-21
32-chars UIDs

ruggi:
Part of #5654 

Sharded off of https://github.com/concrete-utopia/utopia/pull/5702

**Problem:**

Our hashes are 3 characters long, which means a lot of chance for collisions. In order to be able to parallelize the parser we need to stop relying on collision lookups, which needs longer UIDs as a prerequisite.

**Fix:**

This is sharded off of the spike in https://github.com/concrete-utopia/utopia/pull/5702, making the UIDs 32-chars long (matching both UUID v4's and the 128-bit object hashes).

This is sharded in order to make the changes understandable, since the next steps involve changing the invocation of the hashes (including the filename and the bounds) as well as dropping the existing UIDs lookups.

**👀 Reviewers**

The relevant logic changes are in https://github.com/concrete-utopia/utopia/pull/5712/commits/c2d1882fac8269a6259b4933a5a9d99b4bcfadd5 .

**Manual Tests:**
I hereby swear that:


bkrmendy (approved):

ruggi:
closing for now, will revisit later

PR #5716, opened 2024-05-21
NavigatorTree

balazsbajorics:
**Context:**
We want the navigator structure to be less scary to change. The primary short term need is that we want to be able to take the existing navigator tree, and play with it: condense some parts of it into a single row, or maybe even skip elements so they don't appear in the navigator but their ancestors show up correctly indented.

I was exploring how to achieve a condensed row without losing all of the existing navigator rows in the process. I think I found a way that takes us iteratively to an easier-to-maintain navigator codebase.

# This PR:
This PR Introduces 3 major concepts:
### Problem:
On master, `getNavigatorTargets` takes the `ElementPathTree`, and does a tree walk using the recursive `walkAndAddKeys` function. 
getNavigatorTargets was originally written to be a simple tree-flattening function in the shape of `ElementPathTree -> Array<NavigatorTarget>`, and the inner walkAndAddKeys does this by pushing flattened rows into an accumulator array.
A problem happened when we started using getNavigatorTargets to add extra functionality to the navigator: support for Conditionals and Maps and render prop keys were all added superficially onto walkAndAddKeys, keeping its original tree flattening walk structure, but using it to add extra elements along the way, and combine various data sources. 
We ended up having a system which _implicitly_ created a navigator tree while flattening the ElementPathTree. This meant that reading the code became quite a challenge: you had to build a mental model of the implicit tree, and then immediately consider how would it look like if it was flattened. 
One example of the headache caused by this: since NavigatorTarget for the individual rows is not tree-aware, it meant that the [`navigatorDepth()`](https://github.com/concrete-utopia/utopia/blob/87bb87abb0f1c84e63b2ed6cfdf6421b97ae0516/editor/src/components/navigator/navigator-utils.ts#L81-L107) helper function which gets the indentation level of an individual row has to back-solve whether the element has a conditional true-false label as a parent because then it needs to go an extra level deeper. 
All these little paper cuts add up: the simple ask of "what if these three elements are now condensed to a single row and all their children are then indented three over to the left" means adding even more patches and backsolvers all over the navigator codebase.

### Solution:
getNavigatorTargets is replaced by a `getNavigatorTrees` which transforms the `ElementPathTree` into a semantic `NavigatorTree`. The `NavigatorTree` combines the data sources and adds the extra information the navigator needs: it has dedicated types for the Conditional and Map and Render Props, and it makes it easy to add new kinds of navigator tree types. The sole purpose of `getNavigatorTrees` is to create the NavigatorTree, it doesn't deal with hiding invisible elements or omitting collapsed children or adding fake navigator entries that correspond to the labels for the conditional clause or the render prop name.
Then there is a second step in `getNavigatorTargets` which ONLY deals with flattening the tree: it's an easy to read tree walk with a single purpose.

The end result is that `getNavigatorTargets` is functionally identical to the original output of getNavigatorTargets, and all of the navigator tests (of which there are quite a few!) pass without modification.

### Problem:
`NavigatorTarget` corresponds to a NavigatorItem which corresponds to a navigator row. This is implicit in the current navigator design, and makes it pretty hard to introduce rows with multiple elements, or extra rows which don't correspond to any element. This is why the Conditional Clause and Render Prop Labels had to be injected as pseudo elements in the NavigatorTarget array. This obviously makes it very painful to introduce designs where a single row can contain multiple elements.
### Solution:
Introducing the `NavigatorRow` type which represents a "physical" navigator row.
The new full data flow is this:
`(ElementPathTree + Metadata + ProjectContentsTree) -> Array<NavigatorTree> -> Array<NavigatorRow> -> extracting individual NavigatorTargets`
This refactor is fully transparent because so far we only use the `RegularNavigatorRow` which contains a single `NavigatorTarget` per row, so retains the original design of one item per row. 
I left the pseudo NavigatorTargets as they were, to minimize the size of this PR, but in the future those can be turned into components rendered by various row types, without them having to be navigator items themselves.

## Actual Condensed Rows?
This PR also introduces `condenseNavigatorTree` which walks the NavigatorTree and demonstrates that now we have the ability to transform the tree as an isolated concern, and we can experiment with various rulesets without having to touch `getNavigatorTrees` or `getNavigatorRowsForTree`. If the feature switch `'Condensed Navigator Entries'` is off, `condenseNavigatorTree` does nothing. 
If 'Condensed Navigator Entries' is on, we generate a `CondensedNavigatorRow` which is rendered as an ugly placeholder in the navigator. The condenseNavigatorTree and the CondensedNavigatorRow will be the entry points for @ruggi 's upcoming PR.


editor/src/components/navigator/navigator-item/navigator-item-wrapper.tsx
```diff
@@ -218,8 +220,57 @@ export function getNavigatorEntryLabel(
...
+          windowStyle={props.windowStyle}
+          navigatorEntry={navigatorEntry}
+        />
+      )
+    }
+    return (
+      <div
+        style={{
+          ...props.windowStyle,
+          left: 5 + 12 + 6 + BasePaddingUnit * props.navigatorRow.indentation,
```
bkrmendy, line 244: what do these numbers mean?

balazsbajorics: this is the placeholder code which we will throw away in the next PR :) it's only visible with the FS on, which is not on by default

editor/src/components/navigator/navigator-utils.ts
```diff
@@ -79,317 +106,677 @@ export function navigatorDepth(
...
+  navigatorEntry: NavigatorEntry
+  mappedEntries: Array<NavigatorTree>
+}
+
+type ConditionalNavigatorTree = {
+  type: 'conditional-entry'
+  elementHidden: boolean
+  subtreeHidden: boolean
+  navigatorEntry: NavigatorEntry
+  trueCase: Array<NavigatorTree>
```
gbalint, line 138: Do we really need an array here? Conditional branches contain a single element.

editor/src/components/navigator/navigator-utils.ts
```diff
@@ -79,317 +106,677 @@ export function navigatorDepth(
...
+  let currentChild: NavigatorTree = entry.child
+  while (currentChild.type === 'condensed-trunk') {
+    singleRow.push(currentChild.navigatorEntry)
+    currentChild = currentChild.child
+  }
+  return { singleRow: singleRow, child: currentChild }
+}
+
+function getNavigatorRowsForTree(
+  navigatorTree: Array<NavigatorTree>,
```
gbalint, line 643: In most places where you use a tree, the type is an array of trees. Is that really necessary? It could be a single tree with a root.
If it is really necessary, let's decide if we use singular or plural in the name, because e.g. getNavigatorRowsForTree is singular, but getNavigatorTrees is plural, although they use the same type.

balazsbajorics: I followed the inconsistency of the `elementPathTree: ElementPathTrees` :D

editor/src/components/navigator/navigator-utils.ts
balazsbajorics: the Conditional branch being a single element is a limitation of the parser, felt easier and cheaper to not assume that it can only be a single element (and then put in guards to ensure that we never encounter a conditional with multiple elements)

editor/src/components/editor/store/editor-state.ts
```diff
@@ -2364,6 +2365,16 @@ export function navigatorEntriesEqual(
   }
 }
 
+export function navigatorRowToKey(row: NavigatorRow): string {
+  if (row.type === 'regular-row') {
```
ruggi, line 2369: why not a `switch` here?

balazsbajorics: fixed

ruggi (approved):

liady (approved):

gbalint (approved):

Issue #5696, opened 2024-05-16
Extend wrap functionality


liady:
We need the new "Wrap..." component picker to have these fuctionalities:

- [x] Wrap in non-elements (conditionals, etc)
- [x] Wrap multiple elements

PR #5722, opened 2024-05-22
Wrap in List

### LINKED ISSUE #5419: Wrap in List!
Related to #5442

liady:
**Problem:**
We cannot currently wrap with a List

**Fix:**
This PR fixes wrapping a single element with a list.

**Details:**
1. Adding a List as an option for "Wrap..."
2. Adjust the code so that wrapping in a map (which is currently two `INSERT`s in disguise) will actually `REPLACE` the map contents
3. NOTE - this PR supports a single element wrapping.

For multiple element wrapping see #5696

<video src="https://github.com/concrete-utopia/utopia/assets/7003853/0dfc0062-8c9a-4e19-9bfc-5a138f0a2df3"></video>

Fixes #5419


editor/src/components/editor/store/insertion-path.ts
```diff
@@ -68,6 +68,23 @@ export function childInsertionPath(
   }
 }
 
+export interface MapInsertionPath {
+  type: 'MAP_INSERTION'
+  intendedParentPath: StaticElementPath
+  insertBehavior: ConditionalClauseInsertBehavior
```
Rheeseyb, line 74: This type should probably be renamed

bkrmendy: for the note, we agreed not to go with `MapInsertionPath` at all

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -2318,6 +2318,15 @@ export const MetadataUtils = {
     const element = MetadataUtils.findElementByElementPath(metadata, target)
     return MetadataUtils.isJSXMapExpressionFromMetadata(element)
   },
+  isJSXElementFromMetadata(element: ElementInstanceMetadata | null): boolean {
```
Rheeseyb, line 2321: I'm surprised these two don't already exist, though we do have `getJSXElementFromMetadata` and `getJSXElementFromElementInstanceMetadata`

liady: yeah, I followed the `isConditionalFromMetadata`, `isJSXMapExpressionFromMetadata`, etc pattern here - should I leave it like this (for completion and for the next developer who will need it and look for it) or do you think I need to use the `getJSXElement...` methods?

bkrmendy (approved):

editor/src/core/model/element-metadata-utils.ts
liady: so maybe we should leave this method for completion and for next developers, if that makes sense

Rheeseyb: Yeah, keep the function 👍

Issue #5635, opened 2024-05-08
Make style section hideable and suppressable


maltenuhn:
**Problems**
- The style section and its subsections are difficult to work with and understand, especially for components. 
- Additionally, the completely hiding the style section was not well received.

**Solution**
_Part 1_
- [x] Create a new "Styles" (sic) Section
- [x] Make that capable of folding in and out **the entire set of existing style subsections**
- [x] use the Utopia API to expand, collapse, or hide this new section (bikeshed the naming)

_Part 2_
- [x] Create a new Section called "Advanced"
- [x] Relabel the Target Selector subsection / component as "Target" (not 'style')
- [x] Relabel the CSS Classname subsection (the one we show when Tailwind is enabled) as 'CSS'
- [x] Move them into the "Advanced" section
- [x] Make the 'advanced' subsection closed contracted by default (so you don't see the Target and CSS subsections

_Part 3: Defaults and Component API_ 
- [x] For components, collapse the new Styles section by default. 
- [x] API should allow showing it as collapsed, expanded, and hidden (bikedshed labels...), and section should respect that
- [x] for elements, expand the Styles section by default

bkrmendy:
https://github.com/concrete-utopia/utopia/pull/5714 adds the component API prop to expand/collapse/hide the Style section

bkrmendy:
https://github.com/concrete-utopia/utopia/pull/5723 adds the component API parts

PR #5723, opened 2024-05-22
Inspector sections - component API

bkrmendy:
# [Project with updated inspector annotations](https://utopia.fish/p/89e72656-humorous-whimsey/?branch_name=feature-style-section-component-api-defaults)

https://github.com/concrete-utopia/utopia/issues/5635

## Description
After https://github.com/concrete-utopia/utopia/pull/5719 added the foldable style sections, this PR makes them configurable via the component API through the existing `inspector` prop.

The new schema of the `inspector` prop is
```
inspector?:
    'hidden' |
    {
        display?: 'collapsed' | 'expanded',
        sections: Array<"layout" | "layout-system" | "visual" | "typography">
    }
```

- if the `inspector` prop is not specified for the selected element(s), both the Styles and the Advanced sections are shown. If the selected element is an intrinsic element, a Fragment or a Group from `utopia-api`, the style section is expanded, otherwise it's collapsed by default
- `hidden` hides both the Styles section and the Advanced section (the Component section will still be visible)
- `display?: 'collapsed' | 'expanded'` shows the style section collapsed/expanded by default. If it's omitted, the defaults from the first bullet point apply
- `sections` specifies which sections are visible in the Styles section. If it's omitted, all sections are visible

### Manual Tests
I hereby swear that:


editor/src/components/inspector/inspector.spec.browser2.tsx
```diff
@@ -444,7 +444,7 @@ export var storyboard = (
         </Storyboard>
       )
       `,
-        ['components.utopia.js']: `import { Card } from './storyboard'
+        ['/utopia/components.utopia.js']: `import { Card } from './storyboard'
```
gbalint, line 447: wow, how could this work earlier?

editor/src/components/editor/store/store-deep-equality-instances.ts
```diff
@@ -3480,7 +3468,7 @@ export const ComponentDescriptorKeepDeepEquality: KeepDeepEqualityCall<Component
     (descriptor) => descriptor.focus,
     createCallWithTripleEquals<Focus>(),
     (descriptor) => descriptor.inspector,
-    InspectorSpecKeepDeepEquality,
+    createCallFromIntrospectiveKeepDeep(),
```
Rheeseyb, line 3471: IIRC this is quite expensive, so we try to avoid it unless it's absolutely necessary

editor/src/components/inspector/inspector.spec.browser2.tsx
bkrmendy: it didn't 🤦‍♂️🙈 the tests only asserted what was visible, not what wasn't visible. I'm going back and updating the tests to be more resilient

Rheeseyb: Yeah this is disturbing. I just checked the code and we always check that the filename is prefixed with `/utopia/`, implying that this didn't work

gbalint (approved):

editor/src/components/inspector/inspector.tsx
```diff
@@ -904,3 +929,76 @@ function useBoolean(starting: boolean): {
...
+          isIntrinsicElementMetadata(instance)
+        )
+      }),
+    'useShouldExpandStyleSection shouldExpandFromElementOrComponent',
+  )
+
+  return useEditorState(
+    Substores.propertyControlsInfo,
+    (store) => {
+      for (const target of store.editor.selectedViews) {
```
Rheeseyb, line 954: You've wrote a for loop here but return on the first result? Also, you could early return here if `shouldExpandFromElementOrComponent` is true

balazsbajorics: a few weeks ago I created `mapFirstApplicable` which has the syntax of map, and runs until it gets the first non-null return value

Rheeseyb: This has no non-null return values though

balazsbajorics: oh

seanparsons (approved):

editor/src/components/inspector/inspector.spec.browser2.tsx
bkrmendy: I updated the tests in https://github.com/concrete-utopia/utopia/pull/5723/commits/9cc3fecda2031deb6c6f7e485675d6da91fbe0bd

editor/src/components/inspector/inspector.tsx
bkrmendy: The for loop was a mistake, I updated it to an `every` https://github.com/concrete-utopia/utopia/pull/5723/commits/f638923e97bde849807457aea6d394139e2e4fab. The intention is to bias towards collapsing the Styles section. Regarding the early return, I'd prefer to keep it because I'd have to use a fallback for the `descriptor == null || descriptor.inspector == null` case anyway

balazsbajorics (approved):

PR #5725, opened 2024-05-22
Update react-contexify to 6.0.0

benwolfram:
**Problem:**
As apart of modernizing and updating context menus, bumping react-contexify to the latest version following their [migration guide](https://fkhadra.github.io/react-contexify/migrate/).

**Fix:**'

This PR fixes keyboard navigation for the context menu, however, events are still bubbling up to the editor. Will look at this next.

- Bumping to version 6.0.0
- Changing how `show` is called, with the event
- Updating classes and redoing the styles
- Removing explicit 28px heights
- Using `onVisibilityChange`
- Typescript type fixes

- Removing the existing patch?
  - I am not sure what this is even, but I realized it was tied to 5.0.0 version, so it seemed irrelevant now. I assume there is a good reason why it exists, so please let me know if this needs to be redone for v6 and how that is done

**Manual Tests:**
I hereby swear that:

- [x] I opened a hydrogen project and it loaded
- [x] I could navigate to various routes in Preview mode

Relates to https://github.com/concrete-utopia/utopia/issues/5643

PR #5736, opened 2024-05-23
Manage react-contexify's body class state with an atom

benwolfram:
**Problem:**
Following https://github.com/concrete-utopia/utopia/pull/5725, fixing keyboard navigation with react-contexify's Context Menu, there is a regression is keyboard events bleeding into the canvas. This occurs because `onVisibilityChange` triggers with `true` when one context menu opens, but also `false` for all the other menus at the same time.

**Fix:**
Keep track of which menus is open (using ids) and apply the body class dynamically based on this state.

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/context-menu-state.tsx
```diff
@@ -0,0 +1,41 @@
+import { atom, useAtom } from 'jotai'
+
+import { BodyMenuOpenClass } from '../../core/shared/utils'
+import React from 'react'
+
+export const ContextMenuStateAtom = atom([] as string[])
+export const ContextMenuOpen = atom((get) => {
+  const contextMenusOpen = get(ContextMenuStateAtom)
+  return contextMenusOpen.length > 0
+})
```
benwolfram, line 10: Would love to use this state [here](https://github.com/concrete-utopia/utopia/blob/403e73d4da4a260fb41f19c229ace7f488e99867/editor/src/components/editor/editor-component-common.tsx#L57) (possible elsewhere) so I decided to put this in it's own derived atom. But I don't believe I am consume it there as the function isn't a hook and I didn't really wanna parameter drill it down. Open to feedback tho.

liady: We can use the atoms outside of React by using [stores](https://jotai.org/docs/guides/using-store-outside-react) -
```ts
const store = getDefaultStore()
const isMenuOpen = store.get(ContextMenuOpen)
```
and then we don't have to rely on class name for the keyboard event check

editor/src/components/canvas/context-menu-state.tsx
```diff
@@ -0,0 +1,41 @@
...
+  React.useEffect(() => {
+    if (isContextMenuOpen) {
+      document.body.classList.add(BodyMenuOpenClass)
+    } else {
+      document.body.classList.remove(BodyMenuOpenClass)
+    }
+  }, [isContextMenuOpen])
+
+  const add = (id: string) => {
+    if (values.includes(id)) return
```
liady, line 25: style - I think the convention is for always using if blocks (instead of a single line)

editor/src/components/context-menu-wrapper.tsx
```diff
@@ -147,6 +143,15 @@ export const ContextMenu = <T,>({ dispatch, getData, id, items }: ContextMenuPro
     [getData, dispatch, isDisabled, isHidden],
   )
 
+  const { add, remove } = useContextMenuState()
```
liady, line 146: just an idea - you can return an `onVisibilityChange` function from the hook and use it directly

editor/src/components/canvas/context-menu-state.tsx
benwolfram: Sick, this is exactly what I was looking for. Thanks @liady, def my first time using this library and atoms!

editor/src/components/context-menu-wrapper.tsx
benwolfram: I really prefer not to do that tbh, I know there is a lot of 'returning functions' in the codebase. Personally, my hot take is that it's probably unnecessary and is a lot more confusion to deal with.

Feel like this interface does a good job of decoupling state/data from the UI component.

editor/src/components/canvas/context-menu-state.tsx
benwolfram: We should put this stuff in eslint 🙈

editor/src/templates/editor.tsx
```diff
@@ -693,7 +695,7 @@ export const EditorRoot: React.FunctionComponent<{
 }) => {
   return (
     <AtomsDevtools>
-      <JotaiProvider>
+      <JotaiProvider store={jotaiStore}>
```
benwolfram, line 698: Using an explicit store instead of relying on internal default, so that this can be accessed.

editor/src/components/canvas/context-menu-state.tsx
```diff
@@ -0,0 +1,47 @@
+import { atom, useAtom } from 'jotai'
+
+import { BodyMenuOpenClass } from '../../core/shared/utils'
+import React from 'react'
+
+export const ContextMenuStateAtom = atom([] as string[])
```
balazsbajorics, line 6: in general if we can use the generic parameter `<P>`, we prefer it to casting with `as P` for increased type safety.

Even minorer, but we use `Array<string>` instead of the `string[]` notation.

```suggestion
export const ContextMenuStateAtom = atom<Array<string>>([])
```

editor/src/components/canvas/context-menu-state.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const useContextMenuState = () => {
+  const [values, setValues] = useAtom(ContextMenuStateAtom)
+
+  React.useEffect(() => {
+    if (values.length > 0) {
+      document.body.classList.add(BodyMenuOpenClass)
+    } else {
+      document.body.classList.remove(BodyMenuOpenClass)
+    }
+  }, [values])
```
balazsbajorics, line 21: now that we have this atom, I think we should kill `BodyMenuOpenClass`. it can be done in a follow-up PR tho

balazsbajorics (approved):

balazsbajorics:
the `adjust opacity with the keyboard` Karma tests seem to have been broken

benwolfram:
Fixed by https://github.com/concrete-utopia/utopia/pull/5889

Issue #5740, opened 2024-05-24
Component annotation variants should be added to the insert menu even when the component can not be parsed


gbalint:
Sample project contained components using forwardRef, these are parsed as arbitrary Javascript right now and shouldn't be.
See https://github.com/concrete-utopia/utopia/issues/5739

When a component has annotations and there are variants in it, the variants should be added to the insert menu even when the component is not parseable.
How it works today is that we add components to the insert menu from the parsed model, and we only check the component annotations for these components. However, for external components we add them to the insert menu just because they have component annotations. This should work the same way for internal components too.

PR #5741, opened 2024-05-24
Even unparsed components should be added to the insert menu when they have component annotations

gbalint:
**Problem:**
See https://github.com/concrete-utopia/utopia/issues/5740
Sometimes components can not be parsed (e.g. when they are in a React.forwardRef or React.memo).
However, when they have component annotations, we can trust that they are actually components, and we can add them (or their variants, when those are defined in the annotations) to the insert menu

**Manual Tests:**
I hereby swear that:

Fixes https://github.com/concrete-utopia/utopia/issues/5740


bkrmendy (approved):

editor/src/components/editor/export-utils.ts
```diff
@@ -58,17 +62,27 @@ export function getExportedComponentImportsFromParseSuccess(
     listingName: string,
     importDetailsToAdd: ImportDetails,
   ): void {
-    for (const topLevelElement of success.topLevelElements) {
-      if (
+    const isExportedComponent = success.topLevelElements.some(
```
Rheeseyb, line 65: This was throwing me off and it's because this isn't actually checking if the component is exported, but is just checking that it is parsed. So I think we should be renaming this to `isParsedComponent`

Rheeseyb (approved):

PR #5745, opened 2024-05-24
NavigatorTree: subtreeHidden support in CondensedTrunk 

balazsbajorics:
**Problem:**
The `CondensedTrunk` navigator tree entry type did not support subtreeHidden, which meant a hack had to be implemented in the form of `filterCollapsedNavigatorRows`. 

Since we have collapsed navigator row management built into `getNavigatorRowsForTree`, we shouldn't have _two_ places managing it.

**Fix:**
Add support for `subtreeHidden` and `elementHidden` in `CondensedTrunk`. 

**Commit Details:** 

- add these two fields to CondensedTrunk
- rejig `walkTree` so it only recurses if subtreeHidden is false

**Manual Tests:**
I hereby swear that:


editor/src/components/navigator/navigator-utils.ts
```diff
@@ -672,33 +678,37 @@ function getNavigatorRowsForTree(
   const condensedTree = condenseNavigatorTree(navigatorTree)
 
   function walkTree(entry: NavigatorTree, indentation: number): Array<NavigatorRow> {
+    function walkIfSubtreeVisible(e: NavigatorTree, i: number): Array<NavigatorRow> {
+      if (
+        filterVisible === 'visible-navigator-targets' &&
+        'subtreeHidden' in entry &&
```
ruggi, line 684: I missed this yesterday, I think this is a little too opaque, a helper function would be more readable imo

ruggi (approved):

editor/src/components/navigator/navigator-utils.ts
balazsbajorics: fixed in 65ce1aed1

PR #5758, opened 2024-05-27
Feat: Wrap multiple elements using the new picker

### LINKED ISSUE #5696: Extend wrap functionality

liady:
**Problem:**
Our current picker does not support multiple selections, we need this functionality for wrapping multiple elements

**Fix:**
Make the new picker accept multiple selections.

Important note: this PR adds the functionality of wrapping multiple elements, but for replacment and inserts the target remains only the first selection (as it was before).
We need to decide later on what to do about insertion/replacements

<video src="https://github.com/concrete-utopia/utopia/assets/7003853/1825e5a1-b695-4c87-8bb1-89516728436c"></video>

**Manual Tests:**
I hereby swear that:

Fixes #5696


balazsbajorics:
I've left a message on discord about special casing some wrap behaviors for div, fragment and group

Rheeseyb (approved):

balazsbajorics (approved):
we should not forget about re adding the special cased wrapping behavior for div, group, fragment

editor/src/components/navigator/navigator-item/component-picker-context-menu.tsx
```diff
@@ -485,7 +489,7 @@ function insertComponentPickerItem(
       if (isRenderPropTarget(insertionTarget)) {
         return [
           setProp_UNSAFE(
-            target,
+            targets[0],
```
seanparsons, line 492: Better to capture the value with `targets.at(0)`, which handles the empty array case correctly. Otherwise this will shove an undefined value all the way down until something breaks because the subscript type is a little naughty.

seanparsons (changes requested):
Better to avoid those subscripts because they're a bug waiting to happen.

editor/src/components/navigator/navigator-item/component-picker-context-menu.tsx
liady: yes, that was my first try (with the "at") - byt this requires a much larger refactor in many places to support undefined (which they shouldn't). Since the original code expected `ElementPath` up in the calling chain (and I changed it to `ElementPath[]`), I think there won't be any empty array there.

image

seanparsons (approved):

PR #5760, opened 2024-05-27
Allow loading project assets from the root

### LINKED ISSUE #5759: Allow loading project assets that reference the project root

ruggi:
**Problem:**

Projects may reference assets relative to the project root, e.g. a font being loaded from a CSS file.

In this case they wouldn't be found on the server, because they are not referencing an actual project ID.

**Fix:**

Add a catchall route that once hit:

1. checks that the requested resource is an allowed extension
2. checks that the request's referer is compatible to a project's url, and extract the project ID from there
3. checks that the requester can access the project
4. if all the above pass, the request is proxied to `/p/{id}/...` so it can be correctly resolved

**Manual Tests:**
I hereby swear that:

Fixes #5759

### DIFF
--- a/utopia-remix/app/handlers/validators.ts
+++ b/utopia-remix/app/handlers/validators.ts
@@ -26,51 +26,73 @@ export function validateProjectAccess(
       return validationError(new ApiError('Invalid project id', Status.BAD_REQUEST))
     }
 
-    const ownership = await getProjectOwnership(
-      { id: projectId },
-      { includeDeleted: includeDeleted },
-    )
-    if (ownership == null) {
-      return validationError(new ApiError('Project not found', Status.NOT_FOUND))
-    }
+    return canAccessProject({
+      projectId: projectId,
+      includeDeleted: includeDeleted,
+      canRequestAccess: canRequestAccess,
+      request: req,
+      permission: permission,
+    })
+  }
+}
 
-    const user = await getUser(req)
-    const userId = user?.user_id ?? null
+export async function canAccessProject({
+  projectId,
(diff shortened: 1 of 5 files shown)


utopia-remix/app/handlers/validators.ts
```diff
@@ -26,51 +26,73 @@ export function validateProjectAccess(
...
     }
 
-    const ownership = await getProjectOwnership(
-      { id: projectId },
-      { includeDeleted: includeDeleted },
-    )
-    if (ownership == null) {
-      return validationError(new ApiError('Project not found', Status.NOT_FOUND))
-    }
+    return canAccessProject({
```
ruggi, line 29: Nothing changed here, just factored out the logic into `canAccessProject`

utopia-remix/app/routes/$.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * This function will look at the incoming request and if the requested resource
+ * has an allowed extension it will try to derive the project ID
+ * from the request's referer header.
+ * If a match is found, and the project can be accessed, the request will be proxied to
+ * /p/<id>/<resource>.
+ */
+async function handleLoadAssetFromRoot(req: Request) {
+  const proxyProjectAssetPath = await getProxyAssetPath(req)
+  if (proxyProjectAssetPath == null) {
+    return {}
```
ruggi, line 27: not returning anything specific here to avoid leaking any information

balazsbajorics (approved):

gbalint (approved):

PR #5790, opened 2024-05-30
Data Picker: be able to pick global variables

balazsbajorics:
**Problem:**
Globally declared variables were missing from the data picker.

To investigate, I created a sample project with nested scopes and I've identified two missing cases, the first one is: globally declared variables in the code file. 
I had to re-acquaint myself with the render-core-element codebase, and with @seanparsons 's help I realized what we are missing was `runBlockUpdatingScope` returning spied variables.

**Fix:**
The solution was to extend the file-scope-level `mutableContext` and add a new field `spiedVariablesDeclaredInRootScope: VariableData` in there.

**Commit Details:** 
- Added a new FileRootPath type so it's now `insertionCeiling: FileRootPath | ElementPath`, this lets us not use `null` or `EP.emptyPath` as the "file root" reference
- New easier-to-read test case in `scoped-variables.spec.tsx`
- runBlockUpdatingScope returns `spiedVariablesDeclaredWithinBlock: VariableData`
- feed spiedVariablesDeclaredWithinBlock into `mutableContext.spiedVariablesDeclaredInRootScope`
- extend `spiedVariablesInScope` in `createComponentRendererComponent` to include the root scope variables

**Manual Tests:**
I hereby swear that:


balazsbajorics:
**Note:**

The _other_ missing scoped variable I found is this:
```typescript
<div>
  {(() => {
    const innerVar = ['cica', 'kutya']
    return innerVar.map((inner) => {
      return (
        <div>
          {inner}
        </div>
      )
    })
  })()}
</div>
```

Neither innerVar nor inner shows up in the spied variablesInScope. 

I think the solution is to add `definedWithin` to `JSExpressionOtherJavaScript`, similar to how it's done in `ArbitraryJSBlock`, then `resolveParamsAndRunJsCode` can return a `spiedVariablesDeclaredWithinBlock`, which we can use in `renderCoreElement`.

(Actually at this point I'm not sure why we have a separate JSExpressionOtherJavaScript and ArbitraryJSBlock, feels like they used converged to the point that we'd benefit from them being the same, or ArbitraryJSBlock to just have a single JSExpressionOtherJavaScript in it)

bkrmendy (approved):

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -101,9 +101,43 @@ export const ElementsToRerenderGLOBAL: { current: ElementsToRerender } = {
...
+
+export function fileRootPathFromString(fileRootPathString: string): FileRootPath {
+  if (fileRootPathString === 'file-root') {
+    return { type: 'file-root' }
+  } else {
+    throw new Error(`Unknown file root path string: ${fileRootPathString}`)
+  }
+}
+
+export function insertionCeilingToString(insertionCeiling: ElementPath | FileRootPath): string {
```
ruggi, line 120: this is not used, is it? 🤔

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -101,9 +101,43 @@ export const ElementsToRerenderGLOBAL: { current: ElementsToRerender } = {
...
 
+export type FileRootPath = {
+  type: 'file-root'
+}
+
+export function fileRootPathToString(fileRootPath: FileRootPath): string {
+  return 'file-root'
+}
+
+export function fileRootPathFromString(fileRootPathString: string): FileRootPath {
```
ruggi, line 112: if `insertionCeilingToString` is not used, this is not used either

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -101,9 +101,43 @@ export const ElementsToRerenderGLOBAL: { current: ElementsToRerender } = {
   current: 'rerender-all-elements',
 }
 
+export type FileRootPath = {
+  type: 'file-root'
+}
+
+export function fileRootPathToString(fileRootPath: FileRootPath): string {
```
ruggi, line 108: this is effectively not used as well

editor/src/components/canvas/ui-jsx-canvas.tsx
balazsbajorics: it is used in `scoped-variables.spec.tsx`

ruggi: oh apologies, I misclicked - I meant `insertionCeilingFromString`

ruggi (approved):

PR #5794, opened 2024-05-30
Fix: wrap tests

### LINKED ISSUE #5779: Restore removed tests of Wrap functionality
Restore the tests that were removed here: https://github.com/concrete-utopia/utopia/pull/5778/commits/75c4e6c85b7457a2a1ca90aadb7fa60d3583f7d3

liady:
This PR re-adds the skipped tests and fixes them. Now we pass the specific mode to the picker to show only the components that support children, an in addition `List` is being [special cased](https://github.com/concrete-utopia/utopia/pull/5794/files#diff-aae3a971fe8ad7522e15665e7202a4386a58a51bb5d7ec3feaa1d87f9fb073a5R675) to appear in the wrapping component list since we special-cased it in the insertion process

**Manual Tests:**
I hereby swear that:

Fixes #5779


editor/src/components/shared/project-components.ts
```diff
@@ -565,7 +565,7 @@ export const fragmentElementsDescriptors: ComponentDescriptorsForFile = {
 export const mapElementDescriptors: ComponentDescriptorsForFile = {
   map: {
     properties: {},
-    supportsChildren: false,
+    supportsChildren: true,
```
liady, line 568: since we do allow wrapping in a list now

ruggi (approved):

editor/src/components/shared/project-components.ts
gbalint: This can be problematic, because it enables insertion (e.g. the plus button in the navigator), even though we can not insert into a list.

liady: reverted it back to not supporting children, and replaced the logic to special-case the list element

PR #5812, opened 2024-06-02
silent builds

liady:
Try reducing the console output for our CI builds


liady:
Temporarily closing as Spike

PR #5815, opened 2024-06-03
Data Picker: support scope slices

balazsbajorics:
image

image

**Problem:**
When picking data, if the data picker shows all the variables that are potentially in scope, it can be overwhelming, and hard to navigate.
The classic devtools approach is to show scopes below each other in a tree-like structure, each scope toggleable to be visible or hidden. This approach clashes with the current design of the data picker, and it also felt too developery for our needs.

**Solution**
Show a breadcrumb-style row which lists the available scopes. By default, the innermost scope is activated because most of the time that is the most relevant scope to pick data from for the UI. The user can click around in the breadcrumb UI to check other scopes.

**Details**
- `lowestInsertionCeiling: ElementPath` added to `DataSelectorModalProps`
- `insertionCeiling: ElementPath` to VariableOption
- the `DataSelectorModal` organizes variables into scope buckets, and only shows one bucket at a time
- I added @bkrmendy 's breadcrumb UI and related helper functions from #5802

**Follow Up Tasks:**
- [ ] when opening the data picker for an existing value, we should pre-navigate to the correct scope
- [ ] switching scopes makes the UI jump around -> we should make the data picker fixed size
- [ ] The component scope is sometimes shown as Outlet, we should fix it to properly show the component's name. This limitation comes from the current navigator's labeling helper function


editor/src/components/inspector/sections/component-section/data-picker-utils.tsx
```diff
@@ -1,50 +1,48 @@
...
+import type { AllElementProps } from '../../../editor/store/editor-state'
 import type { ArrayInfo, JSXInfo, ObjectInfo, PrimitiveInfo } from './variables-in-scope-utils'
-import { processJSPropertyAccessors } from '../../../../core/data-tracing/data-tracing'
-import { foldEither } from '../../../../core/shared/either'
 
-export interface PrimitiveOption {
-  type: 'primitive'
-  variableInfo: PrimitiveInfo
-  definedElsewhere: string
+interface VariableOptionBase {
```
balazsbajorics, line 18: VariableOptionBase is a new common base type all VariableOptions extend, this makes the code a little bit easier to read

bkrmendy (approved):

editor/src/components/inspector/sections/component-section/data-selector-modal.tsx
```diff
@@ -356,6 +417,28 @@ export const DataSelectorModal = React.memo(
...
+                        padding: '2px 4px',
+                        borderRadius: 4,
+                        cursor: 'pointer',
+                        fontSize: 12,
+                        fontWeight: EP.pathsEqual(selectedScope, scope) ? 800 : undefined,
+                      }}
+                    >
+                      {label}
+                    </div>
+                    {idx < a.length - 1 ? (
```
ruggi, line 436:
```suggestion
                    {when(idx < a.length - 1), ( ...
```

ruggi (approved):

PR #5832, opened 2024-06-04
Fix/data picker fixes more

balazsbajorics:
image

An assortment of small data picker fixes:

- flipping breadcrumbs and the "data path input field", because the logical scope is breadcrumb -> data path, for example: `Reviews >  data.reviews[0].quote`. How it is on master makes it look like we are inside data.reviews[0].quote and we have options File > Reviews to further narrow the scope, which is backwards.
- adding an empty state
- various minor css tweaks
- not allowing clicking on breadcrumbs that point at an empty scope slice, these breadcrumbs are also subdued
- fixing vite hot reload for tweaking the data modal by unexporting `function pathBreadcrumbs`
- When opening the data picker with an already selected data, navigate into the object hierarchy to be able to show the selected Cartouche
- Outlet Name Hack – if we find that the Scope Breadcrumb would be named Outlet, we try to manually look up the component name from the parsed model

### DIFF
--- a/editor/src/components/inspector/sections/component-section/data-selector-modal.tsx
+++ b/editor/src/components/inspector/sections/component-section/data-selector-modal.tsx
@@ -182,6 +182,7 @@ export const DataSelectorModal = React.memo(
             store.editor.jsxMetadata,
             store.editor.allElementProps,
             store.editor.elementPathTree,
+            store.editor.projectContents,
             Object.keys(scopeBuckets),
             lowestInsertionCeiling ?? EP.emptyElementPath,
           )
--- a/editor/src/components/inspector/sections/component-section/data-selector-modal.tsx
+++ b/editor/src/components/inspector/sections/component-section/data-selector-modal.tsx
@@ -411,6 +411,32 @@ export const DataSelectorModal = React.memo(
                 ...style,
               }}
             >
+              {/* Scope Selector Breadcrumbs */}
+              <FlexRow style={{ gap: 2, paddingBottom: 8 }}>
+                {elementLabelsWithScopes.map(({ label, scope, hasContent }, idx, a) => (
+                  <React.Fragment key={`label-${idx}`}>
+                    <div
+                      onClick={setSelectedScopeCurried(scope, hasContent)}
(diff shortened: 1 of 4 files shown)


editor/src/core/model/element-template-utils.ts
```diff
@@ -1657,3 +1658,16 @@ export function findContainingComponentForPath(
 
   return null
 }
+
+export function findContainingComponentForPathInProjectContents(
```
balazsbajorics, line 1662: this was moved verbatim from data-tracing.ts so it can be shared

editor/src/components/inspector/sections/component-section/data-selector-modal.tsx
```diff
@@ -398,6 +411,32 @@ export const DataSelectorModal = React.memo(
...
+                {elementLabelsWithScopes.map(({ label, scope, hasContent }, idx, a) => (
+                  <React.Fragment key={`label-${idx}`}>
+                    <div
+                      onClick={setSelectedScopeCurried(scope, hasContent)}
+                      style={{
+                        width: 'max-content',
+                        padding: '2px 4px',
+                        borderRadius: 4,
+                        cursor: hasContent ? 'pointer' : undefined,
+                        color: hasContent
```
balazsbajorics, line 425: this is the only change I made here, `hasContent == false` makes the color subdued

editor/src/components/inspector/sections/component-section/data-selector-modal.tsx
```diff
@@ -517,7 +536,6 @@ export const DataSelectorModal = React.memo(
                       {primitiveVars.map((variable) => (
                         <CartoucheUI
                           key={variable.valuePath.toString()}
-                          tooltip={variableNameFromPath(variable)}
```
balazsbajorics, line 520: I removed the tooltips because they just repeated the label and made the screenshots look bad

liady (approved):

ruggi (approved):

bkrmendy (approved):

PR #5839, opened 2024-06-05
Children row in inspector props

### LINKED ISSUE #5686: Data Children: show props.children in the inspector component section
**Problem:**
it feels weird to see `name` as a data cartouche in the navigator for `<h2>{name}</h2>`, but the inspector's component section is empty. we want the component section to be able to show anything which might require data picking!

**proposed solution:**
Show props.children as a separate entry at the bottom of the component section!

Question: should we show the component section with props.children for selected elements which would otherwise not show the component section, such as div, h1, h2, etc? I personally feel like yes we should

ruggi:
**Problem:**

The children prop for components is not shown in the inspector.

**Fix:**

- Show the `children` prop in the inspector and allow changing it, via either a cartouche or number/string inputs
- Allow setting children elements when asking for a prop update
- For number and string children, use string/number inputs - or the default JSX component (or cartouche for data references) otherwise
- If the selected component does not support children, don't show it

https://github.com/concrete-utopia/utopia/assets/1081051/a66df108-2812-4c7d-9d65-aa83f8641bc6

**Manual Tests:**
I hereby swear that:

Fixes #5686


editor/src/components/editor/element-children.ts
```diff
@@ -0,0 +1,70 @@
...
+import {
+  isIntrinsicHTMLElement,
+  getJSXElementNameAsString,
+} from '../../core/shared/element-template'
+import type { PropertyControlsInfo } from '../custom-code/code-file'
+import type { EditorAction } from './action-types'
+import { deleteView, replaceElementInScope } from './actions/action-creators'
+import * as EP from '../../core/shared/element-path'
+
+export function elementSupportsChildrenFromPropertyControls(
```
ruggi, line 18: verbatim from `navigator-item-components.tsx`

editor/src/components/inspector/sections/component-section/property-control-controls.tsx
```diff
@@ -348,8 +331,32 @@ export const NumberInputPropertyControl = React.memo(
...
+        // if the target prop is children, dig for its value
+        if (propName === 'children') {
+          const element = MetadataUtils.findElementByElementPath(
+            store.editor.jsxMetadata,
+            props.elementPath,
+          )
+          if (element != null && isRight(element.element) && isJSXElement(element.element.value)) {
+            const child = element.element.value.children[0]
+            if (child.type === 'ATTRIBUTE_OTHER_JAVASCRIPT') {
+              const parsed = parseNumber(child.originalJavascript)
```
ruggi, line 346: this kinda sucks a bit, if you have better ideas please do tell!

bkrmendy (approved):

editor/src/components/inspector/sections/component-section/property-control-controls.tsx
balazsbajorics: I _think_ the info we are looking for is in the `editorState.allElementProps`! IIRC we store props.children in the spy, it's not filtered out

balazsbajorics: maybe we could even put this into `propMetadata.value` so here you don't have to special case it

balazsbajorics (approved):

balazsbajorics (approved):

Issue #5739, opened 2024-05-24
We need to fix focusability for some components that involve forwardRef


maltenuhn:
**Problem**
Sample project contained components using forwardRef, these are parsed as arbitrary Javascript right now and shouldn't be. Specific example (involving cva) here: https://github.com/concrete-utopia/hydrogen-may/blob/sehl/day-one/app/components/hydrogen/new/Layout.jsx#L145

This leads to two issues
1) they are not focusable
2) they don't show up _automatically_ in the insert menu --> has a work-around, by adding them to the component annotations manually

 [Discord Deeplink](https://discord.com/channels/647102474519576589/1160881632350318603/1243532130261274694) to discussion

**Solution**
-  to fix the React.memo, memo, React.forwardRef and forwardRef support, we can probably have a special cased fix in the parser which shouldn't take too long to implement
- accept the fact that  "the parsing of those function parameters is going to be a little spic"

PR #5846, opened 2024-06-06
Handle functions like forwardRef.

seanparsons:
**Problem:**
We need to fix focusability for some components that involve `forwardRef`.

**Fix:**
Extending the previous work from https://github.com/concrete-utopia/utopia/pull/5808, this starts by permitting multiple parameters to a `UtopiaJSXComponent` instance. That necessitates the usual changes to the parser and printer logic. But the bulk of the work was done around `createComponentRendererComponent` as that was written to support a single props parameter specifically. That now supports being called with any number of parameters and then connects those parameters to their names as parsed from the original code.

**Commit Details:**
- `createComponentRendererComponent` now creates a regular function instead of an arrow function and that function analyses the `arguments` global to determine what the actual arguments to the function are. Then applies some light heuristics to then fix the "props" to one of those arguments.
- `applyPropsParamToPassedProps` reworked to handle the multiple function arguments.
- `propsUsedByIdentifierOrAccess` now caters for multiple function parameters.
- `codeUsesProperty` and associated functions now handle multiple function parameters.
- `UtopiaJSXComponent.param` is now `UtopiaJSXComponent.params`.
- `parseCode` now parses multiple parameters for a given function.

**Manual Tests:**
I hereby swear that:

Fixes #[5739](https://github.com/concrete-utopia/utopia/issues/5739)


editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -69,12 +70,34 @@ export function createComponentRendererComponent(params: {
   filePath: string
   mutableContextRef: React.MutableRefObject<MutableUtopiaCtxRefData>
 }): ComponentRendererComponent {
-  const Component = (realPassedPropsIncludingUtopiaSpecialStuff: any) => {
+  function Component() {
+    // eslint-disable-next-line prefer-rest-params
```
balazsbajorics, line 74: why not go with a rest param on Component() ?

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
```diff
@@ -241,6 +241,96 @@ export { ToBeDefaultExported as default }`,
     )
   })
 
+  it('renders a canvas with a component wrapped in a function that has more that one parameter', () => {
+    testCanvasRender(
```
balazsbajorics, line 245: I think it's much more readable to use `testCanvasRenderInline` combined with an inline snapshot, that way we can better see which snapshot corresponds to which test (which helps catch regressions when updating snapshots) and we don't need to snapshot the spy output which has become uselessly large and a burden to maintain

editor/src/core/data-tracing/data-tracing.ts
```diff
@@ -279,63 +279,67 @@ function processJSPropertyAccessorsIntoDataTracingPath(
...
-      default:
-        assertNever(pathDrillSoFar)
-    }
-  }
-  switch (param.boundParam.type) {
-    case 'REGULAR_PARAM': {
-      // in case of a regular prop param, first we want to match the param name to the original identifier
-      if (param.boundParam.paramName !== originalIdentifier.name) {
-        return left('identifier does not match the prop name')
+  for (const param of params) {
```
balazsbajorics, line 286: hm I think this code now erroneously assumes that every parameter is a `props` param. we should instead assume that the first parameter is `props`

balazsbajorics (approved):

gbalint (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
seanparsons: Originally because I was testing things out, but I don't think there should be any reason why it can't be like that now.

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
seanparsons: I think I was going with what's already in this file, so I'll see about changing these tests.

editor/src/core/data-tracing/data-tracing.ts
seanparsons: Let me have a looky see...

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
seanparsons: Made this change in 1645db31b.

editor/src/components/canvas/ui-jsx-canvas.spec.tsx
seanparsons: Made this change in 1645db31b.

editor/src/core/data-tracing/data-tracing.ts
seanparsons: Fixed this in 1645db31b.

PR #5883, opened 2024-06-10
Multi-column data picker

balazsbajorics:
**Problem:**
The data picker is still somewhat crummy for exploring multiple deep objects in the scope

**Fix:**
First version of the multi-column data picker!

image
image
image

Array picker mode:
image

**Commit Details:**
- Added `DataSelectorColumns` component and file
- Tried to only make minimal changes in `data-selector-modal.tsx` so I only have a small conflict with @bkrmendy 
- When picking a List Source, we show non-pickable options, but without background
- Changed the CartoucheUI so it can have disabled background and disabled highlight

**Manual Tests:**
I hereby swear that:

**TODOs for follow up work**
- [ ] For some reason some objects are not traversable, they become leafs. I believe this is behavior of the existing code, needs investigation:
image

- [ ] Right now we only filter for array type when picking a List Source, we should instead always filter for the correct type
- [ ] the value preview column is a placeholder design at best
- [ ] there is at least one example in the old demo store where we incorrectly show blue cartouches instead of green:
image


seanparsons (approved):

ruggi (approved):
🧮

gbalint (approved):

bkrmendy (approved):

PR #5887, opened 2024-06-11
Override control styles for string children input

### LINKED ISSUE #5886: Text `children` in the inspector show up in blue

ruggi:
**Problem:**

String inputs for children props in the inspector have the blue foreground color even if they shouldn't, because they look as `controlled`.

**Fix:**

Override the control status/styles and make them `simple` if they are used for `children` props.

**Manual Tests:**
I hereby swear that:

Fixes #5886

### DIFF
--- a/editor/src/components/inspector/sections/component-section/property-control-controls.tsx
+++ b/editor/src/components/inspector/sections/component-section/property-control-controls.tsx
@@ -536,9 +536,7 @@ export const StringInputPropertyControl = React.memo(
 
     const controlId = `${propName}-string-input-property-control`
 
-    const isChildren = React.useMemo(() => {
-      return propName === 'children'
-    }, [propName])
+    const isChildren = propName === 'children'
 
     const value = useEditorState(
       Substores.metadata,
--- a/editor/src/components/inspector/sections/component-section/property-control-controls.tsx
+++ b/editor/src/components/inspector/sections/component-section/property-control-controls.tsx
@@ -25,7 +25,7 @@ import {
   useWrappedEmptyOrUnknownOnSubmitValue,
   wrappedEmptyOrUnknownOnSubmitValue,
 } from '../../../../uuiui'
-import type { CSSCursor } from '../../../../uuiui-deps'
+import type { CSSCursor, ControlStatus, ControlStyles } from '../../../../uuiui-deps'
 import { SliderControl, getControlStyles } from '../../../../uuiui-deps'
 import {
   normalisePathSuccessOrThrowError,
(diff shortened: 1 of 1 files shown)


bkrmendy (approved):

liady (approved):

editor/src/components/inspector/sections/component-section/property-control-controls.tsx
```diff
@@ -535,11 +535,16 @@ export const StringInputPropertyControl = React.memo(
     const { propName, propMetadata } = props
 
     const controlId = `${propName}-string-input-property-control`
+
+    const isChildren = React.useMemo(() => {
```
liady, line 539: minor - just wondering what's the performance difference b/w 
```
const x = y === 'foo'
```
and
``` const x = useMemo(() => y === 'foo', [x])
```
just because the creation of a new function and an array on every render (and one comparison happens anyway because of the deps array). 
definitely neglible though.

ruggi: IIUC the point is how frequently `y` changes, so `useMemo` becomes more relevant if the changes are frequent. So in this case is probably negligible, but my muscle memory kicks in on these things. Should I change it? (Or maybe I understood it incorrectly)

ruggi: cool, adjusted in https://github.com/concrete-utopia/utopia/pull/5887/commits/2974ffd942ebcc51456ab558a882688bcd490fd4

balazsbajorics: 1. every time the component renders, the first parameter of usememo is going to be a fresh function, which is created by the arrow function literal. that needs to be allocated etc, even if it's never called
2. the actual code _inside_ the function is cheap, so we don't save CPU cycles by not calling it, because the useMemo memo compare mechanism is probably more expensive (as it has to traverse through a variable length array which means it involves introspection which means it will be orders of magnitude slower than a direct reference equality check. that's why we have that big file of "manually generated" KeepDeepEqualityCalls, to avoid introspection)
3. the return value is a boolean, which is a primitive value (alongside number and string), primitive values are compared by value, we don't need to carefully keep a reference to the previous value. for objects the reason we use useMemo is to keep the reference to the previous object, so that `===` (which is used by React when comparing the props of a `React.memo` memoized component) can be true for previousValue === newValue. since for two booleans previousValue === newValue doesn't have the reference equality problem, we don't benefit from useMemo's ability to retain references.

balazsbajorics (approved):

PR #5900, opened 2024-06-12
Data Picker: Left Sidebar displaying Sources

balazsbajorics:
image

**Commit Details:**
- The root element of the data picker is now a FlexRow
- New `DataSelectorLeftSidebar` component in its own file
- Removed bottom scope breadcrumbs

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/inspector/sections/component-section/data-selector-columns.tsx
+++ b/editor/src/components/inspector/sections/component-section/data-selector-columns.tsx
@@ -195,6 +195,7 @@ const RowWithCartouche = React.memo((props: RowWithCartoucheProps) => {
       style={{
         backgroundColor: onActivePath ? colorTheme.bg4.value : undefined,
       }}
+      disabled={false}
     >
       <span>
         <DataPickerCartouche data={data} forcedDataSource={forcedDataSource} selected={selected} />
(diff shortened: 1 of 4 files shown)


gbalint (approved):

editor/src/components/inspector/sections/component-section/data-selector-columns.tsx
```diff
@@ -232,6 +225,16 @@ const DataSelectorFlexColumn = styled(FlexColumn)({
...
 })
 
+export const DataPickerRow = styled(FlexRow)({
+  alignSelf: 'stretch',
+  justifyContent: 'space-between',
+  fontSize: 10,
+  borderRadius: 4,
+  height: 24,
+  padding: 5,
+  cursor: 'pointer',
```
bkrmendy, line 235: we should only show a pointer when the row is actually clickable, this was a bit confusing when I was trying it

bkrmendy (approved):

editor/src/components/inspector/sections/component-section/data-selector-columns.tsx
balazsbajorics: ah good point!

PR #5928, opened 2024-06-13
Consistent border radii for navigator selection

### LINKED ISSUE #5927: navigator selection applied border radii are all over the place

ruggi:
**Problem:**

Selected rows in the navigator have inconsistent / potato border radii applied to pretty much any selected row. Moreover, the children selection also does not have any border radius.

**Fix:**

Contiguous regions of selection now share just 4 rounded corners. This is done with a hook which is used on both the regular and the condensed navigator rows, so they share all the underlying logic to figure out which element should have `borderRadius` applied or not.

Note: there's a reasonable chance there will be corner (pun intended) cases here and there, but it feels decently solid.

Screenshot 2024-06-13 at 4 08 41 PM

**Manual Tests:**
I hereby swear that:

Fixes #5927

### DIFF
--- a/editor/src/components/navigator/navigator-item/navigator-condensed-entry.tsx
+++ b/editor/src/components/navigator/navigator-item/navigator-condensed-entry.tsx
@@ -18,6 +18,7 @@ import {
   NavigatorRowClickableWrapper,
   useGetNavigatorClickActions,
 } from './navigator-item-clickable-wrapper'
+import { useNavigatorSelectionBoundsForEntry } from './use-navigator-selection-bounds-for-entry'
 
 function useEntryLabel(entry: NavigatorEntry) {
   const labelForTheElement = useEditorState(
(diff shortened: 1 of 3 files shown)


editor/src/components/navigator/navigator-item/navigator-condensed-entry.tsx
```diff
@@ -87,6 +88,14 @@ export const CondensedEntryItemWrapper = React.memo(
       )
     }, [selectedViews, props.navigatorRow])
 
+    const { isTopOfSelection, isBottomOfSelection } = useNavigatorSelectionBoundsForEntry(
+      props.navigatorRow.entries[0],
+      rowRootSelected,
+      0,
+    )
+
+    const borderRadius = 5
```
bkrmendy, line 97: I'd put this into a const, this number also shows up in other places too

bkrmendy (approved):

editor/src/components/navigator/navigator-item/navigator-condensed-entry.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/5928/commits/c3c0db05158427cb6341959753a1bb13234d1a16

liady (approved):
Nice consolidation of all of the border logic!

balazsbajorics (approved):

PR #5934, opened 2024-06-13
Move arrow indicator under mouse

### LINKED ISSUE #5868: Create a mini-arrow to show by the mouse pointer when a distant element is selected
#### Problem:
We show an arrow in the toolbar when a selected element is out of visible bounds on the canvas, but it would be more helpful to show that arrow next to your cursor

#### Solution:

- [ ] remove the arrow from the toolbar entirely
- [ ] when a selected element is out of bounds, show the little blue arrow next to the cursor, pointing towards the element. Find it here: `editor/resources/editor/icons/light/semantic/arrowOutOfBounds@2x.png`

It should look like this, turning to point in different directions accordinly
Image
Image

ruggi:
**Problem:**

The arrow indicator for elements outside of the visible area could become more useful if moved under the mouse cursor rather than being in the canvas toolbar.

**Fix:**

1. Move the indicator under the mouse cursor
2. Remove the canvas toolbar button

https://github.com/concrete-utopia/utopia/assets/1081051/b40f96ea-0038-46a6-9f2c-02054576ee92

At some point we could revisit this and reinstate the toolbar button to scroll to the element, but since this is experimental territory anyways we can evaluate pros and cons of the two approaches and maybe come up with a hybrid.

**Manual Tests:**
I hereby swear that:

Fixes #5868


bkrmendy (approved):

balazsbajorics (approved):

liady (commented):
👏

PR #5986, opened 2024-06-18
Full length UIDs

ruggi:
Part of #5654

**Problem:**

> The UIDs are too short

…and therefore prone to collisions.

**Fix:**

> I made them longer

…so they won't collide.

This PR adds support for full-length UIDs. This is done with a split approach between production and test environments, with the theoretical ability to configure the desired UID length (but realistically they'll just stay as they are - 3 chars for tests, 32 chars for prod).

When generating a consistent UID (remember! we'll want to get rid of the existing UID checks in the future, so we can parallelize the parsing):
- for test environments reuse a good chunk of the previous logic, generating 3-chars long UIDs that are either a portion of the original or an incremental alphabetical string (which is now done as an actual increment rather than doing nested loops which would also be a nightmare to parametrize).
- for production environments, use either the verbatim hash (128 bit, 32 chars) or a random UUID v4 (with stripped dashes).

**Manual Tests:**
I hereby swear that:


liady (approved):

bkrmendy (approved):

editor/src/core/shared/uid-utils.ts
```diff
@@ -863,3 +810,95 @@ export function getUtopiaID(element: JSXElementChild | ElementInstanceMetadata):
...
+    atozUID = nextTestUID(atozUID)
+  }
+  return atozUID
+}
+
+// For test UIDs, increment the input uid deterministically ('aaab'->'aaac', 'aazz'->'abaa', etc.) while keeping the original length.
+// Non alphabetical characters will be defaulted to 'a'.
+// If the UID cannot be incremented (because it's entirely made of `z`'s), an error will be thrown.
+export function nextTestUID(uid: string): string {
+  let result = uid.split('')
```
seanparsons, line 878: Possibly clearer alternative:
```suggestion
  let result = Array.from(uid)
```

seanparsons (approved):

editor/src/core/shared/uid-utils.ts
ruggi: oh nice, TIL https://github.com/concrete-utopia/utopia/pull/5986/commits/653afe48c7f1d004f6bbe5e8c40aa3172c832f78

ruggi:
Screenshot 2024-06-19 at 11 52 56 AM
I misclicked and inhadverently deleted the original comment.

This is the original
Screenshot 2024-06-19 at 11 53 20 AM

gbalint (approved):

editor/src/core/shared/uid-utils.spec.ts
```diff
@@ -86,3 +86,18 @@ describe('updateHighlightBounds', () => {
...
+    expect(nextTestUID('aaa')).toEqual('aab')
+    expect(nextTestUID('aab')).toEqual('aac')
+    expect(nextTestUID('aba')).toEqual('abb')
+    expect(nextTestUID('aaz')).toEqual('aba')
+    expect(nextTestUID('azz')).toEqual('baa')
+    expect(nextTestUID('a9a')).toEqual('aab')
+    expect(nextTestUID('aa9')).toEqual('aab')
+  })
+  it('throws if it cannot generate anymore', async () => {
+    expect(() => nextTestUID('zzz')).toThrow('Unable to generate a UID')
```
balazsbajorics, line 101: at this point I think we should roll over to 4 character UIDs, right?

ruggi: we sure could, this was meant to be 1:1 with the previous implementation which only goes as far as 3 chars, but no problem in extending it

editor/src/utils/utils.ts
```diff
@@ -281,6 +281,11 @@ export function generateUUID(): string {
   return UUID().replace(/-/g, '_')
 }
 
+// Returns a UUID v4 with dashes removed.
+export function strippedUUID(): string {
+  return UUID().replace(/-/g, '')
```
balazsbajorics, line 286: wouldn't we be better off (for debuggability) to keep the dashes?

ruggi: it would create a discrepancy I don't like with the hashed IDs which have the nice property of being overlappable to v4 (same length) but don't have dashes 🤔

balazsbajorics: hmm, wouldn't it be a useful information to know that the uid was hashed or a UUID? for debugging purposes

ruggi: yeah it's a reasonable point, no issue from my side

editor/src/core/shared/uid-utils.spec.ts
balazsbajorics: oh keeping the existing behavior is a good idea!

editor/src/utils/utils.ts
ruggi: (although at this point we could even prefix the ids)

ruggi: https://github.com/concrete-utopia/utopia/pull/5986/commits/7058bed0e78debe856e98d038d31800288beb1c3

PR #5987, opened 2024-06-18
Refactor variable matching code

bkrmendy:
## Problem
The variable matching code evolved in an "organic" way and needs trimming:
- matching for the children prop, arbitrary props and for mapped values are mushed together into one function
- we don't actaually want to order variables according to relevance anymore (we want to order them alphabetically, this is coming on a follow-up PR)

## Fix
- Revamp the matching code into a more general form, and derive specific matching functions to match for the children prop, arbitrary props and for mapped values
- adjust the call sites to provide the data for these matchers

## Manual Tests
I hereby swear that:

### DIFF
--- a/editor/src/components/inspector/sections/component-section/data-reference-cartouche.tsx
+++ b/editor/src/components/inspector/sections/component-section/data-reference-cartouche.tsx
@@ -18,14 +18,18 @@ import type { RenderedAt } from '../../../editor/store/editor-state'
 import { replaceElementInScope } from '../../../editor/actions/action-creators'
 import {
   getCartoucheDataTypeForExpression,
+  matchForChildrenProp,
+  matchForPropertyValue,
+  usePropertyControlDescriptions,
+  usePropertyValue,
   useVariablesInScopeForSelectedElement,
 } from './variables-in-scope-utils'
 import { jsxElementChildToValuePath } from './data-picker-utils'
-import { useAtom } from 'jotai'
 import type { CartoucheDataType, CartoucheHighlight, CartoucheUIProps } from './cartouche-ui'
 import { CartoucheUI } from './cartouche-ui'
 import * as PP from '../../../../core/shared/property-path'
 import { AllHtmlEntities } from 'html-entities'
+import { optionalMap } from '../../../../core/shared/optional-utils'
 
 const htmlEntities = new AllHtmlEntities()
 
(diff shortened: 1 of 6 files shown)


editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
```diff
@@ -443,49 +437,16 @@ const filterObjectPropFromVariablesInScope =
...
-      }
-    })
-
-    return result
-  }
-
 export function useVariablesInScopeForSelectedElement(
   elementPath: ElementPath | null,
-  propertyPath: PropertyPath | null,
+  matchFn: (_: Array<VariableInfo>) => Array<VariableInfo>,
```
balazsbajorics, line 442: could you make this a named type that you also use in the matchers? that way a right click -> Find All Implementations can quickly take us to all matchers, no need to jump around in the codebase

editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
```diff
@@ -322,102 +320,98 @@ function createPropertyAccessExpressionString(expressionSoFar: string, toAppend:
...
-      variable.props = orderVariablesForRelevance(
-        variable.props,
-        controlDescription,
-        currentPropertyValue,
-        targetPropertyName,
-      )
+      variable.props = matchVariables(variable.props, matchers)
+    }
+
+    for (const [matcher, idx] of matchersWithIndex) {
```
balazsbajorics, line 341: ultra nitpick but I think a standard for loop is a more idiomatic way to iterate over an array while getting an index :)

editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
bkrmendy: good idea, pushed it in https://github.com/concrete-utopia/utopia/pull/5987/commits/4abb909eba9f1ebe5ab863a2d5ac217cba897ef8

editor/src/components/inspector/sections/component-section/data-reference-cartouche.tsx
```diff
@@ -99,14 +103,29 @@ export const DataReferenceCartoucheControl = React.memo(
...
-        : assertNever(props.renderedAt)
+        : null
 
-    const variableNamesInScope = useVariablesInScopeForSelectedElement(elementPath, propertyPath)
+    const controlDescriptions = usePropertyControlDescriptions(maybePropertyPath)
+    const currentPropertyValue = usePropertyValue(elementPath, maybePropertyPath)
+
+    const matcher = React.useMemo(
+      () =>
+        props.renderedAt.type === 'element-property-path'
```
seanparsons, line 116: This looks a bit like a switch in disguise. :)

seanparsons (approved):

editor/src/components/inspector/sections/component-section/variables-in-scope-utils.ts
bkrmendy: I replaced it with `matchers.entries()` (which seems to be the recommened way to do this)

editor/src/components/inspector/sections/component-section/data-reference-cartouche.tsx
bkrmendy: I wish TS had "switch expressions" :( https://github.com/concrete-utopia/utopia/pull/5987/commits/2f41b156c6dcdf2e949a4b5cafe105ca917fd162

balazsbajorics (approved):

PR #5993, opened 2024-06-18
Feautre/cartouche context menu

### LINKED ISSUE #5972: Context menu on data cartouche
add a context menu for the data cartouche.

it should have two basic entries: edit which just opens the Data Picker, and delete in places where we support deletion of data (ie the optional props of a component)

It should have, _greyed out,_ three more items:
• Jump to data source
• Open in CMS
• Edit Value

balazsbajorics:
This PR adds a basic context menu to the Data Reference Cartouche.
image

Most of my evening was spent wrestling ContextMenuWrapper. I decided to "give up" and create a new component which uses a proper portal instead of horrible x-hacks and whatevers.

**Commit Details:**
- Renamed `ContextMenuWrapper` to `ContextMenuWrapper_DEPRECATED` also marked it as deprecated
- Added new `ContextMenuWrapper` which takes way less props, uses less wrapping divs, and uses a Portal.
- `DataCartoucheInner` uses `ContextMenuWrapper`. 
- The only actually working option is `Replace...` which just opens the data picker
- Everything else is a placeholder for future capability

Fixes #5972


bkrmendy (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -127,7 +127,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`1007`)
```
seanparsons, line 130: That seems like quite a jump.

seanparsons (approved):

Issue #5892, opened 2024-06-11
Create a combo button+dropdown control, and use it to kill the on-canvas "show / hide inspector", "show / hide code editor", and "show / hide navigator" buttons


maltenuhn:
**Problem**
We have an unconventional way to show / hide different panels. Turns out that _everyone_  by default looks in the toolbar, so we should have toggleable menu entries there.

**Implementation**
This requires a button that can invoke a "context" menu (not actually a context menu, but positioned below the button. #5633 and #5643 have prior art. Unfortunately #5643 is currently on hold, so this will require either a partial solution for that, or a stopgap one otherwise.

**Design**
- for the button, add another button to the toolbar, left of the zoom indicator. Ask @lankaukk for an icon!
- for the context menu, this needs the three-column pattern (checkmark, label, keyboard shortcut). I don't know if we have this standardised, our react-select popupMenu (which we do _not_ want to use here!) might be prior art.
- Menu entry labels:  `Navigator`, `Code Editor`, `Inspector`.  (NB: these aren't perfect, but neither are 'left menu' and 'right menu')

(checked indicators like this, plus space for keyboard shortcuts)
![Image](https://github.com/concrete-utopia/utopia/assets/2945037/fe8eac0f-0f50-427e-9e78-a56d9b92d623)

PR #6004, opened 2024-06-19
Add Manage Panels to Toolbar

benwolfram:
**Problem:**
We'd like to have a more prominent, easily-accessible way to manage panel visibility located in the toolbar

**Fix:**
As a v1, leveraging a lot of similarity to the insert mode's dropdown in the toolbar. Will iterate to use leverage a newly created button + dropdown menu component in the future.

**Manual Tests:**
I hereby swear that:

Fixes https://github.com/concrete-utopia/utopia/issues/5892


balazsbajorics:
image
the colors are broken in light mode (we will use light mode for all the demos)

balazsbajorics:
@benwolfram I pushed a quick fix fb49141fa to this branch: I re-added the TestMenu (which we need for debugging and which is why the System Test was breaking), and moved it to a new place that is actually better for us.

bkrmendy (approved):

seanparsons (approved):

PR #6011, opened 2024-06-20
Fix resizing reordered panels

### LINKED ISSUE #6005: Cannot resize the navigator (Probably trivial fix! Easy repro steps! Show Balazs how to write good code!)
**Problem**
Sometimes the navigator is no longer resizable.

and I have repro steps
1. Open utopia, new project
2. show code editor, it should show up underneath the navigator
3. move code editor to the right of the navigator, it full-sizes
4. close the code editor

Now try resizing the navigator. It fails!

ruggi:
**Problem:**

A reordered panel that is the only element of a column cannot be resized.

**Fix:**

When calculating non empty columns, take into account the visibility of the column panels too, which would otherwise return a zombie column index and mess up the resizing calculations.

https://github.com/concrete-utopia/utopia/assets/1081051/4b07009e-6849-4008-b1f8-8e3479126b2c

**Manual Tests:**
I hereby swear that:

Fixes #6005

### DIFF
--- a/editor/src/components/canvas/grid-panels-container.tsx
+++ b/editor/src/components/canvas/grid-panels-container.tsx
@@ -15,7 +15,7 @@ import {
   useResolvedGridPanels,
   wrapAroundColIndex,
 } from './grid-panels-state'
-import type { LayoutUpdate, StoredPanel } from './stored-layout'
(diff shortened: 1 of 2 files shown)


bkrmendy (approved):

gbalint (approved):

balazsbajorics (commented):
+1

PR #6023, opened 2024-06-25
fix(homepage): link to projects page instead of the editor

liady:
Change the link in our site to point to `utopia.app/projects` instead of `utopia.app/p/36ae27be-welcome-to-utopia`


ruggi:
there's also a (broken) link in `readme.md` that would be worth updating/removing

ruggi (approved):

balazsbajorics (approved):

PR #6046, opened 2024-07-09
Feat/animate grid

ruggi:
**Problem:**

When rearranging grid elements, they should animate along the X/Y axii to indicate the completed movement.
Moreover, we don't have a way to animate rendered canvas elements.

**Fix:**

This PR introduces a way to animate canvas elements programmatically, without having to inject anything inside the rendered components themselves. After doing that, the PR uses the new logic to animate the grid rearrange as a first test subject.

The goal is to be able to arbitrarily animate canvas elements using the same framer motion API we use elsewhere, explicitly. In order to avoid manipulating the rendered elements, the targeting capabilities of `useAnimate` are used here.

1. Create the animation scope and animation function with `useAnimate` 
2. Connect the scope with the `DesignPanelRoot` component
3. Store the animation function in a new `AnimationContext` context
4. `useCanvasAnimation` can now be used to target specific canvas elements and animate them, selecting them from the DOM by their `data-uid` prop
5. 🎈 

https://github.com/concrete-utopia/utopia/assets/1081051/ab70b016-a91b-47f8-95ec-bec00bdca838

### DIFF
--- a/editor/src/components/canvas/controls/grid-controls.tsx
+++ b/editor/src/components/canvas/controls/grid-controls.tsx
@@ -49,6 +49,7 @@ import { CanvasOffsetWrapper } from './canvas-offset-wrapper'
 import { useColorTheme } from '../../../uuiui'
 import { gridCellTargetId } from '../canvas-strategies/strategies/grid-helpers'
 import { useCanvasAnimation } from '../ui-jsx-canvas-renderer/animation-context'
+import { CanvasLabel } from './select-mode/controls-common'
 
 const CELL_ANIMATION_DURATION = 0.15 // seconds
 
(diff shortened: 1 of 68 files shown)


liady:
this is super slick. well done 👏

bkrmendy (approved):

liady (approved):

editor/src/core/model/element-metadata.spec.browser2.tsx
```diff
@@ -1479,7 +1479,8 @@ describe('record variable values', () => {
 })
 
 describe('specialSizeMeasurements.globalFrameWithTextContent', () => {
-  it('includes the size of a contained text node', async () => {
```
ruggi, line 1482: cc @seanparsons for future reference

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -59,7 +62,7 @@ export function gridCellCoordinates(row: number, column: number): GridCellCoordi
 }
 
 export function gridCellCoordinatesToString(coords: GridCellCoordinates): string {
-  return `${coords.row}:${coords.column}`
```
balazsbajorics, line 62: why this change?

ruggi: because there's no reason to have those serialized in two different ways (hyphen vs colon) and in my WIP code it was a meaningful difference – **however** I now realized it's actually never used anymore, so I removed it [`1ed278b` (#6046)](https://github.com/concrete-utopia/utopia/pull/6046/commits/1ed278b3e453c8bbe6f77587af683d779fdd1add)

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -412,13 +406,27 @@ export const GridControls = controlForStrategyMemoized(() => {
...
+      return null
+    }
+    return EP.parentPath(shadow.elementPath)
+  }, [shadow])
+
+  useSnapAnimation({
+    targetRootCell: targetRootCell,
+    controls: controls,
+    shadowFrame: initialShadowFrame,
+    gridPath: gridPath,
```
balazsbajorics, line 427: does useSnapAnimation depend on reference equality here? if not, then there's no point in useMemo'ing gridPath

ruggi: https://github.com/concrete-utopia/utopia/pull/6046/commits/5c156aed30404a2a62c0b5d45c80f843e347196a

editor/src/components/editor/editor-component.tsx
```diff
@@ -516,7 +519,7 @@ export const EditorComponentInner = React.memo((props: EditorProps) => {
                   overflowX: 'hidden',
                 }}
               >
-                <DesignPanelRoot />
+                <DesignPanelRoot animationScope={props.animationScope} />
```
balazsbajorics, line 522: why the prop drilling here? couldn't / shouldn't this be in a context? is props.animationScope referentially stable?

balazsbajorics (approved):
👨‍🎨

editor/src/components/editor/editor-component.tsx
ruggi: what about this https://github.com/concrete-utopia/utopia/pull/6046/commits/b60e41e87183dca88ebe0e6f513178a4fc01fbef ?

PR #6058, opened 2024-07-10
Grid resize better handles

ruggi:
**Problem:**

The resize grid strategy has some UI issues:

1. the handles are only shown when a grid cell is selected, not the grid itself
2. the handles are not positioned correctly relatively to the grid itself
3. it's hard to discern precisely the resize effects

**Fix:**

1. Enable the grid resizing if the grid is selected too
2. Position the handles centered relatively to the grid rows/columns
3. Replace the "numbered handles" with more subtle draggable circles
4. During drag, show a striped background for the affected row or column, with the measurements centered in the striped area

**Note**
The handles icons are placeholders/ideas.

| Before | After |
|--------|-----------|
| Kapture 2024-07-10 at 14 19 33 | Kapture 2024-07-10 at 14 17 58 |

### DIFF
--- a/editor/src/components/canvas/controls/grid-controls.tsx
+++ b/editor/src/components/canvas/controls/grid-controls.tsx
@@ -231,7 +231,7 @@ export const GridResizingControl = React.memo((props: GridResizingControlProps) 
           fontSize: 8,
         }}
         css={{
-          opacity: 0.5,
+          opacity: resizing ? 1 : 0.5,
           ':hover': {
             opacity: 1,
           },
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -214,49 +266,75 @@ export interface GridResizingProps {
...
+              />
+            )
+          })}
+        </div>
+      )
+    case 'FALLBACK':
+      return null
+    default:
+      assertNever(props.axisValues)
+      return null
```
bkrmendy, line 337: The `return null` is unreachable because of the `assertNever(props.axisValues)`, let's get rid of the return

ruggi: 👍 https://github.com/concrete-utopia/utopia/pull/6058/commits/4d0637ba9a4b5e64617125b5e852926779d72829

bkrmendy (approved):

Issue #6066, opened 2024-07-11
Canvas Performance: ComponentRendererComponent shouldUpdate() proeblems


balazsbajorics:
When resizing a PromiseCard component on the canvas,

![Image](https://github.com/concrete-utopia/utopia/assets/2226774/1ac18076-bd10-4fa6-9ae2-497088cf6103)

we spend 6ms on `shouldUpdate` inside the componentRendererComponent. Bottom up, it spends most of that time in `isElementInChildrenOrPropsTree` -> `React.Children.toArray`.

This should be blazing fast, because this is called in a very tight loop.

PR #6070, opened 2024-07-12
Faster shouldUpdate

gbalint:
**Problem:**
See https://github.com/concrete-utopia/utopia/issues/6066

**Fix:**
The major contributor to the slowness of `isElementInChildrenOrPropsTree ` is `React.children.toArray`.
I tried to use `React.children.forEach` or `React.children.map` but those just made things worse.
However, we know that in lot of cases `props.children` is just an array of elements or a React element itself. In these cases there is no need to run the fully general and slow toArray. Even though this is not the official use of the API (it is actually discouraged: https://react.dev/reference/react/Children#why-is-the-children-prop-not-always-an-array), it makes sense to try to get the elements without calling toArray (and we can still call `toArray` when that fails).
I made the function even more faster by switching to simple `for` loops instead of using array functions and allocating even more arrays.
With these changes I never saw >1.3ms shouldUpdate times when resizing a PromiseCard.

**Manual Tests:**
I hereby swear that:

Fixes https://github.com/concrete-utopia/utopia/issues/6066

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
@@ -408,7 +408,7 @@ function isElementInChildrenOrPropsTree(elementPath: string, props: any): boolea
   return false
 }
 
-function fastReactChildrenToArray(children: React.ReactNode | React.ReactNode[]) {
+function fastReactChildrenToArray(children: any) {
   if (React.isValidElement(children)) {
     return [children]
   }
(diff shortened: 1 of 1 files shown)


bkrmendy (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -380,20 +380,40 @@ function isRenderProp(prop: any): prop is { props: { [UTOPIA_PATH_KEY]: string }
...
+  for (let c of childrenArr) {
+    if ((c.props as any)[UTOPIA_PATH_KEY] === elementPath) {
+      return true
+    }
+
+    for (let p in props) {
+      if (React.isValidElement(p) && (p.props as any)[UTOPIA_PATH_KEY] === elementPath) {
+        return true
+      }
+    }
```
Rheeseyb, line 393: This `for` loop should be outside of the containing `for` loop, and likewise for the `for (let p in props)` below

ruggi (approved):
looks great, but see @Rheeseyb 's comment for the unnecessarily nested loops

Rheeseyb (approved):

balazsbajorics (approved):

PR #6073, opened 2024-07-12
skip calling getRemixValidPathsGenerationContext if not in remixScene

### LINKED ISSUE #6065: Canvas Performance: `useGetStoryboardRoot` / `getValidElementPaths` performance problems
When resizing a PromiseCard, it takes 7-8 ms to run `useGetStoryboardRoot` / `getValidElementPaths` which seems to spend most of its time on Remix `matchRoutes`.

Sounds like for every frame we repeat a task that should only be done whenever the routing configuration changes.

balazsbajorics:
**Problem:**
When resizing a PromiseCard, it takes 7-8 ms to run useGetStoryboardRoot / getValidElementPaths which seems to spend most of its time on Remix matchRoutes.

**Diagnosis:**
Turns out, we had this three lines of code running for _every single element visited_ during the recursive `getValidElementPathsFromElement` walk:
```    
const isRemixScene = isRemixSceneElement(element, filePath, projectContents)
    const remixPathGenerationContext = getRemixValidPathsGenerationContext(path)
    if (remixPathGenerationContext.type === 'active' && isRemixScene) {
```
`getRemixValidPathsGenerationContext` is not trivially cheap to call, and it only returns a sensible result if `path` points at a remix scene.

**Fix:**
Only call getRemixValidPathsGenerationContext if `isRemixScene` is true.

Now `useGetStoryboardRoot` only takes 0.7ms instead of 7ms.

image

**Note:** 0.7ms still feels excessive to run on every single frame, I'll continue the investigation and make a follow up PR

**Manual Tests:**
I hereby swear that:

Fixes #6065

### DIFF
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -1774,55 +1774,57 @@ function getValidElementPathsFromElement(
   let paths = includeElementInPath ? [path] : []
   if (isJSXElementLike(element)) {
     const isRemixScene = isRemixSceneElement(element, filePath, projectContents)
-    const remixPathGenerationContext = getRemixValidPathsGenerationContext(path)
-    if (remixPathGenerationContext.type === 'active' && isRemixScene) {
-      function makeValidPathsFromModule(routeModulePath: string, parentPathInner: ElementPath) {
-        const file = getProjectFileByFilePath(projectContents, routeModulePath)
-        if (file == null || file.type !== 'TEXT_FILE') {
-          return
+    if (isRemixScene) {
+      const remixPathGenerationContext = getRemixValidPathsGenerationContext(path)
+      if (remixPathGenerationContext.type === 'active') {
+        function makeValidPathsFromModule(routeModulePath: string, parentPathInner: ElementPath) {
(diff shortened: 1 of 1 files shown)


Rheeseyb (approved):

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1774,55 +1774,57 @@ function getValidElementPathsFromElement(
...
-    const remixPathGenerationContext = getRemixValidPathsGenerationContext(path)
-    if (remixPathGenerationContext.type === 'active' && isRemixScene) {
-      function makeValidPathsFromModule(routeModulePath: string, parentPathInner: ElementPath) {
-        const file = getProjectFileByFilePath(projectContents, routeModulePath)
-        if (file == null || file.type !== 'TEXT_FILE') {
-          return
-        }
+    if (isRemixScene) {
+      const remixPathGenerationContext = getRemixValidPathsGenerationContext(path)
+      if (remixPathGenerationContext.type === 'active') {
```
balazsbajorics, line 1779: Hide Whitespace recommended

ruggi (approved):

seanparsons (approved):

PR #6075, opened 2024-07-12
More shouldUpdate optimizations

gbalint:
**Problem:**
Follow-up to https://github.com/concrete-utopia/utopia/pull/6070

**Fix:**
I introduced two more optimizations:

1. The bigger one is that shouldUpdate called `isElementInChildrenOrPropsTree` separately for each of the element paths in `ElementsToRerenderGLOBAL`, even though all these runs traversed the same subtree, because they were called with the same `props` parameter. I updated `isElementInChildrenOrPropsTree` to work on an array of element paths, and check for all of them in a single execution of the function. I was a bit worried that the added complexity of this function  may slow down the most typical single element use case, but I did not find the effect measurable.

This really heavily speeds up multiselection interactions, basically a multiselection with n elements is nearly n times faster with this code. 

3. I added an early return to not do anything at all when the props.children is null or a string (happens frequently).

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
+++ b/editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
@@ -102,15 +102,29 @@ export function createComponentRendererComponent(params: {
     const instancePath: ElementPath | null = tryToGetInstancePath(instancePathAny, pathsString)
 
     function shouldUpdate() {
-      return (
-        ElementsToRerenderGLOBAL.current === 'rerender-all-elements' ||
-        ElementsToRerenderGLOBAL.current.some((er) => {
-          return (
-            (instancePath != null &&
-              (EP.pathsEqual(er, instancePath) || EP.isParentComponentOf(instancePath, er))) ||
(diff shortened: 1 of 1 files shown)


balazsbajorics (approved):

bkrmendy (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-component-renderer.tsx
```diff
@@ -379,32 +386,36 @@ function isRenderProp(prop: any): prop is { props: { [UTOPIA_PATH_KEY]: string }
   )
 }
 
-function isElementInChildrenOrPropsTree(elementPath: string, props: any): boolean {
+function areElementsInChildrenOrPropsTree(elementPaths: Array<string>, props: any): boolean {
+  if (props.children == null || typeof props.children === 'string') {
+    return false
+  }
```
Rheeseyb, line 392: This condition is wrong because it is only ruling out the children. The element may still exist in the props

Rheeseyb (changes requested):

Rheeseyb (approved):

PR #6080, opened 2024-07-15
Support area names in grid templates

### LINKED ISSUE #6079: Grid template parsing should support labels

ruggi:
**Problem:**

Parsed CSS grid templates should support area names.

**Fix:**

Parse area names along with the CSS number for grid templates. The label is stored in the number itself, so we can reuse it later (e.g. for targeting via other style props).

To make it more apparent, I added the area name under the resize handles, but it's definitely not a requirement (although I think it could/should be shown somewhere!).

https://github.com/user-attachments/assets/5aba975d-3b24-48be-8d69-c19f774bb351

Fixes #6079

### DIFF
--- a/editor/src/components/inspector/common/css-utils.ts
+++ b/editor/src/components/inspector/common/css-utils.ts
@@ -589,18 +589,18 @@ const GridCSSNumberUnits: Array<GridCSSNumberUnit> = [...LengthUnits, ...Resolut
 export interface GridCSSNumber {
   value: number
   unit: GridCSSNumberUnit | null
-  label: string | null
+  areaName: string | null
 }
 
 export function gridCSSNumber(
   value: number,
   unit: GridCSSNumberUnit | null,
-  label: string | null,
+  areaName: string | null,
 ): GridCSSNumber {
   return {
-    value,
-    unit,
-    label,
+    value: value,
+    unit: unit,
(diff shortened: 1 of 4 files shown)


bkrmendy (approved):

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -762,7 +768,9 @@ export function printArrayCSSNumber(array: Array<GridCSSNumber>): string {
   return array
     .map((dimension) => {
       const printed = printCSSNumber(dimension, null)
-      return typeof printed === 'string' ? printed : `${printed}`
+      const areaName = dimension.areaName != null ? `[${dimension.areaName}] ` : ''
+      const value = typeof printed === 'string' ? printed : `${printed}`
```
liady, line 772: minor - why do we need to use the string template here? can we just pass `printed `to
`${areaName}${printed}` next line?

liady (approved):

editor/src/components/inspector/common/css-utils.ts
ruggi: sounds good https://github.com/concrete-utopia/utopia/pull/6080/commits/f5573860c5356073d009f153df9ea3a60b7b7b9a

PR #6087, opened 2024-07-16
grid-specific hug strategy

bkrmendy:
## Problem
The basic hug contents strategy doesn't really make sense for grids.

## Fix
Create a hug contents strategy tailored to grids. When setting a grid to hug along an axis, if the grid doesn't use any `fr` units in `gridTemplate{Row | Column}` along that axis, the width/height prop is removed from that axis, so the grid is sized by its template rows or columns. If there's an `fr` unit  in`gridTemplate{Row | Column}`, the strategy is not applicable because the `fr` only works if the grid is sized


editor/src/components/inspector/inspector-common.ts
```diff
@@ -272,7 +272,12 @@ export const hugContentsApplicableForContainer = (
           MetadataUtils.isPositionAbsolute(element)
         ),
     ).length > 0
+
+  const isElementGrid = MetadataUtils.isGridLayoutedContainer(
```
ruggi, line 276: I would rename this to `isGrid` so it does not stutter with `isGridElement` & co

ruggi (approved):

editor/src/components/inspector/inspector-strategies/hug-contents-basic-strategy.ts
```diff
@@ -1,5 +1,6 @@
 import type { ElementPathTrees } from '../../../core/shared/element-path-tree'
```
balazsbajorics, line 1: minor nitpick, but could you rename this file?

editor/src/components/inspector/inspector-common.ts
```diff
@@ -255,12 +255,12 @@ export function detectAreElementsFlexContainers(
...
   flexDirection.startsWith('column')
 
-export const hugContentsApplicableForContainer = (
+export const basicHugContentsApplicableForContainer = (
   metadata: ElementInstanceMetadataMap,
   pathTrees: ElementPathTrees,
   elementPath: ElementPath,
 ): boolean => {
-  return (
+  const hasNonFixStickyOrAbsolute =
```
balazsbajorics, line 263: isNonFixStickOrAbsolute? hasNoFixStickyOrAbsolutePosition? positionNotFixStickyOrAbsolute?

balazsbajorics (approved):

editor/src/components/inspector/inspector-common.ts
bkrmendy: [naming](https://github.com/concrete-utopia/utopia/pull/6087/commits/d1bb61b01ac64ba66620fed0f65dfd95510ab77e)

editor/src/components/inspector/inspector-strategies/hug-contents-basic-strategy.ts
bkrmendy: done in https://github.com/concrete-utopia/utopia/pull/6087/commits/2731834981717d7df9403fc9f979a960cadf0043 and https://github.com/concrete-utopia/utopia/pull/6087/commits/917e7767d63d180f9d1dba0b3c5f21fa694109fd

PR #6090, opened 2024-07-17
Support `repeat()` in grid templates

### LINKED ISSUE #6089: support `repeat()` functions in grid templates

ruggi:
**Problem:**

Grids with `repeat()` in their template are correctly displayed, but cannot be resized because the dimensions are `FALLBACK`.

**Fix:**

Expand the `repeat()` functions inside the templates when parsing them, so they are calculated correctly.

https://github.com/user-attachments/assets/7744262c-f701-4685-a357-6ddaf9685bca

Fixes #6089

### DIFF
--- a/editor/src/components/inspector/common/css-utils.ts
+++ b/editor/src/components/inspector/common/css-utils.ts
@@ -953,16 +953,23 @@ export function expandRepeatFunctions(str: string): string {
 
 const reGridAreaNameBrackets = /^\[.+\]$/
 
-export function tokenizeGridTemplate(str: string): string[] {
-  let tokens: string[] = []
-  let parts =
-    // 1. expand any `repeat()` functions
-    expandRepeatFunctions(str)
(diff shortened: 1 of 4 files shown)


editor/src/components/inspector/common/css-utils.ts
```diff
@@ -943,11 +943,35 @@ export function parseGridRange(
   }
 }
 
+// this regex matches 'repeat ( [unit] , [value] )', capturing groups around the unit (1) and the value (2)
+const reRepeatFunction = /repeat\s*\(\s*(\d+)\s*,\s*([^,\)]+)\s*\)/
```
liady, line 947: just a note - maybe add a TODO to support non-units repeaters (i.e `auto-fit`, `auto-fill`, etc)?

ruggi: makes sense, done ✅

bkrmendy (approved):

seanparsons (approved):

PR #6091, opened 2024-07-17
fix(editor): clear execution cache on canvas reset

liady:
**Problem:**
Due to caching of execution scope, when resetting the canvas some files are not being `import`ed - resulting in missing assets:

<video src="https://github.com/user-attachments/assets/e0541178-2d38-4174-9996-14af6f9b54d8"></video>

**Fix:**
Clear the execution scope cache before resetting the canvas. This fixes the issue.
This PR also lifts the invalidation check higher up the component tree, since we're changing the key of the canvas wrapper, which effectively means that all the `useRef`s are clearing up. Now the check happens higher than the key change.

<video src="https://github.com/user-attachments/assets/2ebd7f8e-f59a-4a17-af9f-6e94b3a6fabb"></video>

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/canvas-component-entry.tsx
```diff
@@ -93,26 +94,74 @@ const CanvasComponentEntryInner = React.memo((props: CanvasComponentEntryProps)
...
+            onRuntimeError={onRuntimeError}
+            localClearRuntimeErrors={localClearRuntimeErrors}
+          />
         )}
       </div>
     </>
   )
 })
 
+function CanvasInner({
```
liady, line 108: extracting it to a component so that we can be sure that `canvasProps` are defined, since the `useInvalidatedCanvasRemount` hook also assigns the current count values to the `ref`s, and when they are undefined it creates a false positive on the next render (which breaks a lot of tests)

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -454,7 +467,10 @@ export const UiJsxCanvas = React.memo<UiJsxCanvasPropsWithErrorCallback>((props)
     )
 
     // IMPORTANT this assumes createExecutionScope ran and did a full walk of the transitive imports!!
-    if (shouldRerenderRef.current) {
+    const isRemounted =
+      props.invalidatedCanvasData.mountCountInvalidated ||
+      props.invalidatedCanvasData.domWalkerInvalidated
```
balazsbajorics, line 472: if we only clear the cache on `invalidatedCanvasData.mountCountInvalidated`, isn't it a problem that we `unimport...` on `domWalkerInvalidated` too?

liady: so when the `proejctContents` are changed, the cache is automatically invalidated (no need to clear it)

editor/src/components/canvas/canvas-component-entry.tsx
ruggi: should this be put inside a `React.memo`? honest question, I don't know if it would create issues with the refs

liady: Since it's an extracted inline component, I think that now without a `memo` it works exactly as the same before (meaning it will render at the same times when the online component was rendered, not more or less).
There might be a performance gain to adding `memo`, but this needs to be checked, and is unrelated to the extraction (again - since now it works exactly as before, and the parent component is `memo`ed)

balazsbajorics (approved):

ruggi (approved):

gbalint (approved):

PR #6093, opened 2024-07-17
Drag grid elements that don't have grid props set

bkrmendy:
## Problem
When a grid child without grid props is dragged, it jumps back to the top left of the grid no matter where it's dragged

## Fix
If a grid child doesn't have grid props set, we take it as if was set to occupy the grid column/row under the mouse pointer at start of the interaction (since that's where it's positioned by the browser)

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/grid-rearrange-move-strategy.spec.browser2.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/grid-rearrange-move-strategy.spec.browser2.tsx
@@ -9,8 +9,11 @@ import { gridCellTargetId } from './grid-helpers'
 
 describe('grid rearrange move strategy', () => {
   it('can rearrange elements on a grid', async () => {
+    const testId = 'aaa'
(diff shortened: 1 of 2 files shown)


ruggi (approved):
nice

editor/src/components/canvas/canvas-strategies/strategies/grid-rearrange-move-strategy.spec.browser2.tsx
```diff
@@ -125,16 +150,13 @@ export var storyboard = (
           data-uid='aaa'
           data-testid='aaa'
         />
-        <Placeholder
```
balazsbajorics, line 128: was the fact that this is a Placeholder causing a test issue, or is this just an aesthetic fix?

bkrmendy: I replaced it with a div because Placeholders don't pass down the `data-testid` attribute

balazsbajorics (approved):

PR #6097, opened 2024-07-18
Convert to grid

bkrmendy:
## Description
This PR adds a basic convert-to-grid strategy, makes the flex <-> grid toggle work again, and turns the layout system + button into a menu-on-button where on can pick between grid or flex

### Details/breakdown
- add the convert-to-grid strategy (and a remove grid strategy)
- make the flex <-> grid toggle work again
- add the menu-on-button to choose between grid and flex
- add a shortcut that turns a layout into a grid
- expose a prop on the radix-based dropdown to align the dropdown element relative to the opener
- update tests


editor/src/components/inspector/inspector-strategies/inspector-strategies.ts
```diff
@@ -157,6 +158,20 @@ export const addFlexLayoutStrategies = (
...
 ]
 
+export const addGridLayoutStrategies = (
+  metadata: ElementInstanceMetadataMap,
+  elementPaths: ElementPath[],
+  elementPathTree: ElementPathTrees,
+  allElementProps: AllElementProps,
+): Array<InspectorStrategy> => [
+  {
+    name: 'Add flex layout',
```
ruggi, line 168:
```suggestion
    name: 'Add grid layout',
```

editor/src/components/common/shared-strategies/convert-to-grid-strategy.ts
```diff
@@ -0,0 +1,112 @@
...
+      allElementProps,
+    )
+    const childFrames: Array<CanvasFrameAndTarget> = childrenPaths.map((child) => ({
+      target: child,
+      frame: MetadataUtils.getFrameOrZeroRectInCanvasCoords(child, metadata),
+    }))
+
+    const { gap, numberOfColumns, numberOfRows } = guessMatchingGridSetup(childFrames)
+
+    return [
```
ruggi, line 93: this looks great! although I would also go and remove any `position`-related props from the children

bkrmendy: added it in [remove positioning from children](https://github.com/concrete-utopia/utopia/pull/6097/commits/2968526c20ea884c59f046a3efd8276b2e1d425e)

editor/src/components/common/shared-strategies/convert-to-grid-strategy.ts
```diff
@@ -0,0 +1,118 @@
...
+  const averageGap =
+    gaps.length === 0 ? 0 : Math.floor(gaps.reduce((a, b) => a + b, 0) / gaps.length)
+
+  return {
+    nChildren: childrenAlongAxis,
+    averageGap: averageGap,
+  }
+}
+
+function guessMatchingGridSetup(children: Array<CanvasFrameAndTarget>): {
```
ruggi, line 50: I think this should also have a logic branch so that it makes sure that the combined number of resulting cells is `>=` than the number of children. For example if you have N children resulting from duplication, all having the exact same coordinates, the resulting grid template would be incorrect (e.g. 1x1 instead of 1xN)

bkrmendy: Discussed on a side channel, I'll implement this on a separate PR because that way we'll have a holistic overview of the convert-to-grid strategy

editor/src/components/common/shared-strategies/convert-to-grid-strategy.ts
ruggi: keeping this unresolved so we can find it quickly later

ruggi (approved):

editor/src/components/common/shared-strategies/convert-to-grid-strategy.ts
ruggi: keeping this unresolved so we can find it quickly later

editor/src/components/common/shared-strategies/convert-strategies-common.ts
```diff
@@ -0,0 +1,30 @@
...
+import { MetadataUtils } from '../../../core/model/element-metadata-utils'
+import type { ElementPathTrees } from '../../../core/shared/element-path-tree'
+import type { ElementInstanceMetadataMap } from '../../../core/shared/element-template'
+import {
+  isElementNonDOMElement,
+  replaceNonDOMElementPathsWithTheirChildrenRecursive,
+} from '../../canvas/canvas-strategies/strategies/fragment-like-helpers'
+import type { AllElementProps } from '../../editor/store/editor-state'
+
+export type FlexDirectionRowColumn = 'row' | 'column' // a limited subset as we won't never guess row-reverse or column-reverse
```
Rheeseyb, line 11:
```suggestion
export type FlexDirectionRowColumn = 'row' | 'column' // a limited subset as we never guess row-reverse or column-reverse
```

liady (approved):

balazsbajorics (approved):

editor/src/components/common/shared-strategies/convert-to-grid-strategy.ts
```diff
@@ -0,0 +1,118 @@
...
+  gap: number
+  numberOfColumns: number
+  numberOfRows: number
+} {
+  const horizontalData = guessLayoutInfoAlongAxis(
+    children,
+    (a, b) => a.frame.x - b.frame.x,
```
Rheeseyb, line 57: This works, and I guess is fine, but it might have been clearer to pass in an object with named keys for the various functions. Not worth changing in the PR now, just something to keep in mind for the future.

editor/src/components/inspector/flex-section.tsx
```diff
@@ -1,64 +1,153 @@
...
-export const areElementsFlexContainersSelector = createSelector(
+export const layoutSystemSelector = createSelector(
   metadataSelector,
   selectedViewsSelector,
-  detectAreElementsFlexContainers,
+  (metadata, selectedViews) => {
+    const detectedLayoutSystems = selectedViews.map(
+      (path) =>
+        MetadataUtils.findElementByElementPath(metadata, path)?.specialSizeMeasurements
+          .layoutSystemForChildren ?? null,
```
Rheeseyb, line 35: Should this return `'none'` rather than `null`? I'm genuinely unsure

Rheeseyb (approved):

editor/src/components/inspector/flex-section.tsx
bkrmendy: Completely fair question, I added the `?? null` to remove the `| undefined` from the type of `specialSizeMeasurements.layoutSystemForChildren` to make it a less noisy type. I didn't add `'none'` to make the type of `detectedLayoutSystems` similar to what we'd get if we checked `specialSizeMeasurements.layoutSystemForChildren` for a single element

PR #6099, opened 2024-07-18
Fix/use editor state inside canvas

balazsbajorics:
**Problem:**
When anything in the projectContents changes, the DerivedState.remixData gets recreated. (This should be improved to only recreate the remixData if we actually change the file-based routing)

The `UtopiaRemixRootComponent` uses **illegal** (🙂) useEditorState hooks to access `DerivedState.remixData`.

Since `DerivedState.remixData` is referentially new, the useEditorState hooks will trigger a re-render of every UtopiaRemixRootComponent. 

The UtopiaRemixRootComponent rerenders then basically trigger an almost full rerender of the canvas before the "real" canvas rerender step. 

on master:
image
with Sean's createExecutionScope cache fix (currently disabled and blocked by #6091):
image

On the image you can see that in the sample store's `/` route this takes 15ms every single frame.

**Fix:**
**20ms -> 10ms -> 5ms**
image
When the canvas is in selective rerender mode, DO NOT let `UtopiaRemixRootComponent` rerender itself. Not updating UtopiaRemixRootComponent during the selective mode should be safe, since the continuous canvas interactions never cause the file based routing or loaders etc to update.

- Wrapping the canvas in a EditorStateContext provider with the value forced to `null` which means we will throw an error in the future every time new code accidentally uses the useEditorState hooks (which are _not_ compatible with selective re-rendering). Hopefully this means selective re-rendering will be less brittle in the future, but in a follow up PR I will write tests for this too
- Added new hook `useCanvasState` which is almost the same as `useEditorState` except it takes an extra callback `allowRerender`. If AllowRerender is false, then no matter what changes in the editor state, we will passively reject the component update and the hook will return the _old_ state value.
- Using the new hook in `utopia-remix-root-component`
- Profit!

**Notes:**
This PR depends on #6091 because the selective re-rendering of the remix root component causes the same css issue as Sean's original #6076

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -558,19 +558,22 @@ export const UiJsxCanvas = React.memo<UiJsxCanvasPropsWithErrorCallback>((props)
...
-            validRootPaths={rootValidPathsArray}
-            canvasRootElementElementPath={storyboardRootElementPath}
-          >
-            <SceneLevelUtopiaCtxAtom.Provider value={sceneLevelUtopiaContextValue}>
-              {StoryboardRoot}
-            </SceneLevelUtopiaCtxAtom.Provider>
-          </CanvasContainer>
-        </UtopiaProjectCtxAtom.Provider>
-      </RerenderUtopiaCtxAtom.Provider>
+      {/* deliberately breaking useEditorState and useRefEditorState to enforce the usage of useCanvasState */}
```
ruggi, line 561: 👍 👍

ruggi (approved):

bkrmendy (approved):

gbalint (approved):

PR #6102, opened 2024-07-18
Support `rowGap`/`columnGap` in grid controls

### LINKED ISSUE #6101: support rowGap/columnGap in grid canvas controls

ruggi:
**Problem:**

Grid controls don't support/show correctly grids with `rowGap` and `columnGap` properties.
Specifically, they both show incorrect grid lines as well as not having correct drag targets for the mouse-based interactions.

**Fix:**

1. Include those two props in the `specialSizeMeasurements`
2. Carry them over when displaying the grid shadow and targets

| Before | After |
|--------|-----------|
| Screenshot 2024-07-18 at 18 02 19 | Screenshot 2024-07-18 at 18 02 16 |

Fixes #6101

### DIFF
--- a/editor/src/core/shared/element-template.ts
+++ b/editor/src/core/shared/element-template.ts
@@ -2720,6 +2720,8 @@ export interface SpecialSizeMeasurements {
   elementGridProperties: GridElementProperties
   containerGridPropertiesFromProps: GridContainerProperties
   elementGridPropertiesFromProps: GridElementProperties
+  rowGap: number | null
+  columnGap: number | null
 }
 
 export function specialSizeMeasurements(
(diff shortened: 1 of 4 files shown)


editor/src/core/shared/element-template.ts
```diff
@@ -2720,6 +2720,8 @@ export interface SpecialSizeMeasurements {
   elementGridProperties: GridElementProperties
   containerGridPropertiesFromProps: GridContainerProperties
   elementGridPropertiesFromProps: GridElementProperties
+  rowGap: number | null
```
balazsbajorics, line 2723: shouldn't these be part of GridContainerProperties?

ruggi: Great question and I struggled with it a bit too. I ended up going this route because it felt to me that those are not grid measurements but akin to the other `[...]gap[...]` ones (gridGap, gap). Also, they are "raw values" while the `Grid[...]Properties` are somewhat derived.

(This also would mean the `Grid[...]Properties` could be renamed to `Grid[...]SomethingElse` where `SomethingElse` could be `measurements`, `layout`, …)

balazsbajorics: ah I just searched and apparently this applies to flex too! okay then

balazsbajorics (approved):

PR #6106, opened 2024-07-19
Fix canvas toolbar rerenders

balazsbajorics:
**Problem:**
The CanvasToolbar and some of its children components would unnecessarily rerender during canvas interactions.

**Fix:**
Fixing some useState hooks that caused the problems. 

- no longer get the entire currentStrategyState just so we can derive an icon name
- CanvasStrategyPicker no longer gets the full reference to the array of strategies, only their name and ID (which makes equality checks easier)
- similarly, MoveReorderReparentIndicator only gets the string to display instead of the full state


Rheeseyb (approved):

liady (approved):
very nice 
🫴 🍏

gbalint (approved):

PR #6107, opened 2024-07-19
Fix/remove navigator derived state

balazsbajorics:
**Problem**
Currently, we recalculate `getNavigatorTargets` twice per frame, even during canvas interactions where we know we won't use the results for rendering the frame.

**Spike discovery**
What if we remove getNavigatorTargets from DerivedState and instead turn it into a memoized selector?

**Details**
* create selector `navigatorTargetsSelector`, and try to use it from every component that accessed derivedState.navigatorTargets or derivedState.navigatorRows or derivedState.visibleNavigatorTargets
* create helper `getNavigatorTargetsFromEditorState` for all the Action handler functions so they can keep using navigator targets without it being in the DerivedState
* fix _a ton_ of test code so it uses `getNavigatorTargetsFromEditorState`. this is the majority of the diff

**Evaluation**
I think the approach is very promising! Especially for frames during canvas interactions where we want to disable most of the editor chrome, the selectors give us a very clean way to _completely avoid_ calculations that are not directly necessary for updating the canvas. I would like to continue to kill the rest of derived state in the same vein. I believe I am over the halfway point, because navigatorTargets was the majority of uses in our codebase.


editor/src/components/editor/store/editor-state.ts
```diff
@@ -2422,9 +2421,6 @@ export function isSlotNavigatorEntry(entry: NavigatorEntry): entry is SlotNaviga
 }
 
 export interface DerivedState {
-  navigatorRows: Array<NavigatorRow>
```
balazsbajorics, line 2425: this is one of the important changes

editor/src/components/navigator/navigator-utils.ts
```diff
@@ -793,7 +805,57 @@ function getNavigatorRowsForTree(
...
+export function useGetNavigatorTargets(): GetNavigatorTargetsResults {
+  return useEditorState(
+    // TODO try not to use fullStore here
+    Substores.fullStore,
+    navigatorTargetsSelector,
+    'useGetNavigatorTargets',
+  )
+}
+
+export const navigatorTargetsSelector = createSelector(
```
balazsbajorics, line 823: the most important part of the PR is this new selector, and the helper functions using it

editor/src/components/inspector/sections/layout-section/conditional-section.tsx
```diff
@@ -80,8 +80,15 @@ const branchNavigatorEntriesSelector = createCachedSelector(
   (store: ProjectContentAndMetadataSubstate) => store.editor.jsxMetadata,
   (store: ProjectContentAndMetadataSubstate) => store.editor.elementPathTree,
   (store: ProjectContentAndMetadataSubstate) => store.editor.projectContents,
+  navigatorTargetsSelector,
   (_store: ProjectContentAndMetadataSubstate, paths: ElementPath[]) => paths,
-  (jsxMetadata, elementPathTree, projectContents, paths): BranchNavigatorEntries | null => {
+  (
+    jsxMetadata,
+    elementPathTree,
+    projectContents,
```
ruggi, line 88: these two are unused, and so it's possible to remove lines 81-82 as well

ruggi (approved):

gbalint (approved):

seanparsons (approved):

PR #6114, opened 2024-07-23
Handle fragment sizing including ancestors.

### LINKED ISSUE #5856: the Layout component in hydrogen-may shows up with the wrong size, until it is focused, then its displayed size changes
This is a problem in how we measure the size of unfocused and focused components which wrap their children in a thick layer of wrapper elements.

It is particularly problematic in the sample project, because it _always_ shows up, and looks clearly incorrect because the outline visibly does not line up with what's rendered on the screen.

seanparsons:
**Problem:**
The Layout component in our sample project shows up with the wrong size, until it is focused, then its displayed size changes.

**Cause:**
This mainly originates because when capturing dimensions of elements we only handled DOM walking for immediate children (by element path). As a result for the canvas controls specifically they would not see any dimensions for children of an element which are not immediate children by element path. This is caused by fragments, arbitrary code or even special components like `Suspense` as we parse them out and give them a UID but they don't materialise into real DOM elements.

**Fix:**
The fixes for handling this effectively run in this order:
- Collect the metadata for DOM elements that are children of other elements being collected but that haven't had their parent by element path be collected. This means that as well as collecting `a/b` and `a/b/c` we'd also collect `a/b/somefragment/arbitrarycode~~~1/d`.
- We build up some faked metadata entries by backfilling from the lowest most entries that don't have parents. During this we also correct the frames of their eventual ancestors which do exist.
- For the controls we now check if the target is a component instance and if the paths were replaced, if that is the case then prevent the canvas strategies from applying.

**Commit Details:**
- Slightly reworked `replaceFragmentLikePathsWithTheirChildrenRecursiveInner` to include the unfurled components because the elements attached to those paths can affect the bounds of the parent/ancestor.
- Implemented `backfillDomMetadata` which tweaks the frames of existing elements by working its way up from descendants and puts entries in a separate `reconstructedMetadata` for those cases when bounds are calculated by looking at the (in effect) missing metadata entries.
- `collectMetadata` now includes DOM elements that are child nodes of the DOM elements but haven't seen their parent (by element path) be visited by the DOM walker. This means that elements are collected when things like fragments and arbitrary code introduce gaps in the hierarchy by element path.
- Added `getChildrenOrderedMultiMetadata` as a way to avoid combining multiple values of `ElementInstanceMetadataMap`.
- Removed a parameter from `getChildrenPathsOrdered`.
- `fillSpyOnlyMetadata` changed to work with the reconstructed metadata values as well as the regular DOM metadata.
- Tweaked `buildTree` to have it support varargs of `ElementInstanceMetadataMap`.
- Improved the performance of `isDescendantOf` by exploiting the string representation of the paths.
- Changed `getElementFragmentLikeType` to use the reconstructed metadata as this function was used extensively to identify the children of an element to calculate its bounding box for canvas controls.
- Fixed the setting of the `componentInstance` property.

**Manual Tests:**
I hereby swear that:

Fixes #5856


editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-strategy.ts
```diff
@@ -44,6 +44,16 @@ export function keyboardAbsoluteMoveStrategy(
 ): CanvasStrategy | null {
   const { pathsWereReplaced, paths: selectedElements } =
     retargetStrategyToChildrenOfFragmentLikeElements(canvasState)
+
+  if (
+    pathsWereReplaced &&
+    selectedElements.some((originalTarget) =>
+      MetadataUtils.isComponentInstanceFromMetadata(canvasState.startingMetadata, originalTarget),
+    )
+  ) {
```
balazsbajorics, line 53: is there any strategy left that uses `retargetStrategyToChildrenOfFragmentLikeElements` and _doesn't_ bail out on isComponentInstanceFromMetadata? because if not, then I think we should move the component check inside!

not sure about the fallout, but it very much looks like `getElementFragmentLikeType` can be changed so it only returns `'sizeless-div'` as a result if `isComponentInstanceFromMetadata` is false.

editor/src/components/canvas/dom-walker.ts
```diff
@@ -729,14 +876,21 @@ function collectMetadata(
   cachedPaths: Array<ElementPath>
   collectedPaths: Array<ElementPath>
 } {
-  if (pathsForElement.length === 0) {
+  // Determine if the parent has been visited so far.
```
balazsbajorics, line 879: please expand this to contain the 'why?'!

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -724,12 +727,28 @@ export const MetadataUtils = {
     }
     return result
   },
+  getChildrenOrderedMultiMetadata(
```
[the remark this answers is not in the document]
Rheeseyb: Yeah I've just expanded all of the diffs and it looks like you're right

editor/src/components/canvas/canvas-strategies/strategies/keyboard-absolute-move-strategy.ts
seanparsons: There's other utility functions which also invoke `retargetStrategyToChildrenOfFragmentLikeElements` which is why I left the logic like this.

I think if we want to refactor `getElementFragmentLikeType` it should be done in a separate piece of work, so that this doesn't become a forever PR.

editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
```diff
@@ -38,6 +38,15 @@ export function absoluteMoveStrategy(
...
     retargetStrategyToChildrenOfFragmentLikeElements(canvasState)
 
+  if (
+    pathsWereReplaced &&
+    originalTargets.some((originalTarget) =>
+      MetadataUtils.isComponentInstanceFromMetadata(canvasState.startingMetadata, originalTarget),
+    )
+  ) {
+    return null
+  }
```
Rheeseyb, line 48: Does this case happen? I'm pretty sure that if it does then this is wrong, as we just shouldn't be replacing the paths at all if we're operating on a component instance. Instead the strategy should be applied to the component instance itself (rather than returning `null`)

editor/src/components/canvas/dom-walker.ts
```diff
@@ -428,6 +436,135 @@ function runSelectiveDomWalker(
...
+        const pathStringToFill = EP.toString(pathToFill)
+        if (!missingPaths.has(pathStringToFill) && !(pathStringToFill in metadata)) {
+          missingPaths.add(pathStringToFill)
+          const parentPath = EP.parentPath(pathToFill)
+          if (parentPath != null) {
+            fillPath(parentPath)
+          }
+        }
+      }
+    }
```
Rheeseyb, line 477: These functions should live outside of the `for` loop

editor/src/components/canvas/dom-walker.ts
seanparsons: Fixed in 2351319d3.

editor/src/core/model/element-metadata-utils.ts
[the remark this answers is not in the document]
seanparsons: Fixed in 2351319d3.

editor/src/components/canvas/canvas-strategies/strategies/absolute-move-strategy.tsx
balazsbajorics: we should change retargetStrategyToChildrenOfFragmentLikeElements and specifically `getElementFragmentLikeType` to not return `sizeless-div` for components, and these checks then can be removed. I agree that we shouldn't do that as part of this PR, it can either be a follow-up, or a pre-PR.

Rheeseyb: > I agree that we shouldn't do that as part of this PR, it can either be a follow-up, or a pre-PR.

If we do that then this new check above shouldn't exist in any of the strategies as we'll have to immediately remove it anyway.

editor/src/components/canvas/dom-walker.ts
seanparsons: Fixed in aeae4fd8f.

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser2.tsx
```diff
@@ -333,6 +333,24 @@ describe('Spy Wrapper Template Path Tests', () => {
         "storyboard/scene-1/app:other-app-root/other-inner-div/other-card-instance": Object {
           "name": "div",
         },
+        "storyboard/scene-1/app:other-app-root/other-inner-div/other-card-instance:other-button-instance/other-expr/other-hi-element~~~1:c33/hi-element-fragment-child-1": Object {
```
Rheeseyb, line 336: Are these test changes a leftover from the previous approach?

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.tsx
```diff
@@ -75,7 +76,7 @@ export function addFakeSpyEntry(
       globalFrame: null,
       localFrame: null,
       nonRoundedGlobalFrame: null,
-      componentInstance: false,
+      componentInstance: isComponentInstance(elementOrAttribute),
```
Rheeseyb, line 79: Thanks for fixing this!

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser2.tsx
seanparsons: Nope, we collect more than we used to.

editor/src/core/shared/element-path-tree.ts
```diff
@@ -36,21 +36,23 @@ export function elementPathTree(
...
-export function buildTree(metadata: ElementInstanceMetadataMap): ElementPathTrees {
-  const elementPaths = Object.values(metadata).map((m) => m.elementPath)
-  if (elementPaths.length === 0) {
-    return {}
-  }
-  const root = EP.getStoryboardPathFromPath(elementPaths[0])
-  if (root == null) {
-    return {}
-  }
+export function buildTree(...metadatas: Array<ElementInstanceMetadataMap>): ElementPathTrees {
```
Rheeseyb, line 39: Do we still want this?

seanparsons: No harm in it.

editor/src/core/shared/element-path.ts
```diff
@@ -655,33 +642,9 @@ export function findAmongAncestorsOfPath<T>(
...
-        }
-      }
-    }
-    return true
-  } else {
-    return false
-  }
+  const targetString = toString(target)
+  const maybeAncestorString = toString(maybeAncestor)
+  return targetString !== maybeAncestorString && targetString.startsWith(maybeAncestorString)
```
Rheeseyb, line 647: Nice

editor/src/components/canvas/ui-jsx-canvas-renderer/ui-jsx-canvas-spy-wrapper.spec.browser2.tsx
balazsbajorics: we definitely do not want new metadata entries introduced by this PR.

Rheeseyb (approved):

PR #6118, opened 2024-07-24
Feat/grid container subsection

### LINKED ISSUE #6117: basic grid container inspector

ruggi:
**Problem:**

It should be possible to interact with grid container rows/cols templates from the inspector.

**Fix:**

1. if a grid element is selected, show its rows and columns in the inspector
2. allow editing their values
3. allow removing them
4. allow adding new cols/rows with a default `1fr` size
5. rearrange the grid children so they go "back into the grid" upon removing a col/row

https://github.com/user-attachments/assets/2f28b870-a0a4-4a65-9880-73031ce9d1ee

Fixes #6117

### DIFF
--- a/editor/src/components/inspector/flex-section.tsx
+++ b/editor/src/components/inspector/flex-section.tsx
@@ -32,7 +32,11 @@ import { gridCSSNumber, isCSSNumber, type GridCSSNumber } from './common/css-uti
 import { applyCommandsAction } from '../editor/actions/action-creators'
 import { setProperty } from '../canvas/commands/set-property-command'
 import * as PP from '../../core/shared/property-path'
-import type { GridContainerProperties, GridPosition } from '../../core/shared/element-template'
+import type {
+  GridAutoOrTemplateBase,
+  GridContainerProperties,
+  GridPosition,
+} from '../../core/shared/element-template'
 import {
   gridPositionValue,
   type ElementInstanceMetadata,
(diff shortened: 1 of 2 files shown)


bkrmendy (approved):

editor/src/components/inspector/flex-section.tsx
```diff
@@ -106,10 +122,80 @@ export const FlexSection = React.memo(() => {
...
+
+    if (fromProps?.type !== 'DIMENSIONS' && calculated?.type !== 'DIMENSIONS') {
+      return []
+    }
+
+    let dimensions: GridCSSNumber[] = []
+
+    if (calculated?.type === 'DIMENSIONS') {
+      dimensions.push(...calculated.dimensions)
+    }
```
gbalint, line 154: nitpick, but I think this simple if does not warrant using a let a mutate, and copying the array. I would just use

`const dimensions = calculated?.type === 'DIMENSIONS' ? calculated.dimensions : null`

editor/src/components/inspector/flex-section.tsx
```diff
@@ -106,10 +122,80 @@ export const FlexSection = React.memo(() => {
...
+
+    if (fromProps?.type === 'DIMENSIONS') {
+      if (merged.length === 0) {
+        return fromProps.dimensions
+      } else if (merged.length === fromProps.dimensions.length) {
+        return fromProps.dimensions
+      }
+    }
+
+    return merged
```
gbalint, line 192: looks like a copy paste of the above useMemo, it could be extracted into a function where fromProps and calculated are the parameters (or a parameter whether it is row or column)

gbalint (approved):

editor/src/components/inspector/flex-section.tsx
ruggi: good point! https://github.com/concrete-utopia/utopia/pull/6118/commits/b7cfd566e9c4781df314c0c90f8f92e70c8b2e67

editor/src/components/inspector/flex-section.tsx
ruggi: https://github.com/concrete-utopia/utopia/pull/6118/commits/b7cfd566e9c4781df314c0c90f8f92e70c8b2e67

PR #6131, opened 2024-07-26
Grid row/col gap inspector control

### LINKED ISSUE #6130: Inspector controls for row/col gap

ruggi:
**Problem:**

It should be possible to control `rowGap` and `columnGap` from the grid subsection in the inspector.

**Fix:**

- Add two number controls for col/row gaps
- If `gap` is specified, get rid of it when updating those two gaps
- Added a utility command to perform bulk set/delete operations at the same time for element properties

**Notes**
- The icons for the controls are placeholders, I just used what felt like a good option for now.

https://github.com/user-attachments/assets/caa7667a-3b36-4caa-a1b5-35d5d32fc2e9

Fixes #6130

### DIFF
--- a/editor/src/components/canvas/controls/grid-controls.tsx
+++ b/editor/src/components/canvas/controls/grid-controls.tsx
@@ -425,7 +425,7 @@ export const GridControls = controlForStrategyMemoized(() => {
   )
 
   const grids = useEditorState(
-    Substores.metadataAndPropertyControlsInfo,
+    Substores.metadata,
     (store) => {
       return mapDropNulls((view) => {
         const element = MetadataUtils.findElementByElementPath(jsxMetadata, view)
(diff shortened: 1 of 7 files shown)


bkrmendy (approved):

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -648,35 +649,45 @@ export const GridControls = controlForStrategyMemoized(() => {
...
+              activelyDraggingOrResizingCell != null ? colorTheme.primary.value : 'transparent'
+            }`,
+            padding:
+              grid.padding == null
+                ? 0
+                : `${grid.padding.top}px ${grid.padding.right}px ${grid.padding.bottom}px ${grid.padding.left}px`,
+          }
+
+          // Gap needs to be set only if the other two are not present or we'll have rendering issues
+          // due to how measurements are calculated.
```
liady, line 673: minor - but if we're always only allowing to play with `rowGap` and `columnGap` - why do we need to refer to `gap` at all?
meaning why not have:
```
style.rowGap = grid.rowGap || grid.gap
style.columnGap = grid.columnGap || grid.gap
```
since we don't refer to `gap` anywhere anyway?

liady (approved):

seanparsons (approved):

editor/src/components/canvas/controls/grid-controls.tsx
ruggi: we need to, because `gap` can override the other twos

PR #6132, opened 2024-07-26
Honour props check now works correctly for fragment components.

### LINKED ISSUE #6124: Support updating a component's style attribute when possible or prevent updating full stop.
Taking this project as an example: https://utopia.pizza/p/1f34c834-brindle-output/

It should be possible to move and resize both `MyCompDiv` and `MyCompFrag` and have those update the style attribute in identical ways, so e.g. resizing both would result in:
```ts
      <MyCompDiv style={{ height: 20 }} />
      <MyCompFrag style={{ height: 20 }} />
```

This should be achieved by having `getElementFragmentLikeType` not return `sizeless-div` for components at all, so if there is a case where that would actually be incorrect then that needs to be flagged (though that would also likely be a separate issue to be dealt with later, and potentially via the component annotation API).

As well as that, the `componentHonoursProps*` functions only check that the root element uses those props, so that will need to be addressed (https://github.com/concrete-utopia/utopia/blob/master/editor/src/core/model/element-template-utils.ts#L876-L896), and the selective DOM walking doesn't reconstruct the frames, which will be required to show the selection outline update in real time during interactions.

seanparsons:
**Problem:**
With a component that honours `props.style` but inside a fragment, it should be possible to update an instance of the component to resize/move it.

**Fix:**
This extends the coverage of the `componentHonoursProps*` functions to include delving inside root fragments.

As well as a slight change to the DOM walker to check for descendant of valid paths not just the exact valid paths.

**Commit Details:**
- Copied `isDescendantOf` optimisation to `isDescendantOfOrEqualTo`.
- `componentHonoursPropsPosition` and `componentHonoursPropsSize` now drill inside fragments at the root of the component.
- `runSelectiveDomWalker` now checks for descendants of the valid paths so as to catch the content within fragments, where the fragment is at the root of a component.
- Slight tweak to `collectAndCreateMetadataForElement` to not both collecting anything if `pathsForElement` is empty.

**Manual Tests:**
I hereby swear that:

Fixes #6124


seanparsons:
Superceded by a slightly different approach.

PR #6133, opened 2024-07-26
Simplify the optics.

seanparsons:
**Problem:**
The optics code was comprehensive but most of the core benefits we didn't particularly use and it made the code more difficult to improve.

**Fix:**
Replaced the four different optic representations with a single one.

Adding a benchmark suite to check the performance results of the changes.

**Performance Results:**
  array from nested array:
    952,354 ops/s, ±0.33% -> 11,938,203 ops/s, ±0.51%
  first value from nested array:
    950,064 ops/s, ±0.16% -> 12,339,498 ops/s, ±0.20%
  modify a nested array:
    11,456,420 ops/s, ±0.44% -> 11,622,899 ops/s, ±0.39%
  array from project contents:
    513,388 ops/s, ±3.17% -> 863,201 ops/s, ±0.37%
  modify project contents:
    37,821 ops/s, ±0.99% -> 38,053 ops/s, ±0.16%

**Commit Details:**
- Removed the four different representations of the optics.
- Simplified the composition and created a new utility function `makeOptic` to handle adding the `compose` function.

**Manual Tests:**
I hereby swear that:


editor/src/benchmarks.benchmark.ts
```diff
@@ -1,9 +1,13 @@
 import { benchmarkBuildTree } from './core/shared/element-path-tree.benchmark'
 import { benchmarkElementPathFunction } from './core/shared/element-path.benchmark'
 import { benchmarkAttributes } from './core/shared/jsx-attributes.benchmark'
+import { benchmarkOptics } from './core/shared/optics.benchmark'
 import { benchmarkPropertyPathFunction } from './core/shared/property-path.benchmark'
 
+/*
```
Rheeseyb, line 7: Delete me?

Rheeseyb (approved):
Nice!

editor/src/benchmarks.benchmark.ts
seanparsons: Needs to be kept, because of refactoring the `jsx-attributes.benchmark.ts` benchmark (at least) that causes the benchmarks to fail.

balazsbajorics (approved):

PR #6147, opened 2024-07-30
Draw to insert into grid

bkrmendy:
## Problem
While draw to insert technically works when drawing into a grid, the existing strategies don't assign the grid positioning properties (`gridRow`, `gridColumn`), so the newly drawn element is positioned relatively its parent, not relative to the grid cell it was drawn into. This results in weird behaviour when the newly drawn element is dragged into a different grid cell.

https://github.com/user-attachments/assets/dca5641a-2e7e-4254-9b1b-3be59bd5386f

## Fix
This PR fixes the problem by creating a grid-aware draw to insert strategy, which correctly assigns the grid positioning props, and which produces grid elements that work as expected.

https://github.com/user-attachments/assets/0d03a908-6dcd-4ba7-99f9-7ec16d58b6af

### Out of Scope
The existing grid move strategy doesn't change the absolute positioning props of grid elements (yet). Because of this, if an absolutely positioned grid element is moved to a different grid cell, it'll retain its absolute positioning props. This will be fixed on a follow-up PR

### Implementation notes
- The grid-specific draw to insert strategy is implemented as a separate strategy for clarity (so that it's easier to read/maintain it)
- `drawToInsertMetaStrategy` and `gridDrawToInsertStrategy` are created by `drawToInsertMetaMetaStrategy`, which is basically an if that checks whether the element under the cursor is a grid, and returns the appropriate strategy
- the code for inserting text into a grid is in `gridDrawToInsertStrategy`. The command that kicks off the text edit mode could be taken out of the strategy's `apply` function, but at this point it was easier to put in there compared to manually calling the apply function and appending the text edit mode command to the commands returned by it


editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-meta-meta-strategy.ts
```diff
@@ -0,0 +1,27 @@
...
+      MetadataUtils.findElementByElementPath(canvasState.startingMetadata, targetParent),
+    )
+  ) {
+    return stripNulls([
+      gridDrawToInsertStrategy(canvasState, interactionSession, customStrategyState),
+    ])
+  }
+
+  return drawToInsertMetaStrategy(canvasState, interactionSession, customStrategyState)
+}
```
gbalint, line 27: This metametastrategy looks so simple, did you think of just putting this at the beginning of drawToInsertMetaStrategy? It looks unnecessary to me to add another level of meta strategy here.

ruggi (approved):
This is fantastic. The only piece of advice I have is that it'd be neat to see the grid wireframe when hovering it, without having to start dragging on it. Especially for the text edit mode it would make it a lot better imho.

editor/src/components/canvas/canvas-strategies/strategies/draw-to-insert-meta-meta-strategy.ts
bkrmendy: good idea, I did it here https://github.com/concrete-utopia/utopia/pull/6147/commits/cfc3a7259a307657b9bf53dde0c55b8100a713c3

gbalint (approved):

Issue #6155, opened 2024-08-01
combined number/keyword/dropdown input for grid cols/rows


ruggi:
requires #6156 to be merged in

PR #6157, opened 2024-08-01
Grid dimensions supporting keywords

### LINKED ISSUE #6156: support keywords in grid dimensions

ruggi:
**Problem:**

We need to support keywords inside grid template dimensions.

**Fix:**

Extend the grid dimensions type (renamed from `GridCSSNumber` to `GridDimension`) to be either a CSSNumber or a CSSKeyword, and the related parsing.

Important note: this is an intermediate PR towards #6155 and **temporarily** this PR defaults a col/row value to `0` if it's not a CSSNumber (as commented in the code).

No functional changes otherwise.

Fixes #6156 
Prerequisite for #6155

### DIFF
--- a/editor/src/components/inspector/common/css-utils.ts
+++ b/editor/src/components/inspector/common/css-utils.ts
@@ -597,14 +597,17 @@ export type GridCSSNumber = BaseGridDimension & {
 
 export type GridCSSKeyword = BaseGridDimension & {
   type: 'KEYWORD'
-  value: CSSKeyword<string>
+  value: CSSKeyword<ValidGridDimensionKeyword>
 }
 
 export function isGridCSSKeyword(dim: GridDimension): dim is GridCSSKeyword {
   return dim.type === 'KEYWORD'
 }
 
-export function gridCSSKeyword(value: CSSKeyword<string>, areaName: string | null): GridCSSKeyword {
+export function gridCSSKeyword(
+  value: CSSKeyword<ValidGridDimensionKeyword>,
+  areaName: string | null,
+): GridCSSKeyword {
   return {
     type: 'KEYWORD',
     value: value,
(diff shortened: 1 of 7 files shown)


editor/src/components/inspector/flex-section.tsx
```diff
@@ -194,28 +194,43 @@ const TemplateDimensionControl = React.memo(
...
-
-        const newValues = [...values]
-        newValues[index] = gridCSSNumber(
-          value.value,
-          (value.unit as GridCSSNumberUnit) ?? values[index].unit,
-          values[index].areaName,
-        )
+      (index: number) => (value: UnknownOrEmptyInput<CSSNumber | CSSKeyword<string>>) => {
+        if (isCSSKeyword(value)) {
+          const newValues = [...values]
```
liady, line 199: we can use [`with`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/with) here (and in line 217 as well)
```
const newValues = values.with(index, isCSSNumber(value)
            ? gridCSSNumber(value, null)
            : gridCSSKeyword(value, null))
```

liady (approved):

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -841,9 +882,27 @@ export const parseCSSUnitlessAsNumber = (input: unknown): Either<string, number>
   }
 }
 
+const validGridDimensionKeywords: string[] = [
```
gbalint, line 885: You could make this a type, and then it would be possible to make the code typesafe

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -841,9 +882,27 @@ export const parseCSSUnitlessAsNumber = (input: unknown): Either<string, number>
...
+  'inherit',
+  'initial',
+  'unset',
+  'subgrid',
+  'auto-fit',
+  'auto-fill',
+  // NOTE: function keywords are omitted as they are treated separately
+]
+
+export function isValidGridDimensionKeyword(s: string): boolean {
```
gbalint, line 899: This could return `s is GridDimensionKeyword` if you had a type for that

editor/src/components/inspector/flex-section.tsx
liady: resolving until we'll support `es2023` syntax

editor/src/components/inspector/common/css-utils.ts
ruggi: done in [`d6ffcc7` (#6157)](https://github.com/concrete-utopia/utopia/pull/6157/commits/d6ffcc73f4a76200d91a3efe3769d0404a8461dd)

liady (approved):

gbalint (approved):

Issue #5964, opened 2024-06-17
selecting content underneath the header does not work


maltenuhn:
**Problem**
(open ended) trying to select content underneath stuff rendered by the Layout component is not possible even w/ cmd-click
We probably need a way to annotate such extraneous, non-editable content in some way

Project link: https://utopia.pizza/p/960ac92c-concrete-utopia-hydrogen-editions-24/

![Image](https://github.com/concrete-utopia/utopia/assets/2945037/3cda1d98-be47-4813-9c67-3bacd4eaef04)

gbalint:
First step:
[ ] fix locking so you can select the element behind a locked element

After this fix, we can persist the locking and/or add component annotation about autolocking the internals of a component.

PR #6158, opened 2024-08-01
Selection through locked elements

gbalint:
**Problem:**
When an element is locked, you can not select it. However, it is not implemented that you should be able to select an element behind a locked element.
First task of https://github.com/concrete-utopia/utopia/issues/5964

**Fix:**
`findValidTarget` iterates through the elements in `document.elementsFromPoint` (which is in z-order), and always tries to find a selectable utopia element connected to the element (using `firstAncestorOrItselfWithValidElementPath`). The first one which is found will be the selected element.

However, when you click on a locked element, it is going to find its parent (or other ancestor) and select that, even if there is an unlocked element behind it.

My solution was that `firstAncestorOrItselfWithValidElementPath` should also return back whether the found element was coming from a search starting from a locked element. If yes, we continue to search in the elements from `document.elementsFromPoint`, maybe we can find an element which was behind the locked element earlier and is unlocked. If there is, we can return that, if there isn't, we can still return the element we found first.

This fix solves the problem for embedded elements, but not good enough in cases when  hierarchy locked element is inside a component which is not focused. `getSelectableViews` uses the locked element list coming from `getAllLockedElementPaths`, which returns all the locked elements and all the descendants of the hierarchy locked elements coming from the metadata. This method does not return the descendants which are in unfocused components, because those elements are not in the metadata. This problem caused some of the shenanigans in the sample store Layout component, where hierarchy locking some of internals did not work well.

As a simple solution I just don't rely on the all locked elements list to decide what is locked in `firstAncestorOrItselfWithValidElementPath`, but I check it there if the element under the mouse is a descendant of a hierarchy locked element or not.

TODO: let's remove `getAllLockedElementPaths` and use a memoized `isElementLocked` instead of that in getSelectableViews too.

**Manual Tests:**
I hereby swear that:


ruggi (approved):

liady (approved):

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -274,26 +293,35 @@ function findFirstValidParentForSingleElementUncached(
...
       point,
+      lockedElements,
     )
-    if (foundValidElementPath != null) {
-      return foundValidElementPath
+    if (p != null) {
+      foundValidElementPath = p.path
+      if (!p.locked) {
+        break
+      }
```
balazsbajorics, line 320: nice

balazsbajorics (commented):
👌

PR #6160, opened 2024-08-01
feat(canvas): grid gap controls

### LINKED ISSUE #6163: Grid gap canvas controls
We want to have grid gap on-canvas controls similar to the flex ones

liady:
**Problem:**
We currently don't have inline grid gap handlers for the canvas

**Fix:**
This PR adds the inline gap handlers, calculating them according to the grid that is used to display the placeholder cells.

<video src="https://github.com/user-attachments/assets/eb219d35-5c37-4d1c-898d-73ab67ce5354"></video>

Fixes #6163


seanparsons (approved):

editor/src/core/shared/array-utils.ts
```diff
@@ -525,3 +525,25 @@ export function isPrefixOf<T>(
...
 }
+
+export function createArrayWithLength<T>(length: number, value: T | ((index: number) => T)): T[] {
+  return Array.from({ length }, (_, index) => {
+    // see issue https://github.com/microsoft/TypeScript/issues/37750
+    return value instanceof Function ? value(index) : value
+  })
+}
+
+export function matrixFrom<T>(iterable: Iterable<T>, width: number): T[][] {
```
bkrmendy, line 536: This seems to be unused

editor/src/core/shared/array-utils.ts
```diff
@@ -525,3 +525,25 @@ export function isPrefixOf<T>(
 export function valueOrArrayToArray<T>(ts: T | T[]): T[] {
   return Array.isArray(ts) ? ts : [ts]
 }
+
+export function createArrayWithLength<T>(length: number, value: T | ((index: number) => T)): T[] {
```
bkrmendy, line 529: I'd keep the function signature simple and only accept functions (this is the same with the builtin `map` function)

editor/src/components/canvas/gap-utils.ts
```diff
@@ -153,6 +171,157 @@ export function gapControlBoundsFromMetadata(
   )
 }
 
+export function gridGapControlBoundsFromMetadata(
+  elementMetadata: ElementInstanceMetadataMap,
```
bkrmendy, line 175: this param is unused

editor/src/components/canvas/gap-utils.ts
```diff
@@ -153,6 +171,157 @@ export function gapControlBoundsFromMetadata(
...
+  }>
+  rows: number
+  columns: number
+  cellBounds: CanvasRectangle
+  gapValues: { row: CSSNumber; column: CSSNumber }
+  gridTemplateRows: string
+  gridTemplateColumns: string
+} {
+  const parentGridPlaceholderId = `grid-${EP.toString(parentPath)}`
+  const parentGrid = document.getElementById(parentGridPlaceholderId)
```
bkrmendy, line 195: as discussed on Discord, let's not use the DOM query to get the parent grid. We can do this based on the parent path, let's use that to stay consistent

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -749,7 +765,11 @@ export const GridControls = controlForStrategyMemoized<GridControlsProps>(({ tar
           }
 
           return (
-            <div key={`grid-${EP.toString(grid.elementPath)}`} style={style}>
+            <div
+              key={`grid-${EP.toString(grid.elementPath)}`}
+              id={`grid-${EP.toString(grid.elementPath)}`}
```
bkrmendy, line 770: This isn't necessary if we use the metadata to look up grids

editor/src/components/canvas/gap-utils.ts
liady: this is not getting the actual element here, but actually the grid placeholder - the one that is drawn in the grid controls, so no path for it.
(also mentioned it in my reply on Discord)

editor/src/components/canvas/controls/grid-controls.tsx
liady: same here - we need to query the grid placeholder (the one that displays the placeholder cells, since we need the cells and not the elements), so we can't use the metadata for it

editor/src/components/canvas/gap-utils.ts
bkrmendy: Makes sense, the main problem with this is that this is a huge side effect, so if we're going to start using this from multiple places we should extract this into a function that has a big exclamation mark attached
otherwise I'm afraid that this will become a little gotcha that we'll have to keep in mind all the time when we're working on the grid code

editor/src/components/canvas/controls/grid-controls.tsx
liady: done in https://github.com/concrete-utopia/utopia/pull/6160/commits/dd7e663d8985a530ad6b33d7f481fdb5a4fd2563

bkrmendy (approved):

PR #6172, opened 2024-08-05
fix: getAllUniqueUids speedup

balazsbajorics:
**Problem:**

image

Dragging the PromiseCard in the sample store takes about 22ms, of which 3-4 ms is a rebuilding of the uid -> file mapping of the entire project contents, including all the files that did not even change:

image

**Investigation:**
The function responsible for creating the uid -> file mapping is called `getAllUniqueUids`, because its original duty was to check for duplicate uids across the project. It was written in a way that it had to run for the entire project contents, and it was memoized by the projectContents reference. This meant that changing a single file would force us to have to re-run `getAllUniqueUidsInner` for the entire project
Over time, the UID uniqueness check became less important, but this function became crucial in creating a lookup of uid -> filePath. This lookup is used to speed up most of our helper functions that reach into the project contents to read or change the value of an element or component. It is crucial to keep this lookup because otherwise the speed of those lookups would suffer especially for large projects.

**Better Caching**
So what we need is a better cacheable way that can store results for a given file in ProjectContents, and return those if we see the same ProjectContents file. Some of our ad-hoc caching in the DerivedState worked in a way that we passed in the previous DerivedState as the cache's data source. This makes memory management very easy, because then there is no global cache object which may accidentally introduce a memory leak by retaining references to old unneeded derivedStates. Unfortunately this was not an option here, because this function is now called from hooks, so there's no centralized "previous value" we can use as the cache source.

So instead I decided to try something we haven't really used in Utopia before, a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap)! I created a global WeakMap object that stores the previous results for each file, the key is the file reference, but it only holds a weak reference to the key which means if the file reference is garbage collected, the WeakMap automatically forgets that cached value too! Pretty neat!

**Details:**
- Since we very rarely used `getAllUniqueUids` to check the duplicated uids, I renamed it to `getUidMappings`
- To achieve easy and quick caching, `getUidMappings` returns `filePathToUids` instead of a uid -> filePath lookup, so I added a helper `getFilePathForUid`
- similarly, to get an array of all UIDs (something we rarely need and we used to calculate for every frame _twice_), we now need to call the new helper `getAllUniqueUidsFromMapping`
- the file is now named `get-uid-mappings.ts` which caused a bunch of diffs, but it also means you can review the code zero-based instead of an inscrutable botched diff.


editor/webpack-benchmarks.config.js
```diff
@@ -30,6 +30,8 @@ module.exports = {
           },
         ],
       },
+      // CSS Loading.
+      { test: /\.css$/, use: 'null-loader' },
```
balazsbajorics, line 34: I fixed the old benchmarks btw!

editor/src/core/model/get-uid-mappings.ts
```diff
@@ -0,0 +1,277 @@
...
+}
+
+function collectUidsForEachFile(projectContents: ProjectContentTreeRoot): GetAllUniqueUIDsResult {
+  let workingFileToUidMapping: FileToUidMapping = new Map()
+  let workingDupliactedUIDs: DuplicateUIDsResult = new Map()
+  walkContentsTreeForParseSuccess(projectContents, (filePath, parseSuccess) => {
+    const mappingsForFile = collectUidsForFile(filePath, parseSuccess)
+    // mappingsForFile.uniqueIDs.forEach((uid) => {
+    //   workingAllUids.add(uid)
+    // })
```
gbalint, line 209: Did you intend to leave this commented out code here?

balazsbajorics: oops,fixed!

editor/src/core/model/get-uid-mappings.ts
```diff
@@ -0,0 +1,274 @@
...
+    duplicateIDs: workingDupliactedUIDs,
+  }
+}
+
+export function clearCachedUidsPerFileForTests() {
+  CachedUidsPerFile = new WeakMap()
+}
+
+export function getFilePathForUid(mapping: FileToUidMapping, uid: string): string | null {
+  // let result: string | null | undefined = null
```
bkrmendy, line 223: Was this pushed accidentally?

bkrmendy (approved):

editor/src/core/model/get-uid-mappings.ts
balazsbajorics: I think so, fixed

editor/src/components/editor/store/editor-state.ts
```diff
@@ -3501,12 +3501,12 @@ export function getElementPathsInBounds(
   }
 }
 
-export function modifyParseSuccessAtPath(
+export function modifyParseSuccessAtPath<E extends { projectContents: ProjectContentTreeRoot }>(
```
seanparsons, line 3504: Probably want to rename this now.

editor/src/core/model/get-uid-mappings.ts
```diff
@@ -0,0 +1,273 @@
...
+  working: Set<string>,
+  duplicatedIDs: DuplicateUIDsResult,
+  filePath: string,
+  uid: string,
+): void {
+  if (working.has(uid)) {
+    duplicatedIDs.set(uid, filePath)
+  }
+
+  working.add(uid)
```
seanparsons, line 38: Can put this in an else, so we don't run it unnecessarily.

editor/src/core/model/get-uid-mappings.ts
```diff
@@ -0,0 +1,273 @@
...
+): GetAllUniqueUIDsResult {
+  return collectUidsForEachFile(projectContents)
+}
+
+export const getUidMappings = memoize(getUniqueUidsMappingInner)
+
+export function getAllUniqueUidsFromMapping(mapping: FileToUidMapping): Array<string> {
+  let result: Array<string> = []
+  for (const [filePath, uids] of mapping.entries()) {
+    result.push(...Array.from(uids))
```
seanparsons, line 242: Save creating an intermediary array:
```suggestion
    uids.forEach(uid => result.push(uid))
```

seanparsons (approved):

editor/src/core/model/get-uid-mappings.ts
balazsbajorics: `result.push(...Array.from(uids))` is from somewhere on the internet, it is accepted as the fastest way to achieve copying. I ran a Benny benchmark and with the change the code would be somewhere between 2-6% slower

PR #6184, opened 2024-08-06
Unified prop and comment flags

gbalint:
**Problem:**
I would like to save the locked state of elements into the project.

We have two ways to persist extra utopia specific information into the project code:
- comment flags
- data attributes.

For locking I would need both, because you can lock all different kinds of elements, some support props, other comments.

This PR is a step to unify these two ways of storing information in the code: you can store the same information either in a comment or in a prop based on the element's type. JSX elements support props, while other kind of elements support comments.

Unfortunately, there are element types which don't support neither props nor comments (e.g. JSXExpressionFunctionCall or JSXTextBlock). It is still not possible to save these flags onto these elements.
To fix that we need to make sure the parser adds the comments to these element types too.

- I extended the comment-flags file to handle data flags. I added unified methods which can read or add flags, and based on the element type they either use comments or props.
- First I renamed comment-flags.ts to utopia-flags.ts, but I realized that it destroys the readability of the diff, so I plan to do that in a subsequent PR.
- I modified the can-condense handling to use the new methods, so after this PR it is possible to set a conditional or a code element to condensable (however, it seems the navigator does not handle that, that can also be fixed in a subsequent PR, my goal was not to work on the condensing feature).

**Manual Tests:**
I hereby swear that:

**TODO:**

- [ ] Rename comment-flags.ts
- [ ] Check why non-jsx-elements are still not condensed in the navigator
- [ ] Use the unified approach with other comment flags (not just can-condense). This is debatable in cases where the flag is only supported by a single element type (conditional override, map count)


editor/src/utils/can-condense.spec.browser2.ts
Rheeseyb, line 1: Did you mean to commit this?

gbalint: oh, i did not

PR #6210, opened 2024-08-08
Grid move: fallback to prevent flicker

### LINKED ISSUE #6209: fix flicker when moving a cell inside a grid

ruggi:
**Problem:**

When moving a grid element, if the mouse goes over a non-grid-target area the element will temporarily jump back to its original position, creating effectively a flicker effect during the interaction.

**Fix:**

If the cell under the mouse is not found, instead of bailing try to fallback to the last known valid cell target instead.

| Before | After |
|----------|------------|
| Kapture 2024-08-08 at 15 34 38 | Kapture 2024-08-08 at 15 34 04 |

Fixes #6209


bkrmendy (approved):

liady (approved):
👏

PR #6211, opened 2024-08-08
decouple dom-walker from canvas re-render

balazsbajorics:
**Problem:**
Today dom-walker runs when the canvas re-renders. But instead we want to be able to sometimes re-render the canvas and not run the dom-walker (when we scroll the canvas, for example), and sometimes we want to re-run the dom-walker but not re-render the canvas (when a `RUN_DOM_WALKER` action is dispatched)

**Fix:**
- Decouple `shouldRerender` and `runDomWalker`.
- Revert `shouldRerender` to how it was a few months ago: if anything in the dispatch result changes, re-render the canvas subtree. the Canvas code has enough internal memoization to be resilient
- `shouldRunDOMWalker` instead of checking the full EditorState.canvas, just checks a few important keys from it, most importantly, it does NOT check the scroll offset.
- We want to be able to check a few keys of an object but ensure that we provided an exhaustive list of what to check and what to deliberately ignore – because otherwise we may silently fail to check important new keys! To achieve this, I added a new `keysEqualityExhaustive` that uses some typescript dark magic.

**Warning**
These changes were made in the parts of the dispatch which is not testable by any of our test suites. (Something we should address in the future), so please help in being vigilant and looking for potential performance regressions this change might cause!

### DIFF
--- a/editor/src/core/shared/equality-utils.ts
+++ b/editor/src/core/shared/equality-utils.ts
@@ -71,3 +71,27 @@ export function keepReferenceIfShallowEqual<T>(original: T, maybeNew: T) {
     return maybeNew
   }
 }
+
+function keyEquality<T>(key: keyof T, a: T, b: T): boolean {
+  return a[key] === b[key]
+}
+
+export function keysEquality<T>(keys: ReadonlyArray<keyof T>, a: T, b: T): boolean {
+  return keys.every((key) => keyEquality(key, a, b))
+}
+
+type AtLeastOne<T> = [T, ...T[]]
+
+export const keysEqualityExhaustive =
+  <O extends { [key: string]: any }>() =>
+  <L1 extends AtLeastOne<keyof O>, L2 extends Array<Exclude<keyof O, L1[number]>>>(options: {
+    include: L1
+    exclude: L2 extends any
+      ? Exclude<keyof O, L1[number] | L2[number]> extends never
+        ? L2
+        : Exclude<keyof O, L1[number] | L2[number]>[]
+      : never
+  }) =>
+  (a: O, b: O): boolean => {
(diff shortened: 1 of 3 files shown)


editor/src/core/shared/equality-utils.ts
```diff
@@ -71,3 +71,27 @@ export function keepReferenceIfShallowEqual<T>(original: T, maybeNew: T) {
...
+  return keys.every((key) => keyEquality(key, a, b))
+}
+
+type AtLeastOne<T> = [T, ...T[]]
+
+export const keysEqualityExhaustive =
+  <O extends { [key: string]: any }>() =>
+  <L1 extends AtLeastOne<keyof O>, L2 extends Array<Exclude<keyof O, L1[number]>>>(options: {
+    include: L1
+    exclude: L2 extends any
```
seanparsons, line 89: What is this field for?

seanparsons (approved):

editor/src/core/shared/equality-utils.ts
balazsbajorics: its an exhaustive exclude list: the idea is that you must exhaustively enumerate all keys of the tested object: you have to explicitly say if they should be included in the comparison, or ignored. (maybe ignoredKeys would be a better name?)

seanparsons: Is this so that if we added another field to the type then it's a compile error until it's added to the include or exclude list?

balazsbajorics: exactly!

gbalint: this is awesome!

gbalint (approved):

PR #6217, opened 2024-08-08
fix(inspector): keep opener icon when dropdown is open

### LINKED ISSUE #6164: disappearing grid row context menu ellipses
To open the context menu for a grid row or column, you click on the `...` that becomes visible when hovering over the row. However, that `...` should stay visible when the context menu is open. The button should actually get a gray background when the context menu is open as well.

Image

liady:
**Problem:**
When a grid dropdown is open, its icon disappears
Image

**Fix:**
The icon will stay and have a subdued color when the dropdown is open

<video src="https://github.com/user-attachments/assets/aca01425-2f69-4cbe-890c-f75449d9b982"></video>

**Commit Details:** 
- pass the open state to the icon opener
- take the open state into consideration for visibility

**Manual Tests:**
I hereby swear that:

Fixes #6164

### DIFF
--- a/editor/src/components/inspector/flex-section.tsx
+++ b/editor/src/components/inspector/flex-section.tsx
@@ -500,12 +500,11 @@ function AxisDimensionControl({
     <div
       key={`col-${value}-${index}`}
       style={{ display: 'flex', alignItems: 'center', gap: 6 }}
-      className={isOpen ? 'openMenu' : ''}
       css={{
         [`& > .${axisDropdownMenuButton}`]: {
-          visibility: 'hidden',
+          visibility: isOpen ? 'visible' : 'hidden',
         },
-        ':hover, &.openMenu': {
+        ':hover': {
           [`& > .${axisDropdownMenuButton}`]: {
             visibility: 'visible',
           },
--- a/editor/src/components/inspector/flex-section.tsx
+++ b/editor/src/components/inspector/flex-section.tsx
@@ -430,14 +430,14 @@ const TemplateDimensionControl = React.memo(
     )
 
     const openDropdown = React.useCallback(
-      () => (
+      (isOpen: boolean) => (
         <SquareButton
           data-testid={'openDropdown'}
           highlight
           onClick={NO_OP}
           style={{ width: 12, height: 22 }}
         >
-          <Icons.Threedots />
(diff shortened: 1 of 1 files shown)


editor/src/components/inspector/flex-section.tsx
```diff
@@ -457,59 +457,92 @@ const TemplateDimensionControl = React.memo(
...
+            opener={openDropdown}
+          />
+        ))}
       </div>
     )
   },
 )
 TemplateDimensionControl.displayName = 'TemplateDimensionControl'
 
+function AxisDimensionControl({
```
liady, line 477: extracted to a component to have an "open" state for every dropdown trigger

editor/src/components/inspector/flex-section.tsx
```diff
@@ -457,59 +457,92 @@ const TemplateDimensionControl = React.memo(
...
+      css={{
+        [`& > .${axisDropdownMenuButton}`]: {
+          visibility: 'hidden',
+        },
+        ':hover, &.openMenu': {
+          [`& > .${axisDropdownMenuButton}`]: {
+            visibility: 'visible',
+          },
+        },
+      }}
```
bkrmendy, line 513: Why not use the `open` variable instead of the classname-based code here?

liady: i was debating myself on that. 
to keep the css specificity logic we have to do it here (using the `open` variable in the `style` prop will be overriding this definition.
so the other way I considered was:
```tsx
css={{
  [`& > .${axisDropdownMenuButton}`]: {
    visibility: isOpen ? 'visible' : 'hidden', // using it here
  },
  ':hover': {
    [`& > .${axisDropdownMenuButton}`]: {
      visibility: 'visible',
    },
  },
}}
```
but this felt a bit clunky and also not very extensible if we'd want other cases. also if in the future we will want another thing to happen, we will have to change both places.

I thought that adding it as another case in the selector next to the `:hover` class makes most sense, since it really is "either it is hovered or the menu is opened".
We can also use a data-attribute instead of a class (`data-dropdown-open`), but I would still add it to the `:hover` css selector, imo.

Rheeseyb (approved):

bkrmendy (approved):

PR #6222, opened 2024-08-09
Update VS Code and enable Intellisense

Rheeseyb:
**Problem:**
Our version of VS Code is extremely outdated and built using a reverse engineered build, but since then the web version of VS Code has come a long way and can now support Intellisense.

**Fix:**

**Commit Details:**
- Update the vscode-build's `shell.nix` to use newer versions of node and yarn, and supply an extra package required for building it
- Update the pulled version of VS Code to 1.91.1
- Remove the reliance on IndexedDB for communicating between the outer edge of the iframe and the extension, and instead use VS Code's command service for passing messages between the two, meaning that we no longer need to use polling to watch for file changes, and can instead directly trigger them
  - This required the registering of some extra commands, and the code for then communicating those across the iframe boundary
- Remove the `fs` files from `utopia-vscode-common`, and instead use an in memory fs implementation in the extension. This stores the files in a `Map<string, FSNode>`, and stores both the saved and unsaved content of text files
- Since we are now passing messages directly between Utopia and VS Code, I've flattened the message types (previously there were two types: one for file changes; one for other changes such as cursor position, highlight bounds etc)
- The server now passes down the headers required to enable [`crossOriginIsolated`](https://developer.mozilla.org/en-US/docs/Web/API/Window/crossOriginIsolated). `crossOriginIsolated` is required for full Intellisense on the web, which requires 2 new cross origin headers, and for the code editor iframe to be served up from the same TLD as the rest of the editor.
  - This does mean that the code editor will be running on the same process as Utopia
- To actually build this version of VS Code I had to:
  - Disable code mangling. I'm, not sure why, but with that enabled the build was just silently hanging on my machine
  - Remove some native keyboard mapping code. This was causing an issue related to electron, which we are clearly not interested in, so I just ripped it out after failing to fix it.
- To simplify the VS Code UI I have:
  - Forcibly enabled Zen mode (the `isZenModeActive` check always returns true)
  - Changed some Zen mode defaults to prevent the code editor trying to fullscreen itself, or centre the code pane inside itself
  - Forcibly hidden other parts of the UI that would normally still be shown when in non-fullscreen Zen mode

**Manual Tests:**
I hereby swear that:


editor/src/templates/vscode-editor-iframe/index.html
```diff
@@ -10,9 +10,9 @@
     <meta id="vscode-workbench-builtin-extensions" data-settings="[]" />
 
     <link
-      data-name="vs/workbench/workbench.web.api"
+      data-name="vs/workbench/workbench.web.main"
```
Rheeseyb, line 13: This is just because the name of the output of the build has changed

server/src/Utopia/Web/Endpoints.hs
```diff
@@ -277,7 +277,7 @@ injectIntoPage toInject@(_, _, _, _, editorScriptTags) (TagComment "editorScript
 injectIntoPage toInject (firstTag : remainder) = firstTag : injectIntoPage toInject remainder
 injectIntoPage _ [] = []
 
-renderPageWithMetadata :: Maybe ProjectIdWithSuffix -> Maybe ProjectMetadata -> Maybe DB.DecodedProject -> Maybe Text -> Text -> ServerMonad H.Html
+renderPageWithMetadata :: Maybe ProjectIdWithSuffix -> Maybe ProjectMetadata -> Maybe DB.DecodedProject -> Maybe Text -> Text -> ServerMonad ProjectPageResponse
```
Rheeseyb, line 280: The new `ProjectPageResponse` type is so that the return values of these functions can include the new headers required for `crossOriginIsolated`

utopia-vscode-common/src/fs/fs-core.ts
Rheeseyb, line 1: Since we no longer need to use IndexedDB to back our file system, and we no longer need to use it for passing messages, these files went in the bin and the remaining required functionality was moved into the extension

utopia-vscode-extension/src/extension.ts
```diff
@@ -2,60 +2,56 @@ import * as vscode from 'vscode'
...
+  exists,
+  readFileAsUTF8,
+  setErrorHandler,
+  writeFileUnsavedContentAsUTF8,
+} from './in-mem-fs'
 
 const FollowSelectionConfigKey = 'utopia.editor.followSelection.enabled'
 
-export async function activate(context: vscode.ExtensionContext): Promise<void> {
+export function activate(context: vscode.ExtensionContext) {
```
Rheeseyb, line 35: Since we're no longer relying on IndexedDB, all of the FS functionality is now sync

utopia-vscode-extension/src/extension.ts
```diff
@@ -244,13 +218,15 @@ async function doSubscriptionWork(work: SubscriptionWork): Promise<void> {
...
+function runPendingSubscriptionChanges() {
   minimisePendingWork()
   for (const work of pendingWork) {
-    await doSubscriptionWork(work)
+    doSubscriptionWork(work)
   }
 
   pendingWork = []
+
+  // TODO should we still do it like this, or instead follow the pattern used by queueEvents?
```
Rheeseyb, line 229: In `utopia-fs.ts` we emit events to inform VS Code of certain file changes, and we batch those together using a 5ms timeout which is reset whenever another event is added to the queue. This just means that when e.g. a ton of new files are added (when initialising the project) we will emit all of those events at the same time. I _think_ we should take the same approach here rather than having a poll that runs every 100ms regardless of the amount of pending work. I'll test that out in a follow up PR.

utopia-vscode-extension/src/extension.ts
```diff
@@ -334,56 +307,106 @@ function getFullConfig(): UtopiaVSCodeConfig {
...
-        sendMessage(vsCodeReady())
-        break
-      default:
-        const _exhaustiveCheck: never = message
-        console.error(`Unhandled message type ${JSON.stringify(message)}`)
-    }
-  }
+function sendMessageToUtopia(message: FromVSCodeToUtopiaMessage): void {
+  vscode.commands.executeCommand('utopia.toUtopiaMessage', message)
+}
```
Rheeseyb, line 312: This part is the crux of the new communication approach. We have this extension and VS Code register a pair of new commands for passing messages between themselves, with the VS Code side then forwarding those messages straight onto Utopia via a `window.postMessage` (since that isn't possible from the extension).

utopia-vscode-extension/src/utopia-fs.ts
```diff
@@ -89,13 +91,9 @@ export class UtopiaFSExtension
...
   // FileSystemProvider
   readonly onDidChangeFile: Event<FileChangeEvent[]> = this.fileChangeEventQueue.emitter.event
-  readonly onUtopiaDidChangeSavedContent: Event<Uri[]> =
-    this.utopiaSavedChangeEventQueue.emitter.event
-  readonly onUtopiaDidChangeUnsavedContent: Event<Uri[]> =
-    this.utopiaUnsavedChangeEventQueue.emitter.event
 
-  private queueEvent<T>(event: T, eventQueue: EventQueue<T>): void {
-    eventQueue.queue.push(event)
+  private queueEvents<T>(events: Array<T>, eventQueue: EventQueue<T>): void {
```
Rheeseyb, line 95: This is the event queue for notifying VS Code of file changes

utopia-vscode-extension/src/utopia-fs.ts
```diff
@@ -193,127 +205,189 @@ export class UtopiaFSExtension
...
+
+  delete(uri: Uri, options: { recursive: boolean }) {
+    this.silentDelete(uri.path, options)
+    commands.executeCommand('utopia.toUtopiaMessage', vsCodeFileDelete(uri.path))
   }
 
-  async delete(uri: Uri, options: { recursive: boolean }): Promise<void> {
-    const path = fromUtopiaURI(uri)
-    await deletePath(path, options.recursive)
+  silentDelete(path: string, options: { recursive: boolean }) {
```
Rheeseyb, line 301: "silent" because it doesn't send the message to Utopia. It still emits the event. I couldn't think of a better name here tbh

vscode-build/vscode.patch
```diff
@@ -1,35 +1,29 @@
...
+diff --git a/build/gulpfile.compile.js b/build/gulpfile.compile.js
+index c4947e76cbf..e46583b9073 100644
+--- a/build/gulpfile.compile.js
++++ b/build/gulpfile.compile.js
+@@ -22,7 +22,7 @@ function makeCompileBuildTask(disableMangle) {
+ }
+ 
+ // Full compile, including nls and inline sources in sourcemaps, mangling, minification, for build
+-const compileBuildTask = task.define('compile-build', makeCompileBuildTask(false));
++const compileBuildTask = task.define('compile-build', makeCompileBuildTask(true));
```
Rheeseyb, line 10: The `true` here is to disable mangling. Minification still happens. For some reason the mangling was causing the build to completely hang on my machine.

vscode-build/vscode.patch
```diff
@@ -38,59 +32,27 @@ index 9e26dfeeb6e..00000000000
...
-     "extensions-ci": "node --max_old_space_size=4095 ./node_modules/gulp/bin/gulp.js extensions-ci"
-@@ -77,6 +77,7 @@
-     "spdlog": "^0.13.0",
-     "sudo-prompt": "9.2.1",
-     "tas-client-umd": "0.1.4",
+@@ -96,10 +96,10 @@
+     "kerberos": "^2.0.1",
+     "minimist": "^1.2.6",
+     "native-is-elevated": "0.7.0",
+-    "native-keymap": "^3.3.5",
```
Rheeseyb, line 43: I removed everything involving `native-keymap` as that was causing issues with the build involving Electron (which we obviously don't care about)

seanparsons (approved):

vscode-build/vscode.patch
Rheeseyb, line 1: You're best off view the raw file here rather than a diff of a diff. The original patch is fairly irrelevant

vscode-build/vscode.patch
```diff
@@ -867,228 +791,245 @@ index 7d8189c50be..125b6c89d0b 100644
...
++import { isFromUtopiaToVSCodeMessage, messageListenersReady } from 'utopia-vscode-common';
+ 
+ export interface IWorkbenchOptions {
+ 
+@@ -193,6 +195,37 @@ export class Workbench extends Layout {
+ 				this.restore(lifecycleService);
+ 			});
+ 
++			// Chain off of the previous one to ensure the ordering of changes is maintained.
++			// FIXME Do we still need this?
```
Rheeseyb, line 983: I don't think we still need this. IIRC the whole reason for this was to ensure writes to IndexedDB were applied in the correct order, and that required this setup because the writes would use a sequence of `await` steps (to check if the file existed, to check if the directory existed and create it if not). Now that's no longer an issue, I'm fairly certain we can remove the project changes coordination, so I'll remove it in a follow up PR

vscode-build/vscode.patch
```diff
@@ -1121,148 +1062,75 @@ index b9beba774a5..40627803528 100644
...
-+		"https://utopia95.com"
- 	];
+-				if (environmentService.isBuilt) {
+-					// Built time configuration (do NOT modify)
+-					bundledExtensions = [/*BUILD->INSERT_BUILTIN_EXTENSIONS*/];
+-				} else {
++				// if (environmentService.isBuilt) {
++				// 	// Built time configuration (do NOT modify)
++				// 	bundledExtensions = [/*BUILD->INSERT_BUILTIN_EXTENSIONS*/];
++				// } else {
```
Rheeseyb, line 1101: This annoying piece of code was preventing us from slipping our extension into the list of built in extensions

seanparsons: This comment and probably some of the others would be helpful if they were in the code.

Rheeseyb: Good shout, I'll add them

utopia-vscode-common/src/messages-to-utopia.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+}
+
+export function isClearLoadingScreen(messageData: unknown): messageData is ClearLoadingScreen {
+  return (
+    typeof messageData === 'object' && (messageData as any)?.['type'] === 'CLEAR_LOADING_SCREEN'
+  )
+}
+
+export type FromVSCodeToUtopiaMessage =
```
balazsbajorics, line 154: much better naming, thanks!

utopia-vscode-extension/package.json
```diff
@@ -14,15 +17,15 @@
   ],
   "browser": "./dist/browser/extension",
   "engines": {
-    "vscode": "^1.61.2"
+    "vscode": "^1.74.0"
```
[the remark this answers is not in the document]
balazsbajorics: oh I realize this is an `engines` field. WTF!

utopia-vscode-extension/src/in-mem-fs.ts
```diff
@@ -0,0 +1,400 @@
...
+    markModified(newParent)
+  }
+}
+
+export function deletePath(path: string, recursive: boolean) {
+  const parent = getParent(path)
+  const targets = targetsForOperation(path, recursive)
+
+  // Really this should fail if recursive isn't set to true when trying to delete a
+  // non-empty directory, but for some reason VSCode doesn't provide an error suitable for that
```
balazsbajorics, line 391: maybe throwing an error is the expected behavior?

balazsbajorics (approved):

balazsbajorics:
very nice!

utopia-vscode-extension/src/extension.ts
```diff
@@ -334,56 +308,110 @@ function getFullConfig(): UtopiaVSCodeConfig {
...
-  initMailbox(VSCodeInbox, parseToVSCodeMessage, handleMessage)
+function initMessaging(
+  context: vscode.ExtensionContext,
+  workspaceRootUri: vscode.Uri,
+  utopiaFS: UtopiaFSExtension,
+): void {
+  context.subscriptions.push(
+    vscode.commands.registerCommand(
+      'utopia.toVSCodeMessage',
+      (message: FromUtopiaToVSCodeMessage) => {
```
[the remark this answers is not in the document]
Rheeseyb: That's a valid point. I'll make that change in a follow up PR

utopia-vscode-extension/src/in-mem-fs.ts
Rheeseyb: It still needs to tell VS Code why it failed though, which is why this is awkward

PR #6225, opened 2024-08-12
Handle Remix Config Errors

### LINKED ISSUE #6182: feed node_modules to remix.config.js, handle errors properly
Our internal helper function `getRemixRootDir` needs to evaluate the remix.config.js, but it does so in a way that it doesn't provide node_modules to the evaluator:

https://github.com/concrete-utopia/utopia/blob/3c899113975efa02126c6f5f02639f89083ae812/editor/src/components/editor/store/remix-derived-data.tsx#L65-L72

This means that if the user's remix.config.js imports (or requires) anything from node_modules, that import will fail.

Repro case: take the existing example project remix.config.js, and try adding this line at the top:
`const withBuilderDevTools = require('@builder.io/dev-tools/remix')();`

Furthermore, if there is an error during the evaluation of remix.config.js, we should show a proper on-canvas error instead of a weird error on the console output.

**Fix 1:** We need to feed `EditorState.nodeModules` or `EditorState.codeResultCache.curriedRequireFn` to `getRemixRootDir`, and see if it solves the problem.

**Fix 2:** Add proper error handling around the evaluator call in getRemixRootDir, and if there is an error, run it through `processErrorWithSourceMap` and put it in `addToRuntimeErrors`

seanparsons:
**Problem:**
Currently we have a simplistic way of executing the Remix config file which doesn't give it access to the project or any of the dependencies, so if the file uses those it just fails.

**Fix:**
The main fix was to use some of our existing logic for executing arbitrary code, so that the Remix config is run as close to any other project code as possible.

Additionally if an error occurs while running the config code it is recorded and displayed the same way as any other error is displayed on the canvas error overlay.

**Commit Details:**
- Refactored out `buildBaseExecutionScope` from the canvas code, so that it can be used when executing the Remix config.
- `DerivedState.remixData` is now an instance of `FancyError` or the remix derived data.
- Rewrote `getRemixRootDir` to have it execute the code similar to how any other code would be run. So that it can require dependencies and do anything like that it might need to.
- Added `getDefaultedRemixRootDir` to help with handling some cases that don't care about the error.
- `useReadOnlyRuntimeErrors` now pulls in the error from `remixData` if there is one.
- A few small fixes to handle the new type of `remixData`.

**Manual Tests:**
I hereby swear that:

Fixes #6182

### DIFF
--- a/editor/src/components/navigator/left-pane/index.tsx
+++ b/editor/src/components/navigator/left-pane/index.tsx
@@ -22,6 +22,7 @@ import { when } from '../../../utils/react-conditionals'
 import { PagesPane } from './pages-pane'
 import { getRemixRootFile } from '../../canvas/remix/remix-utils'
 import { getRemixRootDir } from '../../editor/store/remix-derived-data'
+import { foldEither } from '../../../core/shared/either'
 
 export interface LeftPaneProps {
   editorState: EditorState
@@ -74,13 +75,22 @@ export const LeftPaneComponent = React.memo<LeftPaneComponentProps>((props) => {
 
   const isMyProject = useIsMyProject()
 
+  const curriedRequireFn = useEditorState(
+    Substores.restOfEditor,
+    (store) => store.editor.codeResultCache.curriedRequireFn,
+    'LeftPaneComponent curriedRequireFn',
+  )
+
   const isRemixProject = useEditorState(
     Substores.projectContents,
-    (store) =>
-      getRemixRootFile(
-        getRemixRootDir(store.editor.projectContents),
-        store.editor.projectContents,
-      ) != null,
+    (store) => {
+      const rootDir = getRemixRootDir(store.editor.projectContents, curriedRequireFn)
+      return foldEither(
+        () => false,
+        (dir) => getRemixRootFile(dir, store.editor.projectContents) != null,
+        rootDir,
+      )
+    },
     'LeftPaneComponent isRemixProject',
   )
   return (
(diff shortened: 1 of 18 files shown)


Rheeseyb (approved):

editor/src/components/navigator/left-pane/index.tsx
```diff
@@ -74,13 +75,22 @@ export const LeftPaneComponent = React.memo<LeftPaneComponentProps>((props) => {
...
+
   const isRemixProject = useEditorState(
     Substores.projectContents,
-    (store) =>
-      getRemixRootFile(
-        getRemixRootDir(store.editor.projectContents),
-        store.editor.projectContents,
-      ) != null,
+    (store) => {
+      const rootDir = getRemixRootDir(store.editor.projectContents, curriedRequireFn)
```
balazsbajorics, line 87: why not just use `store.editor.curriedRequireFn` here? and then you could delete the `const curriedRequireFn = useEditorState...` line

balazsbajorics (approved):
can I ask you to do a quick performance measurement of the before-after? I am curious if there's any noticeable change with the more involved evaluation and error handling

editor/src/components/navigator/left-pane/index.tsx
seanparsons: Because it's not part of that substore?

balazsbajorics: ah right. boo! how frequently does the curriedRequireFn change? it may be beneficial to change the substore type so we do not rerender the left-pane every time curriedRequireFn changes

seanparsons: We could conceivably disconnect it from changes to `curriedRequireFn`, but that definitely makes me a little antsy around it not updating when it might need to.

seanparsons:
> can I ask you to do a quick performance measurement of the before-after? I am curious if there's any noticeable change with the more involved evaluation and error handling

Seems like calling `getRemixRootDir` with this branch takes 0.16ms, with no obvious additional costs that I can see. Obviously if someone tries to calculate a million digits of pi in that file it'll be different, but that's kinda their fault.

PR #6233, opened 2024-08-15
Fix CDN usage in webWorkerExtensionHost

Rheeseyb:
**Problem:**
VS Code now uses an iframe for running the web workers used for hosting extensions. In staging we have VS Code use the CDN for loading everything by providing the CDN as part of the base URL that is used via its require function. That in turn meant that whilst VS Code was being run in an iframe with the regular utopia domain, the extensions host iframe was using cdn subdomain, which caused CORS issues.

**Fix:**
At the point where the iframe's URI is constructed we now strip the `cdn.` from it's authority.
Included in this PR are some changes to local dev to use `cdn.localhost` in all places where we would use the CDN subdomain, since it turns out you don't need to do anything to make that work. ~~Also I was in the process of tidying up some code after feedback from the previous PR, so I've included that rather than separating it into another PR.~~

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/utopia-vscode-extension/src/extension.ts
+++ b/utopia-vscode-extension/src/extension.ts
@@ -141,8 +141,25 @@ function didClose(path: string): DidClose {
 
 type SubscriptionWork = UpdateDirtyContentChange | DidChangeTextChange | WillSaveText | DidClose
 let pendingWork: Array<SubscriptionWork> = []
+let pendingWorkHandle: number
 
-function minimisePendingWork(): void {
+function pushWorkToQueue(work: SubscriptionWork) {
+  pendingWork.push(work)
+
+  if (pendingWorkHandle != null) {
+    clearTimeout(pendingWorkHandle)
+  }
+
+  pendingWorkHandle = setTimeout(() => {
+    const minimisedWork = minimisePendingWork()
+    pendingWork = []
+    for (const innerWork of minimisedWork) {
+      doSubscriptionWork(innerWork)
+    }
+  }, 5)
+}
+
+function minimisePendingWork(): Array<SubscriptionWork> {
   let newPendingWork: Array<SubscriptionWork> = []
   for (const workItem of pendingWork) {
     const latestWorkItem = newPendingWork[newPendingWork.length - 1]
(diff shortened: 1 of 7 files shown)


utopia-vscode-extension/src/extension.ts
```diff
@@ -141,8 +141,25 @@ function didClose(path: string): DidClose {
 
 type SubscriptionWork = UpdateDirtyContentChange | DidChangeTextChange | WillSaveText | DidClose
 let pendingWork: Array<SubscriptionWork> = []
+let pendingWorkHandle: number
 
-function minimisePendingWork(): void {
+function pushWorkToQueue(work: SubscriptionWork) {
```
Rheeseyb, line 146: This was to remove an unnecessary busy wait, which I brought up in the original PR

utopia-vscode-extension/src/extension.ts
```diff
@@ -316,6 +317,92 @@ function sendMessageToUtopia(message: FromVSCodeToUtopiaMessage): void {
   vscode.commands.executeCommand('utopia.toUtopiaMessage', message)
 }
 
+function handleFromUtopiaToVSCodeMessage(
```
Rheeseyb, line 320: This is just extracting an inner anonymous function into its own top level function, as requested in the previous PR

balazsbajorics (approved):

vscode-build/vscode.patch
```diff
@@ -1114,6 +1096,25 @@ index e1913359976..3834f2dde41 100644
...
+ 
++		// We want to use the cdn for loading all assets, but if we create this iframe using the cdn
++		// path then the cross origin settings prevent it from calling postmessage on its parent
+ 		const relativeExtensionHostIframeSrc = FileAccess.asBrowserUri(iframeModulePath);
+-		return `${relativeExtensionHostIframeSrc.toString(true)}${suffix}`;
++		const authority = relativeExtensionHostIframeSrc.authority;
++		const authorityWithoutCDN = authority.startsWith('cdn.') ? authority.slice(4) : authority;
++		const srcWithoutCDN = relativeExtensionHostIframeSrc.with({authority: authorityWithoutCDN});
++		return `${srcWithoutCDN.toString(true)}${suffix}`;
+ 	}
```
Rheeseyb, line 1115: This is the actual fix

PR #6239, opened 2024-08-19
Use Pointer Lock For Scrubbing

### LINKED ISSUE #6196: Label scrubbing should use pointerLock and wrap around the screen
**Problem**
Utopia input labels are scrubbable, but this scrubbing suffers from a number of issues:

- it is slow to update the inspector and canvas
- since the inspector is close to the edge of the screen, it sometimes means you run out of room (esp for large positive numbers like `left`)
- the mouse can trigger hover effects on the canvas or other inputs, which looks unprofessional

**Solution**
- Use the pointerLock API
- Draw a simulated mouse pointer
- wrap that pointer around the screen horizontally

**Existing work**
- for the performance, there's #6067 
- There's a draft PR: #6195 with severe performance issues (and some notes / todos) - perhaps it just exacerbates existing performance problems?
-  There's a working prototype implementation you can follow: https://github.com/maltenuhn/scrub-a-dub-dub/blob/master/src/DraggableRectangle.jsx

Image

seanparsons:
**Problem:**
Utopia input labels are scrubbable, but this scrubbing suffers from a number of issues:

- it is slow to update the inspector and canvas
- since the inspector is close to the edge of the screen, it sometimes means you run out of room (esp for large positive numbers like left)
- the mouse can trigger hover effects on the canvas or other inputs, which looks unprofessional

**Fix:**
The previous scrubbing implementation has been completely replaced in favour of an implementation which uses pointer locking to achieve the same effect. This has resulted in some slightly more complicated handling around tracking the accumulated movement.

Handling the "cleanup" of event handlers is also a little more complicated than before because it repeatedly tended towards callbacks that were circularly referenced.

**Commit Details:**
- Changed some const functions to regular functions.
- Removed the redundant y dragging for scrubbing as it wasn't used.
- Added some additional React refs to help keep track of what has occurred.
- Reworked the old scrub event handlers with the pointer lock based event handlers.

**Manual Tests:**
I hereby swear that:

Fixes #6196


editor/src/uuiui/inputs/number-input.tsx
```diff
@@ -389,53 +362,79 @@ export const NumberInput = React.memo<NumberInputProps>(
...
+              pointerOriginRef.current.requestPointerLock()
+              scrubbingCleanupCallbacks.current.push(() => {
+                window.removeEventListener('mouseup', scrubOnMouseUp)
+              })
+              scrubbingCleanupCallbacks.current.push(() => {
+                window.removeEventListener('mousemove', scrubOnMouseMove)
+              })
+              scrubbingCleanupCallbacks.current.push(() => {
+                document.removeEventListener('pointerlockchange', checkPointerLockChange, true)
+              })
```
Rheeseyb, line 432: Is there a reason to push these cleanups to a ref of an array like this? It seems to me that in both cleanup sites (`checkPointerLockChange` and `scrubOnMouseUp`) we could just call a single function that removes these 3 event listeners regardless of the circumstances

Rheeseyb (approved):

editor/src/uuiui/inputs/number-input.tsx
seanparsons: I referenced this in the description, the issue was that I kept ending up with circular references between the callbacks, as their dependencies need to be declared before the callbacks which use them. This means that the function which triggers the cleanup doesn't depend on the event listeners, which snips a convenient gap in the circle.

Rheeseyb: Ohhhh.... ew.... ok, can you add that in a comment here please too

seanparsons: Fixed in 8e2c64b2b.

ruggi (approved):
lgtm, but there's an issue that I'm not sure whether should be addressed in this PR or in an upcoming one (I'd say the latter tbh): if while scrubbing a value the cursor ends up over another scrubbable input, the scrubbing bails

Rheeseyb:
> lgtm, but there's an issue that I'm not sure whether should be addressed in this PR or in an upcoming one (I'd say the latter tbh): if while scrubbing a value the cursor ends up over another scrubbable input, the scrubbing bails

That to me sounds like it should be addressed before merging this PR, as I don't believe that's an issue with the existing implementation (though I may be wrong)

seanparsons:
> lgtm, but there's an issue that I'm not sure whether should be addressed in this PR or in an upcoming one (I'd say the latter tbh): if while scrubbing a value the cursor ends up over another scrubbable input, the scrubbing bails

Interesting, definitely should be addressed in this, I don't remember ever seeing this but I wonder if the browser is still tracking the real position and I've coincidentally dragged around the other fields.

seanparsons:
> lgtm, but there's an issue that I'm not sure whether should be addressed in this PR or in an upcoming one (I'd say the latter tbh): if while scrubbing a value the cursor ends up over another scrubbable input, the scrubbing bails

I've created a new bug for this as it's a pre-existing issue and can be triggered in other ways: https://github.com/concrete-utopia/utopia/issues/6249

PR #6240, opened 2024-08-20
Spike: dom sampler

balazsbajorics:
**Problem:**
The current dom-walker's code is hard to read and hard to maintain, and as a direct consequence of this, it is slow, does a lot of unnecessary things and makes changing the ElementInstanceMetadata type an undue burden.

The original dom-walker implemented a tree walk of the DOM using recursive functions. The recursive tree walk unto itself is hard to maintain, because VSCode can't spot unused parameters that are just passed around. It was also necessary because 4 years ago we didn't have the `data-path` printed into the DOM for each element, so the dom-walker itself had to assemble paths based on the `data-uid`s it found.
 
Then on top of this code we grafted various layers of optimizations: one layer attempts to cache results for each element while keeping the recursive tree walk. As part of this optimization the dom-walker heavily relies on its own previous results which need to be passed in.

The other problem is that the dom-walker is only half of the story, the other half is the SpyWrapper system, which uses a react higher-order-component wrapping every user element rendered by the Canvas. Originally we had to keep these two data structures apart because the canvas render was asynchronous, and we couldn't trust to run the dom-walker only when the spy metadata was ready. But this constraint has not been true for over two years now!

These two data sources are used to produce two data structures, which are individually memoized, and then the two memoized data structures are unified into a single structure in a costly step called `reconstructJSXMetadata`. 
`reconstructJSXMetadata` itself is hard to read and hard to maintain, so unsurprisingly it is also pretty wasteful performance-wise, and changing it is a MASSIVE pain in the ass every single time. Anecdotally, almost everyone spent at least a few weeks on adding new types of elements that required slightly different handling in the reconstruct function.

All in all, for a long time now the original constraints that made the dom-walker as-is no longer applied, but refactoring the dom-walker into something more manageable was always deemed a task too daunting.

**Fix:**
This PR introduces the `dom-sampler`. It tries to fix all of the above problems:
- it's no longer a tree walk! turns out, ahead of time we have an exhaustive array of ElementPaths of elements we want to collect metadata for, called the `validPaths`. And we already print an attribute called `data-path` for each dom element containint their ElementPath. Also turns out, using `querySelectorAll` to query all relevant elements that match a certain data-path is surprisingly quick! This means all of the recursive tree walking functions can be skipped! This also means that most of the hard-to-maintain code is gone.
- No more `reconstructJSXMetadata`! the dom-sampler returns a final JsxMetadata map, and the necessary merging of the spy and dom information sources are done internally during the processing of each element. There are a few important features that were implemented as part of `reconstructJSXMetadata`, namely `fillMissingDataFromAncestors`. This is now brought into the dom-sampler as a step of fixes that can be extended in the future. **Note**: this means that the dom-sampler uses the spyMetadata as one of its inputs! This is not an issue because we _synchronously_ render the canvas before we run the dom-sampler. This was one of the keys to deleting a lot of code.

**Notes:**
- There is a new `EditorStore.elementMetadata` which for now we only use to feed the dom-sampler with the previous results as a cache source. BUT in the future, ALL other metadata from EditorState should be deleted, further simplifying the update flow and the `dispatch.tsx` code
- I am yet to make conclusive measurements about this, but it looks like `getAttributesComingFromStyleSheets` is 1 - 1.5ms – this should not even be in the metadata, because we only use this information in the inspector and only for the selected element, and we only want to update it once every few seconds, yet FOR EACH FRAME we spend 1ms calculating it. (optimizations like this can now be much easier implemented because the new code is easier to change and measure)
- As I was reimplementing some of the features I realized we should delete the whole concept of `infinityCanvasRectangle`, it was a well meaning attempt to make the canvas appear infinite for the sake of the canvas controls, but in reality it makes a lot of frame handling unnecessarily complicated, and the canvas does have a size, although it can expand infinitely, but if an element on the canvas root sizes itself `left: 0, right: 0, top: 0, bottom: 0` it very much will have a size on the screen which is not infinite. 

**Befores and afters:**

A full re-walk of all elements, uncached:
dom-walker 67ms: 
image

dom-sampler 32ms:
image

The frame when a reparent happens (only re-measuring the potentially affected elements):
dom-walker 26ms:
image

dom-sampler 11.5ms:
image

Moving or resizing a single element (only re-measuring the one moved / resized element):
dom-walker 18ms:
image

 dom-sampler 6ms:
image


editor/src/components/canvas/canvas-utils.ts
```diff
@@ -1843,6 +1843,48 @@ function getValidElementPathsFromElement(
     const isScene = isSceneElement(element, filePath, projectContents)
     const isSceneWithOneChild = isScene && element.children.length === 1
 
+    const name = isJSXElement(element) ? getJSXElementNameAsString(element.name) : 'Fragment'
```
balazsbajorics, line 1846: this code is unchanged, just reordered

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -417,34 +424,45 @@ export async function renderTestEditorWithModel(
...
+          selectedViews: workingEditorState.patchedEditor.selectedViews,
+          spyCollector: spyCollector,
         },
-      })
+      )
+
+      // console.log(
+      //   'collect metadata called, spyCollector.current.spyValues.variablesInScope:',
+      //   spyCollector.current.spyValues.variablesInScope,
+      // )
```
gbalint, line 445: did you deliberately left this commented out console.log here?

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -417,34 +424,45 @@ export async function renderTestEditorWithModel(
...
 
-      if (domWalkerResult != null) {
-        const saveDomReportAction = saveDOMReport(
-          domWalkerResult.metadata,
-          domWalkerResult.cachedPaths,
-          domWalkerResult.invalidatedPaths,
+      const deepequalityResult = ElementInstanceMetadataMapKeepDeepEquality(
```
gbalint, line 447: camel case error `deepEqualityResult`

editor/src/core/shared/element-path.ts
```diff
@@ -1203,3 +1203,33 @@ export function isIndexedElement(path: ElementPath): boolean {
 export function getFirstPath(pathArray: Array<ElementPath>): ElementPath {
   return forceNotNull('Element path array is empty.', pathArray.at(0))
 }
+
+export function humanReadableDebugPath(path: ElementPath): string {
```
gbalint, line 1207: Nice!

gbalint:
Did you deliberately leave all the unused old dom walker code there?

editor/src/components/canvas/dom-sampler.tsx
```diff
@@ -0,0 +1,449 @@
...
+    // if there are no spy elements, but we have valid paths, we probably encountered a transient case during a canvas runtime error
+    // the intended behavior of the editor during a runtime error is to preserve the existing metadata to avoid flickering
+    // so we return the existing metadata in that case
+    return {
+      metadata: options.metadataToUpdate,
+      tree: MetadataUtils.createElementPathTreeFromMetadata(options.metadataToUpdate), // TODO this should return a cached tree
+    }
+  }
+
+  let result
```
gbalint, line 75: missing type

balazsbajorics:
> Did you deliberately leave all the unused old dom walker code there?

Good question!
Yes!
I didn't want to bloat this PR, but right after this is merged, the dom-walker and reconstructJSXMetadata can go straight in the bin

gbalint (approved):

editor/src/components/canvas/dom-sampler.tsx
```diff
@@ -0,0 +1,455 @@
...
+import { CanvasContainerID } from './canvas-types'
+import {
+  collectDomElementMetadataForElement,
+  getAttributesComingFromStyleSheets,
+} from './dom-walker'
+import type { UiJsxCanvasContextData } from './ui-jsx-canvas'
+
+export function runDomSampler(
+  elementsToFocusOn: ElementsToRerender,
+  options: {
```
seanparsons, line 47: Why wrap these in an object?

editor/src/components/canvas/dom-sampler.tsx
```diff
@@ -0,0 +1,455 @@
...
+  tree: ElementPathTrees
+} {
+  const containerRect = getCanvasRectangleFromElement(
+    canvasRootContainer,
+    options.scale,
+    'without-text-content',
+    'nearest-half',
+  )
+
+  const dynamicPathsToCollect: Array<ElementPath> = pathsToCollect
```
seanparsons, line 124: You can probably save on building this array by doing the `forEach` on `pathsToCollect` instead.

editor/src/components/canvas/dom-sampler.tsx
```diff
@@ -0,0 +1,455 @@
...
+    infinityCanvasRectangle,
+    emptySpecialSizeMeasurements,
+    null,
+  )
+}
+
+function getValidPathsFromCanvasContainer(canvasRootContainer: HTMLElement): Array<ElementPath> {
+  const validPaths: Array<ElementPath> | null = optionalMap(
+    (paths) => paths.split(' ').map(EP.fromString),
+    canvasRootContainer.getAttribute('data-utopia-valid-paths'),
```
seanparsons, line 378: Huh, we don't have a constant for this like a lot of the others.

editor/src/core/shared/element-template.ts
```diff
@@ -2521,6 +2521,22 @@ export interface ElementInstanceMetadata {
   assignedToProp: string | null
 }
 
+export interface DomElementMetadata {
+  element: Either<string, JSXElementChild>
+  globalFrame: MaybeInfinityCanvasRectangle | null
+  nonRoundedGlobalFrame: MaybeInfinityCanvasRectangle | null
+  specialSizeMeasurements: SpecialSizeMeasurements
+  textContent: string | null
+
```
seanparsons, line 2530: Intentional gap?

editor/src/core/shared/element-template.ts
```diff
@@ -2521,6 +2521,22 @@ export interface ElementInstanceMetadata {
...
+  computedStyle: ComputedStyle | null
+  attributeMetadatada: StyleAttributeMetadata | null // sic fix typo
```
seanparsons, line 2532: Rename refactor surely faster than adding the comment? :)

seanparsons (approved):
Nice work!

editor/src/core/shared/element-template.ts
balazsbajorics: maybe in a separate PR, the diff is _massive_

editor/src/components/canvas/dom-sampler.tsx
balazsbajorics: fixed, now all parameters are in the object. in my opinion this makes future refactors easier, and makes the caller side more readable. makes it harder to forget that you missed passing some important parameter. Case in point: I just added `domWalkerAdditionalElementsToFocusOn` and the atomic commit to do that has a very readable diff, and it made it easier for me to see where I am still yet to pass it

editor/src/components/canvas/dom-sampler.tsx
balazsbajorics: fixed this too, now all params are in an object

editor/src/components/canvas/dom-sampler.tsx
```diff
@@ -0,0 +1,455 @@
...
+
+  return result
+}
+
+function collectMetadataForPaths(
+  canvasRootContainer: HTMLElement,
+  pathsToCollect: Array<ElementPath>,
+  validPaths: Array<ElementPath>,
+  metadataToUpdate_MUTATE: ElementInstanceMetadataMap,
+  options: {
```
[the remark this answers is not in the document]
balazsbajorics: fixed this too, all params are now in the object

Rheeseyb:
> As I was reimplementing some of the features I realized we should delete the whole concept of infinityCanvasRectangle, it was a well meaning attempt to make the canvas appear infinite for the sake of the canvas controls, but in reality it makes a lot of frame handling unnecessarily complicated, and the canvas does have a size, although it can expand infinitely, but if an element on the canvas root sizes itself left: 0, right: 0, top: 0, bottom: 0 it very much will have a size on the screen which is not infinite.

Careful now. The canvas being zero sized caused a lot of runtime issues that were fixed by that awful code

editor/src/components/canvas/dom-sampler.spec.browser2.tsx
```diff
@@ -0,0 +1,450 @@
...
+        "sb/pg-sc/pg:aaa/bbb/266",
+        "sb/pg-sc/pg:aaa/bbb/266/ccc~~~1",
+        "sb/pg-sc/pg:aaa/bbb/266/ccc~~~2",
+        "sb/pg-sc/pg:aaa/bbb/266/ccc~~~3",
+        "sb/pg-sc/pg:aaa/bbb/266/ccc~~~1/ddd",
+        "sb/pg-sc/pg:aaa/bbb/266/ccc~~~2/ddd",
+        "sb/pg-sc/pg:aaa/bbb/266/ccc~~~3/ddd",
+        "sb/pg-sc/pg:aaa/bbb/6cc",
+      ]
+    `, // TODO before merge isn't the fragment children missing from here?
```
Rheeseyb, line 105: This seems like a significant thing to address before merging

editor/src/components/canvas/dom-sampler.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+    if (domMetadata == null) {
+      // if we couldn't find any dom elements for the path, we must scan through all the spy elements to find a fallback with a potentially dynamic path
+      const spyElem = spyCollector.current.spyValues.metadata[EP.toString(path)]
+      if (spyElem != null) {
+        metadataToUpdate_MUTATE[EP.toString(path)] = {
+          ...spyElem,
+        }
+      }
+      return // we couldn't find a fallback spy element, so we bail out
```
Rheeseyb, line 168: Shouldn't this be wrapped in an `else` ?

editor/src/components/canvas/dom-sampler.tsx
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    const validDynamicPath = path
+    const spyMetadata = spyCollector.current.spyValues.metadata[EP.toString(validDynamicPath)]
```
Rheeseyb, line 172: The above could be simplified by moving this and the associated bailing before the above `if (domMetadata == null)` since you're doing the same check inside that

editor/src/components/canvas/ui-jsx-canvas.tsx
```diff
@@ -537,7 +537,7 @@ export const UiJsxCanvas = React.memo<UiJsxCanvasPropsWithErrorCallback>((props)
     props.focusedElementPath,
     topLevelElementsMap,
     executionScope,
-    projectContentsForRequireFn,
+    projectContents,
```
Rheeseyb, line 540: 👀  why has this changed?

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -417,34 +424,40 @@ export async function renderTestEditorWithModel(
...
+          metadataResult.tree,
         )
         recordedActions.push(saveDomReportAction)
         const editorWithNewMetadata = editorDispatchActionRunner(
           asyncTestDispatch,
           [saveDomReportAction],
           workingEditorState,
           spyCollector,
-          domWalkerResult.reconstructedMetadata,
+          {}, // TODO delete before merge
```
Rheeseyb, line 460: 👀

editor/src/core/shared/element-path.ts
```diff
@@ -644,6 +644,7 @@ export function findAmongAncestorsOfPath<T>(
 export function isDescendantOf(target: ElementPath, maybeAncestor: ElementPath): boolean {
   const targetString = toString(target)
   const maybeAncestorString = toString(maybeAncestor)
+  // TODO FIX OMG this will return true for aaa/bbb vs aaa/bbb-1, which is a problem for user-written UIDs, especially common in tests
```
Rheeseyb, line 647: 👀

Rheeseyb (commented):
Very impressive work!

PR #6241, opened 2024-08-20
Chore: data-static-path

balazsbajorics:
**Context:**
```typescript
<div data-uid='a'>
  {[1, 2, 3].map(() => (<div data-uid='b'><div data-uid='c' /></div>)}
</div>
```

This code snippet will render three of the `<div data-uid='b' />` elements. If you look at the rendered elements, their resulting DOM will be:
```
<div data-path='a'>
  <div data-path='a/b~~~1'><div data-path='a/b~~~1/c' /></div>
  <div data-path='a/b~~~2'><div data-path='a/b~~~2/c' /></div>
  <div data-path='a/b~~~3'><div data-path='a/b~~~3/c' /></div>
</div>
```

and so the resulting paths will be
```
a
a/b~~~1
a/b~~~1/c
a/b~~~2
a/b~~~2/c
a/b~~~3
a/b~~~3/c
```

Paths containing the `~~~1` are what we call _dynamic_ ElementPaths. 
Because it would be very hard for Utopia to predict the number of indexes using static analysis, we have the concept of a _static_ path: any dynamic path can be converted to a static path by stripping the indexes. The parts of Utopia which generate paths based on static analysis use the `StaticElementPath`. 
The indexed element path's indexes are assigned at runtime and the information only exists printed to the DOM data-path attribute. 

The flow of data looks like this:
parsed model -> canvas evaluation -> indexed paths assigned -> dom-walker reads these paths -> saved to JsxMetadata.

Other places of the editor state, such as the `validPaths` data use the StaticElementPaths. So when the dom-walker determines if a path is valid or not, we take the dynamic path, and convert it to static (strip the indexes) and compare _that_ to the array of valid paths.

**Problem:**
The upcoming reworked dom-walker will work the other way around: take an array of validPaths, and use document.querySelector to find a corresponding element for which to store metadata. This means the new dom-sampler will need to be able to find elements based on their static path, using querySelector.

but querySelector does not support regexes (which is probably for the best, because then we would have to take the performance hit of regexes). The main relevant feature of querySelector is that it can match prefixes. It accidentally works to match `a/b` to `a/b~~~1`, but it stops working to match `a/b/c` to `a/b~~~1/c`.

**Solution:**
The most straightforward solution is to add a `data-static-path` attribute to the elements rendered by utopia, which is a sibling to `data-path`. If `data-path` is already static, then the contents of `data-path` and `data-static-path` will match. If `data-path` has a ~~~1 indexed element, those will be stripped. 

This way we can querySelector data-static-path for `data-static-path='a/b/c'` and get a hit for `data-path='a/b~~~1/c'`


editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -65,7 +65,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`749`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`880`)
```
Rheeseyb, line 68: Errrr what's happened here?!

Rheeseyb (approved):

seanparsons (approved):

balazsbajorics:
closing this in favor of a solution using the spyMetadata keys

PR #6248, opened 2024-08-21
Check that a valid GITHUB_TOKEN is supplied when running start-full

Rheeseyb:
**Problem:**
The VS Code build can fail due to GitHub rate limiting if the user hasn't supplied a `GITHUB_TOKEN`, or if they have but it has expired. When this happens, the failure message is very obscure, leaving the user unclear how to progress.

**Fix:**
Previously we checked for the existence of the token, but that alone isn't enough, since the an expired token will also cause the same unclear failure case. Instead, we now perform a request against the GH API using the token, and check that a 200 response code is received. If no token was provided, or if the token has expired, a 40X response will be received, in which case the process bails early and prints a helpful message:

```
$ GITHUB_TOKEN=dirtman start-full                    
pg_ctl: PID file "utopia-db/postmaster.pid" does not exist
Is server running?
no server running on /private/tmp/tmux-501/default
A valid GITHUB_TOKEN is required when running the full Utopia build.
You may be missing one, or it may have expired.
Please see the readme for instructions.
```

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -289,6 +289,18 @@ let
       cd $(${pkgs.git}/bin/git rev-parse --show-toplevel)/vscode-build
       nix-shell --run pull-extension-inner
     '')
+    (pkgs.writeScriptBin "check-github-token" ''
+      #!/usr/bin/env bash
+      set -e
+      GITHUB_TOKEN_STATUS=$(curl -s -o /dev/null -w "%{http_code}" --request GET --url "https://api.github.com/octocat" --header "Authorization: Bearer $GITHUB_TOKEN" --header "X-GitHub-Api-Version: 2022-11-28")
+      if [ "$GITHUB_TOKEN_STATUS" = "200" ]
+      then
+        echo "A valid GITHUB_TOKEN was found"
+      else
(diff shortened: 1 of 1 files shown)


liady (approved):
awesome!

balazsbajorics (approved):
```
           MMM.           .MMM
           MMMMMMMMMMMMMMMMMMM
           MMMMMMMMMMMMMMMMMMM      ___________________________________
          MMMMMMMMMMMMMMMMMMMMM    |                                   |
         MMMMMMMMMMMMMMMMMMMMMMM   |  :(  |
        MMMMMMMMMMMMMMMMMMMMMMMM   |_   _______________________________|
        MMMM::- -:::::::- -::MMMM    |/
         MM~:~   ~:::::~   ~:~MM
    .. MMMMM::. .:::+:::. .::MMMMM ..
          .MM::::: ._. :::::MM.
             MMMM;:::::;MMMM
      -MM        MMMMMMM
      ^  M+     MMMMMMMMM
          MMMMMMM MM MM MM
               MM MM MM MM
               MM MM MM MM
            .~~MM~MM~MM~MM~~.
         ~~~~MM:~MM~~~MM~:MM~~~~
        ~~~~~~==~==~~~==~==~~~~~~
         ~~~~~~==~==~==~==~~~~~~
             :~==~==~==~==~~
```

ruggi (approved):

PR #6256, opened 2024-08-23
Remove ElementInstanceMetadata.localFrame.

seanparsons:
**Problem:**
It would appear that we don't _really_ need `ElementInstanceMetadata.localFrame` and can instead easily calculate it.

**Fix:**
The supposition that we don't need `localFrame` was actually true, it can instead be calculated comparatively simply from the `globalFrame` of the target and that of one of the ancestors. Originally it was thought that the latter would be the parent and in most cases it is, however if the target is the root component of an instance, then the parent will have an identical value for `globalFrame` and instead we need the grandparent of the target.

**Commit Details:**
- Removed the field `localFrame` from `ElementInstanceMetadata` and also from the relevant factory function.
- Included the whole editor metadata in the metadata that `pressKeys` creates so that it's possible to obtain the element metadata for the parent.
- Replaced lookups of `localFrame` with calls to `MetadataUtils.getFrame`.
- Implemented `MetadataUtils.getFrame` to lookup the global frames for an element and its parent (mostly) and to then calculate the local frame from their difference.

**Manual Tests:**
I hereby swear that:


editor/src/components/editor/store/editor-update.tsx
```diff
@@ -30,6 +30,7 @@ import type { Optic } from '../../../core/shared/optics/optics'
 import { fromField } from '../../../core/shared/optics/optic-creators'
 import { modify } from '../../../core/shared/optics/optic-utilities'
 import { ProjectServerStateKeepDeepEquality } from './store-deep-equality-instances'
+import { getContentsTreeFileFromElements } from '../../../components/assets'
```
ruggi, line 33:
```suggestion
```

ruggi (approved):

editor/src/components/editor/store/editor-update.tsx
seanparsons: Included in 40a456834.

editor/src/components/canvas/canvas-strategies/strategies/shared-move-strategies-helpers.ts
```diff
@@ -274,7 +274,7 @@ export function getMoveCommandsForSelectedElement(
     MetadataUtils.getFrameInCanvasCoords(selectedElement, startingMetadata),
   )
 
-  const localFrame = nullIfInfinity(MetadataUtils.getFrame(selectedElement, startingMetadata))
```
balazsbajorics, line 277: this is odd! was there a getFrame on master?

balazsbajorics: aa I see there was, and it returned the localFrame. makes sense

seanparsons: Ah yes, that's why it was named the same in this branch.

balazsbajorics (approved):

PR #6265, opened 2024-08-27
Polish the grid column/row resize controls

bkrmendy:
## Description
This PR removes some visual glitches from the grid column/row resize controls:
- when the resize handle is being dragged, only the grid column/row is highlighted, not the grid column/row + the handle
- when a handle is hovered, elements under the handle aren't highlighted
- handle shadows are properly scaled

### Manual Tests:
I hereby swear that:


liady (approved):

ruggi (approved):
🏅

PR #6273, opened 2024-08-29
Keep grid rearrange shadow inside container

### LINKED ISSUE #6272: keep grid rearrange shadow inside container

ruggi:
**Problem:**

When rearranging grid elements, the shadow can go anywhere the mouse goes, making it seem like an absolute move.

**Fix:**

Wrap the shadow coords so they are restricted to the grid container bounds.

| Before | After |
|----------|--------------|
| Kapture 2024-08-29 at 13 11 21 | Kapture 2024-08-29 at 13 10 43 |

Fixes #6272


seanparsons (approved):

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -722,8 +738,34 @@ export const GridControls = controlForStrategyMemoized<GridControlsProps>(({ tar
       }
     }
 
-    return { x: getCoord('x', 'width'), y: getCoord('y', 'height') }
-  }, [features, initialShadowFrame, interactionData, shadow, hoveringStart, mouseCanvasPosition])
+    // make sure the shadow is displayed only inside the grid container bounds
+    function wrapCoord(c: number, min: number, max: number, shadowSize: number) {
+      return Math.min(Math.max(c, min), max - shadowSize)
```
liady, line 743: nice

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -618,6 +620,20 @@ export const GridControls = controlForStrategyMemoized<GridControlsProps>(({ tar
 
   const gridPath = optionalMap(EP.parentPath, shadow?.elementPath)
 
+  const gridFrame = React.useMemo(() => {
```
liady, line 623: we can maybe take padding here into consideration (so that the shadow won't get over the paddings)

liady (approved):

editor/src/components/canvas/controls/grid-controls.tsx
ruggi: true, I tried it when writing the code and it makes sense too – I just did not know which was best tbh, so I just went with the minimum option, but I guess it's easy to change if we find that it would feel better being "stuck to the inner grid" 🤔

PR #6274, opened 2024-08-29
Improve Canvas Context Menu

seanparsons:
**Problem:**
`getNameAndIconResult` runs fairly regularly and consumes 1ms at least of runtime during interactions.

**Fix:**
`getNameAndIconResult` is the basis of the canvas context menu entries, which means the result of it and the selector that calls it are not necessary until the context menu is displayed. That logic was refactored to be call on demand, only when the context menu is displayed does the lookup get invoked.

Alongside this, some minor performance improvements were made to some of the logic that underpins the menu items lookup.

In the process of doing this, an issue was found where the tests for the context menus would fail because the portal element didn't exist when the logic for obtaining a reference to it ran. So the lookup fails as the test logic doesn't re-render the editor a subsequent time. The editor proper doesn't have this issue because the portal element is part of the hardcoded HTML of the editor page itself and so exists before React has even warmed up. Fixing this for the `react-contextify` menus involved targeting a container element and having it be rendered before the main editor render. This broke the radix components as the portal for that ended up in a different part of the document hierarchy, so a similar fix was then applied for those.

**Commit Details:**
- Refactored `namesAndIconsAllPathsResultSelector` into `getNamesAndIconsAllPaths` so that it can be invoked as required.
- Support added to `ContextMenu` for the items to be defined with a function.
- `ContextMenu` keeps track of its own visibility so that the items to be displayed are not constructed when the menu is not visible.
- `useCanvasContextMenuItems` refactored into `getCanvasContextMenuItems`.
- `ElementContextMenu` now creates a callback for `ContextMenu` to invoke.
- Fixed render tests so that the context menu portals are instantiated upfront similar to the way they are in the editor proper.
- Improved performance of `isButtonFromMetadata`.
- Improved performance of `firstLetterIsLowerCase` as it is the basis of `isIntrinsicElement`.

**Manual Tests:**
I hereby swear that:


editor/src/core/shared/string-utils.ts
```diff
@@ -3,7 +3,8 @@ export function capitalize(value: string): string {
 }
 
 export function firstLetterIsLowerCase(value: string): boolean {
-  return value.length > 0 && isLowerCase(value[0])
+  const charCode = value.charCodeAt(0)
+  return !isNaN(charCode) && charCode >= 97 && charCode <= 122
```
ruggi, line 7: just for clarity would it be possible to use `'a'` and `'z'` instead of their numeric charcodes here? maybe something like `'a'.charCodeAt(0)` and put it into a const 🤔

editor/src/uuiui/radix-components.tsx
```diff
@@ -10,7 +10,9 @@ import { colorTheme, UtopiaStyles, UtopiaTheme } from './styles/theme'
 import { Icons, SmallerIcons } from './icons'
 import { when } from '../utils/react-conditionals'
 import { Icn, type IcnProps } from './icn'
-import { borderRadius } from './styles/utopitrons'
+import { forceNotNull } from '../core/shared/optional-utils'
+
+export const RadixComponentsPortalId = 'radix-components-portal'
```
ruggi, line 15: Maybe worth adding a comment here (on in the `.html` file?) to make sure if one is changed the other one changes too?

editor/src/components/element-context-menu.tsx
```diff
@@ -51,6 +51,13 @@ import { useDispatch } from './editor/store/dispatch-context'
 import { useCreateCallbackToShowComponentPicker } from './navigator/navigator-item/component-picker-context-menu'
 import { navigatorTargetsSelector, useGetNavigatorTargets } from './navigator/navigator-utils'
 import { foldEither } from '../core/shared/either'
+import type { AllElementProps, EditorStorePatched } from './editor/store/editor-state'
+import type { ElementInstanceMetadataMap } from '../core/shared/element-template'
+import type { ElementPathTrees } from '../core/shared/element-path-tree'
+import type { FilePathMappings } from '../core/model/project-file-utils'
+import type { PropertyControlsInfo } from './custom-code/code-file'
+import type { ProjectContentTreeRoot } from './assets'
+import { use } from 'chai'
```
ruggi, line 60:
```suggestion
```

ruggi (approved):

editor/src/core/shared/string-utils.ts
seanparsons: Yeah, that's a good idea.

editor/src/components/canvas/ui-jsx.test-utils.tsx
```diff
@@ -609,6 +610,20 @@ export async function renderTestEditorWithModel(
 
   let numberOfCommits = 0
 
+  // This results in the portal element existing before the subsequent render,
+  // which is necessary as any component that attempts to look for the portal as part of that render
+  // will not find it as it is not yet in the DOM.
+  const renderTargetDiv = document.createElement('div')
+  document.body.appendChild(renderTargetDiv)
+  const firstRender = render(
```
bkrmendy, line 618: this is unused

editor/src/components/inspector/common/name-and-icon-hook.ts
```diff
@@ -8,25 +8,16 @@ import type { ElementPath } from '../../../core/shared/project-file-types'
...
-import type {
-  CanvasSubstate,
-  MetadataSubstate,
-  ProjectContentSubstate,
-} from '../../editor/store/store-hook-substore-types'
 import type { ElementPathTrees } from '../../../core/shared/element-path-tree'
 import type { FilePathMappings } from '../../../core/model/project-file-utils'
 import type { PropertyControlsInfo } from '../../custom-code/code-file'
 import type { ProjectContentTreeRoot } from '../../assets'
+import * as EP from '../../../core/shared/element-path'
```
bkrmendy, line 20: This import is unused

editor/src/uuiui/radix-components.tsx
seanparsons: Fixed in 045fe83f1.

editor/src/core/shared/string-utils.ts
seanparsons: Fixed in 045fe83f1.

editor/src/components/inspector/common/name-and-icon-hook.ts
seanparsons: Fixed in 045fe83f1.

editor/src/components/canvas/ui-jsx.test-utils.tsx
seanparsons: Fixed in 045fe83f1.

bkrmendy (approved):

liady (approved):

PR #6276, opened 2024-08-29
feat(grid): show gaps immediately on handle hover

liady:
This PR shows the grid gap background:

1. when hovering the the grid gap handle,
-or-
2. when hovering the gap itself for more than 1500ms. this is due to the fact that sometimes the gaps are pretty large, so the handle might get "lost" and it's not easy to find and hover it

<video src="https://github.com/user-attachments/assets/e3223e40-70a2-460c-a5a2-a30602670da3"></video>


seanparsons (approved):

Rheeseyb (approved):
The timeout is maybe a bit long, but that's extremely minor

PR #6286, opened 2024-09-02
Cache Or Avoid Normalisation Of Paths

seanparsons:
**Problem:**
The function `normalisePathToUnderlyingTarget` is called quite a lot transitively by the strategies logic and often with the same values. But in general this function is triggered by pretty much anything the user does in the editor.

**Fix:**
Given that `normalisePathToUnderlyingTarget` relies on the project contents and an element path and the former changes in a stepwise fashion, we can cache based on that value so that when it changes we clear the cache with the cache being indexed by the element path. A utility function `valueDependentCache` was created which encapsulates this logic.

`valueDependentCache` was instrumented temporarily with a cache hit indicator which showed it hovering around 90% of calls being cache hits during a canvas interaction. When no changes were made to the project (like when the mouse is moved over the canvas) it rapidly shot to above 97%.

**Commit Details:**
- `deleteConflictingPropsForWidthHeight` now avoids calling `deleteValuesAtPath` if there is nothing to delete.
- Removed the null from the `elementPath` parameter of `normalisePathToUnderlyingTarget`.
- Added `valueDependentCache` utility function.
- Wrapped the original implementation of `normalisePathToUnderlyingTarget` with `valueDependentCache`

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/custom-code/code-file.spec.ts
+++ b/editor/src/components/custom-code/code-file.spec.ts
@@ -480,10 +480,6 @@ describe('normalisePathToUnderlyingTarget', () => {
     )
     expect(actualResult).toEqual(expectedResult)
   })
-  it('gives an error when the element path is empty', () => {
-    const actualResult = normalisePathToUnderlyingTarget(projectContents, null)
-    expect(actualResult.type).toEqual('NORMALISE_PATH_ERROR')
-  })
   it('flags elements that can not be found', () => {
     const actualResult = normalisePathToUnderlyingTarget(
       projectContents,
--- a/editor/src/components/editor/store/editor-state.ts
+++ b/editor/src/components/editor/store/editor-state.ts
@@ -3582,6 +3582,9 @@ export function modifyUnderlyingTarget(
     underlyingFilePath: string,
   ) => ParseSuccess = defaultModifyParseSuccess,
 ): EditorState {
+  if (target == null) {
+    throw new Error(`Target is null.`)
+  }
   const underlyingTarget = normalisePathToUnderlyingTarget(editor.projectContents, target)
   const targetSuccess = normalisePathSuccessOrThrowError(underlyingTarget)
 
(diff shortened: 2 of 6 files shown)


gbalint (approved):

editor/src/core/shared/memoize.ts
```diff
@@ -23,3 +23,30 @@ export function memoize<T extends Moizeable>(func: T, options?: Partial<Options<
   const memoizeOptions = getMemoizeOptions(options)
   return moize(func, memoizeOptions)
 }
+
+export function valueDependentCache<Value, Input, Result>(
+  fallback: (value: Value, input: Input) => Result,
+  inputToString: (input: Input) => string,
+): (value: Value, input: Input) => Result {
+  let cache: { [key: string]: Result } = {}
```
liady, line 31: maybe consider a `Map` here with arbitrary keys? (and instead the `inputToString` we can have an equality function)

liady (approved):

editor/src/components/editor/store/editor-state.ts
```diff
@@ -3582,6 +3582,9 @@ export function modifyUnderlyingTarget(
     underlyingFilePath: string,
   ) => ParseSuccess = defaultModifyParseSuccess,
 ): EditorState {
+  if (target == null) {
```
Rheeseyb, line 3585: Surely the function signature should be updated to remove the `| null` then?!

editor/src/components/custom-code/code-file.spec.ts
```diff
@@ -480,10 +480,6 @@ describe('normalisePathToUnderlyingTarget', () => {
     )
     expect(actualResult).toEqual(expectedResult)
   })
-  it('gives an error when the element path is empty', () => {
```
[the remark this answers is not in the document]
seanparsons: Urgh, I didn't look at the test description...

editor/src/components/editor/store/editor-state.ts
seanparsons: I somewhat assumed doing that would be the proverbial pulling on a thread, turns out nothing passes null for that parameter. Fixed in 380a604af.

editor/src/components/custom-code/code-file.spec.ts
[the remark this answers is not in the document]
seanparsons: Fixed in 380a604af.

editor/src/components/editor/store/editor-state.ts
Rheeseyb: Beautiful!

Rheeseyb (approved):

PR #6302, opened 2024-09-03
feat(grid): controls on inspector hover

liady:
This PR displays subdued Grid controls on inspector hover, according to the axis

<video src="https://github.com/user-attachments/assets/6db0c006-f4ba-44d1-a025-99fffe719a78"></video>

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/controls/select-mode/subdued-grid-gap-controls.tsx
```diff
@@ -0,0 +1,135 @@
...
+    1,
+  )
+
+  return (
+    <>
+      {controlBounds.gaps
+        .filter((gap) => gap.axis === axis || axis === 'both')
+        .map((gap, i) => (
+          <GridGapControl
+            key={i}
```
gbalint, line 55: Using the index as a React key is discouraged, because it doesn't serve the purpose to identify an element even when it is returned on a different index.
This may not be practically relevant in this case if you always have the same gap controls, but I would still choose meaningful identifiers and not the index itself.

editor/src/components/canvas/controls/select-mode/subdued-grid-gap-controls.tsx
```diff
@@ -0,0 +1,135 @@
...
+  if (gridGap == null) {
+    return null
+  }
+
+  const gridGapRow = gridGap.row
+  const gridGapColumn = gridGap.column
+
+  const controlBounds = gridGapControlBoundsFromMetadata(
+    selectedElement,
+    gridRowColumnInfo[0],
```
seanparsons, line 41: Unsafe array indexing, it would appear in this case that it would break `gridGapControlBoundsFromMetadata` as that doesn't allow undefined to be passed to this parameter:
```suggestion
    gridRowColumnInfo.at(0),
```

editor/src/components/canvas/controls/select-mode/subdued-grid-gap-controls.tsx
```diff
@@ -0,0 +1,135 @@
...
+      row: fallbackEmptyValue(gridGapRow),
+      column: fallbackEmptyValue(gridGapColumn),
+    },
+    1,
+  )
+
+  return (
+    <>
+      {controlBounds.gaps
+        .filter((gap) => gap.axis === axis || axis === 'both')
```
seanparsons, line 52: This looks like it should be wrapped up in a `React.useMemo` as `gridGapControlBoundsFromMetadata` generates new values each time.

seanparsons (approved):

Rheeseyb (approved):
Approved with the proviso that Sean's comments are addressed 👍

editor/src/components/canvas/controls/select-mode/subdued-grid-gap-controls.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const metadata = useRefEditorState((store) => store.editor.jsxMetadata)
+  const gridRowColumnInfo = useGridData([selectedElement])
+
+  const sideRef = useBoundingBox([selectedElement], (ref, parentBoundingBox) => {
+    const gridGap = maybeGridGapData(metadata.current, selectedElement)
+    if (gridGap == null) {
+      return
+    }
+
+    const controlBounds = gridGapControlBoundsFromMetadata(
```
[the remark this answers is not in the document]
liady: this actually happens after a gap update (where `SubduedGridGapControl` itself does not re-render, only this code runs again)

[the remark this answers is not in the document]
seanparsons: Fine by me!

PR #6309, opened 2024-09-04
fix(editor): remove error messages when deleting a file

### LINKED ISSUE #6300: Deleting a file with a syntax error leaves the canvas permanently broken
We should clear errors and re-evaluate the project when a file is deleted

liady:
**Problem:**
We currently clear errors only when a descriptor file is deleted. This can cause syntax errors to remain even after the file that they're in was deleted.

**Fix:**
Clear errors on every file delete

**Manual Tests:**
I hereby swear that:

Fixes #6300

### DIFF
--- a/editor/src/components/editor/actions/actions.tsx
+++ b/editor/src/components/editor/actions/actions.tsx
@@ -4222,19 +4222,19 @@ export const UPDATE_FNS = {
         }, updatedEditor)
       }
       case 'TEXT_FILE': {
-        const nextEditor = {
+        let nextEditor = {
           ...editor,
           projectContents: updatedProjectContents,
         }
         if (isComponentDescriptorFile(action.filename)) {
-          const withUpdatedPropertyControls = {
+          // update property controls
+          nextEditor = {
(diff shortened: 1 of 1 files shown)


gbalint (approved):

seanparsons (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -4222,21 +4222,21 @@ export const UPDATE_FNS = {
...
-          const withUpdatedPropertyControls = {
+          // update property controls
+          nextEditor = {
             ...nextEditor,
             propertyControlsInfo: updatePropertyControlsOnDescriptorFileDelete(
               editor.propertyControlsInfo,
               action.filename,
             ),
           }
-          return removeErrorMessagesForFile(withUpdatedPropertyControls, action.filename)
```
Rheeseyb, line 4237: We must have been assuming that any other change would trigger a canvas render

Rheeseyb (approved):

PR #6315, opened 2024-09-04
fix(parse): add variables altered inside a code to the scope

### LINKED ISSUE #5762: Mutating a prop in the component code does not change its value
If you mutate a prop in the component implementaion, the prop value does not change.

See https://utopia.pizza/p/4ec59759-chain-loaf/

Image

liady:
**Problem:**
Today code blocks that re-assign variables defined elsewhere are not updating the variables, since we only take into consideration variable declarations. For example this code will throw an error on our canvas:
```tsx
const Foo = ({ prop }) => {
  prop = { title: 'test' }
  return <div>{prop.title}</div>
}
```

**Fix:**
1. Treat re-assigned variables inside arbitrary code as identifiers that need to be returned to update the scope. More specifically - this PR adds them to the `utopiaCanvasBlockRanToEnd` call that runs in the end of every code block.
2. In addition - in the `runBlockUpdatingScope` function where we actually update the scope - don't filter them out (currently it only takes `definedWithin` into consideration, so this PR adds any re-assigned variables that were defined elsewhere). 

**Manual Tests:**
I hereby swear that:

Fixes #5762


bkrmendy (approved):

Rheeseyb (approved):
Nice!

PR #6325, opened 2024-09-05
Improved `buildTree`.

seanparsons:
**Problem:**
`buildTree` is another function on our hot path for any interaction and can often cost around 2ms of frame time.

**Fix:**
By combining the logic from `buildTreeRecursive_MUTATE` and `getChildrenPaths`, we are able to turn the logic for adding an item to the tree into a function which is executed for the two cases where paths are added. Then it's possible to reshape the recursive call into a loop which avoids some unnecessary array construction. Along with this some tweaks were made to avoid some repetition in the logic which checks the types of the elements as they were working from the `ElementInstanceMetadata` each time for 5 different cases.

To help with ordering and to avoid having to do a lot of re-ordering the `children` field of `ElementPathTree` is now split into two distinct fields. As the inner children (from within the component) should be listed ahead of the props children (from within the JSX element in the code).

**Results:**
Benchmarks before:
```
Running "buildTree - deeply nested elements" suite...
  deeply nested elements:
    2,605 ops/s, ±2.16%
Running "buildTree - very wide elements" suite...
  very wide elements:
    9,462 ops/s, ±1.88%
```

Benchmarks after:
```
Running "buildTree - deeply nested elements" suite...
  deeply nested elements:
    5,431 ops/s, ±1.39%
Running "buildTree - very wide elements" suite...
  very wide elements:
    23,598 ops/s, ±0.73%
```

The above translates to `buildTree` taking about 0.5ms to run consistently.

**Commit Details:**
- Reworked `buildTree_MUTATE` to build the tree from the root downwards, combining it with some of the logic taken from `getChildrenPaths`.
- Further improvement to `isDescendantOf` replacing the previous implementation.
- `getReorderedPaths` now avoids a lot of repeated logic for checking the element type.
- Broke apart `ElementPathTree.children` into two separate fields
  to make for easier ordering.
- Added `getElementPathTreeChildren` and `forEachElementPathTreeChild` utility functions.
- `buildTree_MUTATE` now adds to the two different children fields based
  on the element path representing inner children over property children.
- `getReorderedIndexInPaths` further reworked to handle shifting by the siblings.
- Fixed some tests that had expressions incorrectly ordered.

**Manual Tests:**
I hereby swear that:

Fixes #[6324](https://github.com/concrete-utopia/utopia/issues/6324)


editor/src/core/shared/element-path.ts
```diff
@@ -642,14 +642,37 @@ export function findAmongAncestorsOfPath<T>(
...
+        // This caters for the case where the paths are the same.
+        return false
+      }
+      return true
+    } else {
+      return false
+    }
+  } else {
+    return false
+  }
```
gbalint, line 675: This is matter of taste, and feel free to ignore, but imho these last two `return false` else branches would be much more readable at the beginning, according to the "happy path to the left" principle (right now you need to search back and find the actual indentation level to check what does this else mean)

seanparsons: For whatever reason I find it easier (within reason) to have it more nested. I've found especially when I'm changing the code a lot that early returns can be a roadblock as I have to rework them into regular if..else statements.

editor/src/core/shared/element-path-tree.ts
```diff
@@ -1,140 +1,148 @@
...
-        .concat(otherChildrenFromPaths)
-        .sort(
-          (a, b) =>
-            paths.findIndex((p) => EP.pathsEqual(p, a)) -
-            paths.findIndex((p) => EP.pathsEqual(p, b)),
-        )
-    : childrenFromElement
+  for (const elementMetadata of Object.values(metadata)) {
+    forEachRight(elementMetadata.element, (element) => {
+      switch (element.type) {
```
gbalint, line 113: Since you only do something when element is jsxelement or jsxfragment, I would just use an `isJSXElementLike` here (without an else)

editor/src/core/shared/element-path-tree.ts
```diff
@@ -159,67 +167,93 @@ function getReorderedPaths(
...
-    return (
-      MetadataUtils.isConditionalFromMetadata(element) ||
-      MetadataUtils.isExpressionOtherJavascriptFromMetadata(element) ||
-      MetadataUtils.isJSXMapExpressionFromMetadata(element) ||
-      MetadataUtils.isIdentifierFromMetadata(element) ||
-      MetadataUtils.isPropertyAccessFromMetadata(element) ||
-      MetadataUtils.isElementAccessFromMetadata(element)
-    )
+  const pathsToBeReordered = original.filter((path) => {
+    // Omit elements that have a missing parent.
```
gbalint, line 171: The code and the comment is not consistent with each other: comment says missing parent, but code checks for missing ancestor

editor/src/core/shared/element-path-tree.ts
```diff
@@ -159,67 +167,93 @@ function getReorderedPaths(
...
-  return elementsToReorder.reduce((paths, path) => {
+
+  return original.reduce((paths, path) => {
     const index = getReorderedIndexInPaths(paths, metadata, path)
     if (index === 'do-not-reorder') {
       return paths
     }
-    const before = paths.slice(0, index)
-    const after = paths.slice(index)
+    const pathsWithout = paths.filter((pathEntry) => !EP.pathsEqual(pathEntry, path))
```
gbalint, line 180: Minor, but if getReorderedIndexInPaths would return the original index of path too, we would not have to search for it again with this filter

gbalint (approved):

editor/src/core/shared/element-path-tree.ts
seanparsons: Hmmm, it'll make this logic a bit funkier, I'll leave it for the moment.

editor/src/core/shared/element-path-tree.ts
seanparsons: Fixed in 8c91af726.

editor/src/core/shared/element-path-tree.ts
seanparsons: Fixed in 8c91af726.

editor/src/benchmarks.benchmark.ts
```diff
@@ -5,9 +5,9 @@ import { benchmarkOptics } from './core/shared/optics.benchmark'
 import { benchmarkPropertyPathFunction } from './core/shared/property-path.benchmark'
 import { benchmarkGetUniqueUids } from './core/model/get-unique-ids.benchmark'
 
-// await benchmarkBuildTree()
+await benchmarkBuildTree()
```
Rheeseyb, line 8: Did you intend to commit this? Or was this just you running the benchmark locally?

seanparsons: It was from me running it locally, not sure if there's a good answer to not changing this all the time.

editor/src/core/shared/element-path.ts
balazsbajorics: I have to agree with Balint, I opened this resolved comment because I was about to comment here myself, that at least here in the diff viewer this does not look refactor-proof to me: 
image

editor/src/core/shared/element-path-tree.ts
```diff
@@ -1,140 +1,146 @@
...
         } else {
-          return (
-            !isJSXTextBlock(child) &&
-            !isJSExpressionMapOrOtherJavaScript(child) &&
-            !isJSIdentifier(child) &&
-            !isJSPropertyAccess(child) &&
-            !isJSElementAccess(child)
-          )
+          elementChildren = element.children.filter((child) => {
+            switch (child.type) {
```
Rheeseyb, line 124: Given the number of cases here, is it worth making this exhaustive?

seanparsons: Sure, it only ended up like this as it was translated from the original condition.

editor/src/core/shared/element-path.ts
Rheeseyb: Yeah I'm going to unresolve this comment, because I came here for the same reason. Historically I've been against early returns, but I certainly prefer the "happy path to the left" principle that we've been following in the code for the past year+. I think it's fair to say that that preference is now shared by the majority of the team (based on conversations with Berci and Federico too), so please can we collectively stick to it.

editor/src/core/shared/element-path-tree.ts
seanparsons: Fixed in e0d58e345.

editor/src/core/shared/element-path.ts
Rheeseyb: Thank you (and sorry!)

Rheeseyb (approved):

PR #6335, opened 2024-09-07
chore(test): remove warnings

liady:
**Problem:**
Today we have a lot of warnings in our Karma output which really clutters it

**Fix:**
This PR tries to address some of them:
1. import the default export of `package.json` instead of `import * from`
2. change `moduleIds` from `hashed` to `deterministic` (as Webpack recommends)
3. import `{ act }` from `@testing-library/react` instead of `react-dom/test-utils`

**Manual Tests:**
I hereby swear that:


liady:
Temporarily closing as a spike to be picked up later

PR #6344, opened 2024-09-09
Fix subscribing mutation/resize observers during a build error

gbalint:
**Problem:**
Followup to https://github.com/concrete-utopia/utopia/pull/6326
In a Remix project it can happen that in the dispatch flow we don't have the results of the rendering in the dom yet. It can even happen, that after a build error is fixed, we still don't have the canvas root container in the dom.
That is a critical issue, because we attach the mutation/resize observers to the canvas root container element, so when we don't have that we can subscribe them, so the dom sample is not going to run when the rendering is finished.

**Fix:**
I needed an element, which is an ancestor of the canvas root container and
1 is always rendered
2 does not contain canvas controls or other content
I choose canvas-container-outer, but I needed to modify its ancestors so it is always rendered, even when there is a build error.
To make sure we don't observe unnecessary elements, I filtered the query for the resize observer to only include elements with the `data-uid` attribute.

I needed to update the tests to make sure the observers from the previous domwalkermutablestate don't continue to fire when there is a change, because the outer canvas container can stay persistent between test runs.

**Manual Tests:**
I hereby swear that:

Fixes https://github.com/concrete-utopia/utopia/issues/6345

### DIFF
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -295,11 +295,11 @@ export function resubscribeObservers(domWalkerMutableState: {
   let rootElementId = CanvasContainerID
   let rootElement = document.getElementById(rootElementId)
 
-  // if (rootElement == null) {
-  //   // when there is a build error the canvas root container is not rendered, so we attach the observers to the canvas wrapper component
-  //   rootElementId = CanvasWrapperComponentId
-  //   rootElement = document.getElementById(rootElementId)
-  // }
+  if (rootElement == null) {
+    // when there is a build error the canvas root container is not rendered, so we attach the observers to the canvas wrapper component
+    rootElementId = CanvasWrapperComponentId
+    rootElement = document.getElementById(rootElementId)
+  }
 
   if (
     ObserversAvailable &&
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -299,6 +299,8 @@ export function resubscribeObservers(domWalkerMutableState: {
     domWalkerMutableState.resizeObserver != null &&
     domWalkerMutableState.mutationObserver != null
   ) {
+    domWalkerMutableState.resizeObserver.disconnect()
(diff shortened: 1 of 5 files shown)


editor/src/components/canvas/dom-walker.ts
```diff
@@ -291,17 +292,25 @@ export function resubscribeObservers(domWalkerMutableState: {
   mutationObserver: MutationObserver
   resizeObserver: ResizeObserver
 }) {
-  const canvasRootContainer = document.getElementById(CanvasContainerID)
+  let rootElementId = CanvasContainerID
+  let rootElement = document.getElementById(rootElementId)
+
+  if (rootElement == null) {
+    // when there is a build error the canvas root container is not rendered, so we attach the observers to the canvas wrapper component
```
balazsbajorics, line 299: hmm, why not use `canvas-container-outer` or another div which is wrapping the CanvasContainer?

liady (approved):

editor/src/templates/editor-canvas.tsx
```diff
@@ -871,10 +876,9 @@ export class EditorCanvas extends React.Component<EditorCanvasProps> {
 
     const canvasIsLive = isLiveMode(this.props.editor.mode)
 
-    const canvasControls = React.createElement(NewCanvasControls, {
-      windowToCanvasPosition: this.getPosition,
-      cursor,
-    })
+    const canvasControls = (
```
balazsbajorics, line 879: so I am only reading this code and haven't tried it yet, but doesn't this mean this will become _another_ div that is mounted a bit higher in the tree?

balazsbajorics (approved):

seanparsons (approved):

editor/src/templates/editor-canvas.tsx
Rheeseyb: This was always there. Balint has just updated this to use JSX syntax rather than `React.createElement`

Rheeseyb (approved):

PR #6350, opened 2024-09-11
fix(grid): move controls when changing content placement

liady:
**Problem:**
Today the grid controls placeholder don't move when changing `content` placement
10-45-psczv-wnkpg
 

**Fix:**
Copy `justifyContent` and `alignContent` props to the grid placeholder as well.

<video src="https://github.com/user-attachments/assets/0e8dfd6e-f6ca-4b46-bd83-89b2472d03a8"></video>

**Note:**
We should discuss how we can avoid these kinds of issues in the future - where there might be a future grid style prop that will not be copied to the grid control

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/controls/grid-controls.tsx
+++ b/editor/src/components/canvas/controls/grid-controls.tsx
@@ -375,6 +375,8 @@ export type GridData = {
   gridTemplateColumnsFromProps: GridAutoOrTemplateBase | null
   gridTemplateRowsFromProps: GridAutoOrTemplateBase | null
   gap: number | null
+  justifyContent: string | null
+  alignContent: string | null
   rowGap: number | null
   columnGap: number | null
   padding: Sides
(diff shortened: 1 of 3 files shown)


editor/src/components/editor/store/store-deep-equality-instances-3.spec.ts
```diff
@@ -265,6 +265,7 @@ describe('SpecialSizeMeasurementsKeepDeepEquality', () => {
     gap: 11,
     flexDirection: 'column',
     justifyContent: 'center',
+    alignContent: 'auto',
```
seanparsons, line 268: Looks like you've updated the spec but not the implementation?

liady: yeah, I wonder why TS didn't warn about it. fixed!

seanparsons (approved):

editor/src/components/editor/store/store-deep-equality-instances-3.spec.ts
seanparsons: It definitely did on the CI server! :)

PR #6351, opened 2024-09-11
Grid reorder

### LINKED ISSUE #6355: Reorder grid

ruggi:
**Problem:**
Moving grid elements currently means only explicitly setting their grid positioning props (row, column), which has two issues:
1. it does not reorder elements in the code nor the navigator
2. it does not leverage the grid intrinsic ordering, those properties should be set only when strictly necessary

**Fix:**

Extend the strategy so that it prioritizes reordering instead of explicit positioning, and even in the cases where it needs to use row/col props it will still reorder based on the starting index position in the grid hierarchy (left-right, top-bottom).

Fixes #6355


seanparsons (approved):
Nice comments!

bkrmendy (approved):

PR #6353, opened 2024-09-11
Execution Count Limiter

seanparsons:
**Problem:**
Sometimes we want to ensure that a function does not run more than X number of times during a render or similar.

**Fix:**
Created a utility function that invisibly wraps around an existing function to provide an execution count limited version of it, along with a function to potentially reset the count.

**Commit Details:**
- Added `limitExecutionCount`, which can wrap around a function and prevent repeated execution with an exception being thrown.
- Wrapped `runDomSampler` with `limitExecutionCount`.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/templates/editor.tsx
+++ b/editor/src/templates/editor.tsx
@@ -136,7 +136,7 @@ import {
   traverseReadOnlyArray,
 } from '../core/shared/optics/optic-creators'
 import { keysEqualityExhaustive, shallowEqual } from '../core/shared/equality-utils'
-import { runDomSampler } from '../components/canvas/dom-sampler'
+import { resetRunDomSamplerLimit, runDomSampler } from '../components/canvas/dom-sampler'
 import { omitWithPredicate } from '../core/shared/object-utils'
 
 if (PROBABLY_ELECTRON) {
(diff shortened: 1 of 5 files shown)


Rheeseyb (approved):

editor/src/templates/editor.tsx
```diff
@@ -431,6 +431,7 @@ export class Editor {
   ): {
     entireUpdateFinished: Promise<any>
   } => {
+    resetRunDomSamplerLimit()
```
balazsbajorics, line 434: I wonder if we should have one function here that resets _all_ execution count limited wrapped functions, otherwise there will be a long list here

seanparsons: In `LimitExecutionCountOptions` there's a `addToResetArray` property which can be used to put the reset calls into a common array for that kind of usage. Since there's only one and this is partly to demonstrate things, I didn't use that in this PR (other than in the tests). But also I foresee that it's likely we'll have different reset points through the code and not just the one big one here at the start of the dispatch.

balazsbajorics: I have a feeling that we benefit if the first version is more specific to the dispatch flow (and thus we use the resetArray), rather than an overly generic function, but we'll see once we add the second use, I guess

balazsbajorics (approved):

editor/src/templates/editor.tsx
seanparsons: Changes from elsewhere discussion made in [0ac7c0f](https://github.com/concrete-utopia/utopia/pull/6353/commits/0ac7c0f616f9068ee4663efde369ddcfabeec8f2).

PR #6354, opened 2024-09-11
feat(grid): advanced grid controls

liady:
This PR adds advanced grid controls:
* add inspector modal for advanced controls
* justify + Align items - as chain controls
* justify + Align content - as dropdowns
* adding the ability to unset by clicking on the same button

<video src="https://github.com/user-attachments/assets/17fc9368-ad77-4802-9f50-faa0c798136a"></video>

Reviewers: the main logic is in [`editor/src/components/inspector/controls/advanced-grid-modal.tsx`](https://github.com/concrete-utopia/utopia/pull/6354/files#diff-0c658c76e054ebe3b417768607e4f5b50552cef5825e33d0cf694a966a04bc36)

**Manual Tests:**
I hereby swear that:


editor/src/components/inspector/controls/advanced-grid-modal.tsx
```diff
@@ -0,0 +1,251 @@
...
+  closePopup?: () => void
+  style?: React.CSSProperties
+  pickerOffset?: {
+    x: number
+    y: number
+  }
+}
+
+export const AdvancedGridModal = React.memo((props: AdvancedGridModalProps) => {
+  const pickerOffset = props.pickerOffset ?? { x: -280, y: -20 }
```
balazsbajorics, line 33: this feels very hacky, can't we anchor the picker properly so it doesn't need to know about a magic number which is approximately the inspector width?

editor/src/components/inspector/controls/advanced-grid-modal.tsx
```diff
@@ -0,0 +1,251 @@
...
+      id={`trigger-${props.id}`}
+      className={`ignore-react-onclickoutside-${props.id}`}
+      style={props.style}
+    >
+      {props.popupOpen ? picker : null}
+    </div>
+  )
+})
+
+export const itemsOptions = (alignOrJustify: 'align' | 'justify') =>
```
balazsbajorics, line 186: minor: this doesn't need to be exported

editor/src/components/inspector/controls/advanced-grid-modal.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const closePopup = React.useCallback(() => {
+    if (!dropdownOpen.justifyContent && !dropdownOpen.alignContent) {
+      props.closePopup?.()
+    }
+  }, [dropdownOpen, props])
+
+  const alignItemsControl = useInspectorLayoutInfo('alignItems')
+  const justifyItemsControl = useInspectorLayoutInfo('justifyItems')
+  const alignContentControl = useInspectorLayoutInfo('alignContent')
+  const justifyContentControl = useInspectorLayoutInfo('justifyContent')
```
balazsbajorics, line 53: I don't think we call the return value of `useInspectorLayoutInfo` "control" anywhere, we usually reserve the word Control for an actual UI element

editor/src/uuiui/radix-components.tsx
```diff
@@ -286,6 +293,7 @@ export const RadixSelect = React.memo(
               borderRadius: 6,
               color: 'white',
               boxShadow: UtopiaStyles.shadowStyles.high.boxShadow,
+              zIndex: 3,
```
balazsbajorics, line 296: this looks scary

editor/src/core/performance/__snapshots__/performance-regression-tests.spec.tsx.snap
```diff
@@ -576,6 +576,29 @@ Array [
...
+  "/Symbol(react.forward_ref)(Styled(div))/div//Symbol(react.forward_ref)(UIGridRow)",
+  "/Symbol(react.forward_ref)(Styled(div))/div//span",
+  "/Symbol(react.forward_ref)(Styled(div))/div//Symbol(react.memo)(RadixSelect)",
+  "/Symbol(react.forward_ref)(Styled(div))/div//Symbol(react.forward_ref)(UIGridRow)",
+  "/Symbol(react.forward_ref)(Styled(div))/div//span",
+  "/Symbol(react.forward_ref)(Styled(div))/div//Symbol(react.memo)(RadixSelect)",
+  "/Symbol(react.forward_ref)(Styled(div))/div//Symbol(react.forward_ref)(UIGridRow)",
+  "/Symbol(react.forward_ref)(Styled(div))/div//Symbol(react.forward_ref)(Styled(div))",
+  "/Symbol(react.forward_ref)(Styled(div))/div//UtopiaSpiedFunctionComponent(InspectorModal)",
+  "/Symbol(react.forward_ref)(Styled(div))/div//div",
```
balazsbajorics, line 601: it looks like we are now rendering a bunch of extra elements even when the advanced modal is closed? that is possibly impacting performance negatively. 

could you add `.displayName` props to the new components, so we can know more? thanks!

balazsbajorics:
image
clicking on the ... to open this dropdown prints a flushSync warning to the console:
image

99% of the time this means that a `dispatch` was called from a react lifecycle method which is illegal in our dispatch flow. Instead the dispatch should be called from a setTimeout(0) or the fancy new `queueMicrotask`

editor/src/components/inspector/controls/advanced-grid-modal.tsx
liady: definitely, this was already in the original code so I missed it - I'll fix

editor/src/uuiui/radix-components.tsx
liady: was set to `auto` to make sure the dropdown will always be the on top

editor/src/uuiui/radix-components.tsx
```diff
@@ -293,7 +293,7 @@ export const RadixSelect = React.memo(
               borderRadius: 6,
               color: 'white',
               boxShadow: UtopiaStyles.shadowStyles.high.boxShadow,
-              zIndex: 3,
+              zIndex: 'auto',
```
balazsbajorics, line 296: this is the initial value, are you setting this here in case the className would override it to something non-initial?

liady: so it needs an explicit z-index value, otherwise it is hidden behind the modal, even though their portals are in the correct order in the DOM. however simply 'auto' didn't fix it as well, so I'm passing it from outside now.

editor/src/core/performance/__snapshots__/performance-regression-tests.spec.tsx.snap
liady: another leftover from the previous code - fixed 🙏

liady:
> image clicking on the ... to open this dropdown prints a flushSync warning to the console: image
> 99% of the time this means that a `dispatch` was called from a react lifecycle method which is illegal in our dispatch flow. Instead the dispatch should be called from a setTimeout(0) or the fancy new `queueMicrotask`

@balazsbajorics this actually happens on master as well (it's unrelated to the advanced grid popup) - I'll try to fix it in a separate PR perhaps?

editor/src/components/inspector/controls/advanced-grid-modal.tsx
liady: automatically getting the offset from the panel width now - and just setting the desired distance from it

editor/src/components/inspector/controls/advanced-grid-modal.tsx
```diff
@@ -0,0 +1,283 @@
...
+import {
+  RadixSelect,
+  regularRadixSelectOption,
+  separatorRadixSelectOption,
+} from '../../../uuiui/radix-components'
+import { optionalMap } from '../../../core/shared/optional-utils'
+import type { FlexAlignment } from 'utopia-api/core'
+import { FlexJustifyContent } from 'utopia-api/core'
+
+export interface AdvancedGridModalProps {
```
ruggi, line 19: it seems to me this needs some cleaning up now, for example `key` is not needed, `popupOpen`/`popupClose` don't need to be optional etc.

editor/src/components/inspector/controls/advanced-grid-modal.tsx
```diff
@@ -0,0 +1,283 @@
...
+      if (value === 'unset') {
+        alignContentLayoutInfo.onUnsetValues?.()
+      } else {
+        alignContentLayoutInfo.onSubmitValue(value as FlexJustifyContent)
+      }
+    },
+    [alignContentLayoutInfo],
+  )
+
+  const justifyItemsOptions = React.useMemo(() => itemsOptions('justify'), [])
```
ruggi, line 104: I think these two can just be constants, right?

editor/src/components/inspector/controls/advanced-grid-modal.tsx
```diff
@@ -0,0 +1,283 @@
...
+  const alignItemsOptions = React.useMemo(() => itemsOptions('align'), [])
+
+  const rowVariant = '|--67px--|<--------1fr-------->'
```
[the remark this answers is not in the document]
liady: added the type (and moved out of the render function)

editor/src/components/inspector/controls/advanced-grid-modal.tsx
liady: yes, moved out of the render function

balazsbajorics (approved):

PR #6368, opened 2024-09-13
Grid cell bounds in metadata

gbalint:
**Problem:**
Grid strategies use DOM api to find grid cells under the mouse, find their bounding box, etc.
That is against the strategy contract: strategies should rely on their input and should not inspect the dom.
This PR is a first step to eliminate these: it stores the global frames of grid cells in SpecialSizeMeasurements.

To calculate that, it relies on the `padding`, `rowGap`, `columnGap, `gridTemplateRows` and `gridTemplateColumns` values of the computed style.

We already parsed these into SpecialSizeMeasurements.containerGridProperties`, but with a way more general type than necessary (it uses the same type as the parsed style prop, so it allows keywords and not just numbers).

**TODO:**
- [ ] refactor the types, and create a `ComputedGridContainerProperties` which only allows numbers in the dimensions
- [ ] refactor grid helpers to rely on metadata and not on the dom api
- [ ] consider just using the `getGlobalFramesOfGridCells` function directly from the strategies and do not store the values in the metadata. In this case we have to be sure we memoize the function well, because it will be called in every frame.

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/dom-walker.ts
```diff
@@ -956,3 +970,51 @@ function getClosestOffsetParent(element: HTMLElement): Element | null {
...
+
+  return cellRects
+}
+
+function gridTemplateToNumbers(gridTemplate: GridTemplate | null): Array<number> | null {
+  if (gridTemplate?.type !== 'DIMENSIONS') {
+    return null
+  }
+
+  const result: Array<number> = []
```
ruggi, line 1010:
```suggestion
  let result: Array<number> = []
```

editor/src/components/canvas/dom-walker.ts
```diff
@@ -956,3 +970,51 @@ function getClosestOffsetParent(element: HTMLElement): Element | null {
...
+
+  if (columnWidths == null || rowHeights == null) {
+    return null
+  }
+
+  const cellRects: Array<Array<CanvasRectangle>> = []
+  let yOffset = padding.top ?? 0
+  rowHeights.forEach((height) => {
+    let xOffset = padding.left ?? 0
+    const rowRects: CanvasRectangle[] = []
```
ruggi, line 992:
```suggestion
    let rowRects: CanvasRectangle[] = []
```

editor/src/components/canvas/dom-walker.ts
```diff
@@ -956,3 +970,51 @@ function getClosestOffsetParent(element: HTMLElement): Element | null {
...
+): Array<Array<CanvasRectangle>> | null {
+  const columnWidths = gridTemplateToNumbers(grid.gridTemplateColumns)
+
+  const rowHeights = gridTemplateToNumbers(grid.gridTemplateRows)
+
+  if (columnWidths == null || rowHeights == null) {
+    return null
+  }
+
+  const cellRects: Array<Array<CanvasRectangle>> = []
```
ruggi, line 988:
```suggestion
  let cellRects: Array<Array<CanvasRectangle>> = []
```

PR #6371, opened 2024-09-13
Calculate global frame of grid cells

gbalint:
**Problem:**
Grid strategies use DOM api to find grid cells under the mouse, find their bounding box, etc.
That is against the strategy contract: strategies should rely on their input and should not inspect the dom.
This PR is a first step to eliminate these: it introduces a function which calculates the global frames of grid cells from some fields of SpecialSizeMeasurements.

Note: we use SpecialSizeMeasurements.containerGridProperties`, and that has a way more general type than necessary (it uses the same type as the parsed style prop, so it allows keywords and not just numbers).

**TODO:**
- [ ] refactor the types, and create a `ComputedGridContainerProperties` which only allows numbers in the dimensions
- [ ] memoize the function
- [ ] refactor grid helpers to rely on metadata and not on the dom api

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
@@ -4,6 +4,7 @@ import * as EP from '../../../../core/shared/element-path'
 import type {
   ElementInstanceMetadataMap,
   GridPositionValue,
+  GridTemplate,
 } from '../../../../core/shared/element-template'
 import {
   gridPositionValue,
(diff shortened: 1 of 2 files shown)


ruggi (approved):
very nice!

editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
```diff
@@ -604,3 +611,62 @@ function getGridPositionIndex(props: {
...
+
+  const cellRects: Array<Array<CanvasRectangle>> = []
+  let yOffset = padding.top ?? 0
+  rowHeights.forEach((height) => {
+    let xOffset = padding.left ?? 0
+    const rowRects: CanvasRectangle[] = []
+    columnWidths.forEach((width) => {
+      const rect = canvasRectangle({ x: xOffset, y: yOffset, width: width, height: height })
+      rowRects.push(rect)
+      xOffset += width + (columnGap ?? 0)
```
liady, line 648: a small note - the gap shouldn't be added to the last column/row dimensions

liady, line 648: A small note - the gap shouldn't be added to the first offset of row/column

gbalint: It is not added to the first one, the offset is increased _after_ I added the first rect in line 647.

seanparsons (approved):

liady (approved):

PR #6373, opened 2024-09-16
No dom query in grid element resize strategy

gbalint:
**Problem:**
Another step to remove dom api dependency from grid strategies: the grid-resize-element strategy uses the new `getGlobalFramesOfGridCells` to detect the grid cell under the mouse.

**Commit Details:** 

- Fixed `getGlobalFramesOfGridCells` problem: it calculated the local frame not the global frame....
- Fixed and improved `getGlobalFramesOfGridCells` test to include differend padding/row-padding/column-padding values
- Memoized `getGlobalFramesOfGridCells`: this is a quite fast function so it wouldn't be important to memoize it, but in most cases we run it for the same grid cell (because we are doing an interaction in it), so it still makes sense to memoize. I just uses `memoize` with the default config (which allows memoization of 5 values)
- Created a new `getGridCellUnderMouse` version called `getGridCellUnderMouseFromMetadata`, and call that from the strategy
- I modified the strategy to use the canvas and not the window coordinate system. Unfortunately, `GridCustomStrategyState.targetCellData` uses the window coordinate system, and that is used in multiple strategies, so I postponed to change that too. I introduced a new type `TargetGridCellDataCanvas` (similar to `TargetGridCellData`), I plan to migrate to this new type everywhere.

**TODO:**

The strategy is still not the best. A downside of querying the grid cells under the mouse is that the strategy is broken when the mouse is not over the grid (even though the resize rectangle follows the mouse), see https://screenshot.click/16-30-s1avn-hgu83.mp4

Using the metadata it is possible to implement a much better strategy which can calculate where to resize to even when the mouse is not over the grid. That can come in a subsequent PR.

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/canvas-strategies/strategies/grid-cell-bounds.ts
```diff
@@ -27,6 +33,38 @@ export function gridCellTargetId(
...
+  const gridCellGlobalFrames = getGlobalFramesOfGridCells(grid)
+
+  if (gridCellGlobalFrames == null) {
+    return null
+  }
+
+  return getGridCellUnderPoint(gridCellGlobalFrames, canvasPoint)
+}
+
+// TODO: optimization: binary search
```
Rheeseyb, line 49: Unless we have a grid with a ton of rows / columns this really won't matter

editor/src/components/canvas/dom-lookup.ts
```diff
@@ -636,3 +639,18 @@ export function canvasPointToWindowPoint(
     throw new Error('calling screenToElementCoordinates() before being mounted')
   }
 }
+
+export function canvasRectangleToWindowRectangle(
```
Rheeseyb, line 643: Huh, I thought we already had this function

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/grid-cell-bounds.ts
```diff
@@ -27,6 +33,38 @@ export function gridCellTargetId(
...
+          gridCellCoordinates: gridCellCoordinates(i + 1, j + 1),
+          cellCanvasRectangle: gridCellGlobalFrames[i][j],
+        }
+      }
+    }
+  }
+  return null
+}
+
+// TODO: should be superseded by getGridCellUnderMouseNoDomLookup
```
balazsbajorics, line 67: do you mean

```suggestion
// TODO: should be superseded by getGridCellUnderMouseFromMetadata
```


?

liady (approved):

balazsbajorics (approved):

editor/src/components/canvas/dom-lookup.ts
gbalint: me too :)

editor/src/components/canvas/canvas-strategies/strategies/grid-cell-bounds.ts
gbalint: yes, meanwhile I renamed the function :)

PR #6380, opened 2024-09-17
Preserve repeats when resizing grids

### LINKED ISSUE #6379: Preserve repeats when resizing grids

ruggi:
**Problem:**

**Fix:**

- Did [other things]

**Manual Tests:**
I hereby swear that:

Fixes #6379


editor/src/components/canvas/canvas-strategies/strategies/resize-grid-strategy.ts
```diff
@@ -169,6 +199,64 @@ export const resizeGridStrategy: CanvasStrategyFactory = (
   }
 }
 
+type ExpandedGridDimension = GridDimension & {
+  repeated: boolean
+  originalIndex: number
+}
+
+// Reconstruct the final, resized dimensions trying to preserve as much as possible the original CSS.
```
Rheeseyb, line 207: Thank you for this comment!

editor/src/components/canvas/canvas-strategies/strategies/resize-grid-strategy.ts
```diff
@@ -169,6 +199,64 @@ export const resizeGridStrategy: CanvasStrategyFactory = (
...
+  // 3.
+  // midChunkStart is the index in the expanded values matching the first occurrence of the target index in the
+  // expanded values.
+  const midChunkStart = params.expandedOriginalValues.findIndex(
+    (c) => c.originalIndex === targetIndex,
+  )
+  // midChunkEnd is the start + the number of repeated elements for this `repeat` function
+  const midChunkEnd =
+    midChunkStart +
+    params.expandedOriginalValues.filter((c) => c.originalIndex === targetIndex).length
```
Rheeseyb, line 237: I think the name `midChunkEnd` is now incorrect if this is a `repeat`, because of the case e.g. `repeat(2, 5px, 10px)`. It might be clearer to have `const midChunkSize = params.expandedOriginalValues.filter((c) => c.originalIndex === targetIndex).length` and then only introduce `midChunkEnd` in the `else` statement

ruggi: [`32c1467` (#6380)](https://github.com/concrete-utopia/utopia/pull/6380/commits/32c1467d962b501ec45a63871df1f4ebf5088d49)

editor/src/components/canvas/canvas-strategies/strategies/resize-grid-strategy.ts
```diff
@@ -149,7 +172,14 @@ export const resizeGridStrategy: CanvasStrategyFactory = (
           ),
         mergedValues.dimensions,
       )
-      const propertyValueAsString = printArrayGridDimension(newSetting)
+
+      const newDimensions = buildResizedDimensionsArray({
```
balazsbajorics, line 176: I haven't read the code in detail yet but _it feels like_ what is happening is that `resizedDimensions` is trash and you wrote an elaborate helper function `buildResizedDimensionsArray` to essentially untrash it. can't we make it so that `resizedDimensions` is not weird to begin with? I feel like `buildResizedDimensionsArray` could be remade to do the actual resizing if you would give it a single `resizedValue` instead of all of `resizedDimensions`, and then we could ditch `getNewDragValue`

ruggi: Makes sense, updated the code to replace directly instead 👍

balazsbajorics (approved):

PR #6382, opened 2024-09-17
feat(parser): parallel file parsing

### LINKED ISSUE #5655: Parser performance improvement: Look into parallelising the parsing of multiple files
Depends on #5654 

If we no longer need to keep track of the existing UIDs when parsing a file, we should be able to parse each file in parallel, so we should investigate that and see how possible that is and how much work it would take. I've suggested a timebox of 2 days here, but we can tweak that as necessary depending how performance is looking after #5652 #5653 and #5654 have been completed.

liady:
This PR enables parallel parsing of files, utilizing multiple parser workers.
Until now we had only one `ParserPrinterWorker`, which had to handle parsing all of the files serially. In our sample store this could be ~10-15s spent only on parsing (depending on the machine).

In this PR we're introducting the ability to spin multiple workers, and divide the file load between them, allowing the parsing to be parallel.

**Note to reviewers:** Main logic changes are in `worker-types.ts` and `workers.ts`

Some important points:
1. Since parsing the files happens on the load critical path, we can easily hit a hardware limitation due to multiple workers/main thread that take up all the processes. This limits us both in the number of new workers we can spin up, and also causing N workers to work for more that 1/N of the time in parallel. Even though - there are still meaningful gains.
2. There is an overhead to spinning up workers, since our parsing code itself is heavy and takes some time to evaluate. A proper threadpool solution might be able to solve that (spin workers when needed and keep them alive as long as the parse load is large). However this requires changing our entire worker mechansim, and can even hurt performance since spinning up new workers in the "middle" of the parsing flow creates an overhead. However this can be a future optimization - to start with 3-4 workers in advance and if needed spin up more using a threadpool library (`workerspool` or `threads`, for example).
3. Our current chunking algorithm splits the parsable files (not all files are parsable) into N chunks that are more-or-less equal in terms of the sum of their file length, and sends them in parallel to N workers. This makes sure that all of the workers will work for roughly the same amount of time. A proper threadpool queue-based solution could be more efficient here (since it will always send the next file to an available worker) but the difference to the current "naive" chunking implementation wouldn't be that large, and it will also add a small messaging overhead (sending each file in turn).
4. The current concurrency is 3, and this will be configurable from the settings/RYO menu after #6381 will be merged.

**Commit Details:**

- Added support for multiple parse workers in `UtopiaTsWorkers`
- Changed `getParseResult` to chunk the files, and call on each chunk `getParseResultSerial`, waiting for all of them to return before returning the result
- `getParseResultSerial` now looks for the next available `ParserPrinterWorker` to send the next chunk to, for parsing
- Worker code itself remained unchanged
- The feature can be toggles on/off from the settings menu

**Manual Tests:**
I hereby swear that:

Fixes #5655


editor/src/core/performance/performance-utils.ts
```diff
@@ -8,3 +11,38 @@ export function timeFunction(fnName: string, fn: () => any, iterations: number =
   // eslint-disable-next-line no-console
   console.log(`${fnName} took ${timeTaken}ms`)
 }
+
+export function canMeasurePerformance(): boolean {
```
seanparsons, line 15: Nice.

seanparsons (approved):

editor/src/core/workers/common/concurrency-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  return isConcurrencyEnabled() ? PARSE_CONCURRENCY : 1
+}
+
+export function getParserChunkCount() {
+  return isConcurrencyEnabled() ? PARSE_CONCURRENCY : 1
+}
+
+// TODO: this will be configurable from the RYO menu
+export function isConcurrencyLoggingEnabled() {
+  return isConcurrencyEnabled()
```
balazsbajorics, line 21: we should _definitely_ not be logging these by default :) logging a couple hundred lines is a surprising toll on Chrome

liady: fixed. this will also be configurable from the RYO menu once #6381 will be merged

bkrmendy (approved):

editor/src/core/shared/array-utils.ts
```diff
@@ -538,3 +538,41 @@ export function matrixGetter<T>(array: T[], width: number): (row: number, column
...
+  const permutation = array.map((_, index) => index)
+  permutation.sort((a, b) => {
+    const sortResult = sortFn(array[a]) - sortFn(array[b])
+    return ascending ? sortResult : -sortResult
+  })
+  const sortedArray = permutation.map((index) => array[index])
+  return { sortedArray, permutation }
+}
+
+export function sortArrayByReversePermutation<T>(array: T[], permutation: number[]): T[] {
```
Rheeseyb, line 576: This naming is throwing me off a bit, because it reads like it will be reversing the array's order, but it's actually reverting the array to its original order. So maybe a better name would be `revertArrayOrder`?

Rheeseyb (approved):

PR #6396, opened 2024-09-23
Expression editing in grid row/column controls

gbalint:
**Problem:**
Instead of showing the generated rows/columns in the inspector, show the values from templates directly, and allow expression editing there. 
This means `repeat` expression affect multiple generated rows, and this is displayed in the row's label:

image

- the logic is much simpler this way, because one control is directly paired with one  value from gridTemplateRows/Columns
- I could delete `mergeGridTemplateValues`, because the only source of information now are the template row/col values from props
- New GridExpressionInput component
- Some logic to calculate the control label, because now the generated row/col index can be different from the index of the control itself
- changed how rename/delete works, because the index affects the template row/col from the props, not the generated row/col
- deleted flex-section.spec.tsx, because it only tested the removed `mergeGridTemplateValues`
- Renamed Rows and Columns label to "Template Rows" and "Template Columns"
- Removed the ... menu when the input is focused so we have bigger space to edit the expression:
25-50-i20ib-aftoa

**TODO:**
- we don't take gridAutoRows and gridAutoColumns into account at all (needs design)
- the control is just an input, we don't show a dropdown to choose from allowed keywords

**Manual Tests:**
I hereby swear that:


editor/src/uuiui/inputs/grid-expression-input.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const [printValue, setPrintValue] = React.useState<string>(printGridDimension(value))
+    React.useEffect(() => setPrintValue(printGridDimension(value)), [value])
+
+    const onChange = React.useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+      setPrintValue(e.target.value)
+    }, [])
+
+    const onKeyDown = React.useCallback(
+      (e: React.KeyboardEvent) => {
+        if (e.key === 'Enter') {
```
bkrmendy, line 44: happy-path-to-the-left would reduce the nesting here

bkrmendy (approved):

editor/src/components/inspector/common/css-utils.ts
```diff
@@ -652,6 +652,21 @@ export type GridCSSMinmax = BaseGridDimension & {
...
 }
 
+export function parseGridCSSMinmaxOrRepeat(input: string): GridCSSMinmax | GridCSSRepeat | null {
+  const parsed = csstree.parse(input, { context: 'value' })
+  if (parsed.type === 'Value') {
+    const parsedDimensions = parseGridChildren(parsed.children)
+    if (
+      isRight(parsedDimensions) &&
+      parsedDimensions.value.length === 1 &&
+      (parsedDimensions.value[0].type === 'MINMAX' || parsedDimensions.value[0].type === 'REPEAT')
```
ruggi, line 662: maybe it'd be nice to use `isGridCSSMinmax` and `isGridCSSRepeat`?

editor/src/components/inspector/flex-section.spec.browser2.tsx
```diff
@@ -1,4 +1,4 @@
-import { selectComponentsForTest } from '../../utils/utils.test-utils'
+import { selectComponentsForTest, wait } from '../../utils/utils.test-utils'
```
ruggi, line 1:
```suggestion
import { selectComponentsForTest } from '../../utils/utils.test-utils'
```

ruggi (approved):

PR #6397, opened 2024-09-23
Color Picker Responsiveness

seanparsons:
**Problem:**
The color picker can be laggy to use and feels slow.

**Fix:**
The four main changes are:
- Updates to the indicator controls are done directly to the elements.
- The mouse move is prevented from stacking up calls by using the same logic that is used in the canvas.
- A slight refactor for `shouldIncludeVSCodeDecorations` that takes its runtime from 3ms to about a tenth of that.
- Removed focusing of color picker input as it was the root cause of an initial janky dragging, as it fired a non-transient change which then caused a chain of worker updates to fire as the user continued to drag.

`getDerivedStateFromProps` was also very slightly simplified mostly to see what's going on.

**Commit Details:**
- Simplified `getDerivedStateFromProps` as it was setting a property that was never used.
- Added `setIndicators` which moves the color circle and the sliders based on the current state.
- Added optional trigger for `setIndicators` to `componentDidUpdate`.
- The various mouse move handlers in the color picker check `mouseMoveHandled`.
- Added some utility functionality so that it is possible to check if an event target is inside the color picker.
- Added call for `setIndicators` to `componentDidMount`.
- Made a slight refactor to `shouldIncludeVSCodeDecorations` that radically improves its performance.
- Removed focusing of input in color picker that didn't appear to work as it was causing a non-transient change to fire.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/inspector/controls/color-picker.tsx
+++ b/editor/src/components/inspector/controls/color-picker.tsx
@@ -198,8 +198,6 @@ export class ColorPickerInner extends React.Component<
   ColorPickerInnerProps,
   ColorPickerInnerState
 > {
-  private RefFirstControl = React.createRef<HTMLInputElement>()
-
   private fullWidth = colorPickerWidth
   private fullHeight = MetadataEditorModalPreviewHeight
   private paddedWidth = this.fullWidth - inspectorEdgePadding * 2
@@ -237,9 +235,6 @@ export class ColorPickerInner extends React.Component<
   componentDidMount() {
     setTimeout(() => {
       // wrapping in a setTimeout so we don't dispatch from inside React lifecycle
-      if (this.RefFirstControl.current != null) {
-        this.RefFirstControl.current.focus()
-      }
       this.setIndicators()
     }, 0)
   }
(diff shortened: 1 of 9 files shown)


editor/src/components/inspector/controls/color-picker.tsx
```diff
@@ -322,16 +338,21 @@ export class ColorPickerInner extends React.Component<
...
-    const newValue = toPassedInColorType(
-      chromaColor,
-      this.props.value,
-      this.state.normalisedHuePosition,
-    )
-    if (transient) {
-      this.props.onTransientSubmitValue(newValue)
-    } else {
-      this.props.onSubmitValue(newValue)
+    if (this.updateTimeoutId.current != null) {
```
balazsbajorics, line 341: this will only update the color on the canvas once you let go of dragging, right?

seanparsons: Nope, if you stop dragging for a moment it'll fire the update as well.

balazsbajorics: I updated my comment, I think we should use onTransientSubmitValue here

balazsbajorics (approved):

PR #6398, opened 2024-09-23
Strategy Controls Improvement

seanparsons:
**Problem:**
We often generate identical values for the canvas controls from the strategies, but because they're generated values they cause the controls to re-render.

**Fix:**
With `useKeepReferenceEqualityIfPossible` to make value equal values become reference equal values to make it so that React doesn't re-render the controls.

**Commit Details:**
- Wrap `applicableStrategies` in `useGetApplicableStrategyControls` with a `useKeepReferenceEqualityIfPossible` so that repeated identical values for the controls don't cause repeated re-renders.

**Manual Tests:**
I hereby swear that:


seanparsons:
Seems to cause a bunch of weird test failures.

PR #6399, opened 2024-09-23
chore(editor): remove unneeded network call

liady:
**Problem:**
We have a `/permissions` network call on our loading critical path, but its result is not being used (yet?)

**Fix:**
Remove the call for now since it is sometimes blocking loading the editor (competing for network)


seanparsons (approved):

balazsbajorics (approved):
I wonder if this is leftover from something that has been deleted, or this will be needed in the future once we resume working on permissions

liady:
@balazsbajorics I think this was a part of the "granular editor permissions" feature (for granular control on text edit, design controls, etc) - but this phase was put aside and we decided to stay only with the general "edit" permission for now

PR #6401, opened 2024-09-23
Tailwind class compilation

bkrmendy:
image

## Description

This PR removes twind (as it's not maintained anymore), and replaces it with https://github.com/mhsdesign/jit-browser-tailwindcss/, which builds on tailwindcss to compile tailwind classes. 

### Details
- The code wrapping `twind` is removed
- Wrapper code is added for https://github.com/mhsdesign/jit-browser-tailwindcss/ (`useTailwindCompilation`)
- A Karma test is added to check whether some tailwind classes are compiled correctly (`tailwind.spec.browser2.tsx`). The test is intentionally coarse-grained, because exhaustively testing every tailwind feature is out of scope, and the example store heavily uses tailwind anyway
- Generating the tailwind styles depends on whether the project has a tailwind.config.js file or not. Down the line we'll change this to checking in a flag in the `"utopia"` section of `package.json`, the current implementation is there to keep the PR self-contained
- `isTailwindEnabled` and `TAILWIND_INSTANCE.current`: `isTailwindEnabled` should be a pure function of `projectContents`. However, that implementation caused a lot of re-renders, so instead I mirrored the original implementation that checked whether a module-private variable was set or not
- Stubbed `@mhsdesign/jit-browser-tailwindcss` in Jest


seanparsons (approved):

editor/src/core/tailwind/tailwind-compilation.ts
```diff
@@ -0,0 +1,110 @@
...
+  const allCSSFiles = getCssFilesFromProjectContents(projectContents).join('\n')
+  const rawConfig = importDefault(requireFn('/', TailwindConfigPath))
+  const tailwindCss = createTailwindcss({ tailwindConfig: rawConfig as TailwindConfig })
+  TAILWIND_INSTANCE.current = tailwindCss
+  void generateTailwindStyles(tailwindCss, allCSSFiles)
+}
+
+const generateTailwindClasses = memoize(generateTailwindClassesInner, {
+  maxSize: 1,
+  matchesArg: equalityFnForProjectContents,
```
balazsbajorics, line 70: I really don't like that we are spreading this pattern, this is horrible! 

we should use reselect to create the memoized selector, it gives a much cleaner interface for memoizing parameters.

editor/src/core/tailwind/tailwind-compilation.ts
```diff
@@ -0,0 +1,110 @@
...
+  return tag
+}
+
+async function generateTailwindStyles(tailwindCss: Tailwindcss, allCSSFiles: string) {
+  const contentElement = document.getElementById(CanvasContainerID)
+  if (contentElement == null) {
+    return
+  }
+  const content = contentElement.outerHTML
+  const styleString = await tailwindCss.generateStylesFromContent(allCSSFiles, [content])
```
balazsbajorics, line 42: what does allCSSFiles do here? is generateStylesFromContent using it to avoid duplicates, or is it appending them all to the styleString? if the latter, does it mean if I have any css anywhere in the project, it will be attached to the DOM? that's certainly not something desired

editor/src/core/tailwind/tailwind-compilation.ts
```diff
@@ -0,0 +1,110 @@
...
+})
+
+export const useTailwindCompilation = (requireFn: RequireFn) => {
+  const projectContents = useEditorState(
+    Substores.projectContents,
+    (store) => store.editor.projectContents,
+    'useTailwindCompilation projectContents',
+  )
+
+  const isInteractionActiveRef = useRefEditorState((store) =>
```
balazsbajorics, line 80: I think a safer proxy is ElementsToRerenderGLOBAL.current, but we should definitely have an official function for this

editor/src/templates/preview/preview.tsx
```diff
@@ -280,8 +279,6 @@ const initPreview = () => {
       }
     }
 
-    injectTwind(projectContents, requireFn)
```
balazsbajorics, line 283: shouldn't we replace this with something pointing at the tailwind compiler? 
does the preview.tsx still work at all? if not, we should probably remove this file instead of letting it slowly regress

balazsbajorics (commented):
please use reselect selectors, otherwise looks good!

editor/src/core/tailwind/tailwind-compilation.ts
bkrmendy: https://github.com/concrete-utopia/utopia/pull/6401/commits/d2643f20690a3edafca8dbb7dbc8e68f8c8d22e7

editor/src/templates/preview/preview.tsx
bkrmendy: https://github.com/concrete-utopia/utopia/pull/6401/commits/ef7ec9c60bf04122b82ff35739183678928e460e

editor/src/core/tailwind/tailwind-compilation.ts
bkrmendy: https://github.com/concrete-utopia/utopia/pull/6401/commits/ef7ec9c60bf04122b82ff35739183678928e460e

editor/src/core/tailwind/tailwind-compilation.ts
bkrmendy: Context on what `tailwindCss.generateStylesFromContent` does:

Tailwind styles are generated based on the tailwind config (passed to the constructor of the `tailwindCss` instance), the css files (that contain the `@tailwind` directives, custom font faces, [layers](https://tailwindcss.com/docs/adding-custom-styles#using-css-and-layer)), and the actual tailwind classes in the elements we're interested in (the second argument of `generateStylesFromContent`). Based on all this, `generateStylesFromContent` returns the CSS generated by the Tailwind JIT compiler, which has been optimized for the given content.

Specific answers:
- `generateStylesFromContent` is not using `allCSSFiles` to avoid duplicates
- I haven't specifically debugged the output of `generateStylesFromContent`, since it's been treated as a black box so far

balazsbajorics: thanks! how does "real" tailwind ingest these @tailwind directives? if I write a directive in a css file that I never reference anywhere, will it be picked up by the tailwind/scss precompile step in a real application? As far as I understand, in this implementation here, it _will_ be picked up automatically. I just want to make sure if that matches the real tailwind implementation

bkrmendy: Tailwind is a used as a PostCSS plugin, and the files that PostCSS processes are determined by the build system (see examples here: https://github.com/postcss/postcss?tab=readme-ov-file#usage). Replicating that functionality is out of scope for this PR, we'll polish this on a follow-up PR

PR #6405, opened 2024-09-24
Grid move: absolute move without positioning

### LINKED ISSUE #6404: grid absolute move relative to the grid

ruggi:
**Problem:**

Grid elements with `position: absolute` are always moved relative to the cell they appear in.

**Fix:**

For cell elements that don't have explicit grid positioning (`gridRow` / `gridColumn`), move them relative to the grid itself instead, and don't show the cell target outline during the move.

(Elements that have positioning will still be absolutely positioned relatively to their cell, and the positioning will be kept).

Fixes #6404

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
@@ -5,7 +5,6 @@ import type {
   ElementInstanceMetadataMap,
   GridPositionValue,
   GridTemplate,
-  SpecialSizeMeasurements,
 } from '../../../../core/shared/element-template'
 import {
   gridPositionValue,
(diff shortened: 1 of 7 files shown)


editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
```diff
@@ -509,6 +527,9 @@ function getGridMoveType(params: {
   // For absolute move, just use rearrange.
   // TODO: maybe worth reconsidering in the future?
```
liady, line 528: we can maybe remove this comment now? or is it still relevant for when it's absolute with grid positioning?

ruggi: good point! removing

bkrmendy (approved):

PR #6413, opened 2024-09-25
Grid alignment buttons

### LINKED ISSUE #6412: Grid alignment buttons

ruggi:
**Problem:**

The alignment buttons at the top of the inspector could be repurposed to also be useful when dealing with self align/justify for grids and flex.

**Fix:**

This PR does a bunch of things:

1. Rename the `T` and `L` labels in the frame positioning to `X` and `Y`
2. Move the alignment buttons from the top of the editor to above the `X` and `Y` controls, splitting them into two chain controls
3. Make those buttons control the positioning for flow elements, as well as controlling justify self/align self for flex or grid children
4. Make those buttons display the current state of the property they're controlling
5. Make those buttons unset the property if they are clicked again while active
6. Add a context menu to the right of the buttons to unset the values (the entry will only show if there _are_ set values) as well as distribute horizontally/vertically
7. Adjust the radix dropdown menu items so that they are less opaque when disabled

https://github.com/user-attachments/assets/e749df9a-a94c-4844-ad0e-cbfecf81d08e

## 👀 For the kind reviewers 👀
The actual meat of this PR is in `editor/src/components/editor/actions/actions.tsx` and `editor/src/components/inspector/alignment-buttons.tsx`.

Fixes #6412


editor/src/components/editor/actions/actions.tsx
```diff
@@ -6607,3 +6580,109 @@ function removeErrorMessagesForFile(editor: EditorState, filename: string): Edit
...
+        workingEditorState = apply(justify, 'end')
+        break
+      default:
+        assertNever(alignment)
+    }
+  }
+  return workingEditorState
+}
+
+function alignFlowChildren(
```
ruggi, line 6624: nothing changed here, just factored out from `alignOrDistributeSelectedViews`

bkrmendy (approved):

gbalint (approved):

PR #6414, opened 2024-09-25
Mutation and Resize observers now async the DOM walker.

seanparsons:
**Problem:**
Both the mutation and resize observers can trigger a DOM walker run, which happens immediately and in process of that observer. Which can result in some situations where the DOM walker is running a bit more than necessary with all the implications that can bring.

**Fix:**
With a timeout, we now delay the running of the DOM walker until all of that queued observer processing is complete. Which is also shared between the two observers, so anything that triggers both wont result in 2 runs of the DOM walker where the result of the first is almost immediately discarded.

**Results:**
Before:
Before

After:
After

There's a reduction in the time spent handling the mouse up from around 750ms to 400ms.

**Commit Details:**
- Added `queueUpDomWalker` function that manages the triggering of the dispatch of the dom walker action.
- Observers now call `queueUpDomWalker` so that only one run is triggered at some point soon.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/dom-walker.spec.browser2.tsx
+++ b/editor/src/components/canvas/dom-walker.spec.browser2.tsx
@@ -529,9 +529,9 @@ describe('Observing runtime changes', () => {
     }
 
     await mouseClickAtPoint(buttonToClick, clickPoint)
-    clock.current.tick(10)
+    clock.current.tick(100)
     await renderResult.getDispatchFollowUpActionsFinished()
-    clock.current.tick(10)
+    clock.current.tick(100)
 
     // Check there is metadata for the target at the end
     const metadataAfter = MetadataUtils.findElementByElementPath(


editor/src/core/performance/performance-regression-tests.spec.tsx
```diff
@@ -127,7 +127,7 @@ describe('React Render Count Tests -', () => {
 
     const renderCountAfter = renderResult.getNumberOfRenders()
     // if this breaks, GREAT NEWS but update the test please :)
-    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`989`)
+    expect(renderCountAfter - renderCountBefore).toMatchInlineSnapshot(`898`)
```
balazsbajorics, line 130: interestingly, seems like on the CI these new snapshot values broke and it is indeed 989

balazsbajorics (approved):

editor/src/core/performance/performance-regression-tests.spec.tsx
seanparsons: Yeah, that's weird, reran them and the snapshots updated back again.

gbalint (approved):

PR #6416, opened 2024-09-26
Combined mouse up handling 

seanparsons:
**Problem:**
We have handlers for some events in two places, which causes some upper bounding on performance as both have to process updates separately.

**Fix:**
This attempts to combine the handling so that we don't have a single dispatch flow when that is possible.

**Commit Details:**
- `useSelectOrLiveModeSelectAndHover` now calls the "global" `handleMouseUp` function that gathers actions from potentially multiple functions.
- `EditorCanvas` now hands off it's mouse up handling via the `addMouseUpHandler` utility function.

**Manual Tests:**
I hereby swear that:


ruggi (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
```diff
@@ -667,17 +668,22 @@ function useSelectOrLiveModeSelectAndHover(
         (!hadInteractionSessionThatWasCancelled || !draggedOverThreshold.current) &&
         (activeControl == null || activeControl.type === 'BOUNDING_AREA')
 
+      let editorActions: Array<EditorAction> = []
+
       if (event.type === 'mousedown') {
         didWeHandleMouseDown.current = true
       }
       if (event.type === 'mouseup') {
+        const handleMouseUpActions = handleMouseUp(event.nativeEvent)
```
balazsbajorics, line 677: to make this code clearer here, I think we should call it `handleGlobalMouseUp` or something similar, so it doesn't look like we are just calling off to a local helper

balazsbajorics (approved):

editor/src/components/canvas/controls/select-mode/select-mode-hooks.tsx
seanparsons: Yeah, that's a better name.

seanparsons: Fixed in db2f65791.

gbalint (approved):

PR #6418, opened 2024-09-26
chore(editor): remove duplicate UID check

### LINKED ISSUE #5654: Parser performance improvement: Simplify the hashing function by using full length UUIDs and remove the need to check for clashes

liady:
The WIP PR removed all of the duplicate UID code and checks

**Manual Tests:**
I hereby swear that:

Fixes #5654


liady:
Temporarily closing as Spike

PR #6425, opened 2024-09-27
Use node module resolution

### LINKED ISSUE #6187: we have trouble importing `import splideStyle from '@splidejs/react-splide/css/core';`
the package json has these dependencies:

```
"@splidejs/react-splide": "^0.7.12",
"@splidejs/splide-extension-auto-scroll": "^0.5.3",
```

put the line 

`import splideStyle from '@splidejs/react-splide/css/core';` anywhere in a remix route, observe the error message: 

Image

Rheeseyb:
**Problem**
Utopia's module resolution logic is based on the [2019 node module resolution](https://web.archive.org/web/20190213102857/https://nodejs.org/api/modules.html#modules_all_together), but that logic has significantly evolved since then. This means that there are plenty of cases where the module resolution whilst running a project in Utopia would not match that whilst running elsewhere, causing certain projects to break in unexpected ways.

**Fix:**
I've implemented the latest node.js module resolution logic from https://nodejs.org/api/modules.html#all-together. Since this is of course running in the browser, this also still needs to take into account the [`browser` field of the `package.json`](https://github.com/defunctzombie/package-browser-field-spec), which was already implemented but I wanted to call it out as that is one area where this differs from the node.js spec.

There is still a remaining `FIXME` in here about partial path matching when checking the `imports` and `exports` fields. I'll tackle that in a followup PR.

Fixes #6187


editor/src/core/es-modules/package-manager/module-resolution.ts
```diff
@@ -476,15 +691,75 @@ function resolveModuleAndApplySubstitutions(
   importOrigin: string,
   toImport: string,
 ): FileLookupResult {
-  function lookupFn(path: string[], direct: boolean): FileLookupResult {
-    const nodeModulesResult = nodeModulesFileLookup(nodeModules, path, direct)
-    if (isResolveSuccess(nodeModulesResult)) {
-      return nodeModulesResult
-    } else {
-      return projectContentsFileLookup(projectContents, path, direct)
+  // 1 Check if it's a core module - Node JS only
```
seanparsons, line 694: Could these comments be moved to the respective code?

seanparsons (approved):

editor/src/core/es-modules/package-manager/module-resolution.ts
```diff
@@ -476,15 +691,75 @@ function resolveModuleAndApplySubstitutions(
...
+  // 7 Throw "not found"
+
+  function lookupFn(path: string, innerImportOrigin: string): FileLookupResult {
+    if (path.startsWith('/') || path.startsWith('./') || path.startsWith('../')) {
+      const pathForLookup = path.startsWith('/')
+        ? path
+        : `${stripTrailingSlash(innerImportOrigin)}/${path}`
+      const pathParts = getPartsFromPath(pathForLookup)
+      const lastPart = last(pathParts)
+      const projectContentsFileLookupResult = projectContentsFileLookup(projectContents, pathParts)
```
gbalint, line 713: I think it would make it easier to compare this to the specification if we kept the function names from there. That would be especially useful later down the line when we have to update this according to a change in the spec.
e.g.:
projectContentsFileLookup -> loadAsFile
projectContentsDirectoryLookup -> loadAsDirectory
packageExportsLookup -> loadPackageSelf
etc.

Rheeseyb: Yeah that's a fair comment. I avoided that since this logic isn't actually loading the result, it is just resolving the file, but it makes it easier to track back to the linked algorithm

editor/src/core/es-modules/package-manager/module-resolution.ts
```diff
@@ -1,180 +1,261 @@
...
+// LOAD_INDEX(X)
+function loadIndex(fileForPathParts: FileForPathParts, pathParts: string[]): FileLookupResult {
+  // 1. If X/index.js is a file
+  //   a. Find the closest package scope SCOPE to X.
+  //   b. If no scope was found, load X/index.js as a CommonJS module. STOP.
+  //   c. If the SCOPE/package.json contains "type" field,
+  //     1. If the "type" field is "module", load X/index.js as an ECMAScript module. STOP.
+  //     2. Else, load X/index.js as an CommonJS module. STOP.
+  // Note: We don't attempt to load as an ECMAScript, as we can't - instead we'll try to load whatever
+  // it is as CommonJS, and if it turns out to be an esm we'll attempt to transpile it to CommonJS
```
balazsbajorics, line 171: can you put the comments to the actual code lines?

editor/src/core/es-modules/package-manager/module-resolution.spec.ts
```diff
@@ -51,6 +80,18 @@ const sampleProjectContents: ProjectContentTreeRoot = contentsToTree({
     null,
     0,
   ),
+  '/src/dep-object.js': textFile(
+    textFileContents('export const DepObject = 1', unparsed, RevisionsState.CodeAhead),
```
balazsbajorics, line 84: shouldn't this export an object? :)

editor/src/core/es-modules/package-manager/module-resolution.ts
Rheeseyb: I left these comments here as a reminder that this is how node does this, but since we don't actually use the "type" field, we just load `X/index.js` as is, we don't need to look up the closest package scope here (hence the note at the bottom of the comment)

editor/src/core/es-modules/package-manager/module-resolution.spec.ts
Rheeseyb: Nah this was just for testing the shape of `package.json`. The exported const here is ignored

PR #6428, opened 2024-09-30
TransientActions: making elementsToRerender mandatory

balazsbajorics:
**Problem:**
TransientActions is the inspector's way of creating a continuous interaction. This properly collates all undo events into a single undo step representing the entire series of dispatched transient actions, and it _optionally_ lets you set an array of ElementPaths that will make the editor and the canvas switch to fast rerender mode. The problem is that this elementsToRerender path is optional, which means some inspector interactions will be slow!

**Fix:**
Make elementsToRerender mandatory and fill it in all places.

### DIFF
--- a/editor/src/templates/editor.tsx
+++ b/editor/src/templates/editor.tsx
@@ -159,7 +159,7 @@ function collectElementsToRerenderForTransientActions(
   action: EditorAction,
 ): Array<ElementPath> {
   if (action.action === 'TRANSIENT_ACTIONS') {
-    if (action.elementsToRerender != null) {
+    if (action.elementsToRerender !== 'rerender-all-elements') {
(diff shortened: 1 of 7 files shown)


editor/src/templates/editor.tsx
```diff
@@ -159,7 +159,7 @@ function collectElementsToRerenderForTransientActions(
   action: EditorAction,
 ): Array<ElementPath> {
   if (action.action === 'TRANSIENT_ACTIONS') {
-    if (action.elementsToRerender != null) {
+    if (action.elementsToRerender !== 'rerender-all-elements') {
```
bkrmendy, line 162: this if is always true

liady (approved):

seanparsons (approved):

PR #6443, opened 2024-10-01
Specify elements to rerender in grid strategies

bkrmendy:
## Problem
The grid strategies opt to rerender all elements in their strategy result, which is adds a performance penalty.

## Fix
Only re-render the elements affected by a given strategy


ruggi (approved):

seanparsons (approved):
Lovely.

PR #6445, opened 2024-10-02
fix(css): dont set dimensions on canvas element

liady:
**Problem:**
Today all css declarations on `body` or `html` are being applied to `#canvas-element`, including dimesions declarations (width, height, etc). This can cause weird canvas appearance.

**Fix:**
Filter out dimesions declarations when applying styles to the canvas element.
This is a part of the larger Vite import task, but it's a general fix.

**Manual Tests:**
I hereby swear that:

- [x] I could navigate to various routes in Play mode.

### DIFF
--- a/editor/src/core/shared/css-utils.ts
+++ b/editor/src/core/shared/css-utils.ts
@@ -28,13 +28,28 @@ export function rescopeCSSToTargetCanvasOnly(input: string): string {
 
   let ast = csstree.parse(scopedInput)
 
-  csstree.walk(ast, (node, item, list) => {
+  csstree.walk(ast, function (node, item, list) {
     // As we are wrapping in an @scope, we need to redirect certain selectors to :scope
     if (isSelectorToChange(node) && list != null) {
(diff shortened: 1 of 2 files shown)


editor/src/core/shared/css-utils.ts
```diff
@@ -28,13 +28,28 @@ export function rescopeCSSToTargetCanvasOnly(input: string): string {
...
+      removeDimensionsFromCssRule(this.rule)
     }
   })
 
   return csstree.generate(ast)
 }
+
+const propertiesToRemove = ['width', 'height', 'max-width', 'max-height', 'min-width', 'min-height']
+function removeDimensionsFromCssRule(rule: csstree.Rule | null): void {
+  if (rule != null) {
```
bkrmendy, line 46: I'd remove the `| null` from the param and move the null check out of the function

liady: yup. done

balazsbajorics (approved):

bkrmendy (approved):

PR #6408, opened 2024-09-24
Mandatory Elements To Rerender For Strategies

seanparsons:
**Problem:**
When adding new canvas strategies it is very easy to forget to use the `setElementsToRerenderCommand` when required (which is most of the time).

**Fix:**
Now instead of commands being the source of the elements to rerender, now that value is returned specifically by the strategy and is required to be passed to `strategyApplicationResult`. That value is then applied to the editor state taking it from the `StrategyApplicationResult`.

**Commit Details:**
- Added `combineElementsToRerender` utility function.
- Added `appendElementsToRerenderToApplyResult` utility function.
- Added `StrategyApplicationResult.elementsToRerender`.
- Removed `setElementsToRerenderCommand` and `appendElementsToRerenderCommand` and the associated supporting code.
- Moved values originally passed to `setElementsToRerenderCommand` to the new parameter that `strategyApplicationResult` now has.
- Moved values originally passed to `appendElementsToRerenderCommand` to a call to `appendElementsToRerenderToApplyResult`.
- Added `applyElementsToRerenderFromStrategyResult` utility function.
- Hooked in the `applyElementsToRerenderFromStrategyResult` function
  into various strategy functions to update the `EditorState`.

**Manual Tests:**
I hereby swear that:

- [x] I opened a hydrogen project and it loaded
- [x] I could navigate to various routes in Preview mode

Fixes #6372

PR #2755, opened 2022-11-03
Feature/gap-controls

bkrmendy:
Fixes #[ticket_number]

## Problem:
Flex gap cannot be adjusted via the canvas

## Fix:
Implement controls for flex gap along the lines of padding controls

PR #6446, opened 2024-10-02
Specify elements to rerender in keyboard & gap strategies

bkrmendy:
## Problem
The keyboard font strategies and the gap setter strategies don't correctly declare the elements to rerender in the strategy application result

## Fix
Fill it in with the correct element path array

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/set-flex-gap-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/set-flex-gap-strategy.tsx
@@ -174,10 +174,7 @@ export const setFlexGapStrategy: CanvasStrategyFactory = (
       if (shouldTearOffGap) {
         return strategyApplicationResult(
           [deleteProperties('always', selectedElement, [StyleGapProp])],
-          // FIXME: This was added as a default value in https://github.com/concrete-utopia/utopia/pull/6408
-          // This was to maintain the existing behaviour, but it should be replaced with a more specific value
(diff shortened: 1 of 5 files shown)


ruggi (approved):

editor/src/components/canvas/canvas-strategies/strategies/set-flex-gap-strategy.tsx
```diff
@@ -200,7 +197,7 @@ export const setFlexGapStrategy: CanvasStrategyFactory = (
             },
           ]),
         ],
-        [...selectedElements, ...children.map((c) => c.elementPath)],
+        selectedElements,
```
balazsbajorics, line 200: do you remember why were the children included here to begin with? I'm afraid this was to fix a behavior.

seanparsons (approved):

editor/src/components/canvas/canvas-strategies/strategies/set-flex-gap-strategy.tsx
bkrmendy: this line was taken from the original `setElementsToRerender` (see https://github.com/concrete-utopia/utopia/pull/6408), and it was added with the original implementation of the control (https://github.com/concrete-utopia/utopia/pull/2755).

We don't actually need to explicitly pass the children paths here, because they aren't affected by the strategy

balazsbajorics (approved):

PR #6466, opened 2024-10-03
Revert "Switch Event Handlers"

seanparsons:
Reverts concrete-utopia/utopia#6385


balazsbajorics:
the reason for the revert was two bugs discovered by @gbalint:
1. the navigator drag and drop stopped working
2. clicking into the navigator did not steal focus from VSCode, so pressing backspace to delete a selected element would delete a character in VSCode instead

PR #6472, opened 2024-10-04
Better grid rearrange strategies

gbalint:
**Problem:**
The grid rearrange strategies were stateful: they used the grid custom strategy state to store the cell under the mouse at the beginning, and used that to check how much the element has to be moved. It also cached the last cell under the mouse, so the element does not jump back to its original position while you are over a gap between cells.

**Fix:**
I modified the strategy so the behavior does not rely on the custom strategy state. It still writes into the strategy state though, because that is the input for the grid control and the cell animation, but the strategy result does not depend on it. (migrating those to prop controls can come in a subsequent PR)

The new algorithm:
- strategy checks the grid cell position of the drag start mouse position in the grid child
- strategy uses the offset from above to get the new grid cell coords of the elements compared to the current mouse position

**Changes**:
- I needed a way to find a cell under the mouse which returns the closest cell even when you are over a gap. (`getClosestGridCellToPointFromMetadata`)
- Removed `draggingFromCell` and `originalRootCell` from `GridCustomStrategyState`
- Custom state is not an input to `runGridRearrangeMove` anymore

**Manual Tests:**
I hereby swear that:


bkrmendy (approved):

editor/src/components/canvas/canvas-strategies/strategies/grid-cell-bounds.ts
```diff
@@ -57,16 +58,50 @@ function getGridCellUnderPoint(
...
+export function getClosestGridCellToPoint(
+  gridCellGlobalFrames: GridCellGlobalFrames,
+  point: CanvasPoint,
+): TargetGridCellData | null {
+  let closestCell: TargetGridCellData | null = null
+  let closestDistance = Infinity
+
+  for (let i = 0; i < gridCellGlobalFrames.length; i++) {
+    for (let j = 0; j < gridCellGlobalFrames[i].length; j++) {
+      if (distanceFromPointToRectangle(point, gridCellGlobalFrames[i][j]) < closestDistance) {
```
ruggi, line 83: I would store the calculated distance into a variable here

ruggi (approved):

seanparsons (approved):

editor/src/core/shared/math-utils.ts
```diff
@@ -408,6 +408,32 @@ export function distance<C extends CoordinateMarker>(from: Point<C>, to: Point<C
   return magnitude({ x: from.x - to.x, y: from.y - to.y } as Vector<C>)
 }
 
+export function distanceFromPointToRectangle<C extends CoordinateMarker>(
```
seanparsons, line 411: I had to think this through when reading it, but this is nice.

PR #6476, opened 2024-10-05
fix(parse): don't lookup non-parsable files in the cache

liady:
**Problem:**
Today we try to read every file from the cache, even if it's non-parsable, and always get a cache miss (since we don't parse it and thus not storing anything). We shouldn't try and access the cache for non-parsable files at all.
image

**Fix:**
Ignore cache for non-parsable files.

**Commit Details:**
- Other than tidying the code, the only change is to also check for `isParsableFile()` inside `shouldUseCacheForFile()`. 
- As part of this change, since I now import `isParsableFile`, I also moved `gitBlobChecksumFromBuffer` out of `assets.ts` and into `file-utils.ts`. `file-utils` is in `shared` but was importing it from the editor's `assets.ts`, and along with it the entire editor code, unnecessarily.

**Manual Tests:**
I hereby swear that:


editor/src/core/workers/parser-printer/parse-cache-utils.worker.ts
```diff
@@ -33,15 +34,22 @@ function logCacheMessage(parsingCacheOptions: ParseCacheOptions, ...messages: (s
...
-  if (filename === ARBITRARY_CODE_FILE_NAME) {
+  if (isArbitraryCodeFile(filename)) {
     return ['code', [{ content: content }]]
   }
   return [filename]
 }
 
-function shouldSkipCacheForFile(filename: string, parsingCacheOptions: ParseCacheOptions): boolean {
-  return filename === ARBITRARY_CODE_FILE_NAME && !parsingCacheOptions.cacheArbitraryCode
+function shouldUseCacheForFile(filename: string, parsingCacheOptions: ParseCacheOptions): boolean {
```
liady, line 48: this is the main change - checking also for `isParsableFile` inside `shouldUseCacheForFile`

editor/src/core/workers/parser-printer/parse-cache-utils.worker.ts
```diff
@@ -33,15 +34,22 @@ function logCacheMessage(parsingCacheOptions: ParseCacheOptions, ...messages: (s
...
+  return (
+    isParseableFile(filename) &&
```
liady, line 50: this is the main change - checking also for `isParsableFile` inside `shouldUseCacheForFile`

bkrmendy (approved):

Rheeseyb (approved):

seanparsons (approved):

PR #6477, opened 2024-10-05
chore(types): remove duplicate type definitions

liady:
**Problem:**
Today we have shared types between the editor and our Remix code (`utopia-shared`), however some of these definitions are still duplicated in the editor code (even if identical), which can lead to divergence over time.

**Fix:**
Make sure the duplicated types are identical and remove them, reference `utopia-shared` as the source of truth.
Actual code (functions) was not moved, only types, since `utopia-shared` should be kept separated from editor code and imports.

**Note:**
In the future this can allow us to have `utopia-shared` code hash (or package version) as the cache version for `ParseSuccess` objects.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/utopia-shared/src/types/project-file-types.ts
+++ b/utopia-shared/src/types/project-file-types.ts
@@ -74,6 +74,7 @@ export interface ExportVariablesWithModifier {
   variables: Array<string>
 }
 
+// export function functionName(){...}
 export interface ExportFunction {
   type: 'EXPORT_FUNCTION'
   functionName: string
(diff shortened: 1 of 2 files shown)


bkrmendy (approved):

utopia-shared/src/types/project-file-types.ts
```diff
@@ -74,6 +74,7 @@ export interface ExportVariablesWithModifier {
   variables: Array<string>
 }
 
+// export function functionName(){...}
```
seanparsons, line 77: Huh?

seanparsons (approved):

PR #6484, opened 2024-10-07
Correctly Aligned Grid Controls

seanparsons:
**Problem:**
It was initially found that adding `alignContent: 'center'` causes a skew in the placeholder boundaries, but this also affects the resize handles.

**Fix:**
This involves several fixes:
- Changing out `border` for `outline` in `GridControl` so that it would not affect the content held within and removed the slight shifting of that element to cater for the `border`.
- Being more permissive with what is allowed in `alignContent` as there are values which were being filtered out.

**Commit Details:**
- `alignContent` in `getSpecialMeasurements` now isn't filtered by `getFlexJustifyContent`.
- Added `justifyContent` and `alignContent` to `GridResizingProps`.
- In `GridResizing`, the main div now includes `paddingRight`, `paddingBottom`,
  `justifyContent` and `alignContent`.
- `GridControl` now uses `range` to build the `placeholders` value.
- Removed position shifting from `GridControl` and changed the `border` to
  an `outline` as those were causing the content to be repositioned.
- Added `range` utility function.
- `SpecialSizeMeasurements.alignContent` is now typed to `string | null` instead
  of `FlexJustifyContent | null` as that type is wildly incorrect.

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -892,19 +908,19 @@ export const GridControl = React.memo<GridControlProps>(({ grid }) => {
     gridPath: gridPath,
   })
 
-  const placeholders = Array.from(Array(grid.cells).keys())
+  const placeholders = range(0, grid.cells)
```
ruggi, line 911: 🪄

ruggi (approved):

bkrmendy (approved):

balazsbajorics (approved):

PR #6486, opened 2024-10-07
Don't show grid resize handles on top of scene labels

### LINKED ISSUE #6485: don't display grid resize handles on top of scene labels

ruggi:
**Problem:**

The grid resize handles are shown on top of the scene label.

image

**Fix:**

Move the handles north of the scene label if the grid is also a scene.

Screenshot 2024-10-07 at 14 04 32

Fixes #6485

### DIFF
--- a/editor/src/components/canvas/controls/grid-controls.tsx
+++ b/editor/src/components/canvas/controls/grid-controls.tsx
@@ -498,6 +498,7 @@ export type GridData = {
   columns: number
   cells: number
   metadata: ElementInstanceMetadata
+  isScene: boolean
 }
 
 export function useGridData(elementPaths: ElementPath[]): GridData[] {
@@ -557,6 +558,9 @@ export function useGridData(elementPaths: ElementPath[]): GridData[] {
           rows: rows,
           columns: columns,
           cells: rows * columns,
+          isScene:
+            MetadataUtils.isProbablySceneFromMetadata(targetGridContainer) ||
+            MetadataUtils.isProbablyRemixSceneFromMetadata(targetGridContainer),
         }
       }, elementPaths)
     },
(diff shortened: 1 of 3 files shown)


editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -1848,3 +1857,9 @@ export function controlsForGridPlaceholders(gridPath: ElementPath): ControlWithP
     priority: 'bottom',
   }
 }
+
+function sceneTopOffset(canvasScale: number) {
+  const labelHeight = document.getElementById(SceneLabelTestID)?.getBoundingClientRect().height ?? 0
```
Rheeseyb, line 1862: I don't remember if there was a reason why we didn't give the scene label an explicit height (I just checked assuming we would have). I think it would make sense to do that and then export it as a const rather than querying the DOM here

ruggi: I thought so too, but now I see its offset is calculated based on the font size, so I did not want to change that. But of course I see the point so maybe it's worth simplifying that whole thing and just give it an explicit height.

ruggi: done in [`74a7c44` (#6486)](https://github.com/concrete-utopia/utopia/pull/6486/commits/74a7c44bae86e65d98b25598ac309b5c0dce1a6c)

seanparsons (approved):

Rheeseyb (approved):

editor/src/components/canvas/controls/grid-controls.tsx
Rheeseyb: Thank you!

PR #6490, opened 2024-10-07
No more stateful grid strategies

gbalint:
**Description:**
I removed custom state strategy usage from grid reparent strategy:
- the target cell was cached in custom strategy state to make sure we can reparent over gaps between cells. I removed this and just reparent into the closest cell.

Removed custom strategy state usage from grid controls, and moved the necessary information to `EditorStateCanvasControls`. Instead of custom strategy state patches i needed to return showGridControls commands.

Removed all fields from `GridCustomStrategyState` except `metadataCacheForGrids`, which is still necessary

**Manual Tests:**
I hereby swear that:


bkrmendy (approved):

seanparsons (approved):
Feels so much cleaner!

ruggi (approved):

PR #6492, opened 2024-10-08
Style plugins - tracking PR

bkrmendy:
🔌


editor/src/components/canvas/gap-utils.ts
```diff
@@ -328,39 +329,17 @@ export interface FlexGapData {
...
-  elementPath: ElementPath,
-): FlexGapData | null {
-  const element = MetadataUtils.findElementByElementPath(metadata, elementPath)
-  if (
-    element == null ||
-    element.specialSizeMeasurements.display !== 'flex' ||
-    isLeft(element.element) ||
-    !isJSXElement(element.element.value)
-  ) {
+export function getFlexData(info: StyleInfo): FlexGapData | null {
```
balazsbajorics, line 332:
```suggestion
export function getFlexGapData(info: StyleInfo): FlexGapData | null {
```

editor/src/components/canvas/plugins/inline-style-plugin.ts
```diff
@@ -0,0 +1,66 @@
...
+    undefined,
+    getLayoutProperty('gap', right(element.element.value.props), styleStringInArray),
+  )
+
+  const flexDirection = element.specialSizeMeasurements.flexDirection ?? 'row'
+
+  return {
+    value: {
+      renderedValuePx: gap,
+      value: gapFromProps ?? null,
```
balazsbajorics, line 45: do you plan to standardize these values across the properties? here we have `renderedValuePx` – coming from the specialSizeMeasurements, and the gapFromProps coming from inline style. do we want to bring specialSizeMeasurements to the mix here?

alternatively, referencing `specialSizeMeasurements` could remain in `getFlexGapData` in gap-utils.ts: it could read the style info and specialSizeMeasurements and combine it.

I have a feeling that this inline-style-plugin should only deal with getting values from inline style, and we should only change the existing strategy code to the level of replacing all inline style references with the StyleInfo lookups.

bkrmendy: This an interesting question and we'll have to make a tradeoff no matter which route we choose. The problem is the following: the strategies need to rendered value in pixels so they can calculate the delta when a handle is dragged. This code needs to be written somewhere, so the decision is around where we put it.

- Pros of not involving `specialSizeMeasurements` in the plugins: the plugin code is shorter and it's more scoped to the given styling system.
- Cons of not involving `specialSizeMeasurements` in the plugin: the element styles provided by `styleInfoReader` are not a one-stop shop, so the strategies still need to piece together the info they need from `styleInfoReader` + the metadata. Doing that in the plugin makes the strategy code cleaner, and the logic that reads from `specialSizeMeasurements` can be abstracted out anyway, to be called from anywhere

My main argument for keeping `specialSizeMeasurements` out of the strategy is to make it possible for the strategy code to focus on feature logic as much as possible, and let the underlying code provide the data it has to work with. This way the strategies stay easy to tweak/review

editor/src/components/canvas/gap-utils.ts
bkrmendy: https://github.com/concrete-utopia/utopia/pull/6492/commits/450130409ada79698d36596d9c74ca7aecec24b6

bkrmendy:
implemented in PRs split out from this one

PR #6497, opened 2024-10-08
Store the Tailwind config in the editor state

bkrmendy:
split out from https://github.com/concrete-utopia/utopia/pull/6492

## Problem
The Tailwind config is only accessible inside `useTailwindCompilation`, which means that the upcoming element style plugins won't be able to access it

## Fix
Store the Tailwind config in the editor state, the same way that we store the property control descriptors. This way, the tailwind config will be accessible from hooks, actions, the strategy dispatch etc.

### DIFF
--- a/editor/package.json
+++ b/editor/package.json
@@ -424,6 +424,7 @@
     "source-map-loader": "0.2.3",
     "string-replace-loader": "2.2.0",
     "style-loader": "0.18.2",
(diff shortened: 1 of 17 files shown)


bkrmendy:
closed in favor of https://github.com/concrete-utopia/utopia/pull/6500

PR #6499, opened 2024-10-08
Better grid reparent

gbalint:
**Problem:**
One of the root causes of some major issues is that the grid reparent strategies were not created by the general reparent-metastrategy, but the grid-rearrange-move strategy itself.

**Fix:**
- Removed reparent strategy creation from grid-rearrange-move. After this grid-rearrange-move is always selectable when dragging from a grid, even when outside of a grid (so you can manually switch to this strategy and the child will jump back to the original grid, into the cell which is closest to the mouse). This is a behavior which is similar to flex reorder.
- reparent-metastrategy can create a reparent into grid strategy. To achieve this I just had to allow `GRID_CELL_HANDLE` controls in the strategy factory.
- Fortunately reparent-metastrategy does not allow reparenting into the same parent, so this automagically solved the problem that reparenting inside the original grid should just do a rearrange (because the reparent strategy is not even selectable in that case)
- With this solution, when reparenting from one grid to a different grid, the grid controls were shown on both the original and the new parent grid too. Unfortunately, the grid controls do not support showing the controls on multiple grid instances at the same time (e.g. they shared the state that which cell was highlighted), so I had to switch the grid control of the rearrange strategy to `visible-only-while-active` (so it is not visible on the original grid when the reparent strategy is active)
- I had to make sure the grid specific style props are removed during all kinds of reparent (earlier this was manually handled in the grid specific code)
- I could remove plenty of reparent specific code lines from `grid-rearrange-move-strategy.ts`
- In the earlier solution reparent to flow was the default when repareing into a flow target, but in our regular reparent flow reparent is never the default. Because of this I had to rewrite one test to select the flow strategy manually.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/drag-to-move-metastrategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/drag-to-move-metastrategy.tsx
@@ -58,7 +58,8 @@ export const dragToMoveMetaStrategy: MetaCanvasStrategy = (
     selectedElements.length === 0 ||
     selectedElements.some(EP.isRootElementOfInstance) ||
     interactionSession == null ||
-    interactionSession.activeControl.type !== 'BOUNDING_AREA' ||
+    (interactionSession.activeControl.type !== 'BOUNDING_AREA' &&
+      interactionSession.activeControl.type !== 'GRID_CELL_HANDLE') ||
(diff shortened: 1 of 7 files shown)


bkrmendy (approved):

ruggi (approved):

balazsbajorics (approved):
beautiful!

PR #6500, opened 2024-10-09
Memoized Tailwind config evaluation

bkrmendy:
## Problem
The Tailwind config is only accessible inside `useTailwindCompilation`, which means that the upcoming element style plugins won't be able to access it

## Fix
Store the last seen text contents of the Tailwind config file along with the last evaluation result. If the new contents of the Tailwind config file are the same as the last seen contents, the last evaluation result, otherwise, re-evaluate the Tailwind config file, and store it, and return it


gbalint (approved):

editor/src/core/tailwind/tailwind-compilation.ts
```diff
@@ -16,6 +16,30 @@ import type { RequireFn } from '../shared/npm-dependency-types'
...
+  }
+  const cached =
+    LatestConfig.current == null || LatestConfig.current.code !== tailwindConfig.fileContents.code
+      ? null
+      : LatestConfig.current.config
+
+  if (cached != null) {
+    return cached
+  }
+  const { customRequire } = createRequireFn(editorState, TailwindConfigPath)
```
balazsbajorics, line 37: let's add a FIXME here that this should use a shared long-lived require function instead of creating a brand new one

editor/src/core/tailwind/tailwind-compilation.ts
```diff
@@ -16,6 +16,30 @@ import type { RequireFn } from '../shared/npm-dependency-types'
 import { TailwindConfigPath } from './tailwind-config'
 import { ElementsToRerenderGLOBAL } from '../../components/canvas/ui-jsx-canvas'
 import { isFeatureEnabled } from '../../utils/feature-switches'
+import type { Config } from 'tailwindcss/types/config'
+import type { EditorState } from '../../components/editor/store/editor-state'
+import { createRequireFn } from '../property-controls/property-controls-local'
+
+const LatestConfig: { current: { code: string; config: Config } | null } = { current: null }
+export function getTailwindConfig(editorState: EditorState): Config | null {
```
balazsbajorics, line 24: maybe putting cached in the name helps us remember that there is memoization happening here
```suggestion
export function getTailwindConfigCached(editorState: EditorState): Config | null {
```

balazsbajorics (approved):

editor/src/core/tailwind/tailwind-compilation.ts
bkrmendy: https://github.com/concrete-utopia/utopia/pull/6500/commits/0b2a8404813c53b004a1dbb2566c627c9e3ed596

editor/src/core/tailwind/tailwind-compilation.ts
bkrmendy: https://github.com/concrete-utopia/utopia/pull/6500/commits/df49f33e57e08514accf914bed2bda2d45f1d20e

PR #6505, opened 2024-10-09
Grid reparent uses rearrange helper

gbalint:
**Description:**
Grid reparent uses the grid rearrange helper to set the grid specific properties.

When you reparent into a grid container, moving the element to the target cell, setting the grid control properties to support animation, etc. is really similar to grid rearrange. Despite that, the two strategies did not share code.

I modified the grid reparent strategy to call runGridRearrangeMove. This automatically added animation to grid reparent, and fixed some jumping around.

- `runGridRearrangeMove` has a new parameter for the grid metadata (before that it could assume that the grid is parent of the selected element, but during reparent this is not true)
- some of the `runGridRearrangeMove` which handles multicell elements is not needed for grid reparent, I needed to handle that (e.g. during grid reparent we always put the element into a single target cell, so no need to handle the original size in cells, the offset between the dragged cell and root cell, etc)
- Removed `getGridPositioningCommands` which is replaced by `runGridRearrangeMove`

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
@@ -57,7 +57,7 @@ export function runGridRearrangeMove(
     return []
   }
 
-  const isReparent = !EP.pathsEqual(EP.parentPath(selectedElement), grid.elementPath)
+  const isReparent = !EP.isParentOf(grid.elementPath, selectedElement)
 
   const { gridCellGlobalFrames, containerGridProperties: gridTemplate } =
     grid.specialSizeMeasurements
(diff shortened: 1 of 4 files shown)


balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
```diff
@@ -51,17 +51,13 @@ export function runGridRearrangeMove(
...
     return []
   }
 
-  const parentGridPath = EP.parentPath(selectedElement)
-  const grid = MetadataUtils.findElementByElementPath(jsxMetadata, parentGridPath)
-
-  if (grid == null) {
-    return []
-  }
+  const isReparent = !EP.pathsEqual(EP.parentPath(selectedElement), grid.elementPath)
```
seanparsons, line 60: I think this is equivalent:
```suggestion
  const isReparent = !EP.isParentOf(grid.elementPath, selectedElement)
```

ruggi (approved):

seanparsons (approved):

PR #6506, opened 2024-10-09
Update class list command

bkrmendy:
## Problem

In order to add/remove/update tailwind classes on an element, we need a way to edit the `className` property in a granular way. 

## Constraints/considerations
- we can't treat the className property as a string, because we want to edit the list of classnames in it
- we can't treat the className property as an array of strings, because Tailwind classes have their own syntax (negative values, selectors) that we need to respect
- not all classes in className property are parseable, Tailwind class name updates should leave the non-tailwind classes alone

## Fix
This PR introduces `tailwind-class-list-utils`, which provides a toolkit to edit the Tailwind classes in the className property. On top of that, that PR adds a command that provides a high-level wrapper for `tailwind-class-list-utils` that the rest of the editor can use


editor/src/core/tailwind/tailwind-class-list-utils.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+          Object {
+            "property": "margin",
+            "type": "parsed",
+            "value": "1rem",
+          },
+          Object {
+            "property": "width",
+            "type": "parsed",
+            "value": "1rem",
+          },
```
balazsbajorics, line 49: this doesn't capture the `lg:` prefix. is that going to lead to data loss when we manipulate the entries?

bkrmendy: updated the tests here https://github.com/concrete-utopia/utopia/pull/6506/commits/97f400a7e10d410a10a7f6d5237b540fc8700885 to make it explicit that we do preserve the prefixes

balazsbajorics (approved):
this is very nice!

gbalint (approved):

editor/src/core/tailwind/tailwind-class-list-utils.ts
```diff
@@ -0,0 +1,106 @@
...
+      return {
+        type: 'parsed',
+        ast: { property: cls.ast.property, value: updatedProperty, variants: [], negative: false },
+      }
+    })
+    return classListWithUpdatedClasses
+  }
+
+export const removeClasses =
+  (propertiesToRemove: string[]): ClassListTransform =>
```
ruggi, line 96: I feel it would be better to have this as a `Set` directly instead of going `string[] → Set`, but ¯\_(ツ)_/¯

PR #6507, opened 2024-10-09
feat(import): github import wizard

liady:
This PR adds a Github import wizard, tracking the progress of importing a project from Github, showing errors or warnings. The design is still preliminary, but it is hidden behind a feature switch so that the logic can be reviewed and merges, in order not to create a long living PR.
<video src="https://github.com/user-attachments/assets/65c484b2-815a-4c07-8fb2-2346f06a466e"></video>

It supports error/warning states:
image

## PR Includes:
1. Adding three actions
    - `SetImportWizardOpen` - controlling the wizard UI
    - `UpdateImportOperations` - updating the results of individual import operations (desc. to follow).
    - `UpdateProjectRequirements` - updating the results of checking and fixing Utopia specific requirements
2. Utilities to start/end tracking of import operations, their results and timings.
3.  Tracked import operations - tracking hooks are being called at appropriate places in the code.
    - Loading the code from Github
    - Parsing the files
    - Checking specific Utopia requirements
    - Loading dependencies
4. Utopia checks:
    - storyboard file (created if not present)
    - `utopia` entry in package.json
    - project composition (JS/TS)
 5. Not tracked in this PR but will in subsequent ones (can be added modularily):
    - React version - currently a placeholder
    - Server side packages - mainly Node builtins
    - Style frameworks
 6. The wizard UI itself - showing progress and results (still preliminary)

**Note for reviewers:** there are a lot of changed files due to adding new actions and passing down the `dispatch` parameter, but the main logic is contained in the new services and the wizard React components.

**Manual Tests:**
I hereby swear that:


editor/src/components/editor/editor-component.tsx
```diff
@@ -677,12 +679,19 @@ const LockedOverlay = React.memo(() => {
 
   const editorLocked = React.useMemo(() => githubOperations.length > 0, [githubOperations])
 
-  const refreshingDependencies = useEditorState(
+  const refreshingDependencies = false
+  useEditorState(
```
bkrmendy, line 683: The result of the `useEditorState` call is discarded here, was this an intentional change?

editor/src/components/editor/import-wizard/components.tsx
```diff
@@ -0,0 +1,320 @@
...
+  ImportOperation,
+} from '../../../core/shared/import/import-operation-types'
+import { ImportOperationResult } from '../../../core/shared/import/import-operation-types'
+import { assertNever } from '../../../core/shared/utils'
+import { Icons } from '../../../uuiui'
+import { GithubSpinner } from '../../../components/navigator/left-pane/github-pane/github-spinner'
+import { RequirementResolutionResult } from '../../../core/shared/import/proejct-health-check/utopia-requirements-types'
+
+export function OperationLine({ operation }: { operation: ImportOperation }) {
+  const operationRunningStatus = React.useMemo(() => {
```
bkrmendy, line 17: the `useMemo` is not necessary here, because the return value is a string, which has value equality semantics by default

editor/src/core/es-modules/package-manager/fetch-packages.ts
```diff
@@ -287,13 +292,31 @@ export async function fetchNodeModules(
   const nodeModulesArr = await Promise.all(
     dependenciesToDownload.map(
       async (newDep): Promise<Either<DependencyFetchError, NodeModules>> => {
+        function notifyFetchEnd(result: ImportOperationResult) {
```
bkrmendy, line 295: an alternaite approach here would be to wrap `fetchNodeModules` into a higher-order function that calls `notifyOperationFinished`. That way `notifyFetchEnd` wouldn't have to be added to each early return

editor/src/core/shared/import/import-operation-service.ts
```diff
@@ -0,0 +1,146 @@
...
+      ImportOperationAction.Replace,
+    ),
+  ]
+  if (isFeatureEnabled('Import Wizard')) {
+    actions.push(setImportWizardOpen(true))
+  }
+  dispatch(actions)
+}
+
+export function showImportWizard() {
```
bkrmendy, line 36: this seems to be unused

editor/src/core/shared/import/import-operation-service.ts
```diff
@@ -0,0 +1,146 @@
...
+} from '../../../components/editor/actions/action-creators'
+import { ImportOperationAction } from './import-operation-types'
+import type {
+  ImportOperation,
+  ImportOperationResult,
+  ImportOperationType,
+} from './import-operation-types'
+import { isFeatureEnabled } from '../../../utils/feature-switches'
+
+let editorDispatch: EditorDispatch | null = null
```
bkrmendy, line 15: The dispatch function is available in the code almost everywhere, so I'd try and get rid of the global state here and check whether the dispatch function can be passed to functions in this file

liady: got it. is the preferred practice to pass it down? (asking since `fetchNodeModules`, for example, is being called from several places that don't have this `dispatch` function available)

editor/src/components/editor/editor-component.tsx
liady: no, it's an error! (should have been assigned to `refreshingDependencies` instead of `false` which I only used temporarily before the feature switch. Thank you!

editor/src/core/shared/import/import-operation-service.ts
bkrmendy: I'd prefer passing it down because I'd avoid global state as much as possible, but in terms of functionality if doesn't really matter, since the dispatch function always stays the same (afaik)

editor/src/core/es-modules/package-manager/fetch-packages.ts
Rheeseyb: I believe the point here is to call `notifyOperationFinished` as and when each of these inner promises complete

Rheeseyb: This inner anon function could however be extracted and wrapped in that way though

editor/src/core/shared/import/import-operation-service.ts
Rheeseyb: Yes please, +1 to avoiding global state. If nothing else, it makes testing a nightmare

editor/src/core/shared/import/proejct-health-check/check-utopia-requirements.ts
```diff
@@ -0,0 +1,128 @@
...
+      'packageJsonEntries',
+      RequirementResolutionResult.Found,
+      'Valid package.json found',
+    )
+  }
+
+  return projectContents
+}
+
+function checkProjectLanguage(projectContents: ProjectContentTreeRoot): ProjectContentTreeRoot {
```
Rheeseyb, line 76: There's no need for this to return anything

editor/src/core/shared/import/proejct-health-check/check-utopia-requirements.ts
```diff
@@ -0,0 +1,128 @@
...
+  let tsCount = 0
+  applyToAllUIJSFiles(projectContents, (filename, uiJSFile) => {
+    if (filename.endsWith('.ts') || filename.endsWith('.tsx')) {
+      tsCount++
+    } else if (filename.endsWith('.js') || filename.endsWith('.jsx')) {
+      jsCount++
+    }
+    return uiJSFile
+  })
+  if (tsCount > jsCount) {
```
Rheeseyb, line 88: Would we ever want to pass this check if `tsCount > 0`? I can't think of a reason why we would 🤔

liady: yes I was trying to be a bit permissive here with an rough heuristic, in case they have several `.ts` files that are maybe unrelated to the main app flow. But if we're giving them the option to "continue anyway" even with errors, then they can maybe mark this as a non-issue if that's the case. I'll change to `>0`.

editor/src/components/editor/import-wizard/components.tsx
liady: I was actually trying to save the internal comparisons from running every render if they don't need to (similar to what a computed field would have done)

editor/src/core/shared/import/import-operation-service.ts
liady: fixed it here - https://github.com/concrete-utopia/utopia/pull/6507/commits/36cadbab90aef63b1b1c788ee73b3584fb24a8f8, it required a refactor of multiple files (some are unrelated to the actual functionality), but now the `dispatch` is passed down instead of being stored as a state

editor/src/components/editor/store/editor-state.ts
```diff
@@ -1452,6 +1457,9 @@ export interface EditorState {
   githubSettings: ProjectGithubSettings
   imageDragSessionState: ImageDragSessionState
   githubOperations: Array<GithubOperation>
+  importOperations: Array<ImportOperation>
```
seanparsons, line 1460: I wonder if these should be under some umbrella field rather than at the root of `EditorState`.

editor/src/components/editor/import-wizard/components.tsx
```diff
@@ -0,0 +1,320 @@
...
+      assertNever(operation)
+  }
+}
+
+function getTextColor(
+  operationRunningStatus: 'waiting' | 'running' | 'done',
+  operation: ImportOperation,
+) {
+  if (operationRunningStatus === 'waiting') {
+    return 'gray'
```
seanparsons, line 291: Should this be using our color themes? As I would surmise that it will look a little weird in dark mode.

editor/src/components/editor/import-wizard/components.tsx
```diff
@@ -0,0 +1,320 @@
...
+      css={{
+        '.import-wizard-operation-children > &': {
+          paddingLeft: 26,
+          fontSize: 12,
+          img: {
+            width: 12,
+            height: 12,
+          },
+        },
+        '.import-wizard-operation-children .operation-done [data-short-time=true]': {
```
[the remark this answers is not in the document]
liady: good point! however the design is still temporary (and pending @lankaukk 's insights) - including whether or not to even have these kind of timings..
this is also the reason it's currently behind a (turned-off) flag - so we could merge the logic and iterate on the design later on..

editor/src/components/editor/import-wizard/components.tsx
liady: yes, definitely should come from the theme... this is also temporary design and definitely going to be changed in the design iteration

editor/src/components/editor/store/editor-state.ts
liady: interesting point! which field, for example? I followed the `githubOperations` line, which also describe a temporary chain of operations, very similar to the `import` one

bkrmendy (approved):

balazsbajorics (approved):

Rheeseyb (approved):

PR #6512, opened 2024-10-10
Insert better default grids

### LINKED ISSUE #6511: insert menu grid + 2x2 when converting

ruggi:
**Problem:**

1. When creating a grid layout from the inspector the default grid should be 2x2 with a 10px gap
2. There should be an insert menu entry in the canvas toolbar for a 3x3 grid with a 10px gap
3. There should be an entry for a grid in the floating component picker too

**Fix:**

Implement all of the above:

https://github.com/user-attachments/assets/6214c47d-ae77-4735-b9ec-ab00198e4b90

Note: the icon for the grid is the one used for the navigator, there should be a dedicated one with the right size (16x16) for the canvas toolbar as well.

Fixes #6511

### DIFF
--- a/editor/src/components/shared/project-components.ts
+++ b/editor/src/components/shared/project-components.ts
@@ -126,6 +126,14 @@ export function insertableComponentGroupDiv(): InsertableComponentGroupDiv {
   return { type: 'HTML_DIV' }
 }
 
+export interface InsertableComponentGroupGrid {
+  type: 'HTML_GRID'
+}
+
+export function insertableComponentGroupGrid(): InsertableComponentGroupGrid {
+  return { type: 'HTML_GRID' }
+}
+
 export function insertableComponentGroupHTML(): InsertableComponentGroupHTML {
   return {
     type: 'HTML_GROUP',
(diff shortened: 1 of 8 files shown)


editor/src/components/shared/project-components.ts
```diff
@@ -462,6 +474,41 @@ const divComponentGroup = {
...
+                  display: 'grid',
+                  gridTemplateRows: '1fr 1fr 1fr',
+                  gridTemplateColumns: '1fr 1fr 1fr',
+                  gap: 10,
+                  position: 'absolute',
+                  width: 200,
+                  height: 200,
+                },
+                emptyComments,
+              ),
```
balazsbajorics, line 501: this should be a shared function for the floating menu and the insert button, so that they remain consistent even if we tweak the default template

ruggi: done in [`8279ecf` (#6512)](https://github.com/concrete-utopia/utopia/pull/6512/commits/8279ecf03f587201a89c4ffb32bb793eb9f3b618)

liady (approved):

PR #6513, opened 2024-10-10
Style plugins

bkrmendy:
# [Example project with a flex gap](https://utopia.fish/p/71b4ada2-boatneck-lamp/?branch_name=feature-style-plugins)

## Problem
The canvas controls can't edit elements that are styled with Tailwind

## Fix
Add a plugin system, where each plugin provides a data source that strategies can read from (so as of this PR, either Tailwind or the inline style), and a function that takes props from the inline style, and moves them to the right place the style element. For example, the plugin that implements Tailwind styling takes the inline style props, converts them to the right tailwind classes, and adds them to the `className` prop.

The flex gap strategy and the flex gap control is updated to use the data provided by the plugin system to read the necessary data.

### Out of scope
There's one known limitation of this system as implemented in this PR: props that are removed by a strategy don't get removed by the normalization step in plugins. This is only a problem for the Tailwind plugin (since the inline style plugin leaves the inline style prop as the strategies left them, so the fix for this will come on a follow-up PR after this one (the actual fix is already implemented here for anyone interested https://github.com/concrete-utopia/utopia/pull/6510/commits/b2fc141d411d684c618792c6b4963e11c7a00f54)

### Details
- Types for the plugin system, and two concrete plugin implementations are added in the `canvas/plugins` folder
- The data provided by the plugins is added to `InteractionCanvasState` as the `styleInfoReader` function
- The normalization step provided by the plugins is hooked into the strategy lifecycle in `interactionFinished` 
- The flex gap strategy/controls are updated to use `styleInfoReader`
- `UpdateClassList` is fixed so that it can write the `className` prop even if it doesn't exist on element at the start of the command

### DIFF
--- a/editor/src/components/canvas/controls/select-mode/flex-gap-control.tsx
+++ b/editor/src/components/canvas/controls/select-mode/flex-gap-control.tsx
@@ -24,8 +24,8 @@ import { createInteractionViaMouse, flexGapHandle } from '../../canvas-strategie
 import { windowToCanvasCoordinates } from '../../dom-lookup'
 import {
   cursorFromFlexDirection,
-  maybeFlexGapData,
   gapControlBoundsFromMetadata,
+  getFlexGapData,
   recurseIntoChildrenOfMapOrFragment,
 } from '../../gap-utils'
 import { CanvasOffsetWrapper } from '../canvas-offset-wrapper'
@@ -46,6 +46,7 @@ import {
   reverseJustifyContent,
 } from '../../../../core/model/flex-utils'
 import { optionalMap } from '../../../../core/shared/optional-utils'
+import { getActivePlugin } from '../../plugins/style-plugins'
 
 interface FlexGapControlProps {
   selectedElement: ElementPath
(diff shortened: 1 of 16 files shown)


editor/src/components/canvas/controls/select-mode/flex-gap-control.tsx
```diff
@@ -130,21 +132,40 @@ export const FlexGapControl = controlForStrategyMemoized<FlexGapControlProps>((p
     elementPathTrees,
     selectedElement,
   )
-  const flexGap = maybeFlexGapData(metadata, selectedElement)
-  if (flexGap == null) {
-    return null
-  }
 
-  const flexGapValue = updatedGapValue ?? flexGap.value
+  const flexGapFromEditor = useEditorState(
```
bkrmendy, line 136: the refactor here was necessary because the early return didn't respect the rules of hooks

balazsbajorics (approved):

ruggi (commented):
note for future selves: `gap` is not only for flex, but also for grids (https://developer.mozilla.org/en-US/docs/Web/CSS/gap), so maybe we should take that into account when naming things (e.g. `flexGap`)

editor/src/components/canvas/controls/select-mode/flex-gap-control.tsx
```diff
@@ -130,21 +132,40 @@ export const FlexGapControl = controlForStrategyMemoized<FlexGapControlProps>((p
...
+    (store) =>
+      maybeFlexGapData(
+        getActivePlugin(store.editor).styleInfoFactory({
+          projectContents: store.editor.projectContents,
+          metadata: metadata,
+          elementPathTree: store.editor.elementPathTree,
+        })(selectedElement),
+        MetadataUtils.findElementByElementPath(store.editor.jsxMetadata, selectedElement),
+      ),
+    'FlexGapControl flexGap',
```
ruggi, line 147:
```suggestion
    'FlexGapControl flexGapFromEditor',
```

editor/src/core/tailwind/tailwind-class-list-utils.ts
```diff
@@ -40,6 +40,7 @@ export function getClassListFromParsedClassList(
         config ?? undefined,
       )
     })
+    .filter((part) => part != null && part?.length > 0)
```
ruggi, line 43:
```suggestion
    .filter((part) => part != null && part.length > 0)
```

ruggi (approved):

gbalint (approved):

PR #6514, opened 2024-10-10
Remove unnecessary custom require function boilerplate

bkrmendy:
## Problem
In multiple places there's unnecessary boilerplate code around creating the custom `require` function.

## Fix
Remove the boilerplate code, since it's only necessary where we want to inject `ComponentRendererComponent`s or or collect data in `MutableUtopiaCtxRefData`

### Details
- updated snapshot tests in the property control descriptions tests, which kept the file extension of the module key in a way that wasn't consistent with the rest of the tests
- added a tests that makes sure that components in `RemixScene`s can be focused (an earlier iteration of this PR broke component focusing in `RemixScene`s)

**Manual Tests:**
I hereby swear that:


Rheeseyb (approved):

liady (approved):

Rheeseyb (approved):
Thanks for adding the test!

PR #6516, opened 2024-10-10
Paste grid cell as sibling

### LINKED ISSUE #6515: better grid cell paste

ruggi:
**Problem:**

Copy-pasting a grid cell ends up pasting it inside the selected element instead of next to it.

**Fix:**

If the copied element is a grid cell and the target is another grid cell, paste next to it.

| Before | After |
|------------|-------------|
| Kapture 2024-10-10 at 16 18 06 | Kapture 2024-10-10 at 16 16 49 |

Fixes #6515

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/reparent-utils.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/reparent-utils.ts
@@ -538,6 +538,36 @@ function pasteNextToSameSizedElement(
   return null
 }
 
+function pasteIntoNextGridCell(
+  copyData: ParsedCopyData,
+  selectedViews: NonEmptyArray<ElementPath>,
+  metadata: ElementInstanceMetadataMap,
+): ReparentTargetForPaste | null {
+  if (selectedViews.length === 0) {
+    return null
+  }
+  const selectedView = lastOfNonEmptyArray(selectedViews)
+
+  // if the copied elements are grid cells and the target is a grid cell, paste next to it
+
+  const copyDataElementsAreGridCells = copyData.elementPaste.every(({ originalElementPath }) =>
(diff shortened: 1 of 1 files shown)


balazsbajorics (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-utils.ts
```diff
@@ -538,6 +538,36 @@ function pasteNextToSameSizedElement(
...
+  // if the copied elements are grid cells and the target is a grid cell, paste next to it
+
+  const copyDataElementsAreGridCells = copyData.elementPaste.every(({ originalElementPath }) =>
+    MetadataUtils.isGridCell(copyData.originalContextMetadata, originalElementPath),
+  )
+  if (!copyDataElementsAreGridCells) {
+    return null
+  }
+
+  const targetIsGridCell = MetadataUtils.isGridCell(metadata, selectedView)
```
liady, line 560: just a small unrelated naming nitpick - the element itself should probably be called "grid child" (since a "grid cell" is usually the actual cell and not the element being copied in this case)

liady (approved):

editor/src/components/canvas/canvas-strategies/strategies/reparent-utils.ts
ruggi: true, but we should also update the method name (`isGridChild`). We have a bunch of naming cleanup to do, (e.g. `tracks` vs `dimensions`)

ruggi: [`c170a70` (#6516)](https://github.com/concrete-utopia/utopia/pull/6516/commits/c170a70c7ab22c9b3a1434a01fbf56cfd7cfe786)

PR #6517, opened 2024-10-10
GridMeasurementHelper component to continuously measure grid cells

gbalint:
**Problem:**
When you reparent from a grid to another grid, for a single frame the element blinks at the same cell position as it was in the original grid: https://screenshot.click/09-31-10wrd-ysw94.mp4 

**Root cause:**
As you hover over the target grid, the grid-reparent-strategy is selected. It triggers a showGridControl command, so the grid control is shown and the cells can be measured. However, it won't be able to position to the real target cell in this frame yet, because the control is not enabled yet, so we don't have the cell bounds in the metadata.

**Fix:**
Instead of using grid controls to measure let show permanent grid measurement helpers so grids can always be measured.

- Added memoized function `getAllGrids` which returns all the grids from the metadata
- Added `GridMeasurementHelper` component which renders a grid with placeholder divs in its cells (similarly to `GridControl`, but much-much simpler)
- Rendering `GridMeasurementHelper` for all grids
- Use rendered grid measurement helpers in dom-walker to measure the size of grid cell bounds
- Remove all the hacks (most notably `getMetadataWithGridCellBounds`) which made it possible to use latestMetadata in interaction when startingMetadata did not include grid cell bounds. This is not needed anymore, because the grid cells are always measured using the always-on `GridMeasurementHelper` components

**Manual Tests:**
I hereby swear that:


ruggi (approved):

editor/src/components/canvas/controls/new-canvas-controls.tsx
```diff
@@ -596,6 +606,20 @@ const NewCanvasControlsInner = (props: NewCanvasControlsInnerProps) => {
                   isSelectOrInsertMode(editorMode) &&
                     !EP.multiplePathsAllWithTheSameUID(localSelectedViews),
                   <>
+                    {when(
+                      dragInteractionActive,
+                      <CanvasOffsetWrapper>
+                        {gridsWithoutControl.map((grid) => {
```
balazsbajorics, line 612: can't these controls be added by the reparent metastrategy? we only want this to happen when it's a drag-to-reparent!

ruggi (approved):

bkrmendy (approved):

PR #6521, opened 2024-10-11
Fix Grid Controls HMR

balazsbajorics:
**Problem:**
The strategy controls were wrapped in `controlForStrategyMemoized` which [defeats Vite's HMR.](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react#consistent-components-exports)

**Fix:**
Move all the component internals into a dedicated file which doesn't export other functions to prevent breaking the rules of hmr.

**Commit Details:**

- Moved all grid control components and their helper code into `grid-controls-components`
- `grid-controls` now imports these components
- No code was changed, only moved

I verified that with this change, HMR started working!

## Bonus

At @liady 's suggestion, I added `eslint-plugin-react-refresh` to the list of our ESLint rules. It is not rock solid, but definitely helps in remembering that there _are_ rules for HMR / Fast Refresh to work correctly.


eslint-config-utopia/index.js
```diff
@@ -168,6 +168,13 @@ module.exports = {
     'jest/no-conditional-expect': 'off',
     'jest/no-done-callback': 'off',
     'jest/no-test-prefixes': 'off',
+    'react-refresh/only-export-components': [
+      'warn',
+      {
+        allowConstantExport: true,
+        allowExportNames: ['meta', 'links', 'headers', 'loader', 'action'], // For Remix compatibility
+      },
+    ],
```
rbshop, line 177: Nice

rbshop (approved):

seanparsons (approved):

PR #6526, opened 2024-10-11
Handle Nested Grid Dragging

seanparsons:
**Problem:**
Rearranging a grid inside a grid only works if dragging a cell of the nested grid.

**Fix:**
The main fix was the removal of a `pointerEvents` property which was inadvertently preventing the mousedown from triggering as it should.

But a secondary issue was discovered while testing that, which is that the mousedown is also not handled correctly in this case:
image
When the mousedown is triggered it selects the parent grid item, but this is too late as there's no handling present for the parent grid item as that gets initialised by the selection change. To fix this, ancestor grids have been added to the grids currently present so that the ancestor grids can catch the mouse down.

However, doing the above caused an issue of duplicated controls which broke some of our tests. To fix that `combineApplicableControls` was implemented to combine different instances of `GridControls`, thereby deduplicating some of the controls that would render.

**Commit Details:**
- Implemented `combineApplicableControls` to collate disparate instances of `GridControls`.
- `controlsForGridPlaceholders` has been slightly tweaked to cater for the change in the type of `GridControlsProps` and also to handle a new option suffix.
- Removed `pointerEvents` setting which was bizarrely the cause of the original issue.
- Removed spurious property.
- `GridControlsComponent` now includes ancestor paths.

**Manual Tests:**
I hereby swear that:

- [x] ] I opened a hydrogen project and it loaded

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
+++ b/editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
@@ -19,6 +19,7 @@ import type {
   StrategyApplicationResult,
   InteractionLifecycle,
   CustomStrategyState,
+  WhenToShowControl,
 } from './canvas-strategy-types'
 import {
   ControlDelay,
@@ -83,6 +84,11 @@ import { getReparentTargetUnified } from './strategies/reparent-helpers/reparent
 import { gridRearrangeResizeKeyboardStrategy } from './strategies/grid-rearrange-keyboard-strategy'
 import createCachedSelector from 're-reselect'
 import { getActivePlugin } from '../plugins/style-plugins'
+import {
+  controlsForGridPlaceholders,
+  GridControls,
+  isGridControlsProps,
+} from '../controls/grid-controls-for-strategies'
 
 export type CanvasStrategyFactory = (
   canvasState: InteractionCanvasState,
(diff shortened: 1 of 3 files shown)


balazsbajorics (changes requested):
if I try to rearrange the child of a grid inside a grid, something goes wrong:
https://github.com/user-attachments/assets/0c749364-c674-4bdc-90f2-4500e2a623e7

in general if I have the child of a grid-in-a-grid selected, we shouldn't show visible grid outlines for the outer grid, as that grandparent element is irrelevant for the selected element

balazsbajorics (changes requested):
_visually_, everything should look like as it does on master:
https://github.com/user-attachments/assets/f5dddbc1-fefb-4957-ad22-0ba7c802a634

I tried the Try Me link, and when I have the inner rectangle selected, we shouldn't show the grid cells for the grandparent grid (as it is not relevant), the pink target cell outlines are hidden, and we also show too much of the blue helper background colors obscuring the content:
https://github.com/user-attachments/assets/4cfc3f4e-47ce-400e-840c-302993040c26

seanparsons:
> _visually_, everything should look like as it does on master: https://github.com/user-attachments/assets/f5dddbc1-fefb-4957-ad22-0ba7c802a634
> 
> I tried the Try Me link, and when I have the inner rectangle selected, we shouldn't show the grid cells for the grandparent grid (as it is not relevant), the pink target cell outlines are hidden, and we also show too much of the blue helper background colors obscuring the content: https://github.com/user-attachments/assets/4cfc3f4e-47ce-400e-840c-302993040c26

Fixed in 2583a84be.

PR #6530, opened 2024-10-14
grid rearrange's pink outline no longer partially overlapped

balazsbajorics:
**Problem:**
image

The active grid cell's pink outline would sometimes appear under the grey outline drawn for non-hovered cells.

**Fix:**
image

- The cell that has the pink outline is now drawn with a z-index: 1
- The outline no longer uses opacity which causes errors in color when they overlap
- This means the cell outlines no longer need to predict overlap and avoid drawing right and bottom borders

### DIFF
--- a/editor/src/components/canvas/controls/grid-controls.tsx
+++ b/editor/src/components/canvas/controls/grid-controls.tsx
@@ -840,7 +840,7 @@ const GridControl = React.memo<GridControlProps>(({ grid }) => {
               style={{
                 position: 'relative',
                 pointerEvents: 'initial',
-                zIndex: activePositioningTarget ? 1 : 0,
+                zIndex: activePositioningTarget ? 1 : undefined,
               }}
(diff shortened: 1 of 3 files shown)


gbalint (approved):

ruggi (approved):

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -839,6 +840,7 @@ const GridControl = React.memo<GridControlProps>(({ grid }) => {
               style={{
                 position: 'relative',
                 pointerEvents: 'initial',
+                zIndex: activePositioningTarget ? 1 : 0,
```
ruggi, line 843: would it be better to leave it as `undefined` or `inherit` in the right hand case?

balazsbajorics: hm yeah that might be better

PR #6537, opened 2024-10-15
Tailwind Padding

bkrmendy:
## Problem
Paddings defined from Tailwind cannot be edited on the canvas.

Other problems that came up in this PR:
1. The plugin prop removal code removed props that were deleted at the start of a strategy, and then later set again
2. In some cases, the padding strategy removes both the shorthand `padding` prop and all four longhand padding props (`padding{ Top | Left | Right | Bottom }`), and sets them one or more of these later. When this happens, the property patching code kicks in, and sets all 5 props to their default value (`0px`). However I found an unknown unknown here: it seems like if the shorthand padding prop is set last, it takes precedence over the longhand padding props, so setting `padding: 0` last negates a `paddingTop: 10` for example. This is not a Utopia-specific problem, it can be reproed in [codesandbox](https://codesandbox.io/p/sandbox/naughty-leavitt-djhdvv) by commenting out and re-adding one of the padding props ([utopia project](https://utopia.pizza/p/acd1ed02-rigorous-magnesium/) for reference)

### Tailwind paddings:
- Extend `StyleInfo` and add the necessary reader code
- update `simplePaddingFromMetadata` to read from `StyleInfo` and make the necessary refactors

### Plugin prop removal code:
The immediate problem is fixed by tracking the prop setting/deleting commands, and simulate them to figure out which props are reset after they are deleted.

This solutions works for the specific problem in this PR, but it poses a bigger question: what if, in the command runner function, a command tries to remove a prop from some other element that's not in the command constructor function? The commands are only aware of the style prop, so in a Tailwind project, the remove just won't work (and we couldn't find out even if we diffed the project contents before and after a command has run)

### Padding prop weirdness:
I fixed this by only using the longhand versions of a prop to patch the shorthand (see https://github.com/concrete-utopia/utopia/blob/c4933758b8b333861dbb08f9164d8996df842dd2/editor/src/components/editor/actions/action-utils.ts#L532-L622)

**Manual Tests:**
I hereby swear that:

- [ ] I could navigate to various routes in Play mode


bkrmendy:
https://github.com/concrete-utopia/utopia/pull/6637

PR #6538, opened 2024-10-15
Grid Gap Control 2

balazsbajorics:
**Problem:**
The orange grid gap controls are sometimes 1px off, they are one frame behind, they show weird delays when the canvas zooms.

**Fix:**
A lot of the problem came down to `gridGapControlBoundsFromMetadata` trying to calculate the bounding boxes of where the gaps are in the selected Grid. However this is error prone because a lot of these boxes will fall on fractional pixel values, and we were not rounding them the same way Chrome was rounding the real grid. 

Instead of trying to fix the math, I had an idea to replicate our hack for the grid cell outline controls: take the ComputedStyle.gridTemplateRows / Columns, and create a helper grid which has the correct position and sizing for our gap outlines. The trick is that we want to take the original grid's gap value, and turn those into grid tracks too, setting the helper grid's gap to zero.

So if my original grid's ComputedStyle has tracks [5px 10px 15px 5px] and a gap of 13px, we would create a helper grid with the tracks [5px 13px 10px 13px 15px 13px 5px]!

Now the only job is to fill it with the elements and make sure to only draw anything in the tracks that correspond to gaps in the original grid (they are always the odd numbered tracks).

This means we can offload the entire positioning and rendering to Chrome, and makes our life much simpler going forward.

**Commit Details:** 

- Fixed Hot Reload by creating a new file `grid-gap-control-component.tsx`
- Brand new `GridGapControlComponent`, `GridPaddingOutlineForDimension` and `GridRowHighlight` components. 
- `GridPaddingOutlineForDimension` creates the helper grid with the inlined gaps as tracks
- `GridRowHighlight` is responsible for the visual lines, and it recreates the original grid in the given dimension so it can have a handle for each original row
- `GridGapHandle` mostly unchanged


editor/src/components/canvas/controls/select-mode/grid-gap-control-component.tsx
```diff
@@ -0,0 +1,477 @@
...
+    onMouseDown,
+    beingDragged,
+    onMouseOver,
+    zIndexPriority,
+    elementHovered,
+  } = props
+
+  let style: CSSProperties = {
+    ...getStyleMatchingTargetGrid(grid),
+    zIndex: zIndexPriority ? 1 : 0,
```
ruggi, line 165: I believe here as well might be worth leaving it
```suggestion
    zIndex: zIndexPriority ? 1 : undefined,
```

editor/src/components/canvas/controls/select-mode/grid-gap-control-component.tsx
```diff
@@ -0,0 +1,477 @@
...
+        height={height}
+        pillColor={colorTheme.brandNeonOrange.value}
+        borderWidth={borderWidth}
+      />
+    </div>
+  )
+}
+
+function handleDimensions(axis: Axis, scale: number): Size {
+  if (axis === 'row') {
```
ruggi, line 470: why not a `switch` here?

editor/src/components/canvas/controls/select-mode/grid-gap-control-helpers.tsx
```diff
@@ -0,0 +1,35 @@
+import type { CanvasVector } from '../../../../core/shared/math-utils'
+import { windowPoint } from '../../../../core/shared/math-utils'
+import { Modifier } from '../../../../utils/modifiers'
+import type { EditorDispatch } from '../../../editor/action-types'
+import CanvasActions from '../../canvas-actions'
+import { createInteractionViaMouse, gridGapHandle } from '../../canvas-strategies/interaction-state'
+import { windowToCanvasCoordinates } from '../../dom-lookup'
+import type { Axis } from '../../gap-utils'
+
+export function startInteraction(
```
ruggi, line 10: I think it'd be better to give this a more specific name, e.g. `startGapControlInteraction` (or whatever makes most sense to you) since it's exported

ruggi (approved):

editor/src/components/canvas/controls/select-mode/grid-gap-control-helpers.tsx
balazsbajorics: oh yes this is a very good shout

gbalint (approved):

editor/src/components/canvas/controls/select-mode/grid-gap-control-component.tsx
balazsbajorics: this is just moved code, I wanted to leave most it intact for forensic reasons

PR #6546, opened 2024-10-15
Partial grid resize mixed with absolute resize

### LINKED ISSUE #6544: Mixed/partial grid resize and absolute resize for when stretching on a single direction

ruggi:
**Problem:**

When a grid child is stretching on a single direction it should be possible to cell-resize it along that direction, while element-resize it in the other direction. The diagonal resize controls should be disabled in that case.

**Fix:**

Do that.

https://github.com/user-attachments/assets/e53be7c8-08fc-4ffe-8248-d147e01981f9

Fixes #6544


balazsbajorics (approved):
nice!

liady (approved):

PR #6549, opened 2024-10-16
Move normalization step into the dispatch flow

bkrmendy:
https://github.com/user-attachments/assets/d184b155-e18a-42d6-ba84-1532f6ebd280

## Problem
The tailwind style plugin can't differentiate between a prop being removed from the inline style prop, and that prop never having been there at all. For example, when a flex gap is removed by dragging it in the negative direction, it goes from positive -> 0 -> being removed from the style prop. When the normalisation step kicks in, there's no gap prop in the inline style, so the tailwind plugin doesn't do anything. To make matters worse, when the gap property is removed while the interaction is in progress, the value coming from the generated tailwind class takes precedence, so the flex gap jumps back to its original size.

## Fix
The problem of removing props in the normalization step is fixed by
1. moving the normalization step in the display flow
2. making the normalization step aware that some props need to be removed
3. find out which elements to normalize/which props to remove from the dispatched actions and the strategies, and running the normalization on those elements and prop.

The problem of prevent the CSS classes to kick in when a prop is removed (for example, when the inline `gap` prop is removed in the flex gap strategy), this PR adds a postprocessing step alongside the normalization step, that patches an allowlisted set of props to a "zero" value, which simulates the prop not being there at all

### Details
- added a test for removing the gap prop set from Tailwind
- added a `propertiesToRemove` param to `normalizeFromInlineStyle`, and updated the tailwind style plugin to use that prop to remove classes from the `className` prop
- added the code that finds out which elements/props to normalize from the dispatched actions
- added the code that finds out which elements/props to normalize from the strategies, and that tells which elements to patch during an interaction

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/plugins/style-plugins.ts
+++ b/editor/src/components/canvas/plugins/style-plugins.ts
@@ -8,6 +8,7 @@ import {
   isTailwindEnabled,
 } from '../../../core/tailwind/tailwind-compilation'
 import type { PropertiesToUnsetForElement } from '../../editor/actions/action-utils'
+import { isFeatureEnabled } from '../../../utils/feature-switches'
 
 export interface StylePlugin {
   name: string
@@ -25,7 +26,7 @@ export const Plugins = {
 } as const
 
 export function getActivePlugin(editorState: EditorState): StylePlugin {
-  if (isTailwindEnabled()) {
+  if (isFeatureEnabled('Tailwind') && isTailwindEnabled()) {
     return TailwindPlugin(getTailwindConfigCached(editorState))
   }
   return InlineStylePlugin
--- a/editor/src/components/canvas/plugins/style-plugins.ts
+++ b/editor/src/components/canvas/plugins/style-plugins.ts
@@ -7,6 +7,7 @@ import {
   getTailwindConfigCached,
   isTailwindEnabled,
 } from '../../../core/tailwind/tailwind-compilation'
+import type { PropertiesToUnsetForElement } from '../../editor/actions/action-utils'
 
 export interface StylePlugin {
   name: string
@@ -14,6 +15,7 @@ export interface StylePlugin {
   normalizeFromInlineStyle: (
     editorState: EditorState,
     elementsToNormalize: ElementPath[],
+    propertiesToRemove: PropertiesToUnsetForElement[],
   ) => EditorState
 }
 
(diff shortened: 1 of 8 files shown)


balazsbajorics (approved):

gbalint (approved):

editor/src/components/canvas/plugins/tailwind-style-plugin.ts
```diff
@@ -59,6 +60,55 @@ function getTailwindClassMapping(classes: string[], config: Config | null) {
...
+    return mapDropNulls(
+      ([key, value]) => (typeof key !== 'object' ? null : { key, value }),
+      Object.entries(styleProp.value),
+    )
+  }
+
+  return null
+}
+
+function getRemoveUpdates(properties: PropertyPath[]) {
```
Rheeseyb, line 88: Some of these functions are missing return types. Can you add them please

bkrmendy: added them in https://github.com/concrete-utopia/utopia/pull/6549/commits/c84c9525c010928f343789d907b974835e0fdb9c

Rheeseyb (approved):

editor/src/components/canvas/plugins/style-plugins.ts
```diff
@@ -23,7 +26,7 @@ export const Plugins = {
 } as const
 
 export function getActivePlugin(editorState: EditorState): StylePlugin {
-  if (isTailwindEnabled()) {
+  if (isFeatureEnabled('Tailwind') && isTailwindEnabled()) {
```
seanparsons, line 29: Should that feature switch be inside `isTailwindEnabled`?

bkrmendy: good point, I put the feature flag check here to be explicit about the fact that this won't kick in until the FS is removed

editor/src/components/editor/actions/action-utils.ts
```diff
@@ -362,3 +364,74 @@ export function simpleStringifyActions(
     .map((a) => simpleStringifyAction(a, indentation))
     .join(`,\n${spacing}`)}\n${spacingBeforeClose}]`
 }
+
+export function getElementsToNormalizeFromCommands(commands: CanvasCommand[]): ElementPath[] {
+  return mapDropNulls((command) => {
+    switch (command.type) {
```
seanparsons, line 370: Minor, but this being exhaustive ensures we have to fix it when adding a new `CanvasCommand`.

seanparsons (approved):

PR #6567, opened 2024-10-18
Handle Changing Imports In Canvas

seanparsons:
**Problem:**
If we do a partial canvas re-render but a necessary import is also added, then the canvas errors out momentarily.

**Cause:**
The resolved values of the imports are buried in a property called `requireResult` which originates with require and resolve functions. Those functions are only rebuilt when the elements to re-render value is `rerender-all-elements` or when it has changed between the current canvas render and the prior one. With our recent changes to the canvas strategies to make them never supply the `rerender-all-elements` value, the failure case of the imports changing but the elements to re-render value not changing has become more likely.

In the case of grid insertion, the elements to re-render becomes the target path on hover, then when the user clicks it stays with that target path but the imports are added. Since there is no change in the former, the canvas logic is unable to find the new import in the old `requireResult`.

**Fix:**
Compare the old and current project contents and if their imports, combined top level arbitrary block or exports have changed then recreate the require and resolve functions. This replaces the previous check which looked at the current and previous versions of the elements to re-render property.

**Commit Details:**
- Removed the previous handling of `shouldRerenderRef`.
- Implemented `projectContentsSameForRefreshRequire`.
- Changed out old `haveProjectImportsChanged` call to use the new
  `projectContentsSameForRefreshRequire`.

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/grid-draw-to-insert-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/grid-draw-to-insert-strategy.tsx
@@ -171,7 +171,7 @@ const gridDrawToInsertStrategyInner =
               ),
               updateHighlightedViews('mid-interaction', [targetParent]),
             ],
-            [targetParent],
+            [],
           )
         }
 
--- a/editor/src/components/canvas/canvas-utils.ts
+++ b/editor/src/components/canvas/canvas-utils.ts
@@ -63,7 +63,12 @@ import type {
   HighlightBoundsForUids,
   ExportsDetail,
 } from '../../core/shared/project-file-types'
-import { isExportDefault, isParseSuccess, isTextFile } from '../../core/shared/project-file-types'
+import {
+  importsEquals,
+  isExportDefault,
+  isParseSuccess,
+  isTextFile,
+} from '../../core/shared/project-file-types'
 import {
   applyUtopiaJSXComponentsChanges,
   getDefaultExportedTopLevelElement,
@@ -140,7 +145,11 @@ import { getStoryboardUID } from '../../core/model/scene-utils'
 import { optionalMap } from '../../core/shared/optional-utils'
 import { assertNever, fastForEach } from '../../core/shared/utils'
 import type { ProjectContentTreeRoot } from '../assets'
-import { getProjectFileByFilePath } from '../assets'
+import {
+  getProjectFileByFilePath,
+  isProjectContentDirectory,
+  isProjectContentFile,
+} from '../assets'
 import type { CSSNumber } from '../inspector/common/css-utils'
 import { parseCSSLengthPercent, printCSSNumber } from '../inspector/common/css-utils'
 import { getTopLevelName, importedFromWhere } from '../editor/import-utils'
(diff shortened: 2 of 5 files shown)


editor/src/components/canvas/canvas-strategies/strategies/grid-draw-to-insert-strategy.tsx
```diff
@@ -171,7 +171,7 @@ const gridDrawToInsertStrategyInner =
               ),
               updateHighlightedViews('mid-interaction', [targetParent]),
             ],
-            [targetParent],
+            [],
```
Rheeseyb, line 174: Are you sure we no longer need to pass in the `targetParent` here? Or was this maybe an old investigation that was accidentally included?

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -2180,3 +2189,69 @@ export function canvasPanelOffsets(): {
...
+  newProjectContents: ProjectContentTreeRoot,
+): boolean {
+  if (oldProjectContents === newProjectContents) {
+    // Identical references, so the imports are the same.
+    return true
+  } else {
+    for (const [filename, oldProjectTree] of Object.entries(oldProjectContents)) {
+      const newProjectTree = newProjectContents[filename]
+      // If the file can't be found in the other tree, the imports are not the same.
+      if (newProjectTree == null) {
```
Rheeseyb, line 2204: I think because we use `ProjectContentTreeRootKeepDeepEquality` we could also have a check here for `if (oldProjectTree === newProjectTree)`

editor/src/components/canvas/canvas-utils.ts
```diff
@@ -2180,3 +2189,69 @@ export function canvasPanelOffsets(): {
...
+        isProjectContentDirectory(oldProjectTree) &&
+        isProjectContentDirectory(newProjectTree)
+      ) {
+        // Both entries are directories.
+        if (
+          !projectContentsSameForRefreshRequire(oldProjectTree.children, newProjectTree.children)
+        ) {
+          // The imports of the subdirectories differ.
+          return false
+        }
```
Rheeseyb, line 2247: Could just `return projectContentsSameForRefreshRequire(oldProjectTree.children, newProjectTree.children)` here

editor/src/components/editor/import-utils.ts
```diff
@@ -366,3 +366,13 @@ function removeImportDetails(
     importedFromWithin: importedFromWithin,
   }
 }
+
+export function getProjectImports(projectContents: ProjectContentTreeRoot): {
```
Rheeseyb, line 370: Looks like you never ended up using this function

Rheeseyb (approved):

editor/src/components/canvas/canvas-strategies/strategies/grid-draw-to-insert-strategy.tsx
seanparsons: This is for the stage where insertion has been activated but not started, so there's no need to re-render the parent.

editor/src/components/canvas/canvas-utils.ts
seanparsons: Can't do that because we're in a for loop that would exit on the first case where those values were the same. But we need to check those entries exhaustively.

editor/src/components/canvas/canvas-utils.ts
seanparsons: Improvement added in 112d15b20.

editor/src/components/editor/import-utils.ts
seanparsons: Removed in 112d15b20.

editor/src/components/canvas/canvas-utils.ts
Rheeseyb: My bad 🤦

ruggi (approved):

PR #6573, opened 2024-10-21
Open grid template row/col inspector menu with right click on title

### LINKED ISSUE #6572: Open grid template context menu right clicking the title

ruggi:
**Problem:**

It should be possible to open the dropdown/context menu for grid template rows/cols in the inspector by right-clicking the item's title.

**Fix:**

Add an invisible context menu in the title component and trigger it conditionally by right clicking the title itself.
This solution works better than conditionally opening the same dropdown menu as the three dots button because it will correctly position it.

https://github.com/user-attachments/assets/2c25fc94-b039-4423-9a3f-0c6e0f71b562

Fixes #6572

### DIFF
--- a/editor/src/uuiui/radix-components.tsx
+++ b/editor/src/uuiui/radix-components.tsx
@@ -91,7 +91,7 @@ export interface DropdownMenuProps {
   alignOffset?: number
   onOpenChange?: (open: boolean) => void
   style?: CSSProperties
-  isOpen?: boolean
+  forceOpen?: boolean
 }
 
 export const ItemContainerTestId = (id: string) => `item-container-${id}`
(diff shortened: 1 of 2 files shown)


editor/src/uuiui/radix-components.tsx
```diff
@@ -103,7 +104,7 @@ export const DropdownMenu = React.memo<DropdownMenuProps>((props) => {
   }, [])
   const onEscapeKeyDown = React.useCallback((e: KeyboardEvent) => e.stopPropagation(), [])
 
-  const [open, setOpen] = React.useState(false)
+  const [open, setOpen] = useIsDropdownMenuOpen(props.isOpen ?? null)
```
ruggi, line 107: Moved into a dedicated hook so we can keep track of the `useEffect` (which I don't think can be avoided).

bkrmendy (approved):

balazsbajorics (approved):

PR #6574, opened 2024-10-22
Tailwind MegaSpike

bkrmendy:
# Problems explored on this spike

## Patching removed props
During canvas interactions, we write props to the inline style even in Tailwind projects, so that the updates happen as fast as possible. However, when a prop is removed during a canvas interaction, we want to override it with a 0px (or similar) value so that the relevant Tailwind class doesn’t take over. Example: while the gap prop is being adjusted, the gap property value is coming from the inline style gap prop, but as soon as it’s removed, the gap value from the tailwind class takes over, making it look like the gap is jumping back to its original size.  Setting 0px works for simple values, but for more complex props like padding, this approach broke down in a surprising way: React doesn’t like it when the longhand and the shorthand version of the same CSS prop are set at the same time. This manifested in the paddings randomly disappearing. This is a well-documented wontfix in React, see https://github.com/facebook/react/issues/8689. We need to tackle this problem in way that won't infect all other code around it.

The solution to this is [patchRemovedProperties](https://github.com/concrete-utopia/utopia/blob/0266f942153d8a7e35fa7b4dee843a318326709f/editor/src/components/canvas/plugins/style-plugins.ts#L185-L205) (and everything else it calls). While internally it's still a pretty heuristic way to solve the problem, it isolates it in a way that
- only `runStyleUpdateMidInteraction` is involved in figuring out which properties need to be zeroed
- only the strategy dispatch in involved in zeroing these props, in an explicit, opt-in way (the opt-in happens through calling `patchRemovedProperties`)

## Code that has reading/writing from the style prop baked in
Utopia has reading/writing from the inline style prop baked in throughout the editor code, we should investigate how widespread this is and how this could be dealt with.

This lead to the following explorations:
- I refactored [`AdjustCssLengthProperties`](https://github.com/concrete-utopia/utopia/blob/0266f942153d8a7e35fa7b4dee843a318326709f/editor/src/components/canvas/commands/adjust-css-length-command.ts#L39) to use `styleInfoFactory` instead of reading from element props (see the diff for details), to showcase how we can refactor a command from reading directly from the inline style prop to a style-prop agnostic approach. This refactor also showed that the StyleInfo system doesn't express the state that a prop is set, but it's a value that we can't consume (for example if instead of a CSS number, it's a JS expression or identifier), so I made the necessary [adjustments](https://github.com/concrete-utopia/utopia/blob/0266f942153d8a7e35fa7b4dee843a318326709f/editor/src/components/canvas/canvas-types.ts#L540-L548). 
- The `SET_PROP`/`UNSET_PROP` actions write into the inline style prop too. I updated these actions to [use the commands under the hood](https://github.com/concrete-utopia/utopia/blob/0266f942153d8a7e35fa7b4dee843a318326709f/editor/src/components/editor/actions/actions.tsx#L1719-L1745), so if we wanted to add some extra logic related to setting/unsetting props, the actions get those automatically.
- There are some functions that write the inline style prop, and they are so high up the call stack that it would be very laborious to pass them down all the info the plugins might need (concretely, the Tailwind config for the Tailwind plugin). To deal with this, I added a Jotai [atom + store to store the Tailwind config](https://github.com/concrete-utopia/utopia/blob/0266f942153d8a7e35fa7b4dee843a318326709f/editor/src/core/tailwind/tailwind-compilation.ts#L47-L56), and make it accessible without massive amounts of refactoring

## Normalising props removed by an action or strategy / Alternative approaches to normalization
Commands (and the Utopia code in general) could modify elements that aren’t apparent from the command descriptor object. This is problematic because the normalisation step needs to know which elements to normalize, but this has to be communicated by the commands somehow.  Since the commands and prop setting in general have to be audited anyway (because of the above), an alternative to the normalization approach (which we have now) could be investigated.

One way to approach this is that in a command/action that might remove a prop, we make a note somewhere that a prop has been removed, and then later down the line remove that prop from wherever it needs to be removed (inline style prop or `className`). However, instead of "making a note", what about removing the prop straight away? This is underpinned by the fact that most commands already use `applyValuesAtPath` to set props, and `deleteValuesAtPath` to remove props. I updated the StylePlugin type with the equivalents of `applyValuesAtPath` and `deleteValuesAtPath`, turning them into an [official API](https://github.com/concrete-utopia/utopia/blob/0266f942153d8a7e35fa7b4dee843a318326709f/editor/src/components/canvas/plugins/style-plugins.ts#L40-L50) to be implemented by the style plugins. Then, [these functions can be called](https://github.com/concrete-utopia/utopia/blob/0266f942153d8a7e35fa7b4dee843a318326709f/editor/src/components/canvas/commands/set-property-command.ts#L109-L118) instead of writing the prop directly into the inline style.

The question of hard-to-refactor, high-up-the-call-stack functions comes up here too. We can take advantage of the fact that if a function reads/writes from/to element props (either via the props themselves or via `ElementInstanceMetadata`), because the `className` prop that Tailwind needs is also among the element props (this would be a harder nut to crack if we wanted to support `styled-components`). Style plugins already parse element styles from props, so by creating another entrypoint for [reading](https://github.com/concrete-utopia/utopia/blob/0266f942153d8a7e35fa7b4dee843a318326709f/editor/src/components/canvas/plugins/style-plugins.ts#L41-L44) and [writing](https://github.com/concrete-utopia/utopia/blob/0266f942153d8a7e35fa7b4dee843a318326709f/editor/src/components/canvas/plugins/style-plugins.ts#L50), and we use [Jotai store-based approach to get the right plugin to use](https://github.com/concrete-utopia/utopia/blob/0266f942153d8a7e35fa7b4dee843a318326709f/editor/src/components/canvas/plugins/style-plugins.ts#L60-L65), this becomes a tractable problem.


editor/src/components/canvas/commands/adjust-css-length-command.ts
```diff
@@ -70,237 +58,159 @@ export function adjustCssLengthProperties(
   }
 }
 
-interface UpdatedPropsAndCommandDescription {
-  updatedProps: JSXAttributes
-  commandDescription: string
-}
-
-export const runAdjustCssLengthProperties: CommandFunction<AdjustCssLengthProperties> = (
+export const runAdjustCssLengthProperties = (
```
bkrmendy, line 61: `runAdjustCssLengthProperties` is refactored not to read directly from the inline style prop

editor/src/components/canvas/commands/set-css-length-command.ts
```diff
@@ -75,45 +67,45 @@ export function setCssLengthProperty(
   }
 }
 
-export const runSetCssLengthProperty: CommandFunction<SetCssLengthProperty> = (
+export const runSetCssLengthProperty = (
```
bkrmendy, line 70: `runSetCssLengthProperty` is refactored not to read directly from the inline style prop

editor/src/components/canvas/plugins/style-plugins.ts
```diff
@@ -1,33 +1,207 @@
...
+) {
+  switch (commandLifecycle) {
+    case 'mid-interaction':
+      return runStyleUpdateMidInteraction(editorState, elementPath, updates)
+    case 'end-interaction':
+      return getActivePlugin(editorState).updateStyles(editorState, elementPath, updates)
+    default:
+      assertNever(commandLifecycle)
+  }
+}
```
bkrmendy, line 207: writing the inline styles instead of the whatever the active plugins needs during an interaction

PR #6583, opened 2024-10-23
Rename areaName to lineName

### LINKED ISSUE #6582: Rename areaName to lineName

ruggi:
**Problem:**

Grid dimension (which should be named tracks 🙃 , but that's for another day) have a field called `areaName` but it should actually be called `lineName` (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Grid_layout_using_named_grid_lines).

**Fix:**

Rename `areaName` to `lineName`, and all its related mentions and (indirect) references.

Fixes #6582


bkrmendy (approved):

balazsbajorics (approved):
thank you!!!!!!

PR #6586, opened 2024-10-24
Fixed handling of parent frames to get the right positioning.

seanparsons:
**Problem:**
When you draw-to-insert or reparent into an element which is in a grid cell (so not into the grid cell itself), then its absolute position is offseted with the position of the grid cell in the grid.

**Cause:**
The root cause here is that the reparent is happening into an element which then has `contain: 'layout'` added to it, which changes the containing block of the element inserted to that parent. But the calculations occur against some ancestor of the element being inserted into. This is why the offset becomes increasingly more dramatic the further down and to the right of the grid the new parent is, as the offset is being calculated against the top left of the grid.

**Fix:**
Now the reparent logic checks to see if `contain: 'layout'` _will_ be added and changes the bounds that the offset is calculated against to the new parent.

There is also an additional fallback to get the bounds when calculating the move commands, so that the right offset is calculated.

**Commit Details:**
- Split `maybeAddContainLayout` into that and another function `shouldAddContainLayout`.
- Corrected two tests which were inserting into the second flex child and were also offset similar to how the grid items were.
- `getAbsoluteReparentPropertyChanges` now takes a parameter which specifies if `contain: 'layout'` will be added to the container.
- `getMoveCommandsForSelectedElement` now has an additional fallback for `elementParentBounds` when the closest non fragment parent provides bounds for absolute children.
- `getLocalFrame` now has an optional parameter to specify the parent to target, to allow for handling the lookups where the element has yet to be reparented.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.tsx
@@ -261,7 +261,7 @@ export function createAbsoluteReparentAndOffsetCommands(
   projectContents: ProjectContentTreeRoot,
   nodeModules: NodeModules,
   forcePins: ForcePins,
-  containLayout: ShouldAddContainLayout,
+  willContainLayoutBeAdded: ShouldAddContainLayout,
 ) {
   const reparentResult = getReparentOutcome(
     metadata,
@@ -293,7 +293,7 @@ export function createAbsoluteReparentAndOffsetCommands(
         metadata,
         projectContents,
         forcePins,
-        containLayout,
+        willContainLayoutBeAdded,
       )
     })
 
--- a/editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.tsx
+++ b/editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.tsx
@@ -320,7 +320,7 @@ export function shouldAddContainLayout(
   )
 
   if (EP.isStoryboardPath(closestNonFragmentParent)) {
-    return 'add-contain-layout'
+    return 'dont-add-contain-layout'
   }
 
   const parentProvidesBoundsForAbsoluteChildren =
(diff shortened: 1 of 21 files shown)


editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.tsx
```diff
@@ -254,6 +260,7 @@ export function createAbsoluteReparentAndOffsetCommands(
   projectContents: ProjectContentTreeRoot,
   nodeModules: NodeModules,
   forcePins: ForcePins,
+  containLayout: 'should-add' | 'should-not-add',
```
ruggi, line 263: Given how it's used, may I suggest renaming these values to something that indicates more specifically what is or is not to be added? (and moving it to a dedicated type so it can be reused in e.g. `getAbsoluteReparentPropertyChanges`).
Thinking of something like
```ts
type ShouldAddContainLayout = 'add-contain-layout' | 'dont-add-contain-layout'
```
but whatever makes sense to you would work.

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-property-changes.ts
```diff
@@ -90,6 +90,7 @@ export function getAbsoluteReparentPropertyChanges(
   newParentStartingMetadata: ElementInstanceMetadataMap,
   projectContents: ProjectContentTreeRoot,
   forcePins: ForcePins,
+  containLayout: 'should-add' | 'should-not-add',
```
ruggi, line 93: see my earlier comment, I would also rename this var to `shouldAddContainLayout` or something like that

ruggi (approved):
works a treat

editor/src/components/canvas/canvas-strategies/strategies/absolute-reparent-strategy.tsx
seanparsons: Yeah, that's a better naming, change made in 9fe611e74.

editor/src/components/canvas/canvas-strategies/strategies/reparent-helpers/reparent-property-changes.ts
seanparsons: Parameter renamed in 788e1b602.

balazsbajorics (approved):
it would be so nice if we could verify that we managed to add contain-layout and have some sort of fallback / disabled behavior if not

gbalint (approved):

PR #6598, opened 2024-10-29
A Less Aggressive Auto-Collapse

balazsbajorics:
**Problem:**
We want a slightly more refined auto-collapse behaviour.

**Fix:**
The previous behaviour has been re-enabled with an additional check. The additional check is supported with data passed through from before any changes have been made to the element, so that it's possible to identify the change in situation.

**Commit Details:**
- Reverted the changes which disabled `pushIntendedBoundsAndUpdateHuggingElements` along with the changes from that PR.
- Added toast when fixed width and height added.
- Updated some tests.
- Moved `getHuggingElementContentsStatus` into its own file and
  extracted out the type `HuggingElementContentsStatus`.
- Changed type of `PushIntendedBoundsAndUpdateHuggingElements.value`.
- Added additional escape case in `applyUpdateResizeHuggingElementsCommands`.
- `deleteElements` stores some more values upfront in the `trueUpHuggingElement`
  value.
- Added `elementFrame` and `huggingElementContentsStatus` to `TrueUpHuggingElement`.


editor/src/components/canvas/commands/push-intended-bounds-and-update-hugging-elements-command.ts
```diff
@@ -128,39 +133,8 @@ function applyUpdateResizeHuggingElementsCommands(
           'height',
           frameAndTarget.frame.height,
         ),
+        showToastCommand('Added fixed width and height', 'NOTICE', 'added-width-height'), // TODO before merge verify this toast shows up
```
ruggi, line 136: 🥁

ruggi (approved):

editor/src/components/canvas/commands/push-intended-bounds-and-update-hugging-elements-command.ts
seanparsons: I knew there was something I had to check...

seanparsons: TODO removed in c3ea88573.

editor/src/components/canvas/commands/push-intended-bounds-and-update-targets-command.spec.browser2.tsx
```diff
@@ -251,6 +251,46 @@ describe('push intended bounds', () => {
...
+			export var storyboard = (
+				<Storyboard data-uid='sb'>
+					<div data-uid='grid' style={{ display: 'grid', gridTemplateColumns: '1fr', gridRowColumns: '1fr', alignItems: 'center', justifyContent: 'center', padding: 10, gap: 10, backgroundColor: 'white', width: 100, height: 100, position: 'absolute', left: 21, top: 17 }}>
+						<div data-uid='container' data-testid='container' style={{ backgroundColor: 'red', width: 80, height: 80 }} />
+					</div>
+				</Storyboard>
+			)
+		`),
+        )
+      })
```
balazsbajorics, line 293: @seanparsons can you add a test where we check if deleting the last grid-item of a grid prevents the grid from collapsing? (similar to this test, but without the intermediary `container` element). the grid can collapse if it has auto rows and columns, or if the column/row template has a `repeat` function with auto repeat count, and maybe some use of `minmax` can also lead to collapse

seanparsons: Tests added in [827f338](https://github.com/concrete-utopia/utopia/pull/6598/commits/827f338da189f579bfa41efcdbf095585fa1ca1b).

bkrmendy (approved):

gbalint (approved):

PR #6599, opened 2024-10-30
Set props through style plugin

bkrmendy:
See https://github.com/concrete-utopia/utopia/pull/6574 for a long-form writeup on the full background of this PR.

## The problem addressed on this PR
The normalisation step cannot easily handle commands/actions removing properties from elements.

## Fix
Instead of working around the limitation of the normalisation step, this PR introduces an approach where the style plugins themselves implement setting/removing style props on elements, and the `SetProperty` / `DeleteProperty` command uses the style plugin API to set/remove style props. The PR deals with patching props that are removed during a canvas interaction (details in https://github.com/concrete-utopia/utopia/pull/6574, TLDR: if `gap` is removed with the gap control, it needs to be set to `0px` while the interaction is in progress, otherwise the value from the Tailwind class will be applied which would be janky).

This is an under-the-hood PR, no user-facing functionality is added.

### Details
- Updated the `StylePlugin` interface with the `updateStyles` function, which can be used to set/delete element styles. The inline style plugin and the Tailwind style plugin are updated accordingly
- Injected `StylePlugin.updateStyles` into `runSetProperty` and `runDeleteProperties` via the new `runStyleUpdateForStrategy` function. This function encapsulates a performance-related feature: if an interaction is ongoing, it sets props in the inline style prop (and patches removed props), so updates are real-time, and when an interaction is done, it uses the active plugin. This trick works because the active strategy is re-run on interaction finish, producing the right patches for a style update.
- The machinery dealing with zeroing props is added. The entrypoint for adding the zeroed props is `patchRemovedProperties` (called from `dispatch-strategies.tsx`, and the entrypoint for recording which props need to be zeroed is `runStyleUpdateMidInteraction`
- The code related to the normalisation step is removed
- `UpdateClassListCommand` is removed and turned into a helper function (since it's not called as a command anymore, just as a helper in the Tailwind style plugin)

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/plugins/style-plugins.ts
```diff
@@ -1,33 +1,227 @@
...
+  property: string
+  value: string | number
+}
+
+interface DeleteCSSProp {
+  type: 'delete'
+  property: string
+}
+
+export const updateCSSProp = (property: string, value: string | number): UpdateCSSProp => ({
```
seanparsons, line 34: Can these not be regular functions?

editor/src/components/canvas/plugins/style-plugins.ts
```diff
@@ -1,33 +1,227 @@
...
+  }
+
+  return updatedPropertiesToExtend
+}
+
+function runStyleUpdateMidInteraction(
+  editorState: EditorState,
+  elementPath: ElementPath,
+  updates: StyleUpdate[],
+): { editorStateWithChanges: EditorState; editorStatePatches: EditorStatePatch[] } {
```
seanparsons, line 101: Is this the return type `EditorStateWithPatch`?

seanparsons (approved):

editor/src/components/canvas/plugins/style-plugins.ts
bkrmendy: `EditorStateWithPatch` only has a single patch. I factored out the return type into it's own interface, `EditorStateWithPatches`

editor/src/components/canvas/plugins/style-plugins.ts
```diff
@@ -1,33 +1,227 @@
...
+    editorState,
+  )
+}
+
+export function runStyleUpdateForStrategy(
+  commandLifecycle: InteractionLifecycle,
+  editorState: EditorState,
+  elementPath: ElementPath,
+  updates: StyleUpdate[],
+) {
```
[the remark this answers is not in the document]
bkrmendy: I added an explicit return type to that function in https://github.com/concrete-utopia/utopia/pull/6599/commits/d5913cac6f0d8c5153952b4651ff42936c2e75ea

editor/src/components/canvas/canvas-strategies/canvas-strategies.tsx
```diff
@@ -495,6 +495,16 @@ export function applyCanvasStrategy(
   return strategy.apply(strategyLifecycle)
 }
 
+export function applyElementsToRerenderFromStrategyResultAndZeroProps(
```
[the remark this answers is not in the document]
bkrmendy: I updated the name to `applyElementsToRerenderFromStrategyResultAndPatchRemovedProps` in https://github.com/concrete-utopia/utopia/pull/6599/commits/7ff07b72643b87e580c3719fe4f8112b8d80bb71

editor/src/components/canvas/commands/delete-properties-command.ts
```diff
@@ -24,19 +26,40 @@ export function deleteProperties(
...
+    },
+    { styleProps: [], rest: [] },
+  )
+
   // Apply the update to the properties.
-  const { editorStatePatch: propertyUpdatePatch } = deleteValuesAtPath(
+  const { editorStateWithChanges } = deleteValuesAtPath(
     editorState,
     command.element,
     command.properties,
```
balazsbajorics, line 51: shouldn't this only run for `propsToDelete.rest`?

bkrmendy: great catch, fixed in https://github.com/concrete-utopia/utopia/pull/6599/commits/09086039ca30b2d337f1b424ce71355a0bbf980c

editor/src/components/editor/store/editor-state.ts
```diff
@@ -1471,6 +1472,7 @@ export interface EditorState {
   collaborators: Collaborator[]
   sharingDialogOpen: boolean
   editorRemixConfig: EditorRemixConfig
+  propertiesUpdatedDuringInteraction: UpdatedProperties
```
balazsbajorics, line 1475: isn't this leftover?

balazsbajorics:
> The machinery dealing with zeroing props is added. The entrypoint for adding the zeroed props is patchRemovedProperties (called from dispatch-strategies.tsx, and the entrypoint for recording which props need to be zeroed is runStyleUpdateMidInteraction

why doesn't the `runStyleUpdateMidInteraction` inject the zeroing? feels like it is a leftover idea to run it in a later part of the update flow – IIUC the style plugin has all the information and all the power necessary to internally deal with adding these changes

editor/src/components/editor/store/editor-state.ts
bkrmendy: It's very much not a leftover, this prop is read by these functions https://github.com/concrete-utopia/utopia/blob/09086039ca30b2d337f1b424ce71355a0bbf980c/editor/src/components/canvas/plugins/style-plugins.ts#L77-L83 and written to here https://github.com/concrete-utopia/utopia/blob/09086039ca30b2d337f1b424ce71355a0bbf980c/editor/src/components/canvas/plugins/style-plugins.ts#L132-L138

balazsbajorics (approved):

bkrmendy:
> > The machinery dealing with zeroing props is added. The entrypoint for adding the zeroed props is patchRemovedProperties (called from dispatch-strategies.tsx, and the entrypoint for recording which props need to be zeroed is runStyleUpdateMidInteraction
> 
> why doesn't the `runStyleUpdateMidInteraction` inject the zeroing? feels like it is a leftover idea to run it in a later part of the update flow – IIUC the style plugin has all the information and all the power necessary to internally deal with adding these changes

We talked about this IRL, the reason this is needed is because the code that does the zeroing for padding needs to know about all the props that have been set/unset (see [this code](https://github.com/concrete-utopia/utopia/blob/0f4b889cd40f3c34041ac4e21c6a3cdfbdb877eb/editor/src/components/canvas/plugins/style-plugins.ts#L153-L170) from the spike)

PR #6602, opened 2024-10-30
Inserting/reparenting into a grid component

gbalint:
**Problem:**
Inserting/reparenting into a grid component doesn't work.

**Root cause:**
The problem is the following:
- We collect the global frames of grid cells using the measurement helper components (which render an invisible clone of the grid with placeholders in each cell, so they are measurable html elements)
- getSpecialSizeMeasurements needs to find the appropriate measurement helper in the dom
- Measurement helpers are identifiable because they have an id which includes the element path of the grid they refer to. This a path which is coming from the metadata of the element.
- getSpecialSizeMeasurements uses the element path from the data-path attribute of the measured grid element to find the measurement helper
- When the grid is a component, the rendered grid layouted div's data-path contains the element path of the root element of the component and not the path of the component itself (e.g. foo/bar/grid:root instead of foo/bar/grid). But the measurement helper's id includes the element path of the component (e.g. foo/bar/grid). Because of this, getSpecialMeasurements does not find the measurement helper for this grid, and it does not store the grid cell global frames in specialSizeMeasurements

**Fix:**
- Instead of using the data-path of the measured element, use the paths coming from `getPathsOnDomElement`, which lists all the possible utopia element paths connected to this rendered html element. Try to find the measurement helper using these paths (it is enough to use the first one which is found).
- For the measurement of parentGridCellGlobalFrames, just do the same thing on the parent element in the dom.
- I renamed `measureGlobalFramesOfGridCellsFromControl` because it doesn't use the grid control anymore (but the measurement helper).

This change fixes all draw-to-insert/click-to-insert/reparent functionalities.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/dom-walker.ts
+++ b/editor/src/components/canvas/dom-walker.ts
@@ -923,7 +923,7 @@ function getSpecialMeasurements(
 
   const gridCellGlobalFrames =
     layoutSystemForChildren === 'grid'
-      ? measureGlobalFramesOfGridCellsFromControl(
+      ? measureGlobalFramesOfGridCells(
           element,
           scale,
           containerRectLazy,
@@ -933,7 +933,7 @@ function getSpecialMeasurements(
 
   const parentGridCellGlobalFrames =
     element.parentElement != null && elementLayoutSystem(parentElementStyle) === 'grid'
-      ? measureGlobalFramesOfGridCellsFromControl(
+      ? measureGlobalFramesOfGridCells(
           element.parentElement,
           scale,
           containerRectLazy,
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/dom-walker.ts
```diff
@@ -1081,15 +1076,25 @@ function getClosestOffsetParent(element: HTMLElement): Element | null {
...
   containerRectLazy: CanvasPoint | (() => CanvasPoint),
   elementCanvasRectangleCache: ElementCanvasRectangleCache,
 ): GridCellGlobalFrames | null {
-  const path = EP.toString(gridPath)
+  const paths = getPathsOnDomElement(element)
+
   let gridCellGlobalFrames: GridCellGlobalFrames | null = null
-  const gridControlElement = document.getElementById(GridMeasurementHelperKey(EP.fromString(path)))
+  const gridControlElement = (() => {
+    for (let p of paths) {
```
seanparsons, line 1089: Does this work with nested grids?

gbalint: Absolutely! These paths are all utopia element paths which render exactly to this html element, so there is no overlap between the paths of different grids.

bkrmendy (approved):

liady (approved):

PR #6603, opened 2024-10-30
feat(import): check for server dependencies

liady:
**Note for reviewers:**
Other than refactors, tests and name changes, the **actual logic** is only in [requirement-server-packages.ts](https://github.com/concrete-utopia/utopia/compare/feat/check-server-dependencies?expand=1#diff-a2a88b7fd33ba102c98948e8efcb2c1771061df92f39e2b99af55560a1aceffc).
(I had to change the resolution name from `'Found'` to `'Passed'` since here a passing check means that server packages weren't actually found)

**PR details:**
This PR adds a check for server dependencies to the import process - currently a very closed list, to be added more incrementally.
For example when importing a Next project:
<video src="https://github.com/user-attachments/assets/4db242d8-bb51-46e0-892c-d0a84d4c5231"></video>

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/assets.ts
+++ b/editor/src/components/assets.ts
@@ -761,3 +761,36 @@ export function ensureDirectoriesExist(projectContents: ProjectContents): Projec
     return result
   }
 }
+
+function getFileAsJson<T>(projectContents: ProjectContentTreeRoot, fileName: string): T | null {
+  const file = getProjectFileByFilePath(projectContents, fileName)
+  if (file != null && isTextFile(file)) {
+    return JSON.parse(file.fileContents.code) as T
+  }
+  return null
+}
+
+type PackageJson = { utopia?: Record<string, string>; dependencies?: Record<string, string> }
+export function getPackageJson(projectContents: ProjectContentTreeRoot): PackageJson | null {
+  return getFileAsJson<PackageJson>(projectContents, '/package.json')
+}
+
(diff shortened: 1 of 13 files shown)


editor/src/components/assets.ts
```diff
@@ -761,3 +761,36 @@ export function ensureDirectoriesExist(projectContents: ProjectContents): Projec
     return result
   }
 }
+
```
liady, line 764: these were just moved here from `check-utopia-requiremenets.ts`

editor/src/core/shared/import/proejct-health-check/requirements/requirement-server-packages.ts
```diff
@@ -0,0 +1,29 @@
+import type { ProjectContentTreeRoot } from 'utopia-shared/src/types'
+import {
+  RequirementResolutionResult,
+  type RequirementCheck,
+  type RequirementCheckResult,
+} from '../utopia-requirements-types'
+import { getProjectDependencies } from '../../../../../components/assets'
+
+const serverPackagesRestrictionList: RegExp[] = [/^next/, /^remix/, /^astro/, /^svelte/]
```
Rheeseyb, line 9: Why is remix in this list?

Rheeseyb (approved):

PR #6609, opened 2024-10-31
feat(import): detect node built-ins

liady:
This PR adds the ability to detect Node built-ins that are being imported.

**Notes:**
- We currently detect all imports in project files, without checking if the importing file is in the import chain of the app itself. This _may_ result in false positives, if one of the files in the project that does not belong to the app imports a Node module.
- For that reason - currently this only triggers a "warning" and not a critical error.
- We compare the built-ins that we find with the project dependencies, in case some of them are shimmed.
- We're using a static list taken from here: https://github.com/sindresorhus/builtin-modules. We can't use the package directly since it is an ESM module, and our tests cannot import it, and in any case all it does is exporting this JSON.

If you clone this branch you can test it using http://localhost:8000/p?accessLevel=public&clone=liady/utopia-vite-project - I added an `import * as fs from 'fs'` there to test this functionality. (don't forget to turn on the `Import Wizard` feature flag)

image

**Manual Tests:**
I hereby swear that:


editor/pnpm-lock.yaml
```diff
@@ -4640,7 +4640,7 @@ packages:
     dependencies:
       '@rollup/pluginutils': 3.1.0
       '@types/resolve': 1.17.1
-      builtin-modules: 3.2.0
+      builtin-modules: 3.3.0
```
Rheeseyb, line 4643: Looks like this was added in error?

Rheeseyb (approved):

editor/pnpm-lock.yaml
liady: oh yes, looks like a leftover that didn't clean when I `pnpm uninstall`-ed it. thanks!

seanparsons (approved):

PR #6610, opened 2024-10-31
Disable grid control strategies for components which don't support style props

gbalint:
**Problem:**
Now it is possible to use the gap and the grid resize controls on components which do not support style props.

**Fix:**
We somehow have to decide whether a component supports style props or not. I implemented the following heuristics, but we can improve that later if necessary:
1. If a component has a descriptor and it allows 'layout-system' section to be shown then it supports style props.
2. If it doesn't have a descriptor then we check if the component implementation takes the necessary style props (similarly to the already existing `targetSupportsPropsSize` function). This returns true when the component implementation can not be found (if it is an external component)

Note: When the component does not support the necessary style props:
- I don't show the gap control, but it is just misleading if it is there
- I show the grid resize controls, because they show information (row/column size), which is meaningful even if the control is not interactive.
These are just draft solution, so a followup task is necessary to design/implement the the non-interactive state of the controls

-`targetRegisteredStyleControlsOrHonoursStyleProps` is the main function which decides whether a component supports style props for the actual control. You can give an inspector section to check (if that is enabled from the annotation, it returns true), and  a list of props (if they are referred inside the component we return true).

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/core/model/element-metadata-utils.ts
+++ b/editor/src/core/model/element-metadata-utils.ts
@@ -1,5 +1,5 @@
 import * as OPI from 'object-path-immutable'
-import type { Emphasis, FlexLength, Icon, Sides } from 'utopia-api/core'
+import type { Emphasis, FlexLength, Icon, Sides, Styling } from 'utopia-api/core'
 import { sides } from 'utopia-api/core'
 import { getReorderDirection } from '../../components/canvas/controls/select-mode/yoga-utils'
 import { getImageSize, scaleImageDimensions } from '../../components/images'
@@ -113,6 +113,7 @@ import {
   componentUsesProperty,
   findJSXElementChildAtPath,
   elementChildSupportsChildrenAlsoText,
+  componentHonoursStyleProps,
 } from './element-template-utils'
 import {
   isImportedComponent,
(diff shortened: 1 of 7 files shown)


editor/src/core/model/element-template-utils.ts
```diff
@@ -969,6 +969,46 @@ export function componentHonoursPropsSize(component: UtopiaJSXComponent): boolea
...
 }
 
+export function componentHonoursStyleProps(
+  component: UtopiaJSXComponent,
+  props: Array<string>,
+  everyOrSome: 'every' | 'some',
+): boolean {
+  if (component.params == null) {
+    return false
+  } else {
```
ruggi, line 979: this case imho is not needed here, happy path to the left would be more readable in this case imo

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1212,6 +1213,70 @@ export const MetadataUtils = {
...
+        descriptor.inspector.type === 'hidden' ||
+        !descriptor.inspector.sections.includes(inspectorSection)
+      ) {
+        return 'registered-to-hide'
+      }
+      return 'registered-to-show'
+    })()
+
+    switch (inspectorSectionRegistered) {
+      case 'registered-to-show': //
```
ruggi, line 1270: is `//` meant to have a comment or can it be removed?

editor/src/core/model/element-metadata-utils.ts
```diff
@@ -1212,6 +1213,70 @@ export const MetadataUtils = {
...
+    metadata: ElementInstanceMetadata | null,
+    props: Array<string>,
+    everyOrSome: 'every' | 'some',
+  ): boolean {
+    if (metadata == null) {
+      return false
+    }
+    if (
+      isLeft(metadata.element) ||
+      (isRight(metadata.element) && !isJSXElement(metadata.element.value))
```
ruggi, line 1227:
```suggestion
      (!isJSXElement(metadata.element.value))
```

ruggi (approved):

liady (approved):

PR #6613, opened 2024-11-01
Restore Toast Renderer

seanparsons:
**Problem:**
Toasts aren't displaying.

**Fix:**
The `ToastRenderer` has been re-added into the editor hierarchy.

**Commit Details:**
- Readded `ToastRenderer` to `CanvasFloatingToolbars`.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/canvas-floating-toolbars.tsx
+++ b/editor/src/components/canvas/canvas-floating-toolbars.tsx
@@ -6,6 +6,7 @@ import { ErrorOverlayComponent } from './canvas-error-overlay'
 import { SafeModeErrorOverlay } from './canvas-wrapper-component'
 import { CanvasStrategyPicker } from './controls/select-mode/canvas-strategy-picker'
 import { TestMenu } from '../titlebar/test-menu'
+import { ToastRenderer } from '../editor/editor-component'
 
 export const CanvasFloatingToolbars = React.memo((props: { style: React.CSSProperties }) => {
   const safeMode = useEditorState(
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-floating-toolbars.tsx
```diff
@@ -43,6 +44,11 @@ export const CanvasFloatingToolbars = React.memo((props: { style: React.CSSPrope
           <CanvasStrategyPicker />
         </FlexRow>
       </FlexRow>
+      <FlexRow style={{ marginTop: `auto` }}>
+        <FlexRow style={{ marginBottom: '6.5px' }}>
```
Rheeseyb, line 48: Can these be squashed into a single `FlexRow`, or does that actually mess up the style?

Rheeseyb (approved):

editor/src/components/canvas/canvas-floating-toolbars.tsx
seanparsons: I don't believe so, I tried using `calc-size` but that didn't work at all.

liady: btw, I tried a single `FlexRow` with `style: {{ marginTop: 'auto', marginBottom: '6.5px' }}` and it seems like it looks the same... unless there is another special case where they are differ?
this is with this toast:
```tsx
<React.Fragment>
  <div>this is a test</div>
  <div>this is another test</div>
</React.Fragment>
```
image

seanparsons: I stand corrected, applied this change in 933bccd4d.

bkrmendy (approved):

liady (approved):

PR #6617, opened 2024-11-05
Grid Child Pin Outlines

seanparsons:
**Problem:**
For the children of a grid, the pin outlines work in some slightly unusual ways, which means they don't work as they currently are implemented for just those elements.

**Fix:**
This mostly revolves around building an element which is sized to the containing block as described here:
https://www.w3.org/TR/css-grid-2/#abspos-items
Then render the pin outlines within that, so that the browser does most of the work for us. This utilises the component `PinOutline` but not it's containing components as they are used normally but has a grid specific implementation instead.

Additionally it was noticed that `parseGridRange` was creating a value which didn't match the spec or browser behaviour where a single value such as `2` would be treated as `2 / 3` when it should be treated as `2 / auto`.

**Screenshot:**
image

**Commit Details:**
- Fixed `parseGridRange` to more closely reflect the behaviour in browsers.
- Renamed `elementStyle` to `computedStyle` in `getSpecialMeasurements` for clarity.
- `PinLines` excludes grid cells from those elements which will get regular pin outlines.
- `PinOutlineProps` now has `endX` and `endY` properties, along with supporting `string` values for most of the properties.
- `PinOutline` now defers a lot of the calculations down into the CSS.
- Added `GridElementContainingBlocks` into the controls hierarchy.
- Implemented `collectGridPinOutlines` which handles the more complicated case that grids present.
- Implemented `GridElementContainingBlock` to render `PinOutline` instances for the grid children.

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/canvas-strategies/strategies/grid-rearrange-move-strategy.spec.browser2.tsx
```diff
@@ -506,8 +506,8 @@ export var storyboard = (
         {
           const { top, left, gridColumn, gridRow } = child.style
           expect({ top, left, gridColumn, gridRow }).toEqual({
-            gridColumn: '1',
-            gridRow: '1',
+            gridColumn: '1 / 3',
+            gridRow: '1 / 3',
```
balazsbajorics, line 510: where do these `/ 3`s come from?

seanparsons: I looked into this a bit and the element being dragged around spans over two cells in each direction.

Previously these `1`s would've been treated as `1 / 2`, but because of the change I made to `parseGridRange` it would be treated as `1 / auto`. So I assume the logic is correcting those to what they are as a result. I tried the code and moved it to other locations and these changed to `2 / 4` and so forth, I didn't dig into it any further because it appears to be correct.

balazsbajorics: (we chatted in DMs and realized it's not based on the size of the element, the size 2 seems to come from a mistake)

editor/src/components/canvas/canvas-strategies/strategies/grid-rearrange-move-strategy.spec.browser2.tsx
```diff
@@ -504,10 +504,10 @@ export var storyboard = (
         await mouseUpAtPoint(dragTarget, endPoint)
 
         {
-          const { top, left, gridColumn, gridRow } = child.style
-          expect({ top, left, gridColumn, gridRow }).toEqual({
-            gridColumn: '1',
-            gridRow: '1',
+          const { top, left, gridColumnStart, gridRowStart } = child.style
+          expect({ top, left, gridColumnStart, gridRowStart }).toEqual({
+            gridColumnStart: '1',
```
balazsbajorics, line 509: if we do the longhand route, we should also check gridColumnEnd and gridRowEnd here (I presume they are undefined!)

editor/src/components/canvas/controls/grid-controls.spec.browser2.tsx
```diff
@@ -0,0 +1,343 @@
...
+      gridColumnEnd: 'auto',
+      width: 50,
+      height: 50,
+    })
+    expect(result).toEqual({
+      left: { x: 688.5, y: 315.5, width: 12, height: 1 },
+      top: { x: 725.5, y: 285.5, width: 1, height: 5 },
+    })
+  })
+  it('pinned top and right, grid column end set to auto and not absolutely positioned', async () => {
```
balazsbajorics, line 253: for the "...end set to auto" tests, we should add a padding to the grid parent – because `auto` corrseponds to the padding edge, which is only different from the last line if there is a padding set

editor/src/components/canvas/canvas-strategies/strategies/grid-rearrange-move-strategy.spec.browser2.tsx
seanparsons: Added check for those fields to 7838f8df1.

editor/src/components/canvas/controls/position-outline.tsx
```diff
@@ -182,36 +183,62 @@ const collectPinOutlines = (
...
+          PinOutlineUnscaledSize / 2 / props.scale
+        }px)`
+  const lineRight =
+    props.endX == null
+      ? undefined
+      : `calc(${numberOrStringToSize(props.endX)} - ${PinOutlineUnscaledSize / 2 / props.scale}px)`
+  const lineBottom =
+    props.endY == null
+      ? undefined
+      : `calc(${numberOrStringToSize(props.endY)} - ${PinOutlineUnscaledSize / 2 / props.scale}px)`
```
bkrmendy, line 232: I would extract the ternaries here into a function for better readability

bkrmendy (approved):

editor/src/components/canvas/controls/position-outline.tsx
seanparsons: Refactored in e722c0823.

PR #6618, opened 2024-11-06
feat(import): split checks

liady:
This PR splits the validation checks to allow an early break before parsing the files.
In the UI they are split to these phases:
- **Validating code** (which are actually the pre-parse validations)
- Parse files (as today)
- **Checking Utopia requirements** (which are the post-parse validations)

**Details:**
1. The requirements are still aggregated in the editor state, but they are separated to phases for running as well as for displaying in the wizard ([see the separation here](https://github.com/concrete-utopia/utopia/pull/6618/files#diff-15c53165ea0f64ccf6f092783f2dce88639666dd1f4fd29e46035dc96ca4f2b3R20-R30)).
2. Main changes are in [editor/src/core/shared/import/project-health-check/utopia-requirements-service.ts](https://github.com/concrete-utopia/utopia/pull/6618/files#diff-9691feb390abe09b4d2b2aca374b464cf8bcb91a10942f8f6f07a2b4e9355fa6), where the split actually happens (`startPreParseValidation` and `startPostParseValidation`).
3. Renaming the requirement phases added changes in several other files, but they are simply refactors.
4. Minor - we also remove the shimmer when pausing the wizard (to make it clearer that the import has stopped).

See here - first a successful project and then one with errors:
<video src="https://github.com/user-attachments/assets/a907fe6b-68be-44a9-8a01-7e951ed90001"></video>

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/canvas-loading-screen.tsx
+++ b/editor/src/components/canvas/canvas-loading-screen.tsx
@@ -3,7 +3,6 @@ import { Global, css } from '@emotion/react'
 import { useColorTheme } from '../../uuiui'
 import { useEditorState } from '../editor/store/store-hook'
 import { Substores } from '../editor/store/store-hook'
-import cx from 'classnames'
 
 export const CanvasLoadingScreen = React.memo(() => {
   const colorTheme = useColorTheme()
@@ -18,9 +17,18 @@ export const CanvasLoadingScreen = React.memo(() => {
     'CanvasLoadingScreen importWizardOpen',
   )
 
-  const importingStopped =
-    (importWizardOpen && importState.importStatus.status === 'done') ||
-    importState.importStatus.status === 'paused'
+  const importingStoppedStyleOverride = React.useMemo(
+    () =>
+      // if the importing was stopped, we want to pause the shimmer animation
+      (importWizardOpen && importState.importStatus.status === 'done') ||
+      importState.importStatus.status === 'paused'
+        ? {
+            background: colorTheme.codeEditorShimmerPrimary.value,
+            animation: 'none',
+          }
+        : {},
+    [importWizardOpen, importState.importStatus.status, colorTheme.codeEditorShimmerPrimary.value],
+  )
 
   return (
     <React.Fragment>
@@ -64,13 +72,14 @@ export const CanvasLoadingScreen = React.memo(() => {
         style={{ height: '100%', width: '100%', backgroundColor: colorTheme.bg1.value }}
       >
         <div
-          className={cx('shimmer', { 'no-shimmer': importingStopped })}
+          className='shimmer'
           style={{
             position: 'absolute',
             left: 0,
             top: 0,
             width: '100vw',
             height: '100vh',
+            ...importingStoppedStyleOverride,
           }}
         ></div>
       </div>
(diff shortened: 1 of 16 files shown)


editor/src/components/canvas/canvas-loading-screen.tsx
```diff
@@ -34,14 +52,19 @@ export const CanvasLoadingScreen = React.memo(() => {
             background-size: 1468px 104px;
             position: relative;
           }
+
+          .no-shimmer {
```
seanparsons, line 56: Couldn't this be conditionally included in the style prop?

editor/src/components/editor/import-wizard/components.tsx
```diff
@@ -82,7 +80,8 @@ function OperationChildrenList({ operation }: { operation: ImportOperation }) {
           successFn={dependenciesSuccessFn}
           successTextFn={dependenciesSuccessTextFn}
         />
-      ) : operation.type === 'checkRequirements' ? (
+      ) : operation.type === 'checkRequirementsPreParse' ||
+        operation.type === 'checkRequirementsPostParse' ? (
```
seanparsons, line 84: I think this would be better served with a function that returns these components that has an exhaustive switch, rather than chaining ternaries. Then new operation types get handled appropriately and I think it would make this expression a little bit less complicated.

editor/src/components/editor/import-wizard/import-wizard.tsx
```diff
@@ -238,34 +219,45 @@ function ActionButtons({ importResult }: { importResult: TotalImportResult }) {
       </React.Fragment>
     )
   }
-  if (
-    importResult.result == ImportOperationResult.Error ||
-    importResult.result == ImportOperationResult.CriticalError
-  ) {
+  if (importResult.result == ImportOperationResult.CriticalError) {
```
seanparsons, line 222: This is definitely a case where a switch statement would make more sense.

editor/src/core/shared/github/operations/load-branch.ts
```diff
@@ -186,20 +191,33 @@ export const updateProjectWithBranchContent =
             }
             notifyOperationFinished(dispatch, { type: 'loadBranch' }, ImportOperationResult.Success)
 
+            // pre parse validation
```
seanparsons, line 194: Love these comments!

editor/src/core/shared/import/project-health-check/utopia-requirements-service.ts
```diff
@@ -33,12 +36,29 @@ export function updateProjectRequirementsStatus(
...
   notifyOperationStarted(dispatch, {
-    type: 'checkRequirements',
-    children: Object.keys(projectRequirements).map((key) =>
-      importCheckRequirementAndFix(
+    type: 'checkRequirementsPreParse',
+    children: Object.keys(checks).map((key) =>
+      importCheckRequirementAndFixPreParse(
         key as ProjectRequirement,
-        initialTexts[key as ProjectRequirement],
+        checks[key as PreParseValidationRequirement].initialText,
```
seanparsons, line 48: I would be very hesitant to use these casts because they absolutely will break things in weird ways if the value of `checks` changes.

seanparsons (approved):

Rheeseyb (approved):

PR #6622, opened 2024-11-06
Include border widths when measuring grid templates for controls

### LINKED ISSUE #6619: Include border widths for grid measurement

ruggi:
**Problem:**

Border width is not taken into account when calculating grid template measurements, which results in the grid placeholders being misplaced.

**Fix:**

Include the border widths in the special size measurements and use them (transparent) for the grid placeholder styles.

Fixes #6619


bkrmendy (approved):

gbalint:
Instead of rendering the grid helper/control from the metadata, maybe we could just find the actual grid in the DOM and use its computed style to render the appropriate helper grid. It would be simpler (we don't have to save the style data in the metadata and then recreate that again in the control) and better too (now theoretically there is a frame delay before the control is updated, because we need to dom-walk the grid and store the results in the metadata first). WDYT?

balazsbajorics (approved):

seanparsons (approved):

liady (approved):

PR #6623, opened 2024-11-06
Rename grid cell to grid item

gbalint:
**Problem:**
Sometimes we call the grid items "grid cells", which is incorrect (grid cells are the positions in the grid, and not the child items).

**Fix:**
Rename

**Manual Tests:**
I hereby swear that:


bkrmendy (approved):

balazsbajorics (approved):

liady (approved):
😌

PR #6627, opened 2024-11-08
Direct grid lookup.

seanparsons:
**Problem:**
Currently with the grid controls there is a 1-frame behind problem, where they will always render one frame after their respective grids. This is because they get their data from the DOM walker which in turn relies on the rendered content of the canvas.

**Fix:**
Rather than relying on the metadata that comes from the DOM walker, this moves the lookups to be directly called by the controls themselves. The two hooks `useGridData` and `useGridMeasurementHelperData` mostly got the same information from the metadata and that data is now obtained by `gridMeasurementHelperDataFromElement`. Which is supported by `getFromElement`.

As the re-render triggering has been lost because it's not connected to the metadata, this PR adds `useMonitorChangesToElements` as a way to regain that re-render triggering.

**Commit Details:**
- Moved some fields from `GridData` up to `GridMeasurementHelperData`.
- Implemented `gridMeasurementHelperDataFromElement` and `getGridMeasurementHelperData`, which perform the calculations needed by the controls.
- `useGridMeasurementHelperData` now reimplemented to use `getGridMeasurementHelperData`.
- `useGridData` now reimplemented to use `getGridMeasurementHelperData`.
- Implemented `getFromElement` as a hopefully reusable function for similar solutions.
- Implemented `useMonitorChangesToElements` as a solution for how to trigger the re-renders.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/controls/grid-controls-for-strategies.tsx
+++ b/editor/src/components/canvas/controls/grid-controls-for-strategies.tsx
@@ -53,7 +53,11 @@ import {
 import { applicative4Either, defaultEither, isRight, mapEither } from '../../../core/shared/either'
 import { domRectToScaledCanvasRectangle, getRoundingFn } from '../../../core/shared/dom-utils'
 import Utils from '../../../utils/utils'
-import { useMonitorChangesToElements } from '../../../components/editor/store/store-monitor'
+import {
+  useMonitorChangesToEditor,
+  useMonitorChangesToElements,
+} from '../../../components/editor/store/store-monitor'
+import { useKeepReferenceEqualityIfPossible } from '../../../utils/react-performance'
 
 export const GridCellTestId = (elementPath: ElementPath) => `grid-cell-${EP.toString(elementPath)}`
 
@@ -196,18 +200,18 @@ export function gridMeasurementHelperDataFromElement(
   }
 }
 
-export function useGridMeasurentHelperData(
+export function useGridMeasurementHelperData(
   elementPath: ElementPath,
 ): GridMeasurementHelperData | undefined {
   const scale = useEditorState(
     Substores.canvas,
     (store) => store.editor.canvas.scale,
-    'useGridMeasurentHelperData scale',
+    'useGridMeasurementHelperData scale',
   )
 
-  useMonitorChangesToElements([elementPath])
+  useMonitorChangesToEditor()
 
-  return getGridMeasurementHelperData(elementPath, scale)
+  return useKeepReferenceEqualityIfPossible(getGridMeasurementHelperData(elementPath, scale))
 }
 
 export type GridData = GridMeasurementHelperData & {
@@ -220,8 +224,6 @@ export function useGridData(elementPaths: ElementPath[]): GridData[] {
     (store) => store.editor.canvas.scale,
     'useGridData scale',
   )
-
-  useMonitorChangesToElements(elementPaths)
 
   const grids = useEditorState(
     Substores.metadata,
@@ -257,7 +259,7 @@ export function useGridData(elementPaths: ElementPath[]): GridData[] {
     'useGridData',
   )
 
-  return grids
+  return useKeepReferenceEqualityIfPossible(grids)
 }
 
 interface GridRowColumnResizingControlsProps {
--- a/editor/src/components/canvas/controls/grid-controls-for-strategies.tsx
+++ b/editor/src/components/canvas/controls/grid-controls-for-strategies.tsx
@@ -53,11 +53,9 @@ import {
 import { applicative4Either, defaultEither, isRight, mapEither } from '../../../core/shared/either'
 import { domRectToScaledCanvasRectangle, getRoundingFn } from '../../../core/shared/dom-utils'
 import Utils from '../../../utils/utils'
-import {
-  useMonitorChangesToEditor,
-  useMonitorChangesToElements,
-} from '../../../components/editor/store/store-monitor'
+import { useMonitorChangesToEditor } from '../../../components/editor/store/store-monitor'
 import { useKeepReferenceEqualityIfPossible } from '../../../utils/react-performance'
+import type { CSSProperties } from 'react'
 
 export const GridCellTestId = (elementPath: ElementPath) => `grid-cell-${EP.toString(elementPath)}`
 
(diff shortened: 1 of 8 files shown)


editor/src/components/editor/store/store-monitor.ts
```diff
@@ -0,0 +1,23 @@
...
+          null,
+          (_, element) => {
+            return element
+          },
+        )
+      }, elementPaths)
+    },
+    'useMonitorChangesToElements elements',
+  )
+}
```
gbalint, line 23: This means that the rerender is dependent on project code change, but the positioning of a grid can change without that if it depends on the state of the component, or on loading external data (e.g. what if you select a grid before the metaobjects which fill it are loaded?), maybe even on remix route

editor/src/components/canvas/controls/grid-controls-for-strategies.tsx
```diff
@@ -68,136 +88,170 @@ export type GridMeasurementHelperData = {
...
-    'useGridMeasurentHelperData',
-    fastDeepEqual, //TODO: this should not be needed, but it seems EditorStateKeepDeepEquality is not running, and metadata reference is always updated
+): GridMeasurementHelperData | undefined {
+  const scale = useEditorState(
+    Substores.canvas,
+    (store) => store.editor.canvas.scale,
+    'useGridMeasurentHelperData scale',
   )
+
+  useMonitorChangesToElements([elementPath])
```
gbalint, line 208: I'm afraid this is going to rerender the helpers/controls even during interactions when the grid is not in elements-to-rerender

seanparsons: If there's a change to the grid, then it should be in elements-to-rerender surely?

editor/src/components/editor/store/store-monitor.ts
seanparsons: I've implemented an alternative approach for triggering in 952e26eff. But also added a `useKeepReferenceEqualityIfPossible` call around the result of both `useGridData` and `useGridMeasurementHelperData` so that we don't re-render the controls themselves unnecessarily.

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -1057,8 +1051,7 @@ export interface GridMeasurementHelperProps {
 }
 
 const GridMeasurementHelper = React.memo<{ elementPath: ElementPath }>(({ elementPath }) => {
-  const gridData = useGridMeasurentHelperData(elementPath)
-
+  const gridData = useGridMeasurementHelperData(elementPath)
```
gbalint, line 1054: This feels overcomplicated to me. We basically dom-walk the element, collect a subset of the metadata from the computedStyle, and then recreate the styling from that with the reversal of that.
I think it would be way simpler to just copy the computedStyle values of the found grid dom element into the style props of the helper/control directly.

gbalint: So I would not use the same code in the dom-sampler and here, because here it is not necessary to have the metadata, we just need the raw computedStyle.

seanparsons: We can't copy the properties over directly from one to the other as `getGridHelperStyleMatchingTargetGrid` which uses this data does do some manipulation like in the case of making the borders transparent.

There's also the `XFromProps` values which are constructed from some of the same functions as those values are passed down into the `GridResizing` component.

gbalint: Even if we don't directly copy all computedStyle, just property by property, even then it is way simpler and safer to e.g. just copy the gridTemplateRows instead of parsing/converting it to our metadata format and then print it back.

editor/src/components/canvas/controls/grid-controls-for-strategies.tsx
```diff
@@ -68,142 +92,174 @@ export type GridMeasurementHelperData = {
...
+export function useGridMeasurementHelperData(
+  elementPath: ElementPath,
+): GridMeasurementHelperData | undefined {
+  const scale = useEditorState(
+    Substores.canvas,
+    (store) => store.editor.canvas.scale,
+    'useGridMeasurementHelperData scale',
   )
+
+  useMonitorChangesToEditor()
```
balazsbajorics, line 212: isn't useMonitorChangesToElements good enough here?
even if you call useKeepReferenceEqualityIfPossible on the return value, the component calling this hook will be forced to rerender by useMonitorChangesToEditor

balazsbajorics (approved):
I agree with Balint that we can probably simplify some types in the style getter, but I don't think that's a blocker of this PR. I haven't (yet) had the time to actually play with this, but the code looks solid to me

editor/src/components/canvas/controls/grid-controls-for-strategies.tsx
seanparsons: Without this, it meant anything slightly more funky like asynchronously updated values would leave the helpers out of date. Even something simple like a sibling of the grid expanding and moving the grid as a result would leave the helpers in a weird state.

gbalint (approved):

editor/src/components/canvas/controls/grid-controls.tsx
seanparsons: I've done a slightly blended approach here in 0cc20b3e7. It attempts to use as much as seemed reasonable in taking the computed style directly, but still keeps a lot of the fields from this as a lot of the controls still use parsed and calculated values.

editor/src/components/canvas/controls/select-mode/grid-gap-control-component.tsx
```diff
@@ -220,8 +220,8 @@ const GridRowOrColumnHighlight = (props: {
   beingDragged: boolean
   onMouseOver: () => void
   elementHovered: boolean
-  gridJustifyContent: FlexJustifyContent | null
-  gridAlignContent: AlignContent | null
+  gridJustifyContent: string | undefined
```
liady, line 223: do we want to keep them untyped? (since for example in the metadata we do use typings for justify/align). we can maybe (later) have typed functions (getJustifyValue, getAlignValue) that return the value from the computed style as the appropriate type

liady (approved):

editor/src/components/canvas/direct-dom-lookups.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getDeepestPathOnDomElement } from '../../core/shared/uid-utils'
+import * as EP from '../../core/shared/element-path'
+import type { ElementPath } from 'utopia-shared/src/types'
+
+export type FromElement<T> = (element: HTMLElement) => T
+
+export function getFromElement<T>(path: ElementPath, fromElement: FromElement<T>): T | undefined {
+  const pathString = EP.toString(path)
+  const elements = document.querySelectorAll(
+    `#${CanvasContainerID} [${UTOPIA_PATH_KEY}^="${pathString}"]`,
```
liady, line 12: in a nested query it's slightly more efficient to use 
```js
document.getElementById(CanvasContainerID).querySelectorAll(`[${UTOPIA_PATH_KEY}^="${pathString}"]`)
```
(since there is no upwards lookup),
but it's really minor.

editor/src/components/canvas/controls/select-mode/grid-gap-control-component.tsx
seanparsons: These are just matching the values from the computed style, so that we can use them directly.

PR #6630, opened 2024-11-11
feat(loading): merge github fetch with the loading bar

liady:
This PR adds the loading stages to the loading bar screen and progress bar.

Stages for GH import:
1. Loading Editor...
2. Fetching <repo name>
3. Parsing files
4. Fetching dependencies

Stages for project load (we reverse the order in our project load):
1. Fetching dependencies
2. Parsing files

If everything is validated correctly - it just shows the editor with the project ready to be worked on. If not - it shows an error modal (its UI was slightly changed in #6677 to better reflect errors).

It also adds support for Dark mode.

**Note:**
- The new logic is almost entirely in [editor/src/components/editor/loading-screen.tsx](https://github.com/concrete-utopia/utopia/pull/6630/files#diff-5aa4d7e811e0bfb07a60910e517fe3a01cd270c9b4fd05132e513d939d69f769)
- There is a jump in the progress bar since these are actually two DOM elements, one that is rendered before the React is loaded and a React component that replaces it after the React code is loaded. This jump will be fixed in a subsequent PR.
- Currently the progress bar is "static" (advances by time and not stages). This will be changed in (the same) subsequent PR.

<video src="https://github.com/user-attachments/assets/a6d52f3d-b27b-4110-be1e-a7f213f563ab"></video>

**Manual Tests:**
I hereby swear that:


editor/src/components/editor/loading-screen.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+  if (!hasMounted.current) {
+    portal.innerHTML = ''
+    hasMounted.current = true
+  }
+
+  const themeStyle =
+    currentTheme === 'dark'
+      ? `
+    .editor-loading-screen { background-color: ${colorTheme.bg6.value} }
```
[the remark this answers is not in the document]
liady: in a sense this is "patching", since we want to match the CSS behavior in [index.html](https://github.com/concrete-utopia/utopia/blob/a1c03084719bdfff4757fb8fc3d2950817941f87/editor/src/templates/index.html#L94-L112) where the dark mode "overrides" the light mode defaults (that also apply here).
I didn't want to change anything in case we have a light theme, so to not create a possible shift in color from the light "default" that is defined in the base `index.html` css.

editor/src/components/editor/actions/actions.tsx
```diff
@@ -6319,6 +6322,7 @@ export async function load(
   // this action is now async!
   const migratedModel = applyMigrations(model)
   const npmDependencies = dependenciesWithEditorRequirements(migratedModel.projectContents)
+  notifyOperationStarted(dispatch, { type: 'refreshDependencies' })
```
balazsbajorics, line 6325: in actions.tsx, please mark all side effects with a comment to make future debugging easier – by default we assume the update functions are pure

liady: thanks! marked.

PR #6632, opened 2024-11-12
Make `AdjustCssLengthProperties` style prop-agnostic

bkrmendy:
## Problem
The padding strategy relies on the `AdjustCssLengthProperties` command to update elements, both for reading and writing styles. In order for that to work in Tailwind projects, where elements don't have an inline style prop, `AdjustCssLengthProperties`  needs to read element styles from the "right" property, otherwise the command wouldn't work as intended.

## Fix
Use the StyleInfo system from to read the element style info.

### Details
- Refactored `AdjustCssLengthProperties`  to read elements styles through `styleInfo`
...

### Manual Tests:
I hereby swear that:

- [ ] I could navigate to various routes in Play mode

### DIFF
--- a/editor/src/components/canvas/commands/utils/property-utils.ts
+++ b/editor/src/components/canvas/commands/utils/property-utils.ts
@@ -5,6 +5,9 @@ import { setJSXValuesAtPaths, unsetJSXValuesAtPaths } from '../../../../core/sha
 import type { EditorState, EditorStatePatch } from '../../../editor/store/editor-state'
 import { modifyUnderlyingElementForOpenFile } from '../../../editor/store/editor-state'
 import { patchParseSuccessAtElementPath } from '../patch-utils'
+import type { CSSNumber } from '../../../inspector/common/css-utils'
+import { isCSSNumber } from '../../../inspector/common/css-utils'
+import type { StyleInfo, CSSStyleProperty } from '../../canvas-types'
 
 export interface EditorStateWithPatch {
   editorStateWithChanges: EditorState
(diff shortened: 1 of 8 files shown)


editor/src/components/canvas/commands/utils/property-utils.ts
```diff
@@ -99,3 +102,24 @@ export function maybeCssPropertyFromInlineStyle(property: PropertyPath): string
...
+export type GetCSSNumberFromStyleInfoResult =
+  | { type: 'not-found' }
+  | { type: 'not-css-number' }
+  | { type: 'css-number'; number: CSSNumber }
+
+export function getCSSNumberFromStyleInfo(
+  styleInfo: StyleInfo,
+  property: string,
+): GetCSSNumberFromStyleInfoResult {
+  let styleInfoUntyped = styleInfo as any as Record<string, CSSStyleProperty<unknown>>
```
seanparsons, line 115: I believe this can be done without the need for a cast:
```typescript
  if (!isStyleInfoKey(property)) {
    return { type: 'not-found' }
  }

  const prop = styleInfo[property]
  if (prop == null || prop.type === 'not-found') {
    return { type: 'not-found' }
  }

  if (prop.type === 'not-parsable' || !isCSSNumber(prop.value)) {
    return { type: 'not-css-number' }
  }
  return { type: 'css-number', number: prop.value }
```

seanparsons (approved):

gbalint (approved):

editor/src/components/canvas/commands/utils/property-utils.ts
bkrmendy: pushed the code without a cast in https://github.com/concrete-utopia/utopia/pull/6632/commits/886103bbb5390c7abbe3fb454fd8291e706e133e

PR #6634, opened 2024-11-13
Remove unnecessary props from `StyleInfoFactory` context param

bkrmendy:
## Problem
`metadata` and `elementPathTree` are required by the context param of `StylePlugin.styleInfoFactory`, but they weren't used by the plugins. Because of this, having to pass `metadata` and `elementPathTree` in the context param created unnecessary boilerplate

## Fix
Remove `metadata` and `elementPathTree` from the context param of `StylePlugin.styleInfoFactory`, but keep the context param itself so other data can be added to the context later on.


balazsbajorics (approved):
nice, lean!

PR #6640, opened 2024-11-14
Grid component support

gbalint:
**Problem:**
Followup to https://github.com/concrete-utopia/utopia/pull/6636

The goal is to support interaction with grid item in the following scenario:
- the grid is in a component
- the grid renders the component's children
- the grid is not the root element of the component

**Fix:**
In https://github.com/concrete-utopia/utopia/pull/6636 I introduced a way to indentify a grid using the element path of its item (which is necessary because when the grid is in a non-focused component).

- `GridMeasurementHelper` collects the grids from the metadata. These grids in non-focused components are not in the metadata, so they are not collected. I fixed this by collecting component items from the grids too. 
- To get the measurement data for these grids which identified by their items, I added a new parameter to `getGridMeasurementHelperData` so it can get the styling of the parentElement of the item's element.
- The measurement helper was identified by its element path, and the dom-sampler could retrieve the helper from the dom using that. However, this is again not a good solution for these grids in non-focused components, so I added global WeakMap which stores the pairing between HTMLElements of grids and the id attribute of their GridMeasurementHelper.
- I updated the `path` fiels names of `GridContainerIdentifier` and `GridItemIdentifier`: it was not a good idea to use the same name in the two paths, because it was possible to not check the type and use the path field.

**Missing:**
`gridMoveStrategiesExtraCommands` still tries to set grid props on the parent, and when it is a component, those can be absolutely useless. We should detect when that is the case, and not generate these commands.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/controls/grid-controls-for-strategies.tsx
+++ b/editor/src/components/canvas/controls/grid-controls-for-strategies.tsx
@@ -50,6 +50,8 @@ import {
 } from '../../editor/store/editor-state'
 import { findOriginalGrid } from '../canvas-strategies/strategies/grid-helpers'
 
+export const GridMeasurementHelperMap = { current: new WeakMap<HTMLElement, string>() }
+
 export const GridCellTestId = (elementPath: ElementPath) => `grid-cell-${EP.toString(elementPath)}`
 
 function getCellsCount(template: GridAutoOrTemplateBase | null): number {
@@ -94,6 +96,7 @@ export type GridMeasurementHelperData = {
   rowGap: number | null
   columnGap: number | null
   padding: Sides
+  element: HTMLElement
 }
 
 export function getGridMeasurementHelperData(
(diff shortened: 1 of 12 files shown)


editor/src/components/canvas/controls/grid-controls-for-strategies.tsx
```diff
@@ -90,13 +96,17 @@ export type GridMeasurementHelperData = {
   rowGap: number | null
   columnGap: number | null
   padding: Sides
+  element: HTMLElement
```
seanparsons, line 99: Passing around and holding onto DOM elements makes me a little twitchy. It can shift around underneath us as React updates it and we're holding onto a reference to something which could hold onto a lot of memory potentially.

seanparsons (approved):

bkrmendy (approved):

ruggi (approved):

PR #6649, opened 2024-11-15
Handle Non-Offset Parent Grids.

seanparsons:
**Problem:**
Dragging of `position: 'absolute'` elements is broken, when those elements are in a grid which itself does not have `position: 'absolute'` set.

**Cause:**
The grid cell is assumed to be the offset parent of sorts, which is not the case for these elements.

**Fix:**
In this particular situation, the closest offset parent is identified and that is used to calculate the new frame values instead of the frame of the grid cell, as it currently is.

**Commit Details:**
- `runGridMoveAbsolute` now checks to see if the grid provides layouting, if it does not then the closest offset global frame is identified.
- Slightly widened the type of `zeroRectIfNullOrInfinity`.

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/canvas-strategies/strategies/grid-move-absolute.ts
```diff
@@ -215,6 +215,19 @@ function runGridMoveAbsolute(
...
 
+  function getContainingRect(): CanvasRectangle {
+    if (selectedElementMetadata.specialSizeMeasurements.immediateParentProvidesLayout) {
+      const gridCellGlobalFrame = getGlobalFrameOfGridCell(gridCellGlobalFrames, targetRootCell)
+      return zeroRectIfNullOrInfinity(gridCellGlobalFrame)
+    } else {
+      const closestOffsetGlobalFrame = MetadataUtils.findElementByElementPath(
+        jsxMetadata,
+        selectedElementMetadata.specialSizeMeasurements.closestOffsetParentPath,
+      )?.globalFrame
```
balazsbajorics, line 226: `selectedElementMetadata.specialSizeMeasurements.coordinateSystemBounds` should give you this info!

balazsbajorics: we should rename it to offsetParentBounds for readability...

balazsbajorics: furthermore, it also works if the selected element's offsetParent does not have metadata because it is internal to an unfocused component

seanparsons: Fixed in 8b8a5e8a7.

balazsbajorics: I believe you did not push 8b8a5e8a7 !

seanparsons: My mistake, fixed now.

bkrmendy (approved):

PR #6650, opened 2024-11-18
Tailwind-compatible flex section: round 1

bkrmendy:
## Problem
The controls to set `justify-contents`, `align-items` and `flex-direction` don't work with Tailwind

## Fix
- Add the necessary class names to `TailwindClassNameMapping`, so that the Tailwind style plugin can write these classes. These controls read props from `specialSizeMeasurements`, so I'm leaving `StyleInfo` unchanged in this PR, and add the flex-related classes I added to `TailwindClassNameMapping` when something actually needs to read them through `StyleInfo`.
- Add tests for Tailwind editing in the affected controls

### Commit Details
- Add the necessary class names to `TailwindClassNameMapping`
- Add tests for `NineBlockControl` (used to set `justify-contents` and `align-items`)
- Add tests for `ThreebarControl` (used to set `align-items` when `justify-contents` is set to `space-between`)
- Add tests for `SpacedPackedControls` (sets `justify-contents` to either `space-between` for spaced or to `flex-start` for `packed`)
- Add tests for `FlexDirectionToggle` (used to set flex direction)

### Out of scope
There are two other controls in the flex section, used for setting `flex-wrap` and  `flex-gap`. These controls use `useInspectorInfo` under the hood, making that work is left to a follow-up PR

## Manual Tests
I hereby swear that:


seanparsons (approved):
All of the tests!

PR #6655, opened 2024-11-19
Style plugin-aware `useInspectorInfo`

bkrmendy:
## Problem
`useInspectorInfo`, a widely used helper in the inspector, cannot read element styles set by Tailwind

## Fix
The fix is made up from multiple pieces:
- I extended `StylePlugin` with a new function, `readStyleFromElementProps`, which reads a given style info key from the `JSXAttributes` passed to it. I updated both `InlineStylePlugin` and `TailwindStylePlugin` to support this function, and updated all affected tests. This change was originally prototyped on the [Megaspike](https://github.com/concrete-utopia/utopia/pull/6574)
- I updated `useGetMultiselectedProps` to use `StylePlugin.readStyleFromElementProps` when a style prop is read.
- I updated `CSSStylePropertyNotParsable` and `ParsedCSSStyleProperty` to preserve the original value read from `projectContents`, in addition to the parsed representation. This way, `CSSStyleProperties` can be used by code that expect to work with this lower-level representation (such as the internals of `useInspectorInfo`)
- I updated the `SET_PROP` and `UNSET_PROP` actions to use the `setProperty` and `deleteProperty` commands under the hood. This way, any editor code using these actions will be able to use the style plugins to write element style. This change was originally prototyped on the [Megaspike](https://github.com/concrete-utopia/utopia/pull/6574)

### Out of scope
This change only touches `useGetMultiselectedProps` from the internals of `useInspectorInfo`. `useGetSpiedProps` is left unchanged, since the values element style props set through Tailwind don't show up in `allElementProps` (which is the data source for spied props)

## Manual Tests:
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/plugins/inline-style-plugin.ts
+++ b/editor/src/components/canvas/plugins/inline-style-plugin.ts
@@ -9,7 +9,7 @@ import type { ModifiableAttribute } from '../../../core/shared/jsx-attributes'
 import { getJSXElementFromProjectContents } from '../../editor/store/editor-state'
 import { cssParsers, type ParsedCSSProperties } from '../../inspector/common/css-utils'
 import { stylePropPathMappingFn } from '../../inspector/common/property-path-hooks'
-import type { CSSStyleProperty } from '../canvas-types'
+import type { CSSStyleProperty, StyleInfo } from '../canvas-types'
 import {
   cssStyleProperty,
   cssStylePropertyNotParsable,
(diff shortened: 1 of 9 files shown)


gbalint (approved):

editor/src/components/editor/actions/actions.tsx
```diff
@@ -1727,50 +1732,40 @@ export const UPDATE_FNS = {
...
-      (success) => success,
-    )
-    if (unsetPropFailedMessage != null) {
-      const toastAction = showToast(notice(unsetPropFailedMessage, 'ERROR'))
-      return UPDATE_FNS.ADD_TOAST(toastAction, editor)
-    } else {
-      return updatedEditor
-    }
+    return foldAndApplyCommandsSimple(editor, [
+      deleteProperties('always', action.element, [action.property]),
```
ruggi, line 1736: I think this makes sense, however we're losing the toast this way

ruggi (approved):

editor/src/components/canvas/plugins/inline-style-plugin.ts
```diff
@@ -39,18 +39,21 @@ function getPropertyFromInstance<P extends StyleLayoutProp, T = ParsedCSSPropert
...
+    return cssStylePropertyNotParsable(attribute)
   }
-  return cssStyleProperty(parsed.value)
+  return cssStyleProperty(parsed.value, attribute)
 }
 
 export const InlineStylePlugin: StylePlugin = {
   name: 'Inline Style',
+  readStyleFromElementProps: <T extends keyof StyleInfo>(props: JSXAttributes, key: T) => {
+    return getPropertyFromInstance(key, props) as StyleInfo[T]
```
ruggi, line 55: `as StyleInfo[T]` 🧐

bkrmendy: fixed here https://github.com/concrete-utopia/utopia/pull/6655/commits/eb66da96c2d0a4b5c4104f4cb2318a464604cbda

PR #6662, opened 2024-11-20
feat(import): send import results to discord

liady:
This PR adds the ability to send github import results (errors/warnings) to Discord (to the channel `#site-import-webhook`).

**Details:**
1. Added a generic API endpoint for sending messages to Discord from our code. We use `discord-webhook-node` to format the message. The generic logic is here: [utopia-remix/app/util/discordWebhookUtils.ts](https://github.com/concrete-utopia/utopia/pull/6662/files#diff-77fb1f38c2e0b74476d2dac0b094d0c3f56cd598ca46f5ba2f348dd1afd8c4ae)
2. The code in the client that sends the results, here: [editor/src/core/shared/import/import-operation-service.ts](https://github.com/concrete-utopia/utopia/pull/6662/files#diff-426bff74c9da628f136029668cad91728508ef49a539c96a379c8c1e3f70e150R250-R265)
3. The code in the server that builds the textual site import message here: [utopia-remix/app/handlers/discordMessageBuilder.ts](https://github.com/concrete-utopia/utopia/pull/6662/files#diff-d2102489b75734896e5f1e06678c3f0ca926650eb08cdd14e00e79315f947302R22-R80)

No other logic was changed, the rest of the code changes are mainly refactors / types.

The code is generic and we can later [add](https://github.com/concrete-utopia/utopia/pull/6662/files#diff-77fb1f38c2e0b74476d2dac0b094d0c3f56cd598ca46f5ba2f348dd1afd8c4aeR20-R22) more Discord webhooks for other functionalities, easily. 
The webhook url for the specific channel is hidden in an env var - `DISCORD_WEBHOOK_SITE_IMPORT`, it is already set on our staging/production servers, and currently not on local machines.

**Messages Example:**

image

**Manual Tests:**
I hereby swear that:


utopia-remix/app/util/discord-webhook-node.d.ts
```diff
@@ -0,0 +1,69 @@
+declare module 'discord-webhook-node' {
```
liady, line 1: this is required since `discord-webhook-node` does not expose its types in an ESM compliant way

editor/src/components/editor/import-wizard/import-wizard-helpers.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react'
+import {
+  ImportOperationResult,
+  type ImportOperation,
+} from '../../../core/shared/import/import-operation-types'
+import { assertNever } from '../../../core/shared/utils'
+
+export function getImportOperationText(operation: ImportOperation): string {
```
liady, line 8: just moved here to be shared

editor/src/components/editor/server.ts
```diff
@@ -761,3 +767,24 @@ export function getBranchProjectContents(operationContext: GithubOperationContex
     return response.json()
   }
 }
+
+export async function sendDiscordMessage(
+  webhookType: DiscordWebhookType,
+  messageType: DiscordMessageType,
+  message: DiscordMessage,
+) {
+  const body: DiscordWebhookBody = { webhookType, messageType, message }
```
Rheeseyb, line 776: We should be building the message on the back end as much as possible

liady: I wanted to keep the endpoint as generic as possible (so we could use it for other channels and notifications as well) - so in the client I only build the "site import" specific information, and the generic things are in the server.
(for example the user id is being added in the backend and not in the client).
do you think there is additional information that should be moved to the server? (note that everything in the client is very specific to the site import process - https://github.com/concrete-utopia/utopia/pull/6662/files#diff-426bff74c9da628f136029668cad91728508ef49a539c96a379c8c1e3f70e150R208-R282)

liady: refactored it so that the message is being built on the server:
https://github.com/concrete-utopia/utopia/pull/6662/files#diff-d2102489b75734896e5f1e06678c3f0ca926650eb08cdd14e00e79315f947302R22

and the client just sends what it needs to:
https://github.com/concrete-utopia/utopia/pull/6662/files#diff-d2102489b75734896e5f1e06678c3f0ca926650eb08cdd14e00e79315f947302R22-R80

utopia-remix/app/util/discord-webhook-node.d.ts
Rheeseyb: That's disappointing

utopia-remix/package.json
```diff
@@ -93,6 +94,7 @@
     "eslint-plugin-prettier": "3.1.2",
     "eslint-plugin-react": "7.33.2",
     "eslint-plugin-react-hooks": "4.6.0",
+    "eslint-plugin-react-refresh": "0.4.12",
```
Rheeseyb, line 97: Interesting. I'm guessing this wasn't intentionally committed since it hasn't been added to our eslint config? It looks like it would be useful to us, so worth bringing up on Discord (unless you already did and I missed it!)

Rheeseyb (approved):

utopia-remix/package.json
liady: so this is the plugin that we discussed about a month ago that prevents accidentally breaking the HMR. Balazs was adding it to the eslint config and to the editor, but we needed to install it in the `utopia-remix` as well, otherwise the eslint didn't work at all.

editor/src/components/editor/actions/actions.tsx
```diff
@@ -2191,12 +2194,14 @@ export const UPDATE_FNS = {
     }
   },
   UPDATE_IMPORT_STATUS: (action: UpdateImportStatus, editor: EditorModel): EditorModel => {
+    const newImportState = {
+      ...editor.importState,
+      importStatus: action.importStatus,
+    }
+    notifyImportStatusToDiscord(newImportState, editor.projectName)
```
balazsbajorics, line 2201:
```suggestion
    // side effect ☢️
    notifyImportStatusToDiscord(newImportState, editor.projectName)
```

balazsbajorics (approved):

utopia-remix/app/util/discord-webhook-node.d.ts
liady: yes, my initial thought was forking it and fixing its package.json (it's quite simple), but it seemed like an overkill for it

PR #6667, opened 2024-11-20
Grid Element Direct Lookup Ordered Correctly.

seanparsons:
**Problem:**
Grid children should have their properties and dimensions retrieved directly the way we've done for grids.

**Fix:**
Some additional supporting functions similar to those implemented for grids were implemented for the children of grids. Those were implemented into the controls that still referred to the metadata for the dimensions and grid properties of those elements.

In the process of testing the above work, it was discovered that the flow of data wasn't quite what was expected. With the DOM sampler running before the helper controls were re-rendered, which resulted in some values suffering from the 1-frame behind problem. To solve that an additional store was added just for the helper controls so that now the following ordering is observed:
- Update changes the model.
- The canvas is rendered.
- The helper controls are rendered.
- The DOM sampler runs.
- Everything else in the editor is rendered.

**Commit Details:**
- Moved a lot of common types and functions used when obtaining values from elements for grids into `grid-measurements.tsx`.
- Added `getGridElementMeasurementHelperData` to retrieve the necessary details for a grid element, along with some supporting functions.
- Extracted out a `getGlobalFrame` function for the common lookup used.
- Changed some test ID values in two components as they created the same ones.
- `PaddingResizeControl` set to the `bottom` priority so that it doesn't end up on top of the track resizing controls.
- `GridControl` now gets the cell dimensions directly from the grid element.
- `GridElementContainingBlock` gets the grid child dimensions directly.
- Added `HelperControlsStateContext` and made the connected change to `StoreName`.
- Added `helperControlsStore` to `Editor`.
- Added contexts for the helper store to `EditorRoot` and `HotRoot`.
- Wrapped the contents of `GridElementContainingBlocks` and `GridMeasurementHelpers` with the new helper store context.
- Added calls to update the helper store in the regular and test dispatch flows.

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -1064,22 +1069,24 @@ export const GridMeasurementHelpers = React.memo(() => {
...
-      ))}
-      {gridItems.map((gridItem) => (
-        <GridMeasurementHelper
-          key={GridMeasurementHelpersKey(gridItem)}
-          elementPath={gridItem}
-          source='parent'
-        />
-      ))}
-    </CanvasOffsetWrapper>
+    <EditorStateContext.Provider value={helperControlsStore == null ? null : helperControlsStore}>
```
ruggi, line 1072:
```suggestion
    <EditorStateContext.Provider value={helperControlsStore}>
```
maybe?

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -2096,17 +2069,19 @@ export const GridElementContainingBlocks = React.memo<GridElementContainingBlock
...
-        return (
-          <GridElementContainingBlock
-            key={GridElementContainingBlockKey(selectedGridChildElement)}
-            gridPath={EP.parentPath(selectedGridChildElement)}
-            gridChild={selectedGridChildElement}
-          />
-        )
-      })}
-    </CanvasOffsetWrapper>
+    <EditorStateContext.Provider value={helperControlsStore == null ? null : helperControlsStore}>
```
ruggi, line 2072: also here ✋

ruggi (approved):

editor/src/components/canvas/controls/grid-controls.tsx
seanparsons: Fixed in ed216aa53.

editor/src/components/canvas/controls/grid-controls.tsx
seanparsons: Fixed in ed216aa53.

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -1064,22 +1069,24 @@ export const GridMeasurementHelpers = React.memo(() => {
...
-      ))}
-      {gridItems.map((gridItem) => (
-        <GridMeasurementHelper
-          key={GridMeasurementHelpersKey(gridItem)}
-          elementPath={gridItem}
-          source='parent'
-        />
-      ))}
-    </CanvasOffsetWrapper>
+    <EditorStateContext.Provider value={helperControlsStore}>
```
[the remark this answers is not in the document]
balazsbajorics: generally, I like to place the store provider around a component which takes no props, that way we avoid mixing props from different stores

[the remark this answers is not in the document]
seanparsons: Refactor in e27933034.

balazsbajorics (approved):

PR #6678, opened 2024-11-25
Grid Markers Mouse Down Behaviour

seanparsons:
**Problem:**
When mouse down on one of the grid markers:
- The ruler’s white background and .2px black outline becomes visible.
- The black ruler marks for other rows/columns become visible.
- The grid lines become visible.

**Fix:**
Primarily the markers now are children of a newly added `GridRuler` component, so that can manage some of the positioning and they can be visually placed within it.

Otherwise there were mostly ancillary changes related to now needing to be able to scale the markers instead of relying on `zoom` as it was causing trouble with the layout. Along with adding the logic to build up all the extra markers which are displayed when the mouse down is triggered.

**Commit Details:**
- Ruler markers now take a colour and scale directly.
- Added a short vertical and horizontal pipe marker.
- Grid outlines shown when triggered from the grid markers.
- `RulerMarkers` now takes a callback to trigger the showing of grid markers.
- Additional properties added to `RulerMarkerData` and `RulerMarkerPositionData`.
- The additional markers are built up using `addOtherMarker`.
- When the row and column grid markers are mouse downed, their respective ruler and additional markers are shown.
- A `GridRuler` component was added and the `RulerMarkerIndicator` components respective to their axis are now children of those rulers.
- `RulerMarkerIndicator` components were reworked to cater for their new functionality and also that they are now within a `GridRuler` component.
- Added `getRulerMarkerColor` utility function.
- Deleted `skewMarkerPosition` as now all of the markers are a fixed size.
- Removed `priority` setting in `controlsForGridPlaceholders` so the grid
  controls sit above other controls.
- Introduced `MarkerSVG` component to handle scaling of the SVGs.
- Put the selected ruler markers on top of the other ruler markers.

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/controls/grid-controls-ruler-markers.tsx
```diff
@@ -1,82 +1,116 @@
...
-  <svg width='11' height='11' xmlns='http://www.w3.org/2000/svg'>
-    <polygon
-      points='5,1 10,10 0,10'
-      fill={colorTheme.primary.value}
-      stroke={colorTheme.white.value}
-      strokeWidth='0.5'
-    />
-  </svg>
-)
+function upFacingTriangle(fillColor: UtopiColor, scale: number): React.ReactNode {
```
liady, line 6: small nitpick - we can have a component:
```tsx
function MarkerSVG({ scale, children }: MarkerSVGProps) {
  return (
    <svg
      width={`${11 / scale}`}
      height={`${11 / scale}`}
      viewBox='0 0 11 11'
      xmlns='http://www.w3.org/2000/svg'
    >
      {children}
    </svg>
  )
}
```
to be reused instead of all of the svgs:
```tsx
function upFacingTriangle(fillColor: UtopiColor, scale: number): React.ReactNode {
  return (
    <MarkerSVG scale={scale}>
      <polygon points='5.5,0 11,11 0,11' fill={fillColor.value} />
    </MarkerSVG>
  )
}

function rightFacingTriangle(fillColor: UtopiColor, scale: number): React.ReactNode {
  return (
    <MarkerSVG scale={scale}>
      <polygon points='11,5.5 0,0 0,11' fill={fillColor.value} />
    </MarkerSVG>
  )
}

function downFacingTriangle(fillColor: UtopiColor, scale: number): React.ReactNode {
  return (
    <MarkerSVG scale={scale}>
      <polygon points='5.5,11 0,0 11,0' fill={fillColor.value} />
    </MarkerSVG>
  )
}

function leftFacingTriangle(fillColor: UtopiColor, scale: number): React.ReactNode {
  return (
    <MarkerSVG scale={scale}>
      <polygon points='0,5.5 11,0 11,11' fill={fillColor.value} />
    </MarkerSVG>
  )
}

function regularVerticalPipe(fillColor: UtopiColor, scale: number): React.ReactNode {
  return (
    <MarkerSVG scale={scale}>
      <rect x='4' y='0' width='3' height='11' fill={fillColor.value} />
    </MarkerSVG>
  )
}

function regularHorizontalPipe(fillColor: UtopiColor, scale: number): React.ReactNode {
  return (
    <MarkerSVG scale={scale}>
      <rect x='0' y='4' width='11' height='3' fill={fillColor.value} />
    </MarkerSVG>
  )
}

function shortVerticalPipe(fillColor: UtopiColor, scale: number): React.ReactNode {
  return (
    <MarkerSVG scale={scale}>
      <rect x='4' y='2' width='3' height='9' fill={fillColor.value} />
    </MarkerSVG>
  )
}

function shortHorizontalPipe(fillColor: UtopiColor, scale: number): React.ReactNode {
  return (
    <MarkerSVG scale={scale}>
      <rect x='2' y='4' width='9' height='3' fill={fillColor.value} />
    </MarkerSVG>
  )
}
```

(but that's definitely just an optimization)

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -2157,29 +2168,75 @@ function useSelectedGridItems(): ElementPath[] {
   )
 }
 
-const rulerMarkerIconSize = 12 // px
+const rulerMarkerIconSize = 11 // px
```
ruggi, line 2171: this is something tangentially related, but maybe it'd be good to export this and use it in the SVG definitions (e.g. `${11 / scale}`.

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -2254,128 +2300,344 @@ const RulerMarkers = React.memo((props: { path: ElementPath }) => {
         height: height,
       })
 
+      let otherColumnMarkers: Array<RulerMarkerPositionData> = []
```
ruggi, line 2303: I think this would become a lot simpler if you would just collect all the notch positions independently as canvas points. For example for the horizontal axis, you would just need to grab all the columns from the grid frames (`parentGridCellGlobalFrames[0]`), and concat them with one extra entry for the final notch with `x: lastCol.x + lastCol.width`. Then, you can just:
```ts
notches.map(notch => <div key={...} style={{ x: notch.x, y: grid.y - notchHeight, ... }} />)
```
(if you don't want to display the triangle marker on top of the notch you can just filter it out, but I believe the ruler should display _all_ the notches corresponding to the grid lines)

seanparsons: Ah yes, I did start thinking about this, but then because of the grid gaps it didn't seem to make sense at the time.

seanparsons: It also needed to have the selected start and end values called out specifically as some of the existing logic relies on those.

editor/src/components/canvas/controls/grid-controls.tsx
seanparsons: Fixed in https://github.com/concrete-utopia/utopia/pull/6678/commits/6425c8b185eac8790ce7e6277e16d2380609ccb2.

editor/src/components/canvas/controls/grid-controls-ruler-markers.tsx
seanparsons: Fixed in https://github.com/concrete-utopia/utopia/pull/6678/commits/6425c8b185eac8790ce7e6277e16d2380609ccb2.

ruggi (approved):
I think this is ok, however the "duplicate" notches for grids with gaps are not necessarily what we want, but imho we can tweak down the line.

PR #6684, opened 2024-11-26
feat(css): change media queries to container queries

liady:
This PR changes the loaded project CSS to convert media queries to container queries.

Note: almost all of the changes are snapshots, these are the important ones:
- the adjustment code in [css-utils.ts](https://github.com/concrete-utopia/utopia/pull/6684/files#diff-20815c549e15a5fdfab5e15884455d2647c2d2ee5bac4002d740960e6070677aR55-R97) 
- the tests added to [css-utils.spec.ts](https://github.com/concrete-utopia/utopia/pull/6684/files#diff-a6f48747618684f57b1fa5c924b893cb2c1cc399d6ffacac772dbd3018369656R203-R270)
- a small adjustment to the color parser in [css-parser-utils.ts](https://github.com/concrete-utopia/utopia/pull/6684/files#diff-7e1f4da9bbdd89ac2c178aeadca1037b8ffd70d81f2ce906bf1bdac0382eed87R344-R348 )

Details:
- Upgrading to `css-tree@3`, we can now parse container queries. 
- This PR adds a check to the AST walk that identifies `@media` queries that include `width` condition, and converts only the `width` related conditions to a `@container` query (nested inside the `@media` query, if necessary). The relevant code is [here](https://github.com/concrete-utopia/utopia/pull/6684/files#diff-20815c549e15a5fdfab5e15884455d2647c2d2ee5bac4002d740960e6070677aR55-R97).
- We also add the container name `scene` to the `Scene/RemixScene` elements, to be treated as the container boundary.

Note: as part of the upgrade to `css-tree@3` the color parsing slightly changed to match `CSS Color Level 5` [spec](https://drafts.csswg.org/css-color-5/#color-syntax), so the lexed result is wrapped in `color-base`, there was a small fix for it [added](https://github.com/concrete-utopia/utopia/pull/6684/files#diff-7e1f4da9bbdd89ac2c178aeadca1037b8ffd70d81f2ce906bf1bdac0382eed87R344-R348) to the PR.

<video src="https://github.com/user-attachments/assets/91d9bca1-4a64-4cba-8a02-29d65160d974"></video>

**Manual Tests:**
I hereby swear that:


editor/src/components/canvas/ui-jsx-canvas-renderer/remix-scene-component.tsx
```diff
@@ -24,6 +24,9 @@ export const RemixSceneComponent = React.memo((props: React.PropsWithChildren<Re
   const { style, getLoadContext, startingRoute, ...remainingProps } = props
 
   const sceneStyle: React.CSSProperties = {
+    // eslint-disable-next-line
+    // @ts-ignore
+    container: `${SceneContainerName} / inline-size`,
```
bkrmendy, line 29: why is the `@ts-ignore` needed here? If it's because of the `React.CSSProperties` type, I'd rather extend the type:
```typescript
type CSSPropertiesWithContainer = React.CSSProperties & { container: string }
```
and use that instead of `React.CSSProperties`

Rheeseyb (approved):

editor/src/components/canvas/ui-jsx-canvas-renderer/remix-scene-component.tsx
Rheeseyb: Yeah, this please

liady: yeah, React 19 already supports it, but it's a good idea to monkey patch it for now.
probably 
```
type CSSPropertiesWithContainer = { container: string } & React.CSSProperties
```
to not override if we'll upgrade and forget to change it back.
will do!

liady: ok, after some digging, turns out we had `csstype` locked to `3.0.3`. simply [upgrading](https://github.com/concrete-utopia/utopia/pull/6684/commits/7b6478c158512768581676b9489c675727d53b56) it to `3.1.3` did the trick, without the need for the React upgrade (that uses it).

bkrmendy (approved):

PR #6685, opened 2024-11-26
Respect spans when resizing grid cells

### LINKED ISSUE #6683: Respect spans when resizing grid items

ruggi:
**Problem:**

Resizing grid cells always forces explicit numerical placement pins, chomping any `span`s that may have been defined on the cell.

**Fix:**

After determining the right numerical positions for the resized cell bounds, do a normalization pass so that the new grid props are rewritten to respect the new bounds but also to express them with `spans` if the original pins were spans in the first place.

For example, assuming enlarging to the right by 1 cell:

| Initial | Result |
|--------|----------|
| `gridColumn: span 2` | `gridColumn: span 3` |
| `gridColumn 3 / span 2` | `gridColumn: 3 / span 3` |

**Manual Tests:**
I hereby swear that:

Fixes #6683

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/grid-resize-element-strategy.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/grid-resize-element-strategy.ts
@@ -1,11 +1,18 @@
 import { MetadataUtils } from '../../../../core/model/element-metadata-utils'
 import * as EP from '../../../../core/shared/element-path'
+import type {
+  GridElementProperties,
+  GridPositionOrSpan,
+  GridPositionValue,
+} from '../../../../core/shared/element-template'
+import { gridSpanNumeric, isGridSpan } from '../../../../core/shared/element-template'
 import {
   type CanvasRectangle,
   isInfinityRectangle,
   rectangleIntersection,
 } from '../../../../core/shared/math-utils'
 import { gridContainerIdentifier, gridItemIdentifier } from '../../../editor/store/editor-state'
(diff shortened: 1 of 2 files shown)


editor/src/components/canvas/canvas-strategies/strategies/grid-resize-element-strategy.ts
```diff
@@ -104,15 +111,60 @@ export const gridResizeElementStrategy: CanvasStrategyFactory = (
...
         return emptyStrategyApplicationResult
       }
 
       const gridTemplate =
         selectedElementMetadata.specialSizeMeasurements.parentContainerGridProperties
 
+      const elementGridPropertiesFromProps =
+        selectedElementMetadata.specialSizeMeasurements.elementGridPropertiesFromProps
+
+      const width =
```
gbalint, line 126: I don't like the term `width`/`height` because I'm always mistaken that it is a geometrical width/height (in pixels). Is there a better name for this? Maybe `columnCount` and `rowCount`

gbalint (approved):

editor/src/components/canvas/canvas-strategies/strategies/grid-resize-element-strategy.ts
ruggi: makes total sense, fixed in [`6be5569` (#6685)](https://github.com/concrete-utopia/utopia/pull/6685/commits/6be5569b638bf408abf840f96f655e53a785ddd4)

PR #6689, opened 2024-11-29
Cell resize via ruler markers

ruggi:
**Problem:**

It should be possible to resize grid cells using the ruler markers.

**Fix:**

When dragging a ruler marker:
1. start a new interaction for the new resize strategy variant which detects the closest cell on a given axis
2. freeze the existing markers so they don't move until the interaction is complete
3. show a pink snap line on the interaction target, with a label on top indicating the pin value
4. show a live offset line following the drag position

**Notes:**
- The snap line positioning might be tweaked in the future based on what we decide to do with "duplicate" ruler notches.
- The resize is only enabled for stretching grid items, same as we do with the regular resize. We should (probably?) incrementally add the ability to resize cells even if their child is not stretching when using the rulers.

**Manual Tests:**
I hereby swear that:

- [x] I opened a hydrogen project and it loaded
- [x] I could navigate to various routes in Play mode

Fixes #6688

PR #6692, opened 2024-12-03
Grid ruler resize labels sticky

ruggi:
**Problem:**

Markers should not move while resize is in progress, and the active one's label should be visible during resize.

**Fix:**

As a followup to https://github.com/concrete-utopia/utopia/pull/6689, this PR makes sure that markers are fixed when the ruler resize strategy is active, and that the label for the marker being dragged stays visible as well.

### DIFF
--- a/editor/src/components/canvas/controls/grid-controls.tsx
+++ b/editor/src/components/canvas/controls/grid-controls.tsx
@@ -2529,7 +2529,7 @@ const RulerMarkers = React.memo((props: RulerMarkersProps) => {
         <RulerMarkerIndicator
           gridRect={rulerMarkerData.gridRect}
           parentGrid={rulerMarkerData.parentGrid}
-          marker={rulerMarkerData.columnStart}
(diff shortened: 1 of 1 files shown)


liady (approved):

seanparsons (approved):

editor/src/components/canvas/controls/grid-controls.tsx
```diff
@@ -2859,6 +2859,21 @@ const RulerMarkerIndicator = React.memo(
 
     const labelClass = 'ruler-marker-label'
 
+    const edge: GridResizeEdge =
```
seanparsons, line 2862: Could this be made into: `${props.axis}-${props.marker.bound}`?

ruggi: oh, very nice idea! I did not even know TS would allow that without complaining about types! ✅

bkrmendy (approved):

PR #6696, opened 2024-12-04
Ruler marker resize for non-stretching grid items

### LINKED ISSUE #6695: Grid ruler marker resize for fixed-size elements

ruggi:
**Problem:**

It should be possible to resize non-stretching grid items via the ruler markers.

**Fix:**

Because of how grids are rendered, similarly to what happened in https://github.com/concrete-utopia/utopia/pull/6691, we need to calculate the rendered dimensions of grid _cells_ (not their contained items!) after they are rendered, since for non-stretching grid items their dimensions (via `getGridChildCellCoordBoundsFromCanvas`) don't match the ones of the cell(s) they are configured to target.

This PR uses the logic introduced in https://github.com/concrete-utopia/utopia/pull/6691 to calculate the size of a grid cell starting from its item, and uses that to correctly place the ruler markers for non-stretching items.

**Manual Tests:**
I hereby swear that:

Fixes #6695

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
@@ -752,6 +752,9 @@ export function getGridRelativeContainingBlock(
   gridMetadata: ElementInstanceMetadata,
   cellMetadata: ElementInstanceMetadata,
   cellProperties: GridElementProperties,
+  options?: {
+    forcePositionRelative?: boolean
+  },
 ): LocalRectangle {
   const gridProperties = gridMetadata.specialSizeMeasurements.containerGridProperties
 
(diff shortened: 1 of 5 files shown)


bkrmendy (approved):

editor/src/components/canvas/canvas-strategies/strategies/grid-helpers.ts
```diff
@@ -779,17 +778,19 @@ export function getGridRelativeContainingBlock(
...
   // Gap needs to be set only if the other two are not present or we'll have rendering issues
   // due to how measurements are calculated.
-  if (
-    gridMetadata.specialSizeMeasurements.rowGap != null &&
-    gridMetadata.specialSizeMeasurements.columnGap != null
-  ) {
-    const gap = gridMetadata.specialSizeMeasurements.gap
-    gridElement.style.gap = gap == null ? 'initial' : `${gap}px`
-  } else {
+  const hasRowGap = gridMetadata.specialSizeMeasurements.rowGap != null
```
ruggi, line 781: this is a fix tangential to this PR, the problem is that `gap` itself might be null even if `rowGap` and `columnGap` are defined - the example is a configuration with uneven row/col gaps.

seanparsons (approved):

PR #6702, opened 2024-12-05
Check Element Against Containing Block

seanparsons:
**Problem:**
If an element is dragged outside of the current cell it occupies but not the containing block, then we will move the element to the new cell and calculate the pins for that new position. But this means we are often unnecessarily changing those values which the user may have set specifically.

**Fix:**
Now we check first if the element is outside the containing block by checking that it is fully inside the containing block, if it is inside the containing block then no changes are made to the grid position.

**Commit Details:**
- Added `rectangleContainsRectangleInclusive` utility function.
- Added `getNewGridElementPropsCheckingOriginalGrid` as a wrapper around `getNewGridElementProps` which checks if the element is outside of the containing block anywhere.

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/canvas-strategies/strategies/grid-move-absolute.ts
+++ b/editor/src/components/canvas/canvas-strategies/strategies/grid-move-absolute.ts
@@ -8,7 +8,7 @@ import {
   type ElementInstanceMetadataMap,
   type GridContainerProperties,
 } from '../../../../core/shared/element-template'
-import type { CanvasVector } from '../../../../core/shared/math-utils'
+import type { CanvasRectangle, CanvasVector } from '../../../../core/shared/math-utils'
 import {
   canvasRectangle,
   canvasRectangleToLocalRectangle,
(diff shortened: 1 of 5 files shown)


bkrmendy (approved):

editor/src/components/canvas/canvas-strategies/strategies/grid-move-absolute.ts
```diff
@@ -285,6 +335,19 @@ function runGridMoveAbsolute(
     return []
   }
 
+  const coordinateSystemBounds =
```
ruggi, line 338: Unless I'm mistaken I think this could be passed to the `getNewGridElementPropsCheckingOriginalGrid` function here, and then it would also make sense to check for it to be null here at line `340` instead of `?? zeroCanvasRect` since `getNewGridElementPropsCheckingOriginalGrid` would return `null` anyways and do nothing on line `370`.

ruggi (approved):

editor/src/components/canvas/canvas-strategies/strategies/grid-move-absolute.ts
seanparsons: Yeah, that's cleaner, change made in b2004f0b8.

PR #6723, opened 2024-12-12
feat(editor): show correct responsive value according to breakpoint

liady:
This PR adds the ability to display the correct value in the controller according to the Scene size.
For example - if the element has `className='pt-[20px] lg:pt-[150px] md:pt-[110px]`, and the Scene has a width of `1000px` (which is larger than `md` but smaller than `lg`) - the `paddingTop` control will correctly show `110px`.

**Details:**
- This PR augments `ParsedCSSStyleProperty`, so instead of containing just the `value` it now holds:
  - a list of `variants` (the possible values and for each one the `modifier`s that when applied, the value is chosen. a screen size is a modifier)
  - the `currentVariant` - which is the value that is currently selected according to the Scene size.
- After parsing the Tailwind classes using the 3rd-party parser (as before), we call `getModifiers`, that converts the Tailwind specific variants from the parser representation that looks like:
```ts
{type: 'media', value: 'sm'}
``` 
to our generic representation:
```ts
{
  type: 'media-size',
  size: {
    min: { value: 0, unit: 'px' },
    max: { value: 100, unit: 'em' }
  }
}
```
- The `getModifiers` function uses an interal `screensConfigToScreenSizes` function - that parses the Tailwind config to have a map of `<screenAlias>` to `ScreenSize`
- Both are covered in tests in `tailwind-responsive-utils.spec.ts`
- This PR uses the utils that were merged in #6716 (specifically `selectValueByBreakpoint` to select the best matching variant)

**Example (note that this PR *does not* contain the Scene resize buttons):**
<video src="https://github.com/user-attachments/assets/90bb37ee-9aae-4563-9ddc-57a869a69ad5"></video>

**Manual Tests:**
I hereby swear that:

### DIFF
--- a/editor/src/components/canvas/plugins/tailwind-style-plugin-utils/tailwind-responsive-utils.spec.ts
+++ b/editor/src/components/canvas/plugins/tailwind-style-plugin-utils/tailwind-responsive-utils.spec.ts
@@ -48,6 +48,40 @@ describe('getModifiers', () => {
     ])
   })
 
+  it('handles custom screen sizes from config with mixed units', () => {
+    const variants = [
+      { type: 'media', value: 'custom' },
+      { type: 'media', value: 'custom2' },
+    ]
+    const config = {
+      theme: {
+        screens: {
+          custom: { min: '10px', max: '20em' },
+          custom2: '30vh',
+        },
+      },
+    } as unknown as Config
+
+    const result = getModifiers(variants, config)
+    expect(result).toEqual([
+      {
+        type: 'media-size',
+        size: {
+          min: { value: 10, unit: 'px' },
+          max: { value: 20, unit: 'em' },
+        },
+        modifierOrigin: { type: 'tailwind', variant: 'custom' },
+      },
+      {
+        type: 'media-size',
+        size: {
+          min: { value: 30, unit: 'vh' },
+        },
+        modifierOrigin: { type: 'tailwind', variant: 'custom2' },
+      },
+    ])
+  })
+
   it('handles min-max range screen sizes', () => {
     const variants = [{ type: 'media', value: 'tablet' }]
     const config = {
--- a/editor/src/components/canvas/plugins/tailwind-style-plugin-utils/tailwind-responsive-utils.spec.ts
+++ b/editor/src/components/canvas/plugins/tailwind-style-plugin-utils/tailwind-responsive-utils.spec.ts
@@ -1,0 +1,229 @@
+import type { Config } from 'tailwindcss/types/config'
+import { getModifiers, screensConfigToScreenSizes } from './tailwind-responsive-utils'
+
+describe('getModifiers', () => {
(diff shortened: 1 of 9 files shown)


editor/src/components/canvas/plugins/tailwind-style-plugin-utils/tailwind-responsive-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+                .filter((s) => s != '')
+                .join(' and ')}`
+
+        const screenSize = extractScreenSizeFromCss(mediaString)
+        if (screenSize == null) {
+          return null
+        }
+        return [key, screenSize]
+      })
+      .filter((entry): entry is [string, ScreenSize] => entry != null),
```
balazsbajorics, line 47: [nitpick] we have `mapDropNulls` to skip iterating over the array twice

editor/src/components/canvas/plugins/tailwind-style-plugin-utils/tailwind-responsive-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+          typeof size === 'string'
+            ? `@media (min-width: ${size})`
+            : `@media ${[
+                size.min != null ? `(min-width: ${size.min})` : '',
+                size.max != null ? `(max-width: ${size.max})` : '',
+              ]
+                .filter((s) => s != '')
+                .join(' and ')}`
+
+        const screenSize = extractScreenSizeFromCss(mediaString)
```
balazsbajorics, line 41: `screenSizes` already comes in the form of min or {min, max}. If I understand this correctly,  we create a proper media query string from it, and then `extractScreenSizeFromCss` strips the min and max and returns it to us. Am I missing some nuance here? feels like unnecessary double work

editor/src/components/canvas/plugins/tailwind-style-plugin-utils/tailwind-responsive-utils.spec.ts
```diff
@@ -0,0 +1,229 @@
+import type { Config } from 'tailwindcss/types/config'
+import { getModifiers, screensConfigToScreenSizes } from './tailwind-responsive-utils'
+
+describe('getModifiers', () => {
```
balazsbajorics, line 4: can you add a test which uses a non-px unit?

editor/src/components/canvas/plugins/tailwind-style-plugin-utils/tailwind-responsive-utils.ts
liady: It's made intentionally to decouple the Tailwind specific code from the generic media query parsing code. This two-steps process allows us to use the generic part for any media query parsing (even if it didn't come from Tailwind), and also be tested separately and self-contained

bkrmendy (approved):

